PR #13, opened 2016-01-03
Transformime

rgbkrk:
Originally I was going to use nteract/transformime as is, but realized it was a strange beast in a React land. Now using https://github.com/nteract/transformime-react to create React Components for all the display data that Jupyter has.

screenshot 2016-01-02 22 15 13


rgbkrk:
Markdown isn't rendered above, though that's not a big deal here.

PR #20, opened 2016-01-05
Console text

rgbkrk:
This converts ANSI colored console text into HTML as part of a React component.

Eventually, we'll want to replace ansi-to-html with our own ansi-to-react, creating React components instead of HTML strings and relying on `dangerouslySetInnerHTML`.


rgbkrk:
With this last one, I think it's time to extract the display area out into a well tested React component.

PR #30, opened 2016-01-06
Bump electron compile, make babelrc

rgbkrk:
Ran into issues, as noted in https://github.com/jupyter/atom-notebook/issues/26#issuecomment-169490529. Ended up moving the `babelrc` out and upgrading to the recent release of electron-compile.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,10 @@
+  {
+    "presets": [
+      "es2015",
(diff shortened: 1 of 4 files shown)


rgbkrk:
Verified this is good to go with a fresh install and can be run with either `electron .` or `npm run start`.

PR #38, opened 2016-01-08
Use commutable for creating the immutable notebook

rgbkrk:
I've created a package to help with operations on the notebook, called `commutable`. It creates an Immutable.js structure out of a raw notebook object. It also turns all the "multi-line" strings into just strings for use in-memory. That's standard practice amongst Jupyter as on disk they're arrays of strings for source, outputs, etc. to make git and the notebook easier to work with.


rgbkrk:
While I was at it, moved [`commutable` to nteract](https://github.com/nteract/commutable).

PR #43, opened 2016-01-10
Rely on react-jupyter-display-area

rgbkrk:
Extracted Display.jsx and related components out of composition then started relying on them here.


rgbkrk:
Went ahead and cached both node_modules and electron-prebuilt to speed up builds.

PR #47, opened 2016-01-16
[WIP] Notebook pass cell

rgbkrk:
As a cell changes, update the notebook model.

Don't merge me, I was written in poor taste late at night.


rgbkrk:
Since we're going to be relying on Rx for kernel messaging, it's ok if we scrap this and expose the right parts there.

PR #49, opened 2016-01-18
Notebook rx initial

kenwheeler:


rgbkrk:
:shipit: Thanks @kenwheeler!

PR #54, opened 2016-01-22
code of conduct!

rgbkrk:
Added in a code of conduct


rgbkrk:
Had no idea what email to use with TODO group's, used my own. If anyone else in the org would like to be added in as a potential contact, please add yourself in a follow up PR!

PR #57, opened 2016-01-23
[WIP] Provide the LAUNCH_KERNEL action.

rgbkrk:
Woo, execution is really close now. This launches a kernel based on the kernel specified in the notebook. This provides the core channels for communicating with the backend runtime - shell, control, iopub, and stdin. Plenty to do here as we'll have to figure out data flow with this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,47 @@ node_js:
 before_script:
   - export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start
 env:
-  - CXX=g++-4.8
+  - ZMQ="4.1.3" SODIUM="1.0.5"
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
-      - libzmq3-dev
     packages:
       - g++-4.8
+      - gcc-4.8
+      - uuid-dev
+
+before_install:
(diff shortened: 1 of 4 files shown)


rgbkrk:
I knew as soon as I introduced `zmq` into composition I would want `zmq-prebuilt` ready for us...

rgbkrk:
Alright, @kenwheeler and @knowbody let me know what you think here.

rgbkrk:
You know, caching the built zmq in Travis may be ok...

rgbkrk:
I must be missing some fundamental Hot and Cold Observables stuff here. When passing these new channel subjects around, if I subscribe on them and use `onNext` nothing. Yet if I assign them to `window` while in debug mode, later when I rely on them in other code they just work

rgbkrk:
The channels get created over here by the way:

https://github.com/nteract/enchannel-zmq-backend/blob/master/src/subjection.js

Also noting that Ben Lesh had some feedback here (about the same code): http://stackoverflow.com/questions/34845778/create-an-rx-subject-using-subject-create-that-allows-onnext-without-subscriptio/34862286#34862286

rgbkrk:
Closing this to implement it again with the current arch.

PR #58, opened 2016-01-25
Bring on RxJS 5.

rgbkrk:
This switches us over to RxJS 5.

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -1,7 +1,7 @@
 import fs from 'fs';
 
-export function getJSON(filepath) {
-  return new Promise((resolve, reject) => {
+export const getJSON = filepath =>
+  new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


src/api/index.js
```diff
@@ -1,7 +1,7 @@
 import fs from 'fs';
 
-export const getJSON = filepath =>
-  new Promise((resolve, reject) => {
+export function getJSON(filepath) {
```
kenwheeler: assuming this is just for claritys sake?

rgbkrk: Whoop, I actually meant to ask you about these. While I was trying to figure out that thrown `Header` issue, I was making these explicit so I could read tracebacks more easily while injecting some `debugger` statements neatly.

What does declaring these as anonymous functions provide when they're exported? It seemed like this was a clean way to prevent monkey patching by declaring this as a `const`, though I wasn't sure what it was for.

PR #68, opened 2016-01-30
Problem: We don't state what our current condition is

rgbkrk:
Solution: Let everyone know it's not working all the way, link to issues.


jdfreder:
Begin discussion

PR #69, opened 2016-01-30
Kernel action

rgbkrk:
Problem: We weren't launching kernels before

Solution: Provide an API for launching a kernel as well as the packages that help us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,47 @@ node_js:
 before_script:
   - export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start
 env:
-  - CXX=g++-4.8
+  - ZMQ="4.1.3" SODIUM="1.0.5"
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
-      - libzmq3-dev
     packages:
       - g++-4.8
+      - gcc-4.8
+      - uuid-dev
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -4,14 +4,47 @@ node_js:
...
+  - '[ -z "$SODIUM" ] || make'
+  - '[ -z "$SODIUM" ] || make install'
+  - '[ -z "$SODIUM" ] || cd ..'
+  - '[ -z "$SODIUM" ] || export LIBS=-lsodium && export sodium_CFLAGS=$CFLAGS && export sodium_LIBS=$LDFLAGS'
+  - cd zeromq-$ZMQ
+  - ./autogen.sh
+  - if [[ -z "$SODIUM" ]]; then ./configure --prefix=$LDHACK; else ./configure --prefix=$LDHACK --with-libsodium=$LDHACK; fi
+  - make
+  - make install
+  - cd ..
```
jdfreder, line 44: I suggest putting this in a batch file under `./scripts` and having the travis.yml call out to it

jdfreder, line 44: because it's just annoying to code in a yml or json file array

rgbkrk, line 44: Good idea, we should break that out.

PR #71, opened 2016-01-30
Travis script

rgbkrk:
Problem: The preinstall was ugly within Travis

Solution: Make a scripts dir

/cc @jdfreder


jdfreder:
Thanks!

Issue #67, opened 2016-01-30
`npm install` => `npm ERR! code E500`


ghost:
Hi!

![screen shot 2016-01-30 at 08 51 04](https://cloud.githubusercontent.com/assets/10885649/12694453/e6220a9c-c72e-11e5-8807-6f0995c92234.png)

[npm-debug.log.zip](https://github.com/nteract/composition/files/110717/npm-debug.log.zip)

rgbkrk:
That makes me wonder if you have to be logged in. I just switched to a fresh node version with a fresh clone and the install went fine for me. Trying again in a Docker container.

rgbkrk:
Well, that worked too. Time for me to check out the log you posted. Thanks for being thorough.

rgbkrk:
Doing some digging, I see that others report a similar thing when their connection is slow: https://github.com/npm/npm/issues/11169 I'll see about limiting my bandwidth to attempt to reproduce. Update: [Comcast is awesome](https://github.com/tylertreat/Comcast), limited my connection to "why would I do this to myself levels" and still had no troubles.

Possible workaround: try to `npm install @reactivex/rxjs` directly. We're using the scoped module to get all the module types: https://www.npmjs.com/package/@reactivex/rxjs#all-module-types-cjses6amdtypescript-via-npm

You're already on npm3, so I haven't suggested trying `npm install @reactivex/rxjs@5.0.0-beta.1`. That's worth a shot as well.

Thanks for your patience, hopefully we'll make the developer experience better through this. :smile:

rgbkrk:
Another thing to try based on a [comment on npm/npm](https://github.com/npm/npm/issues/6669#issuecomment-62469018):

> I'm going to guess that this is caused by your login credentials being stale or otherwise incorrect. Run npm login again, and if that doesn't work, move $HOME/.npmrc out of the way and run npm login again.

ghost:
Wow :smile: that's reactivity!

I did: `$ npm install @reactivex/rxjs@5.0.0-beta.1; npm install;` and it worked!

rgbkrk:
:tada:

rgbkrk:
It would be awesome if you add that suggestion to the README as an alternative/troubleshooting thing.

ghost:
Here you are: #72

PR #72, opened 2016-01-30
Add troubleshooting to readme matching #67

ghost:


rgbkrk:
Thanks @phfrohring!

We follow [C4.1](http://rfc.zeromq.org/spec:22), so we're really happy to welcome you to the nteract party! :tada: Would you like to join as a maintainer, following the C4.1 contract?

ghost:
Dear @rgbkrk, I cannot take this responsibility but I'm very glad you offered it. C4.1 is a great process :+1: .

rgbkrk:
Thank you @phfrohring and no problem at all.

Issue #74, opened 2016-01-30
Launching kernels as an event on notebook load


rgbkrk:
It's a bit unclear to me how to launch the kernel as a result of a new notebook load. The current flow of loading is unidirectional and very Redux like. 

The problem I feel like I'm facing in creating the initial connections is that I would expect that the first `READ_JSON` event would propagate a `NEW_KERNEL` event. I'll post a PR with my current code and hopefully we'll figure out something clean.

rgbkrk:
Done in #75.

PR #75, opened 2016-01-30
Problem: need way to launch kernel w/ Redux setup

rgbkrk:
Solution: provide action `NEW_KERNEL`

Not all the way there, needs thinking in #74.


rgbkrk:
Tests pass locally and the channels are all set up for me.

rgbkrk:
The Travis error is from nvm stalling out, I think we can ignore it.

captainsafia:
Awesome!

PR #79, opened 2016-02-06
Problem: we don't have simple ways of creating messages

rgbkrk:
Solution: Write a simple `createMessage` and `createExecuteRequest`

Note: this treats the session id as a module level closure. I honestly don't know if we need to track on sessions.

### DIFF
--- a/src/api/messaging.js
+++ b/src/api/messaging.js
@@ -1,0 +1,36 @@
+/* eslint camelcase: 0 */ // <-- Per Jupyter message spec
+
+import * as uuid from 'uuid';
+
+const session = uuid.v4();
+
+export function createMessage(msg_type) {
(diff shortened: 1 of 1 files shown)


rgbkrk:
Usage:

```
import { createExecuteRequest } from './api/messaging';

const msg = createExecuteRequest('print(3)');
shell.subscribe(console.log);
iopub.subscribe(console.log);
shell.next(msg);
```

jdfreder:
Lol - `Pull requests that have a failing status can’t be merged from a phone.`

rgbkrk:
Tests pass locally, `nvm` is stalling out again.

rgbkrk:
Of note: something is munging up the channels/subjects, resulting in non-fun Rx errors:

image

PR #82, opened 2016-02-07
Problem: We need to have a central mechanism for registering commands

jdfreder:
Solution: This PR introduces one.

Inspiration from Atom commands, this is a precursor to a native menu, which would be rendered from the commands.

### DIFF
--- a/src/commands/index.js
+++ b/src/commands/index.js
@@ -1,0 +1,80 @@
+const commands = [];
+
+/**
+ * Register a command
+ * @param  {object}   command
+ * @param  {string}   command.name           name of the command, must be unique
+ * @param  {function} command.callback       function to execute
+ * @param  {string}   [command.help]         help string for the command
+ * @param  {string}   [command.accelerator]  name of the command plus an
+ *                                           accelerator which allows the
+ *                                           command to be accessed quickly via
+ *                                           the keyboard when it's in a menu.
+ *                                           The accelerator is specified using
+ *                                           a '&' character.  i.e. '&File'
+ */
+export function registerCommand(command) {
+  if (!command) throw new Error('Command not specified');
+  if (!command.name) throw new Error('A command must have a name');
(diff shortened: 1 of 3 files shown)


rgbkrk:
Great idea, glad you looked over at what Atom did.

src/commands/index.js
```diff
@@ -0,0 +1,80 @@
...
+    return commands.splice(index, 1)[0];
+  }
+}
+
+/**
+ * Find a command's index by name
+ * @param  {string} name command name
+ * @return {number}      index of the command in the internal command array
+ */
+function _find(name) {
```
rgbkrk, line 49: Since this function isn't exported, don't worry about the pseudo-hiding with leading underscores.

rgbkrk:
What do you think about Atom's `CompositeDisposable`?

rgbkrk:
_inspired by https://github.com/atom/atom/blob/master/src/command-registry.coffee_

rgbkrk:
Since Travis hasn't been happy with `nvm`, here's the locally run tests:

```
$ npm t

> composition@0.0.1 test /Users/kyle6475/code/src/github.com/nteract/composition
> electron-mocha --renderer --compilers js:babel-core/register 'test/**/*.@(js|jsx)'



  commands
    registerCommand

      ✓ registers a command

      ✓ does not allow overwrite

      ✓ accepts metadata

      ✓ checks validity of the command
    unregisterCommand

      ✓ removes a command

      ✓ is case sensitive

      ✓ does not error if a command does not exist
    get

      ✓ gets a command

      ✓ is case sensitive

      ✓ returns undefined for commands that do not exist
    invoke

      ✓ invokes a command

      ✓ is case sensitive

      ✓ does not fail if a command does not exist

  Notebook

    ✓ accepts an Immutable.List of cells (79ms)

  createStore

    ✓ sets up our store model


  15 passing (229ms)
```

Tempted to set up strider for us.

PR #83, opened 2016-02-07
Problem: launch kernel error not displayed to user

rgbkrk:
Solution: Log the error itself

We need to think on what streams our errors end up in. Is this a userland popup? How does it get represented? At the very least, I just didn't want to see the error get squashed.

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -14,7 +14,7 @@ export function newKernel(kernelSpecName) {
           spawn,
         });
       })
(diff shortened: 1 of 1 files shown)


jdfreder:
Thanks, I was seeing "err" in my console which made me go "sigh...".  I think you're right though, make errors display as user land popups will save a lot of frustration in the future.  The same kind of frustration we have in the live notebook.

PR #86, opened 2016-02-08
Problem: Command module not necessary

jdfreder:
Sln: Remove it

Talked to @rgbkrk and came to the realization that the additional metadata required for the actions to be rendered in the menu is small enough to live in the menu code itself.  When we decide to implement a command pallet, we will refactor the code so that metadata either lives on the actions or we will re-implement the commands code.


rgbkrk:
IIRC, what it seemed like we got to was that our current `actions` were roughly equivalent to the command palette and that the command palette would be built upon our central actions.

PR #87, opened 2016-02-08
Add code that creates a menu

jdfreder:
Problem: We don't have an application menu
Sln: Create code that can generate one and dispatch store actions

cc: @rgbkrk

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -31,11 +31,6 @@ export function readJSON(filePath) {
   };
 }
 
-export function openFile() {
-  /* eslint no-alert: 0 */
-  return readJSON(prompt('File path:'));
-}
-
 export function updateCell(notebook, index, cell) {
   return {
     type: 'UPDATE_CELL',
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -31,6 +31,11 @@ export function readJSON(filePath) {
   };
 }
 
+export function openFile() {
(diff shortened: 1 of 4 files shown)


rgbkrk:
I definitely see how we can rely on some of this with a web based version too, straddling similar actions. Don't activate on coding or designing for that yet, just a thought.

jdfreder:
Right now the 'Open' menu item is bound to the `readJSON` method which requires a single parameter, the JSON file path.  Really what we need here is an action that displays an open dialog and then invokes readJSON for the user selected path.

It's important to note that with the menu, only actions that accept 0 parameters can be invoked!

rgbkrk:
Weird that the first menu item is `Electron` here (on Mac at least):

screenshot 2016-02-08 10 40 12

rgbkrk:
> Really what we need here is an action that displays an open dialog and then invokes readJSON for the user selected path.
> It's important to note that with the menu, only actions that accept 0 parameters can be invoked!

ohhhh, so the action needs to be one that triggers the open dialog and then results in reading the notebook

jdfreder:
Yeah, but I bound to readJSON just to show that the dispatcher is working.

rgbkrk:
Yeah, that's cool. I'm stating it as an understanding mostly for myself and others to know what follow on PRs for other actions needs to be.

PR #88, opened 2016-02-08
Adding base cell styles via mocks

kenwheeler:
This PR has a dependency on https://github.com/nteract/react-jupyter-display-area/pull/4 being merged.

This adds baseline styles from the mocks, selected states and toolbar/toolbar styles will be in the cell selection and toolbar PRs, respectively. /cc @rgbkrk

### DIFF
--- a/src/components/Cell/Inputs.jsx
+++ b/src/components/Cell/Inputs.jsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+export default class Inputs extends React.Component {
+  static displayName = 'Inputs';
+
+  static propTypes = {
+    index: React.PropTypes.any
+  };
+
+  render() {
+    return (
+      <div className="cell_inputs">
+        [{this.props.index}]
+      </div>
+    );
+  }
+}
--- a/src/components/Cell/Cell.jsx
+++ b/src/components/Cell/Cell.jsx
@@ -15,16 +15,12 @@ export default class Cell extends React.Component {
 
   render() {
     return (
-      <div style={{
-        paddingLeft: '10px',
-        paddingRight: '10px',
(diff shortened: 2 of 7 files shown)


rgbkrk:
Looking awesome. I've merged nteract/react-jupyter-display-area#4, are you currently linking? We can make a patch release on react-jupyter-display-area - I've also added you on npm for it.

src/components/Cell/Inputs.jsx
```diff
@@ -0,0 +1,17 @@
...
+  static displayName = 'Inputs';
+
+  static propTypes = {
+    index: React.PropTypes.any
+  };
+
+  render() {
+    return (
+      <div className="cell_inputs">
+        [{this.props.index}]
```
rgbkrk, line 13: This field is actually part of the notebook document itself, labeled `execution_count` (check out this [raw notebook document](https://raw.githubusercontent.com/nteract/composition/master/intro.ipynb)).

src/components/Cell/Inputs.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+export default class Inputs extends React.Component {
+  static displayName = 'Inputs';
+
+  static propTypes = {
+    index: React.PropTypes.any
```
rgbkrk, line 7: I'd probably call this prop `executionCount`. It also only belongs to the Code Cells, as they're the ones that run execution.

rgbkrk:
Commenting on this makes me realize that a cell should have it's prop be the actual cell data. If `cell` is passed in (looking at Notebook.jsx), then the `executionCount` is at `cell.get('execution_count')` when its a code cell.

PR #90, opened 2016-02-08
Problem: Stroustrup brace style uncommon

jdfreder:
Solution:  Switch to the one true brace style.  See the styles here: http://eslint.org/docs/rules/brace-style.html


rgbkrk:
I definitely agree with this one, I'm frequently getting tripped up by it.

PR #91, opened 2016-02-08
Problem: Cell should have access to all its props

rgbkrk:
Solution: Pass the cell down

We can use this to pass on `cell.get('execution_count')` the input numbering in a follow on PR. I'll probably end up cleaning up `UPDATE_CELL` as well afterward.


jdfreder:
Thanks!

PR #97, opened 2016-02-08
Problem: In[ ] numbered by index for all cells

rgbkrk:
Solution: Pull execution count from the cell data, adapt CSS, only apply to code cells

/cc @kenwheeler


kenwheeler:
<tips hat/>

jdfreder:
woot!

rgbkrk:
:tada:

PR #98, opened 2016-02-09
Problem: Cannot execute multiple actions from one menu item.

jdfreder:
Sln: Allow menu items to execute multiple actions,
Split kill kernel out of exit

### DIFF
--- a/src/menu/default.js
+++ b/src/menu/default.js
@@ -14,7 +14,7 @@ export default fromJS([
       },
       {
         label: '&Quit',
-        action: 'exit',
+        action: ['killKernel', 'exit'],
         accelerator: 'CmdOrCtrl+Q',
       },
     ],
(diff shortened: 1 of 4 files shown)


src/menu/default.js
```diff
@@ -14,7 +14,7 @@ export default fromJS([
       },
       {
         label: '&Quit',
-        action: 'exit',
+        action: ['killKernel', 'exit'],
```
rgbkrk, line 17: Assumption here is that these get executed in order?

src/menu/index.js
```diff
@@ -28,7 +28,9 @@ export function createMenuTemplate(immutableMenu, dispatch) {
   return immutableMenu
     .update('click', cb => cb ? cb : () => {
       const action = immutableMenu.get('action');
-      if (action) {
+      if (List.isList(action)) {
+        action.map(x => dispatch(actions[x]()));
+      } else if (action) {
         dispatch(actions[action]());
```
rgbkrk, line 34: :+1:

Issue #44, opened 2016-01-10
Selectable cells


rgbkrk:
We need to have the concept of active cells (user has clicked on the cell, is interacting with it in some way, etc.)

rgbkrk:
Done in #108

Issue #22, opened 2016-01-06
Theming


rgbkrk:
I'd really love to see the entire app colorscheme to be dependent on top level variables, similar to Atom's. I'm imagining that, for the time being, we can rely on code mirror choices used as less variables throughout the application.

PR #100, opened 2016-02-09
Problem: inactive cell not flat and flushed

rgbkrk:
Solution: Only darken the execution count section on hover, active, focus

We really need the concept of an active cell, which we're still aiming for.
Wanted to get closer to the mock for people to see.

Problems:
- cell is only "active" according to the mouse (since we can't style
  relatively with the current hierarchy)
- style of the inactive cell is hardcoded to match our current codemirror theme

Those two need to be solved separately: active cells (#44) and theming (#22).

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -88,6 +88,12 @@ img
     box-shadow: 0 1px 3px rgba(0,0,0,.12), 0 1px 2px rgba(0,0,0,.24);
 }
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
screenshot 2016-02-08 20 45 05

jdfreder:
Looks slick!

PR #101, opened 2016-02-09
Add the ability to list kernels

jdfreder:
Problem: I can't change my kernel or list the kernels installed on my machine
Solution: Add a "Language" menu with these capabilities

### DIFF
--- a/src/api/kernel.js
+++ b/src/api/kernel.js
@@ -31,5 +31,5 @@ export function launchKernel(kernelSpecName) {
 }
 
 export function listKernelSpecs() {
-  return findAll().then(kernelResources => Object.keys(kernelResources));
+  return findAll();
 }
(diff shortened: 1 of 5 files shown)


src/api/kernel.js
```diff
@@ -28,3 +29,7 @@ export default function launchKernel(kernelSpecName) {
         };
       });
 }
+
+export function listKernelSpecs() {
+  return findAll().then(kernelResources => Object.keys(kernelResources));
+}
```
rgbkrk, line 35: One cool thing I didn't bother implementing in kernelspecs yet - we could be watching for new kernelspecs all the time and triggering on changes. First boot will be the simplest for us to start with though.

jdfreder, line 35: Yeah, that's another case that the menu API does not handle yet but is a perfect candidate for incremental improvements!

rgbkrk, line 35: Totally!

PR #104, opened 2016-02-10
Problem: UPDATE_CELL was only updating cell source

rgbkrk:
Solution: Change name to UPDATE_CELL_SOURCE

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -43,10 +43,10 @@ export function readJSON(filePath) {
   };
 }
 
-export function updateCell(index, cell) {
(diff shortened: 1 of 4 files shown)


rgbkrk:
I'll be posting `UPDATE_CELL_OUTPUTS` later.

I almost turned this all into _only_ `UPDATE_CELL`, accepting the fully changed cell, then it occurred to me. A single change being propagated through will be simpler for handling changes during multi-user contexts (that would be changing state as well).

jdfreder:
Sweet!  Thanks Kyle!

PR #105, opened 2016-02-10
Problem: We need a way to update cell outputs

rgbkrk:
Solution: Create an action and reducer for updating cell outputs


jdfreder:
Yay!  This is awesome, now we just need to enchannel -> dispatch this on display data

PR #110, opened 2016-02-10
Problem: we didn't like the composition name

### LINKED ISSUE #95: Naming
It's been suggested that we either rename to:
- nteract/nteract

or
- nteract/notebook

The first presumes that the Electron based notebook will be the primary byproduct of the nteract org (similar to Atom). Considering how many people refer to Jupyter as `JupyterHub | jupyter/notebook | tmpnb`, I'm fairly tempted to go along with that. One other issue we've run into in the past is that "the notebook" can refer to the notebook document, the notebook server (python), or the frontend application that is the jupyter notebook.

The composition name came from me thinking of a different way to say notebook; @karissa and I rolled with it and all was happy.

/cc @jackparmer

rgbkrk:
Solution: we :heart: the nteract name

closes #95


jdfreder:
misses README.md

rgbkrk:
Ha! The biggest one.

rgbkrk:
I also avoided changing the CSS for the "composition theme". I leave that as an exercise to the reader.

PR #114, opened 2016-02-10
Problem: execution count wasn't being updated

rgbkrk:
Solution: subscribe to execute_input replies, dispatch to an action for updating, reduce into state

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -66,3 +66,11 @@ export function updateCellOutputs(index, outputs) {
     outputs,
   };
 }
+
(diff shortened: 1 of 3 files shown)


rgbkrk:
This one was going to drive me crazy if it wasn't implemented.

It's worth having a look over our data path after this. Some of this duplication may be easier to encapsulate into reductions on a cell which reflect into reductions on the notebook state. _shrug_

rgbkrk:
cell_execute

tap tap tappity tap, hold that key down

PR #118, opened 2016-02-11
Problem: Cells are hard to move around the document.

jdfreder:
Sln: Use ids instead of indices

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "ansi-to-html": "^0.3.0",
     "codemirror": "^5.10.0",
     "color": "^0.10.1",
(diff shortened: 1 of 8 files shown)


jdfreder:
This is ready now

rgbkrk:
Great! What was the problem you ran into?

jdfreder:
I opened an issue about it #119 , caused by me forgetting we were using immutable - I used `[]` instead of `get()`, see 5a22a27

rgbkrk:
Hmmm, might be seeing something else needing some reflection:

screenshot 2016-02-12 00 02 26

PR #121, opened 2016-02-12
Problem: We only need the first execute_input resp

rgbkrk:
Solution: .first() operator

This will auto-dispose for us, which is awesome.


captainsafia:
:fireworks:

Issue #123, opened 2016-02-12
Accessing channels from actions


rgbkrk:
I'd like to have some way to access our current channels from the actions.

This is motivated by wanting to offload/move execution of a cell to an action function. We currently only have the subject, which is a scan on the state. While we could subscribe, it would be pretty odd to wait for the ​_next_​ change when we really want the ​_last_​ channels. Possibly even the last state in general.

rgbkrk:
Technically, we could pass them in to the dispatch function since they're part of context. That may be what I do here to make this simplified.

rgbkrk:
What's it like if I dispatch while in a dispatch? I realize I need that here too (looking at the code cell component right now).

rgbkrk:
At least for now, the code running is now in an action. @kenwheeler suggested creating a services layer to replace this.

PR #124, opened 2016-02-12
Problem: too much going on in code cell

rgbkrk:
Solution: create an execute cell action

Problem: This is not pretty code, needs guidance from #123

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -82,8 +82,8 @@ export function updateCellExecutionCount(id, count) {
   };
 }
 
-export function executeCell(id, source, dispatch, channels) {
-  return () => {
+export function executeCell(id, source, channels) {
+  return (subject, dispatch) => {
     // TODO: figure out where channels come from
     const { iopub, shell } = channels;
 
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -82,8 +82,8 @@ export function updateCellExecutionCount(id, count) {
   };
 }
 
-export function executeCell(id, source, channels) {
-  return (subject, dispatch) => {
+export function executeCell(id, source) {
+  return (subject, dispatch, channels) => {
     // TODO: figure out where channels come from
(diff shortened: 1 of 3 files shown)


rgbkrk:
What I'd like to do after this is figure out how we can handle not having to pass dispatch or channels in here.

rgbkrk:
This makes it easier for follow on actions that use the kernels like completion, doc requests, etc. while muddying what the story is for our redux-like store.

kenwheeler:
This is great. That logic doesn't necessarily belong in the component. That said, it also doesn't necessarily belong in the action. I recently did something like this on a react native app, where I abstracted service calls to an API module. See below:

```
import Services from "../services";
import * as constants from "../constants";

import config from "../../config/development";
export const API = new Services(config.services);
```

The idea here being, the API (or our channels), can be instantiated and exported right from the actions file. That way we get:
- An abstraction to handle non-actiony logic
- A testable service (or channels) layer
- Access to the service (channel) layer right from actions.

rgbkrk:
Should we be viewing actions as UI triggers?

When a new kernel is loaded, how do we update that service?

jdfreder:
I like @kenwheeler 's suggestion for an `API` module, or something, to store non-action things, but I'd like to know how that would work with something like the declarative menu.  Right now menuitems can define actions by name, and those actions get dispatched on click.  With an API, the menu code would either have to know to look in both the API and actions/index (and we'd have to figure out which had precedence) _or_ you would have to define callbacks instead of actions (loosing some of the declarative goodness).  Do either of you see a way to deal with this in a clean way?

jdfreder:
Let's continue the discussion here, I'm just merging because of C4.1.

rgbkrk:
Yeah I think we can move forward with making services. We have a lot of unanswered questions there.

At the very least, this is no longer coupled to the cell and people can start iterating on some of the other interactions with the kernels.

PR #131, opened 2016-02-12
Problem: rxjs out of date, breaking enchannel

### LINKED ISSUE #115: Bump to RxJS 5.0 beta 2
[The changelog for RxJS 5](https://github.com/ReactiveX/RxJS/blob/master/CHANGELOG.md#500-beta2-2016-02-10) lists a bunch of performance improvements as well as fixing `Subject.create`'s arguments (we ran into this and [they fixed it after our feedback](https://github.com/ReactiveX/RxJS/pull/1241)). It needs adaption in any place where `Subject.create` is used (flip the args). Other than that, these are the two packages that will need to handle the bump:
- [x] [enchannel-zmq-backend](https://github.com/nteract/enchannel-zmq-backend)
- [ ] nteract (here, silly)

rgbkrk:
Solution: update both, pin RxJS beta hard

Closes #115 

Fixed the license field too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,18 +20,18 @@
     "data"
   ],
   "author": "nteract contributors",
-  "license": "MPLv2",
+  "license": "MPL-2.0",
   "bugs": {
     "url": "https://github.com/nteract/nteract/issues"
   },
   "homepage": "https://github.com/nteract/nteract",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


jdfreder:
Woot!  Thanks @rgbkrk !

Issue #130, opened 2016-02-12
Document the getting started hacking process


rgbkrk:
# OS X

```
brew install zmq
brew install pkg-config
```
# Jupyter tooling

```
pip install jupyter_console
jupyter kernelspec install-self
```

Installing the Python 3 kernel:

```
python3 -m pip install ipykernel
python3 -m ipykernel install --user
```

PR #132, opened 2016-02-13
Problem: We don't give enough guidance for install

rgbkrk:
Solution: Provide some more steps


rgbkrk:
Closes #130, as best as I can for now.

jdfreder:
Thanks!

PR #133, opened 2016-02-13
Problem: OS X has case insensitive paths

### LINKED ISSUE #109: Filename Proposal
We have disallowed using capital case at Walmart, due to OSX case insensitivity issues. Additionally, we use the .js extension instead of .jsx.

Do we have good reasons for either? On other projects we've found these patterns to be more trouble than they are worth.

Thoughts?

rgbkrk:
Solution: lower-case-kabobs

Fixes #109


jdfreder:
thanks!

PR #135, opened 2016-02-13
Problem: I can't use my mouse to move cells

jdfreder:
Solution: Implement _basic_ drag and drop


rgbkrk:
Nice! I look forward to using this.

rgbkrk:
Before I opened this PR for viewing I had hoped the drop target would be a region in between cells. Then I read and wondered how the IDs would work, afraid of using them. At the very end when handling the state update, it came out very clean. Well done!

We should probably handle the case where one of the IDs doesn't exist, resulting in no changes.

rgbkrk:
dnd

PR #140, opened 2016-02-13
Problem: DnD color doesn't match theme

rgbkrk:
Solution: Stick to gray for now

Aim for designing in grayscale until we have a theming and accessibility story.


freeman-lab:
looks nice!

PR #141, opened 2016-02-13
Problem: large cells will crash DnD

### LINKED ISSUE #138: Problem: Large cells can't be dragged and dropped
moving_big_cell

Hypothesis: we should be using an icon instead of an HTML snapshot of the element

rgbkrk:
Solution: Use a preview image instead of the default

Fixes #138

### DIFF
--- a/src/components/cell/cell.js
+++ b/src/components/cell/cell.js
@@ -41,6 +41,7 @@ function collectSource(connect, monitor) {
   return {
     connectDragSource: connect.dragSource(),
     isDragging: monitor.isDragging(),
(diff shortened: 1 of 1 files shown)


rgbkrk:
Amusingly, I used the horse from React DnD first. I thought this handled arbitrary DOM nodes for rendering, it didn't like my `<pre>` block. :wink: Now it's a simple PNG of our input numbering.

freeman-lab:
looks nice! maybe we can support the :horse: via a theme =)

rgbkrk:
Heh maybe this needs to be a div with a class for styling.

PR #142, opened 2016-02-13
problem: hovering and selecting can be hard to see

freeman-lab:
solution: use symmetric shadow and speed up transitions

welcome disagreement on this one =) but I find that when I hover over a cell that extends off the bottom, I'm confused as to whether it's actually highlighted because I can't see the bottom, where the shadow is. making the shadows symmetric helps you know what you're hovering. and I found speeding up the transitions helps a bit as well.

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -80,16 +80,16 @@ img
 {
     margin: 10px;
 
-    transition: all .2s ease-in-out;
+    transition: all .1s ease-in-out;
 }
 
 .cell:hover
 {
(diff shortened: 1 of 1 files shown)


rgbkrk:
We try not to disagree too much here, welcoming small changes via [C4.1](http://rfc.zeromq.org/spec:22). Let's give it a spin. If someone doesn't like it, they can change it again and we keep evolving.

Thank you @freeman-lab!

rgbkrk:
Ok, yeah I love having these transitions sped up. It took me a sec to see the difference between selected and hovered, though now I have the cue!

PR #143, opened 2016-02-13
problem: encountering a weird electron-prebuilt error

freeman-lab:
solution: suggest a step that worked in >= 1 environment


rgbkrk:
Weird that it doesn't use the local version. You're launching via `npm run start`?

freeman-lab:
I was... dug deeper, I think it happened because the `zmq` build failed, which meant `electron-prebuilt` postinstall hook never ran (even though it was installed into `node_modules`), so after I switched pythons to successfully build zmq, I then hit the electron error on startup

PR #146, opened 2016-02-14
Problem: We can't open a different file

rgbkrk:
Solution: Allow command line opening of a notebook.

This will definitely require further improvements as well as pinning what our "load this file" strategy is.

### DIFF
--- a/app.js
+++ b/app.js
@@ -6,13 +6,11 @@ import path from 'path';
 
 let windows = [];
 
-// Quit when all windows are closed.
 app.on('window-all-closed', () => {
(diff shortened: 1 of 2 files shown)


rgbkrk:
Options I've been thinking about for loading files after looking at other electron apps:
- Query params
- Hash parameter (as used here)
- The IPC module

jdfreder:
It's a start!

rgbkrk:
Yeah, definitely something. I was really hoping I could just pass in something to the BrowserWindow at the start - some sort of context object that would be available.

PR #148, opened 2016-02-14
Problem: Not fully adapted to our filename standards

rgbkrk:
Solution: Rename and refactor to handle new filenames.

We had some stragglers for #109. While I was at it, fixed the kernel menu ordering (as that's what I originally set out to do).


rgbkrk:
Yay, tests fixed.

jdfreder:
Thanks Kyle!

PR #149, opened 2016-02-14
Problem: Blind spot between cells when dragging and dropping them

jdfreder:
Sln: Refactor the cells into Cell and DraggableCell.  Use padding in DraggableCell to separate cells.


rgbkrk:
Most excellent refactor!

rgbkrk:
I'm going to have to get used to this new drop. I rather liked how flush it was against the cell, as it didn't push the cell up or down on movement.

PR #154, opened 2016-02-15
Lint cell and lower drag thickness

rgbkrk:
Made the DnD code pass lint checks and the border lines a bit less intense.


captainsafia:
Sehr gut, @rgbkrk!

:dancer:

PR #159, opened 2016-02-15
notebooks with titles

ivanov:
So it's hard to keep track of what's what once you start opening multiple notebooks. Here's a set of changes that makes the title of each window be the path of the notebook relative to the current working directory, without the `.ipynb` extension at the end. As a bonus, the Window menu on OS X is now usable.

They look like this:

screen shot 2016-02-15 at 11 22 25 am

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,6 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <title>nteract</title>
     <link rel="stylesheet" href="./node_modules/normalize.css/normalize.css"/>
     <link rel="stylesheet" href="./node_modules/codemirror/lib/codemirror.css"/>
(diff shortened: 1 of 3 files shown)


rgbkrk:
I like how you chose a notebook that has a message we don't implement yet (`clear_output`). :smile:

ivanov:
:thumbsup:

On Mon, Feb 15, 2016 at 11:54 AM, Kyle Kelley notifications@github.com
wrote:

> I like how you chose a notebook that has a message we don't implement yet (
> clear_output). [image: :smile:]
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/nteract/nteract/pull/159#issuecomment-184362785.

## 

Paul Ivanov
http://pirsquared.org | GPG/PGP key id: 0x0F3E28F7

PR #160, opened 2016-02-15
Problem: unregistered actions cripple app

rgbkrk:
Solution: check that an action is well-formed, log errors

This solves the "why doesn't the `alert` for save show up cleanly" issue.

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -5,9 +5,22 @@ export default function createStore(initialState, reducers) {
   const subject = new Rx.Subject();
 
   const store = subject.scan(
-    (state, action) => reducers[action.type].call(null, state, action),
+    (state, action) => {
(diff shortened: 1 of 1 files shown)


src/store/index.js
```diff
@@ -5,10 +5,23 @@ export default function createStore(initialState, reducers) {
...
+        console.error(action);
+        return state; // no reduction
+      }
+
+      return reducers[action.type].call(null, state, action);
+    },
     initialState || {}
   );
 
+  // Debugging time
```
ivanov, line 20: is this like :hammer: :clock1130:

rgbkrk, line 20: We need to define a story for error display, both for users, debugging, etc. Haven't done it yet, so we're logging for now!

PR #162, opened 2016-02-16
Problem: I can't save!

jdfreder:
Solution: Add the ability to save

### DIFF
--- a/src/api/save.js
+++ b/src/api/save.js
@@ -1,0 +1,24 @@
+import * as commutable from 'commutable';
+import * as path from 'path';
+import { writeFileSync } from 'fs';
+
+import remote from 'remote';
+const dialog = remote.require('dialog');
+
+export function showDialog(defaultPath) {
+  const filename = dialog.showSaveDialog({
+    title: 'Save Notebook',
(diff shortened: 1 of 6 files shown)


rgbkrk:
Woo!

src/api/save.js
```diff
@@ -0,0 +1,24 @@
+import * as commutable from 'commutable';
+import * as path from 'path';
+import { writeFileSync } from 'fs';
+
+import remote from 'remote';
```
rgbkrk, line 5: It's at this point that I would wonder if the save dialog should be triggered from within the menu _before_ it reaches the frontend.

src/api/save.js
```diff
@@ -0,0 +1,24 @@
...
+
+  if (filename && path.extname(filename) === '') {
+    return filename + '.ipynb';
+  }
+
+  return filename;
+}
+
+export function save(filename, notebookState) {
+  writeFileSync(filename, JSON.stringify(commutable.toJS(notebookState), null, 2));
```
rgbkrk, line 23: If we block here, it blocks all rendering operations.

PR #163, opened 2016-02-16
Problem: async actions should stay in actions

rgbkrk:
Solution: move 'save' code from reductions into actions, refactor state push

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -1,6 +1,10 @@
 import { getJSON } from '../api';
 
+import * as commutable from 'commutable';
+
 import { launchKernel } from '../api/kernel';
+
+import { writeFile } from 'fs';
 
 import {
   createExecuteRequest,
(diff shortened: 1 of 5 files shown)


rgbkrk:
Follow up to #162

src/actions/index.js
```diff
@@ -38,30 +42,32 @@ export function newKernel(kernelSpecName) {
 }
 
 export function save() {
-  return (subject) => {
+  return (subject, dispatch, state) => {
```
jdfreder, line 45: Ah @rgbkrk this is the part I missed, I wasn't sure how to get state anywhere other than in a reducer (which is why I put the save there)

rgbkrk, line 45: this is also kind of bad if used with anything asynchronous inline, as the state will not be the same deeper in. Basically, we have to tread carefully until we find a better design pattern.

PR #167, opened 2016-02-18
Problem: clear_output needs to be handled

rgbkrk:
Solution: introduce a reduction step based on clear_output

Woo for silly IPython/Jupyter animations:

clear_output

This one is for you @ivanov

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -164,10 +164,13 @@ export function executeCell(id, source) {
 
     // Handle all the nbformattable messages
     childMessages
-         .ofMessageType(['execute_result', 'display_data', 'stream', 'error'])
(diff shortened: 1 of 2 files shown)


ivanov:
:+1: except for the whole Problem Solution business

rgbkrk:
Amusingly, I wrote a PR that I closed that _didn't_ use the Problem/Solution jazz. https://github.com/nteract/react-jupyter-display-area/pull/7

rgbkrk:
Of note - you can write commit messages however you like. :wink: We accept patches here readily.

ivanov:
what's the policy on sarcasm? ;)

rgbkrk:
As long as you're not repeatedly hassling people. I've stopped using Problem/Solution now Paul.

PR #169, opened 2016-02-19
Exported constants for actions

rgbkrk:
This switches us over to explicit constants.


betatim:
Looks good to me, though hard to tell as a lot of lines have changed, but it started and ran stuff.

Can you say a few words on the advantage/disadvantage of doing this?

rgbkrk:
[`Symbol`s](http://blog.keithcirkel.co.uk/metaprogramming-in-es6-symbols) force developers to use the exported constants instead of hard coding strings.

`Symbol("foo") !== Symbol("foo")`, so the constants in the constants file _must_ be used. I've tripped myself up at least once with the strings before. We can always swap back to strings if we have to, since they're exported constants and all usages rely on the constants file.

PR #170, opened 2016-02-20
Refactor to support testing

rgbkrk:
Named functions, appropriate naming for the notebook read function, breaking out some functionality into separate scripts for testing.


jdfreder:
Woah this is big, good work!

rgbkrk:
Ha, yeah. Would have been happy to break it up too. No one was around. :sob: Just kept going for testing sake.

PR #175, opened 2016-02-20
Even more tests of notebook reduction.

rgbkrk:
Following up on #170

### DIFF
--- a/test/reducers/document_spec.js
+++ b/test/reducers/document_spec.js
@@ -1,16 +1,58 @@
 import { expect } from 'chai';
 
+import * as commutable from 'commutable';
+
 import {
   loadNotebook,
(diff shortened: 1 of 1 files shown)


rgbkrk:
Wonder if it's time to turn on Travis? Everything we're testing right now doesn't use zmq, I wonder if there's a way to install most of a package and ignore one.

jdfreder:
I don't see why turning Travis on could hurt.  I don't think we should let it slows us down though, at this point in time.

rgbkrk:
Feel free to turn it back on, hopefully now it's not being flaky. We need CI so if we run into more problems we'll run strider or something on carina.

For now we'll assume people run npm t and try out the notebook locally. If it breaks, we fix it.

PR #176, opened 2016-02-20
[WIP] The Focused Cell

rgbkrk:
DON'T MERGE, this is semi-functional with some bad breaking things.

I may break this up into only providing the code for the actions and reducers, leave the UI bits for later.

This begins adding a way to focus a cell, primarily for the `run-and-move-to-next` action.
- [x] Create the actions and reducers for setting the focused cell
- [ ] On code cells, setting focus will result in honing in on the codemirror ref
- [ ] On Markdown cells, if not in "view rendered" mode, should do... something?
- [ ] Figure out how to reason about that fact that this is always going to have to update based on the user's "active position" on the page

/cc @kenwheeler

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -15,6 +15,7 @@ import {
   UPDATE_CELL_OUTPUTS,
   UPDATE_CELL_SOURCE,
   ERROR_KERNEL_NOT_CONNECTED,
+  SET_FOCUSED,
 } from '../actions/constants';
 
 import {
@@ -41,6 +42,13 @@ reducers[SET_SELECTED] = function setSelected(state, action) {
       action.ids;
   return Object.assign({}, state, {
     selected,
+  });
+};
+
+reducers[SET_FOCUSED] = function setFocused(state, action) {
+  const focused = action.id;
+  return Object.assign({}, state, {
+    focused,
   });
 };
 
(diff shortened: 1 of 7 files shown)


src/reducers/index.js
```diff
@@ -35,6 +41,43 @@ reducers[UPDATE_CELL_SOURCE] = updateSource;
...
+    // Here we'll create a new cell at the end of the document
+    // create a new cell
+    const cell = commutable.emptyCodeCell; // Until configurable
+
+    // This is another time where I wish I could either tell appendCell
+    // what cellID I want or for it to return one to me
+    // notebook = commutable.appendCell(notebook, cell);
+    // doing it here, then writing the idiomatic clean interface later
+    // When writing tests, I ran into this too as the cell ID was wrapped
+    // in a closure and had to be extracted by "knowing" where the next cell was
```
rgbkrk, line 66: Ok, fixed this in https://github.com/nteract/commutable/pull/18

jdfreder, line 66: Awesome!  Also, you'll need to rebase this at some point

rgbkrk, line 66: I'm probably just going to toss it and think about it again renewed. It at least helped me understand some issues in handling focus.

src/reducers/index.js
```diff
@@ -1,5 +1,9 @@
 import * as fs from 'fs';
 
+import * as commutable from 'commutable';
+
+import { v4 as uuid } from 'node-uuid';
```
rgbkrk, line 5: Heh, just noticed that this uses `node-uuid` while our package.json is using `uuid`.

rgbkrk:
Closing this since it needs a healthy rebase and a healthy dose of rethinking.

PR #179, opened 2016-02-21
Problem: No mechanism for listening to keyboard events on elements

jdfreder:
Solution: Add atom's atom-keymap package.  Add basic code to get it to work. 

Still required, the ability to load user keymaps and some useful actions for key bound events.


rgbkrk:
Cool, this is pretty clean.

PR #183, opened 2016-02-26
Cell Toolbar

kenwheeler:
cc @rgbkrk

### DIFF
--- a/src/components/cell/toolbar.js
+++ b/src/components/cell/toolbar.js
@@ -1,0 +1,49 @@
+import React from 'react';
+
+import { executeCell, removeCell } from '../../actions';
+
+class Toolbar extends React.Component {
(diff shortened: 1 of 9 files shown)


src/components/cell/toolbar.js
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  static contextTypes = {
+    dispatch: React.PropTypes.func,
+  };
+
+  constructor(props) {
+    super(props);
+    this.removeCell = this.removeCell.bind(this);
+    this.executeCell = this.executeCell.bind(this);
```
rgbkrk, line 21: What makes this need to bind to this here?

...

Oh, because you pass it to the `onClick`

rgbkrk:
:boom:

PR #187, opened 2016-03-01
Add the ability to create new notebooks

### LINKED ISSUE #155: Implement New Notebook
On the heels of https://github.com/nteract/nteract/pull/153, we need a new notebook menu item:

```
File > 
    New >
        Python
        JavaScript
```

as well as a way to create an empty notebook. This second part is likely work within [commutable](https://github.com/nteract/commutable).

jdfreder:
Change the default behavior of start to open an empty notebook.  Use python2 as the default kernel.

Things to be done in other PRs,
make the default kernel configurable
pick a kernel if the default isn't found
allow the user to new a specific kernel?

closes #155

### DIFF
--- a/main/menu.js
+++ b/main/menu.js
@@ -24,6 +24,11 @@ export const file = {
   label: '&File',
   submenu: [
     {
+      label: '&New',
+      click: () => launch(),
+      accelerator: 'CmdOrCtrl+N',
+    },
+    {
       label: '&Open',
       click: () => {
         const opts = {
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -20,6 +20,7 @@ import {
   NEW_CELL_AFTER,
   REMOVE_CELL,
   UPDATE_CELL_EXECUTION_COUNT,
+  NEW_NOTEBOOK,
   READ_NOTEBOOK,
   ERROR_KERNEL_NOT_CONNECTED,
 } from './constants';
@@ -30,6 +31,10 @@ import {
 } from '../api/messaging';
 
 import Immutable from 'immutable';
+import * as path from 'path';
+
+const remote = require('remote');
+const dialog = remote.require('dialog');
 
 export function exit() {
   return {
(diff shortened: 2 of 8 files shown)


src/actions/index.js
```diff
@@ -59,12 +64,41 @@ export function newKernel(kernelSpecName) {
...
+        if(!filename) {
+          return;
+        }
+
+        if (path.extname(filename) === '') {
+          filename = filename + '.ipynb';
+        }
+        dispatch(saveAs(filename));
+      });
+      return;
```
rgbkrk, line 96: This same logic was in here before we refactored how save-as works, I'm definitely wondering how to make this approachable in general.

main/menu.js
```diff
@@ -24,6 +24,11 @@ export const file = {
   label: '&File',
   submenu: [
     {
+      label: '&New',
+      click: () => launch(),
+      accelerator: 'CmdOrCtrl+N',
+    },
```
rgbkrk, line 30: Let's try to make this like the language menu next.

```
New Notebook
  -->
    Python2
    Python3
```

src/actions/index.js
jdfreder, line 96: I could have IPC'ed back to the main thread, but that seemed ridiculous

rgbkrk, line 96: I mean more like doing the logic in the main thread, since the running notebook doesn't need access. Other reason is that on OS X if I don't have any windows open, these menu items still need to work.

rgbkrk, line 96: Semantically, I'm saying we would make new notebook open a brand new window.

jdfreder, line 96: The new notebook menu item does open a new window.   I don't think the action should open a new window because the action's purpose is to dispatch state reducers.

main/menu.js
jdfreder, line 30: +1

src/actions/index.js
rgbkrk, line 96: Ah, I'm seeing now. I thought this was operating on the notebook we just created. Since we don't have something on disk and we need the structure in place for a default notebook, this is the bootstrap.

jdfreder, line 96: Yup

src/index.js
```diff
@@ -16,7 +17,11 @@ import {
 import { initKeymap } from './actions/keymap';
 
 const filename = decodeURIComponent(window.location.hash.slice(1));
-const { store, dispatch } = createStore({ notebook: null, selected: [], filename }, reducers);
+const { store, dispatch } = createStore({
+  notebook: null,
```
rgbkrk, line 21: This does make me wonder if we can make the default notebook be an empty notebook, relying on the commutable helper.

src/reducers/document.js
```diff
@@ -1,6 +1,11 @@
 import * as commutable from 'commutable';
 import * as uuid from 'uuid';
 
+export function newNotebook(state) {
+  const notebook = commutable.appendCell(commutable.Notebook(), commutable.emptyCodeCell);
```
rgbkrk, line 5: This misses passing the language_info, that's ok for now though.

rgbkrk:
No matter what, this highlights our bootstrapping problem in `nteract` right now in terms of how the notebook gets loaded initially. This does make me reflect on thinking about external actions affecting the state of our rendered notebook (filesystem changes, git status, etc.), though I'll have to come back to this again.

src/reducers/document.js
jdfreder, line 5: Purposefully excluded, kernelspecs does not provide that information, it is obtained from kernel start.

PR #188, opened 2016-03-01
Provide a means for injecting kernelnames into new

rgbkrk:
This merely sets us up for being able to set the new notebook with language selection.

```
File >
  New >
    Python2
    Python3
    JavaScript
```


rgbkrk:
Of note - I left a boilerplate function in to be fixed (needs to launch the right kernel). This was mostly about setting up the menu, not the real action behind it.

Issue #18, opened 2016-01-04
Notebook format


rgbkrk:
Sometime ago, I wrote [nbformatjs](https://github.com/rgbkrk/nbformatjs) (available on npm as `nbformat`) for working with the notebook format. We need to add on to that package and/or build follow ons that use it to handle a variety of things:
- Upgrading from previous formats
- Validation of format
- Converting from the disk format to the in-memory format, e.g. The "multi-line string" array of strings should be `.join('')`ed
- Utilities for working with an Immutable.js notebook (load, convert, save)

rgbkrk:
Also made the simplified version at https://github.com/rgbkrk/nbschema that doesn't enforce any particular validator (there were too many options and I couldn't decide on an interface).

rgbkrk:
Now that [commutable](https://github.com/nteract/commutable) exists, we can move discussion over there. Conversion from disk format --> in-memory format is done, as well as loading and saving.

PR #189, opened 2016-03-02
Constants Helper

StevenLangbroek:
This utility makes creating constants a little less verbose and error-prone. When writing tests for it, I noticed `electron-mocha` doesn't support (and probably never will support) `--watch`, see [#18](https://github.com/jprichardson/electron-mocha/pull/18), so I added `watch` as `devDependency` and a new task for running tests on file changes.
- Add watch as dependency for TDD, plus 'test:watch' task
- Add createConstants utility + specs

### DIFF
--- a/src/actions/constants.js
+++ b/src/actions/constants.js
@@ -1,24 +1,16 @@
-export const CHANGE_FILENAME = Symbol('CHANGE_FILENAME');
+import createConstants from './createConstants';
 
-export const START_SAVING = Symbol('START_SAVING');
-export const DONE_SAVING = Symbol('DONE_SAVING');
-
-export const ERROR_KERNEL_NOT_CONNECTED = Symbol('ERROR_KERNEL_NOT_CONNECTED');
-
-export const NEW_KERNEL = Symbol('NEW_KERNEL');
-export const KILL_KERNEL = Symbol('KILL_KERNEL');
-
-export const EXIT = Symbol('EXIT');
-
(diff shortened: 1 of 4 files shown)


rgbkrk:
:boom:

rgbkrk:
That's clean!

rgbkrk:
Looks like the app isn't launching after this, digging now.

src/actions/constants.js
```diff
@@ -1,24 +1,16 @@
...
+  'NEW_KERNEL', 'KILL_KERNEL',
+  'EXIT',
+  'SET_SELECTED',
+  'NEW_NOTEBOOK', 'READ_NOTEBOOK',
+  'MOVE_CELL', 'REMOVE_CELL',
+  'NEW_CELL_AFTER',
+  'UPDATE_CELL_EXECUTION_COUNT',
+  'UPDATE_CELL_OUTPUTS',
+  'UPDATE_CELL_SOURCE'
+);
```
rgbkrk, line 16: It occurs to me that we no longer get good autocomplete with the constants being generated directly.

If we switch it back to explicit exports, we get named exports to come up complete with type information:

screenshot 2016-03-03 22 38 05

StevenLangbroek, line 16: Well it doesn't know the type anymore. Is that important when working with this piece of code?

rgbkrk, line 16: All about hints and developer experience here. I'm not going to argue either way, I noticed the limitation and pointed it out.

PR #192, opened 2016-03-03
Pass the mimebundle transforms and displayOrder through

rgbkrk:
This sets us up for being able to configure what our default transforms are, motivated by @andrewosh's remote kernel use case and security concerns.

Example sanitized notebook (only allows `text/plain` mimetype):

screenshot 2016-03-02 21 19 46

To do this, I created a new `displayOrder` and passed it through to the notebook:

``` diff
diff --git a/src/index.js b/src/index.js
index 953760e..177c642 100644
--- a/src/index.js
+++ b/src/index.js
@@ -24,6 +24,9 @@ const { store, dispatch } = createStore({
 }, reducers);
 initKeymap(window, dispatch);

+import Immutable from 'immutable';
+const displayOrder = new Immutable.List(['text/plain']);
+
 import { ipcRenderer as ipc } from 'electron';
 ipc.on('menu:new-kernel', (e, name) => dispatch(newKernel(name)));
 ipc.on('menu:save', () => dispatch(save()));
@@ -56,7 +59,9 @@ class App extends React.Component {
             <Notebook
               selected={this.state.selected}
               notebook={this.state.notebook}
-              channels={this.state.channels} />
+              channels={this.state.channels}
+              displayOrder={displayOrder}
+              />
           }
         </div>
       </Provider>
```

This is merely putting it in place so that Andrew can do some more hacking on top.

### DIFF
--- a/src/components/cell/code-cell.js
+++ b/src/components/cell/code-cell.js
@@ -5,6 +5,8 @@ import Inputs from './inputs';
 import Editor from './editor';
 import Display from 'react-jupyter-display-area';
 
+import Immutable from 'immutable';
+
 import {
   executeCell,
 } from '../../actions';
@@ -14,9 +16,11 @@ export default class CodeCell extends React.Component {
 
   static propTypes = {
     cell: React.PropTypes.any,
+    displayOrder: React.PropTypes.instanceOf(Immutable.List),
     id: React.PropTypes.string,
     language: React.PropTypes.string,
     theme: React.PropTypes.string,
+    transforms: React.PropTypes.instanceOf(Immutable.Map),
   };
 
   static contextTypes = {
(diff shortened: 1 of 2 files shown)


jdfreder:
Thanks! :cake:

rgbkrk:
As suggested by @kestertong, we should check out [Google's closure library for making our safe mode include HTML](https://github.com/google/closure-library/blob/master/closure/goog/html/safehtml.js).

betatim:
How do I use this in a nice way with babel and react? I made a [codepen with react, babel and cosure](http://codepen.io/betatim/pen/eZvJeo) and don't understand why I have to put my react rendering inside `$()`. My guess is that `goog.require(...)` only gets resolved/executed much later than the react component is first rendered. Naive question: why can't I just import it like other modules? Need more :mortar_board: :books:

rgbkrk:
New to me too. Looks like the goog require is a global. I'm not a fan of that (non idiomatic for JS/node), though I'm happy to learn from what's here.

PR #193, opened 2016-03-03
Application should not read or erase all notebook data while in use

jdfreder:
Use IPC to pass notebook data into app
Update commutable
Default to kernelspec then fallback on langinfo for kernel name, last resort py3


rgbkrk:
This definitely works for now and breaks us away from that messy encodeURIComponent. :+1:

PR #197, opened 2016-03-04
Fix file open via CLI

jdfreder:


main/launch.js
```diff
@@ -47,7 +47,7 @@ export function launchFilename(filename) {
       if (err) {
         reject(err);
       } else {
-        launch(fromJS(JSON.parse(data)), filename);
+        resolve(launch(fromJS(JSON.parse(data)), filename));
```
rgbkrk, line 50: Ah yeah, the commutable `fromJS` and the immutable `fromJS`.

We wrote commutable to have a fromJS so it was familiar, semantically as `commutable.fromJS(notebookJSON)`

jdfreder, line 50: Yeah, familiarity is nice, but it's easy to forget which you have imported and which is being called.  This may just be a "jon" problem, else we may want to think about a name change, like `fromNotebookJS` ...

rgbkrk, line 50: `fromNotebookJS` seems good

PR #198, opened 2016-03-04
Add notebook window spawn test

jdfreder:


rgbkrk:
Very glad we're testing the main and renderer process now.

jdfreder:
It's a start.  We still need to get electron-mocha using electron-compile instead of babel.  Once we do that I'll be able to check the contents of the spawned window for `.cell` !!  This should catch lots of bugs before they appear.

PR #201, opened 2016-03-05
Add a cell spawning hover context toolbar

jdfreder:


src/components/cell/spawn-tool.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  _spawnTextCell() {
+    if (this.props.spawnBefore) {
+      this.context.dispatch(createCellBefore('markdown', this.props.id));
+    } else {
+      this.context.dispatch(createCellAfter('markdown', this.props.id));
+    }
+  }
```
rgbkrk, line 40: I'm going to go ahead and say that spawn is a weird name for creating cells, since we use spawn elsewhere in the code base for launched processes/kernels.

rgbkrk:
Awesome.

jdfreder:
spawner

betatim:
Woah! +:100:

rgbkrk:
Ok, after playing with this for a bit here's my current feedback:
- If possible, it would be nice for the `+text` and `+code` buttons to not always be shown (possibly based on mouse position?). I'm not sure what the right thing to do is, just that we want to encourage thinking about their current editing.
- New text cells definitely need a min-height - it was hard to discover how to start editing them
  - Maybe we even need a placeholder when an empty markdown cell - "double-click to edit" or we default to edit mode?
- We need to make a mock for the cell creation buttons and work towards a design that works with the rest of our design

rgbkrk:
While I'm at it, I think I made a mistake in updating the mock for cell toolbar to be something flush with the cell. It should hover as a menu too (possibly...). That or we need the gray tones to match.

PR #202, opened 2016-03-05
Problem: newly created text cells are invisible

betatim:
Solution fill new markdown cells with some text.

(I'd like some tutoring about `componentWillReceiveProps`: why is called extremely often? In particular who 'triggers' this when the notebook is loaded from a file?)

### DIFF
--- a/src/components/cell/markdown-cell.js
+++ b/src/components/cell/markdown-cell.js
@@ -22,14 +22,18 @@ export default class MarkdownCell extends React.Component {
     this.state = {
       view: true,
       // HACK: We'll need to handle props and state change better here
-      source: this.props.cell.get('source'),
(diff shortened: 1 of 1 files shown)


rgbkrk:
That's a good start!

I'd instead have this conditionally choose what to `render` based on what's in `this.props.cell.get('source')`. Could display a placeholder div instead of changing state.

rgbkrk:
`componentWillReceiveProps` is going to happen for every character you type within this cell

There may be more props coming in that need to be cleaned up from the `...this.props` passed in by the parent element that we don't need updates from.

betatim:
The other thing that would be cool is to delete that text as soon as someone double clicks (only to make it reappear if they didn't type anything)

PR #203, opened 2016-03-05
Conditionally display empty cell message depending on state

jdfreder:
Suggested in #202

@betatim @rgbkrk

### DIFF
--- a/src/components/cell/markdown-cell.js
+++ b/src/components/cell/markdown-cell.js
@@ -22,18 +22,15 @@ export default class MarkdownCell extends React.Component {
     this.state = {
       view: true,
       // HACK: We'll need to handle props and state change better here
-      source: 'Edit me by double clicking',
+      source: this.props.cell.get('source'),
     };
   }
 
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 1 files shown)


rgbkrk:
This certainly captured the intent of what I said!

I was thinking it would be a div you add in conditionally, this works even better since you get to set the message using Markdown. :wink:

rgbkrk:
One thing I noticed here is that we're not focusing the editor when we double click (were we before?)

jdfreder:
No, I don't think so.  In jupyter/notebook we do

betatim:
Why the `!==`?

```
 this.state.source !== '' ?
                this.state.source :
                `*Empty markdown cell, double click me to add content.*`
```

vs

```
 this.state.source ?
                 this.state.source :
                 `*Empty markdown cell, double click me to add content.*`
```

Does it make any difference in JS? I python I spend my time telling people to use `if some_string` :-/

jdfreder:
Looks like that works too @betatim , thanks!  Want to open a PR?

rgbkrk:
:+1: @betatim

PR #206, opened 2016-03-05
string ?: does the same as string !== '' ?:

betatim:
Test for truthyness of a string instead. #pythonhabits


rgbkrk:
:100:

jdfreder:
Thanks @betatim !

PR #207, opened 2016-03-05
Tweaks to make the spawn cell toolbar look like the cell toolbar

jdfreder:
Also, only show the toolbar when the mouse is 30% away from the border, opposed to the current 50%.


rgbkrk:
Awesome.

Styling matchup :100:

The before/after notion when between cells is a bit odd since it belongs to each cell.

rgbkrk:
Percentages for where my mouse are definitely seem about right for me. :+1:

PR #208, opened 2016-03-05
Second take on cell creator toolbar,

jdfreder:
- cell creator now exists at document level
- use pixel hover boundaries instead of percentage
- show creator when no cells exist


jdfreder:
:tada:

jdfreder:
Licecap doesn't want to record my cursor today, but here's a cap anyways:
spawner

betatim:
I just came here to tell you how cool this is!

jdfreder:
Thanks @betatim !

PR #209, opened 2016-03-06
Remove selected from the state, for now

jdfreder:
Also fix some bugs in the behavior of the hover detection region
for the new cell creator tool.


jdfreder:
:cake:

rgbkrk:
Thanks for the comment on the hot region.

PR #210, opened 2016-03-06
Eslint

rgbkrk:
I bumped to the latest eslint for another project and noticed that a lot of our rules are outdated. To simplify, I bumped this to relying on the airbnb eslint and overrode the "fine IPython, have your snake case sometimes" option.


jdfreder:
I love seeing the codebase shrink!  Thanks! :cake:

PR #211, opened 2016-03-06
Switch to fluorine style agendas for kernel ops

rgbkrk:
This does a partial switch over to fluorine style agendas (observables that we can dispatch) by bridging what we currently have first.


jdfreder:
:tada: :tada: :tada:

rgbkrk:
Hopefully this is easy enough to adapt for use with generators and redux sagas. At the very least, this encapsulates an `Observable` that emits actions.

PR #214, opened 2016-03-06
provide getState()

rgbkrk:
This is a stop gap to start moving some of our native Electron actions and dispatch to rely on less of an awkward API, moving us closer to Redux. I'm going to clean up `save` and `saveAs` after this.

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -18,19 +18,19 @@ export default function createStore(initialState, reducers) {
   ).share();
 
   // Debugging time
-  store.subscribe(null, (err) => {
+
+  const stateSymbol = Symbol('state');
+  store.subscribe(state => {
+    store[stateSymbol] = state;
+  }, (err) => {
     console.error('Error in the store', err);
   });
-
-  let lastState = {};
-  store.subscribe(st => {
-    lastState = st;
-  });
+  store.getState = () => store[stateSymbol];
 
   function dispatch(action) {
     // We need the current state at this time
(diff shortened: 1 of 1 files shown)


jdfreder:
Thanks for doing this Kyle, this is much needed!

src/store/index.js
```diff
@@ -18,19 +18,19 @@ export default function createStore(initialState, reducers) {
...
 
   // Debugging time
-  store.subscribe(null, (err) => {
-    console.error('Error in the store', err);
-  });
 
-  let lastState = {};
-  store.subscribe(st => {
-    lastState = st;
+  const stateSymbol = Symbol('state');
```
StevenLangbroek, line 22: Out of curiosity, is there a reason for using `Symbol` everywhere?

rgbkrk, line 22: Symbols don't show up in `getOwnPropertyNames`, won't conflict with other strings (unique), and are the simplest properties to use in a closure. My favorite introduction to [metaprogramming with Symbols comes from Keith Cirkel's blog](http://blog.keithcirkel.co.uk/metaprogramming-in-es6-symbols/).

rgbkrk, line 22: The benefit from the constants angle is that it requires us to require the constants directly rather than short-circuiting and hard-coding a string anywhere.

StevenLangbroek, line 22: Right I get that, so it's an enforcement mechanism for proper coding.

PR #216, opened 2016-03-07
Shift button position and add titles.

rgbkrk:
Moved the buttons so that they're on either side of "add cell". While I was at it I added titles to the buttons so that hovering over them provides helper text.

Screenshot:

screenshot 2016-03-06 20 33 59

Cursor doesn't show above. May be worth giving [react-a11y](https://github.com/reactjs/react-a11y) a spin after this to highlight other spots we're not providing cues on.

### DIFF
--- a/src/components/cell/cell-creator-buttons.js
+++ b/src/components/cell/cell-creator-buttons.js
@@ -15,11 +15,11 @@ export default class CellCreatorButtons extends React.Component {
   render() {
     return (
       <div className='creator-tool'>
-        <span className='creator-label'>Add cell</span>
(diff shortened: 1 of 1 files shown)


rgbkrk:
Heh, CodeMirror provides the most warnings with `react-a11y`.

To take it for a spin yourself, `npm install react-a11y` then stick these lines in after the React and ReactDOM import:

``` js
const a11y = require('react-a11y');
a11y(React, { includeSrcNode: true, ReactDOM });
```

ivanov:
:+1:

PR #217, opened 2016-03-10
Mention some of our sponsors.

rgbkrk:
We need to be able to acknowledge contributions from everyone somehow, whether contributing code directly, providing funding (or employment!) to developers/designers/doc writers, or other resources (e.g. hosting). This handles the sponsorship of development case.

This will probably belong on our site (which doesn't exist yet). For the time being, I'm adding it as a PR here for the sake of being more transparent.


betatim:
:spades:

PR #218, opened 2016-03-10
Rxdux cleanup

kenwheeler:
- Pulling constants into their own folder, referencing them using `import * as constants`
- Cleaning up reducer model, creating root reducer
- Object.assign -> Object rest spreads
- Pulling key ops into /keys folder
- Fixing linting, configuring eslint to use babel-eslint as parser


src/reducers/index.js
```diff
@@ -1,102 +1,9 @@
 import * as fs from 'fs';
```
rgbkrk, line 1: This is an unused import now we can delete.

PR #221, opened 2016-03-11
Adapt the injected kernelspec to nbformat

### LINKED ISSUE #200: No syntax highlighting in newly created notebooks
Syntax highlighting is broken in new notebooks. If you run `npm run start` without an existing notebook you get one cell connected to a python3 kernel. However no syntax highlighting in the cell.

Digging down to the `<pre class="CodeMirror-line">` via the dev console shows the code without any tags to markup what is keyword or string or ...

rgbkrk:
This slightly fixes #200, for the case of `File -> New Notebook -> Lang`, when it matches the CodeMirror mode. The follow up to this is using the `kernel_info_request` which gives us all the metadata that belongs in the notebook document on launch. It's too bad it's not in the kernelspec itself (to raise up on Jupyter somewhere?).

### DIFF
--- a/main/launch.js
+++ b/main/launch.js
@@ -30,13 +30,25 @@ export function launch(notebook, filename) {
 }
 
 export function launchNewNotebook(kernelspec) {
-  return Promise.resolve(launch(
-    appendCell(emptyNotebook, emptyCodeCell)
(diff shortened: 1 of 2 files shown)


jdfreder:
woot!  thanks @rgbkrk !

PR #222, opened 2016-03-11
Apply lint roller to forehead

rgbkrk:
Using a lint roller.

Fun note from one of the linting errors and our ES6 classes: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#es6-classes

### DIFF
--- a/main/launch.js
+++ b/main/launch.js
@@ -9,8 +9,6 @@ export function launch(notebook, filename) {
   let win = new BrowserWindow({
     width: 800,
     height: 1000,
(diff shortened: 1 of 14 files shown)


rgbkrk:
Alllll done. No more touching lint rollers for a while.

The remaining warnings are all related to `console.*` except for one, which suggests that code cell should be a stateless function. Naively, I set `shouldComponentUpdate` to return `false` and that definitely broke it. Leaving that one alone.

PR #225, opened 2016-03-12
LFS

rgbkrk:
Start introducing LFS usage in this repository so that when developers run `npm install` (after nuking `node_modules` for instance) they don't have to re-download fonts over npm.


jdfreder:
Thanks!

rgbkrk:
9.3 MB for source-code-pro
8.6 MB for source-sans-pro
356M for material-design-icons

rgbkrk:
I... think I'm going to trim down material-design-icons to only include the ones we need.

PR #228, opened 2016-03-12
Break it up

rgbkrk:
This (hopefully) makes it more clear what the main thread is and where the renderer thread lives (notebook here).


rgbkrk:
If for any reason we switch to webpacking this up, this transition makes that a _lot_ simpler.

jdfreder:
Thanks, this is great.  I still need to get the full circle tests working for the different CLI arg patterns

jdfreder:
aka electron-compile working on electron-mocha

rgbkrk:
Sweet, yeah.

I'm writing a README in the src/ directory as well right now to clarify the distinction between `main` and `notebook`.

PR #229, opened 2016-03-12
Clarify the difference between main and renderer

rgbkrk:
Adds a simple README for when people navigate into `src`.


jdfreder:
Awesome!

PR #232, opened 2016-03-13
More requests

rgbkrk:
This adds in a few more helpers for requests and replies that seem ok to do as promises. It should work nicely at first (for us to get going on). If we ever want to cancel the action of the promise though, that's when we'll want an Observable instead.


rgbkrk:
/me whistles

Finally tried these out at the CLI, fixed some bugs.

rgbkrk:
If I spun these out as another package (to be used elsewhere), I'd probably make them return Observables and rely on the user running `toPromise` directly if they want that interface.

`enchannel-messaging` anyone?

PR #233, opened 2016-03-14
Final eslint showdown

rgbkrk:
This finishes up our eslint checks in nteract (I was only running it in `src` originally, which missed main). Also fixed the tests after changing paths earlier.


ivanov:
:tada:

PR #236, opened 2016-03-15
Bump react jupyter display area.

rgbkrk:
Tested out the latest release from @betatim, looking good!


betatim:
:tada:

betatim:
posting emoji is still way more fun than using reactions

PR #237, opened 2016-03-15
Extract menu bits

rgbkrk:
Put menu setup in another file where we handle the init. This makes it a bit easier to test. Native menu actions only emit their event as well as some attached data (it originates in the Electron main thread).

In order for everything to flow through our Flux store here, we need to make `dispatch` and `store` available to the event callback. This can be done in two ways. I went for the explicit bind to create a function that has `store` and `dispatch`.


jdfreder:
:tada:

PR #239, opened 2016-03-16
Set title via native windowing API

rgbkrk:
- Use the remote module to [get the current `BrowserWindow`](https://github.com/atom/electron/blob/master/docs/api/remote.md#remotegetcurrentwindow)
- Subscribe to the Flux store for updates to the filename and execution state

This looks a little funny when hopping between `idle <---> busy`, I look forward to someone making a better representation.

### DIFF
--- a/src/notebook/index.js
+++ b/src/notebook/index.js
@@ -15,6 +15,7 @@ import { initKeymap } from './keys/keymap';
 import { ipcRenderer as ipc } from 'electron';
 
 import { initMenuHandlers } from './menu';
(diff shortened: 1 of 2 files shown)


rgbkrk:
I'm noticing there are a lot of "nice to do" follow ups for the native APIs of OS X, Windows, and Linux like these: https://github.com/atom/electron/blob/master/docs/api/browser-window.md#winsetrepresentedfilenamefilename-os-x

betatim:
:apple: one of your five a day

PR #242, opened 2016-03-18
Removing execution from markdown cell toolbar

kenwheeler:
Because that's silly

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -39,7 +39,7 @@ class Cell extends React.Component {
         onMouseLeave={this.onMouseLeave}
       >
         {
-          this.state.showToolbar ? <Toolbar { ...this.props } /> : null
+          this.state.showToolbar ? <Toolbar type={type} { ...this.props } /> : null
         }
         {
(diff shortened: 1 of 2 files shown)


betatim:
I spent a good few minutes clicking that play button the other day wondering if it should do raw -> rendered markdown.

jdfreder:
Without this can we still go from unrendered markdown to rendered via the mouse?

kenwheeler:
@jdfreder @betatim should we introduce a new toggle button for MD cells for that behavior?

betatim:
For me the play button didn't go from raw -> rendered, which made me wonder if it should or not.

betatim:
Having a button is probably a good idea. Not sure if it should have a different symbol as I can't think of one that really conveys "rendering only" without being totally abstract (but then I also have to think for a moment which one of the new cell buttons gives me a new code cell every time I use it ...)

rgbkrk:
Definitely fine to be deleting it now and letting someone come in to make the button later.

It seems like the pencil/edit icon would work well here for entering edit mode on a markdown cell.

PR #245, opened 2016-03-20
Kernel does not shutdown when quitting app

betatim:
Send `menu:kill-kernel` IPC just before the window is closed (`close`).

If others listen for the `close` event and decide closing isn't the right
thing to do we need to revisit this. Sending it on `closed` is too late :(


rgbkrk:
This is definitely the wisest thing for us to do until we make the shutdown cleaner. Thank you @betatim!

PR #246, opened 2016-03-20
Problem: Can't tell which cell is focused

betatim:
Add ID of the next cell to the state when pressing shift+return

I can't work out how to actually focus a cell once it knows that
it is the currently focused cell. CM has a `autofocus` option
but I don't think this is exposed in `react-code-mirror`. Ideas?

### DIFF
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -168,6 +168,13 @@ export function updateCellExecutionCount(id, count) {
   };
 }
 
+export function focusNextCell(id) {
+  return {
+    type: constants.FOCUS_NEXT_CELL,
+    id,
+  };
+}
+
 export function executeCell(channels, id, source) {
   return (subject) => {
     const obs = agendas.executeCell(channels, id, source);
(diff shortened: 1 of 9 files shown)


rgbkrk:
Neat! I know @kenwheeler is working on a branch to handle the UI/UX portion of focus cell, we could merge just the action, constant, and reducer right now without the next step too.

rgbkrk:
The codemirror react package were using probably does need more (P.S. There were two different codemirror react setups, not sure if this was the best one). We're also going to need to get into the tab completion machinery of codemirror too.

betatim:
Seems like working out how to actually focus a cell (bring it into view and place a cursor in it) is tricky I am :+1: for merging this and solving the hard problem later.

CM: we could either contribute to `react-code-mirror` or be lazy and use `JedWatson/react-codemirror` which has a `options` prop that it passes along.

jdfreder:
Thanks @betatim :cake:

betatim:
:surfer:

PR #247, opened 2016-03-21
[DO NOT MERGE] Anchors for focused cells

kenwheeler:
Was experimenting with using anchors/hashes to navigate to focused cells.

This approach seemed too good to be true, and it was. Using unique IDs as element `id` attributes, and setting a `location.hash` to move the focused cell into view.

This won't work because:
- It always scrolls to the top
- It prevents upwards scrolling when focused if a previous cell is executing on an interval with output
- It doesn't persist position when focusing, i.e: when you click it jumps.

Over the next day or two, I'll be refactoring to
- Include logic and nextCell selection per #246 
- Manage focused scroll position via scrollTop 

Putting this here for reference

### DIFF
--- a/intro.ipynb
+++ b/intro.ipynb
@@ -1,28 +1,8 @@
 {
  "cells": [
   {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "## The Notable Nteract Notebook\n",
-    "\n",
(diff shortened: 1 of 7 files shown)


kenwheeler:
Sorry about the notebook commit :(

PR #251, opened 2016-03-23
Cell focus management

kenwheeler:
## Checklist
- [x] Adding focus on click
- [x] Adding codemirror input focus on cell focus
- [x] Initial focus styling
- [X] scrollTop adjustment on focus

### DIFF
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -52,7 +52,7 @@ export default class Editor extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.focused !== this.props.focused) {
+    if (this.props.focused && prevProps.focused !== this.props.focused) {
       this.refs.codemirror.editor.focus();
     }
   }
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -51,6 +51,12 @@ export default class Editor extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.focused !== this.props.focused) {
+      this.refs.codemirror.editor.focus();
+    }
+  }
+
   onChange(e) {
     if (this.props.onChange) {
       this.props.onChange(e.target.value);
--- a/src/notebook/styles/main.css
+++ b/src/notebook/styles/main.css
@@ -102,7 +102,7 @@ img
 }
 
 .cell.focused {
-  box-shadow: 0px 8px 4px rgba(0,0,0,.12);
+    box-shadow: 3px 3px 9px rgba(0,0,0,.12), -3px -3px 9px rgba(0,0,0,.12);
 }
 
 .cell:hover .input_area .cell_inputs,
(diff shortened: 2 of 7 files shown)


rgbkrk:
OMG this is gorgeous ken. Tried it just now, shift-enter is working beautifully.

Another thing to either do in a follow-on PR or as part of this is to create a new cell when a user runs shift-enter in the last cell.

src/notebook/components/cell/editor.js
```diff
@@ -51,6 +51,12 @@ export default class Editor extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.focused !== this.props.focused) {
+      this.refs.codemirror.editor.focus();
+    }
```
rgbkrk, line 57: The secret to everything here - don't re-focus if focused before.

rgbkrk, line 57: This likely needs to be `this.props.focused && prevProps.focused !== this.props.focused`, for when the component updates yet is not focused (output coming into another cell).

rgbkrk, line 57: Ah, yeah, I just had a weird flip-back-and-forth behavior between two cells just now.

rgbkrk:
Cool, last commit is perfect - able to switch between cells all while output is flowing in.

src/notebook/styles/main.css
```diff
@@ -101,6 +101,10 @@ img
     box-shadow: 1px 1px 3px rgba(0,0,0,.12), -1px -1px 3px rgba(0,0,0,.12);
 }
 
+.cell.focused {
+  box-shadow: 0px 8px 4px rgba(0,0,0,.12);
```
rgbkrk: How about

``` css
.cell.focused {
  box-shadow: 3px 3px 9px rgba(0,0,0,.12), -3px -3px 9px rgba(0,0,0,.12);
}
```

That was our old CSS for selected, courtesy @freeman-lab (I rewound the history to ddeb07d0ac6dbeb4a1a1c1bc43c33e7e5ad74f17 to find it as my refactor of `src/` ended up making finding it not be so great). It was taken out [when the selected cells feature was removed](https://github.com/nteract/nteract/pull/209/files#diff-8c4c6de8f414cb0520c4de4fd2070829R103).

rgbkrk:
One thing I'd love to see is for the hover cell styling of the input prompt to be used instead on the focused cell:

screenshot 2016-03-25 09 10 26

Main reason is that while I was using this branch, noticed that my mouse cursor was keeping the styling fixed on a cell above even though I was executing below. It was hard for me to tell where my cursor had jumped to.

rgbkrk:
For a follow up PR:
- [ ] Arrow key focus jumping 
- [x] Auto create new cell when end of cell list is reached
- [ ] scrollTop adjustment to maintain focus on parent execution

rgbkrk:
Some other related bits that Ken suggested in chat:
- Move the cell execute action to be more top level (can run on a focused cell without codemirror)
- Have the `focusNextCell` action work on Markdown

PR #252, opened 2016-03-23
Hide display area if empty.

Carreau:
Because there is no reason to show anything, and that's make the
notebook a bit more compact.

### DIFF
--- a/src/notebook/styles/main.css
+++ b/src/notebook/styles/main.css
@@ -167,6 +167,11 @@ img
     padding: 10px 10px 10px 60px;
 }
 
+.cell_display:empty
+{
+    display:none;
+}
+
 .code_cell .input_area
 {
     display: flex;


betatim:
It really is gone, can't find it even with a :mag:! :grinning:

rgbkrk:
Whoa, that's a great UI tweak!

jdfreder:
:+1:

rgbkrk:
We could do the longer form React way of doing this which is to not even render a display area on the DOM if the output is empty.

rgbkrk:
``` diff
diff --git a/src/notebook/components/cell/code-cell.js b/src/notebook/components/cell/code-cell.js
index 699c793..73046e7 100644
--- a/src/notebook/components/cell/code-cell.js
+++ b/src/notebook/components/cell/code-cell.js
@@ -46,12 +46,16 @@ const CodeCell = (props, context) => {
           focused={props.id === props.focusedCell}
         />
       </div>
-      <Display
-        className="cell_display"
-        outputs={props.cell.get('outputs')}
-        displayOrder={props.displayOrder}
-        transforms={props.transforms}
-      />
+
+      {
+        props.cell.get('outputs').isEmpty() ? null :
+          <Display
+            className="cell_display"
+            outputs={props.cell.get('outputs')}
+            displayOrder={props.displayOrder}
+            transforms={props.transforms}
+          />
+      }
     </div>
   );
 };
```

PR #256, opened 2016-03-26
Apply same bg color to cell inputs on focused cell

rgbkrk:
Otherwise it's hard to tell we're now focused on a new cell when our mouse is over a different cell.


jdfreder:
woot!

PR #257, opened 2016-03-26
ctrl-enter runs cell without focus

rgbkrk:
This also removes an extra listener we didn't need (can `e.preventDefault` on the code cell).


rgbkrk:
Prior to this, `ctrl-enter` was inserting a newline (and running).

jdfreder:
Awesome

PR #258, opened 2016-03-26
eslint touchups

rgbkrk:
Lint rolling.

We need to investigate `.setRepresentedFilename()` and other OS specific functions to know whether we have to check on them on other OSes, as suggested by @Carreau.

@jdfreder - when you are able to get to your Linux box later, can you pop open an Electron window, open the dev console and run:

```
require('remote').getCurrentWindow().setRepresentedFilename
```


jdfreder:
Will do mildu

jdfreder:
:cake: :fireworks:

PR #259, opened 2016-03-26
When at end of doc, create new cell after run cell

rgbkrk:
This creates a new cell when at the end of the document.

### DIFF
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -24,6 +24,7 @@ export default class Editor extends React.Component {
     lineNumbers: false,
     text: '',
     theme: 'composition',
+    focused: false,
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


rgbkrk:
This _seems_ to come down to a case where on initial componentDidUpdate the `prevProps` are equivalent to the next props since it's the beginning. I need to look at lifecycle stuff here.

rgbkrk:
The React lifecycle cheat sheet was super helpful here, since a new cell won't run through `componentDidUpdate`. For a brand new cell, we need to check for focused after mounted and focus if so.

rgbkrk:
thank you react lifecycle chart

rgbkrk:
I'd love to make it so that on initial notebook load we focus on the first cell. I can do that in a follow up PR.

jdfreder:
Sweet, thanks for sharing the cheat sheet too!

PR #262, opened 2016-03-26
Resolve notebook paths

### LINKED ISSUE #261: Run the kernel in the same directory as the notebook
If you open a notebook from nteract, it launches the kernel process with the current working directory to where nteract was launched from. Child processes need the path to be set properly.

jdfreder:
closes #261

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -4,6 +4,7 @@ import { launchFilename } from './launch';
 
 import { Menu } from 'electron';
 import { defaultMenu, loadFullMenu } from './menu';
+import { resolve } from 'path';
 
 app.on('window-all-closed', () => {
   // On OS X, we want to keep the app and menu bar active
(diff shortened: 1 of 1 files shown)


rgbkrk:
There we go! Now @Carreau's awesome OS X stuff handles the cli opening of a notebook well.

src/main/index.js
```diff
@@ -24,5 +25,5 @@ app.on('ready', () => {
   Menu.setApplicationMenu(defaultMenu);
   // Let the kernels/languages come in after
   loadFullMenu().then(menu => Menu.setApplicationMenu(menu));
-  notebooks.forEach(launchFilename);
+  notebooks.map(f=>resolve(f)).forEach(launchFilename);
```
rgbkrk, line 28: Could turn this into:

``` js
notebooks.forEach(f => {
  launchFilename(resolve(f))
})
```

Carreau, line 28: or `notebooks.map(resolve).forEach(launchFilename);` ?

jdfreder, line 28: You know what I like even more, `.map(resolve).foreach(launchFilename)`

rgbkrk, line 28: that is very readable

PR #267, opened 2016-03-26
Layout ground work for autocompletion

Carreau:
Ok, so for now it completes based on the words in current cell. 
And I've slightly modified the appearance of the completer to me more nteract-like colors/shape.

screen shot 2016-03-26 at 16 06 14

(side note, there is `react-codemirror` and `react-code-mirror`, we use the later one, while the former seem slightly more use and active)

### DIFF
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -36,6 +36,11 @@ export default class Editor extends React.Component {
     this.onChange = this.onChange.bind(this);
   }
 
+  componentWillMount() {
+      require('codemirror/addon/hint/show-hint');
+      require('codemirror/addon/hint/anyword-hint');
+  }
+
   componentDidMount() {
     // On first load, if focused, set codemirror to focus
     if (this.props.focused) {
@@ -82,6 +87,7 @@ export default class Editor extends React.Component {
           lineNumbers={this.props.lineNumbers}
           theme={this.props.theme}
           onChange={this.onChange}
+          extraKeys={{"Ctrl-Space": "autocomplete"}}
         />
       </div>
     );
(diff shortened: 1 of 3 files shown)


Carreau:
completion is bound to ctrl-space for now, as tab is indent.

rgbkrk:
Nice!

rgbkrk:
If you think the other codemirror react is better, feel free to switch it up. I can't recall why we chose one over the other.

src/notebook/components/cell/editor.js
```diff
@@ -36,6 +36,11 @@ export default class Editor extends React.Component {
     this.onChange = this.onChange.bind(this);
   }
 
+  componentWillMount() {
+      require('codemirror/addon/hint/show-hint');
+      require('codemirror/addon/hint/anyword-hint');
```
rgbkrk, line 41: It would be ok I think to make these be at the top level, since require is going to cache the imports anyways.

Carreau, line 41: Ok, i'll do that. I'm looking on how to write a generic codemirror hinter, especially using es6, and promises. CodeMirror seem to still use callback, and  does not use Promises anywhere, so might want to wrap the hinter a bit, as my guess is the author want to keep things compatible with old browsers without shims.

rgbkrk, line 41: That makes sense. The current API that I put in place for this [in messaging/completion](https://github.com/nteract/nteract/blob/master/src/notebook/api/messaging/completion.js) returns a promise though you can also work with the RxJS streams. I'd need to look more deeply at the codemirror hints API to understand the flow.

rgbkrk, line 41: This [little tutorial](http://langnostic.inaimathi.ca/posts/briefly-async-completions-with-code-mirror) helped me a bit, though I suppose I should dive into jupyter/notebook for an implementation. Gonna head out for a bit, good luck and I'll certainly see you tomorrow!

jdfreder:
Dude, awesome!

Issue #270, opened 2016-03-27
Handle payloads


rgbkrk:
Though slated for deprecation, [payloads](http://jupyter-client.readthedocs.org/en/latest/messaging.html#payloads-deprecated) are still in wide use.
- [x] page (via `?` in IPython)
- [x] set_next_input when replace=True (via `%load` in IPython - modifies cell source)
- [x] set_next_input when replace=False

Optional
- edit (`%edit` triggers to open a file)

Very Optional
- ask_exit

rgbkrk:
Finished off the required ones!

Tempting to do edit, though I assume that would be a rather tricky cross platform one so I'm not going to do it anytime soon. Open for someone!

PR #271, opened 2016-03-27
[payloads] set_next_input, stubs for other payloads

rgbkrk:
load

First PR in support of #270

### DIFF
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -140,9 +140,10 @@ export function removeCell(id) {
   };
 }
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
Stubbed out some warnings for the other types. Need to hone in on a design for the pager.

rgbkrk:
As a user, I never thought of the response to `?` or `??` as a pager. It was always, to me, a doc lookup.

rgbkrk:
It would be really cool if we could do the pager inline with block decorations like Atom can do. Not sure if that's possible with codemirror.

PR #280, opened 2016-03-28
Moving cell execution out to the global level

kenwheeler:
This moves cell execution back to the global level. This allows for shift+enter traversal of notebooks without requiring input focus, and supports markdown cells.


rgbkrk:
This is working really well across the whole notebook.

rgbkrk:
Actually, it looks like this is switching cells right yet not running the code in code blocks.

PR #282, opened 2016-03-28
Fixing execution

kenwheeler:
Lets try this again

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -114,14 +114,15 @@ class Notebook extends React.Component {
     e.preventDefault();
 
     const cellMap = this.props.notebook.get('cellMap');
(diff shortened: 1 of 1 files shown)


src/notebook/components/notebook.js
```diff
@@ -114,14 +114,15 @@ class Notebook extends React.Component {
...
+    const id = this.props.focusedCell;
+    const cell = cellMap.get(id);
 
     if (e.shiftKey) {
       this.context.dispatch(focusNextCell(this.props.focusedCell));
     }
 
     this.context.dispatch(
-      executeCell(this.props.channels, cell.get('id'), cell.get('source'))
+      executeCell(this.props.channels, id, cell.get('source'))
```
rgbkrk, line 125: Ah, that makes sense.

PR #284, opened 2016-03-29
Rely on commander to argparse, handle new notebook

rgbkrk:
Switching us over to commander as well as bringing back the "launch new notebook" case.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1,10 +1,23 @@
 import app from 'app';
 
-import { launchFilename } from './launch';
+import {
+  launchFilename,
+  launchNewNotebook,
+} from './launch';
 
 import { Menu } from 'electron';
 import { defaultMenu, loadFullMenu } from './menu';
 import { resolve } from 'path';
+
+const program = require('commander');
+const version = require('../../package.json').version;
+
+program
+  .version(version)
+  .option('-k', '--kernel [kernel]', '')
+  .parse(process.argv);
+
+const notebooks = program.args;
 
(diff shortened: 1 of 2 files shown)


src/main/index.js
```diff
@@ -1,11 +1,24 @@
...
+import {
+  launchFilename,
+  launchNewNotebook,
+} from './launch';
 
 import { Menu } from 'electron';
 import { defaultMenu, loadFullMenu } from './menu';
 import { resolve } from 'path';
 
+const program = require('commander');
```
jdfreder, line 12: I :heart: commander

src/main/index.js
```diff
@@ -13,17 +26,14 @@ app.on('window-all-closed', () => {
...
   // Get the default menu first
   Menu.setApplicationMenu(defaultMenu);
   // Let the kernels/languages come in after
   loadFullMenu().then(menu => Menu.setApplicationMenu(menu));
-  notebooks.filter(Boolean).forEach(f => launchFilename(resolve(f)));
+  if (notebooks <= 0) {
+    launchNewNotebook('python3');
+  } else {
+    notebooks.filter(Boolean).forEach(f => launchFilename(resolve(f)));
+  }
```
jdfreder, line 38: Much better

rgbkrk, line 38: I wondered about providing a `--kernel` argument, then realized it wouldn't make sense when you have a list of notebooks.

jdfreder, line 38: --kernel makes sense for the new notebook feature

PR #285, opened 2016-03-29
Remove unused babel plugin for transforming export

rgbkrk:
We're not using this anywhere as far as I can tell. All the notebook ops I did seemed to be working well.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,6 @@
     ],
     "plugins": [
       "transform-class-properties",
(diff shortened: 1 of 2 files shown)


rgbkrk:
I'm working on packaging right now and recognizing that trimming down our babel plugins makes our releases much simpler (because of defaults in `electron-compile-prebuilt`).

jdfreder:
:fish_cake:

PR #289, opened 2016-03-31
Make sure we break-word on .cell_display

rgbkrk:
Noticed some long flowing sections that needed some wrapping to not escape the cell area.


jdfreder:
Simple enough

PR #290, opened 2016-04-01
State the node 5.x requirement earlier for install

rgbkrk:


rgbkrk:
/cc @willingc 

I'm sure you'll have more substantial and clear docs than what I put together here. This is a stopgap.

betatim:
Merge first, substance later!

PR #293, opened 2016-04-03
Pager

rgbkrk:
screenshot 2016-04-03 10 47 56

Inline docs via the pager part of the message protocol!


rgbkrk:
This is only my first stab at it. I'd love to have an alternative design where docs could appear inline where it was run.

¯\_(ツ)_/¯

PR #296, opened 2016-04-04
More bump and trim

rgbkrk:
Even more dependency bumping and trimming.

Also fixed a test that needed another prop.


rgbkrk:
Went ahead and bumped our version of electron, since it has good bug fixes and changes nothing about the ABI version. It does rely on node 5.10.0, though you can be on a prior 5.x release.

jdfreder:
Thanks :sushi: !

PR #300, opened 2016-04-09
Bump react

rgbkrk:
This is currently working quite well but I noticed that neither of the upstream code-mirror forks have React 15 bumped yet. I'll submit a patch for the one we're using and pay attention to the other one that @Carreau suggested.


rgbkrk:
You will see this warning:

```
$ npm install
npm WARN react-code-mirror@3.0.6 requires a peer of react@>=0.12.0 <0.15.0 but none was installed.
```

Do you think that's ok for now?

rgbkrk:
Alright, patch submitted upstream: https://github.com/ForbesLindesay/react-code-mirror/pull/18

jdfreder:
The warning is probably okay, people may mistake it as an error though

package.json
```diff
@@ -45,7 +45,6 @@
     "react-dnd-html5-backend": "^2.1.2",
     "react-dom": "^15.0.1",
     "react-jupyter-display-area": "^0.3.0",
-    "react-markdown": "^2.1.1",
```
rgbkrk, line 48: Took this out because of https://github.com/rexxars/commonmark-react-renderer/issues/16

rgbkrk:
@jdfreder If you think so, I'm game for going ahead with this then.

jdfreder:
Sweettttt!  :cake:

PR #306, opened 2016-04-13
Initial theming architecture

### LINKED ISSUE #22: Theming

kenwheeler:
This is the initial setup for theming.

TODO:
- [X] Add variables to current CSS file
- [X] Set up pretty variable definitions for light/dark

Closes #22.


src/notebook/index.js
```diff
@@ -52,8 +53,21 @@ ipc.on('main:load', (e, launchData) => {
...
+      storage.get('theme', (error, data) => {
+        if (error) throw error;
+        if (data.length > 0) return;
+        this.setState({
+          theme: data.theme,
+        });
+      });
+      ipc.on('menu:theme', (ev, theme) => {
+        storage.set('theme', { theme });
+        this.setState({ theme });
```
rgbkrk, line 81: This will get overriden by the `this.setState` from the above `store.subscribe` though right?

src/notebook/index.js
```diff
@@ -71,12 +85,15 @@ ipc.on('main:load', (e, launchData) => {
...
               this.state.notebook &&
               <Notebook
+                theme={this.state.theme}
                 notebook={this.state.notebook}
                 channels={this.state.channels}
                 cellPagers={this.state.cellPagers}
                 focusedCell={this.state.focusedCell}
               />
             }
+            <link rel="stylesheet" href={`./styles/theme-${this.state.theme}.css`} />
```
rgbkrk: Ok that's sweet and simple.

src/notebook/index.js
kenwheeler, line 81: Nope. setState merges.

rgbkrk, line 81: I learn much from you, thank you.

rgbkrk:
What's the word on this? Would it help if I put up a mock that had a dark theme that was similar to our current theme?

kenwheeler:
Just pushed up what I have so far, trying to get the first iteration wrapped up this morning. Thoughts on the theme so far?

kenwheeler:
screen shot 2016-04-19 at 6 39 01 am

captainsafia:
It's not in your screenshot, but is the cell toolbar supposed to look a little funky.

screen shot 2016-04-19 at 5 59 26 am

kenwheeler:
Weird, looks like your icons aren't loading: 
screen shot 2016-04-19 at 8 29 16 am

betatim:
@captainsafia the fonts moved to their own package, did you install that?

captainsafia:
@betatim It was because of the conflicts on the branch. I was `npm install`ing off the `package.json` without `nteract-assets`.

rgbkrk:
Our transition to the assets folder off of LFS was awkward. Hopefully we can clear right through that from here.

rgbkrk:
The only thing I'm noticing missing here is that the cells aren't flush when "inactive", as they are with our default theme.

kenwheeler:
Alright, I think we're ready to roll

rgbkrk:
I must be missing something. Assets aren't all loaded.

screenshot 2016-04-21 09 26 03

rgbkrk:
~~There are no errors reported at the command line for `npm run start intro.ipynb` to say anything about the menu. Digging into that now.~~

Nevermind that one, looks like I wasn't looking in the right spot. It's View --> Theme

rgbkrk:
Looks like the CSS is one path too deep, tackling that now.

jdfreder:
:+1:

src/notebook/index.js
```diff
@@ -64,8 +65,21 @@ ipc.on('main:load', (e, launchData) => {
...
     constructor(props) {
       super(props);
-      this.state = {};
+      this.state = {
+        theme: 'light',
+      };
       store.subscribe(state => this.setState(state));
+      storage.get('theme', (error, data) => {
+        if (error) throw error;
+        if (data.length > 0) return;
```
rgbkrk, line 74: Noticed that data.length is not defined on initial state in the storage, which results in this being false and the state being set to undefined.

Disco mode.

PR #307, opened 2016-04-13
[DISCUSS] Execute Javascript at the page level in an IFrame

jdfreder:
DO NOT MERGE 

This PR demonstrates how Javascript can be rerouted from cell outputs to the document level.  I'm opening it as a discussion point, which may be valuable when we begin to explore the idea of document windowing because it solves the problem of what to do when a cell with JS scrolls out of the window and back into the window.

The alternative, which I will open another PR for, is to run the JS in per cell output area iframes.  The benefit to this is that the Javascript will have access to the output area of the cell that executed it.  A complication, but probably manageable, is that the global context will have to be explicitly shared between iframes.


src/notebook/components/html-scriptless.js
```diff
@@ -0,0 +1,30 @@
...
+
+  componentDidMount() {
+    // Remove the script tags from the HTML repr.  Script tags should be handled
+    // by the page level IFrame.  Content must be wrapped in a div in order to
+    // parse correctly.
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(`<div>${this.props.data}</div>`, 'text/xml');
+    Array.prototype.forEach.call(doc.querySelectorAll('script'), scriptEl => {
+      scriptEl.remove();
+    });
```
rgbkrk, line 17: Our current setup relies on setting innerHTML (as is done via React's `setDangerouslyInnerHTML`), which means inline scripts don't actually run. I'm not sure about handlers like `onload` or `onerror` though.

src/notebook/components/javascript-frame.js
```diff
@@ -0,0 +1,36 @@
...
+      .map(bundle => {
+        if (bundle.get('application/javascript')) {
+          return <Javascript data={ bundle.get('application/javascript') } />;
+        } else if (bundle.get('text/html')) {
+          return <HTMLScriptsDisplay data={ bundle.get('text/html') } />;
+        }
+        return null;
+      })
+      .filter(comp => comp !== null);
+
```
rgbkrk, line 31: This looks like it will run all javascripts on every render call.

src/notebook/components/notebook.js
```diff
@@ -22,8 +24,11 @@ class Notebook extends React.Component {
   };
 
   static defaultProps = {
-    displayOrder,
-    transforms,
+    displayOrder: displayOrder
+      .remove('application/javascript'),
+    transforms: transforms
+      .remove('application/javascript')
+      .set('text/html', HTMLScriptlessDisplay),
```
rgbkrk, line 31: :+1:

rgbkrk:
Oh noes! Where is this going!?

jdfreder:
An alternative PR which does IFrames per output area

jdfreder:
I'll attempt to share globals across the IFrames....  Not sure if I will have success though.

PR #309, opened 2016-04-14
More explicit install instructions

betatim:
So that even people who don't read them very carefully notice that there are several steps. _coughcough_

PR #310, opened 2016-04-14
More instructions for git LFS

Carreau:
Apparently you need to `lfs fetch` first or it fails...

And you can not track the file if they are on LFS I guess...
## 

THanks to @betatim , #309 instructions fixed things for me, but I need some extra steps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ build/Release
 node_modules
 
 .ipynb_checkpoints
+
+# LFS things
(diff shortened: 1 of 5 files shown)


rgbkrk:
This totally takes the file out of git _and_ git lfs, as far as I can tell. How about we go with @jdfreder's #311, and make this repo simpler to digest?

rgbkrk:
Since #311 overlap in the right ways, I'll just merge this and clean up the README to take out the LFS usage.

PR #314, opened 2016-04-16
Merge neighbouring cells

betatim:
Merge two cells into one.

Couldn't find a way to hide the merge icon for the last cell 😢

L117 and L123 seem pretty hacky/not pretty to look at. Ideas/advice on good JS style?

### DIFF
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -164,6 +164,13 @@ export function createCellAppend(cellType) {
   };
 }
 
+export function mergeCellAfter(id) {
+  return {
+    type: constants.MERGE_CELL_AFTER,
+    id,
+  };
+}
+
 export function updateCellExecutionCount(id, count) {
   return {
     type: constants.UPDATE_CELL_EXECUTION_COUNT,
--- a/src/notebook/components/cell/cell-creator-buttons.js
+++ b/src/notebook/components/cell/cell-creator-buttons.js
@@ -1,6 +1,10 @@
 import React from 'react';
 
-import { createCellAfter, createCellBefore, createCellAppend } from '../../actions';
+import {
+  createCellAfter,
+  createCellBefore,
+  createCellAppend,
+  mergeCellAfter } from '../../actions';
 
 export default class CellCreatorButtons extends React.Component {
   static propTypes = {
@@ -17,6 +21,7 @@ export default class CellCreatorButtons extends React.Component {
     this.createCodeCell = this.createCell.bind(this, 'code');
     this.createTextCell = this.createCell.bind(this, 'markdown');
     this.createCell = this.createCell.bind(this);
+    this.mergeCell = this.mergeCell.bind(this);
   }
 
   shouldComponentUpdate() {
@@ -36,7 +41,17 @@ export default class CellCreatorButtons extends React.Component {
     }
   }
 
+  mergeCell() {
+    this.context.dispatch(mergeCellAfter(this.props.id));
+  }
+
   render() {
+    const mergeButton = (
+      <button onClick={this.mergeCell} title="merge cells">
+        <i className="material-icons" style={{ transform: 'rotate(90deg)' }}>
+          compare_arrows
+        </i>
+      </button>);
     return (
       <div className="creator-tool">
         <button onClick={this.createTextCell} title="create text cell">
@@ -46,6 +61,7 @@ export default class CellCreatorButtons extends React.Component {
         <button onClick={this.createCodeCell} title="create code cell">
           <i className="material-icons">code</i>
         </button>
+        {this.props.above ? null : mergeButton}
       </div>
     );
   }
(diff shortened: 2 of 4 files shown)


rgbkrk:
Oh great, I've been hoping for that. I'll check out the code thoroughly since you're asking.

betatim:
Don't merge yet, there is a left over action that should have been deleted

src/notebook/actions/index.js
```diff
@@ -164,6 +164,20 @@ export function createCellAppend(cellType) {
...
+    id,
+  };
+}
+
+export function mergeCellAfter(id) {
+  return {
+    type: constants.MERGE_CELL_AFTER,
+    id,
+  };
+}
```
rgbkrk, line 172: 👍

src/notebook/components/cell/cell-creator-buttons.js
```diff
@@ -36,7 +42,21 @@ export default class CellCreatorButtons extends React.Component {
...
+    }
+  }
+
   render() {
+    const mergeButton = (
+      <button onClick={this.mergeCell} title="merge cells">
+        <i className="material-icons" style={{ transform: 'rotate(90deg)' }}>
+          compare_arrows
+        </i>
+      </button>);
```
rgbkrk, line 54: Ah cool, new button.

rgbkrk, line 54: What happens when the cells are of different type?

betatim, line 54: Right now they get merged, and it looks weird ;)

src/notebook/reducers/document.js
```diff
@@ -103,6 +103,29 @@ export default {
...
+    const cellOrder = notebook.get('cellOrder');
+    const cellMap = notebook.get('cellMap');
+    const index = cellOrder.indexOf(id);
+    // do nothing if this the last cell
+    if (cellOrder.size === index + 1) {
+      return state;
+    }
+    const nextId = cellOrder.get(index + 1);
+    const source = `${cellMap.get(id).get('source')}
+${cellMap.get(nextId).get('source')}`;
```
rgbkrk: Would a `\n` style newline still work here or do you have to break it apart for backtick strings?

src/notebook/reducers/document.js
```diff
@@ -103,6 +103,29 @@ export default {
...
   },
+  [constants.MERGE_CELL_AFTER]: function mergeCellAfter(state, action) {
+    const { id } = action;
+    const { notebook } = state;
+    const cellOrder = notebook.get('cellOrder');
+    const cellMap = notebook.get('cellMap');
+    const index = cellOrder.indexOf(id);
+    // do nothing if this the last cell
+    if (cellOrder.size === index + 1) {
+      return state;
```
rgbkrk, line 114: Good to cover this case. In addition, we can make sure the button doesn't appear in the UI for that.

betatim:
Ok, the `mergeCellBefore` is gone.

src/notebook/reducers/document.js
```diff
@@ -103,6 +103,29 @@ export default {
...
+    }
+    const nextId = cellOrder.get(index + 1);
+    const source = `${cellMap.get(id).get('source')}
+${cellMap.get(nextId).get('source')}`;
+
+    return {
+      ...state,
+      notebook:
+        commutable.removeCell(
+          commutable.updateSource(notebook, id, source),
```
rgbkrk, line 124: Looks like the right approach.

src/notebook/reducers/document.js
betatim: Started with

```
const source = cellMap.get(id).get('source') + '\n' + cellMap.get(nextId).get('source');
```

but eslint doesn't like that and recommend this backtick style. Means you can't indent L118 though as that whitespace becomes part of the variable's contents.

rgbkrk:
This reminds me that I need to get my split cell branch up. I still have it sitting around, haven't pushed since I got distracted with right-click context menus.

src/notebook/reducers/document.js
betatim, line 124: It gets clunky when you also want to clear the output of the cell you are keeping. Actually not sure if we should clear it or not ... right now I went for keeping the output as the code for clearing it looked ugly.

src/notebook/reducers/document.js
rgbkrk: Alternatively:

``` js
const source = cellMap.getIn([id, 'source']).concat('\n', cellMap.getIn([nextId, 'source']))
```

src/notebook/components/cell/cell-creator-buttons.js
betatim, line 54: We merge no matter what the type, just like jupyter does.

src/notebook/reducers/document.js
rgbkrk, line 124: Ah, hmmm. Not sure of the right thing to do. Seems like you'd keep the output of the bottom cell?

betatim, line 124: jupyter keeps the output of the bottom cell. Not sure which one of the three options is best.

I'll ponder it while 😴

betatim:
:shipit:'ing done. Merge at will.

rgbkrk:
Merged the Merger!

PR #315, opened 2016-04-16
Publish to gist

rgbkrk:
Adds the ability to publish to GitHub. Only anonymous gists for now.

Considered prototypal.

### DIFF
--- a/src/notebook/publication/github.js
+++ b/src/notebook/publication/github.js
@@ -1,0 +1,63 @@
+const Rx = require('@reactivex/rxjs');
+const commutable = require('commutable');
+const path = require('path');
+
+const shell = require('electron').shell;
+
+const actions = require('../actions');
+
+function notifyUser(filename, gistURL, gistID) {
+  const title = 'Gist uploaded';
+  const options = {
+    body: `${filename} ready at ${gistURL}`,
+  };
+  const notification = new Notification(title, options);
+  notification.onclick = function openGist() {
+    shell.openExternal(`https://nbviewer.jupyter.org/${gistID}`);
+  };
+}
+
+function createGistCallback(hotOffThePresses, agenda, filename) {
+  return function gistCallback(err, response) {
+    if (err) {
+      agenda.error(err);
+      agenda.complete();
+      return;
+    }
+
+    const gistID = response.id;
+    const gistURL = response.html_url;
+
+    notifyUser(filename, gistURL, gistID);
+    if (hotOffThePresses) {
+      agenda.next(actions.overwriteMetadata('gist_id', gistID));
+    }
+    agenda.next(actions.doneUploading);
+  };
+}
+
+export function publish(github, notebook, filepath) {
+  return Rx.Observable.create((agenda) => {
+    const files = {};
+    const notebookString = JSON.stringify(commutable.toJS(notebook), undefined, 1);
+    const filename = path.parse(filepath).base;
+    files[filename] = { content: notebookString };
+
+    agenda.next(actions.startedUploading);
+
+    // Already in a gist, update the gist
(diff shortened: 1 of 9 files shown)


rgbkrk:
gistitnow

Sadly this doesn't show the publish menu, which looks like this:

screenshot 2016-04-16 16 34 41

src/notebook/publication/github.js
```diff
@@ -0,0 +1,63 @@
...
+  return Rx.Observable.create((agenda) => {
+    const files = {};
+    const notebookString = JSON.stringify(commutable.toJS(notebook), undefined, 1);
+    const filename = path.parse(filepath).base;
+    files[filename] = { content: notebookString };
+
+    agenda.next(actions.startedUploading);
+
+    // Already in a gist, update the gist
+    if (notebook.hasIn(['metadata', 'gist_id'])) {
```
betatim, line 53: I like this! Not just publish to gist but also update to gist!

rgbkrk:
Missing pieces to follow on with:
- Auth
  - For GitHub, we could assume an environment variable for the token (as other services use) 
- Public vs. private gist selection

src/notebook/publication/github.js
rgbkrk, line 53: Stolen straight out of @minrk's extension to the Jupyter notebook. :smile:

rgbkrk:
Of note: the gist extension on the jupyter notebook defaults to public. I defaulted to "secret" gist here, as well as anonymous (for the time being).

src/notebook/publication/github.js
rgbkrk, line 53: Since this defaults to anonymous, it may not work for updating unless authed.

rgbkrk:
This now authenticates with GitHub if `GITHUB_TOKEN` is set as an environment variable, which lets you update the same gist each time you publish.

Let me know what you think of the flow.

I'm now mildly irritated with nbviewer's caching. :wink: Tempted to switch the resulting URL to be the gist URL or creating a client side renderer using nteract components.

src/notebook/reducers/document.js
```diff
@@ -3,6 +3,10 @@ import * as uuid from 'uuid';
 
 import * as constants from '../constants';
 
+import Immutable from 'immutable';
+
+const noop = state => state;
```
jdfreder, line 8: Should have used `nop` here ;)

jdfreder:
> This now authenticates with GitHub if GITHUB_TOKEN is set as an environment variable, which lets you update the same gist each time you publish.

I think eventually it will make sense to expose this in the UI somewhere

jdfreder:
> I'm now mildly irritated with nbviewer's caching.  Tempted to switch the resulting URL to be the gist URL or creating a client side renderer using nteract components.

Why?  When you republish, does it not create another guid for the gist?

src/notebook/reducers/document.js
rgbkrk, line 8: As name or was there another function?

rgbkrk:
When you publish again, it updates the gist (new commit).

rgbkrk:
We definitely need to figure out an auth flow here.

src/notebook/publication/github.js
Carreau, line 53: You will need a way to "publish to new gist" or actually show "update gist" otherwise if you copy a notebook, you will have no way to distinguish the original from a copy and always clobber.

rgbkrk, line 53: You're so right @Carreau. Possible flow:

```
Publish
  -->
    to same Gist...
    new Gist...
```

Ugly...

Hmmm... What if it was in the file menu like this:

```
File
  -->
    Save
    Save As...
    Publish
    Publish To...
```

Where `Publish` knows where you were previously publishing to.

PR #317, opened 2016-04-17
Set app name to 'nteract'

captainsafia:
The application name under the menubar listed "Electron" as the app, this sets it to nteract.

Note that this does not change the parent menu item (the thing displayed in the OS X menubar) as that must be configured in `electron-packager`.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -206,7 +206,9 @@ export const help = {
   ],
 };
 
-const name = app.getName();
+const name = 'nteract';
+app.setName(name);
+
 export const named = {
   label: name,
   submenu: [


jdfreder:
Thanks! 💯 🎆

rgbkrk:
This makes me wonder where the getName was pulling it from before.

captainsafia:
> This makes me wonder where the getName was pulling it from before.

Same. It should've used `name` from `package.json` after checking to see if `productName` was null per [this](https://github.com/electron/electron/blob/f403950503eb82bcfa230b13a909572beff75fc0/default_app/main.js#L241). For some strange reason, maybe Electron is pulling the name from _its_ `package.json` inside `node_modules`?

captainsafia:
I found [this](https://github.com/electron/electron/issues/3391), but can't confirm its the cause.

¯_(ツ)_/¯

PR #318, opened 2016-04-18
Try to get completions working.

Carreau:
I get roughly codemirror to display hardcoded completions, i found the completions messages in `api/messaging`, 

But I'm getting stuck in the Rxjs/action/agenda layers, I'll appreciate pointers.

### DIFF
--- a/src/notebook/api/messaging/completion.js
+++ b/src/notebook/api/messaging/completion.js
@@ -36,13 +36,12 @@ export default function complete(channels, code, cursorPos) {
   };
 
   const p = shell
-    .logInfo()
     .childOf(message)
     .ofMessageType('complete_reply')
     .pluck('content')
     .first()
     .toPromise();
-
+  
   shell.next(message);
   return p;
 }
--- a/src/notebook/api/messaging/completion.js
+++ b/src/notebook/api/messaging/completion.js
@@ -36,6 +36,7 @@ export default function complete(channels, code, cursorPos) {
   };
 
   const p = shell
+    .logInfo()
     .childOf(message)
     .ofMessageType('complete_reply')
     .pluck('content')
--- a/src/notebook/api/messaging/index.js
+++ b/src/notebook/api/messaging/index.js
@@ -55,7 +55,7 @@ export function childOf(parentMessage) {
     // since we're in an arrow function `this` is from the outer scope.
     // save our inner subscription
     const subscription = this.subscribe(msg => {
-      console.info('looking for complete_reply...', msg.parent_header.msg_type);
+      console.info('looking for complete_reply...', msg.header.msg_type);
       if (!msg.parent_header || !msg.parent_header.msg_id) {
         subscriber.error(new Error('no parent_header.msg_id on message'));
         return;
--- a/src/notebook/api/messaging/index.js
+++ b/src/notebook/api/messaging/index.js
@@ -44,20 +44,6 @@ export function msgSpecToNotebookFormat(msg) {
   });
 }
 
-export function logInfo(){
-  return Observable.create(subscriber => {
-    const subscription = this.subscribe(msg => {
-      console.info('looking for complete_reply...', msg.header.msg_type);
-    },
-    err => subscriber.error(err),
-    () => subscriber.complete());
-
-    return subscription;
-  });
-}
-
-Observable.prototype.logInfo = logInfo;
-
 /**
  * childOf filters out messages that don't have the parent header matching parentMessage
  * @param  {Object}  parentMessage Jupyter message protocol message
--- a/src/notebook/api/messaging/index.js
+++ b/src/notebook/api/messaging/index.js
@@ -44,6 +44,20 @@ export function msgSpecToNotebookFormat(msg) {
   });
 }
 
+export function logInfo(){
+  return Observable.create(subscriber => {
+    const subscription = this.subscribe(msg => {
+      console.info('looking for complete_reply...', msg.header.msg_type);
+    },
+    err => subscriber.error(err),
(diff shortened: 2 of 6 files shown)


rgbkrk:
Ok, awesome. I'll be glad to help.

Carreau:
> Ok, awesome. I'll be glad to help.

Great, thanks.  I moved forward, I achieved to call successfully to the completion, and send a complete_request though, no complete reply seem to ever reach back.

I'll cleanup what I have an push.

src/notebook/api/messaging/index.js
```diff
@@ -55,6 +55,7 @@ export function childOf(parentMessage) {
     // since we're in an arrow function `this` is from the outer scope.
     // save our inner subscription
     const subscription = this.subscribe(msg => {
+      console.info('looking for complete_reply...', msg.parent_header.msg_type);
```
Carreau: so here I watch for all messages, I never see a complete-reply.

src/notebook/api/messaging/index.js
```diff
@@ -55,6 +55,7 @@ export function childOf(parentMessage) {
     // since we're in an arrow function `this` is from the outer scope.
     // save our inner subscription
     const subscription = this.subscribe(msg => {
+      console.info('looking for complete_reply...', msg.header.msg_type);
```
Carreau: Also, not only do I never see a complete_reply, but I see complete_request only after doing 1 cell execution. 

Is that normal ?

src/notebook/components/notebook.js
```diff
@@ -5,7 +5,9 @@ import HTML5Backend from 'react-dnd-html5-backend';
 
 import DraggableCell from './cell/draggable-cell';
 import CellCreator from './cell/cell-creator';
-import { executeCell, focusNextCell, moveCell } from '../actions';
+import { executeCell, focusNextCell, moveCell, getCompletions} from '../actions';
+
+import complete from '../api/messaging/completion';
```
Carreau, line 10: I'm confused about `import complete` vs `import {complete}` and what default export are, and why sometime modules in the debugger appear as `(0, _something.attribute)`.

rgbkrk, line 10: If the module only exports using `export default function blah()` then you can `import blah from './wherever'`. If it's a named import you have to use the destructuring assignment.

src/notebook/api/messaging/index.js
rgbkrk: Hmmmm.

This sounds like a case of not subscribing before sending. The channels require that you subscribe to the one you expect to receive on _before_ sending. I've probably made this more difficult with that initial promise implementation I threw in api/messaging.

rgbkrk: Where does the send occur?

src/notebook/components/notebook.js
jdfreder, line 10: @Carreau I don't know if this is a valid way of describing it, but I think of it like this:

``` js
import c from './a';
```

is like the following in Python

``` python
import a as c
```

where

``` js
import {b} from './a';
```

is like the following in Python

``` python
from a import b
```

src/notebook/api/messaging/completion.js
```diff
@@ -36,6 +36,7 @@ export default function complete(channels, code, cursorPos) {
   };
 
   const p = shell
+    .logInfo()
     .childOf(message)
     .ofMessageType('complete_reply')
```
Carreau, line 40: @rgbkrk refactored to logInfo only there, the send is done on line 46 below. 

This never get any messages apparently.

src/notebook/components/notebook.js
Carreau, line 10: Make sens, that was roughly what I was thinking. Still weird import syntax, especially for completion, as the completer will have a hard time figuring out what you want to import before you tell it the file or module...

Carreau:
`https://github.com/codemirror/CodeMirror/pull/3974` will allow us to use Promises with codemirror in completions.

Carreau:
Still no clue as of why I don't get a `completion_reply`, @jdfreder ?

jdfreder:
Did you make sure to subscribe to the channel before sending the message?  Also, have to verified that the completion reply is getting sent by the kernel?

Carreau:
> Did you make sure to subscribe to the channel before sending the message?

Yep:

``` js
  /// create message
  const message = createMessage('complete_request');
  message.content = {
    code,
    cursor_pos: cursorPos,
  };

  /// subscribe
  const p = shell
    .logInfo()
    .childOf(message)
    .ofMessageType('complete_reply')
    .pluck('content')
    .first()
    .toPromise();

  /// send message
  shell.next(message);
  return p;
```

I'll modify the ipykernel to see log if the completion request arrives. I assumed you had that tested somewhere.

Carreau:
Yeah ! It works ! But I have to run.

jdfreder:
Sweet!  I'm excited to try this! Thanks @Carreau !

src/notebook/components/cell/editor.js
```diff
@@ -70,8 +72,11 @@ export default class Editor extends React.Component {
       this.context.dispatch(updateCellSource(this.props.id, e.target.value));
     }
   }
-
   render() {
+    const extraKeys = {"Ctrl-Space": "autocomplete"}
```
jdfreder, line 76: At some point we will want the keymap to handle things like this.  For now we may want to change this to tab (if we can easily, without breaking indent).

rgbkrk:
Whoa! It's alive! 

screenshot 2016-04-22 08 46 43

Extremely well done @Carreau!

PR #320, opened 2016-04-18
Add dist scripts for linux build targets

jdfreder:


rgbkrk:
Sweet packaging.

package.json
```diff
@@ -10,10 +10,15 @@
...
     "start": "electron .",
-    "prebuild": "rimraf es5",
-    "build": "babel src --out-dir es5",
+    "prebuild": "rimraf build",
+    "build": "babel src --out-dir build",
     "build:watch": "npm run build -- --watch",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build",
+    "predist": "npm run build && rimraf dist",
+    "postdist": "npm install",
```
jdfreder, line 18: @rgbkrk you may be wondering why this step exists ...  It counters the `--prune` flag which runs `npm prune --production` to remove dev deps

rgbkrk, line 18: Thank you.

The steps for a new contributor right now are:

```
npm install
npm run build
```

and if that same person wants to create a release

```
npm install
npm run build
npm run dist
```

jdfreder, line 18: No,

Clone and 

```
npm install
```

If they want to develop

```
npm run build:watch
```

and if they want to release

```
npm run dist:linux64
```

dist will run build.

rgbkrk, line 18: Ah, cool. If we end up wanting another that chains these together we can do that too. Looking good. Thanks for breaking this down into nice small pieces for us to compose together.

PR #322, opened 2016-04-18
Handle opening files directly from native OS

rgbkrk:
Relying on `open-file` event. Couldn't be any simpler.

You'll need a built copy of nteract to use this (with our new dist hooks).

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -26,6 +26,11 @@ app.on('window-all-closed', () => {
   }
 });
 
+app.on('open-file', (event, path) => {
+  event.preventDefault();
+  launchFilename(resolve(path));
+});
+
 app.on('ready', () => {
   // Get the default menu first
   Menu.setApplicationMenu(defaultMenu);


captainsafia:
This brings tears to my eyes! :')

jdfreder:
This is great!!!

rgbkrk:
I should have recorded a GIF with this, I'll do that later when I'm on one monitor.

rgbkrk:
Not sure how one registers openers, we should figure that out too...

betatim:
Ah, this answer my naive question of: where did you hide the magic of informing OSX that `nteract` is the new game in town for `.ipynb`.

jdfreder:
> Not sure how one registers openers, we should figure that out too...

On Windows that's usually done by the installer, we could put it off for the installer to do on OSX too.

Carreau:
Not sure what that does. Double click on `.ipynb` still open MacVim for me.

jdfreder:
> Not sure what that does. Double click on .ipynb still open MacVim for me.

The double click require the filetype to be associated with the app, which nteract doesn't do yet...

PR #325, opened 2016-04-19
Require testing coverage for each milestone.

rgbkrk:
Earlier @jdfreder talked about making a public pre-alpha (with downloadable links within GitHub releases). I countered that we should likely exhibit some amount of testing coverage now that we've fleshed out more of the experience here, since we would want new developers to have a nice safety harness with tests (not that we'll be able to hit all the functional and integration tests in the same degree - we'll likely need play testing too).

This establishes some baseline coverage for each coming milestone:
- Pre-alpha: > 50%
- Alpha: > 70%
- Beta: > 90%

These can be regarded as pure line-based testing coverage since they're easy to measure. We'll certainly end up handling more edge cases and path coverage. This gets us started so we have something to hold ourselves accountable to.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -16,6 +16,7 @@ usability of our notebook. It is expected that we'll make lots of pre-alpha
 * Specification Compliance
     * Full compliance with Notebook format v4
     * Full compliance with message spec v5
+* Testing coverage > 50%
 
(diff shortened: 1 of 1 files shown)


jdfreder:
:cake: thanks kyle, here's some cake for you

PR #326, opened 2016-04-19
remove osx32 dist

minrk:
The last 32b-only Mac shipped in 2006, so wait until someone asks for a 32b OS X app before committing to supporting one.


rgbkrk:
Great, I wondered the same. I :heart: supporting less.

PR #328, opened 2016-04-19
Why osx64 is there is only 1 osx now ?

Carreau:


rgbkrk:
Yay for less typing on the CLI to build a dist locally for us OS X users. :smile:

Carreau:
In [1]: len('Yay for less typing on the CLI to build a dist locally for us
OS X users. :smile:')/2
Out[1]: 40.5

You just wasted 40 builds.

PR #331, opened 2016-04-19
Fix refs in tests

jdfreder:
Makes the tests pass


jdfreder:
Thanks!

rgbkrk:
While I agree that we should do functional tests, I'd rather test `src/` directly for quicker dev workflow and feedback (as well as not having to run the build to run a single test).

jdfreder:
Lets depend on src for the unit tests, and build for renderer dependent tests

PR #332, opened 2016-04-19
Switch tests back to direct src tests

rgbkrk:
Since we're testing components and logic, not integration or functional tests.


rgbkrk:
Closing in favor of #331

PR #335, opened 2016-04-19
Round trip test, check for at least one cell.

jdfreder:


rgbkrk:
Reasonable first pass.

PR #341, opened 2016-04-20
Side-step electron mocha, use mocha direct.

rgbkrk:
For the unit tests, rely on `mocha` directly. This will allow for `istanbul` testing.

I took out the hokey-pokey notebook test for a few reasons:
- Any code that `require`s CodeMirror relies on a real DOM existing, _even if a CodeMirror instance is not created_
- It doesn't currently test much of anything

I'll bring it back after this in a new PR


jdfreder:
_gasp_ Travis is failing!!   ;)

rgbkrk:
Looking forward to Travis working again

PR #359, opened 2016-04-21
Disco Mode

kenwheeler:
I'm so sorry:

disco-mode

### DIFF
--- a/static/styles/theme-undefined.css
+++ b/static/styles/theme-undefined.css
@@ -1,0 +1,52 @@
+:root {
+  --main-bg-color: white;
+  --main-fg-color: rgb(51,51,51);
+
+  --primary-border: #cbcbcb;
+
+  --cell-bg: white;
+  --cell-bg-hover: #eeedee;
+  --cell-bg-focus: #e2dfe3;
+
+  --toolbar-bg: white;
+  --toolbar-button: #aaa;
+  --toolbar-button-hover: #555;
+
+  --pager-bg: #fafafa;
+  --input-color: #8c8a8e;
+
+  --cm-background: #fafafa;
+  --cm-color: black;
+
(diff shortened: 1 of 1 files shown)


rgbkrk:
Beautiful. 10/10 best use of undefined.

captainsafia:
I stopped doing Important Things to merge this because I agree this is high priority!

captainsafia:
If anyone else merges this I will be a very angry cat.

rgbkrk:
At least we'll very clearly know what code path our users went down to get to the hidden theme.

Carreau:
Hide it behind a konami code and I'll merge it.

captainsafia:
Checks cannot finish quickly enough!

rgbkrk:
It's currently hidden behind the theme variable being unset. :wink:

Carreau:
> It's currently hidden behind the theme variable being unset.

Sure, I still think it should be hidden behind a konami code :-P

rgbkrk:
We should definitely enable a konami code here.

rgbkrk:
RxJS 4 (with RxDOM) has [an example for konami codes](https://github.com/Reactive-Extensions/RxJS/blob/master/examples/konamicode/konamicode.js).

jdfreder:
> Checks cannot finish quickly enough!

Yeah we need to cache the ZMQ stuff

captainsafia:
@jdfreder, how could you take this life-changing merge opportunity from me?!!!?

My life is never going to be satisfying and none of my dreams will ever be fulfilled.

jdfreder:
:trollface:

jdfreder:
> If anyone else merges this I will be a very angry cat.

Actually I didn't see this, sorry!

Issue #180, opened 2016-02-22
User configuration


jdfreder:
For example, user keymap.json files.

@rgbkrk suggests an 'nteract' subfolder of the jupyter user config dirs.

@kenwheeler any opinions here?  Jupyter kind of does it own thing regarding user config folder location.  In the past this has lead to confusion across various operating systems.   Maybe we should use a standard like xdg or something similar.

rgbkrk:
We do still need to implement the XDG specification in `jupyter-paths` as well: https://github.com/nteract/jupyter-paths/issues/8, to, at the very least, find all the kernels available in Linux.

I lean towards the config folder of Jupyter (with our own namespace) just so we don't have to establish yet another standard. I do recognize that `~/.nteract` is far more convenient in docs. At least in Atom, they make it discoverable within both the command palette and the top level Atom menu:

![screenshot 2016-02-21 20 17 06](https://cloud.githubusercontent.com/assets/836375/13207773/1657d1dc-d8d8-11e5-9338-91684b8f79cc.png)

Once it's discoverable that way, it's less relevant where we store them. We can then rely on the Jupyter config directory.

jdfreder:
Yeah, I think you're right about adding a menu shortcut.  That's a neat idea and really frees the user up from having to know the details.

kenwheeler:
In my latest PR, https://github.com/nteract/nteract/pull/306/files, I'm using electron-json-storage https://github.com/jviotti/electron-json-storage to write user prefs to `app.getPath('userData)`.

Reasons for this:
- The user data path varies by OS
- localStorage isn't available in the main process

captainsafia:
This can be closed since it was implemented in #922.

PR #360, opened 2016-04-21
Check for empty object response from storage

rgbkrk:
With [`electron-json-storage`](https://github.com/jviotti/electron-json-storage), if the key doesn't exist in the user data, an empty object is returned.

If you ever want to clear your storage on OS X, it's located at:
 `~/Library/Application\ Support/nteract/`.

Follow up to #306.

### DIFF
--- a/src/notebook/index.js
+++ b/src/notebook/index.js
@@ -71,7 +71,7 @@ ipc.on('main:load', (e, launchData) => {
       store.subscribe(state => this.setState(state));
       storage.get('theme', (error, data) => {
         if (error) throw error;
-        if (data.length > 0) return;
+        if (Object.keys(data).length === 0) return;
         this.setState({
           theme: data.theme,
         });


jdfreder:
:zap: disco mode

Carreau:
> `~/Application\ Support/nteract/.`

Gneh ? `~/Library/...`

rgbkrk:
```
# Enable disco mode
echo '{"theme":"undefined"}' > ~/Library/Application\ Support/nteract/theme.json
```

rgbkrk:
@Carreau Is electron-json-storage putting settings in the wrong spot? We can move things to the right spot. I was thinking in #180 that we would colocate with other jupyter config.

Carreau:
> @Carreau Is electron-json-storage putting settings in the wrong spot? We can move things to the right spot. I was thinking in #180 that we would colocate with other jupyter config.

Nah, you forgot `Library` in your first comment., I fixed it.

rgbkrk:
Ah, cool. Thanks.

PR #337, opened 2016-04-20
[WIP] Add more tests

captainsafia:
Not ready to merge yet.

PR #362, opened 2016-04-21
More tests

jdfreder:


jdfreder:
Pulled @captainsafia 's work in from #337, thanks @captainsafia !

PR #365, opened 2016-04-21
Cache zmq

jdfreder:
In theory this should really speed up the tests....

### DIFF
--- a/travis_before_install
+++ b/travis_before_install
@@ -8,20 +8,30 @@ export LD_RUN_PATH=$LDHACK/lib
 export LD_LIBRARY_PATH=$LDHACK/lib
 export PKG_CONFIG_PATH=$LDHACK/lib/pkgconfig
 echo $PKG_CONFIG_PATH
-wget http://download.zeromq.org/zeromq-$ZMQ.tar.gz
-tar xzvf zeromq-$ZMQ.tar.gz
-'[ -z "$SODIUM" ] || wget https://download.libsodium.org/libsodium/releases/libsodium-$SODIUM.tar.gz'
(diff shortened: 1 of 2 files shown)


jdfreder:
Also caches node_modules...  Which we may not want to do?

jdfreder:
Actually caching node_modules should be fine as long as we don't cache package.json...

rgbkrk:
I've seen caching `node_modules` be a problem at times, though I don't have anything concrete beyond that.

travis_before_install
```diff
@@ -8,20 +8,28 @@ export LD_RUN_PATH=$LDHACK/lib
...
-'[ -z "$SODIUM" ] || tar xzvf libsodium-$SODIUM.tar.gz'
-'[ -z "$SODIUM" ] || cd libsodium-$SODIUM'
-'[ -z "$SODIUM" ] || ./autogen.sh'
-'[ -z "$SODIUM" ] || ./configure --prefix=$LDHACK'
-'[ -z "$SODIUM" ] || make'
-'[ -z "$SODIUM" ] || make install'
-'[ -z "$SODIUM" ] || cd ..'
-'[ -z "$SODIUM" ] || export LIBS=-lsodium && export sodium_CFLAGS=$CFLAGS && export sodium_LIBS=$LDFLAGS'
+
+if [ ! -d "zeromq-$ZMQ" ]; then
```
rgbkrk, line 12: 👍

jdfreder:
4 min 46 sec -> 1 min 46 sec

That's 37% of the uncached time!

rgbkrk:
:boom:

PR #366, opened 2016-04-22
Open URLs within the user's default browser

### LINKED ISSUE #125: Clicking external links should open them in a separate browser window
If you click a link inside of nteract, it opens it in the current Electron window rather than your browser of choice.

rgbkrk:
Closes #125. All links get opened externally within the browser. Before, if you load a URL within nteract, that page would get access to node. This has parallels to the origin story of how I met @kenwheeler.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": ":notebook: the nteract notebook",
   "main": "index.js",
   "scripts": {
+    "prestart": "npm run build",
     "start": "electron .",
     "prepublish": "npm run build",
     "clean": "rimraf build dist",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "description": ":notebook: the nteract notebook",
   "main": "index.js",
   "scripts": {
+    "prestart": "npm run build",
```
rgbkrk, line 7: I snuck this in here to force a build on run. This doesn't impact the packaged app though.

rgbkrk:
Looks like the cached builds are running into issues @jdfreder.

jdfreder:
Test failing because of cache issues, merging

PR #368, opened 2016-04-22
Rebuild cache if make fails

jdfreder:


travis_before_install
```diff
@@ -9,8 +9,8 @@ export LD_LIBRARY_PATH=$LDHACK/lib
 export PKG_CONFIG_PATH=$LDHACK/lib/pkgconfig
 echo $PKG_CONFIG_PATH
 
-if [ ! -d "zeromq-$ZMQ" ]; then
-  echo "ZMQ not found in cache, downloading and building"
+function installZMQ {
```
rgbkrk, line 12: In the future I would use `installZMQ()` instead of `function`

rgbkrk:
For further good bash practices, check out the [Google Bash Style Guide](https://google.github.io/styleguide/shell.xml).

Issue #361, opened 2016-04-21
CodeMirror and Themes placing cursor oddly


rgbkrk:
Sometimes I end up with this 

![screenshot 2016-04-21 10 35 03](https://cloud.githubusercontent.com/assets/836375/14714956/a9ab23a0-07ac-11e6-8d1a-252fe55061b4.png)

and sometimes it's just fine, normal codemirror editing experience.

/cc @kenwheeler

rgbkrk:
Since switching over to [react-codemirror](https://github.com/JedWatson/react-codemirror) I haven't seen this issue again. I'll go ahead and close it. If anyone else notices it, feel free to re-open.

rgbkrk:
This issue. Guh. @jdfreder told me to make sure to take a screenshot with the codemirror dom nodes showing in the inspector next time. Will do.

jdfreder:
Spent some time tracking this down today.  Turns out the theme is getting loaded after codemirror does it's dom calculations.  The fix is to have all of the codemirror instances "refresh" on theme change.

PR #369, opened 2016-04-22
Some light CSS linting

rgbkrk:
Trying to investigate #361, getting nowhere. Oh well, have some semicolons.


jdfreder:
Looks like the testing cache is working again!

rgbkrk:
On a side note, we should switch to `obama` for linting JS.

PR #370, opened 2016-04-22
Shrink completion logic

jdfreder:


src/notebook/components/cell/editor.js
```diff
@@ -72,10 +70,26 @@ export default class Editor extends React.Component {
...
+      },
+    }));
+  }
+
   render() {
-    const extraKeys = {"Ctrl-Space": "autocomplete"}
-    let cmp = (...args) => {return completions(this._complete, ...args);}
-    cmp.async = true;
+    const extraKeys = { 'Ctrl-Space': 'autocomplete' };
+    const hint = this.completions.bind(this);
```
rgbkrk: This binding should be moved up out of `render` and into the constructor.

jdfreder: :+1:

src/notebook/components/cell/editor.js
```diff
@@ -72,10 +70,26 @@ export default class Editor extends React.Component {
...
+    hint.async = true;
```
rgbkrk: I suppose this too.

jdfreder: :heavy_check_mark:

jdfreder:
Ironically this PR net adds one line of code, but it's because of the dictionary expansion

Carreau:
The separate file was on purpose to add extra logic later for multiple completer, but fair enough.

PR #371, opened 2016-04-22
Notifications

kenwheeler:
This is the initial implementation of the notification system.

@rgbkrk Do you want to pull this right in and do styling & more notifications (like gist error) as a separate PR, or get them into this one before merging?
screen shot 2016-04-22 at 9 57 25 am


rgbkrk:
We can put it in now and do styling after.

src/notebook/index.js
```diff
@@ -82,6 +85,7 @@ ipc.on('main:load', (e, launchData) => {
       });
     }
     componentDidMount() {
+      dispatch(setNotificationSystem(this.refs.notificationSystem));
```
rgbkrk, line 88: Can you elaborate on why this ref has to passed around?

rgbkrk:
Not sure why this fails on project coverage with codecov.

src/notebook/index.js
kenwheeler, line 88: So the notification system is a component, and its not available until mount. Therefore we have store the reference to the mounted component in the store, so that we can operate on it from other parts of the application.

PR #373, opened 2016-04-22
Rely on unload for shutdown of kernels.

rgbkrk:
Fixes #245.

We'll want to move out the kernel cleanup function to somewhere within `notebook/api` (which still seems like a misnomer) or a separate library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ cache:
 
 before_install:
   - source travis_before_install
+  - pip install jupyter_console --user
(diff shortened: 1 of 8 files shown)


rgbkrk:
Nice. I've been stopped from patching for not testing. :smile: I'll be on this again later tonight.

rgbkrk:
This is all set now, complete with tests! Ok, slightly tested.

jdfreder:
Wahoo!!!

PR #377, opened 2016-04-23
Keep the kernel fields together.

rgbkrk:
This puts the kernel fields together so the call is symmetric (and so I don't use the wrong args).


rgbkrk:
We definitely need a test kernel... I'm in a bit of a predicament because Travis has a Python2 kernel whereas locally others might have a Python3 kernel.

PR #382, opened 2016-04-23
Always on completions

rgbkrk:
This switches us over to hints all the time, without the users hitting a hot key (though they can still invoke it manually with ctrl-space).

In order to do completions all the time, without tab, we have to set completeSingle to false. Otherwise users will "overtype" the completions with their own text.

For example, let's say a user is typing "import" in Python:

Type `i` (list of completions)

Type `m` ("import" matches, gets filled out)

Type p and now the text reads:

```
importp
```

Switching to this mode makes us behave more like Visual Studio and Atom. The user
must confirm the match in order to accept it. For previous Jupyter users, this
may cause a rift because muscle memory tells them to use tab. This _is_ a different environment
so maybe that's alright.

My personal preference is to have this mode on by default. Thoughts?

Please try it out. It's kind of awesome.

### DIFF
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -63,7 +63,16 @@ export default class Editor extends React.Component {
     //       is deleted
     inputEvents
       .debounceTime(20)
-      .subscribe(event => event.cm.execCommand('autocomplete'));
+      // Filter out whitespace changes, only propagate when a partial token
+      .filter(event => {
+        const editor = event.cm;
+        const tokenRange = editor.findWordAt(editor.getCursor());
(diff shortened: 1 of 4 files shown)


rgbkrk:
We'll want to filter out pure whitespace lines or some other heuristic as well because of IPython's global mode responses.

rgbkrk:
Did some basic whitespace checking. We still need to fix the "open-parens" case. I'm digging into this now.

rgbkrk:
I probably should have put a "please discuss" tag on this one. This needs some more work.

Carreau:
That's likely a codemirror bug. As this behavior should be triggered only when it's the "first" invocation of simpleHint so there is likely something wrong in my implementation.

PR #384, opened 2016-04-23
Autocomplete only when current token has \w

rgbkrk:
This is quite a bit better than the previous token check.

wordy


jdfreder:
Much better!

jdfreder:
🍰 🎆

PR #391, opened 2016-04-25
Prevent the completer from flickering.

Carreau:
Actually trigger an "autocomplete" IIF the completion is not already
shown. Otherwise let codemirror deal with the **update** of the popup,
instead of removing the dom element and reinserting.

Prevent "flicker" of the tooltip.


rgbkrk:
Oh awesome! Didn't know about this.

We could even move this check up into a filter operator prior.

rgbkrk:
This is sooooooo smooth.

PR #392, opened 2016-04-25
Streamline icon generation.

Carreau:
And adjust colors a bit, to be slightly more dynamic. 

Also now it build the PNG dirrectly using inkscape. 

Might want to adjust the SVG for the small icon as well, but heading to lunch for now.

screen shot 2016-04-25 at 13 05 59


rgbkrk:
I'm guessing we'll only be running this from a Linux machine to build the icons.

Carreau:
Or osx. Works for me.

PR #393, opened 2016-04-25
Use a drag handle instead of the whole cell,

### LINKED ISSUE #136: Need to be able to select text in outputs
The current drag and drop is awesome!

I'd love to be able to select text and images in outputs and rendered markdown text, which currently conflicts with our drag handle. You can select code no problem, just nothing that's rendered.

jdfreder:
Allows users to select output
closes #136

### DIFF
--- a/src/notebook/components/cell/draggable-cell.js
+++ b/src/notebook/components/cell/draggable-cell.js
@@ -102,7 +102,7 @@ class DraggableCell extends React.Component {
   }
 
   render() {
-    return this.props.connectDropTarget(this.props.connectDragSource(
+    return this.props.connectDropTarget(
       <div
         style={{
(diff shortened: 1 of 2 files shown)


rgbkrk:
Now that's a sneaky handle

rgbkrk:
Since we're variablizing our CSS anyways, we should probably make this match the input area width directly.

rgbkrk:
Looking great! This works quite well. It feels (as a user) that the selection region should start a bit further in on the left of the output area.

jdfreder:
> It feels (as a user) that the selection region should start a bit further in on the left of the output area.

Do you mean the array of selectable text?  IOW the drag handle width should be thinner?

rgbkrk:
Yeah, thinner drag handle width.

PR #395, opened 2016-04-25
Use move cursor when hovering over drag region

jdfreder:


betatim:
actually, the merge button is green when coverage isn't done yet. :-/ slower merging in the future.

jdfreder:
CodeCov seems to have a slow response time sometimes :/

rgbkrk:
All good!

static/styles/main.css
```diff
@@ -57,6 +57,7 @@ td {
     z-index: 999;
     width: 50px;
     height: 100%;
+    cursor: move;
```
rgbkrk, line 60: I never knew about this and CSS. 😧

jdfreder, line 60: Oh, yeah, there are a bunch of different cursors you can choose from.  It's awesome. :heart: css

rgbkrk:
ok wow. This feels ​_real_​ good. What a great tweak. As a user it feels natural to select within the output area now, regardless of when the `I` appears.

However, when a markdown cell is in edit mode it shouldn't be draggable.

screenshot 2016-04-25 13 05 59

Otherwise you can't select your beginning text very easily.

jdfreder:
The editor should probably be padded to match the left start of the code editor, what do you think?

jdfreder:
#396

rgbkrk:
The only reason I kept the markdown cell as full width before was to match the old notebook. Open to trying this!

PR #398, opened 2016-04-26
Set the running status.

rgbkrk:
Now with more `[*]`. I left this in a setup that made it simpler for us to implement an alternative animation (e.g. spinning brackets).

As I did this one I realized that we've got even more non-notebook format at-runtime cell data (not to be confused with cell metadata):
- running status
- cell pagers
- subscriptions (outputs, etc. that are streaming in from the kernel) - for closing when new code is run

After this we should probably coalesce it into one structure, possibly named cellRuntimeData (I'm unsure of names here).


rgbkrk:
What an angry looking chart.

jdfreder:
> What an angry looking chart.

I know, right!?  I haven't seen the chart until now.  New feature of codecov?

rgbkrk:
Must be. They let you set what you enable and disable, I think I'll turn whatever that is off.

PR #399, opened 2016-04-26
Create a codecov YAML for settings

rgbkrk:
While I was at it, turned off the codecov sunburst

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+comment:
+  layout: header, changes, diff
+coverage:
+  status:
+    patch:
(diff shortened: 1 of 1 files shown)


minrk:
Depending how many repos you want to set this in, there is a [team config](https://codecov.io/account/gh/nteract/yaml) you can edit to set defaults at the org level.

I learned this yesterday when trying to disable the comments that they suddenly enabled on all repos.

rgbkrk:
Awesome, thank you. I just added this same YAML to the whole org.

PR #400, opened 2016-04-26
Rely on nteract/assets for font installation.

### LINKED ISSUE #308: unpin fonts from git release.
The following are making my npm install take **literally** more than 30 minutes on not-that great connection as it does clone the full repo:

```
  "source-code-pro": "git://github.com/adobe-fonts/source-code-pro.git#release",
   "source-sans-pro": "git://github.com/adobe-fonts/source-sans-pro.git#release",
```

Can we avoid that ?

jdfreder:
closes #308

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,8 +68,6 @@
     "react-notification-system": "^0.2.7",
     "remark": "^4.2.0",
     "remark-react": "^2.0.0",
-    "source-code-pro": "git://github.com/adobe-fonts/source-code-pro.git#release",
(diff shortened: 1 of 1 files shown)


rgbkrk:
You'll need to modify index.html too

jdfreder:
No, it's already pointing to the correct location because npm installs dependencies flattly instead of in a hierarchy.

rgbkrk:
Ah, nteract/assets needs to vendorize that then. #308 is explicitly calling for _not_ installing source code pro from git, which is currently being done.

PR #404, opened 2016-04-27
Add ability to run all cells

captainsafia:
Just kidding. I'll try to get this to work but will keep this PR open for question asking.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -127,9 +127,7 @@ export const cell = {
     {
       label: 'Run All',
       accelerator: 'Cmd+E',
-      click: (item, focusedWindow) => {
-        console.log(focusedWindow);
-      },
+      click: createSender('menu:run-all')
     },
   ],
 };
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -121,6 +121,17 @@ export const edit = {
   ],
 };
 
+export const cell = {
+  label: 'Cell',
+  submenu: [
+    {
+      label: 'Run All',
+      accelerator: 'Cmd+E',
+      role: 'run-all'
+    },
+  ],
+};
+
 export const view = {
   label: 'View',
   submenu: [
@@ -350,6 +361,7 @@ export function loadFullMenu() {
 
     template.push(fileWithNew);
     template.push(edit);
+    template.push(cell);
     template.push(view);
 
     // Application specific functionality should go before window and help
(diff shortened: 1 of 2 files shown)


src/main/menu.js
```diff
@@ -121,6 +121,17 @@ export const edit = {
   ],
 };
 
+export const cell = {
+  label: 'Cell',
+  submenu: [
+    {
+      label: 'Run All',
+      accelerator: 'Cmd+E',
+      role: 'run-all'
```
rgbkrk: `role` is a construct for native menu actions, you'll want to use `click` instead, sending a menu action onward to the renderer process.

```
click: createSender('menu:run-all')
```

As an example, check out the language menu below this section.

captainsafia:
Ha! You beat me to it, @rgbkrk! I'm a bad little programmer who doesn't read the docs!

I'll take a look at how things are done in the language menu.

I will cherry-pick my commits when this is done.

rgbkrk:
That's alright, I didn't realize either when I first started with the native menu stuff.

captainsafia:
OK. This "works" but might be 💩. Also, I'm thinking that we could test this (and other menu handlers) with spectron (per #411) eventually.

src/notebook/menu/index.js
```diff
@@ -86,12 +87,20 @@ export function dispatchPublishGist(store, dispatch) {
   });
 }
 
+export function dispatchRunAll(store, dispatch) {
+  const { notebook, channels} = store.getState();
+  notebook.get('cellMap').map((value, key) => {
```
rgbkrk, line 92: You're going to want to rely on `cellOrder` in tandem with `cellMap`, because you can't guarantee order of the `cellMap`.

src/main/menu.js
```diff
@@ -121,6 +121,17 @@ export const edit = {
   ],
 };
 
+export const cell = {
+  label: 'Cell',
+  submenu: [
+    {
+      label: 'Run All',
+      accelerator: 'Cmd+E',
```
rgbkrk, line 129: What led you to this binding?

rgbkrk:
We'll go ahead and iterate on fixing up the cell run ordering after.

:+1:

src/main/menu.js
captainsafia, line 129: I was originally thinking "Cmd+R" but that was already bound to Reload so I picked "Cmd+E" for execute.

I guess maybe I should also change the menu label to "Execute All"? Although that would differ from the language users are generally familiar with.

rgbkrk, line 129: I like "run all" too.

We need to put keymap on the roadmap, if even just to pin down our design of the menu. I'm hesitant to add too many bindings for fear of entrenching a binding we have to support forever.

PR #407, opened 2016-04-27
Wrap code (text/plain) so it fits within a cell

rgbkrk:
Before:

screenshot 2016-04-27 01 37 17

After:

screenshot 2016-04-27 01 37 42

/cc @betatim who asked about this


betatim:
Yay!

PR #414, opened 2016-04-29
Don't Merge: Proposed dep injected subject tests.

kenwheeler:
This is a proposed approach for testing our async flows within actions/agendas that uses `babel-plugin-rewire` to inject mocks into the tested module on the fly.

cc @rgbkrk

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,6 +5,7 @@
     ],
     "plugins": [
       "transform-class-properties",
-      "transform-object-rest-spread"
+      "transform-object-rest-spread",
+      "rewire"
     ]
   }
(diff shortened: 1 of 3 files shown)


rgbkrk:
Whoa, that's pretty neat. I've been wondering about the right way to do mocks with these.

rgbkrk:
Should we go ahead and close this now that we're on redux with redux-observable?

kenwheeler:
I should probably just refactor these right?

rgbkrk:
Sounds good to me!

rgbkrk:
Closing due to inactivity.

PR #415, opened 2016-04-29
Compress path in title bar by un-expanding tilde

Carreau:
we **might** want to actually create the untildify logic ourself that is probably relatively simple, I just don't  node stdlib enough.

### DIFF
--- a/src/notebook/native-window/index.js
+++ b/src/notebook/native-window/index.js
@@ -1,11 +1,27 @@
 import { getCurrentWindow } from 'remote';
+import home from 'home-dir';
+import path from 'path'
+
+const HOME = home();
+
+function _tildify(home, p) {
+  const s = path.normalize(p) + path.sep;
+  return (s.indexOf(home) === 0 ? s.replace(home + path.sep, '~' + path.sep) : s).slice(0, -1);
+}
+
+function tildify(path) {
+  if(!path){
+    return ''
+  }
+  return _tildify(HOME, path)
+}
 
 export function initNativeHandlers(store) {
   store
     .map(state => {
       const { executionState, filename } = state;
       return {
-        title: `${filename || 'Untitled'} - ${executionState}`,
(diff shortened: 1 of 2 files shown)


rgbkrk:
This looks like a good idea to me. I'm not sure how/why this failed the functional tests, so I re-kicked it.

jdfreder:
Looks like the tildify package may be having trouble on the version of electron that we are testing against ??

jdfreder:
Test fails locally too

jdfreder:
Test pass locally on master though...  So yeah, must be the tildify package..

rgbkrk:
We do have our own home directory function (that is used within jupyter-paths). Coupling that with this function

```
function tildify(home, p) {
  const s = path.normalize(p) + path.sep;
  return (s.indexOf(home) === 0 ? s.replace(home + path.sep, '~' + path.sep) : s).slice(0, -1);
}
```

and we're well on our way.

rgbkrk:
Weird, doesn't pass locally for me either. I wonder if it's the synchronous home directory call that `tildify` makes.

Carreau:
I'll update not to use tildify.

Carreau:
Updated.

rgbkrk:
Awesome! I'm going to merge and then make some notes inline.

src/notebook/native-window/index.js
```diff
@@ -1,11 +1,27 @@
...
+  const s = path.normalize(p) + path.sep;
+  return (s.indexOf(home) === 0 ? s.replace(home + path.sep, '~' + path.sep) : s).slice(0, -1);
+}
+
+function tildify(path) {
+  if(!path){
+    return ''
+  }
+  return _tildify(HOME, path)
+}
```
rgbkrk, line 17: Since the function is not exported / not given as a public API, we don't need to use a leading underscore.

PR #418, opened 2016-05-02
Fix run all

captainsafia:
Follow up on #404.


rgbkrk:
Legit, this is good. 👍

PR #421, opened 2016-05-02
"Refresh" codemirror on theme change

### LINKED ISSUE #361: CodeMirror and Themes placing cursor oddly

jdfreder:
closes #361

### DIFF
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -22,6 +22,7 @@ export default class Editor extends React.Component {
     lineWrapping: React.PropTypes.bool,
     onChange: React.PropTypes.func,
     theme: React.PropTypes.string,
+    cmtheme: React.PropTypes.string,
     focused: React.PropTypes.bool,
   };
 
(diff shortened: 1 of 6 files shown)


src/notebook/components/cell/editor.js
```diff
@@ -100,6 +105,11 @@ export default class Editor extends React.Component {
       const cm = this.refs.codemirror.getCodeMirror();
       cm.getInputField().blur();
     }
+
+    if (this.theme !== this.props.theme) {
+      this.theme = this.props.theme;
+      this.refs.codemirror.getCodeMirror().refresh();
```
rgbkrk, line 111: You want this in the props change setup too, likely in `componentDidUpdate`.

jdfreder, line 111: I did, and I tested the life cycle events by hand, this gets fired at all times.  The shouldUpdate life cycle event returns true by default.  Since this is passed as a prop, it triggers the didupdate method.  http://ricostacruz.com/cheatsheets/react.html#updating

rgbkrk, line 111: If `componentDidMount` is firing all the time, we have another problem. From what you linked, you'll want to add this to `componentDidUpdate` as well.

PR #423, opened 2016-05-05
Bump dependencies

rgbkrk:
Primarily for ANSI to React, went ahead
and adopted other newness in our stack (commutable for
migrating v3 -> v4 notebooks e.g.).

PR #424, opened 2016-05-05
Bump rjda

rgbkrk:
Missed one in #423


jdfreder:
Lot's of small version bump PRs tonight!

rgbkrk:
This is where I must admit that I will always make packages start at `1.0.0` and then bump the patch release earnestly. It was such a trickle to get ansi-to-react bits moved up.

PR #426, opened 2016-05-05
Corrent indenture level to match Jupyter/notebook

jdfreder:


rgbkrk:
👍

rgbkrk:
Since we do the same thing with gists, perhaps we need a `stringify` for our commutable notebooks.

jdfreder:
Yeah, agreed, it really is something that belongs at that level because it affects disk representation. https://github.com/nteract/commutable/issues/30

PR #432, opened 2016-05-07
Added Zoom In and Zoom Out function

captainsafia:
zoom-in-zoom-out

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -173,12 +173,12 @@ export const view = {
     },
     {
       label: 'Zoom In',
-      accelerator: 'CmdorCtrl++',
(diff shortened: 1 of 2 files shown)


rgbkrk:
Great! I've been hoping for this.

rgbkrk:
This does make me wonder if there's a zoom role or set of defaults so we don't have to implement this all the way through.

captainsafia:
> This does make me wonder if there's a zoom role or set of defaults so we don't have to implement this all the way through.

I'll have to dig up the links for them again but I found some in-progress issues on electron to add this. For now, the solution seems to be using `webFrame`.

PR #433, opened 2016-05-08
Have CodeMirror emit signal when hitting cell boundary

Carreau:
It might be useful to listen too these signals to select previous/next
cell.

### DIFF
--- a/src/notebook/components/cell/editor.js
+++ b/src/notebook/components/cell/editor.js
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import CodeMirror from 'react-codemirror';
+import CM from 'codemirror'
 
 const Rx = require('@reactivex/rxjs');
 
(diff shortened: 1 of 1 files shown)


jdfreder:
> It might be useful to listen too these signals to select previous/next
> cell.

👍

rgbkrk:
that starts us off!

src/notebook/components/cell/editor.js
```diff
@@ -12,6 +13,34 @@ import 'codemirror/addon/hint/anyword-hint';
 // Hint picker
 const pick = (cm, handle) => handle.pick();
 
+const goLineUpOrEmit = (editor) =>{
```
jdfreder, line 16: When you're not in the context of an object, and don't need to use arrow functions, it makes more sense to use a named function because it's easier to debug in a stack trace.

Carreau, line 16: Thanks, I will next time !

Issue #422, opened 2016-05-02
Cell execution is slow


jdfreder:
Both Ctrl-Enter and Shift-Enter cell execution seems to be really slow.  It's not clear to me at all why this is happening, but it wasn't happening in earlier commits and is even slower than jupyter/notebook.

rgbkrk:
Just "figured out" one of our performance problems. React is always running in dev mode right now because we don't set `process.env.NODE_ENV` to `"production"`.

captainsafia:
💩 ! This makes all our builds fail!

jaredly:
execution doesn't seem slow at all to me :/ @jdfreder are you still experiencing this? And was it on a large, complex notebook, or did it also happen w/ a fresh notebook just evaling e.g. `2+2`?

jdfreder:
Hey @jaredly , thanks for looking into this.  Actually, I think I may have "fixed" (maybe just covered up) the actual problem when I sped up nteract using the PureRenderMixin (in master).

rgbkrk:
This seems to not be an issue right now, feel free to re-open.

Issue #128, opened 2016-02-12
Create mocks for UI of kernel indicators


rgbkrk:
We need to post mocks and UI flows for the state of the kernel.

States:
- [x] Kernel busy
- [x] Current language
- [ ] Icon for language

jdfreder:
#434 makes the kernel busy state available

captainsafia:
I can work on making actual pictures, but I like the idea of having a simple indicator at the bottom of the window in the status bar.

https://github.com/atom/status-bar

rgbkrk:
I like the Atom status bar too.

captainsafia:
Yeah, now I just gotta see if they have an electron API for status bars.

rgbkrk:
If you want fancy graphics (kernel logos), I'm guessing you're better off with something DOM/React based.

captainsafia:
That makes sense. Might makes sense to also use React for the status bar since we used it for context menus...

rgbkrk:
Gutter made this awesome. Closing.

PR #434, opened 2016-05-10
Testing, performance tests, and assorted bug fixes

jdfreder:
Sorry this is an all-in-one, but it was convenient to open the PR as-is.

This PR:
- fixes createStore initialState not setting the state if no actions were trigger
- fixes a bug in the fallback pwd, cwd call.  Missing parens
- listens to kernel status messages and updates the execution state accordingly (cc @betatim I this is kind of what you wanted)
- adds a performance API which is a noop if window.performance does not exist
- marks actions and agendas
- measures the execute_cell agenda
- adds basic performance test to travis
- adds an execute_cell agenda test
- adds convenience methods for testing against a live (kernel backed) state store

related to #422 , #128


src/notebook/actions/index.js
```diff
@@ -28,9 +28,16 @@ export function newKernel(kernelSpecName, cwd) {
...
+        // Listen to the execution status of the kernel
+        channels.iopub
+          .filter(msg => msg.header.msg_type === 'status')
+          .map(msg => msg.content.execution_state)
+          .subscribe(state => subject.next(setExecutionState('idle')));
+
         agendas.acquireKernelInfo(channels)
               .subscribe(action => {
                 subject.next(action);
+                subject.next(setExecutionState('idle'));
```
rgbkrk, line 40: :+1:

src/notebook/agendas/index.js
```diff
@@ -141,11 +146,14 @@ export function executeCell(channels, id, code) {
...
          // Update the outputs with each change
          .subscribe(outputs => {
+           mark('executeCell:output');
+           measure('executeCell:roundtrip', 'executeCell:start','executeCell:output');
            subscriber.next(updateCellOutputs(id, outputs));
          })
     );
 
     shell.next(executeRequest);
+    mark('executeCell:end');
```
rgbkrk, line 156: End is a bit of a misnomer since this is more like "sent", the finish will be later.

rgbkrk:
My hope was that we would send the current time stamp over with the actions, as would be useful for realtime, instead of using marks. I'm going to try this current implementation out soon though.

jdfreder:
We can do that too, and remove the mark calls from everything but the agendas.  In the agendas we will keep the marks and measure call.

PR #435, opened 2016-05-11
Add support for up/down arrow keys to navigate between cells

jdfreder:


rgbkrk:
Well that's awesome. 😄

jdfreder:
Untested :  behavior with rendered MD
Know awkwardness: CM cursor position is set to the top/bottom of the cell correctly and instead assumes last position

rgbkrk:
Ah yeah, rendered markdown focuses the cell and then kind of "gets stuck".

Code cell up/down is awesome though.

PR #437, opened 2016-05-11
Make the marks/measures more useful and less intrusive by

jdfreder:
moving them into the store itself.  Measures are now
automatically made for every reduction.


rgbkrk:
Much better. Way less boilerplate.

PR #438, opened 2016-05-11
Add up/down prev/next cell support to MD cells

jdfreder:


rgbkrk:
I'm glad you made enter work on the markdown cells for getting into edit mode! It does introduce a newline though.

PR #440, opened 2016-05-11
Prevent pesky newlines from showing on enter MD cell

jdfreder:


src/notebook/components/cell/markdown-cell.js
```diff
@@ -78,9 +78,11 @@ export default class MarkdownCell extends React.Component {
         break;
```
rgbkrk, line 84: This function, out of all of them, definitely needs a comment at the top. In context, I know what this is _exactly_ for right now. Later, I'm not sure.

jdfreder, line 84: Yeah, I can add a comment for now.  In the future this entire thing should disappear for a keymap replacement IMHO

jdfreder, line 84: :heavy_check_mark: comments added

PR #441, opened 2016-05-11
WIP: Enable OSX on Travis

jdfreder:


jdfreder:
The Travis OSX builds are too slow, Linux will have to suffice...

PR #450, opened 2016-05-15
Use octicons, remove ADD CELL text

jdfreder:


jdfreder:
I think we need to adjust the offset of the toolbars to make things look nicer with the new icon set.

rgbkrk:
Did you want to adjust the offset now or after?

jdfreder:
Later, I'm not sure how it should be handled yet

PR #454, opened 2016-05-16
Pre-emptively handle deprecated APIs for Electron

rgbkrk:
In advance of switching to Electron 1.1.0, extracted from #453.


jdfreder:
Woohoo!  Thanks Kyle!

PR #463, opened 2016-05-17
Add notebook classic theme,

jdfreder:
also don't outline rendered markdown on focus
and fix output prompt padding

classic-theme

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -192,6 +192,10 @@ export const view = {
           label: 'Dark',
           click: createSender('menu:theme', 'dark'),
         },
+        {
+          label: 'Classic',
+          click: createSender('menu:theme', 'classic'),
+        },
       ],
     },
   ],
--- a/src/notebook/index.js
+++ b/src/notebook/index.js
@@ -111,8 +111,8 @@ ipc.on('main:load', (e, launchData) => {
                 />
             }
             <NotificationSystem ref="notificationSystem" />
-            <link rel="stylesheet" href={`../static/styles/theme-${this.state.theme}.css`} />
             <link rel="stylesheet" href="../static/styles/main.css" />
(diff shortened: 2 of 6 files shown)


rgbkrk:
omg, so classic

src/notebook/index.js
```diff
@@ -111,8 +111,8 @@ ipc.on('main:load', (e, launchData) => {
                 />
             }
             <NotificationSystem ref="notificationSystem" />
-            <link rel="stylesheet" href={`../static/styles/theme-${this.state.theme}.css`} />
             <link rel="stylesheet" href="../static/styles/main.css" />
+            <link rel="stylesheet" href={`../static/styles/theme-${this.state.theme}.css`} />
```
rgbkrk, line 115: What's the reason for flipping these? I'm assuming that the CSS vars aren't defined at first, then Chrome resolves it?

src/main/menu.js
```diff
@@ -192,6 +192,10 @@ export const view = {
           label: 'Dark',
           click: createSender('menu:theme', 'dark'),
         },
+        {
+          label: 'Classic',
+          click: createSender('menu:theme', 'classic'),
+        },
```
rgbkrk, line 198: We're charting into that "should we generate this" territory. :wink:

src/notebook/index.js
jdfreder, line 115: User CSS should always be loaded after all base CSS been loaded.  This increases the specificity of the user's CSS selectors.  IOW it makes it so user CSS doesn't have to use `!important`.  Additionally, it allows CSS vars that are declared in the base to be overridden in the theme files, which is not possible without this change.

rgbkrk, line 115: 👍

PR #469, opened 2016-05-17
Fix title bar text on save

### LINKED ISSUE #462: Title bar does not update on save
Might be related to #330 but saving a file does not update the file name in the title bar.

nteract-save-as

captainsafia:
Fixes #462.

### DIFF
--- a/src/notebook/menu/index.js
+++ b/src/notebook/menu/index.js
@@ -31,7 +31,7 @@ export function triggerSaveAs(store, dispatch) {
       if (!filename) {
         return;
       }
-      const { notebook, executionState} = store.getState();
+      const { notebook, executionState } = store.getState();
       dispatch(saveAs(filename, notebook));
       BrowserWindow.getFocusedWindw().setTitle(`${tildify(filename)} - ${executionState}`);
     }
--- a/src/notebook/menu/index.js
+++ b/src/notebook/menu/index.js
@@ -4,6 +4,8 @@ import {
   showSaveAsDialog,
 } from '../api/save';
 
+import { tildify } from '../native-window';
+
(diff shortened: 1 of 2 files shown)


captainsafia:
TRRRAAAAAAVVVIIIISSSSSSS.

src/notebook/menu/index.js
```diff
@@ -29,8 +31,9 @@ export function triggerSaveAs(store, dispatch) {
       if (!filename) {
         return;
       }
-      const { notebook } = store.getState();
+      const { notebook, executionState } = store.getState();
       dispatch(saveAs(filename, notebook));
+      BrowserWindow.getFocusedWindw().setTitle(`${tildify(filename)} - ${executionState}`);
```
rgbkrk, line 36: This makes me wonder if we should be submitting an action for name change.

captainsafia, line 36: I think that's a good idea. I figure changing the window title is something we'll have to do quite often.

src/notebook/menu/index.js
```diff
@@ -31,7 +31,7 @@ export function triggerSaveAs(store, dispatch) {
       if (!filename) {
         return;
       }
-      const { notebook, executionState} = store.getState();
+      const { notebook, executionState } = store.getState();
```
rgbkrk, line 34: This is a peace offering to the rest of us using `eslint` and wanting a consistent style. :wink:

```
npm run lint
```

If you're in Atom, I recommend installing package `linter-eslint`

captainsafia, line 34: Yeah, I should probably add it as a pre-commit hook or something.

PR #470, opened 2016-05-17
Remove cell toolbar ghosts

### LINKED ISSUE #465: Cell toolbar menu doesn't always hide
celltoolbar

jdfreder:
I hate having to do this, but it's the workaround I've settled on because both React and native mouseleave events are not firing reliably...

closes #465

### DIFF
--- a/src/notebook/components/cell/cell-creator.js
+++ b/src/notebook/components/cell/cell-creator.js
@@ -11,6 +11,7 @@ export default class CellCreator extends React.Component {
   constructor() {
     super();
     this.setHoverElement = this.setHoverElement.bind(this);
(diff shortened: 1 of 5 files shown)


rgbkrk:
We do have access to the DOM with `electron-mocha` for testing. `setCellHoverState` seems like the most likely candidate for testing.

rgbkrk:
One day I look forward to deeper testing than this. Good framing for a later pass. :smile:

rgbkrk:
I can no longer click trash before the upper right cell toolbar disappears.

jdfreder:
Oh you know what, I was testing on the classic theme.  It may be that the mask is different than the other themes.  I can look at this tomorrow

PR #471, opened 2016-05-17
Add uploading notification

### LINKED ISSUE #458: Pop a notification while the gist is uploaded.
I just use the "publish > to gist.." I was confused as it took a few second and I had no visual feedback.

jdfreder:
closes #458

### DIFF
--- a/src/notebook/reducers/document.js
+++ b/src/notebook/reducers/document.js
@@ -207,7 +207,14 @@ export default {
       notebook: notebook.setIn(['metadata', field], Immutable.fromJS(value)),
     };
   },
-  [constants.STARTED_UPLOADING]: noop,
+  [constants.STARTED_UPLOADING]: function startedUploadingGist(state) {
+    const { notificationSystem } = state;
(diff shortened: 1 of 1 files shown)


src/notebook/reducers/document.js
```diff
@@ -207,7 +207,14 @@ export default {
...
-  [constants.STARTED_UPLOADING]: noop,
+  [constants.STARTED_UPLOADING]: function startedUploadingGist(state) {
+    const { notificationSystem } = state;
+    notificationSystem.addNotification({
+      title: 'Uploading gist...',
+      message: 'Your notebook is being uploaded as a GitHub gist',
+      level: 'info',
+    });
+  },
   [constants.DONE_UPLOADING]: noop,
```
Carreau, line 218: Should `done` remove/replace the notification ?

rgbkrk:
This is making me wonder if the notification belongs in actions and that the notification system should be in context, like the channels. Reduction should not be doing anything asynchronous.

PR #472, opened 2016-05-17
Clear execution count and display an error when trying to run a cell with no kernel

### LINKED ISSUE #457: Confused when executing and kernel not connected.
It should at least clear the input prompt number.

(maybe pop a red notification on the upper right, like the gist uploaded one.)

Something else felt weird when this happened,  but it's hard to tell what.

jdfreder:
closes #457

### DIFF
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -212,8 +212,18 @@ export function focusPreviousCell(id) {
   };
 }
 
-export function executeCell(channels, id, source) {
-  return (subject) => {
+export function executeCell(channels, id, source, hasKernel, notificationSystem) {
+  return (subject, dispatch) => {
+    if (!hasKernel) {
+      notificationSystem.addNotification({
+        title: 'Could not executed cell',
+        message: 'The cell could not be executed because the kernel is not connected.',
+        level: 'error',
+      });
+      dispatch(updateCellExecutionCount(id, undefined));
+      return;
+    }
+
     const obs = agendas.executeCell(channels, id, source);
     obs.subscribe(action => {
       subject.next(action);
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -212,9 +212,9 @@ export function focusPreviousCell(id) {
   };
 }
 
-export function executeCell(channels, id, source, hasKernel, notificationSystem) {
+export function executeCell(channels, id, source, kernelConnected, notificationSystem) {
   return (subject, dispatch) => {
-    if (!hasKernel) {
+    if (!kernelConnected) {
       notificationSystem.addNotification({
         title: 'Could not execute cell',
         message: 'The cell could not be executed because the kernel is not connected.',
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -216,7 +216,7 @@ export function executeCell(channels, id, source, hasKernel, notificationSystem)
   return (subject, dispatch) => {
     if (!hasKernel) {
       notificationSystem.addNotification({
-        title: 'Could not executed cell',
+        title: 'Could not execute cell',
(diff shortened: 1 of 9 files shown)


src/notebook/actions/index.js
```diff
@@ -213,13 +213,24 @@ export function focusPreviousCell(id) {
...
-    const obs = agendas.executeCell(channels, id, source);
-    obs.subscribe(action => {
-      subject.next(action);
-    }, (error) => {
-      subject.next({ type: constants.ERROR_KERNEL_NOT_CONNECTED, message: error });
-    });
+  return (subject, dispatch, store) => {
+    const state = store.getState();
+    const { executionState, notificationSystem } = state;
+    if (executionState === 'starting' || executionState === 'not connected') {
```
rgbkrk: Should we be checking on the state of `channels` at this point?

src/notebook/store/index.js
```diff
@@ -40,7 +40,7 @@ export default function createStore(initialState, reducers) {
 
   function dispatch(action) {
     return typeof action === 'function'
-      ? action.call(null, subject, dispatch)
+      ? action.call(null, subject, dispatch, store)
```
rgbkrk: awww, don't sneak the store back in

src/notebook/actions/index.js
jdfreder: Channels is set even if the kernel hasn't started

jdfreder: To clarify, the channels observables exist but may point to a kernel that hasn't started

src/notebook/actions/index.js
```diff
@@ -213,13 +213,24 @@ export function focusPreviousCell(id) {
...
+      notificationSystem.addNotification({
+        title: 'Could not executed cell',
```
rgbkrk: > Could not execute cell

src/notebook/store/index.js
jdfreder: I know, I really didn't want to either

src/notebook/actions/index.js
rgbkrk: This check needs to performed higher up, as it's needed by the run all menu option, the cell toolbar, the shift-enter run, etc. It can be a new prop or a context type.

src/notebook/store/index.js
rgbkrk: Primarily because we're going back to real Redux and can't make this worse before we make it better.

rgbkrk:
From a conversation with Jon (copied with permission):

```
Jon: I'm frustrated in this PR. I shouldn't be accessing the state in the actions
     BUT
     not doing so means I need to pass the execution state down through all of the components, all the way to the cell toolbar with the play button. that or pass down a callback, all the way... ug

Kyle: The `notificationSystem` should be a `context` type, just like the `channels`

Jon: ah yeah, that solves that, but what about execution state? Surely it doesn't logically make sense to pipe it all the way down.

Kyle: Now what you're saying about execution state. Let's take a step back

Jon: execution state is required because the launch of a kernel is only half of the story. so channels are only half of the story.

Kyle: The goal here is to not execute if the kernel is not connected

Jon: we need to hear from the kernel that it has been launched successfully

Kyle: should these executions get queued up ahead

Jon: no, I don't think so

Kyle: k

Jon: because, for example, my computer has a Py3 kernel which launches but never "starts"

Kyle: What if the prop we were passing (or context) was whether the kernel was connected. I realize we can't check that `channels` is available, because it might still be connecting.

Jon: Still seems like the toolbar shouldn't have to know about that

Kyle: I think this check needs to be performed higher up. More than one piece of the app can do a run. It's the core feature

Cell -> Run All
Shift-enter
Cell Toolbar

Jon: Yeah, so the information will have to be propagated to all of those

Kyle: yeah, right off the store you can select that prop off
```

PR #474, opened 2016-05-18
Send pumpkin to the back

Carreau:
It is supposed to be at the bottom of the screen though, not sure why. Electron bug ?


Carreau:
screen shot 2016-05-17 at 20 22 52

PR #476, opened 2016-05-18
Show scrollbar on overflown code cells

### LINKED ISSUE #475: Code cell handles overflow content awkwardly
Code cell should scroll of the content inside it is longer than the width of the window. Instead, you get something like this.

screen shot 2016-05-18 at 2 03 29 am

captainsafia:
Closes #475.

@kenwheeler: I assume this is a legit approach to scrolling overflowing content on flexboxes?


kenwheeler:
👍

PR #477, opened 2016-05-18
Redux rx conversion

kenwheeler:
DONT MERGE YET

This converts our pseudo rxdux to proper redux with redux-observable. Everything _appears_ to be working, and most of the tests are passing, but I would love it if someone could pull this down and give it a spin and try and shake out some issues. Also, having difficulty getting the executeCell agenda test to pass, would love a second set of eyes on that as well.


jdfreder:
Thanks for opening this @kenwheeler , it's looking good!

jdfreder:
I should be able to look at this and the bits you mention later today...

kenwheeler:
Looks like CI is passing now. I'm getting errors about python2 locally, anyone got any ideas?

rgbkrk:
I'm guessing you'll want to install the Python 2 kernel locally. Our default in tests here looks to be `python2`, for the sake of ease on Travis.

```
brew install python
python2 -m pip install ipykernel
python2 -m ipykernel install --user
```

rgbkrk:
We should make our tests work whether python2 or python3 is available so you don't have to have both, though I don't think you have to address that for this PR.

Thanks for this awesomeness @kenwheeler.

src/notebook/actions/index.js
```diff
@@ -32,62 +34,62 @@ export function newKernel(kernelSpecName, cwd) {
...
-  return (subject, dispatch) => {
-    subject.next({
+  return (actions, store) => Rx.Observable.create((subscriber) => {
+    subscriber.next({
       type: constants.CHANGE_FILENAME,
       filename,
     });
-    dispatch(save(filename, notebook));
-  };
+    store.dispatch(save(filename, notebook));
```
rgbkrk, line 84: Later we can probably refactor this to merge the observable from `save` instead of doing another dispatch.

rgbkrk:
Notes so far:
- I haven't been able to switch to another kernel successfully
- Cell focus is a bit... odd. It seems like I can't select into codemirror to type into a cell above - I have to select the cell then click in the codemirror area to execute. This could be related to the changes with the cell toolbar (#465), though I can't tell for sure.

rgbkrk:
Everything else is quite flawless, all execution and cell ops work really well.

rgbkrk:
Ok weird, I can no longer reproduce the lack of ability to switch kernels. Not sure what's up.

jdfreder:
> Ok weird, I can no longer reproduce the lack of ability to switch kernels. Not sure what's up.

I had this problem once locally, but I resolved it by reinstalling the kernels.

jdfreder:
I'll take a look now :)

jdfreder:
:+1: :fireworks: :100: Thanks @kenwheeler !

PR #480, opened 2016-05-19
Pumpkin: absolute->fixed.

Carreau:
Cause I never remember which one is which is which in css and even
actually never remember that the 2 exists. And this one is the one we
want.


jdfreder:
I just had this problem a second ago too in the sticky cell PR!  lol

PR #460, opened 2016-05-17
Spec the css structure

jdfreder:
diagram pending, for now:

```
.notebook
  .cell.code
    .cell-toolbar
    .input-container
      .prompt
      .input
    .pagers
      .pager
    .outputs
  .cell.text
    .cell-toolbar
    .input-container
      .prompt
      .input
    .rendered
  .cell-creator
```

PR #487, opened 2016-05-20
Precursor to widgets, ipywidgets, jupyter-js-widgets

jdfreder:
widgets


captainsafia:
OMG.

jdfreder:
This is working now:

widgets

It requires you to have version 5.x of ipywidgets (latest stable).  Note, you don't need to install widgetsnbextension or activate the nbextension.  Just `pip install ipywidgets`

The `Widget Javascript not detected.  It may not be installed properly. Did you enable the widgetsnbextension? If not, then run "jupyter nbextension enable --py --sys-prefix widgetsnbextension"` message is showing because I haven't implemented the frontend version validation comm yet...  It should be a simple fix though.  

Some widgets wont render correctly because there's a lot of missing CSS.

Also I haven't added any unit tests for this code.

jdfreder:
cc @rgbkrk @sylvaincorlay

jdfreder:
Okay, fixed the version validation:

widgetsworking

rgbkrk:
Whoa! Great to see.

SylvainCorlay:
nice!

Quick question re: styling. In bqplot we have custom styling when the widget view is contained in something that has a css class `theme-dark`.

This made me think of what you mentioned earlier about having (some) standardization for the DOM structure and css classes, which would be useful to replace this.

rgbkrk:
There's a high likelihood there will be other themes, as we already have three themes: 
- light
- dark
- classic (Jupyter)

nteract currently relies on CSS variables, which I'd _love_ to adapt to make frontend widgets and plots adapt dynamically to the user's theme.

SylvainCorlay:
If we have some commonalities about how theming is done between jupyterlab, nteract, the dashboard server extensions etc, widget libraries could easily be rendered correctly in all those cases.

rgbkrk:
That would be really nice.

SylvainCorlay:
Having top-level css classes indicated categorization of themes (dark | light), (grayscale | colored) so that one could adjust without getting into theme idiosyncrasies could be a way to go...

rgbkrk:
nteract is currently dependent on class hierarchy and CSS variables. The DOM structure does have to stay consistent in order for us to write multiple themes.

Current "spec": https://github.com/nteract/nteract/pull/460

If there was a way, I'd prefer for us to rely on more CSS variables instead of class hierarchy. The DOM structure may need to change after all.

jdfreder:
Hey,

I'm just dropping in briefly from Cancun Mexico via paid internet so not much time to type...  I think the idea of specing some type of styling, either CSS var based or DOM structure based, would be great.  I agree that CSS vars are probably the way to go as it allows lab/nteract/notebook to structure widgets differently if needed and also allows custom widget authors to "recylce" theme colors/styles.

On a separate note, if either of you want to jump in on the action while I'm away, or @captainsafia @kenwheeler , feel free to do so!  There's a lot that can still be done.  Otherwise I'll be back later next week and I'll try to drop in every now and then when I'm able to get access to the internet. :)

jdfreder:
Widget theme support via CSS vars:

widgetthemes

cc @SylvainCorlay

SylvainCorlay:
Cool! Does it come with an ipywidgets PR for the css? 
Cc @jasongrout

jdfreder:
> Cool! Does it come with an ipywidgets PR for the css?

I think it would be great to use the CSS there too.  The CSS featured in the gif is a complete replacement for the CSS included with the widgets.  I did this to avoid dependency on bootstrap CSS and jqueryui CSS and to make sure everything used CSS variables where appropriate, for tight integration with themes.

jdfreder:
However, it wouldn't be a direct port - this CSS is for ipywidgets stable, not dev.

SylvainCorlay:
Ok good to know. Some layout related things seem to have changed too.
On Jun 4, 2016 5:54 PM, "Jonathan Frederic" notifications@github.com
wrote:

However, it wouldn't be a direct port - this CSS is for ipywidgets stable,
not dev.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
https://github.com/nteract/nteract/pull/487#issuecomment-223762978, or mute
the thread
https://github.com/notifications/unsubscribe/ACSXFsbu827FwftbicHNY5ODw-Y3V5VOks5qIZ_RgaJpZM4IjdYW
.

jdfreder:
TODO:
- [x] Get sliders to render correctly  
- [x] Associate state update messages to sending cell

jdfreder:
I've hit all the major goals for this PR and now I want to go back through and clean things up.  Also, it seems like there is a race condition somewhere in the RxJS stuff, either at the widget/comm level or lower.

rgbkrk:
Nice, all set with tests.

src/notebook/agendas/index.js
```diff
@@ -83,6 +86,11 @@ export function executeCell(channels, id, code) {
     const subscriptions = [];
 
     const executeRequest = createExecuteRequest(code);
+    subscriber.next(associateCellToMsg(id, executeRequest.header.msg_id));
+
+    // Clear the widget area on cell execution
+    // TODO: Don't special case the widgets
+    subscriber.next(clearWidgets(id));
```
rgbkrk: This should probably be cleared up higher, before the cell association.

src/notebook/agendas/index.js
```diff
@@ -117,10 +125,18 @@ export function executeCell(channels, id, code) {
...
-                               .share();
-
-    childMessages
+    // Messages that should affect the cell's output are both messages child
+    // to the execution request and messages mapped to the cell (from widget
+    // interaction, for example).
+    const cellMessages = iopub
+      .filter(msg => {
+        const state = store.getState();
+        return executeRequest.header.msg_id === msg.parent_header.msg_id || // child msg
```
rgbkrk, line 128: We do have a helper for this - `isChildMessage`.

rgbkrk, line 128: Oh I see, sorry.

src/notebook/agendas/index.js
```diff
@@ -117,10 +125,18 @@ export function executeCell(channels, id, code) {
...
+          state.document.getIn(['cellMsgAssociations', id]) === msg.parent_header.msg_id; // mapped
```
rgbkrk: Ok, cool. This makes the cross cell comms clear to me.

rgbkrk, line 129: This means current `cellID` is the same as the originating message id?

src/notebook/components/cell/cell-creator-buttons.js
```diff
@@ -16,6 +17,7 @@ export class CellCreatorButtons extends React.Component {
 
   constructor() {
     super();
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
```
rgbkrk, line 20: These are the ones I wondered about you extracting to a separate PR.

src/notebook/components/cell/code-cell.js
```diff
@@ -1,76 +1,99 @@
...
+          }
+          </div> : null
+      }
+      {
+        this.props.widgets && !this.props.widgets.isEmpty() ?
+          <WidgetArea
+            id={this.props.id}
+            widgets={this.props.widgets}
+            widgetManager={this.props.widgetManager}
+          /> : null
```
rgbkrk: Ok, cool. This puts the widget area into its own subsection like the current notebook. Still a bit awkward on the document, though it's pretty much the same setup as we do for the pager here.

src/notebook/components/cell/inputs.js
```diff
@@ -1,16 +1,25 @@
...
 
-export default function Inputs({ executionCount, running }) {
-  const count = ! executionCount ? ' ' : executionCount;
-  const input = running ? '*' : count;
-  return (
-    <div className="prompt">
-      [{input}]
-    </div>
-  );
-}
```
rgbkrk, line 11: Any of these that are pure functions use the pure rendering by default if it doesn't use context. You don't need the change for this one.

src/notebook/components/cell/pager.js
```diff
@@ -1,20 +1,28 @@
...
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);
+  }
+
+  render() {
+    const bundle = this.props.pager.get('data');
+    const mimetype = richestMimetype(bundle, this.props.displayOrder, this.props.transforms);
+    const Transform = this.props.transforms.get(mimetype);
+    return <Transform data={bundle.get(mimetype)} />;
+  }
+}
```
rgbkrk, line 27: Same with this one, we shouldn't break these out into classes again. Of note: we can now return `null` in these functions now that we're on React 15.

src/notebook/components/cell/inputs.js
rgbkrk, line 11: Nevermind me, I've got this wrong I think. Ping @kenwheeler for interpretation of pure functions and whether React now optimizes them for `shouldComponentUpdate`.

src/notebook/components/cell/widget.js
```diff
@@ -0,0 +1,54 @@
...
+        this.props.widgetManager.createViewForModel(
+          this.props.id,
+          this.props.cellId
+        ).then(view => {
+          try {
+            placeholder.parentNode.replaceChild(view.el, placeholder);
+            this.disposeView();
+            this.view = view;
+          } catch (err) {
+            // NOMNOM
```
rgbkrk: ಠ_ಠ

src/notebook/widgets/backend-sync.js
```diff
@@ -0,0 +1,229 @@
...
+      // The observable should emit false initially and then true when the
+      // widget framework version is validated.
+      return Rx.Observable.of(false)
+        .merge(Rx.Observable.fromPromise(validationPromise));
+    })
+    .distinctUntilChanged()
+    .publishReplay(1) // Only remember the last state
+    .refCount();
+
+    // Subscribe to the overservable to trigger it to process messages and send
```
rgbkrk: ...observable...

src/notebook/widgets/backend-sync.js
```diff
@@ -0,0 +1,229 @@
...
+    // the validation to the backend.  For convience, store the results on the
```
rgbkrk: ...convenience...

src/notebook/widgets/backend-sync.js
```diff
@@ -0,0 +1,229 @@
...
+      .subscribe(() => {
+        backendStateSubscription.unsubscribe();
+        frontendStateSubscription.unsubscribe();
+        displaySubscription.unsubscribe();
+        customMsgSubscription.unsubscribe();
+        deleteSubscription.unsubscribe();
+        deleteWidget(model.id);
+      });
+  }
+}
```
rgbkrk: This is quite a sweet encapsulation of the widget machinery. :smile:

src/notebook/widgets/comms.js
```diff
@@ -0,0 +1,186 @@
...
+ * For a POC, see http://jsbin.com/tujapaciwo/5/edit?js,console
+ * @param  {Redux.store} store
+ * @return {Observable} observable of the kernel channels object
+ */
+export function getChannels(store) {
+  // For a proof of concept, see the following JSBin:
+  // http://jsbin.com/tujapaciwo/5/edit?js,console
+
+  // Defer the initial value till the channels are subscribed to.  This will
+  // make it so the subscribe event.
```
rgbkrk: What's the end of this comment?

src/notebook/agendas/index.js
jdfreder, line 129: cellMsgAssociations is a map of cell id to msg id

src/notebook/widgets/comms.js
```diff
@@ -0,0 +1,186 @@
...
+      currentChannels = store.app.channels;
+    }
+    return currentChannels;
+  })
+
+  // Listen to channel change events.
+  .merge(Rx.Observable.from(store)
+    .pluck('app')
+    .pluck('channels')
+  )
```
rgbkrk: We actually do something similar at a top level too, we may want to refactor that higher up.

src/notebook/widgets/comms.js
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+ * Get an Observable for comm close messages.
+ * @param  {Redux.store} store
+ * @return {Observable} Observable of comm close messages.
+ */
+export function commCloseMessages(store) {
+  return commRelatedMessages(store)
+    .filter(msg => msg.header.msg_type === 'comm_close');
```
rgbkrk: how is store used here?

src/notebook/agendas/index.js
rgbkrk, line 129: Derp me, it's retrieving the message id by cell id. Thanks.

src/notebook/widgets/comms.js
jdfreder: `store` is passed to `commRelatedMessages` which passes it to `getChannels`

static/styles/theme-light.css
```diff
@@ -1,43 +1,3 @@
...
-
-  --cm-activeline-bg: #e8f2ff;
-  --cm-matchingbracket-outline: grey;
-  --cm-matchingbracket-color: black;
-
-  --cm-hint-color: var(--cm-color);
-  --cm-hint-color-active: var(--cm-color);
-  --cm-hint-bg: var(--main-bg-color);
-  --cm-hint-bg-active: #ABD1FF;
+  /* Uses the defaults in main.css */
```
rgbkrk: Hmmm...

static/styles/main.css
```diff
@@ -4,6 +4,48 @@
 :root {
   --prompt-width: 50px;
   --cell-placeholder-opacity: 0.3;
+
+  --main-bg-color: white;
```
rgbkrk: We should still have the defaults broken out I think, so we're not mixing what's configurable and what's not.

rgbkrk:
I'd love to see the pure component render stuff extracted, but I think I'll go ahead and merge this. I personally feel conflicted in that this directly ties us (yet again) to a specific version of the widgets.

Current cons:
- elevates the widgets to having their own specific section of the document as part of cell areas, not leaving room for how bokeh, matplotlib notebook mode, and others that use the comms
- ties us to a specific version of the widgets. This is the same problem we were facing with Thebe and generally across the notebook; it is shelved here again. We need a static resource spec to fetch. Relatedly...
- our shim layer has to match the version of ipywidgets in use

rgbkrk:
Actually, I'll step away from merging in case others have comments. Following C4.1, this PR is much larger than we should be dealing with. Where indicated above, also in accordance with C4.1, we have ourselves a "lack of spec" issue. This is technical debt brewing in our future.

static/styles/main.css
jdfreder: Yeah, I can un-break this out.  It's a little akward though because the widget theme variables inside main.css depend on these, even though the way CSS evaluates everything will work regardless of order.

jdfreder: As an aside, I don't think it makes sense to add widget CSS vars to every theme, and yet I think they should work with simple theme defaults

jdfreder:
> elevates the widgets to having their own specific section of the document as part of cell areas, not leaving room for how bokeh, matplotlib notebook mode, and others that use the comms

There are a couple of alternative ways to handle this:   
- One is to dump widgets directly into the output area of the cell instead of a "widget area".  This would break simple animation, but this is the path JupyterLab chose.  
- The other is to have a page level widget area, that all cells share.  This could even live in a separate window.

> ties us to a specific version of the widgets. This is the same problem we were facing with Thebe and generally across the notebook is shelved here again. We need a static resource spec to fetch. Relatedly...
> our shim layer has to match the version of ipywidgets in use

There is no way to avoid this- we can move the nteract widget specific JS into another package, but that package will still be dependent on specific backend version of the widgets, this is the nature of the beast.  See the "versionValidation" logic in the backend-sync module.

> Where indicated above, also in accordance with C4.1, we have ourselves a "lack of spec" issue.

I'd be happy to create a widget communication spec here, but it really belongs in Jupyter.

src/notebook/widgets/comms.js
jdfreder: I must have :sleeping: - strange, period and all...

src/notebook/components/cell/widget.js
jdfreder: Whoops, this was left over from early on.

jdfreder:
Here's an idea: let's move away from widget area, dump widgets straight into output, and I can make widgets a true and through npm package "extension" of nteract.  Of course, we'd need to spec how extensions get registered/loaded (my vote is for the npm API).

static/styles/theme-light.css
jdfreder: :heavy_check_mark:

static/styles/main.css
jdfreder: :heavy_check_mark:

rgbkrk:
The npm package approach sounds worth exploring.

rgbkrk:
> I'd be happy to create a widget communication spec here, but it really belongs in Jupyter.

It's ok for it to exist in Jupyter, that's where the notebook format and message spec come from after all.

src/notebook/widgets/comms.js
jdfreder: :heavy_check_mark: 

``` js
// Defer the initial value till the channels are subscribed to.  This will
  // make it so the action of a subscription will cause the current channels to
  // be emitted, regardless of whether or not they were set long before when
  // the subscriber subscribes.
```

src/notebook/widgets/backend-sync.js
jdfreder: :heavy_check_mark:

src/notebook/widgets/backend-sync.js
jdfreder: :heavy_check_mark:

jasongrout:
> One is to dump widgets directly into the output area of the cell instead of a "widget area". This would break simple animation, but this is the path JupyterLab chose.

Right now (this very minute :), I'm working on implementing OutputWidget, which would restore simple animation in a way I think is more powerful than the current notebook.

So far, I like our approach of making widgets just another pluggable renderer for display data. Widgets are not special in this design, and it would be easy to implement other renderers that were just as powerful as widgets (for example, dynamic charting, etc.).

rgbkrk:
> it would be easy to implement other renderers that were just as powerful as widgets (for example, dynamic charting, etc.).

I've definitely been hoping for a vega-lite mimetype that we could use regardless of the rendering environment.

.eslintrc
```diff
@@ -9,6 +9,7 @@
   "rules": {
       "camelcase": [0],
       "no-console": [2, { allow: ["warn", "error"] }],
-      import/no-unresolved: [0]
+      import/no-unresolved: [0],
+      "no-underscore-dangle": [0],
```
rgbkrk, line 13: You too sneaky.

jdfreder, line 13: Ah, that was to deal with the jupyter-js-widgets API which is really Pythonic.  It can be removed, I just missed it on my pass through.

PR #489, opened 2016-05-21
Super upgrade

rgbkrk:
This replaces #453, doing an overall upgrade and fixing the roundtrip test.


rgbkrk:
and so it is. NODE 6

Make sure to nuke your `node_modules` before doing a fresh `npm install` when on node 6.

PR #491, opened 2016-05-21
Reduce our babel footprint, fix electron deprecate

rgbkrk:
I've been hoping to do this for a while! This reduces our babel footprint quite a bit, from > 24 plugins down to ~4 (not including React in either case).

While making sure this was good to go, I fixed up a deprecation from Electron within the menu handlers.


.babelrc
```diff
@@ -1,6 +1,6 @@
   {
     "presets": [
-      "es2015",
+      "es2015-node6/object-rest",
```
rgbkrk, line 3: Note that because we're using `transform-object-rest-spread` below, we have to have a reduced version of this preset.

jdfreder, line 3: 👍

PR #494, opened 2016-05-22
When shift-enter or ctrl-enter is hit, keep view on MD cell

rgbkrk:
Allow for shift-entering your way through a notebook via Markdown Cells. Classic.


rgbkrk:
I'll get some tests on this one while I'm at it.

rgbkrk:
I'll get some tests into here with enzyme that test keyboard actions then this will be ready to go.

Issue #459, opened 2016-05-16
Publish to gist should ask for credential or anonymous.


Carreau:
I just publish to-gist. 

I was expecting to have to enter my credentials/ personal tokens. 

It published as anonymous. 

I would suggest popping a windows with "publish as [anonymous]/[Username]" everytime, [Username], ask for credentials.

captainsafia:
We can close this now!

Yay! Thanks @jdetle!

PR #496, opened 2016-05-23
Add more reducer tests

captainsafia:
Tests are currently failing for `toggleStickyCell` are failing. I don't think it's an issue with the way the tests are structured but will poke around some more.

### DIFF
--- a/test/renderer/agendas/index_spec.js
+++ b/test/renderer/agendas/index_spec.js
@@ -5,7 +5,7 @@ import { liveStore, dispatchQueuePromise, waitForOutputs } from '../../utils';
 
 describe('agendas.executeCell', function() {
   this.timeout(5000);
-  it('produces the right output', (done) => {
+  it('produces the right output', () => {
     return liveStore((kernel, dispatch, store) => {
       const cellId = store.getState().document.getIn(['notebook', 'cellOrder', 0]);
       const source = 'print("a")';
--- a/test/renderer/agendas/index_spec.js
+++ b/test/renderer/agendas/index_spec.js
@@ -5,7 +5,8 @@ import { liveStore, dispatchQueuePromise, waitForOutputs } from '../../utils';
 
 describe('agendas.executeCell', function() {
   this.timeout(5000);
-  it('produces the right output', () => {
+  it('produces the right output', (done) => {
+    setTimeout(done, 5000);
     return liveStore((kernel, dispatch, store) => {
       const cellId = store.getState().document.getIn(['notebook', 'cellOrder', 0]);
       const source = 'print("a")';
--- a/test/renderer/agendas/index_spec.js
+++ b/test/renderer/agendas/index_spec.js
@@ -6,7 +6,6 @@ import { liveStore, dispatchQueuePromise, waitForOutputs } from '../../utils';
 describe('agendas.executeCell', function() {
   this.timeout(5000);
   it('produces the right output', (done) => {
-    setTimeout(done, 5000);
     return liveStore((kernel, dispatch, store) => {
       const cellId = store.getState().document.getIn(['notebook', 'cellOrder', 0]);
       const source = 'print("a")';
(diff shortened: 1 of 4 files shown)


test/renderer/reducers/document_spec.js
```diff
@@ -105,6 +125,26 @@ describe('focusPreviousCell', () => {
...
+
+    const id = originalState.document.notebook.get('cellOrder').first();
+
+    const action = {
+      type: constants.TOGGLE_STICKY_CELL,
+      id,
+    }
+
+    const state = reducers(originalState, action);
+    expect(state.document.stickyCells.get(id)).to.be.true;
```
jdfreder: You need to add stickyCells to the document key of your originalState above.

rgbkrk:
Want to rebase?

captainsafia:
Guh. I'm not sure why this `executeCell` test keeps timing out **and** why it only happens on this PR. I'll investigate later, but these should move us forward on coverage.

test/renderer/agendas/index_spec.js
```diff
@@ -5,7 +5,8 @@ import { liveStore, dispatchQueuePromise, waitForOutputs } from '../../utils';
 
 describe('agendas.executeCell', function() {
   this.timeout(5000);
-  it('produces the right output', () => {
+  it('produces the right output', (done) => {
+    setTimeout(done, 5000);
     return liveStore((kernel, dispatch, store) => {
```
rgbkrk, line 10: This looks like it predates our Redux Observable setup, since `store.dispatch` is our actual dispatch now.

test/renderer/agendas/index_spec.js
```diff
@@ -5,7 +5,8 @@ import { liveStore, dispatchQueuePromise, waitForOutputs } from '../../utils';
 
 describe('agendas.executeCell', function() {
   this.timeout(5000);
-  it('produces the right output', () => {
+  it('produces the right output', (done) => {
+    setTimeout(done, 5000);
```
jdfreder: This setTimeout line will cause the test to pass even if it should fail because of a timeout (set by this.timeout above).

test/renderer/agendas/index_spec.js
jdfreder, line 10: :+1:

jdfreder, line 10: Actually, looking in test/utils.js at the `liveStore` function, that's not the problem.  It's correct here.

test/renderer/agendas/index_spec.js
captainsafia: Yeah, I'm trying to debug a test failure that for some strange reason only happens in this PR. Prior to me making this change, the test would fail with:

`Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.`

I tried a couple of things on my local.
1. Passing `done` to the test case and calling it.
2. Calling `setTimeout(done, 5000)` inside the test case and removing `this.timeout(5000)` in the test suite.
3, Changing the timeout value.

I'm not entirely sure it is something to do with the tests because it does not happen on others PRs where the test is set up this way (#459).

¯_(ツ)_/¯

jdfreder: Ah, it's similar to a bug I just dealt with.  This same test copy and pasted into a test folder called "widgets" would fail because it was being executed after the "components" tests.  The trick in that case was that the bug was in the component tests, specifically with the Latex component.  I added a global flag to disable MathJax per test and that seemed to do the trick.  Maybe the timeout is something similar?  Have you tried making the timeout more verbose?  It may be eating an error which is causing the timeout...

jdfreder: Also, don't forget to rebase on master.  This test will timeout intentionally if code cell execution doesn't work on the branch.

captainsafia: I've already rebased on master, the issue persists before and after a rebase. But I'll look into the trouble you went through with MathJax.

I tried disabling the timeout all together to see if it picked up any errors but it just hangs forever (a long time) and it seems like the Promise is never resolved.

captainsafia:
I think this is good for merge. To resolve the issue with the `executeCell` test I took some code out of @jdfreder's widgets PR which retries the test until whatever weird race condition with executeCell is happening is avoided.

jdfreder:
> @jdfreder's widgets PR which retries the test until whatever weird race condition with executeCell is happening is avoided.

Awesome, sounds good.  I'm working on a fix for the underlying problem in the widgets PR.

PR #498, opened 2016-05-24
Abort execution when a new execute is requested

rgbkrk:
Relying on `takeUntil` paired with `actions` to close down an agenda.

### DIFF
--- a/src/notebook/actions/index.js
+++ b/src/notebook/actions/index.js
@@ -231,6 +231,8 @@ export function toggleStickyCell(id) {
 
 export function executeCell(channels, id, source, kernelConnected, notificationSystem) {
   return (actions, store) => Rx.Observable.create((subscriber) => {
+    store.dispatch({ type: 'ABORT_EXECUTION', id });
+
     if (!kernelConnected) {
       notificationSystem.addNotification({
         title: 'Could not execute cell',
@@ -241,7 +243,10 @@ export function executeCell(channels, id, source, kernelConnected, notificationS
       return;
     }
 
-    const obs = agendas.executeCell(channels, id, source);
+    const obs = agendas.executeCell(channels, id, source).takeUntil(
+      actions.filter(x => x.type === 'ABORT_EXECUTION' && x.id === id)
+    );
+
     obs.subscribe(action => {
       subscriber.next(action);
     }, (error) => {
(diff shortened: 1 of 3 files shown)


src/notebook/actions/index.js
```diff
@@ -231,6 +231,8 @@ export function toggleStickyCell(id) {
 
 export function executeCell(channels, id, source, kernelConnected, notificationSystem) {
   return (actions, store) => Rx.Observable.create((subscriber) => {
+    store.dispatch({ type: 'ABORT_EXECUTION', id });
```
jdfreder, line 234: `.dispatch` undefined here is causing the tests to fail.  Isn't accessing the store in an action an anti pattern?  Also, in actions, don't we normally do this with a store subscriber and `subscriber.next` instead of `store.dispatch`?

rgbkrk, line 234: Our base store is now proper Redux, paired with [redux-observable](https://github.com/redux-observable/redux-observable). The way to [cancel async actions](https://github.com/redux-observable/redux-observable#cancellation) is through other actions:

``` js
dispatch(
  (actions) => Observable.timer(1000)
    .map(() => ({ type: 'TIMER_COMPLETE'}))
    .takeUntil(
      actions.ofType('ABORT_TIMER')
    )
    // `actions.ofType('ABORT_TIMER')` is equivalent to
    // `actions.filter(action => action.type === 'ABORT_TIMER')`
);

// elsewhere in your code you can abort with a simple dispatch
dispatch({ type: 'ABORT_TIMER' });
```

You're absolutely right about using `subscriber.next({ type: 'ABORT_EXECUTION', id })` here. I'll have to update and rebase this PR.

rgbkrk, line 234: We do get [access to the `store`](https://github.com/redux-observable/redux-observable#other-api-notes) as the second item after `actions`.

rgbkrk, line 234: Update: um, what? `store` _is_ `undefined` here

rgbkrk, line 234: Further update: the store needs to be instantiated properly for both redux and Immutable.js in the test using it. See 5dbb43063f05753eb7051075031b501c7b2e8182

jdfreder:
Thanks Kyle! :cake:

PR #504, opened 2016-06-03
Fix references to state.document.notebook

captainsafia:
Immutable.js objects don't play nice with ES6 destructuring.

### DIFF
--- a/src/notebook/menu/index.js
+++ b/src/notebook/menu/index.js
@@ -100,8 +100,8 @@ export function dispatchPublishGist(store, dispatch) {
 
 export function dispatchRunAll(store, dispatch) {
   const state = store.getState();
-  const { channels, executionState } = state.app;
-  const { notebook, notificationSystem } = state.document;
+  const { channels, executionState, notificationSystem } = state.app;
+  const notebook = state.document.get('notebook');
   const cells = notebook.get('cellMap');
   const kernelConnected = channels &&
     !(executionState === 'starting' || executionState === 'not connected');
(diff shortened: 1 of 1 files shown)


src/notebook/menu/index.js
```diff
@@ -36,7 +36,8 @@ export function triggerSaveAs(store, dispatch) {
         return;
       }
       const state = store.getState();
-      const { notebook, executionState } = state.document;
+      const { executionState } = state.document;
```
jdfreder, line 39: Here you use `state.document.executionState` but above you use `state.document.get` to access notebook.  Is `state.document` immutable or not?

jdfreder, line 39: IOW, I think one of the lines is wrong

jdfreder, line 39: Ah, looks like you are using `get` everywhere.  This line should be a `get` call too.  If `get` causes it to break, that means the code setting the execution state on the state.document is doing it wrong

rgbkrk, line 39: Yeah this needs to be extracted as well using `.get`

rgbkrk:
These are all things I missed during the Immutable refactor because they weren't tested (there's coverage, though it doesn't appear to be checking the logic).

src/notebook/menu/index.js
captainsafia, line 39: Whoops! Yeah, I missed this one. No PRs after midnight for me!

captainsafia:
I'll open another PR later today with fixes and do a more thorough pass for this in the code.

PR #506, opened 2016-06-07
Make sure kernel is ready before sending anything to it!

jdfreder:
Extracted from #487


rgbkrk:
How's this work UI/UX wise?

Makes me think we'll want to make this into an observable action later, for the way we use it. I'm guessing for slow startup languages like Julia we'll want to show to the user it's still warming up.

jdfreder:
This seems to make the timeout more ... ug.  Will investigate more later

rgbkrk:
Sounds good!

PR #507, opened 2016-06-08
React redux

kenwheeler:
Switches to redux provider and connect HoC's


rgbkrk:
Slick, this is working quite well. Thank you!

PR #508, opened 2016-06-08
Git ignore test notebook output

rgbkrk:
Keeps the git workspace clean.

In the future, these tests should be using mocks instead.


jdfreder:
:tada:

PR #509, opened 2016-06-08
Setup an Immutable Record for document store

rgbkrk:
We can play a little bit less loosey goosey here and make a specific `Record`. Amusingly, if I would have gone with this originally, we could have kept our destructuring in the reducers. :wink:

Worth thinking about this for the notebook model in `commutable` too, for those that we know the key exactly (instead of a `Map`).


jdfreder:
:fish: :fish_cake:

PR #515, opened 2016-06-12
Update app_spec test format to match document_spec

captainsafia:
I noticed this while working on undo/redo and it annoyed me a bit. We should use the same test format for testing both app and document reducers. I prefer the document-spec-style because it makes the actions more evident.


rgbkrk:
Thanks! That will help when I convert the app state to immutable.

PR #517, opened 2016-06-14
App state should use Immutable.Record

captainsafia:
I heard you liked `Record`s.

### DIFF
--- a/src/notebook/index.js
+++ b/src/notebook/index.js
@@ -37,6 +37,14 @@ if (process.env.GITHUB_TOKEN) {
   }, (x) => console.error(x));
 }
 
+const AppRecord = new Immutable.Record({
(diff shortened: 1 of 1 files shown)


rgbkrk:
Yusssss I dooooo

rgbkrk:
This only keeps the `app` state as a Record before the first reduction. It's a normal object after (need to change all the reducers too).

rgbkrk:
(I'll tackle this right now, amusingly I have this in a branch too).

captainsafia:
> This only keeps the app state as a Record before the first reduction. It's a normal object after (need to change all the reducers too).

OHHHH!

> (I'll tackle this right now, amusingly I have this in a branch too).

Sweet! I'll merge when you're ready.

PR #520, opened 2016-06-14
Cleanup filename state

rgbkrk:
`filename` was being set and used in both the `app` tree and the `document` tree. To keep it consistent, I switched it over to the the `document` tree.


rgbkrk:
This also fixes a bug where the filename state wasn't properly updating in the title. Hoorah!

rgbkrk:
Good catch codecov!

PR #521, opened 2016-06-15
Refactor app/document store structure

captainsafia:
This PR is meant to rework how we currently have the state store set up. Our current setup,

```
{
   app: {}
   document: {}
}
```

presents some difficulties when trying to address things like undo and redo. For undo/redo, we want to store the stack of past and future states in an object that is not `document`, but storing it in `app` forces us to have to expose the overall state tree to some actions in the app or document reducer.

The result is to rework our state store tree and our reducers to something like this.

```
{
    app: {},
    documentMetadata: {
       past: [],
       future: [],
       filename: '',
       document: {}
    }
}
```

This is very WIP, but my goal is that we can have reducers that focus purely on application-level stuff (kernels and stuff), a reducer for managing document metadata and actions on the _entire_ document object (just undo/redo for now, but I imagine there will be other situations where we want to manipulate an entire document object), and reducer on the document itself (cell-level operations).

TL;DR: Reworking state tree so that it better reflects the kind of data access our actions need.

I'm about to BREAK ALL THE THINGS.

### DIFF
--- a/src/notebook/index.js
+++ b/src/notebook/index.js
@@ -47,7 +47,6 @@ const AppRecord = new Immutable.Record({
 
 const DocumentRecord = new Immutable.Record({
   notebook: null,
-  filename: '',
   cellPagers: new Immutable.Map(),
   cellStatuses: new Immutable.Map(),
   stickyCells: new Immutable.Map(),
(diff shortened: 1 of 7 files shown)


captainsafia:
OK. I reworked this a little bit from.

```
{
  app: {}
  documentMetadata: {
    past: [],
    future: [],
    filename: '',
    document: {}
  }
}
```

to

```
{
  app: {},
  metadata: {},
  document: {},
}
```

I did this for a couple of reasons.
1. I read the Redux docs a little bit more and learned that there really isn't a way to do the undo/redo thing across parts of the state tree without referring to the other part of the subtree.
2. I couldn't get the nested reducer thing to work, but that might be because I am a newb.

So the next best thing I figured is to move the information that `document` needs to know that isn't pertinent cells or the actual document to a `metadata` subtree. That way, when that subtree has to (irrevocably) be passed to the `document` reducers, we are at least passing the most pertinent information and not the entire `app` subtree.

rgbkrk:
Cool, looking good.

PR #525, opened 2016-06-17
Clear the output of a single cell

captainsafia:
Zap!

Watch me fumble with the disappearing toolbar.

clear-cell-output

### DIFF
--- a/test/renderer/components/cell/cell-toolbar-spec.js
+++ b/test/renderer/components/cell/cell-toolbar-spec.js
@@ -3,6 +3,8 @@ import React from 'react';
 import { mount } from 'enzyme';
 import {expect} from 'chai';
 
+import { dummyStore } from '../../../utils';
+
 import { Toolbar } from '../../../../src/notebook/components/cell/toolbar';
 
 describe('Toolbar', () => {
@@ -21,5 +23,11 @@ describe('Toolbar', () => {
       clientX: 0,
       clientY: 0,
     })).to.not.throw(Error);
+  });
+  it('clearCellOutput does not throw error', () => {
+    const toolbar = mount(
+      <Toolbar />, { context: { store: dummyStore() }}
+    );
+    expect(() => {toolbar.instance().clearCellOutput()}).to.not.throw(Error);
   });
 });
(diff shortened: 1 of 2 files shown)


captainsafia:
I'm trying to figure out how to test this (and by extension all the other actions in the toolbar), but I'm having trouble because it looks like I need some way to stub the `context` variable in the component.

I found an [npm package](https://github.com/k88hudson/react-test-context) that creates a stub of the context in a component but it ends up overwriting the actions in the component so I can't actually call `clearCellOutput`.

Any other ways to do this?

rgbkrk:
If you split this component into dumb and smart components, you can test with the dumb component + enzyme much more simply (full control of context, props, etc.).

If it makes sense, I'd keep the `default` export:

```
export default connect(mapStateToProps)(Toolbar)
```

While also exporting `Toolbar` as `DumbToolbar` (or some other name like `UnconnectedToolbar`).

rgbkrk:
On a totally _separate_ note, if I saw the :zap: I would have _no_ idea what that was for. We can put it in for now (as the background engineering needed to be done), while I think that several of the other cell actions should be in a dropdown menu `...` with readable text (no requiring hover).

captainsafia:
> On a totally separate note, if I saw the ⚡ I would have no idea what that was for. We can put it in for now (as the background engineering needed to be done),

Yeah, I had trouble picking something from Octicons that fit this action. But yeah, it makes more sense to have some of our actions in a dropdown menu with plaintext, especially when there isn't a meaningful icon like in this case.

captainsafia:
> If you split this component into dumb and smart components, you can test with the dumb component + enzyme much more simply (full control of context, props, etc.).

I'm a silly goose! Enzyme allows you to pass in the context as the second parameter to the `mount` function.

test/renderer/components/cell/cell-toolbar-spec.js
```diff
@@ -22,4 +24,10 @@ describe('Toolbar', () => {
       clientY: 0,
     })).to.not.throw(Error);
   });
+  it('clearCellOutput does not throw error', () => {
+    const toolbar = mount(
+      <Toolbar />, { context: { store: dummyStore() }}
+    );
+    expect(() => {toolbar.instance().clearCellOutput()}).to.not.throw(Error);
```
rgbkrk, line 31: Check out the markdown cell tests if you want to see some action handling. You can get pretty sweet tests out.

PR #526, opened 2016-06-17
Add ability to interrupt kernel

### LINKED ISSUE #405: Ability to interrupt kernel
We only have a way to kill the kernel currently. We also need an "interrupt kernel" menu item.

captainsafia:
Closes #405.

Gotta figure out Windows support....

### DIFF
--- a/src/notebook/reducers/app.js
+++ b/src/notebook/reducers/app.js
@@ -38,7 +38,11 @@ export default handleActions({
   },
   [constants.KILL_KERNEL]: cleanupKernel,
   [constants.INTERRUPT_KERNEL]: function interruptKernel(state) {
-    state.spawn.kill('SIGINT');
+    if (process.platform === 'win32') {
+      console.error('Windows support for kernel interrupt not avaiable yet.');
+    } else {
+      state.spawn.kill('SIGINT');
+    }
     return state;
   },
   [constants.START_SAVING]: function startSaving(state) {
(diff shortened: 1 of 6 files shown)


captainsafia:
Hard to add Windows support when you don't have a Windows machine — so I just set it to error on  Windows for now.

Test for this is very meh. I'm not entirely sure if there is a way to check if a process has been interrupted, at least a way that is evident in `state.app.spawn`.

rgbkrk:
It's ok about the uncertainty with Windows. What does jupyter/notebook do?

src/notebook/reducers/app.js
```diff
@@ -37,6 +37,14 @@ export default handleActions({
     return cleanupKernel(state);
   },
   [constants.KILL_KERNEL]: cleanupKernel,
+  [constants.INTERRUPT_KERNEL]: function interruptKernel(state) {
+    if (process.platform === 'win32') {
+      console.error('Windows support for kernel interrupt not avaiable yet.');
```
rgbkrk, line 42: This should probably be a more user-visible error, using the notification system.

captainsafia:
`jupyter/notebook` uses [its own Windows interrupt module](https://github.com/jupyter/jupyter_client/blob/02f9c580f6533ba97d1fed7157bef80c69f0f4ed/jupyter_client/win_interrupt.py).

src/notebook/reducers/app.js
captainsafia, line 42: Oops!

PR #528, opened 2016-06-18
Clean up references to state

captainsafia:
Found some stale references while using the app.

### DIFF
--- a/src/notebook/menu/index.js
+++ b/src/notebook/menu/index.js
@@ -37,7 +37,7 @@ export function triggerSaveAs(store, dispatch) {
         return;
       }
       const state = store.getState();
-      const { executionState } = state.document;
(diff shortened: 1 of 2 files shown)


src/notebook/menu/index.js
```diff
@@ -136,15 +136,16 @@ export function dispatchInterruptKernel(store, dispatch) {
 
 export function dispatchRestartKernel(store, dispatch) {
   const state = store.getState();
+  const { notificationSystem } = state.app;
   const spawnOptions = {};
-  if (state && state.document && state.document.get('filename')) {
+  if (state && state.document && state.metadata.get('filename')) {
```
rgbkrk, line 141: Should probably be `state && state.metadata && state.metadata.get('filename')`

PR #530, opened 2016-06-20
Add ability to duplicate a notebook

### LINKED ISSUE #523: Add ability to clone a notebook
This is split off from the list that @rgbkrk produced in #416.

We should have a feature where an individual can make a copy of the currently open notebook.

Completely irrelevant, I wonder if Electron has a way for you to do the fun and bouncy `Make a Copy` thing on Mac.

captainsafia:
Closes #523.

WIP because I'm debugging a weird issue where it complains that `electron.BrowserWindow is not a constructor` when launching from a new file.

I also wanna figure out a better way to create multiple copies, as in the `getCopiedFilename` should have a way of creating non-colliding filenames.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -70,6 +70,10 @@ export const fileSubMenus = {
     },
     accelerator: 'CmdOrCtrl+Shift+S',
   },
+  duplicate: {
+    label: '&Duplicate Notebook',
+    click: createSender('menu:duplicate-notebook'),
+  },
 };
 
 export const file = {
@@ -404,6 +408,7 @@ export function loadFullMenu() {
         fileSubMenus.open,
         fileSubMenus.save,
         fileSubMenus.saveAs,
+        fileSubMenus.duplicate,
       ],
     };
 
--- a/src/notebook/native-window/index.js
+++ b/src/notebook/native-window/index.js
@@ -5,6 +5,8 @@ import fs from 'fs';
 import path from 'path';
 import { fromJS } from 'commutable';
 
+import { deferURL } from '../../main/launch';
+
 import Rx from 'rxjs/Rx';
 
 const HOME = home();
--- a/src/notebook/native-window/index.js
+++ b/src/notebook/native-window/index.js
@@ -1,7 +1,9 @@
 import { remote } from 'electron';
-const { getCurrentWindow } = remote;
+const { BrowserWindow, getCurrentWindow } = remote;
 import home from 'home-dir';
+import fs from 'fs';
 import path from 'path';
+import { fromJS } from 'commutable';
 
 import Rx from 'rxjs/Rx';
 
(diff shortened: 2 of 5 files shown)


captainsafia:
This should be ready for review and merge.

I did commit a developer sin in this. Since the duplication happens inside the `renderer` process of Electron, I couldn't use the `launchFilename` function located in `src/main/launch.js` because it relied on the `BrowserWindow` object from `electron` whereas I needed to use the `BrowserWindow` object from `electron`.remote.

My solution to this was very hacky. I copied the `launch` and `launchFilename` functions from `src/main/launch.js` to `src/notebook/native-window/index.js` where they relied on the `BrowserWindow` from the renderer process.

There is probably a way to clean this up and limit the amount of code duplication that is going on across these two files but I can't think of any at the moment.

captainsafia:
@rgbkrk: Peek-a-boo!

src/main/menu.js
```diff
@@ -404,6 +408,7 @@ export function loadFullMenu() {
         fileSubMenus.open,
         fileSubMenus.save,
         fileSubMenus.saveAs,
+        fileSubMenus.duplicate,
       ],
```
rgbkrk, line 412: 👍

src/notebook/utils/index.js
```diff
@@ -0,0 +1,40 @@
+import fs from 'fs';
+import path from 'path';
+
+function fileExists(filename) {
+  if (!filename) {
+    return false;
+  }
+
+  try {
+    return fs.statSync(filename).isFile();
```
rgbkrk: Careful using `statSync` without a timeout. If someone tries to use this against a file that's on a network share (likely on campuses, corporate networks), the _entire application will freeze_ waiting for the response of the stat call.

src/notebook/utils/index.js
```diff
@@ -0,0 +1,40 @@
...
+  const base = path.basename(filename, '.ipynb');
+  const dir = path.dirname(filename);
+  let index = 1;
+  let newFilename = `${dir}/${base}-Copy${index}.ipynb`;
+  while (fileExists(newFilename)) {
+    index = index + 1;
+    newFilename = `${dir}/${base}-Copy${index}.ipynb`;
+  }
+  return newFilename;
+}
```
rgbkrk: Rather than write these two functions, I'd check out https://github.com/raszi/node-tmp using its [async filename generation](https://github.com/raszi/node-tmp#asynchronous-filename-generation-1).

src/notebook/native-window/index.js
```diff
@@ -42,3 +46,38 @@ export function initNativeHandlers(store) {
...
+  return new Promise((resolve, reject) => {
+    fs.readFile(filename, {}, (err, data) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(launch(fromJS(JSON.parse(data)), filename));
+      }
+    });
+  });
+}
```
rgbkrk, line 83: Hmmm, this looks like duplicated logic between the main and renderer process for launching a new window.

captainsafia, line 83: Yeah, see my comment above.

rgbkrk, line 83: Ah, going back to it now.

rgbkrk:
> There is probably a way to clean this up and limit the amount of code duplication that is going on across these two files but I can't think of any at the moment.

All good, I'll happily play with refactoring this later.

captainsafia:
> All good, I'll happily play with refactoring this later.

OK! I will address your other comments and push. 🎉

captainsafia:
OK! That cleans it up a whole lot. Although, it is a bit annoying that you can't configure `tmp` to differentiate using numbers instead of cryptographically generated strings.

PR #531, opened 2016-06-20
Fix issue with travis and clearCellOutput test

captainsafia:
This fixes Travis and the issues that we were having with the clearCellOutput test.

`clearCellOutput` is one of the undoable tasks so when it is called `setBackwardCheckpoint` is called which looks for a `past` key in the `metadata` record. Since we passed a `dummyStore` without a `metadata` record to the context of the toolbar component, the tests got upset.


rgbkrk:
I'll figure out the functional test issues now, trying out spectron.

PR #535, opened 2016-06-24
[WIP] Add a right click menu for cells

captainsafia:
For some strange reason, clicking clear cell output clears the output of the next cell not the one clicked on.


src/notebook/components/cell/cell-context-menu.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { ContextMenu, MenuItem } from 'react-contextmenu';
```
rgbkrk, line 2: Interesting, I had not checked out a React component for a context menu. I was planning on using the context menu and events. Using React for this should prove fruitful!

PR #537, opened 2016-06-28
Add ability to split cell

captainsafia:
Split cell soup!


captainsafia:
OK! I feel like this can be merged because I don't want it to be a huge PR.

The next steps for this are.
- Fix bug in commutable and make it so that the newly created cell is added after the current one not at the end of the file
- Get the position of the cursor and use it when splitting, this will be made easier once I get context menu working

captainsafia:
cc: @rgbkrk @kenwheeler

src/notebook/reducers/document.js
```diff
@@ -138,6 +138,10 @@ export default handleActions({
     const { id } = action;
     return state.update('notebook', (notebook) => commutable.clearCellOutput(notebook, id));
   },
+  [constants.SPLIT_CELL]: function splitCell(state, action) {
+    const { id, position } = action;
+    return state.update('notebook', (notebook) => commutable.splitCell(notebook, id, position));
```
rgbkrk, line 143: It's ok to do the logic here for now and _then_ add it to commutable.

rgbkrk:
I'll merge it as is and look forward to you addressing the rest.

PR #538, opened 2016-06-29
Remove indeterminate executeCell test

jdfreder:
I'm on a test killing crusade.  Each unit test should execute in ~1ms, this one does not, therefor it will die. :skull:


rgbkrk:
You should move these over to the functional suite or an integration suite, since no replacement (proper mocks, better stubs) is provided here.

jdfreder:
I think it may make more sense to do the execution as a GUI test, since our current functional test is a GUI test and it would allow us to cover end-end.

rgbkrk:
Going ahead and merging since the test is passing for now. I'm assuming you'll follow up with moving it to functional or mocking later.

PR #541, opened 2016-07-01
Add context menu

captainsafia:
Yay! This works now.

context-menu

cc: @rgbkrk

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -136,4 +136,6 @@ class Cell extends React.Component {
   }
 }
 
-export default ContextMenuLayer("cell-context-menu")(Cell);
+export default ContextMenuLayer("cell-context-menu", (props) => ({
+  id: props.id,
+}))(Cell);
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -11,7 +11,7 @@ import Toolbar from './toolbar';
 
 import { focusCell, focusPreviousCell, focusNextCell } from '../../actions';
 
-class Cell extends React.Component {
+export class Cell extends React.Component {
   static propTypes = {
     cell: React.PropTypes.any,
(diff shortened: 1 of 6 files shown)


rgbkrk:
What's up with that cell hover state problem? Is the new component grabbing events needed elsewhere?

captainsafia:
> What's up with that cell hover state problem? Is the new component grabbing events needed elsewhere?

Debugging that now. It seems to be specific to the way that the Cell is configured in the test suite.

rgbkrk:
Super sweet by the way, loving it.

zackargyle:
Hey @captainsafia, here to help! What would you like me to look at?

src/notebook/components/cell/cell.js
```diff
@@ -134,4 +136,6 @@ class Cell extends React.Component {
   }
 }
 
-export default Cell;
+export default contextMenuLayer('cell-context-menu', (props) => ({
```
zackargyle, line 139: This will likely mess up any enzyme tests for `Cell`. The reason is that this actually returns a wrapper around Cell so that now it looks like this: `<ContextMenuLayer><Cell /></ContextMenuLayer>`. For your tests the best way to do it is to export the cell class above and then import that.

```
export class Cell extends React.Component { ... }
export default contextMenuLayer(...)(Cell);
```

Then in your test you would import it with the braces.

```
import { Cell } from '../../../../src/notebook/components/cell/cell';
```

captainsafia:
Hello @zackargyle! Thanks for dropping in and making the suggestion. Let me go ahead and try it really quick.

captainsafia:
@zackargyle — It works on local. Failing build is due to Travis being a bad boy. Silly Travis!

rgbkrk:
May want to rebase with the new setup now that #543 has been merged.

captainsafia:
:shipit: !

jdfreder:
negligible decrease, but we should strive for 100% coverage

PR #542, opened 2016-07-06
Add cell copy paste

captainsafia:
For some weird reason, it won't detect `Cmd + C` properly.  `handleKeyDown` seems to be configured properly to detect it but I might be missing something obvious.

Also, this is set up in way to allow multi-cell copy paste in the future.

cell-copy-paste


rgbkrk:
YaY!

src/notebook/components/cell/cell.js
```diff
@@ -74,6 +85,24 @@ class Cell extends React.Component {
...
+    if (this.state.ctrlDown) {
+      if (event.keyCode === 86) {
+        this.setState({ ctrlDown: false });
+        this.pasteCell();
+      } else if (event.keyCode === 67) {
+        this.setState({ ctrlDown: false });
+        this.copyCell();
+      }
+    }
+  }
```
rgbkrk, line 104: I'd look to the markdown test to see how to test these two keyboard driven functions.

rgbkrk:
I'd check that the copy action was dispatch (using `sinon` spies or otherwise). Checking state will work for now if it at least gets the coverage though.

captainsafia:
> I'd check that the copy action was dispatch (using sinon spies or otherwise). Checking state will work for now if it at least gets the coverage though.

I'm gonna do this know just so I don't get in the habit of writing bad tests. Gimme 2 seconds.

captainsafia:
:shipit: !

test/renderer/components/cell/cell-spec.js
```diff
@@ -32,4 +34,39 @@ describe('Cell', () => {
...
+      <Cell cell={commutable.emptyCodeCell} {...sharedProps}/>,
+      { context: { store: dummyStore() } }
+    );
+
+    const spy = sinon.spy(cell.instance(), "pasteCell"); 
+
+    cell.simulate('keydown', { key: 'Ctrl', ctrlKey: true});
+    cell.simulate('keyup', { keyCode: 86 });
+    expect(cell.state('ctrlDown')).to.be.false;
+    expect(spy.called).to.be.true;
```
rgbkrk, line 70: 👍

rgbkrk:
Well this is awesome.

captainsafia:
Me so happy.

PR #547, opened 2016-07-16
Fixed saveAs and duplicate functions

### LINKED ISSUE #546: Save running as Save As even on already saved file
While working on a file that has been already saved, Cmd + S brings up the Save As dialog instead of just immediately saving the file.

captainsafia:
Fixes #546.


rgbkrk:
Since the relevant tests are passing and it's the kernel lifecycle tests that are failing, I've gone ahead and merged this.

PR #554, opened 2016-07-24
Minimalist fix to get packaging working

jdetle:
Commander expects process.argv be `node`name.js`[args]. We are running an executable so`Commander's parse function gobbles up the first and second `process.argv` args. For this reason I spliced in an empty string as a hacky fix to get everything working. I changed .option because if you look at program initially `--kernel [kernel]` was the description of the flag and not an actual option. Now the option name is `program.kernel`.
## Todo?

As mentioned in the slack channel, currently we reject a promise and hang while opening no notebook if a file specified doesn't exist in the path. How do we want to go about fixing this?

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -11,7 +11,7 @@ const program = require('commander');
 const version = require('../../package.json').version;
 
 program._name = 'nteract';
-process.argv.splice(1,0,'')
+process.argv.splice(1, 0, '');
 
 program
   .version(version)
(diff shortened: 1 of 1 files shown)


rgbkrk:
Let's go with it for now. I have a feeling there's still going to be an issue because our `npm run start` args are longer than when it's packaged and run with electron.

¯\_(ツ)_/¯

jdetle:
Oh shoot! Unmerge! That's broken.

rgbkrk:
It's working locally for me here.

EDIT: Well, since I'm specifying a notebook directly.

jdetle:
In my case the raw args go from ['electron', '.' to [electron, '', '.'] and no file can be opened by launchFilename.

PR #555, opened 2016-07-24
Yargs

rgbkrk:
Switch to `yargs`.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -15,9 +15,7 @@ var argv = require('yargs')
               .parse(process.argv.slice(1));
 
 var notebooks = argv._;
-notebooks.filter(function(n){})
-console.log(argv)
-console.log(notebooks)
+
 app.on('window-all-closed', () => {
   // On OS X, we want to keep the app and menu bar active
   if (process.platform !== 'darwin') {
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -9,12 +9,15 @@ import { resolve } from 'path';
 
 const version = require('../../package.json').version;
 
-const argv = require('yargs')
+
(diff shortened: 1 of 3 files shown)


src/main/index.js
```diff
@@ -7,18 +7,13 @@ import { Menu, app } from 'electron';
...
+              .version(version)
+              .argv;
 
-program
-  .version(version)
-  .option('-k, --kernel [kernel]', 'kernel')
-  .parse(process.argv);
-
-const notebooks = program.args;
+const notebooks = argv._;
```
rgbkrk, line 17: Note that I tossed the kernel arg since we weren't even using it.

jdetle:
Linter is just wrong in the tests.

captainsafia:
The linter is still complaining about the `console.log` on line 65 in src/main/launch.js.

jdetle:
Yea I was trying to say that that's wrong. It prints out to the terminal just fine.

rgbkrk:
> Yea I was trying to say that that's wrong. It prints out to the terminal just fine.

It does indeed! However, our linting config explicitly forbids the use of `console.log`, causing Travis to return a big ugly ❌. The options to get around this include:
- using `console.warn`
- using `console.error`
- popping up a dialog box using native electron APIs
- using `eslint disable line`

However, all use of `console` should likely be forbidden in the main process as you can only see this when launching nteract from the command line (my comment will make more sense when you're using the built app).

jdetle:
^ Well I guess I'll listen to the linter next time :)

jdetle:
Should we get a third party involved or can I review?

rgbkrk:
:shipit:

Issue #559, opened 2016-07-25
Linux distribution has no image for nteract shortcut.


jdetle:
The title is self explanatory. Running `npm run dist:linux64` does not appropriately give the nteract shortcut its (very pretty) logo. I'm going to go ahead and try to fix this.

jdetle:
https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#icon

:( icon in the file manager is not yet supported.

rgbkrk:
At the very least, you can add the icon for our `new BrowserWindow` call following the [native image](http://electron.atom.io/docs/api/native-image/) doc. You'll want to case it off so that on windows it uses the `.ico` and on linux it uses `.png` (or so I suspect).

PR #560, opened 2016-07-25
Icon for linux, windows

jdetle:
Addresses #559 somewhat. 
@captainsafia or @rgbkrk should check this out and make sure it doesn't break mac stuff.


jdetle:
Totally misread the problem. Path to static is different if built then if ran using npm run start. I'll think about how to address both use cases.

src/main/launch.js
```diff
@@ -10,12 +10,20 @@ export function deferURL(event, url) {
   event.preventDefault();
   shell.openExternal(url);
 }
+let iconName = '';
+
+if (process.platform == 'win32') {
+  iconName = './static/icon.ico';
+} else if (process.platform == 'linux') {
+  iconName = './static/icon.png';
```
rgbkrk: This should be ../

jdetle: For whatever reason that did not actually work. Is `static/icon.png` better?

jdetle:
@captainsafia can you check this out and make sure it doesn't break the icons for mac machines?

rgbkrk:
I'm going to assume yes, which means we'll find out soon.

PR #562, opened 2016-07-28
Add electron-react-devtools

captainsafia:
This doesn't have to be merged but electron-react-devtools have been really useful to me when debugging and it's been a bit of a bother not having this committed across my different branches.

I think it's fair to include this in master since it's only on dev.


rgbkrk:
Yes please! :bow:

PR #567, opened 2016-07-30
Markdown migrate

rgbkrk:
Switch to the commonmark based markdown renderer, fix up the codemirror tests.


captainsafia:
Beautiful!

PR #569, opened 2016-07-30
upgrades with no breaking changes

rgbkrk:
Light upgrades to get us a little closer to the _one_ thing I really came to work on. :wink:


rgbkrk:
On a separate note, our lint runs shouldn't have to install a kernel (let alone zmq + everything else). ;)

PR #570, opened 2016-07-30
upgrade electron, force rebuild

rgbkrk:
Now on an excellent version of electron and helping all devs by forcing a rebuild.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "electron-mocha": "^2.2.1",
     "electron-osx-sign": "^0.3.1",
     "electron-packager": "^7.1.0",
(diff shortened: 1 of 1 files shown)


rgbkrk:
Interesting.

```
Error: Module version mismatch. Expected 48, got 49.
```

Considering the [node 6.x series](https://nodejs.org/en/download/releases/) has ABI 48 I don't see how electron now has 49 unless they're pulling from master or the future (7.x).

I'll get this working later on. Shelve this for now.

rgbkrk:
Sure enough, others are running into this: https://github.com/electron-userland/electron-prebuilt/issues/172

PR #571, opened 2016-07-31
Upgrade redux observable

rgbkrk:
On a separate note, I guess we should change our "agendas" to "epics".


rgbkrk:
Ruh roh. More work to do here.

screen shot 2016-07-30 at 5 58 33 pm

http://redux-observable.js.org/docs/FAQ.html#why-were-thunkservables-deprecated

rgbkrk:
Need to rework the epics next...

PR #574, opened 2016-08-01
Add modified indicator to titlebar

### LINKED ISSUE #563: Add saved/unsaved indicator
Would be great if there was a way to tell if a file is in an unsaved state as is done in the Jupyter notebook.

captainsafia:
This adds an asterisk to the titlebar if a file has been modified since it was last saved. 

Closes #563.

### DIFF
--- a/src/notebook/middlewares/index.js
+++ b/src/notebook/middlewares/index.js
@@ -1,5 +1,5 @@
 import * as constants from '../constants';
-import { setBackwardCheckpoint } from '../actions';
+import { setBackwardCheckpoint, setModified } from '../actions';
 import { List } from 'immutable';
 
 export const triggerUndo = store => next => action => {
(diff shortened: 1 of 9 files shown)


src/notebook/middlewares/index.js
```diff
@@ -17,3 +17,40 @@ export const triggerUndo = store => next => action => {
...
+export const triggerModified = store => next => action => {
+  const ignore = new List([
+    constants.CHANGE_FILENAME,
+    constants.NEW_KERNEL,
+    constants.KILL_KERNEL,
+    constants.INTERRUPT_KERNEL,
+    constants.SET_NOTEBOOK,
+    constants.UPDATE_CELL_EXECUTION_COUNT,
+    constants.CHANGE_OUTPUT_VISIBILITY,
+    constants.CHANGE_InPUT_VISIBILITY,
```
rgbkrk, line 30: This one seems a little bit off.

src/notebook/middlewares/index.js
```diff
@@ -17,3 +17,40 @@ export const triggerUndo = store => next => action => {
...
+
+  if (ignore.indexOf(action.type) === -1) {
+    store.dispatch(setModified(true));
+  }
+  if (action.type === constants.DONE_SAVING) {
+    store.dispatch(setModified(false));
+  }
+
+  return next(action);
+};
```
rgbkrk, line 56: One reduction we could do, if not too expensive, is to see if the notebook tree has changed.

rgbkrk, line 56: That is, derived instead of computed.

rgbkrk:
An alternative "derived" approach would be to treat the "is it modified" as a view that updates based on if state.notebook is the same as the previous state.notebook (kind of like a `shouldComponentUpdate`). Assuming what's in `state.notebook` is truly what goes to the document that is.

Issue #583, opened 2016-08-10
npm dependencies for eslint-config-airbnb & eslint-plugin-jsx-a11y cannot be met


ankostis:
From yesterday that I'm trying to install `package.json` I'm getting this (note that I'm not experienced with npm):

```
+-- UNMET PEER DEPENDENCY eslint@3.2.2
+-- eslint-config-airbnb@9.0.1
| `-- UNMET PEER DEPENDENCY eslint@^2.9.0
`-- UNMET PEER DEPENDENCY eslint-plugin-jsx-a11y@2.0.1

npm ERR! peer dep missing: eslint@^2.9.0, required by eslint-config-airbnb@9.0.1
npm ERR! peer dep missing: eslint-plugin-jsx-a11y@^1.2.0, required by eslint-config-airbnb@9.0.1
npm ERR! peer dep missing: eslint@^2.9.0, required by eslint-config-airbnb-base@3.0.1
```

I tried to I upgrade `eslint-config-airbnb` from `9.0.1 --> 10.0.0` and `eslint-plugin-react: 5.2.2 --> 6.0.0`, the installation passes. 
Would that work?

captainsafia:
Hello, @ankostis! Thanks for opening this issue. Let's see if we can figure out what's going on.

Let me try and reproduce the error on my machine and see if it occurs and if upgrading resolves the problem without causing any other issues.

To cover our bases, can you let me know what version of `npm` you are using and the commit hash of the version of master that you are on?

captainsafia:
OK! I can reproduce the error but it's simply a warning on my end. Does this issue halt the `npm install` or are you able to complete it?

ankostis:
No, it halts.
Note that I'm on Windows, and using `electron-prebuilt`, if that has any significance.

ankostis:
I managed to built latest 0.0.5 under Windows; now the the outdated versions are just warnings.
I made a quick test build and I confirm that it was the merging of ~~5c8fdec~~ #582 that fixed it.

Still you may consider to upgrade the versions, not to receive the warning.

PR #585, opened 2016-08-11
Some bumps

rgbkrk:
Bumped a few deps, closing #583.

This won't be ready right away since there are new linting rules to adhere to.


rgbkrk:
In a complete reversal, I've downgraded electron prebuilt at the advice of @mafintosh and @maxogden.

.eslintrc
```diff
@@ -9,7 +9,11 @@
   "rules": {
       "camelcase": [0],
       "no-console": [2, { allow: ["warn", "error"] }],
-      "no-underscore-dangle": [0]
+      "no-underscore-dangle": [0],
+      "react/no-string-refs": [0],
+      "react/jsx-filename-extension": [0],
+      "import/prefer-default-export": [0],
+      "react/no-find-dom-node": [0]
```
rgbkrk, line 16: We can dig into these ones later, since they impact the code itself.

.eslintrc
```diff
@@ -9,7 +9,11 @@
   "rules": {
       "camelcase": [0],
       "no-console": [2, { allow: ["warn", "error"] }],
-      "no-underscore-dangle": [0]
+      "no-underscore-dangle": [0],
+      "react/no-string-refs": [0],
```
rgbkrk, line 13: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-string-refs.md

.eslintrc
rgbkrk, line 16: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-find-dom-node.md

.eslintrc
```diff
@@ -9,7 +9,11 @@
...
+      "react/jsx-filename-extension": [0],
```
rgbkrk, line 14: @kenwheeler contests this one, will take airbnb to the ring to duke it out

.eslintrc
```diff
@@ -9,7 +9,11 @@
...
+      "import/prefer-default-export": [0],
```
rgbkrk, line 15: This one I adapted for one package that had a single export. The other will likely follow.

rgbkrk:
screen shot 2016-08-11 at 1 43 45 pm

Forgot to note why certain ones here won't be upgraded:
- We are pinned to commonmark 0.24.0 because of our commonmark React component
- We're using 1.1.3 of electron prebuilt until https://github.com/electron/electron/issues/5851 is resolved
- We need to upgrade to epics for redux-observable 0.9

The only one we can take action on right now is the epics and redux-observable.

rgbkrk:
The commonmark and epics issues mentioned are complete. Now just waiting on electron/electron#5851.

PR #587, opened 2016-08-11
Ignore some dependencies with greenkeeper

rgbkrk:
To let greenkeeper do its thing without massive breakage.


jdfreder:
Thanks Kyle!

Issue #564, opened 2016-07-30
Address weird markdown cell duplication bug


captainsafia:
There is a weird thing that happens where markdown cells are duplicated. This doesn't seem to be a React bug as the cells are duplicated in the actual `cellMap` in the state.

This doesn't happen with code cells as far as I can tell.

EDIT: Looks like this happens with code cells too.

rgbkrk:
Interesting. Any way of pulling this up?

captainsafia:
This should be fixed now that we've moved copy-paste logic into the Notebook component. This means that the cell copy-paste key will not be inadvertently triggered when the person is typing in the cell.

We can close this for now and re-open if the issue persists for whatever reason.

PR #592, opened 2016-08-13
Fix cell copy paste

captainsafia:
This fixes the bugs with cell copy paste. I believe this same issue was the cause of the cell duplication bug in #564 but I'm not sure so I won't close this yet.

Also, this moves the logic for copy-paste via keyboard to the Notebook component and applies it on the focused cell. I think we can add something else via the context menu that allows you to copy an unfocused cell.

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -13,8 +13,6 @@ import {
   focusCell,
   focusPreviousCell,
   focusNextCell,
-  copyCell,
-  pasteCell,
 } from '../../actions';
 
 export class Cell extends React.Component {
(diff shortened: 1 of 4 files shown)


rgbkrk:
There's a bit of linting touch up necessary here as well as a failing test from what looks to be related components:

```
  3 failing
  1) Cell handleKeyDown sets ctrlDown properly:
     AssertionError: expected undefined to be false
      at Context.it (cell-spec.js:47:5)
  2) Cell handleKeyUp responds properly to Ctrl + C:
     TypeError: Attempted to wrap undefined property copyCell as function
      at Object.wrapMethod (node_modules/sinon/lib/sinon/util/core.js:106:29)
      at Object.spy (node_modules/sinon/lib/sinon/spy.js:41:26)
      at Context.it (cell-spec.js:58:23)
  3) Cell handleKeyUp responds properly to Ctrl + V:
     TypeError: Attempted to wrap undefined property pasteCell as function
      at Object.wrapMethod (node_modules/sinon/lib/sinon/util/core.js:106:29)
      at Object.spy (node_modules/sinon/lib/sinon/spy.js:41:26)
      at Context.it (cell-spec.js:72:23)
```

rgbkrk:
Tests pass and coverage is down on the project. That's OK though, as we'll seek to make this better after (as discussed in #613).

PR #594, opened 2016-08-16
Green keeping.

rgbkrk:
I'm a little surprised that @greenkeeperio-bot hasn't handled these bumps.


ivanov:
I'm a registered member of the Green Party - does that count?

rgbkrk:
> I'm a registered member of the Green Party - does that count?

Your PRs for green keeping are given greater credence than the bot.

PR #598, opened 2016-08-16
require, then install

rgbkrk:
This likely shouldn't be a thing, but I got this to work by assigning the result of the require. .___O


jdfreder:
Errr, you change inject to install - maybe that's why it's fixed?

rgbkrk:
Derp. Looks like I fixed it by... fixing it.

Issue #579, opened 2016-08-04
Port from thunkservables to Epics


jayphelps:
redux-observable [switched to using Epics over thunkservables](http://redux-observable.js.org/docs/FAQ.html#why-were-thunkservables-deprecated) a little while back. We've since added [fairly decent docs](http://redux-observable.js.org/docs/basics/Epics.html) on these. Epics are a function that takes a stream of actions in and returns a stream of new actions out. The tl;dr is that instead of dispatching a thunk that returns an observable, you dispatch regular POJO actions and listen for them in process managers called Epics which can then do your side effects in response and then they emit other actions when they're done with any response or signal.

I briefly looked through the nteract source and found two thunkservables as examples, [save/saveAs](https://github.com/nteract/nteract/blob/master/src/notebook/actions/index.js#L57).

They would be ported over to something _like_ this (untested...sorry!):

``` js
// Extracted out for readability, but you don't have to or you can use
// something like the rxjs-fs package
const writeFileObservable = (filename, data, ...args) =>
  Observable.create(observer => {
    fs.writeFile(filename, data, ...args), error => {
      if (error) {
        observer.error(error);
      } else {
        observer.next({ filename, data });
        observer.complete();
      }
    });
  });

const SAVE = 'SAVE';
const SAVE_AS = 'SAVE_AS';

const changeFilename = filename => ({ type: CHANGE_FILENAME, filename });
const save = (filename, notebook) => ({ type: SAVE: filename, notebook });
const saveAs = (filename, notebook) => ({ type: SAVE_AS: filename, notebook });
const doneSaving = () => ({ type: DONE_SAVING }); 

const saveEpic = actions =>
  actions.ofType(SAVE)
    .do(action => {
      // If there isn't a filename, save-as it instead
      if (!action.filename) {
        throw new Error('save needs a filename');
      }
    })
    .mergeMap(action =>
      writeFileObservable(filename, JSON.stringify(commutable.toJS(action.notebook), null, 1))
        .map(doneSaving)
        // .startWith({ type: START_SAVING })
        // since SAVE effectively acts as the same as START_SAVING
        // you could just look for that in your reducers instead of START_SAVING
    );

const saveAsEpic = actions =>
  actions.ofType(SAVE_AS)
    .mergeMap(action => [
      changeFilename(action.filename),
      save(action.filename, action.notebook)
    ]);
```

[...]

PR #605, opened 2016-08-18
Kernel launch conversion

rgbkrk:
kernelLaunch is now an epic in support of #579.

It's not nearly as neat as I'd like it to be. The names of the functions and actions need some fixing to better reflect the lifecycle. I wanted it working and proper first. Later the flow of actions should be:
- `LAUNCH_KERNEL`
- `KERNEL_LAUNCHED`
- `ACQUIRE_KERNEL_INFO`
- `SET_LANGUAGE_INFO`
- `KERNEL_READY`


rgbkrk:
I figured this would drop coverage. If I test 14 lines I can make this pass. Sure why not, I'll go for it.

PR #619, opened 2016-08-21
Expand comments for the cellular epic

rgbkrk:
Added some more comments while talking to @willingc and @ssanderson earlier tonight.


rgbkrk:
The coverage drop is insignificant and a little bit expected.

PR #621, opened 2016-08-21
Filtered for cell type equal to code

### LINKED ISSUE #614: Run all executes Markdown cells
It's very tempting for me to fix this while I'm doing the execution revamp. I'll just post it here instead.

screen shot 2016-08-20 at 8 30 24 am

Need to filter on `code.cell_type === 'code'` (using Immutable.js mechanics).

jdetle:
Closes #614!


rgbkrk:
So awesome. Thank you.

PR #627, opened 2016-08-22
Use redux-undo for undo/redo

captainsafia:
Let's take advantage of the [redux-undo](https://github.com/omnidan/redux-undo) library to manage our undo/redo and avoid nasty bugs in our codebase.

This makes the document reducer tests all wonky so I still have to fix those.


captainsafia:
Ready for merge!

Awkward GIF to show how it works.

undo-redo

rgbkrk:
Slickkkkkkkkk

rgbkrk:
Amusingly, this un-does outputs and numbering which should likely be atomic.

rgbkrk:
Another quirk: because of cell focus, this will only undo until it focuses a codemirror cell which then switches to the undo/redo of the codemirror cell itself.

captainsafia:
> Amusingly, this un-does outputs and numbering which should likely be atomic.

Yeah, I was debating whether or not to add those actions to the exclude list.

> Another quirk: because of cell focus, this will only undo until it focuses a codemirror cell which then switches to the undo/redo of the codemirror cell itself.

Yeah. Cell focus does make this funky. Not sure if a fix is possible.

rgbkrk:
> Yeah, I was debating whether or not to add those actions to the exclude list.

At the very least, since I know how the app works it's kind of hilarious to play with code like:

``` python
import time
for ii in range(20):
  print(ii)
  time.sleep(1)
```

PR #631, opened 2016-08-24
Migrate tf react

rgbkrk:
This brings transformime-react into nteract/nteract.

I went ahead and brought all the commits over as is, perhaps I should nuke the tags or edit them in a rebase...


rgbkrk:
Well this rebase sure got ugly.

rgbkrk:
If I don't rebase, it works out just fine. If I do rebase, even with a reword, git has a bit of a fit. Worth noting that to do this merger I had to use `--allow-unrelated-histories` with git. Messy.

rgbkrk:
Not going to do this directly.

PR #636, opened 2016-08-26
Audit use of state

captainsafia:
Cleaning up how we use `state` across the application.

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -47,7 +47,7 @@ export function triggerSaveAs(store, dispatch) {
         return;
       }
       const state = store.getState();
-      const { executionState } = state.app;
+      const executionState = state.app.get('executionState');
       const notebook = state.document.present.get('notebook');
       dispatch(saveAs(filename, notebook));
       BrowserWindow.getFocusedWindow().setTitle(`${tildify(filename)} - ${executionState}`);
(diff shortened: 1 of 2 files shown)


src/notebook/menu.js
```diff
@@ -47,7 +47,7 @@ export function triggerSaveAs(store, dispatch) {
         return;
       }
       const state = store.getState();
-      const { executionState } = state.app;
+      const executionState = state.app.get('executionState');
```
rgbkrk, line 50: These ones are ok since it's a `Record` (or so I think...).

captainsafia: Yeah, I figured since we consistently use `get` everywhere else in our program we should do so here. More a stylistic fix than anything.

src/notebook/menu.js
```diff
@@ -152,7 +153,7 @@ export function dispatchKillKernel(store, dispatch) {
 
 export function dispatchInterruptKernel(store, dispatch) {
   const state = store.getState();
-  const { notificationSystem } = state.app;
+  const notificationSystem  = state.app.get('notificationSystem');
```
rgbkrk: This is the line the linter is warning on (two spaces then an `=`).

captainsafia: Yep, just pushed a fix.

I still haven't added that pre-commit lint hook. 😝

Issue #639, opened 2016-08-27
new untitled notebooks get a kernel but do not record kernelspec


minrk:
The empty untitled notebook you get when you launch nteract gets hooked up to a default kernel, but that information doesn't get recorded in notebook.metadata.kernelspec.

minrk:
New notebooks with a specific kernel from the menu get the right info.

rgbkrk:
Done in #649.

rgbkrk:
Thanks for alerting on this one Min, this also addresses the issue with new kernels launched from the "language" menu.

PR #640, opened 2016-08-27
include kernelspec displayName in title

minrk:
Since I like to know which kernel I am using:

screen shot 2016-08-27 at 18 26 45

I also ran into #639 putting this together.


rgbkrk:
Addressed the linting in 3a52d60aab74418c0a8e872748b6b39aaf6b7ccd.

rgbkrk:
Thanks Min!

PR #649, opened 2016-08-27
set the kernelspec metadata on NEW_KERNEL

rgbkrk:
When a new kernel comes in, we make sure to set the right metadata on the document.


captainsafia:
Can you add some tests?

rgbkrk:
This makes me realize we were operating on the wrong version of the format I think. The top level structure for this metadata should be:

``` js
{
  "metadata" : {
    "kernel_info": {
        # if kernel_info is defined, its name field is required.
        "name" : "the name of the kernel"
    },
    "language_info": {
        # if language_info is defined, its name field is required.
        "name" : "the programming language of the kernel",
        "version": "the version of the language",
        "codemirror_mode": "The name of the codemirror mode to use [optional]"
    }
  },
  "nbformat": 4,
  "nbformat_minor": 0,
  "cells" : [
      # list of cell dictionaries, see below
  ],
}
```

rgbkrk:
> Can you add some tests?

Can do, captain.

rgbkrk:
Someday later I'll want to check out why we were setting (and reading) `kernelspec` in the metadata... That wasn't an old notebook format thing was it?

rgbkrk:
Closes #639.

PR #652, opened 2016-08-28
Remap cell copy-paste keyboard shortcuts

captainsafia:
Remapping the keys to see if this will work.

However, this results in the Electron-level copy paste not working. I'm working on that now.

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -143,9 +143,9 @@ class Notebook extends React.Component {
   keyDown(e) {
     if (e.keyCode !== 13) {
       const cmdOrCtrl = e.ctrlKey || e.metaKey;
-      if (cmdOrCtrl && e.keyCode === 67) {
+      if (cmdOrCtrl && e.shiftKey && e.keyCode === 67) {
         this.copyCell();
-      } else if (cmdOrCtrl && e.keyCode === 86) {
+      } else if (cmdOrCtrl && e.shiftKey && e.keyCode === 86) {
         this.pasteCell();
       }
       return;


captainsafia:
Ugh. You know what. @rgbkrk go ahead and review and merge this.

It appears that the issue is with the `role` feature in Electron. I'm not sure if this is due to some electron upgrade. I don't think we've made any recently.

Essentially, the role feature that sends special actions for system functions like copy and paste is not working regardless of where the role is set (in an application or context menu).

I'll work on the role issue in another PR.

rgbkrk:
> It appears that the issue is with the role feature in Electron. I'm not sure if this is due to some electron upgrade. I don't think we've made any recently.

We did revert to an old version due to issues with native modules and the more recent versions of electron.

src/notebook/components/notebook.js
```diff
@@ -143,9 +143,9 @@ class Notebook extends React.Component {
   keyDown(e) {
     if (e.keyCode !== 13) {
       const cmdOrCtrl = e.ctrlKey || e.metaKey;
-      if (cmdOrCtrl && e.keyCode === 67) {
+      if (cmdOrCtrl && e.shiftKey && e.keyCode === 67) {
         this.copyCell();
-      } else if (cmdOrCtrl && e.keyCode === 86) {
+      } else if (cmdOrCtrl && e.shiftKey && e.keyCode === 86) {
         this.pasteCell();
```
rgbkrk, line 149: Do we need to `e.preventDefault()` here so that copy and paste doesn't do the normal thing too?

captainsafia: Good point. I'll add it in the other PR that fixes text copy and paste.

PR #653, opened 2016-08-29
Make toolbar fixed

jdetle:
## Changes made
- Changed css such that a slightly transparent toolbar renders inside the cell
- The toolbar should render when the cell is in focus only, but this currently only works for mousevents. I need help to actually just tie this to the cell being focused. 
## Todo
- [x] Cleanup hover logic
- [x] Fix toolbar rendering logic
- [ ] Add button for dropdown menu
- [ ] Update cell tests in the style of the markdown spec

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -98,7 +98,7 @@ export class Cell extends React.Component {
         onContextMenu={this.contextMenu}
       >
         {
-          this.state.hoverCell ? <Toolbar
+          this.state.hoverCell || focused ? <Toolbar
             type={type}
             cell={cell}
             id={this.props.id}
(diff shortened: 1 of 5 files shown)


jdetle:
@captainsafia @rgbkrk, do you think I can go ahead and delete the setHoverState prop of Toolbar since this PR just renders it if the cell is focused or the cell itself is hovered over?

rgbkrk:
That seems fine to me. I'd like to rely on less local state and have less reliance on hover in general.

jdetle:
toolbar_2 2

jdetle:
toolbar_2 3

jdetle:
@captainsafia and I voted to just merge this and work on dropdown in the next few days

rgbkrk:
Woo!

PR #657, opened 2016-08-30
Introduce an error logger.

rgbkrk:
Is there a better way I should be handling this @captainsafia?


captainsafia:
Any reason you aren't using `redux-logger`?

rgbkrk:
Nope, let's use it!

PR #662, opened 2016-08-31
Plotly.js renderer

rgbkrk:
This introduces a Plotly transform for the `application/json+plotly.v1` mimetype.

screen shot 2016-08-30 at 7 19 33 pm

/cc @jackparmer @chriddyp @captainsafia


jdfreder:
slick

src/notebook/components/transforms/plotly.js
```diff
@@ -0,0 +1,30 @@
...
+    Plotly.newPlot(this.el, payload.data, payload.layout);
+  }
+
+  shouldComponentUpdate() {
+    return false;
+  }
+
+  render() {
+    return (
+      <div ref={(el) => this.el = el} /> // eslint-disable-line
```
chriddyp: I'm not sure how the output cells work w.r.t. sizing, but we might need to set `height` here if `layout.height && !layout.autosize`. Here's an example figure:

```
data = [{x: [1, 2], y: [3, 1]}]
layout = {height: 1600}
```

rgbkrk: The cell allows the output to be any size it wants to be. We can set it explicitly here like you said.

``` js
render() {
  const { data, layout } = this.props.data.toJS();
  const style = {}
  if(layout.height && !layout.autosize) {
    style.height = layout.height;
  }
  return (
    <div style={style} ref={(el) => this.el = el} /> // eslint-disable-line
  )
}
```

captainsafia:
Merged! 🎉 

Slight coverage drop but we can address that in subsequent PRs.

PR #667, opened 2016-09-02
React simple dropdown

jdetle:
Here it is! 

A thousand thanks to @captainsafia 

iconmenusimple_2
## Changes Made

-Dropdown CSS
-Toolbar only now shown when focused
-Dropdown button and menu added


rgbkrk:
:boom: This is beautiful! 👍

PR #674, opened 2016-09-03
[Don't merge] Create a backwards compatible plotly mimetype.

rgbkrk:
This allows the plotly mimetype to come in as a string per the current notebook validation.


rgbkrk:
Only opening this branch to store off for reference. I'll close it after Travis finishes.

PR #676, opened 2016-09-04
Add cell cut

captainsafia:
It's time for Safia to get out of her coding funk.

I heard you like to cut cells too.

cut-paste

Also, I should probably change paste so that it pastes after the focused cell.

### DIFF
--- a/src/notebook/actions.js
+++ b/src/notebook/actions.js
@@ -231,6 +231,13 @@ export function copyCell(id) {
   };
 }
 
+export function cutCell(id) {
+  return {
+    type: constants.CUT_CELL,
+    id,
+  };
+}
+
 export function pasteCell() {
   return {
     type: constants.PASTE_CELL,
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -19,6 +19,7 @@ import {
   focusNextCell,
   moveCell,
   copyCell,
+  cutCell,
   pasteCell,
 } from '../actions';
 import { executeCell } from '../epics/execute';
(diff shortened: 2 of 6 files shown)


src/notebook/actions.js
```diff
@@ -231,6 +231,13 @@ export function copyCell(id) {
   };
 }
 
+export function cutCell(id) {
+  return {
+    type: constants.CUT_CELL,
+    id,
+  };
+}
```
rgbkrk, line 239: Testing the action creator will increase coverage in a pretty simple way here (pretty much a copy paste of other action creator tests).

captainsafia: Ah. Whoops. Forgot about action creators tests.

Coming right up!

src/notebook/components/notebook.js
```diff
@@ -148,6 +154,8 @@ class Notebook extends React.Component {
       const cmdOrCtrl = e.ctrlKey || e.metaKey;
       if (cmdOrCtrl && e.shiftKey && e.keyCode === 67) {
         this.copyCell();
+      } else if (cmdOrCtrl && e.shiftKey && e.keyCode === 88) {
+        this.cutCell();
```
rgbkrk, line 158: If you use a similar test to the Markdown cell key handling, you can check to make sure that this function gets called.

src/notebook/components/notebook.js
```diff
@@ -138,6 +140,10 @@ class Notebook extends React.Component {
     this.props.dispatch(copyCell(this.props.focusedCell));
   }
 
+  cutCell() {
+    this.props.dispatch(cutCell(this.props.focusedCell));
```
rgbkrk, line 144: This can be mocked with a fake dispatch to ensure it generates the right action.

rgbkrk:
Added some notes about how to test the other new bits being introduced here.

captainsafia:
Updated. I've got another PR going with some component tests so I'll address the component-related issues there.

rgbkrk:
Woot!

PR #679, opened 2016-09-05
Unit testing Notebook component

rgbkrk:
This sets up a few things to make testing the `Notebook` component much easier:
- Test section for a Drag and Drop Notebook, to test that drag and drop works
- Notebook without Drag and Drop

While doing this I also noticed a bit of an anti-pattern in `SET_NOTEBOOK` for when it sets up the cellStatuses. It boils down to doing a `reduce` to create a map from the `cellOrder`.


jdfreder:
:+1:

PR #680, opened 2016-09-05
Theming epic

rgbkrk:
This moves the theming out of the `setState` and top level mechanics from before and into the main epic setup. This was _mostly_ put into nice async and state flow because the top level setState was colliding with the `Provider` component for redux. Nice and clean now!

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
     "eslint-plugin-react": "^6.0.0",
     "jsdom": "^9.4.2",
     "mocha": "^3.0.2",
+    "mock-require": "^1.3.0",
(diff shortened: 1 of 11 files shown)


jdetle:
Is this done?

rgbkrk:
Not quite, I ended up resorting to switching to electron-mocha which breaks coverage reporting.

rgbkrk:
Now it's ready @jdetle!

betatim:
@rgbkrk made me do it 😬

rgbkrk:
At least only one chainsaw is on fire.

PR #682, opened 2016-09-05
Update styles for dropdown

jdetle:
This should cause everything to look nice for the CSS themes.
css_fixes

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -424,10 +424,9 @@ li.CodeMirror-hint-active {
     opacity:  1.0;
     position: absolute;
     left: 8px;
-    border-color: gray;
+    border-color: rgba(122,122,122,0.3);
     border-width: 1px;
     border-style: solid;
-    background-color: #FFF;
     padding: 10px;
     padding-top: 0px;
     padding-right: 0px;
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -128,8 +128,7 @@ td {
 .cell-toolbar
 {
     display: inline-box;
-    background: var(--main-bg-color);
-    background-color: rgba(255, 255, 255, 0.3);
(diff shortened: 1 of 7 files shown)


rgbkrk:
I wonder when/how the inputs broke for classic mode.

static/styles/main.css
```diff
@@ -435,7 +433,7 @@ li.CodeMirror-hint-active {
     padding-right: 0px;
     font-size: 11px;
     margin: 20px 0;
-    background-color: rgba(255,255,255,1);
+    background-color: var(--dropdown-content);
```
rgbkrk, line 435: 👍

static/styles/theme-christmas.css
```diff
@@ -12,6 +12,9 @@
   --toolbar-button: #aaa;
   --toolbar-button-hover: #555;
 
+  --dropdown-content: white;
+  --dropdown-content-hover: grey;
```
rgbkrk: Heh.

jdetle:
¯_(ツ)_/¯

static/styles/theme-christmas.css
jdetle: do you prefer "gray"?

jdetle: its a little heavy actually

rgbkrk: Nah, I'm giggling because you had to update our hidden themes. :)

captainsafia: I think the gray we originally used for this was #ecf0f1.

rgbkrk:
I seriously have no idea why Travis is not behaving well today. Since these are just CSS changes that don't affect our CSS structure (which is specced), I'm going to go ahead and merge.

PR #683, opened 2016-09-05
Move cell copy/paste to Electron menu

captainsafia:
Looks neater this way I think.

PR #684, opened 2016-09-05
New Code Cell Shortcut

jdetle:
Creating new shortcuts is easy thanks to #683

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -123,6 +123,11 @@ export const edit = {
       type: 'separator',
     },
     {
+      label: 'New Code Cell',
+      accelerator: 'CmdOrCtrl+Shift+N',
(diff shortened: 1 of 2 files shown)


rgbkrk:
You got that right!

Considering we have to support the shortcuts we choose in the long run, what's the reasoning for this key set?

jdetle:
I think `command+shift+char` should imply that you are doing something to the cell, while `command+char` should be reserved for contents of the cell.

rgbkrk:
It just now occurred to me this is the same keybinding as incognito mode in Chrome.

jdetle:
That decision was entirely subconscious

rgbkrk:
I'm calling cmd-shift-p for default publish, leaving normal cmd-p for printing.

PR #685, opened 2016-09-05
Test getLanguageMode, trim out-of-date test

rgbkrk:
Since I still want an example test of how to test the notebook instance, instead of just deleting the out of date test I introduced a new test.


jdetle:
💣 .com

PR #690, opened 2016-09-06
Set the icon path for Leaflet

rgbkrk:
Looks like we definitely needed this! Now I can plot coffee shops.

screen shot 2016-09-05 at 8 32 47 pm


jdetle:
Awesome!

PR #693, opened 2016-09-06
Pandas to geojson

rgbkrk:
Including a sample notebook from @gboeing's Urban Data Science class: https://github.com/gboeing/urban-data-science

nteract-geo

Waiting on https://github.com/gboeing/urban-data-science/issues/1 to determine if we can include this.

This notebook is also a good demonstration of what we need to fix for scrolling as well as for larger notebooks.


rgbkrk:
Included the license in the notebook. This is good to go.

PR #694, opened 2016-09-07
Stop using componentWillMount

rgbkrk:
Per https://github.com/facebook/react/issues/7671

Switched over to `componentDidMount`, which is totally :cool: for our uses here.

### DIFF
--- a/src/notebook/components/cell/cell-creator.js
+++ b/src/notebook/components/cell/cell-creator.js
@@ -20,7 +20,7 @@ export default class CellCreator extends React.Component {
     show: false,
   };
 
-  componentWillMount() {
+  componentDidMount() {
     // Listen to the page level mouse move event and manually check for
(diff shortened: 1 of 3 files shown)


quantizor:
If you want to avoid the extra unnecessary render, you can extend `constructor` instead and perform your logic / state setting there.

rgbkrk:
> If you want to avoid the extra unnecessary render, you can extend constructor instead and perform your logic / state setting there.

Interesting. When does the extra render happen here?

quantizor:
If you perform a setState in cDM.

rgbkrk:
Ah yeah, we're not doing any `setState` in cDM here (it was in cWM before). Thanks for the tip!

quantizor:
Though it doesn't look like that's happening in the diff, so probably fine 😊

PR #705, opened 2016-09-09
sanity: keep index.js around, which loads build

rgbkrk:
Apparently you can't run a dist copy without ./index.js, for electron to run as `electron .`. This reverts #703.


jdetle:
:octocat:

PR #714, opened 2016-09-10
Focus on reducer tests

captainsafia:
Don't merge yet. I have more commits to push.


captainsafia:
OK. This should be good to merge. The only untested line of code is in `moveCell` and I wasn't entirely sure what the function was trying to do so I opted not to do it. For some strange reason, the case where the current ID and the destination ID match is covered but the situation where they are not is not?!!?

rgbkrk:
Awesome! I'll have to check out the move cell tests soon.

PR #716, opened 2016-09-10
Add initial tests for native window

captainsafia:
Starting to touch on uncovered parts of codebase.

### DIFF
--- a/test/renderer/native_window_spec.js
+++ b/test/renderer/native_window_spec.js
@@ -1,0 +1,26 @@
+import * as nativeWindow from '../../src/notebook/native-window';
+
+import chai, { expect } from 'chai';
+import sinon from 'sinon';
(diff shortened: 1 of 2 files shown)


captainsafia:
OCD Safia here — should our spec files be named using '-' instead of '_' to match how it is done in `src/`.

rgbkrk:
@captainsafia your choice, we'll just make it consistent. I prefer the kabob case on files. The `_spec` you see in the wild is "mostly historical", coming from RSpec.

captainsafia:
> @captainsafia your choice, we'll just make it consistent. I prefer the kabob case on files. The _spec you see in the wild is "mostly historical", coming from RSpec.

OK. I prefer kabob case as well (more delicious anyways), I'll submit another PR moving to kabob case on all the test files.

PR #719, opened 2016-09-11
Switch loading to a frontend epic

rgbkrk:
This moves loading into its rightful place in the frontend app.
- [x] Handle loading in an epic fashion
- [x] Take all commutable ops out of the main process
- [x] Make loading be an action triggered via `main:load` or similar (instead of any processing on main thread)
- [x] Make kernel launch actions follow loading
- [x] Clean up `native-window` to rely on functions from the main thread sourcery
- [x] Filename should be registered when notebook loaded
- [x] Handle new notebook case
- [x] Tests for all the new functions!
- [x] Delete tests for the old way!
- [x] Test some new bits that are easier to cover now!

After sanity restored (I've shuffled a lot here):
- [ ] Bring back `duplicateNotebook`


rgbkrk:
Ok, I've cleaned up a lot of this. Need to update my todo list above.

captainsafia:
Oh wow! This is a fantastic refactor! 💯

rgbkrk:
Just wait until you see how fast the app boots now!

captainsafia:
> Just wait until you see how fast the app boots now!

rgbkrk:
Going to let CodeCov finish the next cycle and then see if there are other tests I can instrument.

rgbkrk:
Once codecov updates again, this is ready to go. I need to do some thinking on how to do duplicate notebook.

PR #723, opened 2016-09-11
Allow the main entry point to be run in tests.

rgbkrk:
This will be a tiny bit crazy.

I'm merely `requir(e)`ing `src/notebook/index.js` to force the entire entry point to run (now that it's no longer wrapped in an `ipc` main call.

I'll be interested to see what CodeCov thinks of this.


rgbkrk:
Heh, not much. Clearly not really testing anything other than "can it be run?".

PR #724, opened 2016-09-11
Move coverage path for IDE tools to pick up

rgbkrk:
Makes it handy to use [lcov-info with Atom](https://atom.io/packages/lcov-info). Coverage reporting to codecov should still work, based on their script.


rgbkrk:
screen shot 2016-09-11 at 3 29 01 pm

rgbkrk:
Hopefully this didn't break codecov... Please report back codecov, I miss you.

PR #732, opened 2016-09-12
Add initial comm test suite

captainsafia:
Now everything in `src/notebook` is analyzed by codecov so we have a more accurate coverage value.


jdetle:
👍

rgbkrk:
Amusingly, we don't actually use this code anywhere - it's part of the setup for the widgets.

PR #738, opened 2016-09-12
More menu spec tests

captainsafia:
Don't merge yet. I have more stuff to add after lunch.

🍚

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -65,19 +65,20 @@ export function showSaveAsDialog(defaultPath) {
   });
 }
 
+export function triggerWindowRefresh(store, filename) {
+  if (!filename) {
+    return;
+  }
+  const state = store.getState();
+  const executionState = state.app.get('executionState');
+  const notebook = state.document.get('notebook');
+  store.dispatch(saveAs(filename, notebook));
+  BrowserWindow.getFocusedWindow().setTitle(`${tildify(filename)} - ${executionState}`);
+}
+
 export function triggerSaveAs(store) {
   showSaveAsDialog()
-    .then(filename => {
-      if (!filename) {
-        return;
-      }
-      const state = store.getState();
-      const executionState = state.app.get('executionState');
-      const notebook = state.document.get('notebook');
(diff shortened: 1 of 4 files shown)


captainsafia:
OK. For some strange reason, spies on `triggerWindowRefresh` and `triggerSaveAs` are not working.

I'm instantiating the spy as such.

```
const triggerSaveAs = sinon.spy(menu, 'triggerSaveAs');
```

The test suite claims the functions were never called but that cannot be the case because `console.log`s that I place inside `triggerSaveAs` are printed while the test suite runs.

Anyway, this helps us move along so I guess it can be merged.

src/notebook/menu.js
```diff
@@ -65,19 +65,20 @@ export function showSaveAsDialog(defaultPath) {
...
+export function triggerWindowRefresh(store, filename) {
+  if (!filename) {
+    return;
+  }
+  const state = store.getState();
+  const executionState = state.app.get('executionState');
+  const notebook = state.document.get('notebook');
+  store.dispatch(saveAs(filename, notebook));
+  BrowserWindow.getFocusedWindow().setTitle(`${tildify(filename)} - ${executionState}`);
+}
```
rgbkrk, line 77: 👍

rgbkrk, line 77: This could probably be decoupled a bit more - this is doing two things:
- Setting the title based on filename and execution state
- Dispatching a saveAs

Then you can test those side effects individually

test/renderer/menu-spec.js
```diff
@@ -277,7 +279,35 @@ describe('menu', () => {
...
+
+      menu.triggerWindowRefresh(store, filename);
+
+      expect(store.dispatch.firstCall).to.be.calledWith({
+        type: 'SAVE_AS',
+        notebook: store.getState().document.get('notebook'),
+        filename: filename,
+      });
+    });
+  });
```
rgbkrk, line 312: Where's the `setTitle` check to make sure the title was changed?

rgbkrk:
📔

PR #740, opened 2016-09-13
Live rendered markdown

jdetle:
Check it!

markdown

The live-ness of it has its perks and its downside, we should probably make this a toggleable feature?

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -108,6 +108,7 @@ export class Cell extends React.Component {
             focusBelow={this.focusBelowCell}
             focused={focused}
             cell={cell}
+            cellStatus={this.props.cellStatus}
             id={this.props.id}
             theme={this.props.theme}
           /> :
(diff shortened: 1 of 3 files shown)


jdetle:
markdown_2

^ what happens upon executing the cell.

jdetle:
togglability

rgbkrk:
Awesome!

As for the latex rendering flicker, we can probably make that a lot cleaner by actually coordinating with MathJax. Our LaTeX render is a bit greedy right now.

rgbkrk:
I'm not a fan of using the input/output semantics here, hiding the "input" on markdown seems weird - it _is_ the text after all. We default to render when not in render mode after.

jdetle:
nbd I just increased coverage by 2% 0.o

rgbkrk:
> nbd I just increased coverage by 2% 0.o

and then... not?

jdetle:
I rebased?

rgbkrk:
Codecov so silly.

You have a wonderful day too John!

PR #743, opened 2016-09-13
Add tests to messaging-spec

captainsafia:
Back to testing.

I really need to get a coverage extension for Vim...

### DIFF
--- a/test/renderer/api/messaging-spec.js
+++ b/test/renderer/api/messaging-spec.js
@@ -23,19 +23,20 @@ describe('childOf', () => {
       expect(val).to.equal(3);
     });
   });
-  it('throws an error if msg_id is not present', () => {
+  it('throws an error if msg_id is not present', (done) => {
     return Rx.Observable.from([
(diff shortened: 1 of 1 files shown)


test/renderer/api/messaging-spec.js
```diff
@@ -41,4 +60,40 @@ describe('ofMessageType', () => {
...
     });
   });
+  it('throws an error in msg_type is not present', () => {
+    return Rx.Observable.from([
+      {header: {msg_type_invalid: 'a'}},
+      {header: {msg_type_invalid: 'd'}},
+      {header: {}},
+      {header: {msg_type: 'a'}},
+    ])
+    .ofMessageType(['a', 'd'])
```
rgbkrk, line 71: Either use the `.catch` operator here or use the error callback in `subscribe` (it's the second one).

rgbkrk, line 71: You'll need to use the `done` parameter in the callback to `it` above to take the Observable approach.

PR #744, opened 2016-09-13
Handle bunk mimetypes in pager, test it

rgbkrk:
We were passing in way more than necessary for the pager here. This scopes it a little nicer. Additionally, mimetypes that aren't found return `null` components now and everything is tested!

### DIFF
--- a/src/notebook/components/cell/pager.js
+++ b/src/notebook/components/cell/pager.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import PureRenderMixin from 'react-addons-pure-render-mixin';
 
 import Immutable from 'immutable';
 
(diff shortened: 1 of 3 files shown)


src/notebook/components/cell/pager.js
```diff
@@ -9,17 +8,22 @@ class Pager extends React.Component {
...
+    return false;
   }
 
   render() {
-    const bundle = this.props.pager.get('data');
+    const bundle = this.props.data;
     const mimetype = richestMimetype(bundle, this.props.displayOrder, this.props.transforms);
+    if (!mimetype) {
+      console.error('the fuck');
+      return null;
```
captainsafia, line 23: I think you meant to delete this `console.error`.

test/renderer/components/cell/pager-spec.js
```diff
@@ -0,0 +1,41 @@
...
+      <Pager
+        displayOrder={displayOrder}
+        transforms={transforms}
+        data={Immutable.fromJS({"application/ipynb+json": "{}"})}
+      />
+    );
+
+    expect(pager.type()).to.be.null;
+  })
+})
```
captainsafia, line 41: Much nicer than the tests I had.

src/notebook/components/cell/pager.js
rgbkrk:
```
                                     >_____<
```

test/renderer/components/cell/pager-spec.js
rgbkrk: Thanks! Funny we both decided to test this one right now.

src/notebook/components/cell/pager.js
jdetle: haha!

PR #746, opened 2016-09-14
Rely on Rx for async completion flow

rgbkrk:
Completion only gets used in the editor so I reduced the boilerplate down a bit into one function on the component. We can extract this after.

The flow is way more direct, no more back and forth with promises and callbacks. Also, I cleaned up the use of `debounce` in favor of `switchMap` - user gets exactly their intent and code state.

While I was at this, I noticed that eslint-config-airbnb (or some downstream dep) introduced a new rule that I'll have to clean up on another PR. Dropping it via our config now.


rgbkrk:
Amusingly, I'll bet this increases code coverage. I'll be writing some tests now.

jdetle:
Merging despite the churlishness of codecov. This is really cool!

rgbkrk:
whoa whoa, I'm deep in tests, it's cool

PR #748, opened 2016-09-14
Test completion logic

rgbkrk:
Observable test!

This... seems like a much better way to handle testing other features that use Observables. 💭


jdetle:
#killinit

PR #752, opened 2016-09-14
Fix multiline completion

rgbkrk:
Apparently this was not getting the full cursor position for Jupyter (by index, including newlines), which would affect cells with more than one line. Now it is and using CodeMirror's `indexFromPos`.


rgbkrk:
Ok, _there_ we go.

PR #759, opened 2016-09-14
Show warning message if no kernels

### LINKED ISSUE #490: Provide error messages when no kernelspecs detected
When I run `npm run start` the app get's stuck on the "Loading..." screen below. Also, under the `File > New` menu, I only see `Bash` as an option and no `Python`. 

I don't see any error message in the terminal window. Any ideas on what I am doing wrong?

image

captainsafia:
Closes #490.

Shows a warning when no kernels are installed.

screen shot 2016-09-14 at 12 04 19 pm

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -357,16 +357,19 @@ export function loadFullMenu() {
       };
     }
 
+    let languageMenu = {};
+    let newNotebookItems = {};
+
     if (Object.keys(kernelSpecs).length !== 0) {
       const kernelMenuItems = Object.keys(kernelSpecs).map(generateSubMenu);
 
-      const newNotebookItems = Object.keys(kernelSpecs)
+      newNotebookItems = Object.keys(kernelSpecs)
         .map(kernelName => ({
           label: kernelSpecs[kernelName].spec.display_name,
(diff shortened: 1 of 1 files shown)


src/main/menu.js
```diff
@@ -357,40 +357,55 @@ export function loadFullMenu() {
...
+        ],
+      };
+    } else {
+      dialog.showMessageBox({
+        type: 'warning',
+        title: 'No Kernels Installed',
+        buttons: [],
+        message: 'No kernels are installed on your system.',
+        detail: 'No kernels are installed on your system so you will not be' +
+                'able to execute code cells in any language.',
```
jdetle: should be a space  after be?

captainsafia: Good catch!

src/main/menu.js
```diff
@@ -357,40 +357,55 @@ export function loadFullMenu() {
...
+      }, (index) => {
+        const languageMenu = {
+          label: '&Language',
+        };
+      });
```
rgbkrk, line 414: Excellent!

We should probably link to a help page for this too. They need a next step so we don't leave them lost.

rgbkrk (approved):

PR #761, opened 2016-09-14
Warn and quit if no kernels

captainsafia:
Doing some work on our no kernels cause.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1,4 +1,4 @@
-import { Menu, app, ipcMain as ipc } from 'electron';
+import { Menu, dialog, app, ipcMain as ipc } from 'electron';
 import { resolve } from 'path';
 
 import Rx from 'rxjs/Rx';
@@ -63,6 +63,7 @@ openFile$
 appReady$
   .subscribe(() => {
     kernelspecs.findAll().then(kernelSpecs => {
+      kernelSpecs = {};
       if (Object.keys(kernelSpecs).length !== 0) {
         // Get the default menu first
         Menu.setApplicationMenu(defaultMenu);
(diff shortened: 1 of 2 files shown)


src/main/index.js
```diff
@@ -39,20 +41,25 @@ ipc.on('open-notebook', (event, filename) => {
...
+        launchNewNotebook('python3');
+      } else {
+        notebooks
+          .filter(Boolean)
+          .filter(x => x !== '.') // Ignore the `electron .`
+          // TODO: Consider opening something for directories
+          .forEach(f => launch(resolve(f)));
+      }
+    } else {
+      console.error('Cannot launch the application if no kernels are installed.');
```
rgbkrk: Note: users that run the app from the Dock or Finder can't actually see these `console.error` messages. They get gobbled by `launchctl`. You'll need a full blown dialog.

rgbkrk (changes requested):

src/main/index.js
captainsafia: Fair. I suppose we should move the dialog box from the menu generation here then. Lemme do that now.

src/main/index.js
```diff
@@ -60,20 +62,38 @@ openFile$
...
+      } else {
+        dialog.showMessageBox({
+          type: 'warning',
+          title: 'No Kernels Installed',
+          buttons: [],
+          message: 'No kernels are installed on your system.',
+          detail: 'No kernels are installed on your system so you will not be ' +
+            'able to execute code cells in any language. You can read about ' +
+            'installing kernels at ' + 
+            'https://ipython.readthedocs.io/en/latest/install/kernel_install.html',
```
rgbkrk, line 90: We may want to point to a page with common kernels in the future so we don't hold on to our Python bias.

rgbkrk (approved):

PR #763, opened 2016-09-14
Combine argv & buff to choose if default launch

rgbkrk:
Finally heading off the difference between how files are opened in dist and dev, while waiting on the app to be ready for BrowserWindow launches.


rgbkrk:
Since I had to rebase on top of changes from #761, I re-used the kernelspecs promise to launch the default notebook based on what's available. I could still make the default Python3 but this seems sensible. My default is now 2... Dun dun dunnnnnnn.

PR #765, opened 2016-09-15
Fix channels not connected return type

rgbkrk:


rgbkrk:
This should increase coverage by 0.026%. 👅

rgbkrk:
Fine codecov, fine.

PR #773, opened 2016-09-15
Move cellStatus to cellMap

### LINKED ISSUE #713: Move cellStatuses attributes to cells under cellMap
The attributes stored in the top-level cellStatuses object should be stored in each cell at the cellMap level.

captainsafia:
Closes #713.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -134,7 +134,14 @@ export function createGistCallback(firstTimePublish, observer, filename, notific
  */
 export function publishNotebookObservable(github, notebook, filepath, notificationSystem) {
   return Rx.Observable.create((observer) => {
-    const notebookString = JSON.stringify(commutable.toJS(notebook), undefined, 1);
+    const notebookString = JSON.stringify(
+      commutable.toJS(notebook.update('cellMap', (cells) => {
+        return cells.map((value, key) => {
+          return value.delete('inputHidden').delete('outputHidden').delete('status');
+        });
+      })),
+      undefined,
+      1);
 
     let filename;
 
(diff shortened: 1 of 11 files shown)


rgbkrk:
We'll still need to remove these from the saved notebook too, otherwise jupyter won't be able to open it (probably worth checking on this).

captainsafia:
> We'll still need to remove these from the saved notebook too, otherwise jupyter won't be able to open it (probably worth checking on this).

Oh drat! Totally forgot about that. Let me fix that.

rgbkrk:
> Oh drat! Totally forgot about that. Let me fix that.

All good. Probably worth noting that I _think_ you can do this in nteract, making sure to use the same function in both saveNotebook and publishing.

captainsafia:
OK. Updated. Although we probably wanna move this code to a utility function within commutable.

rgbkrk:
I actually forsee us having other changes to the in-memory model here that we may want to keep experimenting with them here until we're sure about changing the interface in commutable.

src/notebook/epics/github-publish.js
```diff
@@ -134,7 +134,12 @@ export function createGistCallback(firstTimePublish, observer, filename, notific
  */
 export function publishNotebookObservable(github, notebook, filepath, notificationSystem) {
   return Rx.Observable.create((observer) => {
-    const notebookString = JSON.stringify(commutable.toJS(notebook), undefined, 1);
+    const notebookString = JSON.stringify(
+      commutable.toJS(notebook.update('cellMap', cells =>
+        cells.map(value =>
+          value.delete('inputHidden').delete('outputHidden').delete('status')))),
+      undefined,
```
rgbkrk, line 141: Yeah, I'd actually move this code to our own `toJS` helper, maybe with a file named:

`notebook-model.js`
`notebook-memory-model.js`

Ok, I'm out of ideas for naming here.

PR #774, opened 2016-09-16
Upgraded Linter to 3.5.0, AirBnB to 11.1.0

jdetle:
## Before Merge!!

Check if you want to ignore the unused variable exceptions and the no-undef exceptions.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
   "parser": "babel-eslint",
-  "extends": "airbnb",
+  "extends": "airbnb/legacy",
   "env": {
     "browser": true,
     "node": true,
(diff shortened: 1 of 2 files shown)


jdetle:
This is super frustrating because I can't replicate the error on my machine.

.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
   "parser": "babel-eslint",
-  "extends": "airbnb",
+  "extends": "airbnb/legacy",
```
rgbkrk, line 3: There's one way to do it! This must be new.

.eslintrc
```diff
@@ -11,13 +11,27 @@
...
+       "react/jsx-filename-extension": "off",
+       "import/extensions": [0, { "js": "never", "jsx": "never"}],
+       "max-len": [
+            2,
+            100, {
+                "ignoreComments": true,
+                "ignoreTrailingComments": true,
+                "ignoreUrls": true,
+                "ignoreTemplateLiterals": true
+            }
```
rgbkrk, line 33: Thanks for digging deep into these options.

rgbkrk (approved):

rgbkrk:
It's ok if we go with undef, one of these days we'll want to remove some of our escape hatches and line up with upstream (if it makes sense).

PR #775, opened 2016-09-16
Use home dir for dialogs

captainsafia:
This sets the Save As and Open dialogs to the users home dir.

Working on another PR to make sure the current directory in the kernel is the home dir as well.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -35,7 +35,7 @@ export const fileSubMenus = {
         properties: [
           'openFile',
         ],
-        defaultPath: process.cwd(),
+        defaultPath: app.getPath('home'),
       };
       dialog.showOpenDialog(opts, (fname) => {
         if (fname) {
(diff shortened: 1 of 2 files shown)


rgbkrk (commented):
In some of these cases, when you already have a notebook it seems to have defaulted in the past to the other notebook which is really sensible.

captainsafia:
> In some of these cases, when you already have a notebook it seems to have defaulted in the past to the other notebook which is really sensible.

Do you mean if you were trying to Save As an already Saved Notebook the Save As dialog show default to the path of its previous filename as opposed to the home directory?

rgbkrk:
> Do you mean if you were trying to Save As an already Saved Notebook the Save As dialog show default to the path of its previous filename as opposed to the home directory?

Oh yeah, that's one of them. I went ahead and merged so we can iterate on this and get a feel for good defaults (or write logic to detect what we should be doing).

PR #780, opened 2016-09-17
Fix chomping issue

jdetle:
This got lost somewhere along the line but this should fix issues with the dist not taking notebooks as an argument. @lgeiger if you could check this out on linux I'd appreciate it.


rgbkrk:
We can go with this for now, then we can adapt to what atom does for launch.

jdetle:
It's pretty similar I think: https://github.com/atom/atom/blob/master/src/main-process/main.js

PR #785, opened 2016-09-17
test(executeSpec): Remove it.only

captainsafia:
Embarassing!

This should fix CodeCov.


rgbkrk:
Wow, an increase of 42.84% in one commit! :wink:

Thanks for coming back to fix this.

PR #786, opened 2016-09-17
chore(Codecov): Post new comment on coverage change

captainsafia:
This will force Codecov to post a new comment on every coverage change.

As the name might suggest, it's quite spammy so be mindful of how frequently pushes are made.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,6 @@
 comment:
   layout: header, changes, diff
+  behavior: spammy
 coverage:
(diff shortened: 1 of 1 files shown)


rgbkrk:
> be mindful of how frequently pushes are made

Yeah I think people will keep pushing regardless. I actually like the updating comment feature, I wish it just somehow informed you about updates.

rgbkrk (approved):

PR #787, opened 2016-09-17
fix(launch): filter notebook arguments before processing

rgbkrk:
This fixes an issue on dev launch of the notebooks, and is the wise choice given how we check our `buffer.length` and `notebooks.length`.


jdetle:
"I dont wanna work... I just wanna merge PRs all day" 🎶

PR #791, opened 2016-09-18
fix(newKernel): Set kernel directory to ~

### LINKED ISSUE #707: Default launch to ~ when not opened at terminal
On OS X, if I open the latest release from Finder, the current working directory is `/`. Starting out in `~` would be much better for an intro experience.

captainsafia:
This sets the working directory on a new kernel to the home directory. We're now passing the home dir directly to the kernel launcher instead of making it rely on `process.cwd()`.

Closes #707.

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -6,6 +6,8 @@ import {
 
 import * as path from 'path';
 
+import home from 'home-dir';
+
 import { tildify } from './native-window';
 
 import { executeCell } from './epics/execute';
@@ -221,7 +223,7 @@ export function dispatchLoad(store, event, filename) {
 }
 
 export function dispatchNewNotebook(store, event, kernelSpecName) {
-  store.dispatch(newNotebook(kernelSpecName, remote.app.getPath('home')));
+  store.dispatch(newNotebook(kernelSpecName, home()));
 }
 
 export function initMenuHandlers(store) {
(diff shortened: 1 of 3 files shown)


rgbkrk:
Awesome, this is exactly what we needed for a fresh notebook.

captainsafia:
Ugh. What is it with this test failure?!!? Let me look into it.

test/renderer/menu-spec.js
```diff
@@ -248,7 +248,7 @@ describe('menu', () => {
     expect(store.dispatch.firstCall).to.be.calledWith({
       type: 'NEW_NOTEBOOK',
       kernelSpecName: 'perl',
-      cwd: process.cwd(),
+      cwd: require('home-dir')(),
```
rgbkrk, line 251: 👍

src/notebook/menu.js
```diff
@@ -221,7 +223,7 @@ export function dispatchLoad(store, event, filename) {
 }
 
 export function dispatchNewNotebook(store, event, kernelSpecName) {
-  store.dispatch(newNotebook(kernelSpecName, remote.app.getPath('home')));
+  store.dispatch(newNotebook(kernelSpecName, home()));
```
rgbkrk, line 226: 👍

rgbkrk (approved):

rgbkrk:
Wish the spammy collapsed if it's the same message.

PR #796, opened 2016-09-19
fix(dropdown_css): no border for dropdown menu

jdetle:
Admittedly it does look better.


jdetle:
no_border

static/styles/main.css
```diff
@@ -452,62 +450,3 @@ li.CodeMirror-hint-active {
     display: inline-block;
     opacity: 1.0;
 }
-
-/*
-
-The following were written to help with the R kernel formatting
```
rgbkrk, line 458: Oh wait, this ripped out the formatting for nested views that R spits out. Can you add this back in?

Issue #805, opened 2016-09-20
JS error prevents startup on Applicative Apgar alpha...


fperez:
Wanted to take it for a spin, but after downloading the app, once I try to open it I get a window to pop up with this error info (sorry for the screenshot, OSX lets me select the error but not copy it. Boy, do I miss unix-style copy-on-select for text!):

![image](https://cloud.githubusercontent.com/assets/57394/18658609/6ef116e6-7eb9-11e6-88ea-fd9f488de5fc.png)

Am I doing something wrong?

rgbkrk:
Yikes. Definitely not your fault.

That's happening on launch, let me give the release a try again now that I'm on a different machine.

jdetle:
Hi @fperez @rgbkrk , I'm getting the same issue on my end.

rgbkrk:
This release works when you run it from the CLI like `open ~/Downloads/nteract-darwin-x64/nteract.app`, yet not when opened from the icon. Admittedly, that's the only way I tested it after I made the release. 😢

rgbkrk:
This also has an annoying bug of only failing on the first launch from the app icon.

fperez:
Cool! It works if I open it from the cmd line. This looks great, awesome job! I'm going to play with it :)

BTW, kind of ironic that `interact` doesn't work in `nteract` ;)  But I imagine widgets will come later, no rush.

In any case, congrats! Looking forward to continued development.

rgbkrk:
I'd really love to see a spec form from the widgets dedicated to object synchronization from JS <--> kernel before tackling widgets again (we had support for it for a little while then took it out).

rgbkrk:
Basically, I'd love to see widgets carved up in to separate pieces that are currently unsolved specifications:
- Object Synchronization over the messaging protocol
- Frontend asset loading (or expectations)
- UI toolkit

I realize that ipywidgets has an implementation of these, all in one (slightly by necessity), and that there were IPEPs and implementations in flux (as a writer of a custom DOMWidget, I felt that pain).

rgbkrk:
My [current thinking on synchronized objects](https://github.com/rgbkrk/synchronized-object-diff-spec).

rgbkrk:
Wow @jdetle I found out something weird that "explains" what's happening. I went ahead and created a release with some more logging with the arguments and this is what they were:

```
[ '/Users/kylek/Downloads/nteract-darwin-x64/nteract.app/Contents/MacOS/nteract',
[...]

PR #806, opened 2016-09-20
fix(launch): Filter out strange first args

rgbkrk:
Partial fix for #805.

We're seeing a strange arg passed to the application when launched from Finder on OS X, but only on first launch. This filters that junk arg being passed in out of our notebooks collection.

```
[ '/Users/kylek/Downloads/nteract-darwin-x64/nteract.app/Contents/MacOS/nteract',
  '-psn_0_8046508' ]
```

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -26,7 +26,7 @@ const argv = require('yargs')
 
 const notebooks = argv._
   .filter(Boolean)
-  .filter(x => /^(?!-)/.test(x)) // Ignore strangeness on OS X first launch
+  .filter(x => /^(?!-psn)/.test(x)) // Ignore strangeness on OS X first launch
                                  // see #805
   .filter(x => x !== '.'); // Ignore the `electron .`
                            // TODO: Consider opening something for directories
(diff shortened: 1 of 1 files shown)


rgbkrk:
Note: I left some other logging in that will end up in `~/Library/Logs/nteract/` in case we want to debug this further later. Happy to take it out, and we should take this out before the beta.

rgbkrk:
`psn` everywhere

```
$ ps aux | grep psn
kylek             490   2.1  0.8  3948660 128720   ??  U    Fri04PM   6:03.47 /Applications/iTerm.app/Contents/MacOS/iTerm2 -psn_0_45067
kylek            4828   0.0  0.1  2708784  13580   ??  S    Fri04PM   0:30.51 /System/Library/Services/AppleSpell.service/Contents/MacOS/AppleSpell -psn_0_495737
kylek             487   0.0  2.2  4017944 377272   ??  S    Fri04PM 102:20.13 /Applications/Google Chrome.app/Contents/MacOS/Google Chrome -psn_0_36873
```

rgbkrk:
I'm going to go ahead and merge this + :shipit: since it breaks the release.

PR #813, opened 2016-09-21
dx(debug): expose logging and channels when DEBUG=1

rgbkrk:
dx === developer experience

Log redux actions and expose channels directly when `DEBUG` is set as an environment variable.


rgbkrk:
@captainsafia @jdetle What do you think of this? I'm inevitably lowering coverage a bit for code we can't test without setting an environment variable.

jdetle:
I'm all for it, the logger can be very annoying when its not necessary.

PR #818, opened 2016-09-22
chore(CodeCov): Remove spammy behavior

captainsafia:
Gets rid of CodeCov's spammy behavior.

To avoid problems with merging PRs that drop coverage like the one that resulted in original addition, be sure to install the CodeCov Chrome extension and examine commit diffs for their coverage.


rgbkrk (approved):
Approved so hard.

PR #821, opened 2016-09-22
Title cleanup

rgbkrk:
- Only set title in one place
- Test the observable that pulls properties from the store/state
- Go back to using native API on OS X for represented filename

This brings back something @Carreau provided us, which is this native behavior:

screen shot 2016-09-22 at 11 54 17 am

### DIFF
--- a/src/notebook/native-window.js
+++ b/src/notebook/native-window.js
@@ -36,9 +36,7 @@ export function setTitleFromAttributes(attributes) {
   win.setTitle(title);
 }
 
-export function initNativeHandlers(store) {
-  const state$ = Rx.Observable.from(store);
-
+export function createTitleFeed(state$) {
   const modified$ = state$
     .map(state => state.document.get('notebook'))
     .scan((prev, notebook) => ({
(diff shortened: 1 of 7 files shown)


src/notebook/native-window.js
```diff
@@ -22,40 +22,49 @@ export function tildify(p) {
...
 
   const win = getCurrentWindow();
   // TODO: Investigate if setRepresentedFilename() is a no-op on non-OS X
   if (filename && win.setRepresentedFilename) {
-    // TODO: this needs to be the full path to the file
-    win.setRepresentedFilename(filename);
-  }
-  if (win.setDocumentEdited) {
+    win.setRepresentedFilename(attributes.fullpath);
     win.setDocumentEdited(attributes.modified);
```
captainsafia, line 33: Is it safe to no longer check if `setDocumentEdited` is available?

rgbkrk: This should likely be cased off by `process.platform === 'darwin'`

rgbkrk:
This is all set now. I opted not to case off using platform and instead use property checking so it's easier to test (yay for coverage when testing on Linux).

captainsafia:
Oh, I see how you are doing the check now.

Paranoid me wants to check for both properties on `win`. 😨

PR #824, opened 2016-09-23
[WIP]: feat(GithubAuth): Added user authentication flow.

jdetle:
## ToDone

Browser Window pops up with github information, close upon change of page.
Token gets parsed and saved. 
## Todo
- [ ] Figure out how to store the auth token 
- [ ] Getting a weird error from the behavior around line 28... 

```
Uncaught Exception:
Error: Object has been destroyed
    at Error (native)
    at EventEmitter.webContents.on (/Users/johndetlefs/github/nteract/node_modules/electron-prebuilt/dist/Electron.app/Contents/Resources/electron.asar/browser/api/browser-window.js:50:39)
```

Seems to relate to https://github.com/jprichardson/electron-window/issues/5 but none of these fixes help and it actually doesnt stop the code from working. 
- [ ] Remove log statements
- [ ] Add pager for successful authentication
- [ ] Setup real now.sh server to test.

Ask me and I'll give you my server code so you can tests the branch locally. Or I can just create a now.sh server.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -1,4 +1,4 @@
-import { dialog, app, shell, Menu } from 'electron';
+import { dialog, app, shell, Menu, ipcRenderer as ipc, remote} from 'electron';
 import * as path from 'path';
 
 import { launch, launchNewNotebook } from './launch';
@@ -18,6 +18,7 @@ function createSender(eventName, obj) {
     send(focusedWindow, eventName, obj);
   };
 }
+
 
 export const fileSubMenus = {
   new: {
@@ -72,6 +73,10 @@ export const fileSubMenus = {
   publish: {
     label: '&Publish',
     submenu: [
+      {
+        label: '&Authenticate',
+        click: createSender('menu:publish:auth'),
+      },
       {
         label: '&To Gist',
         click: createSender('menu:publish:gist'),
@@ -348,6 +353,7 @@ export function generateDefaultTemplate() {
 }
 
 export const defaultMenu = Menu.buildFromTemplate(generateDefaultTemplate());
+
 
 export function loadFullMenu() {
   return kernelspecs.findAll().then((kernelSpecs) => {
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -1,4 +1,5 @@
-import { dialog, app, shell, Menu, ipcRenderer as ipc, remote} from 'electron';
+import { dialog, app, shell, Menu, ipcMain as ipc,
+  BrowserWindow} from 'electron';
 import * as path from 'path';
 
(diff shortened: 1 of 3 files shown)


src/main/menu.js
```diff
@@ -19,6 +20,31 @@ function createSender(eventName, obj) {
   };
 }
 
+export function githubAuth() {
```
captainsafia, line 23: I think it would be better if this `githubAuth` function was moved to `src/notebook/menu` and if the authentication menu element triggered an IPC event on click that was listened to in `src/notebook/menu`.

src/notebook/menu.js
```diff
@@ -197,6 +166,26 @@ export function dispatchInterruptKernel(store) {
   }
 }
 
+export function dispatchRestartKernel(store) {
```
captainsafia, line 169: Maybe move this back up to where it was before so the diff doesn't have unnecessary moves.

src/notebook/menu.js
```diff
@@ -6,8 +6,6 @@ import {
 
 import * as path from 'path';
 
-import home from 'home-dir';
-
 import { tildify } from './native-window';
```
captainsafia, line 9: I don't think your branch is rebased since this was added in #791.

src/notebook/menu.js
```diff
@@ -241,7 +230,7 @@ export function dispatchLoad(store, event, filename) {
 }
 
 export function dispatchNewNotebook(store, event, kernelSpecName) {
-  store.dispatch(newNotebook(kernelSpecName, home()));
+  store.dispatch(newNotebook(kernelSpecName));
```
captainsafia, line 233: Same as with the import. I don't think this has been rebased to include #791.

captainsafia (changes requested):
I think you need to rebase your branch with the latest changes from master before continuing. There are several diffs in this commit that remove changes made by recent PRs.

src/notebook/menu.js
jdetle: Hmmm this weird thanks.

src/main/menu.js
jdetle: @rgbkrk was saying that we should probably keep this on the backend. In an attempt to paraphrase: "There is too much indirection by having github auth in `src/notebook/menu`, by having it in the backend we can manage authorization across all open windows"

src/notebook/menu.js
jdetle: Yea I have no idea where this came from.

src/main/menu.js
captainsafia: Meh. Fair.

PR #825, opened 2016-09-23
fix(CSS): Update table CSS

captainsafia:
Let's follow traditional table UX guidelines and make sure our text fields are left-aligned.

Number fields should be right-aligned but I have no idea how to do that at the moment.

screen shot 2016-09-22 at 8 50 59 pm

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -50,10 +50,18 @@ img
     margin-left: auto;
 }
 
+table {
+    border-collapse: collapse;
+}
+
 th,
 td {
     padding: 0.5em 1em;
     border: 1px solid var(--primary-border);
+}
+
+th {
+    text-align: left;
 }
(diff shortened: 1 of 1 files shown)


rgbkrk:
Great idea!

rgbkrk:
> Number fields should be right-aligned but I have no idea how to do that at the moment.

The response back from (e.g. Pandas) would have to provide a CSS class for us to know, if it's a plain HTML table. We have no idea what the type is when we get it back.

rgbkrk:
This is also why I'm interested in a JSON Table schema, because then we can layout according to the actual data not a rendered view after the fact.

rgbkrk:
Seems like we need more margins between some outputs

screen shot 2016-09-22 at 8 14 17 pm

PR #826, opened 2016-09-23
feat(GithubAuth): Add initial work for github user authentication

jdetle:
## ToDone

Browser Window pops up with github information, close upon change of page.
Token gets parsed and saved. 
## Todo
- [x] Figure out how to store the auth token 
- [x] Getting a weird error from the behavior around line 28... 

```
Uncaught Exception:
Error: Object has been destroyed
    at Error (native)
    at EventEmitter.webContents.on (/Users/johndetlefs/github/nteract/node_modules/electron-prebuilt/dist/Electron.app/Contents/Resources/electron.asar/browser/api/browser-window.js:50:39)
```

Seems to relate to https://github.com/jprichardson/electron-window/issues/5 but none of these fixes help and it actually doesnt stop the code from working. 
- [x] Remove log statements
- [ ] Add pager for successful authentication
- [ ] Improve user flow for authenticated publish and unauthenticated publish
- [x] Add tests
- [x] Setup real now.sh server to test.

Ask me and I'll give you my server code so you can tests the branch locally. Or I can just create a now.sh server.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -22,13 +22,14 @@ function createSender(eventName, obj) {
 
 export function githubAuth() {
   const win = new BrowserWindow({show: false, webPreferences: {zoomFactor: .75}});
+  let count = 0;
   win.webContents.on('dom-ready', () => {
-    if( win.getURL().match('authorize') ) {
+    if(count == 1){
+      win.destroy();
+      return githubAuth();
+    }
+    if(win.getURL().match('authorize')) {
       win.show();
-      win.on('page-title-updated', () => {
-        win.close()
-        return githubAuth();
-      });
     }
     else {
       win.webContents.executeJavaScript(`
@@ -40,6 +41,7 @@ export function githubAuth() {
         console.log(auth);
       });
     }
+    count += 1;
   });
   win.loadURL('http://localhost:3010/login');
 }
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -20,7 +20,7 @@ function createSender(eventName, obj) {
   };
 }
 
-export function githubAuth() {
+export function githubAuth(item, focusedWindow) {
   const win = new BrowserWindow({show: false, webPreferences: {zoomFactor: .75}});
   win.webContents.on('dom-ready', () => {
     if( win.getURL().indexOf('callback?code=') != -1 ) {
@@ -29,7 +29,7 @@ export function githubAuth() {
         `);
       ipc.on('auth', (event, auth) => {
         auth = JSON.parse(auth)
-        createSender('menu:publish:auth', auth['access_token']);
+        send(focusedWindow, 'menu:publish:auth', auth['access_token']);
         win.close();
         return;
       });
@@ -95,7 +95,7 @@ export const fileSubMenus = {
     submenu: [
       {
         label: '&Authenticate',
-        click: () => githubAuth(),
+        click: githubAuth,
       },
       {
         label: '&Publish To Gist',
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -37,7 +37,7 @@ export function githubAuth(item, focusedWindow) {
       win.show();
     }
   });
-  win.loadURL('https://oauth.nteract.io/github');
+  win.loadURL('https://nteract-oauth.now.sh/github');
 }
 
 export const fileSubMenus = {
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -22,26 +22,20 @@ function createSender(eventName, obj) {
 
 export function githubAuth() {
   const win = new BrowserWindow({show: false, webPreferences: {zoomFactor: .75}});
-  let count = 0;
   win.webContents.on('dom-ready', () => {
-    if(count == 1){
-      win.destroy();
-      return githubAuth();
-    }
-    if(win.getURL().match('authorize')) {
-      win.show();
-    }
-    else {
+    if( win.getURL().indexOf('callback?code=') != -1 ) {
       win.webContents.executeJavaScript(`
         require('electron').ipcRenderer.send('auth', document.body.textContent);
         `);
       ipc.on('auth', (event, auth) => {
-      // hacky parse of JSON
         auth = JSON.parse(auth)
-        console.log(auth['access_token']);
+        createSender('menu:publish:auth', auth['access_token']);
+        win.close();
+        return;
       });
+    } else {
+      win.show();
     }
-    count += 1;
   });
   win.loadURL('http://localhost:3010/login');
 }
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -92,7 +92,7 @@ export const fileSubMenus = {
     accelerator: 'CmdOrCtrl+Shift+S',
   },
   publish: {
-    label: '&Github',
+    label: '&Publish',
     submenu: [
       {
         label: '&Authenticate',
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -21,15 +21,16 @@ function createSender(eventName, obj) {
 }
 
 export function githubAuth(item, focusedWindow) {
-  const win = new BrowserWindow({show: false, webPreferences: {zoomFactor: .75}});
+  const win = new BrowserWindow({ show: false,
+                                  webPreferences: { zoomFactor: 0.75 } });
   win.webContents.on('dom-ready', () => {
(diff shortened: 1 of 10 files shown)


jdetle:
^ the hackiest of hacks :/

src/main/menu.js
```diff
@@ -19,6 +20,32 @@ function createSender(eventName, obj) {
   };
 }
 
+export function githubAuth() {
+  const win = new BrowserWindow({show: false, webPreferences: {zoomFactor: .75}});
+  let count = 0;
+  win.webContents.on('dom-ready', () => {
+    if(count == 1){
+      win.destroy();
+      return githubAuth();
```
rgbkrk: What's this count part?

src/main/menu.js
```diff
@@ -19,6 +20,32 @@ function createSender(eventName, obj) {
...
+      win.show();
+    }
+    else {
+      win.webContents.executeJavaScript(`
+        require('electron').ipcRenderer.send('auth', document.body.innerHTML);
+        `);
+      ipc.on('auth', (event, auth) => {
+      // hacky parse of JSON
+        auth = auth.match(/"\w+"/g)[1].match(/[^"]/g).join('')
+        console.log(auth);
```
rgbkrk: Can you not JSON.parse it?

rgbkrk: *it?

src/main/menu.js
```diff
@@ -19,6 +20,32 @@ function createSender(eventName, obj) {
...
+    if(count == 1){
+      win.destroy();
+      return githubAuth();
+    }
+    if(win.getURL().match('authorize')) {
+      win.show();
+    }
+    else {
+      win.webContents.executeJavaScript(`
+        require('electron').ipcRenderer.send('auth', document.body.innerHTML);
```
rgbkrk: Since this is browser side, you could use document.querySelector to get more specific with the element and what you want to extract.

src/main/menu.js
jdetle: I can probably do that with the query selected text, thanks!

src/main/menu.js
jdetle: So I think this stems from a lack of understanding. In the case of a new user authenticating, there will be multiple 'dom-ready' events, if that's the case I want to destroy the window after the first event. If there is a more elegant way of doing this sort of thing, please let me know.

src/main/menu.js
```diff
@@ -19,6 +20,26 @@ function createSender(eventName, obj) {
...
+        auth = JSON.parse(auth)
+        send(focusedWindow, 'menu:publish:auth', auth['access_token']);
+        win.close();
+        return;
+      });
+    } else {
+      win.show();
+    }
+  });
+  win.loadURL('http://localhost:3010/login');
```
rgbkrk: We can leave it as localhost for now, but we should probably do `https://oauth.nteract.io/github`.

src/notebook/epics/github-publish.js
```diff
@@ -25,51 +21,9 @@ const Github = require('github');
...
-      // TODO: Prompt?
-      // Leaving this here in case authentication becomes more complicated.
-      console.error(err);
-      return new Github(); // Fall back to no auth
-    })
-    .map(setGithub);
-};
-
-/**
  * Notify the notebook user that it has been published as a gist.
```
rgbkrk, line 27: Look at all this cleanup!

src/notebook/reducers/app.js
```diff
@@ -61,7 +63,10 @@ export default handleActions({
...
   },
-  [constants.SET_GITHUB]: function setGithub(state, action) {
-    return state.set('github', action.github);
-  },
+  [constants.SET_GITHUB_TOKEN]: function setGithubToken(state, action) {
+    const {githubToken} = action;
+    const github = new Github();
+    github.authenticate({ type: 'oauth', token: githubToken }) // synchronous, returns immediately
+    return state.set('github', github);
+  }
```
rgbkrk, line 71: Our assumption now (based on the default state) is that if they don't set the token and they publish, it goes out as anonymous. We can figure out the other flows later.

What I'd like to see is:
- `Publish to Gist` will authenticate if not, then publish the gist
- `Publish to Anonymous Gist` will let you always publish with a fresh publish

rgbkrk:
From "trying" this branch out, we'll want to make sure to use the notification system to display information about being logged in. Either that or... the gutter!

rgbkrk:
Open up a new repo for our oauth server too (even though it just uses `github-oauth`). We need a reproducible artifact for deployment.

src/main/menu.js
```diff
@@ -19,6 +20,26 @@ function createSender(eventName, obj) {
...
+        auth = JSON.parse(auth)
+        send(focusedWindow, 'menu:publish:auth', auth['access_token']);
+        win.close();
+        return;
+      });
+    } else {
+      win.show();
+    }
+  });
+  win.loadURL('https://oauth.nteract.io/github');
```
rgbkrk: Alright, use https://nteract-oauth.now.sh for now while we wait for DNS name servers to propagate since I switched us off CloudFlare.

jdetle:
Up to reviewers whether we should add tests and additional flow improvements here or in a separate PR.

src/main/menu.js
```diff
@@ -70,10 +92,14 @@ export const fileSubMenus = {
     accelerator: 'CmdOrCtrl+Shift+S',
   },
   publish: {
-    label: '&Publish',
+    label: '&Github',
     submenu: [
```
captainsafia, line 96: I think the label on this should continue to be 'Publish' since it's an active verb and lends insight into what the user might be able to access from the submenu. A good arrangement might be Publish > Github > (Authenticate, To Gist).

captainsafia:
> Up to reviewers whether we should add tests and additional flow improvements here or in a separate PR.

I think it is fine to add tests. Looking at this the only things that really need to be covered are the `setGithubToken` action creator in src/notebook/actions and the actual reducer for that. Flow improvements are probably better for another PR.

src/notebook/actions.js
```diff
@@ -232,3 +232,9 @@ export function changeCellType(id, to) {
     to,
   };
 }
+export function setGithubToken(githubToken) {
+  return {
+    type: constants.SET_GITHUB_TOKEN,
+    githubToken,
+  };
+}
```
rgbkrk, line 240: This can and should be tested as part of this PR - follow the action creator tests from other action creators. It seems simple, which only makes it more necessary (in case of changing APIs internally).

src/notebook/menu.js
```diff
@@ -244,6 +245,11 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
   store.dispatch(newNotebook(kernelSpecName, home()));
 }
 
+export function dispatchPublishAuth(store, event, githubToken) {
+  store.dispatch(setGithubToken(githubToken));
+}
```
rgbkrk, line 250: This one can be tested with a mock like this:

``` js
const dispatch = sinon.spy();
const store = { dispatch };

dispatchPublishAuth(store, {}, 'TOKEN');
const expectedAction = { type: 'SET_GITHUB_TOKEN', githubToken: 'TOKEN' };
expect(dispatch).to.have.been.calledWith(expectedAction);
```

You'll need the chai sinon boilerplate that's in other tests.

src/notebook/reducers/app.js
```diff
@@ -61,7 +63,10 @@ export default handleActions({
...
   },
-  [constants.SET_GITHUB]: function setGithub(state, action) {
-    return state.set('github', action.github);
-  },
+  [constants.SET_GITHUB_TOKEN]: function setGithubToken(state, action) {
+    const { githubToken } = action;
+    const github = new Github();
+    github.authenticate({ type: 'oauth', token: githubToken }); // synchronous, returns immediately
+    return state.set('github', github);
+  }
```
rgbkrk, line 71: For this one, you'll want to mock `github.authenticate` using a sinon spy.

``` js
const mockGithub = sinon.spy(github, 'authenticate');

// Call into the reducer with your action
// TODO: Code here

// Check that your spy/stub was called with the right args
expect(mockGithub).to.have.been.calledWith({ type: 'oauth', token: 'TOKEN' });

// Check that the state set your new github object
// which means you'll need a sinon stub instead
```

jdetle:
Erroneously thought dispatch would also cover the reducer. Taking care of that now.

PR #828, opened 2016-09-24
Refactor out pager update and test it

captainsafia:
Yay!

### DIFF
--- a/test/renderer/epics/execute-spec.js
+++ b/test/renderer/epics/execute-spec.js
@@ -21,6 +21,7 @@ import {
   executeCellEpic,
   createExecuteRequest,
   msgSpecToNotebookFormat,
+  createPagerActions,
 } from '../../../src/notebook/epics/execute';
 
 describe('executeCell', () => {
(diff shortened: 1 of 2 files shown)


jdetle:
Ayyyyyyyy

rgbkrk:
Woo!

test/renderer/epics/execute-spec.js
```diff
@@ -142,3 +143,21 @@ describe('msgSpecToNotebookFormat', () => {
...
+    }]);
+
+    const pagerActions$ = createPagerActions('1', msgObs);
+
+    pagerActions$.subscribe((action) => {
+      const expected = [{ source: 'page', data: { 'text/html': 'this is a test' } } ];
+      expect(action.id).to.equal('1');
+      expect(action.pagers.toJS()).to.deep.equal(expected);
+      done();
+    });
```
rgbkrk, line 161: Next steps to follow up on (for this and later tests):

### Apply the `take` operator to the observable from `createPagerActions`:

``` js
const pagerActions$ = createPagerActions('1', msgObs).take(1);
```

This is so that the observable "finishes" (otherwise it would stay open forever). This will make more sense at the bottom of this comment.

### Create an `Array` of the actions

``` js
const actions = []
```

### Then the three callbacks to subscribe should get fully used:
- `onNext` - should push each `action` onto `actions`
- `onError` - should fail the test (errors should not happen here)
- `onComplete` - pass done as the callback

``` js
pagerAction$.subscribe(
  (action) => actions.push(action),
  (err) => expect.fail(), // Should not error
  () => {
    // In reality, you want to verify each action within actions
    // Since we used `take` though, that's kind of silly.
    const action = actions[0]; 
    const expected = [{ source: 'page', data: { 'text/html': 'this is a test' } } ];
    expect(action.id).to.equal('1');
    expect(action.pagers.toJS()).to.deep.equal(expected);
    done();
  }
)
```

PR #835, opened 2016-09-25
Refactor cellAction$ and add tests

captainsafia:
Yay tests!

### DIFF
--- a/src/notebook/epics/execute.js
+++ b/src/notebook/epics/execute.js
@@ -92,7 +92,7 @@ export function createCellStatusAction(id, cellMessages) {
     .map(status => updateCellStatus(id, status));
 }
 
-export function updateCellNumbering(id, cellMessages) {
+export function updateCellNumberingAction(id, cellMessages) {
   return cellMessages.ofMessageType(['execute_input'])
     .pluck('content', 'execution_count')
     .first()
@@ -148,7 +148,7 @@ export function executeCellObservable(channels, id, code) {
     // Set the cell status
     createCellStatusAction(id, cellMessages),
     // Update the input numbering: `[ ]`
-    updateCellNumbering(id, cellMessages),
+    updateCellNumberingAction(id, cellMessages),
     // Clear cell outputs
     Rx.Observable.of(updateCellOutputs(id, new Immutable.List())),
     // Handle all nbformattable messages
--- a/src/notebook/epics/execute.js
+++ b/src/notebook/epics/execute.js
@@ -74,6 +74,42 @@ export function createPagerActions(id, payloadStream) {
     .map((pagerDatas) => updateCellPagers(id, pagerDatas));
 }
 
+export function createSourceUpdateAction(id, setInputStream) {
+  return setInputStream.filter(x => x.replace)
+    .pluck('text')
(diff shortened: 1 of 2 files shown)


captainsafia:
CAUGHT BY THE LINTER AGAIN.

rgbkrk:
Woo, Awesome!

src/notebook/epics/execute.js
```diff
@@ -74,6 +74,41 @@ export function createPagerActions(id, payloadStream) {
     .map((pagerDatas) => updateCellPagers(id, pagerDatas));
 }
 
+export function createSourceUpdateAction(id, setInputStream) {
+  return setInputStream.filter(x => x.replace)
+    .pluck('text')
+    .map(text => updateCellSource(id, text));
+}
```
rgbkrk, line 81: This definitely highlights that the code hasn't been run through here. We can test `handleFormattableMessages` and `createSourceUpdateAction` (with `replace` set) in another PR.

However, this does make me notice something we can abstract (if you feel like it) -- we can provide a helper that binds the id as the first argument (or you can bind it directly):

``` js
function bindById(id, func) {
  return func.bind(null, id);
}
```

Then we can `.map(bindById(id, updateCellSource))`.

However, you could also:

``` js
return setInputStream.filter(x => x.replace)
  .pluck('text')
  .map(updateCellSource.bind(null, id));
```

and not worry about a helper (we don't really need it).

src/notebook/epics/execute.js
```diff
@@ -102,36 +137,21 @@ export function executeCellObservable(channels, id, code) {
...
     Rx.Observable.of(updateCellOutputs(id, new Immutable.List())),
     // Handle all nbformattable messages
-    cellMessages
-      .ofMessageType(['execute_result', 'display_data', 'stream', 'error', 'clear_output'])
-      .map(msgSpecToNotebookFormat)
-      // Iteratively reduce on the outputs
-      .scan(reduceOutputs, emptyOutputs)
-      // Update the outputs with each change
-      .map(outputs => updateCellOutputs(id, outputs))
+    handleFormattableMessages(id, cellMessages),
```
rgbkrk, line 154: Yay, thanks for doing this!

test/renderer/epics/execute-spec.js
```diff
@@ -151,13 +155,78 @@ describe('createPagerActions', () => {
...
+    }]);
+
+    const cellAction$ = createCellAfterAction('1', msgObs);
+
+    cellAction$.subscribe((action) => {
+      expect(action.id).to.equal('1');
+      done();
+    });
+  });
+});
```
rgbkrk, line 184: Please set these up like the [review/instructions in the pager actions PR](https://github.com/nteract/nteract/pull/828#discussion_r80351737) so that we make sure no errors are happening.

rgbkrk (approved):

rgbkrk:
Reviewed (with some sections that need testing updates), went ahead and merged since it pushes us ever so forward. 🚀

captainsafia:
> Reviewed (with some sections that need testing updates), went ahead and merged since it pushes us ever so forward.

Most def. I'll see if I can submit a PR to address your comments above later today.

PR #840, opened 2016-09-27
[WIP] Modeling work

rgbkrk:
This more or less follows the ideas in [sync object diff spec](https://github.com/rgbkrk/synchronized-object-diff-spec) for reduction of models. Note that the comms epic has simplified.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "immutable": "^3.7.6",
     "kernelspecs": "^1.0.1",
     "leaflet": "^0.7.7",
+    "lodash": "^4.16.2",
     "mathjax-electron": "^1.1.1",
     "moment": "^2.15.0",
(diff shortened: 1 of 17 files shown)


rgbkrk:
What's missing here is what a `View` for these models are and how we handle change events from the UI.

rgbkrk:
comms

rgbkrk:
image

image

image

rgbkrk:
```
@jdetle: What falls under `models`? Data for a visualization? That sort of stuff?

@rgbkrk: The transient models are associated with `comms`.

@jdetle: You say transient because its so dependent on user input / quick to change?

@rgbkrk: Transient because we don't store it on disk in the notebook format

@jdetle: Ah
```

rgbkrk:
I'm going to see about bringing more of this into a new PR and close this one.

PR #842, opened 2016-09-27
Populate process.env with the user's shell environment

kitten:
This should copy over all environment variables from `env` into `process.env`.


rgbkrk:
Excellent! Thanks @philpl. This sets us up well for several environment issues we need to work through.

rgbkrk:
Welcome to nteract! :tada:

PR #845, opened 2016-09-27
Documentation

jdetle:
This code attempts to document the work done in the execute epic. This is something that should be seriously reviewed because there are certainly mistakes abound since its fairly rxJS heavy code.

### DIFF
--- a/src/notebook/epics/execute.js
+++ b/src/notebook/epics/execute.js
@@ -22,12 +22,25 @@ const Immutable = require('immutable');
 
 const emptyOutputs = new Immutable.List();
 
+/**
+ * Create an object that adheres to the jupyter notebook specification.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html
+ *
+ * @param {Object} msg - Message with the associated output type
+ * @return {Object} msg  -
+ */
 export function msgSpecToNotebookFormat(msg) {
   return Object.assign({}, msg.content, {
     output_type: msg.header.msg_type,
   });
 }
 
+/**
+ * Insert the content requisite for a code request to a kernel message. 
+ *
+ * @param {Object} code - Code to be executed in a message to the kernel.
+ * @return {Object} msg - Message object containing the code to be sent.
+ */
 export function createExecuteRequest(code) {
   const executeRequest = createMessage('execute_request');
   executeRequest.content = {
@@ -41,6 +54,15 @@ export function createExecuteRequest(code) {
   return executeRequest;
 }
 
+/**
+ * An output can be a stream of data that does not arrive at a single time. This
+ * function handles the different types of outputs and accumulates the data
+ * into a reduced output.
+ *
+ * @param {Object} outputs - Kernel output messages
+ * @param {Object} output -
+ * @return {Object} updated_outputs -
+ */
 export function reduceOutputs(outputs, output) {
   if (output.output_type === 'clear_output') {
     return emptyOutputs;
@@ -68,30 +90,71 @@ export function reduceOutputs(outputs, output) {
   return outputs.push(Immutable.fromJS(output));
 }
 
+/**
+ * Reads a payloadStream and inspects it for pager actions dispatched upon
+ * specific cell executions.
+ *
+ * @param {Object} id - Universally Unique ID of the cell message.
+ * @param {Object} payloadStream -
+ * @return {Object} filtered -
+ */
 export function createPagerActions(id, payloadStream) {
   return payloadStream.filter(p => p.source === 'page')
     .scan((acc, pd) => acc.push(Immutable.fromJS(pd)), new Immutable.List())
     .map((pagerDatas) => updateCellPagers(id, pagerDatas));
 }
 
+/**
+ *
+ *
+ *
+ * @param {Object} id - ID of the cell designated for source update.
+ * @param {Observable} setInputStream -
+ * @return {Object} updated_InputStream -
+ */
 export function createSourceUpdateAction(id, setInputStream) {
   return setInputStream.filter(x => x.replace)
     .pluck('text')
     .map(text => updateCellSource(id, text));
 }
 
+/**
+ *
+ *
+ *
+ * @param {String} id - ID of the cell designated for creation of a cell after it.
+ * @param {Object} setInputStream - Stream that contains a subset of messages
+ * from the kernel that instruct the frontend what it should do.
+ * @return {Object} updated_outputs -
+ */
 export function createCellAfterAction(id, setInputStream) {
   return setInputStream.filter(x => !x.replace)
     .pluck('text')
     .map((text) => createCellAfter('code', id, text));
 }
 
+/**
+ *
+ *
+ * @param {String} id -
+ * @param {Object} output -
+ * @return {Object} updated_outputs -
+ */
 export function createCellStatusAction(id, cellMessages) {
   return cellMessages.ofMessageType(['status'])
     .pluck('content', 'execution_state')
     .map(status => updateCellStatus(id, status));
 }
 
+/**
+ * Cells are numbered according to the order in which they were executed.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-counter-prompt-number
+ * This code updates the cell numbering on the frontend.
+ *
+ * @param {String} id -
+ * @param {Object} cellMessages -
+ * @return {Object} cellMessages -
+ */
 export function updateCellNumberingAction(id, cellMessages) {
   return cellMessages.ofMessageType(['execute_input'])
     .pluck('content', 'execution_count')
@@ -99,6 +162,13 @@ export function updateCellNumberingAction(id, cellMessages) {
     .map(ct => updateCellExecutionCount(id, ct));
 }
 
+/**
+ *
+ *
+ * @param {String} id -
+ * @param {Object} cellMessages -
+ * @return {Object} cellMessages -
+ */
 export function handleFormattableMessages(id, cellMessages) {
   return cellMessages
     .ofMessageType(['execute_result', 'display_data', 'stream', 'error', 'clear_output'])
@@ -109,6 +179,15 @@ export function handleFormattableMessages(id, cellMessages) {
     .map(outputs => updateCellOutputs(id, outputs));
 }
 
+/**
+ *
+ *
+ * @param {Object} channels - The standard channels specified in the jupyter
+ * specification.
+ * @param {String} id -
+ * @param {Object} code -
+ * @return {Observable} updated_outputs -
+ */
 export function executeCellObservable(channels, id, code) {
   if (!channels || !channels.iopub || !channels.shell) {
     return Rx.Observable.throw(new Error('kernel not connected'));
@@ -164,6 +243,13 @@ export function executeCellObservable(channels, id, code) {
 
 export const EXECUTE_CELL = 'EXECUTE_CELL';
 
+/**
+ * TODO
+ *
+ * @param {Object} outputs -
+ * @param {Object} output -
+ * @return {Object} updated_outputs -
+ */
 export function executeCell(id, source) {
   return {
     type: EXECUTE_CELL,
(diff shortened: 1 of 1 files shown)


src/notebook/epics/execute.js
```diff
@@ -22,12 +22,25 @@ const Immutable = require('immutable');
 
 const emptyOutputs = new Immutable.List();
 
+/**
+ * Create an object that adheres to the jupyter notebook specification.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html
+ *
+ * @param {Object} msg - Blank message.
```
captainsafia: `msg` isn't blank.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
   return outputs.push(Immutable.fromJS(output));
 }
 
+/**
+ * Reads a payloadStream and inspects it for pager actions dispatched upon
```
captainsafia: Should be pager data, not pager actions.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
+ * specific cell executions.
+ *
+ * @param {String} id - Universally Unique ID of the cell message.
+ * @param {Observable} payloadStream - Stream containing messages from the
+ * kernel.
+ * @return {Observable} pagerActionStream - Observable stream containing
+ * Pager actions.
```
captainsafia: Same as above. Should be pager data.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
+ */
 export function createPagerActions(id, payloadStream) {
   return payloadStream.filter(p => p.source === 'page')
     .scan((acc, pd) => acc.push(Immutable.fromJS(pd)), new Immutable.List())
     .map((pagerDatas) => updateCellPagers(id, pagerDatas));
 }
 
+/**
+ * Insert a create source update action into a group of messages.
```
captainsafia: It actually dispatches an action to update the source of a cell if the kernel has requested so by setting `replace` to true.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
+ * @return {Observable} updateSourceStream - Stream of updateCellSource actions.
+ */
 export function createSourceUpdateAction(id, setInputStream) {
   return setInputStream.filter(x => x.replace)
     .pluck('text')
     .map(text => updateCellSource(id, text));
 }
 
+/**
+ * Insert a create cell after action into a group of messages.
```
captainsafia: I would use dispatches instead of inserts here as well.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
 export function createCellStatusAction(id, cellMessages) {
   return cellMessages.ofMessageType(['status'])
     .pluck('content', 'execution_state')
     .map(status => updateCellStatus(id, status));
 }
 
+/**
+ * Cells are numbered according to the order in which they were executed.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-counter-prompt-number
+ * This code updates the cell numbering on the frontend.
```
captainsafia: Dispatches an action to update the execution count.

src/notebook/epics/execute.js
```diff
@@ -109,6 +185,17 @@ export function handleFormattableMessages(id, cellMessages) {
     .map(outputs => updateCellOutputs(id, outputs));
 }
 
+/**
+ * Creates an observable containing information requisite to lead
+ * to the execution of a cell in the kernel.
```
captainsafia: It returns an observable with a stream of events that need to happen **after** a cell has been executed.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
 }
 
+/**
+ * Cells are numbered according to the order in which they were executed.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-counter-prompt-number
+ * This code updates the cell numbering on the frontend.
+ *
+ * @param {String} id -  Universally Unique Identifier of cell receiving an
+ * update to the execution count.
+ * @param {Object} cellMessages - Nessages to receive updates.
```
captainsafia: `cellMessages` is an Observable not an Object.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
     .map(ct => updateCellExecutionCount(id, ct));
 }
 
+/**
+ * If a message is formattable, give it the notebook format specification and
+ * reduce outputs.
+ *
+ * @param {String} id - Universally Unique Identifier of cell receiving
+ * messages.
+ * @param {Object} cellMessages - Set of sent cell messages.
```
captainsafia: Same as above. `cellMessages` is an Observable.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
+ * @return {Object} cellMessages - Set of updated cell messages.
```
captainsafia: The output is also an Observable.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
 
+/**
+ * Cells are numbered according to the order in which they were executed.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-counter-prompt-number
+ * This code updates the cell numbering on the frontend.
+ *
+ * @param {String} id -  Universally Unique Identifier of cell receiving an
+ * update to the execution count.
+ * @param {Object} cellMessages - Nessages to receive updates.
+ * @return {Object} cellMessages - Updated messages.
```
captainsafia: This is also an Observable.

src/notebook/epics/execute.js
```diff
@@ -68,37 +90,91 @@ export function reduceOutputs(outputs, output) {
...
 }
 
+/**
+ * Insert a create cell after action into a group of messages.
+ *
+ * @param {String} id -  Universally Unique Identifier of cell to have a cell
+ * created after it.
+ * @param {Object} setInputStream - Stream that contains a subset of messages
+ * from the kernel that instruct the frontend what it should do.
+ * @return {Object} updated_outputs -
```
captainsafia: Both `setInputStream` and the returned value are streams.

captainsafia (changes requested):

src/notebook/epics/execute.js
jdetle: Thanks, I was confused about this, does this come from enchannel?

captainsafia: Ish? All of the Observables in execute functions are subsets of the data that is returned by the shell or iopub subjects which are defined in enchannel-zmq-backend. If you read the code you'll notice that in lines 87 through 101.

src/notebook/epics/execute.js
jdetle: This is totally still a huge pain point for me. http://jsfiddle.net/staltz/8jFJH/48/ Playing with this code leaves me to believe all streams are observables. Should I distinguish streams vs observables in the docs or are they synonymous?

captainsafia: IMO, a stream is a general term for a set of events that happen through time. In can be used in computational contexts that don't relate to reactive programming at all. An Observable is a wrapper around a stream that gives you additional functionality like merging and filtering. In our specific case, we are returning streams in the form of Observables in all our code so you need to write.

```
@param {Rx.Observable} setInputStream 
```

@rgbkrk might have better ideas since he works with RxJS all day.

src/notebook/epics/execute.js
rgbkrk: "Message that has content which can be converted to nbformat (e.g. display_data)"

src/notebook/epics/execute.js
rgbkrk: I wouldn't use the word dispatch here, since `dispatch` goes with redux which we are certainly using here. Maybe "emits"?

src/notebook/epics/execute.js
rgbkrk: Let's call them Observables since that's what they are. The name stream I started using to make the variable make sense.

src/notebook/epics/execute.js
captainsafia: I think it would help to show that this connects to Redux dispatching eventually, hence the word use.

src/notebook/epics/execute.js
rgbkrk: Return type can be:

``` js
@return {Observable<Action>}
```

src/notebook/epics/execute.js
rgbkrk: `iopub` and `shell` are both Subjects, created like this:

``` js
/**
 * Helper function for creating a subject from a socket
 * @param {jmp.Socket} socket the jmp/zmq socket connection to a kernel channel
 * @return {Rx.Subject} subject for sending and receiving messages to kernels
 */
export function createSubject(socket) {
  const subj = Subject.create(createSubscriber(socket),
                              createObservable(socket));
  return subj;
}
```

When you chain off of it (not `shell.send`), you're applying functions to create new observables/streams.

``` js
iopub.filter(x => x.header.msg_type === 'awesome') // Rx.Observable
```

src/notebook/epics/execute.js
rgbkrk: Complete mind-shift: "Observe all the reactions to running code for cell with id"

src/notebook/epics/execute.js
jdetle: "Emits a `createSourceUpdateAction` to update the source of a cell if the kernel has requested so by setting replace to true. This is eventually dispatched as a redux action."

...win win?

rgbkrk: "to be dispatched as a redux action" - since there is zero dispatch - it's intended to be consumed by the user via redux-observable.

rgbkrk: Hopefully this doesn't dilute this too much - these observables can still be used with @philpl's https://github.com/philpl/redux-fluorine since they emit actions.

src/notebook/epics/execute.js
jdetle: Observable<Action> doesn't get highlighted in my current state of atom. Okay if I just default to Observable?

src/notebook/epics/execute.js
```diff
@@ -136,12 +138,12 @@ export function createCellAfterAction(id, setInputStream) {
...
 
 /**
- * Insert a create cell status action into a group of messages.
+ * Emits a create cell status action into a group of messages.
  *
  * @param {String} id -  Universally Unique Identifier of cell receiving
  * an update in cell status.
- * @param {Object} cellMessages - Messages to receive create cell status action.
- * @return {Object} updatedCellMessages - Updated messages.
+ * @param {Observable<Actions>} cellMessages - Messages to receive create cell status action.
```
rgbkrk: This should be `Observable<jmp.Message>` or `jupyter.Message` if you prefer.

jdetle: Same with line 160 I presume?

rgbkrk (approved):

src/notebook/epics/execute.js
rgbkrk: Looks like you've handled it already. :smile:

src/notebook/epics/execute.js
```diff
@@ -22,12 +22,25 @@ const Immutable = require('immutable');
 
 const emptyOutputs = new Immutable.List();
 
+/**
+ * Create an object that adheres to the jupyter notebook specification.
+ * http://jupyter-client.readthedocs.io/en/latest/messaging.html
```
rgbkrk, line 27: This should probably have

``` js
@see https://nbformat.readthedocs.io/en/latest/format_description.html#code-cell-outputs
```

PR #846, opened 2016-09-28
improve(GithubAuth): Add authenticated to app record, rework auth flow.

jdetle:
This is initial work towards fixing the github authorization flow. I think adding 'authenticated' to the app record is a good idea. Initially my thought was to go with a function in src/main/menu called publishAuth:

``` javascript
export function githubAuth(item, focusedWindow) {
  const win = new BrowserWindow({ show: false,
                                  webPreferences: { zoomFactor: 0.75 } });
  win.webContents.on('dom-ready', () => {
    if (win.getURL().indexOf('callback?code=') !== -1) {
      win.webContents.executeJavaScript(`
        require('electron').ipcRenderer.send('auth', document.body.textContent);
        `);
      ipc.on('auth', (event, auth) => {
        const authorization = JSON.parse(auth);
        send(focusedWindow, 'menu:publish:auth', authorization.access_token);
        win.close();
        return;
      });
    } else {
      win.show();
    }
  });
  win.loadURL('https://nteract-oauth.now.sh/github');
}
```

and:

``` javascript
publishAuthAndGist(){
   publishAuth();
   createSender('menu:publish:gist');
}
```

but theres some issue with synchronicity. Gonna pair with @captainsafia on this right now.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -103,18 +103,16 @@ export function publishNotebookObservable(github, notebook, filepath,
     } else {
       filename = 'Untitled.ipynb';
     }
-
     const files = {};
     files[filename] = { content: notebookString };
     if (authenticated) {
-      const email = github.users.get({}, function(err, res) {
-        console.log(err, res);
-      });
-
-      notificationSystem.addNotification({
-        title: 'Authenticated',
-        message: 'Authenticated as `${ email}`',
-        level: 'info',
+      github.users.get({}, (err, res) => {
+        if (err) throw err;
+        notificationSystem.addNotification({
+          title: 'Authenticated',
+          message: `Authenticated as ${res.login}`,
+          level: 'info',
+        });
       });
     }
     notificationSystem.addNotification({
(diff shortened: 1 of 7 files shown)


src/notebook/epics/github-publish.js
```diff
@@ -105,7 +106,17 @@ export function publishNotebookObservable(github, notebook, filepath, notificati
...
     const files = {};
     files[filename] = { content: notebookString };
+    if (authenticated) {
+      const email = github.users.get({}, function(err, res) {
+        console.log(err, res);
+      });
 
+      notificationSystem.addNotification({
+        title: 'Authenticated',
+        message: 'Authenticated as `${ email}`',
```
rgbkrk: I'd use their username instead of their email address.

jdetle:
publish-that-gist

jdetle:
Oops, inadvertently created some modern art.

jdetle:
publish-gist

captainsafia:
Now that I think about it we probably want to use a different term from `Authenticated Gist`. I'm not sure what the best word would be but maybe `User Gist` or something?

rgbkrk:
Slick!

jdetle:
```
Publish 
> Gist Anonymously 
> Gist to Github Profile
```

rgbkrk:
User Gist makes sense.

captainsafia:
You need to update your test cases as well.

Issue #357, opened 2016-04-21
Conda environments should show up as kernels


parleur:
I have a anaconda3 install in my home with my python stuff.

launching `jupyter-kernelspec` I obtain:

```
jupyter kernelspec list
Available kernels:
  python3    /home/parleur/anaconda3/lib/python3.5/site-packages/ipykernel/resources
  xonsh      /home/parleur/anaconda3/share/jupyter/kernels/xonsh
```

launching in node:

```
 `require('kernelspecs').findAll().then(console.log)
Promise { <pending> }
> { xonsh: 
   { name: 'xonsh',
     files: [ '/home/parleur/anaconda3/share/jupyter/kernels/xonsh/kernel.json' ],
     resources_dir: '/home/parleur/anaconda3/share/jupyter/kernels/xonsh',
     spec: 
      { argv: [Object],
        codemirror_mode: 'shell',
        display_name: 'Xonsh',
        language: 'xonsh' } } }
```

I only get the xonsh kernel the second time.

minrk:
nteract doesn't include non-installed kernelspecs. Jupyter special-cases the IPython kernelspec _if it hasn't been installed_, but nteract chooses (rightly, I think) to not do this. To install the IPython kernelspec, do:

```
ipython kernel install
```

If you want to specifically install it in the env, like xonsh has done, specify the prefix:

```
ipython kernel install --prefix=/home/parleur/anaconda3
```

minrk:
If you install ipykernel from conda-forge, this will [already be done](https://github.com/conda-forge/ipykernel-feedstock/blob/master/recipe/meta.yaml#L17).

parleur:
Thanks for reactivity.

After doing `ipython kernel install --prefix=/home/parleur/anaconda3`, the kernel appears when:

```
require('kernelspecs').findAll().then(console.log)
Promise { <pending> }
> { python3: 
   { name: 'python3',
     files: 
      [ '/home/parleur/anaconda3/share/jupyter/kernels/python3/kernel.json',
        '/home/parleur/anaconda3/share/jupyter/kernels/python3/logo-32x32.png',
        '/home/parleur/anaconda3/share/jupyter/kernels/python3/logo-64x64.png' ],
     resources_dir: '/home/parleur/anaconda3/share/jupyter/kernels/python3',
     spec: { language: 'python', argv: [Object], display_name: 'Python 3' } },
  xonsh: 
   { name: 'xonsh',
[...]

PR #847, opened 2016-09-28
[WIP] Create a conda env epic

rgbkrk:
WIP. This is just a start to resolving #357.

Precursors:
- [x] Splash screen for slow loading
- [ ] Main thread redux store

Possible precursor:
- [ ] Bring in/refactor spawnteract to use `spawn-rx`

TODO: 
- [ ] Load kernelspecs for conda in main thread (for menu)

/cc @damianavila for review of `conda` integration in the nteract desktop app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "redux-observable": "^0.12.0",
     "rxjs": "^5.0.0-beta.12",
     "shell-env": "^0.2.0",
+    "spawn-rx": "^2.0.1",
     "spawnteract": "^2.2.0",
     "uuid": "^2.0.3",
     "yargs": "^6.0.0"
--- a/src/notebook/epics/conda-kernel-provider-epic.js
+++ b/src/notebook/epics/conda-kernel-provider-epic.js
@@ -19,8 +19,44 @@ export function condaEnvsObservable(condaInfo$) {
   });
 }
 
+export function condaKernelSpecsObservable(condaPrefix$) {
+  return condaPrefix$;
+}
+
+export function createKernelSpecsFromEnvs(envs) {
+  const displayPrefix = 'Python'; // Or R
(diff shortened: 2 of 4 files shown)


rgbkrk:
Now that I've got this working, I'm tempted to move it and all kernelspecs usage to the main thread, only to run once on startup, providing the kernelspecs to all `BrowserWindow`s. The main thread could also `watch` to see if any environments are deleted or added.

jdetle:
What would we lose if you were to do that?

rgbkrk:
Step right up! You have nothing to lose and only something to gain! Conda based kernels! Environment synchronization!

image

Ok, no, startup for the main app is going to become a tad slower while investigating all the conda environments (if you have any). However, running it in the main thread means it's only slow _once_ on app startup.

jdetle:
Speaking of which, should we agree on sort of performance goal/metric where we say startup should never take more than X amount of time on a machine with some set of specifications? Or is that too arbitrary?

rgbkrk:
Yeah, it's pretty arbitrary. We can evaluate based on how things work with users. This slows things down for people with conda, who are also likely the ones to _want_ this. If they don't have conda it's really quick, little cost. I feel like the tradeoff here is worth it if we manage it well. It's still < half a second on my machine with many conda environments.

captainsafia:
> I feel like the tradeoff here is worth it if we manage it well.

Yep! If we end up having other things happening during startup that slow us down, we can go ahead and implement that loading splash screen we've been thinking about.

rgbkrk:
Rebased on master now that a bunch of pre-requisites to making this work well are done or outlined.

Still a WIP, I'm back to hacking on this.

rgbkrk:
Closing since we have part of this implemented for the main process now.

PR #853, opened 2016-09-29
feat(splash): show a splash page on slow loading

rgbkrk:
A splash page!

screenflow

I'd rather this show up after the first window is fully loaded. This at least starts us off for situtations where the users' env takes a while to determine on startup. We may want to take out the extra GIF inside the notebook window as well.


captainsafia:
Wow! Nice!

> We may want to take out the extra GIF inside the notebook window as well.

Yeah definitely.

PR #861, opened 2016-09-30
chore(cleanup): removing errant console.log()

evanuk:
Stumbled upon this when poking through the tests.


jdetle:
Hah that would definitely be me. Thanks @evanuk.

jdetle:
As an aside, linting the tests will be a pain but something we should do soon.

PR #862, opened 2016-09-30
Anonymize after Authentication

jdetle:
## Changed
- App record now has publishAsUser boolean
- Separated setGithubToken and setPublishAsUser
- DispatchPublishUserGist has clearer logic
- DispatchPublishAnonGist has clearer logic
- Add SET_USER_GITHUB, SET_ANON_GITHUB actions
- Add action/reducer menu tests 
- Add menu spec tests
- Add publish user to dummy store

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -23,7 +23,7 @@ export function authAndPublish(item, focusedWindow) {
   const win = new BrowserWindow({ show: false,
                                   webPreferences: { zoomFactor: 0.75 } });
   if (process.env.AUTHENTICATED) {
-    send(focusedWindow, 'menu:publish:gist');
+    send(focusedWindow, 'menu:github:auth');
     return;
   }
   win.webContents.on('dom-ready', () => {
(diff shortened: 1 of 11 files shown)


captainsafia (changes requested):
Tests need updates.

jdetle:
:shipit:

src/main/menu.js
```diff
@@ -23,7 +23,7 @@ export function authAndPublish(item, focusedWindow) {
   const win = new BrowserWindow({ show: false,
                                   webPreferences: { zoomFactor: 0.75 } });
   if (process.env.AUTHENTICATED) {
-    send(focusedWindow, 'menu:publish:gist');
+    send(focusedWindow, 'menu:github:auth');
```
rgbkrk, line 26: Relying on an environment variable for authenticated is a bit strange here, I need to look into the intent soon.

rgbkrk (commented):
I think you can simplify the publish anonymous logic by not making it part of the state tree -- make a new Github() object every time you publish anon, don't override the old one.

PR #869, opened 2016-09-30
Editor: Better intellisense

### LINKED ISSUE #809: Pressing tab-twice auto-completes instead of tabbing twice
Related to https://github.com/nteract/nteract/issues/808

In this GIF, I'm pressing tab twice to bring my expression into the second for loop. The first tab works but then the second tab executes the autocomplete for the first entry in that list (`file`). So, to indent twice, I have to press `tab`, then `escape`, then `tab` again

tab-twice

Maybe we should remove the auto-complete dropdown list from appearing if the cursor is at the start of the line or is preceded by a tab or a white space?

rgbkrk:
First thing I did was set `completeSingle` to false on Autocomplete. This makes us be less aggressive with typeahead, to prevent those cases where the word is automatically completed yet you keep typing by reflex.

Beyond that, I followed what was in http://stackoverflow.com/a/33908969/700228, with a few modifications to make our matching less greedy.

Fixes #809

### DIFF
--- a/src/notebook/components/cell/editor/index.js
+++ b/src/notebook/components/cell/editor/index.js
@@ -35,6 +35,60 @@ function goLineDownOrEmit(editor) {
   }
 }
 
+const excludedIntelliSenseTriggerKeys = {
+  8: 'backspace',
+  9: 'tab',
+  13: 'enter',
+  16: 'shift',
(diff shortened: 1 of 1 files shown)


jdetle:
Love me some increased coverage.

Issue #248, opened 2016-03-21
Handle IPython codemirror mode


rgbkrk:
From https://github.com/jupyter/notebook/blob/a1458ce8fc275a7acdbd47a86fd7072e989d38b1/notebook/static/notebook/js/codemirror-ipython.js#L20-L37:

``` js
CodeMirror.defineMode("ipython", function(conf, parserConf) {
  var pythonConf = {};
  for (var prop in parserConf) {
    if (parserConf.hasOwnProperty(prop)) {
      pythonConf[prop] = parserConf[prop];
    }
  }
  pythonConf.name = 'python';
  pythonConf.singleOperators = new RegExp("^[\\+\\-\\*/%&|\\^~<>!\\?]");
  if (pythonConf.version === 3) {
    pythonConf.identifiers = new RegExp("^[_A-Za-z\u00A1-\uFFFF][_A-Za-z0-9\u00A1-\uFFFF]*");
  } else if (pythonConf.version === 2) {
    pythonConf.identifiers = new RegExp("^[_A-Za-z][_A-Za-z0-9]*");
  }
  return CodeMirror.getMode(conf, pythonConf);
}, 'python');

CodeMirror.defineMIME("text/x-ipython", "ipython");
```

Carreau:
Could/should it be it's own npm package that define an IPython mode ?

rgbkrk:
Since it's just a function, seems like a good idea. Peer dep on codemirror?

rgbkrk:
Duplicate of #389, I'll keep yours @Carreau.

PR #871, opened 2016-09-30
ipython codemirror mode

rgbkrk:
Working on a way to include the IPython codemirror mode. Magics still not picked up properly.


rgbkrk:
Not what I want permanently, but it fixes #248

PR #873, opened 2016-09-30
fix(Codemirror): indentUnit and tabSize conflict

jdetle:
## Adds:
- `tabSize` as a default codecell prop
- `indentUnit` = `tabSize` in codecell

Really don't have opinions on indenting, if I was supposed to do something else please let me know what direction I should go in.

### DIFF
--- a/src/notebook/components/cell/code-cell.js
+++ b/src/notebook/components/cell/code-cell.js
@@ -23,6 +23,7 @@ class CodeCell extends React.Component {
     running: React.PropTypes.bool,
     focusAbove: React.PropTypes.func,
     focusBelow: React.PropTypes.func,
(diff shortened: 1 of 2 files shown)


src/notebook/components/cell/code-cell.js
```diff
@@ -28,6 +28,7 @@ class CodeCell extends React.Component {
   static defaultProps = {
     pagers: new Immutable.List(),
     running: false,
+    tabSize: 4,
```
captainsafia, line 32: You should also add the typing for `tabSize` in `propTypes`.

captainsafia (changes requested):

captainsafia (approved):

PR #875, opened 2016-10-01
feat(editor): per cell find and replace

rgbkrk:
Demo:

https://www.youtube.com/watch?v=VyE0Vb_P9dY

This is just the default codemirror setup for find and replace, so it only works
per cell.

Ctrl-F / Cmd-F
Start searching

Ctrl-G / Cmd-G
Find next

Shift-Ctrl-G / Shift-Cmd-G
Find previous

Shift-Ctrl-F / Cmd-Option-F
Replace

Shift-Ctrl-R / Shift-Cmd-Option-F
Replace all

Alt-F
Persistent search (dialog doesn't autoclose, enter to find next, Shift-Enter to find previous)

Alt-G
Jump to line

### DIFF
--- a/src/notebook/components/cell/editor/index.js
+++ b/src/notebook/components/cell/editor/index.js
@@ -8,6 +8,10 @@ import Rx from 'rxjs/Rx';
 
 import 'codemirror/addon/hint/show-hint';
 import 'codemirror/addon/hint/anyword-hint';
+import 'codemirror/addon/search/search';
+import 'codemirror/addon/search/searchcursor';
+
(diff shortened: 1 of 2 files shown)


captainsafia:
Nice! I'm curious — is that an overlap with the cell at the bottom and status bar in your video?

rgbkrk:
>  is that an overlap with the cell at the bottom and status bar in your video?

Yeah, it's been that way since the status bar was introduced. I assumed you did that on purpose or didn't notice before.

PR #876, opened 2016-10-01
Add Google Analytics tracking

captainsafia:
Setting us up with tracking so we can get some cool stats.

Shamefully, this slows down launch time quite a bit.

Also, it asks the user whether or not they want to be tracked at every launch.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -44,7 +44,10 @@ ipc.on('open-notebook', (event, filename) => {
 });
 
 const electronReady$ = Rx.Observable.fromEvent(app, 'ready');
-
+const webContentsReady$ = Rx.Observable.fromEvent(
+  app,
+  'web-contents-created',
+  (event, webContents) => ({ event, webContents }));
 
 const fullAppReady$ = Rx.Observable.zip(
   electronReady$,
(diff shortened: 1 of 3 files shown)


src/main/launch.js
```diff
@@ -22,8 +22,26 @@ export function launch(filename) {
...
+    message: 'Would you like to enable Google Analytics?',
+    detail: 'We\'d like to enable tracking in this session of nteract. Why are ' +
+      'we doing this? Tracking usage helps us get statistics for grant funding, ' +
+      'the developement of localization features, and more. To read more about ' +
+      'how we use this information, please visit ' +
+      'https://github.com/nteract/nteract/blob/master/TRACKING.md.',
+  }, (choiceIndex) => {
+    let index;
+    if (choiceIndex === 0) {
+      index = path.join(__dirname, '..', '..', 'static', 'index-with-ga.html');
```
rgbkrk: You can do all of the analytics loading as part of a dynamic epic or another way with the redux state tree, coordinating using `ipc` on launch. Then we don't have to duplicate all the code in the html below _and_ startup can be faster.

rgbkrk (changes requested):

src/main/index.js
```diff
@@ -201,3 +204,24 @@ fullAppReady$
       });
     });
   });
+
+browserWindowReady$
+  .skip(1)
```
rgbkrk, line 209: Does this fire more than once then?

captainsafia:
> Does this fire more than once then?

Yeah, the skip is so that it doesn't run this on the initial splash window. But as before, it runs every time the user opens a new notebook.

I'm trying to see if there is some combination of operators wherein I can say "If this is the second time this event is triggered, show the dialog box and then load Google Analytics into the webpage. For all future emissions, just load GA into the webpage."

rgbkrk:
> I'm trying to see if there is some combination of operators wherein I can say "If this is the second time this event is triggered, show the dialog box and then load Google Analytics into the webpage. For all future emissions, just load GA into the webpage."

I'd make this a global "configuration" thing, just like our theme setup - rely on `electron-json-storage` for now until we have a main thread config system. The config for this can likely be:

``` json
{
  "hasBeenPrompted": true,
  "send-metrics": true,
}
```

then again, you may only need the one field (if this wasn't set up, then we can assume a dialog):

``` json
{
  "send-metrics": true,
}
```

captainsafia:
Closing in favor of another PR with the config approach.

PR #877, opened 2016-10-01
Fix text floating on top of status bar

captainsafia:
Prevents text in cells from floating in front of status bar.


jdetle:
TIL about z-indices!

PR #880, opened 2016-10-01
docs(kernel): outline strategy for myriad kernel launching ways

rgbkrk:
Goal: Provide a way to launch and discover kernels in many different ways. This begins some documentation/strategy for the non-standard kernels we want to support.

### DIFF
--- a/src/notebook/epics/kernels.md
+++ b/src/notebook/epics/kernels.md
@@ -84,6 +84,18 @@ Given that we have direct access to node, we could run a kernel in a webworker
 or via another process with ipc. This requires _no_ ZMQ and could be glorious
 out of the box.
 
+Likely way to support this: [`nel`](https://github.com/n-riesco/nel).
+
+```
+> var session = new nel.Session();
+> session.execute('console.log("nice")', { onSuccess: console.log, onError: console.error })
+undefined
+> { mime: { 'text/plain': 'undefined' } }
+```
+
+That's simple enough we could wrap it with Rx and jupyter messaging protocol
+sans ZMQ.
+
 ### Built-in ijavascript kernel using Electron/nteract session
 
 We could likely provide a kernel that is executing in the context of the current
--- a/src/notebook/epics/kernels.md
+++ b/src/notebook/epics/kernels.md
@@ -122,3 +122,13 @@ the menu.
 * `builtin-ijs`
 * `conda`
 * `default`
+
+Right now we use kernelspecs in two places:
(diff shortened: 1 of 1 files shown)


captainsafia:
Mind changing the document name to README.md so people are more likely to read it?

src/notebook/epics/kernels.md
```diff
@@ -0,0 +1,140 @@
...
+Goal: Provide a way to launch and discover kernels in many different ways.
+
+We currently have a method for discovering, launching, and destroying ZMQ based
+kernels.
+
+## Alternative kernel launchers
+
+The next "pilot" kernels I want us to be able to support are:
+
+* "Built-in"/fallback default Python kernel
```
captainsafia, line 12: Maybe specify that it is a built-in JavaScript kernel?

captainsafia: Disregard that, maybe just say fallback to default Python since built-in implies some sort of packaging.

src/notebook/epics/kernels.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+> var session = new nel.Session();
+> session.execute('console.log("nice")', { onSuccess: console.log, onError: console.error })
+undefined
+> { mime: { 'text/plain': 'undefined' } }
+```
+
+That's simple enough we could wrap it with Rx and jupyter messaging protocol
+sans ZMQ.
```
captainsafia, line 97: 😮 ❗️

src/notebook/epics/kernels.md
```diff
@@ -0,0 +1,140 @@
...
+
+* The menu loading (main process)
+* Launching kernels in the notebook window (renderer process)
+
+For native kernels, we could be shutting them down in the main thread instead of
+the browser windows which would make cleanup more consistent. We're stuck with
+a synchronous response on window close - kernels have to close immediately and
+they may not actually be able to.
+
+We could be pushing the connection config across the inter process boundary for
```
captainsafia, line 137: Can you specify what you mean by interprocess boundary? I assume you mean between the main and renderer processes?

rgbkrk:
If this was nested deeply, that would make sense. This isn't a README for the epics.

captainsafia (approved):
Let's merge and iterate.

rgbkrk:
Oh, thanks. I'm planning to work on this over the weekend so I'll certainly fix up the doc here too.

PR #881, opened 2016-10-02
fix(icon): resolve full path for icon relatively, once

rgbkrk:
We can compute an absolute path to the icon using a relative path with `__dirname`.


jdetle:
Thanks for cleaning up after yours truly :)

rgbkrk:
No prob. Figured this was a quick fix while I'm thinking about the store in the main thread (right now thinking on it for the multi kernel specs).

Issue #741, opened 2016-09-13
Support STDIN


rgbkrk:
<img width="798" alt="screen shot 2016-09-12 at 11 33 52 pm" src="https://cloud.githubusercontent.com/assets/836375/18463792/678457ce-7941-11e6-9440-196b4f3ec631.png">

andyndang:
Support for stdin() is crucial to support interactive debugging in IPython (using `%%debug` magic or `.set_trace()` approach)

rgbkrk:
`stdin` support is now on master!

PR #884, opened 2016-10-02
[WIP] feat(stdin): partial stdin handling

rgbkrk:
This implements part of the stdin handling. I mostly wanted to show what it takes for us to handle part of stdin. There's a big unanswered piece here for me which is how we handle the UI for accepting the input.

This PR is not intended to be merged, only open discussion for #741.

### DIFF
--- a/src/notebook/epics/execute.js
+++ b/src/notebook/epics/execute.js
@@ -48,7 +48,7 @@ export function createExecuteRequest(code) {
     silent: false,
     store_history: true,
     user_expressions: {},
-    allow_stdin: false,
+    allow_stdin: true,
     stop_on_error: false,
   };
   return executeRequest;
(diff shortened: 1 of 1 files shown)


src/notebook/epics/execute.js
```diff
@@ -198,13 +198,42 @@ export function handleFormattableMessages(id, cellMessages) {
...
+        parent_header: request.header,
+      });
+      stdin.next(inputReply); // Technically a side-effect, but it's what we have to do
+      return reply;
+    })
+    // We want this stdin stream to be merged in with the stdout/stderr stream
+    // which means adapting the cellMessages below.
+    // That's simple enough, but what's not clear to me is the UI piece.
+    // What would be a good way for us to prompt internally
+    .map(value => ({ type: 'STDIN_NOT_QUITE_IMPLEMENTED', value }));
```
rgbkrk, line 236: How icky would it be if we converted this plain value to ["stdout stream"](http://jupyter-client.readthedocs.io/en/latest/messaging.html#streams-stdout-stderr-etc).

``` js
{
  parent_header: executeRequest.header,
  header: {
    msg_type: 'stream',
  }
  content: {
    name: 'stdout',
    text: value,
  }
}
```

rgbkrk: Then we'd simply fold this observable into the ones that get nbformatted.

What doesn't please me about this stdin setup is that it doesn't appear to map cleanly for multi-user cases in the future, while the rest of our model here could run server side (/main thread).

rgbkrk:
For styling: http://codepen.io/rgbkrk/pen/ALbkPo

rgbkrk:
I'm going to go ahead and close this, will likely make a new PR later.

PR #902, opened 2016-10-04
fix(css): update 'nteract' theme with running cells

stephymiehle:
Remove striped gradient from pinned cells
Add striped gradient (now with animation!) to running cells

Minor changes:
Update whitespace on gradients and box-shadows to make long lines more readable
Add comment about class specificity

### DIFF
--- a/static/styles/theme-nteract.css
+++ b/static/styles/theme-nteract.css
@@ -47,6 +47,12 @@
   --status-bar: var(--main-bg-color);
 }
 
+/*
+  General note on this theme: the class "notebook" is used before several
+  selectors to override collisions with the main stylesheet. Instead of adding
+  !important on al the things, .notebook makes these selectors more specific.
+*/
+
 /* -------------------- */
 /* Focus on Active Card */
 /* -------------------- */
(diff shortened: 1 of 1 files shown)


static/styles/theme-nteract.css
```diff
@@ -47,6 +47,12 @@
   --status-bar: var(--main-bg-color);
 }
 
+/*
+  General note on this theme: the class "notebook" is used before several
+  selectors to override collisions with the main stylesheet. Instead of adding
+  !important on al the things, .notebook makes these selectors more specific.
+*/
```
rgbkrk, line 54: Awesome, thank you for the comment as well. I can barely do CSS, so I like learning from others.

rgbkrk:
This is so great.

nteract-barbershop

rgbkrk:
By the way, if we've made choices that make it hard to do certain things for styling I'd love to know. This is definitely not my domain area, no matter how much I wish I understood more of it.

stephymiehle:
Oh, not a matter of hard-to-do at all!

The trouble is the way that CSS is loaded. Here, the theme CSS is added first (via `src/notebook/components/notebook.js`) and the main stylesheet is loaded after (via `src/notebook/index.js`). CSS is cascading, so (assuming equally-specific selectors) the value declared last wins. To make the theme changes "win", we need to either get more specific (like prepending the `.notebook` class or adding `!important` after the value). If the CSS files were reordered so that the theme CSS came last, the theme would override the main CSS without additional selectors/specificity.

I just did some quickie testing on my end and variables can be used (in `main.css`) before they are declared (in the theme's CSS), so we should be safe to change the order!

rgbkrk:
Oh great, I would hope to be able to apply the CSS for themes afterward. I was unsure how CSS variables behaved. Thank you!

PR #903, opened 2016-10-04
feat(Windows): Add windows support

lgeiger:
Welcome windows devs! 🎉 

Since we don't have `zmq-prebuilt` for electron on windows (yet),  this will rebuilt zmq from source. Windows devs will need Visual Studio 2013 and Python 2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": ":notebook: the nteract notebook",
   "main": "index.js",
   "scripts": {
+    "postinstall": "node scripts/rebuild.js",
     "prestart": "npm run build",
     "start": "cross-env NODE_ENV=development electron .",
     "prepublish": "npm run build",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "dist:linux32": "npm run dist -- --platform=linux --arch=ia32",
     "dist:osx": "npm run dist -- --platform=darwin --arch=x64 --icon ./static/icon.icns --extend-info=./static/extend.plist --app-category-type=public.app-category.developer-tools --app-bundle-id=io.nteract.nteract",
     "dist:osx:signed": "npm run dist:osx -- --osx-sign",
+    "dist:win": "npm run dist -- --platform=win32 --icon ./static/icon.ico",
     "flow": "flow; test $? -eq 0 -o $? -eq 2"
   },
   "repository": {
--- a/scripts/rebuild.js
+++ b/scripts/rebuild.js
@@ -1,9 +1,17 @@
 #!/usr/bin/env node
-const exec = require('child_process').exec;
-
 if (process.platform === 'win32') {
-  exec('npm rebuild zmq-prebuilt --runtime=electron --target=1.1.3 --disturl=https://atom.io/download/atom-shell --build-from-source',
-  function(err, stdout, stderr) {
+  var exec = require('child_process').exec;
+  var deps = require('../package.json').devDependencies;
+  var version = deps.electron || deps['electron-prebuilt'];
+
+  if (typeof version != 'string' || version.search(/>|<|~|\*|x/) > -1) {
+    throw Error('No explicit electron version in package.json found.')
+  };
+
+  var cmd = 'npm rebuild zmq-prebuilt --runtime=electron --target=' + version +
+    ' --disturl=https://atom.io/download/atom-shell --build-from-source'
+
+  exec(cmd, function(err, stdout, stderr) {
     console.log(stdout);
     console.log(stderr);
     if (err) {
--- a/scripts/rebuild.js
+++ b/scripts/rebuild.js
@@ -1,0 +1,17 @@
+#!/usr/bin/env node
+const exec = require('child_process').exec;
+
+if (process.platform === 'win32') {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "description": ":notebook: the nteract notebook",
   "main": "index.js",
   "scripts": {
+    "postinstall": "node scripts/rebuild.js",
```
lgeiger, line 7: Is `postinstall` the right place for this?

rgbkrk: That's probably fine. `prepublish` is a likely candidate, but I don't know the lifecycle/order for `electron-packager`. If it's working, lets go with it for now. I'm assuming this doesn't run within the dist copy - anyone doing an install is in dev mode.

package.json
```diff
@@ -34,6 +35,7 @@
     "dist:linux32": "npm run dist -- --platform=linux --arch=ia32",
     "dist:osx": "npm run dist -- --platform=darwin --arch=x64 --icon ./static/icon.icns --extend-info=./static/extend.plist --app-category-type=public.app-category.developer-tools --app-bundle-id=io.nteract.nteract",
     "dist:osx:signed": "npm run dist:osx -- --osx-sign",
+    "dist:win": "npm run dist -- --platform=win32 --icon ./static/icon.ico",
```
rgbkrk, line 38: Excellent

scripts/rebuild.js
```diff
@@ -0,0 +1,25 @@
...
+  var cmd = 'npm rebuild zmq-prebuilt --runtime=electron --target=' + version +
+    ' --disturl=https://atom.io/download/atom-shell --build-from-source'
+
+  exec(cmd, function(err, stdout, stderr) {
+    console.log(stdout);
+    console.log(stderr);
+    if (err) {
+      throw err;
+    }
+  });
```
rgbkrk, line 20: While I'm a fan of this, what are we re-doing that's already done in [`electron-rebuild`](https://github.com/electron/electron-rebuild)?

rgbkrk (approved):

scripts/rebuild.js
lgeiger: I tried `electron-rebuild` and it didn't work out of the box. Probably it was missing a `--build-from-source` flag somewhere.

rgbkrk:
Do you think we're ready to pull the trigger on switching to `zmq-static`?

lgeiger:
If we switch to `zmq-static` linux and mac devs would need to install the extra [dependencies](https://github.com/nteract/zmq-static#installation).

I'm OK with it, but it would be harder for new contributors. On the other side we could use the latest electron then.

rgbkrk:
I summon @jdetle to the conversation since I know he both wants new Electron and is closer to the feeling that new contributors face.

lgeiger:
And one would need Python 2 for node-gyp.

That said, 👍 for moving to `zmq-static` (though that pretty much means reverting this PR 😄 )

rgbkrk:
Since we're actually shipping builds, at the very least it means that trial users can just roll with nteract directly and not have to build from source.

My original hesitation to enforce all the build tooling setup was barriers to entry for those that are more on the frontend dev side, less astute with Jupyter.

rgbkrk:
> though that pretty much means reverting this PR

We still have to do the rebuild though right?

lgeiger:
> We still have to do the rebuild though right?

Yes that can be done with `electron-rebuild` then.

jdetle:
A wild @jdetle has appeared! I have no idea what the value of zmq-static is right now, but in terms of installing additional dependencies, I think we're fine. We can always post some sort of caveat, "If you are having trouble installing dependencies for nteract, you're always welcome to ask questions in our [Slack](http://slack.nteract.in/)." (Of course I volunteer to help anyone having issues.)

lgeiger:
@jdetle `zmq-static` is pretty much `zmq-prebuilt` without prebuilts 😄 

It would mean we could use the latest electron release at the cost of building zeromq ourself.

rgbkrk:
Is there a reasonable way to make `zmq-prebuilt` rely on `zmq-static`?

rgbkrk:
I'm only asking that because of the node kernel I've been hacking on - I want to keep it relying on prebuild - it makes install so painless.

lgeiger:
@rgbkrk If we find a way of running [`build.js`](https://github.com/nteract/zmq-prebuilt/blob/master/build.js) on `preinstall` if no binaries are found it would even make `zmq-static` obsulet.

Definitely doable ~~but I don't know how~~.

lgeiger:
@rgbkrk It looks like the `preinstall` script can be included into the gyp file.

This means we could rebuild `zmq-prebuilt`. I'll try it tomorrow and make a PR if it works.

rgbkrk:
I need some sort of excited emoji.

lgeiger:
It works! 🎉

I couldn't wait to try it 😉. The PR will come tomorrow.

scripts/rebuild.js
lgeiger: It looks like I ran into this issue: https://github.com/electron/electron-rebuild/issues/53

PR #906, opened 2016-10-05
refactor(geojson): most recent + valid mimetype for geojson

rgbkrk:
BREAKING CHANGE: application/vnd.geo+json is no longer supported

We definitely don't want to make our permanent geojson mimetype be the outdated one. [`application/geo+json` is preferred](http://www.iana.org/assignments/media-types/application/geo+json).

ref: https://github.com/jupyter/jupyter_client/issues/207, https://github.com/jupyter/jupyter/pull/209

### DIFF
--- a/example-notebooks/geojson.ipynb
+++ b/example-notebooks/geojson.ipynb
@@ -3,7 +3,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {
     "collapsed": false
    },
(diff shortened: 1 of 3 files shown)


rgbkrk:
On a separate note, are other people getting 404s for tiles from MapBox? Admittedly, I figured this would come up at some point - where they would reject the client key. There are other tilesets from other vendors available, the potential to include tiles, letting people configure a key, etc.

jdetle:
Didn't really understand the notebook diffs but I'm going to assume everything is in working order!

rgbkrk:
Yeah the notebook diffs are simply me running straight on through to make sure they were updated with the new mime type.

Issue #894, opened 2016-10-03
C4 Conflicts


rgbkrk:
Our README states that we're using C4, but we're not really:
- No longer doing the Problem -- Solution format on commits and PRs
- Commits are now expected to use conventional changelog
- We rely on BSD instead of MPL

We also should consider updating our [version of C4 to spec 42](https://rfc.zeromq.org/spec:42/C4) from zeromq.

This is currently a bit awkward because the things we're saying and the things were doing don't match right now.

I personally prefer specific virtues under C4, which aim to accept people (by accepting their patches) pretty readily (optimistic merging), even if their commit messages don't match directly. I'd prefer to merge and fix after -- while also leading the contributor to the way to fix it.

PR #911, opened 2016-10-05
chore(docs): Update CONTRIBUTING.md

captainsafia:
Addresses some of #894.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,7 +53,7 @@ You can use the following verbs as part of your commit messages/PR titles.
 * fix: For when you have fixed a bug.
 * feat: For when you have added a new future.
 * chore: For when you've done a small chore on the repository such as updating
-a script in package.json.
+a script in package.json or changing your code based on feedback from the linter.
 * docs: For when you've added documentation.
 
 In addition to the active verb, you'll also need to include the affected component
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,16 @@
+Oh, hello there! You're probably reading this because you are interested in
+contributing to nteract. That's great to hear! This document will help you
+through your journey of open source. Love it, cherish it, take it out to
+dinner, but most importantly: read it thoroughly!
+
 ## What do I need to know to help?
 
-You'll need knowledge of JavaScript (ES6), React, RxJS, and CSS to help out
-with this project.
+You'll need knowledge of JavaScript (ES6), React, RxJS, and Redux to help out
+with this project. That's a whole lot of cool stuff! But don't worry, we've
+got some resources to help you out.
+* [Building a voting app with Redux and React](http://teropa.info/blog/2015/09/10/full-stack-redux-tutorial.html)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,4 +45,33 @@ comes in really handy when we need to determine what kinds of information should
...
 Good release messages means more informed users means a better project to use. Yay!
 
+Note that this standard applies to both your commit messages and PR titles so you'll
+need to draft the appropriate commit message when you run `git commit -m` or use the
+interface on your visual git interface.
+
+You can use the following verbs as part of your commit messages/PR titles.
+* fix: For when you have fixed a bug.
+* feat: For when you have added a new future.
+* chore: For when you've done a small chore on the repository such as updating
```
rgbkrk, line 55: chores can also include linting

CONTRIBUTING.md
```diff
@@ -45,4 +45,33 @@ comes in really handy when we need to determine what kinds of information should
 go into our release documentation (as the word changelog in the title might suggest!).
 Good release messages means more informed users means a better project to use. Yay!
 
+Note that this standard applies to both your commit messages and PR titles so you'll
+need to draft the appropriate commit message when you run `git commit -m` or use the
+interface on your visual git interface.
+
+You can use the following verbs as part of your commit messages/PR titles.
+* fix: For when you have fixed a bug.
+* feat: For when you have added a new future.
```
rgbkrk, line 54: The important distinction with these is that `fix` and `feat` are the prefixes that we use to create changelogs (other prefixes are ignored, which means they can be anything).

CONTRIBUTING.md
```diff
@@ -45,4 +45,33 @@ comes in really handy when we need to determine what kinds of information should
...
+* If you've made a change to an epic, use the name of the epic exported, such as
+`executeCellEpic`.
+
+## How fast will my PR be merged?
+
+Your pull request will be merged as soon as there are maintainers to review it and
+after tests have passed. You might have to make some changes before your PR is merged
+but as long as you adhere to the steps above and try your best, you should have no problem
+getting your PR merged.
+
```
rgbkrk, line 76: 👍

rgbkrk:
Excellent.

rgbkrk (approved):

rgbkrk:
Forgot to note - _we can_ declare whatever prefixes we want to use. Right now we're running under the assumption that we're similar to Angular (given what we link to for the conventional changelog standard). For us, a breaking change isn't in terms of the JS APIs but the userland behavior. However, if/when we export our packages then we will have a strict contract of API behavior.

jdetle:
^ Confession time: I had no idea there were specific verbs that we had to use. Oops.

rgbkrk:
I'm guessing the README needs a once-over as well.

CONTRIBUTING.md
stephymiehle: Whoops, typo here on 54! Unless you guys are wizards or something :)

captainsafia: Welp! You're totally right, @blindingstars. I'll update in another PR that includes the updates to the README.

PR #913, opened 2016-10-05
fix(Contributing): Fix typo.

jdetle:
Today is a gift, that's why they call it the present.


rgbkrk:
Prefix should be `docs` or `chore`, rather than `fix`. This doesn't `fix` a bug.

PR #916, opened 2016-10-05
docs(releases): provide an "example" notebook that looks at our download counts

rgbkrk:
Dog fooding our app by taking a look at download counts.

Some improvements I'd love to see (which I can post an issue for, def new-contributor-friendly):
- Rely completely on pandas for fetching the json and stripping out what we need (potentially)
- Plot with x-axis being version, y-axis being count, and a line for each download type

### DIFF
--- a/example-notebooks/download-stats.ipynb
+++ b/example-notebooks/download-stats.ipynb
@@ -1,0 +1,340 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "source": [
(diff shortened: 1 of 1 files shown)


rgbkrk:
screen shot 2016-10-05 at 12 29 00 pm

captainsafia:
Oh wow, what a coincidence! I was just looking into making a command line tool for this so we can get stats to use in grant reports. This is better though!

rgbkrk:
Awesome! I won't update this branch anymore then, I'll leave you to data-science-fu it after.

My ideal for getting the data over time -- we're going to have to run a job that continuously polls the releases API and captures this for us.

PR #921, opened 2016-10-06
feat(ipynb): handle local ipynb links directly

estevanjantsk:
Hey, I added all the logic to handle local ipynb. @rgbkrk told to use open-url, but it looks like only work on mac. So I decided to do it the hard way (don't know if it is the best way). I'm a rails/angular developer so I don't know all the tricks that node has, but I tried. And again, thanks for the opportunity. You guys are awesome.

### DIFF
--- a/src/main/launch.js
+++ b/src/main/launch.js
@@ -2,9 +2,20 @@ import path from 'path';
 
 import { shell, BrowserWindow } from 'electron';
 
+let launchIpynb;
+
+export function getPath(url) {
+  const nUrl = url.substring(url.indexOf('static'), path.length);
+  return path.join(__dirname, '..', '..', nUrl.replace('static/', ''));
+}
+
 export function deferURL(event, url) {
   event.preventDefault();
-  shell.openExternal(url);
+  if (!url.startsWith('file:')) {
+    shell.openExternal(url);
+  } else if (url.endsWith('.ipynb')) {
+    launchIpynb(getPath(url));
+  }
 }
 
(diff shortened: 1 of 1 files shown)


src/main/launch.js
```diff
@@ -36,6 +47,7 @@ export function launch(filename) {
   });
   return win;
 }
+launchIpynb = launch;
```
rgbkrk, line 50: I'm assuming you did this to appease the linter.

estevanjantsk: :bowtie:

src/main/launch.js
```diff
@@ -2,9 +2,20 @@ import path from 'path';
 
 import { shell, BrowserWindow } from 'electron';
 
+let launchIpynb;
+
+export function getPath(url) {
+  const nUrl = url.substring(url.indexOf('static'), path.length);
```
rgbkrk, line 8: Took me a sec to remember why this has to determine the static path.

It would be good to include a comment here about why.

src/main/launch.js
rgbkrk: Yeah, we'll figure out a way to refactor this after.

rgbkrk (approved):

rgbkrk:
Thanks for making the note about open-url only working on OS X. Glad you thought through the problem and didn't rely on my misinformation. :wink:

rgbkrk:
Looks like this isn't working for the case I just tried (while also not reporting a good error). I'll see if I can dig in.

PR #922, opened 2016-10-06
feat(config): Add config management

captainsafia:
Starting to figure out config file loading/reading/saving.


src/notebook/actions.js
```diff
@@ -232,19 +232,30 @@ export function changeCellType(id, to) {
...
   };
 }
+
 export function setGithubToken(githubToken) {
   return {
     type: constants.SET_GITHUB_TOKEN,
     githubToken,
   };
 }
+
```
rgbkrk, line 254: This action name doesn't tell me well enough what it is (other than the context of this PR). Can we make it "setConfigKey" and `SET_CONFIG_KEY` here?

src/notebook/epics/config.js
```diff
@@ -0,0 +1,45 @@
...
+  Observable.create(observer => {
+    fs.readFile(filename, ...args, (error, data) => {
+      if (error) {
+        observer.error(error);
+      } else {
+        observer.next(data);
+        observer.complete();
+      }
+    })
+  });
```
rgbkrk, line 42: A+

We could rely on `readFileObservable` for both save/load notebooks and this one in one file. How about creating `fs-observable.js` in the same directory?

src/notebook/epics/config.js
```diff
@@ -0,0 +1,45 @@
...
+export const loadConfigEpic = actions =>
+  actions.ofType(LOAD_CONFIG)
+    .do(action =>
+      readFileObservable(getConfigFilePath())
+        .map(JSON.parse)
+        .flatMap(({ config }) =>
+          Observable.of(
+            configLoaded(config),
+          )
+        )
```
rgbkrk: Why have a top-level `config` key in the JSON file? Assuming you change it to the file itself being the raw config,  this can be switched to a simple `map`:

``` js
.map(configLoaded)
```

src/notebook/epics/config.js
```diff
@@ -0,0 +1,45 @@
...
+        .catch((err) =>
+          Observable.of({ type: 'ERROR', payload: err, error: true })
```
rgbkrk, line 58: We so need to figure out our error story in nteract, how we'll do these flows.

src/notebook/records.js
```diff
@@ -36,3 +36,8 @@ export const MetadataRecord = new Immutable.Record({
 export const CommsRecord = new Immutable.Record({
   targets: new Immutable.Map(),
 });
+
+export const ConfigRecord = new Immutable.Record({
+  theme: null,
+  sendMetrics: null,
+});
```
rgbkrk, line 43: 👍

src/notebook/reducers/config.js
```diff
@@ -0,0 +1,15 @@
...
+import * as constants from '../constants';
+
+export default handleActions({
+  [constants.SET_KEY]: function setKey(state, action) {
+    const { key, value } = action;
+    return state.set(key, value);
+  },
+  [constants.SET_CONFIG]: function setConfig(state, action) {
+    const { config } = action;
+    return state.merge(config);
```
rgbkrk, line 13: Oooh, good plan on the merge.

Semantically, we should change this to `MERGE_CONFIG` as the constant unless we're directly `set`ting.

rgbkrk (changes requested):
This is looking awesome, fantastic path.

If/when we have a store on the main thread this looks clean enough to extract to in the future, so that config affects all windows. 😄 

Mega 👍 on direction.

src/notebook/reducers/config.js
captainsafia: Oh good point. I was trying to get the verbs we were using in notebook loading and config loading to line up.

src/notebook/epics/config.js
captainsafia: Yeah, I figured this needed to be refactored out. The only main difference is the `.next` call really.

src/notebook/epics/config.js
captainsafia: Oh! I see. This is what I was intending to do but this is a consequence of me just following what was on the notebook observable without thinking critically about what each thing was doing. _embarassed_

src/notebook/epics/config.js
rgbkrk: It's ok if we do this after the PR, just something to note that seems clean to abstract (and you know I hate over architecting abstractions).

captainsafia:
OK. Loading the config on startup works.

Next step is to update the theme setup so that it relies on the theme key in the configuration.

captainsafia:
OK. The theme is now set up to read in from the config.

Next up, set up saving for the config record on app exit.

src/notebook/epics/config.js
```diff
@@ -54,10 +54,14 @@ export const loadConfigEpic = actions =>
...
+
 export const saveConfigEpic = actions =>
   actions.ofType(SAVE_CONFIG)
-    .mergeMap(action =>
-      writeFileObservable(CONFIG_FILE_PATH, action.config.toJS())
+    .mergeAll(action =>
+      writeFileObservable(CONFIG_FILE_PATH, store.getState().config.toJS())
       .catch((err) =>
         Observable.of({ type: 'ERROR', payload: err, error: true })
       )
```
rgbkrk, line 73: Superb!

captainsafia:
Yay! This works now.

Check out [this video](https://www.youtube.com/watch?v=T1e3a_Uf9DY) where I change themes and the nteract.json file automatically updates.

captainsafia:
Coverage drops with this. Gonna try and figure out testing for those epics as well as refactoring out the fs observables in another PR.

captainsafia:
This closes #180.

src/main/launch.js
```diff
@@ -24,6 +24,7 @@ export function launch(filename) {
     if (filename) {
       win.webContents.send('main:load', filename);
     }
+    win.webContents.send('main:load-config');
```
rgbkrk, line 27: Why are we having to load config here?

src/notebook/actions.js
rgbkrk: Resolved.

src/notebook/components/notebook.js
```diff
@@ -55,7 +55,7 @@ export function getLanguageMode(notebook) {
 }
 
 const mapStateToProps = (state) => ({
-  theme: state.app.theme,
+  theme: state.config.theme,
```
rgbkrk, line 58: 👍

src/notebook/epics/config.js
```diff
@@ -0,0 +1,74 @@
...
+  Observable.create(observer => {
+    fs.readFile(filename, ...args, (error, data) => {
+      if (error) {
+        observer.error(error);
+      } else {
+        observer.next(data);
+        observer.complete();
+      }
+    });
+  });
```
rgbkrk, line 30: I have a feeling this gets tested as part of the notebook load epic, so when you do the refactor after this will be covered.

src/notebook/epics/config.js
```diff
@@ -0,0 +1,74 @@
...
+  Observable.create(observer => {
+    fs.writeFile(filename, data, ...args, error => {
+      if (error) {
+        observer.error(error);
+      } else {
+        observer.next({ filename, data });
+        observer.complete();
+      }
+    });
+  });
```
rgbkrk, line 42: Same thing here for saving.

src/notebook/epics/config.js
```diff
@@ -0,0 +1,74 @@
...
+  type: MERGE_CONFIG,
+  config,
+});
+
+export const CONFIG_FILE_PATH = `${jupyterPaths.dataDirs()[0]}/nteract.json`;
+
+export const loadConfigEpic = actions =>
+  actions.ofType(LOAD_CONFIG)
+    .switchMap(action =>
+      readFileObservable(CONFIG_FILE_PATH)
```
rgbkrk, line 54: This whole block inside the `switchMap` can be brought out into a function for testing.

It's basically `action` -> loaded config testing

rgbkrk (approved):
Nothing glaring, I'm assuming you'll tackle the refactor and testing after. This still has decent coverage so 👍

rgbkrk:
Awesome, glad that's in now.

The next hard problem is making our config be `.js` instead of `.json`. :wink: That way we can do the transform registration, etc.

PR #923, opened 2016-10-06
refactor(StickyCell): Refactor sticky cell

### LINKED ISSUE #701: Refactor sticky cell
Sticky cell is currently using some deprecated or unsavory bits of the React APIs, namely string-refs and `ReactDOM.findDOMNode`.

aleury:
Closes #701

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-return-assign */
 import React from 'react';
 import ReactDOM from 'react-dom';
 import PureRenderMixin from 'react-addons-pure-render-mixin';
@@ -112,9 +113,8 @@ export class Notebook extends React.Component {
   componentDidUpdate() {
     // Make sure the document is vertically shifted so the top non-stickied
     // cell is always visible.
-    const placeholder = ReactDOM.findDOMNode(this.refs['sticky-cells-placeholder']);
-    const container = ReactDOM.findDOMNode(this.refs['sticky-cell-container']);
-    placeholder.style.height = `${container.clientHeight}px`;
+    this.stickyCellsPlaceholder.style.height =
+      `${this.stickyCellContainer.clientHeight}px`;
   }
 
   componentWillUnmount() {
@@ -163,7 +163,7 @@ export class Notebook extends React.Component {
     const viewportHeight = window.innerHeight;
     const viewportOffset = document.body.scrollTop;
 
-    const focusedCell = ReactDOM.findDOMNode(this.refs[id]);
+    const focusedCell = this.refs[id];
 
     if (focusedCell) {
       const cellTop = focusedCell.offsetTop;
@@ -243,8 +243,14 @@ export class Notebook extends React.Component {
     return (
       <div>
         <div className="notebook" ref="cells">
-          <div className="sticky-cells-placeholder" ref="sticky-cells-placeholder" />
-          <div className="sticky-cell-container" ref="sticky-cell-container">
+          <div
+            className="sticky-cells-placeholder"
+            ref={(ref) => this.stickyCellsPlaceholder = ref}
+          />
+          <div
+            className="sticky-cell-container"
+            ref={(ref) => this.stickyCellContainer = ref}
+          >
             {cellOrder
               .filter(id => this.props.stickyCells.get(id))
               .map(this.createStickyCellElement)}


src/notebook/components/notebook.js
```diff
@@ -243,8 +243,14 @@ export class Notebook extends React.Component {
...
         <div className="notebook" ref="cells">
-          <div className="sticky-cells-placeholder" ref="sticky-cells-placeholder" />
-          <div className="sticky-cell-container" ref="sticky-cell-container">
+          <div
+            className="sticky-cells-placeholder"
+            ref={(ref) => this.stickyCellsPlaceholder = ref}
+          />
+          <div
+            className="sticky-cell-container"
+            ref={(ref) => this.stickyCellContainer = ref}
```
rgbkrk, line 252: If you want to suppress the linting warning, you can do:

``` js
(ref) => { this.stickyCellContainer = ref; }
```

src/notebook/components/notebook.js
```diff
@@ -112,9 +113,8 @@ export class Notebook extends React.Component {
   componentDidUpdate() {
     // Make sure the document is vertically shifted so the top non-stickied
     // cell is always visible.
-    const placeholder = ReactDOM.findDOMNode(this.refs['sticky-cells-placeholder']);
-    const container = ReactDOM.findDOMNode(this.refs['sticky-cell-container']);
-    placeholder.style.height = `${container.clientHeight}px`;
+    this.stickyCellsPlaceholder.style.height =
+      `${this.stickyCellContainer.clientHeight}px`;
```
rgbkrk, line 117: 👍

src/notebook/components/notebook.js
```diff
@@ -163,7 +163,7 @@ export class Notebook extends React.Component {
     const viewportHeight = window.innerHeight;
     const viewportOffset = document.body.scrollTop;
 
-    const focusedCell = ReactDOM.findDOMNode(this.refs[id]);
+    const focusedCell = this.refs[id];
```
rgbkrk, line 166: Took me a second to recall why we set the focus cell on unmount, then I realized it's about bringing focus to the unpinned cell. We should probably leave a comment here to that effect.

src/notebook/components/notebook.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-return-assign */
```
rgbkrk, line 1: See note below about how to get around this linting warning without disabling it for the entire file.

rgbkrk (approved):
This is excellent! I've left some review feedback, but this is safe to merge -- it can be addressed in a follow on PR.

src/notebook/components/notebook.js
aleury: Oh okay, awesome! Thanks!

rgbkrk:
By the way, I changed the commit message verb to `refactor` since a bug wasn't fixed (though it does close an issue). Thanks for tackling this one!

aleury:
That makes sense. I called it a bug because the issue had the `bug: ui/ux` label.

Should a new issue be made to address your feedback?

rgbkrk:
Ah, that was mislabeled then. Thanks for pointing that out.

It's ok to make a new PR for the feedback, I wouldn't worry about an issue for it.

captainsafia:
> By the way, I changed the commit message verb to refactor since a bug wasn't fixed (though it does close an issue). Thanks for tackling this one!

We actually don't have refactor defined as a verb in our new CONTRIBUTING.md so we probably should update it.

src/notebook/components/notebook.js
aleury: Hi @rgbkrk, actually this line of code is in the `resolveScrollPosition` method. In an attempt to set the focused cell by a callback ref instead, I noticed that `resolveScrollPosition` isn't called when a cell is unpinned because it remains the cell in focus:

``` javascript
  componentWillReceiveProps(nextProps) {
    if (nextProps.focusedCell !== this.props.focusedCell) {
      this.resolveScrollPosition(nextProps.focusedCell);
    }
  }
```

Commenting out `resolveScrollPosition` and `componentWillReceiveProps` has no noticeable affect. The body still adjusts to have the original position of the unpinned cell in view. Should these methods be removed, or am I just missing something? Thanks!

rgbkrk: Oh derp, I didn't realize we were looking at the main notebook component here.

rgbkrk: We... could probably remove these methods. This is the top level notebook component.

rgbkrk: Wait, wait, the folded diff threw me off.

rgbkrk: I'll check this component out directly, I'm having trouble looking it at it through an old PR.

PR #929, opened 2016-10-06
docs(CommitMessages): Add refactor verb to commit message and PR section

aleury:
As requested by @captainsafia https://github.com/nteract/nteract/pull/923


captainsafia (approved):
Thanks so much!

PR #928, opened 2016-10-06
refactor(transform): bring over transformime-react, adapted

rgbkrk:
This brings over transformime-react and closes https://github.com/nteract/transformime-react/issues/18.

Issue #896, opened 2016-10-03
Configure Flow for Travis


rgbkrk:
On Travis, create a `GROUP` for `flow` and:
- [x] Install `flow`
- [x] Install `flow-typed`
- [x] Run `flow-typed install`
- [x] Run `npm run flow`

PR #930, opened 2016-10-06
chore(flowconfig, Components): Fixed Immutable related Flow errors

peggyrayzis:
@rgbkrk Hey Kyle! Looks like I spoke too soon on your last PR - putting the wildcard after the dist/\* folder caused Flow to throw a module not found error for Immutable. I did some digging and found that Immutable's Flow types are located in a dist/ folder, so I limited the scope of flowconfig to just our dist folder.

Also cleaned up the ImmutableMap typing to include 2 args - if the map values are more specific than any, lmk and I can adjust those definitions. I haven't used Immutable much, let alone w Flow, so this is a learning process for me :)

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -23,7 +23,7 @@ type Props = {
   running: boolean,
   theme: string,
   pagers: ImmutableList<any>,
-  transforms: ImmutableMap<any>,
+  transforms: ImmutableMap<string, any>,
 };
 
 type State = {
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -1,8 +1,7 @@
 /* @flow weak */
 import React from 'react';
 import ReactDOM from 'react-dom';
-import PureRenderMixin from 'react-addons-pure-render-mixin';
-
+import { shouldComponentUpdate } from 'react-addons-pure-render-mixin';
 import { List as ImmutableList, Map as ImmutableMap } from 'immutable';
 
 import CodeCell from './code-cell';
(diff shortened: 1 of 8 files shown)


rgbkrk:
Ohhhhh OK, thanks.

rgbkrk:
Do you have any idea why Rx is not showing up in my `flow-typed` folder:

```
$ head flow-typed/npm/rxjs_vx.x.x.js
// flow-typed signature: 92666ab12556f8797740e7ed3562dae4
// flow-typed version: <<STUB>>/rxjs_v^5.0.0-beta.12/flow_v0.33.0

/**
 * This is an autogenerated libdef stub for:
 *
 *   'rxjs'
 *
 * Fill this stub out by replacing all the `any` types.
 *
```

I see that there are Rx definitions inside of https://github.com/flowtype/flow-typed/tree/master/definitions/npm/rxjs_v5.0.x. Is it related to how we pin it?

src/notebook/components/cell/cell.js
```diff
@@ -23,7 +23,7 @@ type Props = {
   running: boolean,
   theme: string,
   pagers: ImmutableList<any>,
-  transforms: ImmutableMap<any>,
+  transforms: ImmutableMap<string, any>,
```
rgbkrk, line 26: These all map to React Components, which can be the big ol' classes or the stateless function.

src/notebook/components/cell/code-cell.js
```diff
@@ -1,6 +1,7 @@
+// @flow
 import React from 'react';
```
rgbkrk, line 2: Also, uh-oh, we're totally going to conflict in https://github.com/nteract/nteract/pull/928.

peggyrayzis:
No idea why Rx isn't working for you, I'll have to look into it. I did notice that the Rx definitions are only tested against flow_v0.29.0-v0.32.x whereas we're running v0.33. Not sure how big of a difference that makes, but my coworker & I have seen some major discrepancies btwn versions before.

src/notebook/components/cell/code-cell.js
rgbkrk: annnd maybe not, looks like we'll be good to go after CI passes

peggyrayzis: No biggie, do you want to merge first?

rgbkrk: @captainsafia already merged and it looks like we're ready to go here!

rgbkrk (approved):

rgbkrk:
Now that we're in this state, we can probably do https://github.com/nteract/nteract/issues/896

PR #932, opened 2016-10-07
chore(deps): fully migrate off of transformime-react

rgbkrk:
Circling back on updating tests and removing the dependency now that it's part of nteract/nteract.


captainsafia (approved):
👍

Issue #885, opened 2016-10-02
Respect image metadata


rgbkrk:
For image media types, they can (optionally) include width and height as part of their metadata. As stated in [the jupyter client docs for `display_data`](http://jupyter-client.readthedocs.io/en/latest/messaging.html#display-data)

> The only metadata keys currently defined in IPython are the width and height of images:

``` js
metadata = {
  'image/png' : {
    'width': 640,
    'height': 480
  }
}
```

For us this means respecting the image metadata within the ImageDisplay in [`transformime-react`](https://github.com/nteract/transformime-react). The transform does receive a metadata prop:

``` jsx
<ImageDisplay metadata={metadata} />
```

rgbkrk:
Done in the referenced issue, #933.

PR #933, opened 2016-10-07
fix(display): pass width and height through to img component

rgbkrk:
screen shot 2016-10-06 at 10 48 01 pm


rgbkrk:
Closes #885.

jdetle:
Nice! Does the laptop have a name?

rgbkrk:
You know, I'm not sure we named lappy. I guess that's lappy's name. It's featured on https://nteract.io/.

jdetle (approved):

jdetle:
:+1: for lappy 3000

PR #937, opened 2016-10-07
chore(deps): bump enchannel-zmq-backend

rgbkrk:
Bringing us up to date with `zmq-prebuilt` from deep within.

/cc @lgeiger @captainsafia 

We can _probably_ evaluate upgrading electron after this too, using the `rebuild` that Lukas outlined.


captainsafia (approved):
Full speed ahead! 🚀

PR #938, opened 2016-10-07
chore(Components): Add Flow typing for more components

peggyrayzis:
Made some edits per Kyle's feedback in CellCreator; also Flow typed draggable cell, toolbar, & notebook.

How deep do you all want to go w/ Flow typing? Do you want me to start typing the Redux code after the components are done? Happy to help wherever I can! 😀

### DIFF
--- a/src/notebook/components/cell/draggable-cell.js
+++ b/src/notebook/components/cell/draggable-cell.js
@@ -1,22 +1,44 @@
+// @flow
 import React from 'react';
-import PureRenderMixin from 'react-addons-pure-render-mixin';
+import { shouldComponentUpdate } from 'react-addons-pure-render-mixin';
 import { DragSource, DropTarget } from 'react-dnd';
 import { findDOMNode } from 'react-dom';
 
-import Immutable from 'immutable';
+import { List as ImmutableList, Map as ImmutableMap } from 'immutable';
 
 import Cell from './cell';
 import { focusCell } from '../../actions';
 
+type Props = {
+  cell: ImmutableMap<string, any>,
+  connectDragPreview: (img: Image) => void,
+  connectDragSource: (el: ?React.Element<any>) => void,
+  connectDropTarget: (el: ?React.Element<any>) => void,
+  displayOrder: ImmutableList<any>,
+  id: string,
+  isDragging: boolean,
+  isOver: boolean,
+  focusedCell: string,
+  transforms: ImmutableMap<string, any>,
+  language: string,
+  running: boolean,
+  theme: string,
+  pagers: ImmutableList<any>,
+};
+
+type State = {
+  hoverUpperHalf: boolean,
+}
+
 const cellSource = {
-  beginDrag(props) {
+  beginDrag(props: Props) {
     return {
       id: props.id,
     };
   },
 };
 
-function isDragUpper(props, monitor, component) {
+function isDragUpper(props: Props, monitor, component): boolean {
   const hoverBoundingRect = findDOMNode(component).getBoundingClientRect();
   const hoverMiddleY = (hoverBoundingRect.bottom - hoverBoundingRect.top) / 2;
 
@@ -37,7 +59,7 @@ const cellTarget = {
   },
 };
 
-function collectSource(connect, monitor) {
+function collectSource(connect, monitor): Object {
   return {
     connectDragSource: connect.dragSource(),
     isDragging: monitor.isDragging(),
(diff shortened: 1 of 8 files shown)


captainsafia:
@peggyrayzis: I'm certainly not opposed to adding typings to our reducers. That would set it up really nicely for new contributors to help out with documentation as well.

Everything looks good! Except it looks like you must make a sacrifice to appease the Great Linter. :bowing_woman:

jdetle:
My guess is that the more flow-typings we have the merrier ^Looks like @captainsafia beat me to it.

src/notebook/components/notebook.js
```diff
@@ -25,10 +26,27 @@ import {
 } from '../actions';
 import { executeCell } from '../epics/execute';
 
+import type { CellProps } from './cell/cell';
```
rgbkrk, line 29: According to the linter, we somehow have to `import Cell from './cell/cell'` as well as `import type { CellProps } from './cell/cell';` on one line. I'm not even sure how to do that with the `type` construct added on.

rgbkrk (approved):

rgbkrk:
Went ahead and added on the commit to see if Travis passes.

rgbkrk:
Welp, I made that worse.

rgbkrk:
I just rebased the commit I added on out. :-/

src/notebook/components/notebook.js
rgbkrk: After reading some more of the flow docs as well as how these get stripped out, I'm beginning to think that maybe we ignore this line with

```
// eslint-disable-line no-duplicate-imports
```

These imports get stripped anyways, as stated in the Flow FAQ:

> Wait, but what happens at runtime after I’ve added an import type declaration?
> 
> Nothing! All import type declarations get stripped away just like other flow syntax

Turns out that there's an open issue for this in `eslint-plugin-babel`: https://github.com/babel/eslint-plugin-babel/issues/59

rgbkrk: Which... reading some more, it looks like we could use `eslint-plugin-import`...

rgbkrk:
Added some linting configuration to handle `import type` appropriately. Time to see what Travis and AppVeyor think.

src/notebook/components/cell/draggable-cell.js
```diff
@@ -1,22 +1,44 @@
+// @flow
 import React from 'react';
-import PureRenderMixin from 'react-addons-pure-render-mixin';
+import { shouldComponentUpdate } from 'react-addons-pure-render-mixin';
 import { DragSource, DropTarget } from 'react-dnd';
 import { findDOMNode } from 'react-dom';
 
-import Immutable from 'immutable';
+import { List as ImmutableList, Map as ImmutableMap } from 'immutable';
```
rgbkrk, line 7: Why use `ImmutableList` instead of just `List`?

peggyrayzis: I aliased them because Flow has its own types for Lists & Maps:
flow: https://github.com/facebook/flow/blob/v0.33.0/lib/core.js#L512-L527
immutable: https://github.com/facebook/immutable-js/blob/master/dist/immutable.js.flow#L466-Lundefined

I took out the alias to test and it still passes, but I thought it would be better to keep them separate because they look to be independent of each other. Do you know if they are fundamentally different? I haven't worked w/ Immutable much.

rgbkrk: Awwww, ok.

rgbkrk: Ohhhhhh, that's the native `Map` type which we don't use here. We want people to use immutable data structures for everything regarding state.

src/notebook/components/notebook.js
peggyrayzis: Thanks for looking into this & fixing it! 👍  I had tried the one-liner before I submitted this PR with no success.

src/notebook/components/cell/draggable-cell.js
peggyrayzis: Ok cool, let's keep the alias then 👍

rgbkrk:
Is this still WIP? I'm happy to merge it and look forward to your next PRs. 😄  Totally happy to get typing on the Redux code.

peggyrayzis:
No longer WIP for these components! There are a lot of untyped components still remaining though (everything in transforms, editor, & display). Would you like to merge or should I just add on to this PR?

Super excited to start typing Redux too! 🎉

rgbkrk:
I'm about to work on some minor refactors on the notebook focus cell mechanics, so I'd love to have all your changes here. I'll go ahead and merge. It's ok to open lots of PRs. :smile:

PR #940, opened 2016-10-08
chore(deps): bump electron@1.4.3

lgeiger:
Bump electron version, build zmq for electron on `npm install`.

Pros:
- It lets us move to a new Electron version (and every new one after)

Cons:
- This requires linux and mac developers to have [build tools](https://github.com/nteract/zmq-prebuilt/#installation---contributors-and-development) installed

I don't expect us to merge this right away since it'll make installation on linux and os x a lot harder. Just try it out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
   "greenkeeper": {
     "ignore": [
       "commonmark",
-      "electron-prebuilt",
       "redux-observable"
     ]
   },
@@ -114,10 +113,10 @@
     "chai": "^3.4.1",
     "chai-immutable": "^1.6.0",
     "cross-env": "^3.0.0",
+    "electron": "1.4.3",
     "electron-mocha": "^3.0.6",
     "electron-osx-sign": "^0.3.1",
     "electron-packager": "^8.0.0",
-    "electron-prebuilt": "1.1.3",
     "electron-react-devtools": "^0.4.0",
     "enzyme": "^2.2.0",
     "esdoc": "^0.4.8",
--- a/scripts/rebuild.js
+++ b/scripts/rebuild.js
@@ -1,25 +1,19 @@
 #!/usr/bin/env node
-if (process.platform === 'win32') {
-  var exec = require('child_process').exec;
-  var deps = require('../package.json').devDependencies;
-  var version = deps.electron || deps['electron-prebuilt'];
+var exec = require('child_process').exec;
+var deps = require('../package.json').devDependencies;
+var version = deps.electron || deps['electron-prebuilt'];
 
-  if (typeof version != 'string' || version.search(/>|<|~|\*|x/) > -1) {
-    throw Error('No explicit electron version in package.json found.')
-  };
+if (typeof version != 'string' || version.search(/>|<|~|\*|x/) > -1) {
+  throw Error('No explicit electron version in package.json found.')
+};
 
-  var cmd = 'npm rebuild zmq-prebuilt --runtime=electron --target=' + version +
-    ' --disturl=https://atom.io/download/atom-shell --build-from-source'
+var cmd = 'npm rebuild zmq-prebuilt --runtime=electron --target=' + version +
+  ' --disturl=https://atom.io/download/atom-shell --build-from-source'
 
-  exec(cmd, function(err, stdout, stderr) {
-    console.log(stdout);
-    console.log(stderr);
-    if (err) {
-      throw err;
-    }
-  });
-} else if (process.platform === 'linux' || process.platform === 'darwin') {
-  console.log('No postinstall step required for platform' + process.platform);
-} else {
-  console.log(process.platform + ' is not yet supported.');
-}
+exec(cmd, function(err, stdout, stderr) {
+  console.log(stdout);
+  console.log(stderr);
+  if (err) {
+    throw err;
+  }
+});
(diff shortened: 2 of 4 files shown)


lgeiger:
Ahhhh this breaks the coverage script since zmq is now compiled for electron but run from within node.

By the way why aren't the unit tests running on travis as well?

rgbkrk:
> By the way why aren't the unit tests running on travis as well?

The unit tests are run in the `coverage` group.

rgbkrk:
> Ahhhh this breaks the coverage script since zmq is now compiled for electron but run from within node.

Gah, foiled again. We shouldn't be using real kernels inside of the unit tests so we should get that cleaned up.

rgbkrk:
Trying this out now, Electron is certainly working well!

rgbkrk:
Ok, it looks like tests aren't using kernels directly (thankfully). It's merely trying to resolve the dependency to zmq as part of the overall import that happens.

I've "mocked" enchannel-zmq-backend so that the require "works" and the tests can pass.

rgbkrk:
The only thing that scares me now is if any Electron APIs are deprecated. I tried out launching the app, running various menu options, and opening an external link that was within an HTML output.

rgbkrk:
It passes!

package.json
```diff
@@ -114,10 +113,10 @@
     "chai": "^3.4.1",
     "chai-immutable": "^1.6.0",
     "cross-env": "^3.0.0",
+    "electron": "1.4.3",
     "electron-mocha": "^3.0.6",
     "electron-osx-sign": "^0.3.1",
     "electron-packager": "^8.0.0",
-    "electron-prebuilt": "1.1.3",
```
rgbkrk, line 120: I'm so happy to switch to the `electron` namespace - we'll be able to clean up some linting config because of this (minor).

scripts/rebuild.js
```diff
@@ -1,25 +1,19 @@
...
+var exec = require('child_process').exec;
+var deps = require('../package.json').devDependencies;
+var version = deps.electron || deps['electron-prebuilt'];
 
-  if (typeof version != 'string' || version.search(/>|<|~|\*|x/) > -1) {
-    throw Error('No explicit electron version in package.json found.')
-  };
+if (typeof version != 'string' || version.search(/>|<|~|\*|x/) > -1) {
+  throw Error('No explicit electron version in package.json found.')
+};
```
rgbkrk, line 8: 👍

scripts/rebuild.js
```diff
@@ -1,25 +1,19 @@
...
-    }
-  });
-} else if (process.platform === 'linux' || process.platform === 'darwin') {
-  console.log('No postinstall step required for platform' + process.platform);
-} else {
-  console.log(process.platform + ' is not yet supported.');
-}
+exec(cmd, function(err, stdout, stderr) {
+  console.log(stdout);
+  console.log(stderr);
```
rgbkrk, line 15: This should probably be `console.error(stderr)`

scripts/rebuild.js
```diff
@@ -1,25 +1,19 @@
...
+  if (err) {
+    throw err;
+  }
```
rgbkrk, line 18: Would it make sense to move the `if (err)` to the top of this callback?

rgbkrk (approved):

rgbkrk:
@nteract/notebook please give this branch a spin and let us know how it is

scripts/rebuild.js
lgeiger: I can change that. I copied it from `zmq-prebuilt` without looking to much at the code.

lgeiger:
@rgbkrk Thanks for fixing the tests!

> The only thing that scares me now is if any Electron APIs are deprecated.

We could upgrade electron one minor version at a time to see if any deprecation warnings come up.

> Windows builds can be done in automation

Windows is ready since #903 

> This requires users to have build tools installed as well as libzmq

libzmq is bundled with `zmq-prebuilt`

rgbkrk:
> libzmq is bundled with zmq-prebuilt

ohhhhhhh

They only really have hassle with needing Python2 then.

rgbkrk:
> Windows is ready since #903

You're too awesome, I forgot.

rgbkrk:
> We could upgrade electron one minor version at a time to see if any deprecation warnings come up.

Nah, let's rip this bandaid off and go with it.

lgeiger:
> They only really have hassle with needing Python2 then.

Yes. And the [deps for building ZeroMQ](https://github.com/nteract/zmq-prebuilt#installation---contributors-and-development).

jdetle:
This was pretty painless to try out. Start up seems marginally slower. Now it is `npm install --python=python2.7` but other than that I already had all the previous developer dependencies. I'll give it a go on my linux machine right now.

lgeiger:
@jdetle Thanks for trying it out.

You can run `npm config set python python2.7` to be able to use `npm install` again.

jdetle:
All good on Linux. I vote for :shipit:.

jdetle:
Also, thanks _so_ much for doing this. I was trying to play around with similar problems for [sidecar](github.com/nteract/sidecar) and I was way out of my comfort zone.

rgbkrk:
And so it is

captainsafia:
So say we all.

PR #943, opened 2016-10-08
chore(docs): Provide ⬆️ "verb"

rgbkrk:


rgbkrk:
Fun fact: PowerShell does not like emoji branch names.

lgeiger:
Sorry for not complying with the contributing standard. It was late yesterday 😄 

@rgbkrk We can still use ⬆️ in the commit message.

rgbkrk:
That's ok, it's a new standard for us. I'm on the fence about whether I care too deeply about the "conventional change log" commit message format. Commit messages can always be changed before merge, feel free to do what makes you productive and happy - we can always change things later.

PR #945, opened 2016-10-08
fix(defaultPath): Allow OS to determine default path

rgbkrk:
Instead of always going to the home directory, which was put in
place to hack around the default path from launchctl being `/`,
let the OS determine the default path.


rgbkrk:
defaultpath

rgbkrk:
Since the flow run had a hiccup, I both re-triggered the build and made our `travis_before_install` skip the IPython + widgets deps.

lgeiger:
Works really well 👍

PR #947, opened 2016-10-08
chore(travis): Install inodejs as the kernel for builds

rgbkrk:
Seeing if this works.


rgbkrk:
I'm surprised that this worked so well. o______o

Guess this is fit to ship.

Issue #942, opened 2016-10-08
Map ⌘-0 to reset zoom


rgbkrk:
At least on Chrome, ⌘-0 resets the zoom. I'm not sure what the generic binding is, but I'd love to be able to reset the zoom level to 100%.

alexandercbooth:
I'm happy to try and tackle this.

alexandercbooth:
For the commit message, is this considered a chore?

alexandercbooth:
or perhaps feature?

rgbkrk:
This is a feature - it's something new exposed to users.

alexandercbooth:
Makes sense, silly of me, thanks!

rgbkrk:
It's ok, not silly at all. Keep asking questions, happy to work with you.

lgeiger:
Closed by #948 

@alexandercbooth Thanks for your work 🎉

alexandercbooth:
You're very welcome!

PR #948, opened 2016-10-08
Problem: issue #942 addressed with feature to add zoom reset

alexandercbooth:
Solution: Added zoom reset to the menu and mapped it to the suggested keys, CmdOrCtrl+0.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -273,6 +273,11 @@ export const view = {
       click: createSender('menu:zoom-out'),
     },
     {
+      label: 'Zoom Reset',
+      accelerator: 'CmdOrCtrl+0',
+      click: createSender('menu:zoom-reset'),
+    },
+    {
       label: 'Theme',
       submenu: theme_menu,
     },
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -263,6 +263,11 @@ export const view = {
       },
     },
     {
+      label: 'Actual Size',
+      accelerator: 'CmdOrCtrl+0',
+      click: createSender('menu:zoom-reset'),
+    },
+    {
       label: 'Zoom In',
       accelerator: 'CmdOrCtrl+=',
       click: createSender('menu:zoom-in'),
@@ -271,11 +276,6 @@ export const view = {
       label: 'Zoom Out',
       accelerator: 'CmdOrCtrl+-',
       click: createSender('menu:zoom-out'),
-    },
-    {
-      label: 'Zoom Reset',
-      accelerator: 'CmdOrCtrl+0',
-      click: createSender('menu:zoom-reset'),
     },
     {
       label: 'Theme',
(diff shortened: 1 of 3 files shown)


src/main/menu.js
```diff
@@ -273,6 +273,11 @@ export const view = {
       click: createSender('menu:zoom-out'),
     },
     {
+      label: 'Zoom Reset',
```
rgbkrk: To match the Chrome menu, I'd make this "Actual Size"

screen shot 2016-10-08 at 11 14 31 am

(I didn't notice this until a second ago)

alexandercbooth: Good thinking, I will change it right away. Would you mind walking me through the best way to do this commit-wise (should I just do another commit and then rebase?).

Thanks!

rgbkrk: New commits are fine, we don't need things squashed here.

src/notebook/menu.js
```diff
@@ -225,6 +225,10 @@ export function dispatchZoomOut() {
   webFrame.setZoomLevel(webFrame.getZoomLevel() - 1);
 }
 
+export function dispatchZoomReset() {
+  webFrame.setZoomLevel(1);
+}
```
rgbkrk, line 230: Would you want to tackle the test for this? There should be a similar test for dispatchZoomOut you can borrow from.

alexandercbooth: Absolutely, happy to!

rgbkrk: By the way, feel free to find me in the table area in the Atrium at PyData DC.

test/renderer/menu-spec.js
```diff
@@ -118,6 +118,15 @@ describe('menu', () => {
     });
   });
 
+  describe('dispatchZoomReset', () => {
+    it('executes zoom reset', () => {
+      const setZoomLevel = sinon.spy(webFrame, 'setZoomLevel');
+      menu.dispatchZoomReset();
+      setZoomLevel.restore();
+      expect(setZoomLevel).to.be.called;
```
rgbkrk, line 126: Awesome thing we can do here:

``` js
expect(setZoomLevel).to.be.calledWith(1);
```

rgbkrk (approved):

rgbkrk:
Super awesome, thank you.

test/renderer/menu-spec.js
rgbkrk: I just merged this, but you can make another PR if you like.

src/notebook/menu.js
alexandercbooth: Will do!

alexandercbooth:
My pleasure, thanks for the help!

test/renderer/menu-spec.js
alexandercbooth: Happy to, what would you suggest for commit message, chore, test, refactor (would have asked in person but had to leave early today, sorry)?

PR #953, opened 2016-10-08
chore(addScript): Add kernelspec script

jdetle:
When debugging kernelspec problems, this script lists all available kernels. We can just ask that a user with problems runs `node scripts/kernelspec.js`. We can also add this to the package.json scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "dist:osx": "npm run dist -- --platform=darwin --arch=x64 --icon ./static/icon.icns --extend-info=./static/extend.plist --app-category-type=public.app-category.developer-tools --app-bundle-id=io.nteract.nteract",
     "dist:osx:signed": "npm run dist:osx -- --osx-sign",
     "dist:win": "npm run dist -- --platform=win32 --icon ./static/icon.ico",
-    "flow": "flow; test $? -eq 0 -o $? -eq 2"
+    "flow": "flow; test $? -eq 0 -o $? -eq 2",
+    "diagnostics": "scripts/kernelspecs.js"
   },
   "repository": {
     "type": "git",
--- a/scripts/kernelspecs.js
+++ b/scripts/kernelspecs.js
@@ -1,1 +1,2 @@
+#!/usr/bin/env node
 require('kernelspecs').findAll().then(console.log.bind(console)).catch(console.error.bind(console))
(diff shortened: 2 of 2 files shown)


rgbkrk:
Solid idea!

rgbkrk:
...since I keep telling people to do this

scripts/kernelspecs.js
```diff
@@ -0,0 +1 @@
+require('kernelspecs').findAll().then(console.log.bind(console)).catch(console.error.bind(console))
```
rgbkrk, line 2: Above this, I would put `#!/usr/bin/env node`

rgbkrk (approved):

jdetle:
What do you think about adding a script to package.json?

rgbkrk:
If you make this executable `chmod +x scripts/kernelspecs.js` and add the `#!` line, we can have people only running this with `scripts/kernelspecs.js`.

rgbkrk:
> What do you think about adding a script to package.json?

Sure! We could make:

```
npm run diagnostics
```

or something, even reporting node versions, etc.

package.json
```diff
@@ -37,7 +37,8 @@
     "dist:osx": "npm run dist -- --platform=darwin --arch=x64 --icon ./static/icon.icns --extend-info=./static/extend.plist --app-category-type=public.app-category.developer-tools --app-bundle-id=io.nteract.nteract",
     "dist:osx:signed": "npm run dist:osx -- --osx-sign",
     "dist:win": "npm run dist -- --platform=win32 --icon ./static/icon.ico",
-    "flow": "flow; test $? -eq 0 -o $? -eq 2"
+    "flow": "flow; test $? -eq 0 -o $? -eq 2",
+    "diagnostics": "scripts/kernelspecs.js"
```
rgbkrk, line 41: Without the `#!` line in the script, this needs to be `"diagnostics": "node scripts/kernelspecs.js"`.

scripts/kernelspecs.js
rgbkrk: Rad.

package.json
jdetle: (Forgot to add the changes in my last commit )

rgbkrk:
I'll merge as soon as appveyor passes, for good measure.

PR #954, opened 2016-10-08
Problem: change test for issue #942 

alexandercbooth:
Solution: At @rgbkrk's suggestion, added zoom level to test.


rgbkrk:
Super!

Issue #931, opened 2016-10-06
Provide an application/json transform


rgbkrk:
Use https://github.com/alexkuz/react-json-tree, have an awesome navigable object.

aleury:
I'd be happy to take on this one.

rgbkrk:
Sweet, this should be pretty simple. I'd start with looking at the text transform:

https://github.com/nteract/nteract/blob/master/src/notebook/components/transforms/text.js

and then import and add it to the list of transforms + display order:

https://github.com/nteract/nteract/blob/master/src/notebook/components/transforms/index.js#L21

aleury:
Awesome, thanks!

aleury:
So I have a `JsonDisplay` component made but am unsure on how to make sure that it's used. Is the plan to check if the data returned is able to be parsed by `JSON.parse` and return the `application/json` mime type if so, or will this be communicated in some other way?

jdetle:
@aleury can you submit a work in progress PR so people can take a look at what you've done?

aleury:
Submitted the WIP PR.

PR #955, opened 2016-10-08
feat(JsonDisplay): Add a JsonDisplay component

aleury:
https://github.com/nteract/nteract/issues/931. 

I don't intend this to be merged at the moment.

### DIFF
--- a/src/notebook/components/transforms/json.js
+++ b/src/notebook/components/transforms/json.js
@@ -1,0 +1,62 @@
+// @flow
+import React from 'react';
+import JSONTree from 'react-json-tree';
+
+const defaultTheme = {
+  base00: 'transparent',
+  base01: '#073642',
+  base02: '#586e75',
+  base03: '#657b83',
+  base04: '#839496',
+  base05: '#93a1a1',
+  base06: '#eee8d5',
+  base07: '#fdf6e3',
+  base08: '#dc322f',
+  base09: '#cb4b16',
+  base0A: '#b58900',
+  base0B: '#66BB6A',
+  base0C: '#2aa198',
+  base0D: '#268bd2',
+  base0E: '#6c71c4',
+  base0F: '#d33682'
+};
+
+function getTheme(themeName) {
+  switch (themeName) {
+    case 'light':
+    case 'classic': {
+      return defaultTheme;
+    }
+    case 'dark':
+    case 'nteract': {
+      return Object.assign({}, defaultTheme, { base0B: '#EDF3F7' });
+    }
+    case 'halloween': {
+      return Object.assign({}, defaultTheme, {
+        base03: '#E28410',
+        base0B: '#A22222',
+        base0D: '#000000',
+      });
+    }
+    default:
+      return defaultTheme;
+  }
+}
+
+type Props = {
+  data: Object;
+  theme: string;
+}
+
+export default class JsonDisplay extends React.Component {
+  props: Props;
+
+  shouldComponentUpdate() {
+    return true;
+  }
+
+  render() {
+    const theme = getTheme(this.props.theme);
+    return <JSONTree data={this.props.data} theme={theme} invertTheme={false} />;
+  }
+}
--- a/src/notebook/components/transforms/json.js
+++ b/src/notebook/components/transforms/json.js
@@ -51,8 +51,11 @@ type Props = {
 export default class JsonDisplay extends React.Component {
   props: Props;
 
-  shouldComponentUpdate() {
-    return true;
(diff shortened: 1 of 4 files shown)


aleury:
If you execute a cell a with the following

``` python
person = { 'name': 'John Smith' }
person
```

the mime type of the result is `text/plain`, so in order to display it as a json tree, I put a check in `richestMimetype` to check if it can be converted to valid json and if so, return the `application/json` mime type.

rgbkrk:
From Python, here's how you do custom raw mimetypes:

``` python
IPython.display.display({'application/json': data}, raw=True)
```

src/notebook/components/cell/display-area/richest-mime.js
```diff
@@ -32,7 +32,8 @@ export default class RichestMime extends React.Component {
     }
 
     const Transform = this.props.transforms.get(mimetype);
-    const data = this.props.bundle.get(mimetype);
+    // const data = this.props.bundle.get(mimetype);
+    const data = this.props.bundle.get('text/plain');
```
rgbkrk: We definitely don't want to case here, it needs to come in as `application/json`.

rgbkrk (changes requested):
Use either the raw way I outlined in an earlier comment or use:

```
import IPython.display
IPython.display.JSON({"a": 2})
```

To try out the JSON. It should be a direct mapping, no special casing with `text/plain`.

aleury:
Ah! Okay, that makes sense. I'll try that. Thanks!

aleury:
Updated the PR and added a couple of tests.

src/notebook/components/transforms/json.js
```diff
@@ -0,0 +1,62 @@
...
+  base09: '#cb4b16',
+  base0A: '#b58900',
+  base0B: '#66BB6A',
+  base0C: '#2aa198',
+  base0D: '#268bd2',
+  base0E: '#6c71c4',
+  base0F: '#d33682'
+};
+
+function getTheme(themeName) {
```
rgbkrk, line 24: Wow, this was definitely some next level work here. Stellar.

src/notebook/components/transforms/json.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+import React from 'react';
+import JSONTree from 'react-json-tree';
+
+const defaultTheme = {
```
rgbkrk, line 5: Somehow we're going to need to map to our CSS variables so that styling happens via CSS, even if translated here. Not for this PR though, we can figure out an approach later.

src/notebook/components/transforms/json.js
```diff
@@ -0,0 +1,62 @@
...
+type Props = {
+  data: Object;
+  theme: string;
+}
+
+export default class JsonDisplay extends React.Component {
+  props: Props;
+
+  shouldComponentUpdate() {
+    return true;
```
rgbkrk: This should likely be `false` right now, or it needs to check on the theme (similar to how the Leaflet one works). `this.props.data` won't change (because of how the Jupyter messages currently work).

rgbkrk (approved):

rgbkrk:
Any screenshots?

src/notebook/components/transforms/json.js
aleury: That's what I was thinking. I just attempted to come up with something that looked halfway decent for the time being.

rgbkrk: According to their docs, these are also class names. I'm happy to play with that after we merge though.

aleury:
screen shot 2016-10-08 at 10 33 59 pm
screen shot 2016-10-08 at 10 34 26 pm
screen shot 2016-10-08 at 10 34 45 pm
screen shot 2016-10-08 at 10 34 59 pm

rgbkrk:
Lookin' excellent.

src/notebook/components/transforms/json.js
aleury: Sounds good! I'll look over the docs some more.

rgbkrk:
Just take the WIP off the title when you think you're ready. I think we're fine to iterate on this one after. Up to you, this is looking solid.

aleury:
Okay, I updated `componentShouldUpdate` and it should be ready to go! Thanks!

rgbkrk:
Thanks!

rgbkrk:
This is quite good, thank you!

screen shot 2016-10-08 at 10 59 06 pm

aleury:
Any time! 😃

PR #959, opened 2016-10-09
style(theme): label halloween as Pumpkin Spice

rgbkrk:
Tis the season


aleury:
Looks like we're missing `style` as commit message type from the contributing docs. I can add that if you like.

rgbkrk:
Yes please, thanks!

PR #966, opened 2016-10-09
docs(CommitMessages): Add the style commit message type

aleury:
Mentioned in https://github.com/nteract/nteract/pull/959


rgbkrk (approved):
Technically, the "semicolons" part listed would be chore(lint) or so I would think. Or... maybe it is style. Either way, 👍

aleury:
I pulled the wording from https://github.com/commitizen/cz-cli, but I'm happy to change it if you would like!

rgbkrk:
Ah that's great. We adopted this convention without knowing too much about standards, so we can evolve to what seems to be consistent across projects.

PR #968, opened 2016-10-10
plot the number of downloads by version

chendaniely:
Looks like this:
download

Not sure how you would trigger the notebook to update in the future.  Maybe some kind of CI trigger?


jdetle:
Hi Daniel!!!!

jdetle:
I'm fairly sure Darwin and OSX are the same, so maybe you can group those two plots together somehow?

chendaniely:
HI @jdetle fancy seeing you here!

@rgbkrk were playing around with this earlier today.  I guess we didn't notice that OSX was represented in 2 different ways.  Would you prefer darwin or osx? It's a fairly simple column recode

chendaniely:
recoded with: `df['os'] = df.os.replace('os-x', 'darwin-x64')` since `os-x` was only seen in the earlier releases

download

jdetle:
Awesome! I really hope that we see more from you over the next days/ months/ years!

rgbkrk:
Thank you so much! This looks excellent.

rgbkrk:
I'm tempted to make a job that continuously updates a gist with a csv (something I know @parente has done), so we can collect download information over time.

rgbkrk:
As for updates, we could probably have a CI trigger. 🤔

PR #972, opened 2016-10-10
feat(Display): Add scrollable display

captainsafia:
I still have to clean up the tests.


rgbkrk:
Excited for this one.

captainsafia:
OK. This should be good for review.

nteract_expanded

I noticed that we weren't using enzyme in the tests for Display. I assume this is a consequence of bringing the DisplayArea in. Do we want to move the tests over to using enzyme so it is consistent with the rest of the code base?

rgbkrk:
> I noticed that we weren't using enzyme in the tests for Display. I assume this is a consequence of bringing the DisplayArea in. Do we want to move the tests over to using enzyme so it is consistent with the rest of the code base?

Yeah, that's a consequence of bringing DisplayArea in. We should definitely switch to `enzyme`.

src/notebook/actions.js
```diff
@@ -251,3 +251,10 @@ export function setConfigKey(key, value) {
 export function setTheme(theme) {
   return setConfigKey('theme', theme);
 }
+
+export function toggleOutputExpansion(id) {
+  return {
+    type: constants.TOGGLE_OUTPUT_EXPANSION,
+    id,
+  };
```
rgbkrk, line 259: This one isn't covered by the tests and should be a quick and easy action creator test.

src/notebook/components/cell/toolbar.js
```diff
@@ -87,17 +89,21 @@ export default class Toolbar extends React.Component {
     this.context.store.dispatch(changeCellType(this.props.id, to));
   }
 
+  toggleOutputExpansion() {
+    this.context.store.dispatch(toggleOutputExpansion(this.props.id));
```
rgbkrk, line 93: Use `enzyme`'s instance to get the React Element then run this function.

rgbkrk (changes requested):

src/notebook/reducers/document.js
```diff
@@ -240,4 +241,10 @@ export default handleActions({
       .delete(['notebook', 'cellMap', id, 'execution_count'])
       .delete(['notebook', 'cellMap', id, 'outputs']);
   },
+  [constants.TOGGLE_OUTPUT_EXPANSION]: function toggleOutputExpansion(state, action) {
+    const { id } = action;
+
+    return state.updateIn(['notebook', 'cellMap'], (cells) => cells.setIn([id, 'outputExpanded'],
+      !cells.getIn([id, 'outputExpanded'])));
```
rgbkrk, line 248: This looks pretty testable as a reducer.

captainsafia:
> Yeah, that's a consequence of bringing DisplayArea in. We should definitely switch to enzyme.

Yep. I can address that and the coverage drop in another PR.

src/notebook/components/cell/display-area/display.js
```diff
@@ -12,29 +12,41 @@ type Props = {
...
-            theme={props.theme}
-          />
-        )
-      }
-    </div>
-  );
+  const style = {
+    height: props.expanded ? '300px' : 'auto',
+    overflow: props.expanded ? 'scroll' : 'overflow',
+  };
```
rgbkrk, line 25: This is pretty awesome. I wonder if we could do something where it doesn't scroll unless you focus on it too.

PR #974, opened 2016-10-11
Refactor out some string refs, improve app perf

rgbkrk:
This refactors out a few components that still relied on string refs and `findDOMNode`.

While I was at it I noticed something that really helps performance in the notebook. Our global handlers for seeing if the mouse was hovering over a cell were firing `setState` _every time the mouse was moved_. 😨


captainsafia:
Solid! 👍

PR #975, opened 2016-10-11
chore(install): Add dependency check

lgeiger:
The script will check if the required system dependencies are installed.

/cc @rgbkrk

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": ":notebook: the nteract notebook",
   "main": "index.js",
   "scripts": {
-    "postinstall": "node scripts/rebuild.js",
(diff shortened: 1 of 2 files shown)


rgbkrk (approved):

rgbkrk:
I think a post install will not run if the build fails.

rgbkrk:
I'll go ahead and merge, and test out my assumption.

lgeiger:
> I think a post install will not run if the build fails.

For node we have prebuilts. zmq is only rebuild for electron inside the postinstall script.

I didn't use the preinstall script because `command-exists` has to be installed first.

PR #977, opened 2016-10-11
refactor(refs): switch to callback refs in more places

rgbkrk:
Trying to cut down on our string refs.

### DIFF
--- a/src/notebook/components/cell/draggable-cell.js
+++ b/src/notebook/components/cell/draggable-cell.js
@@ -38,8 +38,8 @@ const cellSource = {
   },
 };
 
-function isDragUpper(props: Props, monitor, component): boolean {
-  const hoverBoundingRect = findDOMNode(component).getBoundingClientRect();
+function isDragUpper(props: Props, monitor, el: HTMLElement): boolean {
+  const hoverBoundingRect = el.getBoundingClientRect();
   const hoverMiddleY = (hoverBoundingRect.bottom - hoverBoundingRect.top) / 2;
 
   const clientOffset = monitor.getClientOffset();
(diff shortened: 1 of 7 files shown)


src/notebook/components/cell/draggable-cell.js
```diff
@@ -50,12 +50,12 @@ function isDragUpper(props: Props, monitor, component): boolean {
 
 const cellTarget = {
   drop(props, monitor, component) {
-    const hoverUpperHalf = isDragUpper(props, monitor, component);
+    const hoverUpperHalf = isDragUpper(props, monitor, component.el);
```
rgbkrk, line 53: @peggyrayzis do you know how I'd specify this monitor with Flow? It's not an exported type from [react-dnd](http://gaearon.github.io/react-dnd/) yet is [in the flow-typed repo](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-dnd_v2.x.x/flow_v0.23.x-/react-dnd_v2.x.x.js) (I also have no idea which monitor this is at first glance).

peggyrayzis: Have you ran flow typed install yet? I think it's supposed to grab any available libdefs and import them automatically for us based on our npm packages.

As for which monitor, it looks like we're using several different monitors throughout the component:
Lines 52-53 could be any of the 3 since they all have a getClientOffset property. We should log out the monitor to see which one it is.

The monitor in lines 62-68 could be either DragLayerMonitor or DragSourceMonitor based on the isDragging prop.

The monitor in 70-75 is definitely the DropTargetMonitor based on the isOver prop.

rgbkrk: Yeah, I've run `flow-typed install`. I guess I'm not sure how to "require" it?

PR #980, opened 2016-10-11
docs(install): Remove unnecessary dependencies

lgeiger:
Please test this. It sounds to good to be true 😆

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,19 @@ python3 -m ipykernel install --user
 
 Requires [node 6.x and npm 3](https://docs.npmjs.com/getting-started/installing-node).
 
-**Prerequisites for Linux and OS X**
+**Prerequisites for Linux**
+- `python` (`v2.7` recommended, `v3.x.x` is not supported)
+- `make`
+- `wget`
+- A proper C/C++ compiler toolchain, like [GCC](https://gcc.gnu.org/)
 
-If you are running on Linux or OS X, you will need to have `automake`,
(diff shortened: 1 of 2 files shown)


rgbkrk:
Yeah I'm pretty surprised about `pkg-config`.

rgbkrk:
The only true way we can test this on OS X is to put it on a fresh Mac.

For Linux, we can test this in Docker or a VM. 🤔

rgbkrk:
Even if we have to add them back in later, the brilliant thing is that the dep check script here can be updated and we'll still inform devs in automation.

lgeiger:
I'm pretty confident about `pkg-config`. I ran `sudo apt-get remove pkg-config` and everything still works.

lgeiger:
I'm going to test this on a fresh install:
- [x] OS X 10.11.6
- [x] Ubuntu 16.04

lgeiger:
The installation works out of the box on a fresh install of OS X. 🎉 

I get a warning because `python2` is unset. But we can check for `python2.7` which is the default.
Fun fact: 2GB RAM isn't enough to build zeromq on the first try 😄

rgbkrk:
> Fun fact: 2GB RAM isn't enough to build zeromq on the first try

Wow that is a very important fun fact.

lgeiger:
It isn't that bad. On the second try the build starts where it left of and succeeds.

lgeiger:
All good on Ubuntu as well! 🎉

PR #982, opened 2016-10-11
refactor(execute): compose children messages using childOf

rgbkrk:
Noticed this while sitting down with @smashwilson.


jdetle:
👍

PR #985, opened 2016-10-11
chore(components): Added Flow typing to display-area components

peggyrayzis:
Hey! I added typing to all the display-area components & fixed the Flow errors that resulted from the ref refactor earlier today.

All of the components should be Flow typed now 🎉 time to start on the Redux portion!

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -38,6 +38,7 @@ export class Cell extends React.Component {
   focusAboveCell: () => void;
   focusBelowCell: () => void;
   setCellHoverState: (mouseEvent: MouseEvent) => void;
+  cellDiv: HTMLElement;
 
   static contextTypes = {
     store: React.PropTypes.object,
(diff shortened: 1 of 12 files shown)


rgbkrk (approved):

rgbkrk:
These gifts of flow are so incredible. Thanks for following up after my ref refactor. :bow:

peggyrayzis:
Anytime! 😄 I checked the build log and it looks like there's still some errors stemming from flow-typed. I'm going to work on fixing those today before I start on the Redux code.

rgbkrk:
Now I understand why you said there was more to fix for flow after my PRs. I blindly accepted that the green check mark was showing for flow, wondering if my local setup was just not right. After your next fix we should make `npm run flow` explicitly fail the build as needed.

rgbkrk:
Thank you for taking the flow errors on. We'll make sure it falls on all of us to keep `flow` in check.

peggyrayzis:
I didn't realize that flow wasn't failing the builds either until I saw the ref errors pop up after your PR. Should be an easy fix though.

I'm happy to draft up a quick flow style guide for the team if you would like. Flow is kinda weird when you first start using it & the docs are pretty confusing. 😕  The guide could help prevent new contributors from getting frustrated with flow build errors.

rgbkrk:
Thank you so much.

Issue #962, opened 2016-10-09
Autoexpanding application/json


rgbkrk:
As I use nteract's new `application/json` mime type, I wish that by default the entire tree is fully expanded. I'm guessing the component we use has a way to expand them automatically. They start out collapsed.

![screen shot 2016-10-09 at 11 54 09 am](https://cloud.githubusercontent.com/assets/836375/19221746/1adf86ba-8e17-11e6-9bf3-3cc270cdaf7a.png)

<p align="center">⬇️</p>

![screen shot 2016-10-09 at 11 54 27 am](https://cloud.githubusercontent.com/assets/836375/19221748/27b65e2c-8e17-11e6-9a6f-9dbddc92f36b.png)

Similar to how we handle image metadata:

``` js
metadata: {
  'image/png' : {
    width: 640,
    height: 480
  }
}
```

I'd love to make metadata for application/json expanded with:

``` js
metadata: {
  'application/json' : {
    expanded: true,
  }
}
```

which would expand the entire JSON tree.

---

Steps to take (in mostly any order):
- [x] Accept this metadata on our current JSON transform (on `this.props.metadata`)
- [x] Make a PR to ipython/ipython to accept metadata directly _as well as_ an `expanded` parameter
- [x] Make a PR to jupyter-client with this metadata (for the docs)

/cc @aleury

aleury:
Making the json tree fully expanded by default was actually as easy as passing `shouldExpandNode={() => true}`  as a prop to the `JSONTree` component. 

Do you still think we'll need to use the metadata? Will the metadata prop enable the expansion of the `JsonDisplay` component to be configurable by the user? I'm still very new to how the ipython messaging spec works.

rgbkrk:
If we support metadata, then the user could write:

``` python
IPython.display.JSON(myData, expanded=True)
```

When they explicitly want to see it all and

``` python
IPython.display.JSON(myData, expanded=False)
```

When they don't (maybe a lot of data to take in, all at once.)

rgbkrk:
As for the spec, if we want something nice for kernels and frontends to provide to users, we have to  adapt the message spec in Jupyter (we don't run our own specs here in nteract, we keep collaborating in Jupyter -- several of us are devs there too).

rgbkrk:
> as easy as passing `shouldExpandNode={() => true}` as a prop to the JSONTree component

Yeah I'd make it:

``` js
shouldExpandNode={() => {
[...]

PR #987, opened 2016-10-12
fix(JsonDisplay): Metadata prop is an Immutable Map

aleury:
Related to https://github.com/nteract/nteract/issues/962

I didn't notice in the previous PR for `JsonDisplay` that the `metadata` prop is passed in as an Immutable Map.


rgbkrk (approved):

rgbkrk:
Excellent, I'll merge after CI passes.

PR #994, opened 2016-10-13
chore(ci): disable cache for now

rgbkrk:
We're running into some sort of breaking change with a linting rule. If we can clear the cache we can probably get around it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,7 @@ env:
   - GROUP=flow
 
 cache:
-  pip: true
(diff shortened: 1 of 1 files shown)


rgbkrk:
_Now_ master should pass again.

For further context, our cache needed clearing and [the Travis cache API was erroring](https://twitter.com/rgbkrk/status/786381004973846529).

lgeiger:
Thanks! 🎉

PR #999, opened 2016-10-14
feat(CLI): Add shell commands

lgeiger:
Add a shell commands to start nteract by typing `nteract notebook.ipnb`

### DIFF
--- a/src/main/cli.js
+++ b/src/main/cli.js
@@ -11,7 +11,7 @@ function getStartCommand() {
   }
   const dir = join(process.resourcesPath, '..', subdir);
   const nteractPath = join(dir, 'nteract');
-  const electronPath = join(dir, 'Electron');
+  const electronPath = join(dir, 'electron');
 
   if (existsSync(nteractPath)) {
     return nteractPath;
@@ -34,8 +34,14 @@ function writeExecutable(installDir, script, callback) {
 
 export function installShellCommand() {
   const installDir = join('/usr/local/bin', 'nteract');
+  const installDirUbuntu = join(process.env.HOME, '.local', 'bin', 'nteract');
   const script = `#!/bin/bash\nnohup ${getStartCommand()} "$@" > /dev/null 2>&1 &`;
   writeExecutable(installDir, script, (err) => {
-    console.error(err);
+    if (err) {
+      writeExecutable(installDirUbuntu, script, (error) => {
+        console.error(err);
+        console.error(error);
+      });
+    }
   });
 }
--- a/src/main/cli.js
+++ b/src/main/cli.js
@@ -1,0 +1,41 @@
+import { join } from 'path';
+import { existsSync, writeFile, chmod } from 'fs';
+
+
+function getStartCommand() {
+  let subdir;
+  if (process.platform === 'darwin') {
+    subdir = 'MacOS';
+  } else {
+    subdir = '';
+  }
+  const dir = join(process.resourcesPath, '..', subdir);
+  const nteractPath = join(dir, 'nteract');
+  const electronPath = join(dir, 'Electron');
+
+  if (existsSync(nteractPath)) {
+    return nteractPath;
+  } else if (existsSync(electronPath)) {
+    // Developer install
+    const mainDir = dir.split('/node_modules/')[0];
+    return `${electronPath} ${mainDir}`;
+  }
+  return null;
+}
+
+function writeExecutable(installDir, script, callback) {
+  writeFile(installDir, script, (err) => {
+    if (err) callback(err);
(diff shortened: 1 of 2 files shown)


src/main/cli.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+}
+
+export function installShellCommand() {
+  const installDir = join('/usr/local/bin', 'nteract');
+  const script = `#!/bin/bash\nnohup ${getStartCommand()} "$@" > /dev/null 2>&1 &`;
+  writeExecutable(installDir, script, (err) => {
+    console.error(err);
+  });
+}
```
rgbkrk, line 70: Wow this is great.

src/main/menu.js
```diff
@@ -334,6 +335,24 @@ export const help = {
...
+  role: 'help',
+  submenu: [
+    {
+      label: 'Install Shell Commands',
+      click: () => installShellCommand(),
+    },
+    {
+      label: 'Learn More',
+      click: () => { shell.openExternal('http://github.com/nteract/nteract'); },
+    },
```
rgbkrk: 😄

src/main/cli.js
```diff
@@ -0,0 +1,41 @@
...
+      if (error) callback(error);
+    });
+  });
+}
+
+export function installShellCommand() {
+  const installDir = join('/usr/local/bin', 'nteract');
+  const script = `#!/bin/bash\nnohup ${getStartCommand()} "$@" > /dev/null 2>&1 &`;
+  writeExecutable(installDir, script, (err) => {
+    console.error(err);
```
lgeiger: How can I add a notification?
I tried

``` javascript
    const state = window.store.getState();
    const notificationSystem = state.app.get('notificationSystem');
    if (err) {
      notificationSystem.addNotification({
        title: 'Shell Commands not Installed',
        message: err.message,
        level: 'error',
      });
    } else {
      notificationSystem.addNotification({
        title: 'Shell Commands Installed',
        level: 'success',
      });
    }
```

src/main/menu.js
```diff
@@ -391,7 +417,11 @@ export function generateDefaultTemplate() {
   template.push(edit);
   template.push(view);
   template.push(window);
-  template.push(help);
+  if (process.platform === 'linux') {
+    template.push(helpLinux);
+  } else {
+    template.push(help);
+  }
```
rgbkrk: Well done with the cross platform setup here. How did you know which belonged to which menu?

rgbkrk (approved):

src/main/menu.js
lgeiger: Just tried it on linux and it didn't work ;)

rgbkrk:
While you don't have to do it in this PR, it would be great to create a new file if a file doesn't exist.

lgeiger:
I think that's what `fs.writeFile` does, doesn't it?

rgbkrk:
We read the file before we write the file. We'd need a new handler in launch (part of main).

lgeiger:
Oh no on linux I don't have permissions to write to `/usr/local/bin`

lgeiger:
At least on ubuntu we can use `~/.local/bin/`

lgeiger:
Now it should work on Linux as well.

src/main/cli.js
captainsafia: You cannot use the notification system in the main process since it's a React component. I think using a [dialog](https://github.com/electron/electron/blob/master/docs/api/dialog.md) should do the trick though.

lgeiger:
@jdetle Can you test it on your Linux machine? I'm experiencing a very slow startup using `nteract` but I don't know if it's a problem of my VM.

lgeiger:
To workaround the permission issue we could add the shell script to a custom location and add it to the `PATH`. I don't know whats better.

rgbkrk:
What does Atom do?

lgeiger:
They use https://github.com/atom/node-runas

rgbkrk:
Ok, cool. That's a good solution for how we run as admin.

As for setting the location, that's what I'm wondering about Atom. Where do they install `atom` and `apm`?

jdetle:
I was having trouble figuring out how it was supposed to work. Simply running `nteract blah.ipynb` didn't work (if it was supposed to.) I'm heading up to SF (and will be away from my linux machine) but I can continue working on it tomorrow night.

lgeiger:
You need to run `Install Shell Commands` from the menu before using the commandline interface. Thanks for trying it.

jdetle:
I assumed there was some step I was missing, thanks. I love me a good CLI.

Issue #1000, opened 2016-10-14
Add example notebooks to File menu


captainsafia:
After the podcast interview, @shanselman mentioned that it would be a good idea to include the example notebooks in the File > Open menu. We heard similar feedback at the NumFOCUS Summit.

Tagging this as new-contributor-friendly with mentorship from me available.

alexandercbooth:
Hi @captainsafia

I just submitted a PR for this (#1001). Can you check it over when you get a chance and let me know what you think? For some reason, the notebooks don't always open 😕. 

Thanks!

PR #1001, opened 2016-10-14
Problem: issue #1000 addressed

alexandercbooth:
Solution: added example notebooks to the file > open > example notebooks

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -52,27 +52,127 @@ export const fileSubMenus = {
   },
   open: {
     label: '&Open',
-    click: () => {
-      const opts = {
-        title: 'Open a notebook',
-        filters: [
-          { name: 'Notebooks', extensions: ['ipynb'] },
-        ],
-        properties: [
-          'openFile',
-        ],
-      };
-      if (process.cwd() === '/') {
-        opts.defaultPath = app.getPath('home');
-      }
+    submenu: [
+      {
+        label: '&Open',
+        click: () => {
+          const opts = {
+            title: 'Open a notebook',
+            filters: [
+              { name: 'Notebooks', extensions: ['ipynb'] },
+            ],
+            properties: [
+              'openFile',
+            ],
+          };
+          if (process.cwd() === '/') {
+            opts.defaultPath = app.getPath('home');
+          }
 
-      dialog.showOpenDialog(opts, (fname) => {
-        if (fname) {
-          launch(fname[0]);
-        }
-      });
-    },
-    accelerator: 'CmdOrCtrl+O',
+          dialog.showOpenDialog(opts, (fname) => {
+            if (fname) {
+              launch(fname[0]);
+            }
+          });
+        },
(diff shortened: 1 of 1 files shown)


src/main/menu.js
```diff
@@ -52,27 +52,127 @@ export const fileSubMenus = {
...
+              const opts = {
+                title: 'Open a notebook',
+                filters: [
+                  { name: 'Notebooks', extensions: ['ipynb'] },
+                ],
+                properties: [
+                  'openFile',
+                ],
+              };
+              launch('example-notebooks/geojson.ipynb');
```
rgbkrk: You'll want to use `__dirname` to resolve the absolute path to these notebooks.

rgbkrk:
``` js
const path = require('path');
const exampleNotebooksDirectory = path.join(__dirname, '..', '..', 'example-notebooks');
```

alexandercbooth: Thanks Kyle! I figured it was something with resolving the path. 
Since it looks like we already import the path at the top, this is what I have at the moment:

``` jsx
const exampleNotebooksDirectory = path.join(__dirname, '..', '..', 'example-notebooks');
```

at the top. And then changing the launch of the notebooks to:

``` jsx
launch(`${exampleNotebooksDirectory}/geojson.ipynb`);
```

If you're comfortable with this I'll make the changes to the rest and commit. 
Thanks!

alexandercbooth: It loads up and the linter likes it 😸

rgbkrk: Awesome!

One tweak, make sure to use `path.join` when creating the path for the notebook:

``` js
launch(path.join(exampleNotebooksDirectory, 'geojson.ipynb'));
```

This allows the notebooks to load on Windows too.

alexandercbooth: Ah good catch! Thanks for the patience with someone so new to react. I definitely wouldn't want to discourage Windows users (even if it makes me cry on the inside 😉 ).

Thanks Kyle!

src/main/menu.js
```diff
@@ -6,6 +6,8 @@ import { installShellCommand } from './cli';
 
 const kernelspecs = require('kernelspecs');
 
+const exampleNotebooksDirectory = path.join(__dirname, '..', '..', 'example-notebooks');
```
rgbkrk, line 9: I just realized we should probably make this:

``` js
const exampleNotebooksDirectory = path.resolve(path.join(__dirname, '..', '..', 'example-notebooks'));
```

rgbkrk (approved):

rgbkrk:
Thanks!

I left one note that you're free to make a follow on PR for (`path.resolve` of the dir).

src/main/menu.js
alexandercbooth: OK, I'll fix it up and submit a new PR. Thanks!

alexandercbooth:
My pleasure, will do!

PR #1004, opened 2016-10-14
chore(ExampleNotebooks): resolve path in exampleNotebooksDirectory

alexandercbooth:
As per conversation in #1001, changed how the exampleNotebooksDirectory is defined.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -6,8 +6,10 @@ import { installShellCommand } from './cli';
 
 const kernelspecs = require('kernelspecs');
 
-const exampleNotebooksDirectory = path.resolve(path.join(__dirname, '..', '..', 'example-notebooks'));
-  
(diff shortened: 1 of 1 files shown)


alexandercbooth:
Just realized there are some linting issues. Fixing up now, sorry!

alexandercbooth:
OK, changed to this: 

``` jsx
const exampleNotebooksDirectory = path.resolve(path.join(__dirname, '..',
    '..', 'example-notebooks'));
```

Looks kind of ugly, but no screaming from the linter. Is this OK?

rgbkrk:
'Tis fine. Style-wise, you may like this better:

``` js
const exampleNotebooksDirectory = path.resolve(
  path.join(__dirname, '..', '..', 'example-notebooks')
);
```

alexandercbooth:
I like the looks of that much better, thanks!

PR #1007, opened 2016-10-14
feat(CLI): Specify initial kernel via command line

lgeiger:
Example: `nteract --kernel javascript` or `nteract -k javascript` starts a new notebook with the javascript kernel.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -18,15 +18,12 @@ const log = require('electron-log');
 
 const kernelspecs = require('kernelspecs');
 
-const version = require('../../package.json').version;
-
-log.info('args', process.argv);
-
-const sliceAt = process.argv[0].match('nteract') ? 1 : 2;
-
 const argv = require('yargs')
-  .version(version)
-  .parse(process.argv.slice(sliceAt));
+  .alias('k', 'kernel')
+  .default('kernel', 'python3')
+  .argv;
+
+log.info('args', argv);
 
 const notebooks = argv._.filter(x => /(.ipynb)$/.test(x));
 
@@ -119,11 +116,10 @@ openFile$
     if (notebooks.length <= 0 && buffer.length <= 0) {
       log.info('launching an empty notebook by default');
       kernelSpecsPromise.then(specs => {
-        const defaultKernel = 'python3';
-        let kernel = defaultKernel;
+        let kernel;
 
-        if ('python3' in specs) {
-          kernel = 'python3';
+        if (argv.kernel in specs) {
+          kernel = argv.kernel;
         } else if ('python2' in specs) {
           kernel = 'python2';
         } else {


rgbkrk:
Could we do `--kernel kernelName` instead? Otherwise we won't be able to have other arguments later.

lgeiger:
I updated it.

It won't work with `npm run start --kernel python2` because somehow it doesn't recognize the `kernel` flag.

src/main/index.js
```diff
@@ -18,15 +18,12 @@ const log = require('electron-log');
...
 const kernelspecs = require('kernelspecs');
 
-const version = require('../../package.json').version;
-
-log.info('args', process.argv);
-
-const sliceAt = process.argv[0].match('nteract') ? 1 : 2;
-
 const argv = require('yargs')
-  .version(version)
```
lgeiger, line 28: I removed the version since it wasn't working and I couldn't get it to work

src/main/index.js
```diff
@@ -18,15 +18,12 @@ const log = require('electron-log');
 
 const kernelspecs = require('kernelspecs');
 
-const version = require('../../package.json').version;
-
-log.info('args', process.argv);
-
-const sliceAt = process.argv[0].match('nteract') ? 1 : 2;
```
lgeiger, line 25: Since we filter for `.ipynb` files this isn't necessary anymore.

src/main/index.js
```diff
@@ -18,15 +18,12 @@ const log = require('electron-log');
...
-
 const argv = require('yargs')
-  .version(version)
-  .parse(process.argv.slice(sliceAt));
+  .alias('k', 'kernel')
+  .default('kernel', 'python3')
+  .argv;
```
rgbkrk, line 24: should this end with `.parse(process.argv.slice(sliceAt))`?

lgeiger: > Since we filter for `.ipynb` files this isn't necessary anymore.

rgbkrk: Not even the parse?

lgeiger: `yargs` docs say:

> If yargs is executed in an environment that embeds node and there’s no script name (e.g. Electron or nw.js), it will ignore the first parameter since it expects it to be the script name. In order to override this behavior, use `.parse(process.argv.slice(1))` instead of `.argv` and the first parameter won’t be ignored.

I don't know why we would need this.

rgbkrk: Ah, ok.

lgeiger:
I understand it now.

One has to use `npm run start -- --kernel javascript`  to make it work.

rgbkrk:
> I understand it now.
> 
> One has to use npm run start -- --kernel javascript to make it work.

Oh my gosh, why didn't I think of that?

lgeiger:
@rgbkrk That's what I thought as well 😄

PR #1023, opened 2016-10-15
feat(conda): detect conda environments with ipykernels

rgbkrk:
This extracts the conda kernel detection out of #847 in preparation for providing it to the menu and to all launched notebooks.

Usage example:

``` js
> var condaKernel$ = require('./app/build/main/kernels').condaKernelsObservable();
undefined
> var sub = condaKernel$.subscribe(x => console.log(x));
undefined
> { 'conda-env-ipy-py':
   { display_name: 'Python [conda env:ipy]',
     argv:
      [ '/usr/local/opt/python/Frameworks/Python.framework/Versions/2.7/envs/ipy/bin/python',
        '-m',
        'ipykernel',
        '-f',
        '{connection_file}' ],
     language: 'python' },
  'conda-env-microbots-py':
   { display_name: 'Python [conda env:microbots]',
     argv:
      [ '/usr/local/opt/python/Frameworks/Python.framework/Versions/2.7/envs/microbots/bin/python',
        '-m',
        'ipykernel',
        '-f',
        '{connection_file}' ],
     language: 'python' },
  'conda-env-root-py':
   { display_name: 'Python [conda env:root]',
     argv:
      [ '/usr/local/opt/python/Frameworks/Python.framework/Versions/2.7/bin/python',
        '-m',
        'ipykernel',
        '-f',
        '{connection_file}' ],
     language: 'python' } }
```

/cc @nhdaly

### DIFF
--- a/src/main/kernels.js
+++ b/src/main/kernels.js
@@ -25,9 +25,6 @@ export function condaInfoObservable() {
     .map(info => JSON.parse(info));
 }
 
-// var condak = require('./src/notebook/epics/conda-kernel-provider-epic');
-// var envy = condak.condaEnvsObservable(condak.condaInfoObservable());
-
 export function condaEnvsObservable(condaInfo$) {
   return condaInfo$.map(info => {
     const envs = info.envs.map(env => ({ name: path.basename(env), prefix: env }));
--- a/src/main/kernels.js
+++ b/src/main/kernels.js
@@ -1,0 +1,72 @@
+import Rx from 'rxjs/Rx';
+
+import {
+  spawn,
+} from 'spawn-rx';
+
+const path = require('path');
+
+// TODO: Check for sys.prefix/share/jupyter/kernels/*/kernel.json
+/**
+ * ipyKernelTryObservable check for existence of ipykernel in an env
+ * @param  {Object} env [description]
+ * @return {[type]}     [description]
+ */
+export function ipyKernelTryObservable(env) {
+  const executable = path.join(env.prefix, 'bin', 'python');
+  return spawn(executable, ['-m', 'ipykernel', '--version'], { split: true })
+    .filter(x => x.source && x.source === 'stdout')
+    .mapTo(env)
+    .catch(err => Rx.Observable.empty());
+}
+
+export function condaInfoObservable() {
+  return spawn('conda', ['info', '--json'])
+    .map(info => JSON.parse(info));
+}
+
+// var condak = require('./src/notebook/epics/conda-kernel-provider-epic');
+// var envy = condak.condaEnvsObservable(condak.condaInfoObservable());
+
+export function condaEnvsObservable(condaInfo$) {
+  return condaInfo$.map(info => {
+    const envs = info.envs.map(env => ({ name: path.basename(env), prefix: env }));
+    envs.push({ name: 'root', prefix: info.root_prefix });
+    return envs;
+  })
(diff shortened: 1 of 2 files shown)


rgbkrk:
When this gets used in the main process, we'll want to make sure we cache the result (share/publish - I'm spacing on what exactly we have to do to not replay the observable). There are two intended receivers: knowing if the app is still loading - our "keep running the logo" part, as well as the main menu.

src/main/kernels.js
```diff
@@ -0,0 +1,72 @@
...
+    .mapTo(env)
+    .catch(err => Rx.Observable.empty());
+}
+
+export function condaInfoObservable() {
+  return spawn('conda', ['info', '--json'])
+    .map(info => JSON.parse(info));
+}
+
+// var condak = require('./src/notebook/epics/conda-kernel-provider-epic');
```
jdetle: Is this meant to still be here?

rgbkrk: Nah, I can delete that.

dhirschfeld:
Hmm, just seeing this now and want to make the point that not everyone wants all their conda environments showing up as kernels so I hope it's configurable! I dislike that behaviour so much that I've resorted to uninstalling `nb_conda_kernels`.

Either way it should probably respect the kernel whitelist in the jupyter notebook config (something which `nb_conda_kernels` didn't do):

```
# Whitelist of allowed kernel names.
#
# By default, all installed kernels are allowed.
# c.KernelSpecManager.whitelist = set()
```

rgbkrk:
> not everyone wants all their conda environments showing up as kernels so I hope it's configurable

Thank you for raising this up! I've been erring on the side of making things "just work" out of the box for users, so we have less users complaining about python not working (i.e. not having a python kernelspec installed). We don't read any jupyter notebook configuration, as it's implemented in Python with full allowance for Python

Would it be ok for us to detect the root conda environment and provide that? I did feel a bit scummy implementing this - I'd rather people install kernels directly or that conda provides a means of providing kernels without overriding the notebook server.

/cc @minrk for thoughts on conda + jupyter

rgbkrk:
Note: this code that went in provides the ability to locate the kernels and is not enabled. I'll think on configuration for this if we enable it. Open to ideas and PRs.

dhirschfeld:
Most installs I only have  miniconda in the root environment so that wouldn't work for me, however I do explicitly install any kernelspecs I want to actually show up. I like how you have to be explicit and actually install a kernelspec for it to show up - no magic involved. I can see how that might prompt a lot of noise from your average user not understanding why it doesn't just work out of the box though...

This would be fine if there were a configuration where I could easily specify which of the environments I wanted to see as options for a new kernel. I'm thinking of something like a multi-select which could be populated by reading all kernelspecs and conda envs. If any kernelspecs were installed then only those would initially be selected as options for a new kernel. If no kernelspecs were found then all valid conda envs could be initially selected as options for a new kernel and the user could then manually deselect any they didn't want showing up.

The default kernel for an nteract session could then be the one in the environment (first in the path) where it was started, falling back to either root or some globally specified default.

Does that sound reasonable/sensible?

minrk:
You could separate the 'automatically discovered kernels from conda' as a distinct list (sub or sibling), so it looks something like:
- Kernels (+)
  - Python 3
  - Julia
  - conda envs (+)
  - virtualenvs (+)

That way, any auto-discovered kernels are only visible on request, and clearly distinct from properly registered kernels.

PR #1028, opened 2016-10-16
Better cli interface

lgeiger:
This is heavily inspired by Atom.

OS X works, Linux should as well.
Run "Install Shell Comands" from the nteract menu then try `nteract --help` in a new terminal.

Windows is a bit tricky since it doesn't has a `nohup` command for detaching the process from the terminal. Atom uses it's bundled node to start Atom from javascript.

Maybe there is a way of using Electrons node from the command line. I'll try this approach on windows. Lets see if it works.


lgeiger:
Wuhu it works on Windows 🎉 

Tested with cmd.exe, git bash and powershell

Untitled.ipynb
```diff
@@ -0,0 +1,22 @@
+{
```
rgbkrk: I'm guessing you didn't mean to commit this file.

src/main/cli.js
```diff
@@ -1,102 +1,109 @@
...
-    if (err) {
-      callback(err);
-    } else {
-      chmod(installDir, '755', (error) => {
-        callback(error);
-      });
-    }
+const createNewSymlinkObservable = (target, path) =>
+  Rx.Observable.create(observer => {
+    fs.symlink(target, path, (error) => {
```
rgbkrk: I'm beginning to think we should carve out `fs-observable` as a library to share across these.

Untitled.ipynb
lgeiger: Yep. There are a few files here I need to delete. Used git to sync with my windows VM.

src/main/cli.js
```diff
@@ -1,102 +1,109 @@
...
+const setWinPathObservable = (exe, rootDir, binDir) => {
+  // Remove duplicates because SETX throws a error if path is to long
+  const env = process.env.PATH.split(';')
+    .filter(item => !/nteract/.test(item))
+    .filter((item, index, array) => array.indexOf(item) === index);
+  env.push(binDir);
+  const envPath = env.join(';');
+  return spawn('SETX', ['PATH', `${envPath}`])
+    .switchMap(() => spawn('SETX', ['NTERACT_EXE', exe]))
+    .switchMap(() => spawn('SETX', ['NTERACT_DIR', rootDir]));
```
rgbkrk: Depending on what this is for, the behavior you want may be `mergeMap`. What's `SETX` doing?

src/main/cli.js
lgeiger: I was thinking about a new utility folder next to `notebook` and `main` where we could move them

src/main/cli.js
rgbkrk: Ah, are you wanting to `SETX` each of these environment variables in sequence or in parallel?

I would probably do:

``` js
return Rx.Observable.merge(
  spawn('SETX', ['PATH', `${envPath}`]),
  spawn('SETX', ['NTERACT_EXE', exe]),
  spawn('SETX', ['NTERACT_DIR', rootDir])
);
```

If I want to run all of them in parallel.

lgeiger: I'll look at mergeMap 👍 

`SETX` sets a global environment variable (technically a registry key). I use environment variables to make it work for a dev install as well. Otherwise we would need do some guessing based on the default install directories.

src/main/cli.js
rgbkrk: Yeah that sounds good, we keep needing these across the project. I was hoping to use something for checking if a file exists when doing startup as well (so we can create a new file).

src/main/cli.js
```diff
@@ -1,102 +1,109 @@
...
+  unlinkObservable(path)
+    .switchMap(() => createNewSymlinkObservable(target, path));
+
+const installShellCommandsObservable = (exe, rootDir, binDir) => {
+  if (process.platform === 'win32') {
+    return setWinPathObservable(exe, rootDir, binDir);
+  }
+  const envFile = join(binDir, 'nteract-env');
+  return writeFileObservable(envFile, `NTERACT_EXE="${exe}"\nNTERACT_DIR="${rootDir}"`)
+    .switchMap(() => {
```
rgbkrk: Instead of `switchMap`, I'd use `flatMap`/`mergeMap`. [`switchMap`](https://github.com/btroncone/learn-rxjs/blob/master/operators/transformation/switchmap.md) has a particular meaning with new items in the sequence, running the inner observable here.

src/main/cli.js
lgeiger: I'm using `fs.existSync` here because it seems to be fast. It was undeprecated in https://github.com/nodejs/node/pull/8364

src/main/cli.js
```diff
@@ -1,102 +1,109 @@
...
-}
+
+  installShellCommandsObservable(exe, rootDir, binDir)
+    .subscribe(
+      () => {},
+      (err) => dialog.showErrorBox('Could not write shell script.', err.message),
+      () => dialog.showErrorBox(
+        'Command installed.',
+        'The shell command "nteract" is installed.\nOpen notebooks with "nteract notebook.ipynb".'
+        )
```
rgbkrk: This is awesome, well done.

lgeiger: Thanks!
I think I'll use `dialog.messageBox` here. It doesn't make a difference on OS X but on windows there is a red error sign.

src/main/cli.js
rgbkrk: `existsSync` is the one call that is faster than the async version. :smile: Don't worry, I skimmed right on past that one when I saw it - it has a history. :wink: I'm happy about the undeprecation.

src/notebook/epics/config.js
```diff
@@ -52,8 +29,8 @@ export const loadConfigEpic = actions =>
   actions.ofType(LOAD_CONFIG)
     .switchMap(action =>
       readFileObservable(CONFIG_FILE_PATH)
-        .map(JSON.parse)
-        .map(configLoaded)
+        .map(({ filename, data }) => JSON.parse(data))
+        .map(({ filename, data }) => configLoaded(data))
```
rgbkrk: The prior stream is returning parsed JSON data, so you'll want to turn this into:

``` js
.map(configLoaded)
```

src/utils/fs.js
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+          observer.next({ path });
+          observer.complete();
+        }
+      });
+    } else {
+      observer.next({ path });
+      observer.complete();
+    }
+  });
```
rgbkrk, line 19: As for testing these, you can write a sinon spy for each of the functions on `fs`, including `existsSync` and `unlink`.

src/utils/fs.js
```diff
@@ -0,0 +1,60 @@
...
+        if (error) {
+          observer.error(error);
+        } else {
+          observer.next({ path });
+          observer.complete();
+        }
+      });
+    } else {
+      observer.next({ path });
+      observer.complete();
```
rgbkrk, line 17: If the file doesn't exist, we regard it as already gone/deleted?

src/utils/fs.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+  });
+
+export const writeFileObservable = (filename, data, ...args) =>
+  Rx.Observable.create(observer => {
+    fs.writeFile(filename, data, ...args, error => {
+      if (error) {
+        observer.error(error);
+      } else {
+        observer.next({ filename, data });
```
rgbkrk: For `writeFileObservable` we don't need to send the data on (or so I think), since we only have to write it to the file.

rgbkrk: Ok, just double checked the other usages of this and we tend to throw away both args. Returning the filename isn't bad by any means but we don't need to bring this potentially large `data` around (like with notebooks).

src/utils/fs.js
lgeiger: That's the path we are checking for. Maybe we shouldn't return anything at all.

src/utils/fs.js
lgeiger: I'll check what `fs` does.

src/utils/fs.js
lgeiger: I can add test tomorrow

src/utils/fs.js
rgbkrk: 🤔 Hmmmm, well we can always see how we use this in practice.

lgeiger:
I think it's fine to return the arguments even if nothing changed. 

For example `fs.readFile` only returns the data. On the other hand `readFileObservable` returns the filepath (which was passed in as a argument) and the data. This is used [here](https://github.com/nteract/nteract/blob/master/src/notebook/epics/loading.js#L76-L77).

We could either change it and be consistent with `fs` or leave it and decide to allways return everything (which would require no change).

src/utils/fs.js
lgeiger: I tried it with sion spies but couldn't get it to work.

rgbkrk:
Time to have everyone give this a spin on master.

rgbkrk:
Since I wanted to double check after merging, saving, loading, and config are working perfectly.

rgbkrk:
screen shot 2016-10-18 at 8 09 10 am

```
$ nteract --help
open: unrecognized option `--help'
Usage: open [-e] [-t] [-f] [-W] [-R] [-n] [-g] [-h] [-b <bundle identifier>] [-a <application>] [filenames] [--args arguments]
Help: Open opens files from a shell.
      By default, opens each file using the default application for that file.
      If the file is in the form of a URL, the file will be opened as a URL.
Options:
      -a                Opens with the specified application.
      -b                Opens with the specified application bundle identifier.
      -e                Opens with TextEdit.
      -t                Opens with default text editor.
      -f                Reads input from standard input and opens with TextEdit.
      -F  --fresh       Launches the app fresh, that is, without restoring windows. Saved persistent state is lost, excluding Untitled documents.
      -R, --reveal      Selects in the Finder instead of opening.
      -W, --wait-apps   Blocks until the used applications are closed (even if they were already running).
          --args        All remaining arguments are passed in argv to the application's main() function instead of opened.
      -n, --new         Open a new instance of the application even if one is already running.
      -j, --hide        Launches the app hidden.
      -g, --background  Does not bring the application to the foreground.
      -h, --header      Searches header file locations for headers matching the given filenames, and opens them.
```

lgeiger:
Great! Thanks for verifying 👍

lgeiger:
Hmm this is strange. We don't even use `open` in the shell script.

Could you run `wich nteract`?

rgbkrk:
Looks like `--help` ends up being passed through to `open` on Mac instead of providing help info.

rgbkrk:
ohhhhh that's my fault. I aliased `nteract`

rgbkrk:
```
$ nteract --help
Usage: nteract <notebooks> [options]

Options:
  --verbose      Display debug information
  -k, --kernel   Launch a kernel  [default: "python3"]
  -v, --version  Show version number  [boolean]
  -h, --help     Show help  [boolean]

Examples:
  nteract notebook1.ipynb notebook2.ipynb  Open notebooks
  nteract --kernel javascript              Launch a kernel
```

rgbkrk:
Looking great!

lgeiger:
Excellent! 🎉

rgbkrk:
Now the bug reports will come in for when people use this CLI to create new notebooks. :smile:

screen shot 2016-10-18 at 8 15 20 am

We need to do some detection on the main process and an event on the renderer process to handle creating new named notebooks.

rgbkrk:
One thing we probably need to make sure of that I'm not seeing here -- each time you launch `nteract` a new instance is created. On the Mac, this should default to the already created nteract main process.

lgeiger:
Didn't thought of that. We check whether the commandline option has a `.ipynb` extension but not if it exists.

lgeiger:
What would the behavior be if we use `open -a -n nteract_exe` on OS X instead of `nohup nteract_exe &` in the shell script?

That's what [Atom does](https://github.com/atom/atom/blob/master/atom.sh#L88).

rgbkrk:
We probably want it without the `-n`. It does look like they're opening new instances then. It usually seems like they defer to a previous window when I run `atom .` several times in a row.

lgeiger:
OK if we use `open` without `-n` and nteract is running it would only bring nteract to the foreground without opening a new window. Is this the behavior we want?

rgbkrk:
Oh I see, we would want a new window if the notebook wasn't already opened.

Issue #996, opened 2016-10-13
Add user guide for nteract


captainsafia:
This is a follow-up on the question raised in #995.

nteract has an interface that's very different from other interactive notebook applications. We need to create a User Guide that outlines.
- [x] What nteract is
- [x] How to create/open/save files
- [x] How to run, move, and add cells
- [x] How to use the cell dropdown

This guide can reside in a USER_GUIDE.md file at the top directory of this repository.

cc: @willingc

rgbkrk:
Pertinent easy to start part for this issue is updating the [user installation of nteract in the readme](https://github.com/nteract/nteract#installation---users).

alexandercbooth:
I'm happy to work on this. I'll submit a [WIP] PR for some feedback from everyone later tonight.

rgbkrk:
Awesome, looking forward to it.

captainsafia:
This was closed in #1032.

Issue #995, opened 2016-10-13
how to use nteract


hainm:
hi

sounds dummy question but how users would use nteract?

quote from README
`Edit code cells, write markdown, visualize!`

> Double click a .ipynb on the desktop, use Spotlight on the Mac. It Just Works™

What do you mean 'It just Works'? 

thanks.

hainm:
``` bash
npm run start some_notebook.ipynb
```

seems do the trick but took me quite long time to realize that. :D

jdetle:
Hi @hainm! Thanks for trying it out. You bring up a good point, I think this might be meant for releases rather than developer installations, and perhaps we should make that clearer.

rgbkrk:
The "just works" line is likely aimed at how you can treat jupyter notebooks as native files using finder, etc. (For the built release).

This brings up an even more pressing issue - how does a complete beginner to jupyter get introduced at the beginning of the readme? You're a well practiced user of Jupyter @hainm so if we confused you, we confused way more people.

/cc @willingc

captainsafia:
I think the blog post on our Medium is a solid run through.

willingc:
@rgbkrk Perhaps:

It wraps up the best of the web based Jupyter notebook and embeds it as a desktop application. Now, you can open notebooks natively on your system. Double click on an`.ipynb` file to open and begin using the notebook.  It Just Works™

I found the Spotlight reference confuzzling.

captainsafia:
> It wraps up the best of the web based Jupyter notebook and embeds it as a desktop application. Now, you can open notebooks natively on your system. Double click on an.ipynb file to open and begin using the notebook. It Just Works™

That sounds solid. I also think this issue falls under the general need for a User Guide for nteract that outlines things like opening/savings, using the cell dropdown to change cells, adding new cells, etc. I think that can be added as a new-contributor-friendly issue for people who are willing to do some self-testing of that app.

captainsafia:
This is closed via the addition of a user guide in #1032. Let us know if you have any other questions about using the app, @hainm!

hainm:
cool. thanks.

PR #1032, opened 2016-10-18
docs(userGuide): add user guide

alexandercbooth:
Add first draft of user guide in reference to #996. Comments/ideas please!

Just realized I forgot to add [WIP] in title, sorry!

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1,0 +1,75 @@
+# nteract
+
+nteract is a literate coding environment that supports Python, R, JavaScript and other Jupyter kernels. It wraps up the best of the web based Jupyter notebook and embeds it as a desktop application that allows you to open notebooks natively on your system. Double click a .ipynb on the desktop, use Spotlight on the Mac. It Just Works™
+
+
+## Installation
+To get started, download the latest [release](https://github.com/nteract/nteract/releases) for your OS.
+
+## Authoring Notebook Files (.ipynb)
+
+### Create a New Notebook
+A new notebook can be created by accessing the menu,
+
+  File<br/>
+        &nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;New<br/>
+        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;\<Language Kernel\> (e.g. Python 3, R, Julia etc.)
+
+### Opening a Notebook
+There are several ways to open a notebook in nteract:
+
+1. From the menu:
+
+  File<br/>
+        &nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;Open<br/>
+        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;Open<br/>
+
+2. Double-click a notebook file :tada::  
+
+3. From the Command Line (assuming you have shell commands installed), run `nteract notebook.ipynb`:
+
+*Keyboard shortcut ⌘O on macOS and Ctrl-O on Windows/Linux*
+
+### Saving a Notebook
+
+A notebook can be saved in the following ways:
+
+File<br/>
+      &nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;Save
+
+*Keyboard shortcut: ⌘S on macOS and Ctrl-S on Windows/Linux*
+
+File<br/>
+      &nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;Save As
+
+*Keyboard shortcut: ⇧⌘S on macOS and Shift-Ctrl-S on Windows/Linux*
+
+## Notebook Cells
+
+### Adding a Cell
+
+#### Code Cells
+A code cell can be created by accessing the menu,
+
+Edit<br/>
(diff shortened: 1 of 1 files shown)


alexandercbooth:
I noticed when writing out some directions for this that there is not a way to create a markdown cell from the menu/keyboard shortcut. I would be happy to add this as a feature if you're interested (keyboard shortcuts for markdown is something I use all the time in jupyter). What do you think?

rgbkrk:
> I noticed when writing out some directions for this that there is not a way to create a markdown cell from the menu/keyboard shortcut. I would be happy to add this as a feature if you're interested (keyboard shortcuts for markdown is something I use all the time in jupyter). What do you think?

Sure, what's the keybinding?

USER_GUIDE.md
```diff
@@ -0,0 +1,75 @@
+# nteract
+
+nteract is a literate coding environment that supports Python, R, JavaScript and other Jupyter kernels. It wraps up the best of the web based Jupyter notebook and embeds it as a desktop application that allows you to open notebooks natively on your system. Double click a .ipynb on the desktop, use Spotlight on the Mac. It Just Works™
```
rgbkrk: Incorporate the feedback from @willingc in https://github.com/nteract/nteract/issues/995#issuecomment-253535469 here too:

> It wraps up the best of the web based Jupyter notebook and embeds it as a desktop application. Now, you can open notebooks natively on your system. Double click on an.ipynb file to open and begin using the notebook. It Just Works™

USER_GUIDE.md
```diff
@@ -0,0 +1,75 @@
...
+To get started, download the latest [release](https://github.com/nteract/nteract/releases) for your OS.
+
+## Authoring Notebook Files (.ipynb)
+
+### Create a New Notebook
+A new notebook can be created by accessing the menu,
+
+  File<br/>
+        &nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;New<br/>
+        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;\<Language Kernel\> (e.g. Python 3, R, Julia etc.)
```
rgbkrk: You may enjoy using backticks (like a code block) and using literal spaces in markdown instead of `&nbsp;`.

USER_GUIDE.md
```diff
@@ -0,0 +1,75 @@
...
+### Opening a Notebook
+There are several ways to open a notebook in nteract:
+
+1. From the menu:
+
+  File<br/>
+        &nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;Open<br/>
+        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;⮑&nbsp;&nbsp;Open<br/>
+
+2. Double-click a notebook file :tada::
```
rgbkrk: We probably need a caveat here since I think we've only registered the filetypes on OS X.

rgbkrk (approved):

rgbkrk:
Excellent! I pointed out a few things if you have time to address them. Either way, this is great and moves us forward!

USER_GUIDE.md
alexandercbooth: Great point from @willingc, spotlight probably won't register with many users. 

Would it also be helpful to link to all the current available Jupyter kernels here (like this [Jupyter kernels](https://github.com/ipython/ipython/wiki/IPython-kernels-for-other-languages))?

USER_GUIDE.md
alexandercbooth: Haha, true,  that would be more efficient 😸

USER_GUIDE.md
alexandercbooth: OK, I will make sure to note that.

USER_GUIDE.md
rgbkrk: Yeah that would be good.

alexandercbooth:
Thanks, thrilled to help!

The key binding I registered for new markdown cell is shift CmdorCtrl M in #1034

USER_GUIDE.md
alexandercbooth: Just to double check, something like this instead?

```
  File
    ⮑  Open
          ⮑  Open
```

alexandercbooth:
Also, is there a specific name for what I called the "hover menu"?

USER_GUIDE.md
rgbkrk: yep!

rgbkrk:
> is there a specific name for what I called the "hover menu"?

I think the code says "CellCreator". Hover menu is way better for an end user though, so I think you've chosen well. We can always adapt it again later if we think of something better.

USER_GUIDE.md
```diff
@@ -0,0 +1,92 @@
...
+          ⮑  <Language Kernel> (e.g. Python 3, R, Julia etc.)
+```
+### Opening a Notebook
+There are several ways to open a notebook in nteract:
+
+1. From the menu:
+```
+  File
+    ⮑  Open
+          ⮑  Open
```
rgbkrk, line 25: This makes me want to move example notebooks to a separate `File -> "Open Example Notebooks" so we don't have to nest the default.

alexandercbooth: That's a good idea, now that you mention it it does feel a little clunky going through two drop down menus. I'd be happy to change it (in nteract menu & here) if you'd like

rgbkrk: That would be great. I'll go ahead and merge this as well since it's solid progress forward.

alexandercbooth: Awesome, thanks!

PR #1034, opened 2016-10-18
feat(newTextCell): add new markdown cell to menu and create keyboard shortcut

alexandercbooth:
Added new markdown cell feature with keybinding _Shift CmdOrCtrl M_ and added test for it.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -264,6 +264,11 @@ export const edit = {
       click: createSender('menu:new-code-cell'),
     },
     {
+      label: 'New Text Cell',
+      accelerator: 'CmdOrCtrl+Shift+M',
+      click: createSender('menu:new-text-cell'),
+    },
+    {
       label: 'Copy Cell',
       accelerator: 'CmdOrCtrl+Shift+C',
       click: createSender('menu:copy-cell'),
--- a/test/renderer/menu-spec.js
+++ b/test/renderer/menu-spec.js
@@ -31,6 +31,22 @@ describe('menu', () => {
     });
   });
 
+  describe('dispatchCreateTextCellAfter', () => {
+    it('dispatches a CREATE_TEXT_CELL_AFTER action', () => {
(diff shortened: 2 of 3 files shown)


src/main/menu.js
```diff
@@ -264,6 +264,11 @@ export const edit = {
       click: createSender('menu:new-code-cell'),
     },
     {
+      label: 'New Text Cell',
+      accelerator: 'CmdOrCtrl+Shift+M',
+      click: createSender('menu:new-text-cell'),
+    },
```
rgbkrk, line 270: I like that the keys for code cell and text cell are next to each other. I'm a fan of this.

test/renderer/menu-spec.js
```diff
@@ -31,6 +31,22 @@ describe('menu', () => {
...
+      const store = dummyStore();
+      store.dispatch = sinon.spy();
+
+      menu.dispatchCreateTextCellAfter(store);
+
+      expect(store.dispatch.firstCall).to.be.calledWith({
+        type: constants.NEW_CELL_AFTER,
+        cellType: 'markdown',
+        source: '',
+        id: null,
```
rgbkrk, line 45: I thought this part was going to fail CI then I saw that the `NEW_CELL_AFTER` code uses a `null` id as well.

src/main/menu.js
alexandercbooth: Great, thanks!

test/renderer/menu-spec.js
alexandercbooth: Hm, it looks like it did fail with AppVeyor. What should I do?

lgeiger: It looks like AppVeyor didn't start properly and I can't restart it with the UI. I try to close and reopen your PR.

alexandercbooth: Oh OK, thanks Lukas!

PR #1040, opened 2016-10-19
fix(config): Ignore the case of no config file

captainsafia:
I couldn't get the writeFileObservable to work right in the catch so I just had it ignore the case in which there is no config file for now. 

I think it is fine to not create an nteract.json if the user doesn't have one for now (since it is a "hidden" feature), but maybe provide a notification in the future?


rgbkrk:
Is there a way for us to `touch` a file with node?

rgbkrk:
Closing in favor of #1046. Thanks @captainsafia

Issue #944, opened 2016-10-08
Provide right-click -> Inspect


rgbkrk:
This helps two groups:
- Those hacking on nteract
- Those hacking _in_ nteract (outputs they want to inspect)

jkornblum:
Would incorporating this module (electon-context-menu) be acceptable?
https://github.com/sindresorhus/electron-context-menu

It comes with context aware cut, copy, paste, image-save menu items.
It comes with inspect-element menu item.

It has two dependencies, electon-dl and electron-is-dev. Both from the same author and all under MIT license.

---

If this is too much I can modify the source to just include cut, copy, paste, inspect-element. It would not require any dependencies then.

Let me know what you guys think and I can submit a PR

![ss1](https://cloud.githubusercontent.com/assets/5368903/19463033/73e202c4-94b6-11e6-9857-ace7846aa066.png)
![ss2](https://cloud.githubusercontent.com/assets/5368903/19463037/87c78656-94b6-11e6-81cc-69955a85f00b.png)
![ss3](https://cloud.githubusercontent.com/assets/5368903/19463040/8ad02c86-94b6-11e6-9335-521e4f1a87e0.png)
![ss4](https://cloud.githubusercontent.com/assets/5368903/19463043/909920b4-94b6-11e6-8bc2-98a656f203c5.png)

rgbkrk:
@jkornblum Absolutely! Feel free to make a PR. I'd love to use this.

rgbkrk:
Done in #1044.

PR #1044, opened 2016-10-20
feat(context-menu) adding right click context menu

jkornblum:
adding electron-context-menu library to project
addresses issue #944 

adds inspect element while electron is in dev mode
adds context aware cut, copy, paste, save image, copy link

### DIFF
--- a/src/main/launch.js
+++ b/src/main/launch.js
@@ -20,6 +20,8 @@ export function deferURL(event, url) {
 
 const iconPath = path.join(__dirname, '..', '..', 'static', 'icon.png');
 
+const initContextMenu = require('electron-context-menu');
+
 export function launch(filename) {
   let win = new BrowserWindow({
     width: 800,
@@ -30,6 +32,8 @@ export function launch(filename) {
 
   const index = path.join(__dirname, '..', '..', 'static', 'index.html');
   win.loadURL(`file://${index}`);
+
+  initContextMenu();
 
   win.webContents.on('did-finish-load', () => {
     if (filename) {
--- a/app/package.json
+++ b/app/package.json
@@ -26,6 +26,7 @@
     "commonmark": "^0.26.0",
     "commonmark-react-renderer": "^4.3.1",
     "commutable": "^1.1.3",
+    "electron-context-menu": "^0.6.0",
     "electron-json-storage": "^2.0.0",
     "electron-log": "^1.1.0",
     "enchannel": "^1.1.3",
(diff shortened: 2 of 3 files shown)


src/main/index.js
```diff
@@ -13,6 +13,8 @@ import { defaultMenu, loadFullMenu } from './menu';
 
 import prepareEnv from './prepare-env';
 
+require('electron-context-menu')();
+
```
captainsafia: I think it would make more sense to do this instantiation inside the `launch.js` file within the `launch` function so you're only instantiating a context menu if you know a notebook is being successfully loaded.

captainsafia (changes requested):
Great stuff, Joshua! Made some comments inline.

rgbkrk:
Thanks for this @jkornblum! I'm looking forward to it.

In order to make linting pass, you'll need to do the require separate from the call:

``` js
const initContextMenu = require('electron-context-menu');
initContextMenu();
```

src/main/index.js
jkornblum: Okay, I'm pretty new to nteract (and kind of to Electron/JavaScript ha! I'm more C#/Python guy). I've been familiarizing myself with this code base but help me out.

Do we know a notebook successfully loaded in the 'did-finish-load' event listener for webContents? Or should I just throw the init at the top of the launch() function.

rgbkrk: Welcome! 

It's all good! I too used to be more a Python and C person. The javascript has absorbed me ever since I realized I needed to do more of it to express the ideas I wanted to. :smile:

For this PR, I wouldn't worry about knowing if a notebook successfully loaded. We can't actually know from here without more event handling. I think that's out of scope for this PR. 

Safia wants you to move this function call over to `src/main/launch.js` as well as adapt the way it's called.

jkornblum:
Okay I've updated the PR with recommended changes.

captainsafia (approved):

captainsafia:
So awesome, Joshua! Thanks!

rgbkrk:

src/main/launch.js
```diff
@@ -31,6 +33,8 @@ export function launch(filename) {
   const index = path.join(__dirname, '..', '..', 'static', 'index.html');
   win.loadURL(`file://${index}`);
 
+  initContextMenu();
```
rgbkrk, line 36: This should move out of the `launch` function itself. It applies to all BrowserWindows and only needs to be run once, after the `require`.

jkornblum: That's simple, can this PR be reopened or should I submit another PR?

rgbkrk: D'oh! I made a PR for this in #1062. Feel free to merge that one.

PR #1045, opened 2016-10-20
chore(install): Add reinstall script

lgeiger:
Make reinstalling nteract easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ And then opening `docs/index.html` in your favorite browser.
 
 > I upgraded my developer installation and things are broken!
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
I just noticed that if you don't have `rimraf` globally _and_ you didn't have it installed as part of deps here before, this will fail to run the reinstall. Not sure we have to fix that, just a quirk.

lgeiger:
We can add `rimraf` to the development dependencies.

lgeiger:
Oh I see we have it in the dev deps already.

Issue #804, opened 2016-09-20
Setup all the base jupyter filesystem config on first launch if it doesn't exist


rgbkrk:
Carried over from #427.

To ensure a proper environment, we should go ahead and lay the base for a jupyter install on startup.

rgbkrk:
Closed by #1046

PR #1046, opened 2016-10-20
Ensure all paths and config are setup before app launch.

### LINKED ISSUE #1027: Kernels not connecting when missing `nteract.json` file
While trying to get the install up on `master`, my kernel wasn't actually connecting. Popping up the dev tools showed that `~/Library/Jupyter/nteract.json` was missing. After creating the file with an empty object in it, the kernel connected just fine.

I would suggest that the file be automatically created if it doesn't exist, but I don't know the implications to the rest of the system.

cc @rgbkrk and @jdetle who helped diagnose the issue on slack.

rgbkrk:
Fixes #1027, #804.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -4,6 +4,8 @@ import { existsSync } from 'fs';
 
 import Rx from 'rxjs/Rx';
 
+import { mkdirpObservable } from '../utils/fs';
+
 import {
   launch,
   launchNewNotebook,
@@ -16,6 +18,8 @@ import prepareEnv from './prepare-env';
 const log = require('electron-log');
 
 const kernelspecs = require('kernelspecs');
+const jupyterPaths = require('jupyter-paths');
+const path = require('path');
 
 const argv = require('yargs')
   .version()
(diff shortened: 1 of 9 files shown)


src/main/index.js
```diff
@@ -53,7 +61,37 @@ const fullAppReady$ = Rx.Observable.zip(
...
+      mkdirpObservable(jupyterConfigDir)
+    )
+  )
+  // Set up our configuration file
+  .mergeMap(() =>
+    readFileObservable(nteractConfigFilename)
+      .catch((err) => {
+        if (err.code === 'ENOENT') {
+          return writeFileObservable(nteractConfigFilename, JSON.stringify({
+            // TODO: Pull the default config from somewhere
```
captainsafia, line 83: Might also wanna include the sendMetrics in this config.

captainsafia (changes requested):
Looks great!

src/main/index.js
rgbkrk: Setting it to false for now?

rgbkrk: I mostly left it out to provide flexibility for you later. Otherwise how will you prompt for it in the future?

captainsafia: Yep! False by default. In the immediate future we can do the dialog box that asks the user what they want like before. And in the far future, we would have it as part of the config UI I outlined in another issue somewhere....

rgbkrk: If I leave it `undefined` or set it to `null`, when you implement the metrics setup again you can check for nulliness so we're not missing all the early adopters.

captainsafia (approved):

PR #1049, opened 2016-10-20
fix(ConfigRecord): Treat config as a Map

captainsafia:
This will allow us to be more flexible about how our configuration keys change.

### DIFF
--- a/src/notebook/records.js
+++ b/src/notebook/records.js
@@ -35,6 +35,6 @@ export const CommsRecord = new Immutable.Record({
   targets: new Immutable.Map(),
 });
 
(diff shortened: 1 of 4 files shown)


src/notebook/records.js
```diff
@@ -35,6 +35,6 @@ export const CommsRecord = new Immutable.Record({
   targets: new Immutable.Map(),
 });
 
-export const ConfigRecord = new Immutable.Record({
+export const ConfigRecord = new Immutable.Map({
```
rgbkrk, line 38: Scary name to keep!

Tis ok for now though.

PR #1052, opened 2016-10-21
fix(app): Disable asar on build

captainsafia:
It looks like having our example notebooks archived via asar causes some problems with launching the kernel. Currently, asar [has no way of excluding certain directories](https://github.com/electron/asar/issues/2) so I'm settling it for just disabling at all together for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
     "win": {
       "target": "nsis"
     },
-    "extraResources": "bin"
+    "extraResources": "bin",
+    "asar": false
   },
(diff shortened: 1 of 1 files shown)


lgeiger:
You cloud use the `extraResources` option and move the example notebooks out of the `app` directory. This way we could keep using ASAR.

rgbkrk:
> You could use the extraResources option and move the example notebooks out of the app directory. This way we could keep using ASAR.

Love that plan.

PR #1055, opened 2016-10-22
[WIP] fix(CodeCell): Disable input focus on click in CodeCell

### LINKED ISSUE #429: Don't focus on editor when cell clicked
Currently when you click in the output area it focuses the cursor in the editor. This is great for the sake of focusing the cell. It's not so great when you want to select outputs (especially if the overall output is fairly large).

captainsafia:
The Editor now manages its own focused set via CodeMirror instead of receiving props from the parent CodeCell component.

nteract-no-focus

Closes #429.

### DIFF
--- a/src/notebook/components/cell/cell.js
+++ b/src/notebook/components/cell/cell.js
@@ -115,7 +115,6 @@ export class Cell extends React.Component {
           <MarkdownCell
             focusAbove={this.focusAboveCell}
             focusBelow={this.focusBelowCell}
-            focused={focused}
             cell={cell}
             id={this.props.id}
             theme={this.props.theme}
@@ -123,7 +122,6 @@ export class Cell extends React.Component {
             <CodeCell
               focusAbove={this.focusAboveCell}
               focusBelow={this.focusBelowCell}
-              outputFocused={focused}
               cell={cell}
               id={this.props.id}
               theme={this.props.theme}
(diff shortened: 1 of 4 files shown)


rgbkrk:
Oh my gosh finally thank you!

rgbkrk:
According to linting it looks like this outputFocused prop might as well be deleted since it's not used. When it's in use we can bring it back.

captainsafia:
Bleh! Linter just caught me on to the fact that we really don't need to propagate the `focused` property down to the individual CodeCell at all. I can remove `outputFocused` and all that other stuff because the only thing the `focused` is used for is setting the styling at the `<Cell/>` level. So `focused` is really just a relationship between `Notebook` and `Cell` not `Notebook` and `Cell` and `CodeCell` and `MarkdownCell`. I think we can safely remove the property from `MarkdownCell` too.

rgbkrk:
Woo! Love it when the linter helps us delete more and maintain less! The flow types likely need an update too (run locally, our Travis setup doesnt reject for flow yet).

rgbkrk:
What happens when you hit up and down when inside a cell? Does it still navigate to the next and prior cell?

captainsafia:
> What happens when you hit up and down when inside a cell? Does it still navigate to the next and prior cell?

Argh! Good catch. It only navigates well between adjacent cells. So pressing up and down repeatedly doesn't work. Hmmm. 🤔

captainsafia:
I guess this just means that focus needs to be propagated down to `CodeCell` and `MarkdownCell` but not `Editor`.

captainsafia:
Closing in favor of a fresher PR.

PR #1056, opened 2016-10-23
[WIP] Webpack dx

rgbkrk:
Don't merge, I'm only making things worse for a little bit while I figure this out.


rgbkrk:
I'll revisit this one after @lgeiger puts us back on one package.json with the new electron-builder setup.

PR #1058, opened 2016-10-23
refactor(mode): require common languages up front

rgbkrk:
Since we bundle with webpack, we should no longer `require` these dynamically. If we want to load any dynamically later, we should load them with direct script tags in the `index.html`.

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -28,9 +28,21 @@ import { executeCell } from '../epics/execute';
 
 import type { CellProps } from './cell/cell';
 
-// Always set up the markdown mode
+// Always set up the markdown modes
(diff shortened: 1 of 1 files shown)


src/notebook/components/notebook.js
```diff
@@ -54,22 +66,6 @@ export function getLanguageMode(notebook: any): string {
     notebook.getIn(['metadata', 'language_info', 'codemirror_mode'],
     notebook.getIn(['metadata', 'language_info', 'name'],
     'text')));
-
-  if (language !== 'ipython' && language !== 'text') {
-    // HACK: This should give you the heeby-jeebies
```
rgbkrk, line 59: Bye bye `HACK`!

PR #1062, opened 2016-10-24
refactor(context-menu): install in app, init once

rgbkrk:
Follow up to #1044
- Remove context menu install inside of `./package.json`, needed to only be in `./app/package.json`
- [`initContextMenu()` should run at the top level](https://github.com/sindresorhus/electron-context-menu#usage)
- Move electron react dev tools to `app` packaging


app/package.json
```diff
@@ -31,6 +31,7 @@
     "electron-log": "^1.1.0",
     "enchannel": "^1.1.3",
     "enchannel-zmq-backend": "^2.5.0",
+    "electron-react-devtools": "^0.4.0",
```
lgeiger, line 34: We load the devtools only when inside a dev environment: https://github.com/nteract/nteract/blob/master/app/static/index.html#L49

If we want to change that we should probably get rid of this as well.

rgbkrk: This does leave us open to allowing the user to get to dev mode if they want to, without having to run off master. Alternatively, I could move this to the devdeps inside app/package.json.

lgeiger: > This does leave us open to allowing the user to get to dev mode if they want to, without having to run off master.

Oh I see. That makes sense. I'll probably add `nteract --debug` flag for this.

> Alternatively, I could move this to the devdeps inside app/package.json.

This won't make sense since I think `electron-builder` won't install them.

PR #1063, opened 2016-10-24
chore(pack): Enable ASAR

lgeiger:
Follow up to #1052


rgbkrk:
Awesome, that's great.

PR #1066, opened 2016-10-25
fix(splash): Use high DPI mp4

### LINKED ISSUE #969: High DPI splash screen
The splash screen looks very blurry on high DPI screens.

There are a few ways we could fix this:
1. Make it half the size
2. Use something like [Bodymoving](https://creative.adobe.com/addons/products/12557#.V_uqfpOLSHo) to export the animation from [After Effects](https://github.com/nteract/logos/blob/master/nteract_logo_cube_book/source_files/animation/composition/nteract_logo_motion.aep) to HTML or SVG (no more GIF)
3. Use a static ~~high resolution~~SVG splash screen

lgeiger:
Closes #969


rgbkrk:
Sweet. This could probably be a bigger size now too. It's really crisp.

lgeiger:
I can make it bigger if you like. It is now rendered @2x.

PR #1068, opened 2016-10-25
Epics testing

jdetle:
Hello, all! It has been a while! I added some tests for the github-publish epic and refactored out some anonymous functions. 

Is there a better word for chore than writing tests? Sorry if its the wrong verb.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -20,7 +20,6 @@ export const PUBLISH_ANONYMOUS_GIST = 'PUBLISH_ANONYMOUS_GIST';
 /**
  * Notify the notebook user that it has been published as a gist.
  * @param {string} filename - Filename of the notebook.
- * @param {string} gistURL - URL for the published gist.
  * @param {string} gistID - ID of the published gist, given after URL
  * @param {object} notificationSystem - To be passed information for
  * notification of the user that the gist has been published.
@@ -62,7 +61,7 @@ export function createGistCallback(firstTimePublish, observer, filename, notific
     const gistID = response.id;
     const gistURL = response.html_url;
 
-    notifyUser(filename, gistURL, gistID, notificationSystem);
+    notifyUser(filename, gistID, notificationSystem);
     if (firstTimePublish) {
       // TODO: Move this up and out to be handled as a return on the observable
       observer.next(overwriteMetadata('gist_id', gistID));
@@ -140,7 +139,7 @@ export function publishNotebookObservable(github, notebook, filepath,
  * @param  {Immutable.Record} store - Redux store containing current state.
  *
  */
-export function handleGistError(error, store) {
+export function handleGistError(store, error) {
   const state = store.getState();
   const notificationSystem = state.app.get('notificationSystem');
   // TODO: Let this go into the general error flow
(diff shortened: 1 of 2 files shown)


src/notebook/epics/github-publish.js
```diff
@@ -163,31 +198,4 @@ export function handleGistAction(action, store) {
...
-        });
-        return;
-      }
-      notificationSystem.addNotification({
-        title: 'Unknown Publication Error',
-        message: err.toString(),
-        level: 'error',
-      });
-    });
+    .catch((err) => handleGistError(err, store));
```
rgbkrk: If you make the store the first argument you can create this anonymous function earlier and bind it:

const boundHandleGistError = handleGistError.bind(null, store)

Then pass boundHandleGistError to the catch.

rgbkrk (approved):

src/notebook/epics/github-publish.js
jdetle: would that be inside the epic?

src/notebook/epics/github-publish.js
```diff
@@ -135,6 +135,41 @@ export function publishNotebookObservable(github, notebook, filepath,
...
 /**
+ * Handle gist errors for the publish epic.
+ * @param  {String} error - Error response to be parsed and handled.
+ * @param  {Immutable.Record} store - Redux store containing current state.
+ *
+ */
+export function handleGistError(error, store) {
+  const state = store.getState();
+  const notificationSystem = state.app.get('notificationSystem');
+  // TODO: Let this go into the general error flow
```
rgbkrk, line 145: This highlights something important which is that we could be propagating these errors up as actions.

src/notebook/epics/github-publish.js
rgbkrk: Yeah, before the actions of type pary

rgbkrk (approved):

jdetle:
Shucks well this isn't working:

```
export const publishEpic = (action$, store) =>
  const boundHandleGistError = handleGistError.bind(null, store);
  action$.ofType(PUBLISH_USER_GIST, PUBLISH_ANONYMOUS_GIST)
    .mergeMap((action) => handleGistAction(action, store))
    .catch(boundHandleGistError(err));
```

I dont know if you wanted something else?
I got an error at `const boundHandleGistError`

rgbkrk:
Turn

``` js
export const publishEpic = (action$, store) =>
  const boundHandleGistError = handleGistError.bind(null, store);
  action$.ofType(PUBLISH_USER_GIST, PUBLISH_ANONYMOUS_GIST)
    .mergeMap((action) => handleGistAction(action, store))
    .catch(boundHandleGistError(err));
```

into

``` js
export const publishEpic = (action$, store) => {
  const boundHandleGistError = handleGistError.bind(null, store);
  return action$.ofType(PUBLISH_USER_GIST, PUBLISH_ANONYMOUS_GIST)
    .mergeMap((action) => handleGistAction(action, store))
    .catch(boundHandleGistError(err));
}
```

The reason is that the arrow function without braces is an expression.

``` js
() => 2
```

is equivalent to

``` js
() => {
  return 2
}
```

As soon as you need more than one statement you have to use the braces and write an explicit return (if you need it):

``` js
export const publishEpic = (action$, store) => {
  const boundHandleGistError = handleGistError.bind(null, store);
  return action$.ofType(PUBLISH_USER_GIST, PUBLISH_ANONYMOUS_GIST)
    .mergeMap((action) => handleGistAction(action, store))
    .catch(boundHandleGistError(err));
}
```

rgbkrk (approved):

PR #1070, opened 2016-10-26
chore(EpicTesting): Fix issue in github testing.

jdetle:
I checked codecov and some of the tests weren't actually working. This should hopefully fix things!(?)

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -142,30 +142,34 @@ export function publishNotebookObservable(github, notebook, filepath,
 export function handleGistError(store, err) {
   const state = store.getState();
   const notificationSystem = state.app.get('notificationSystem');
-  const error = JSON.parse(err);
-  // TODO: Let this go into the general error flow
-  if (error.message) {
-    if (error.message === 'Bad credentials') {
+  try {
+    const error = JSON.parse(err);
+    // TODO: Let this go into the general error flow
+    if (error.message) {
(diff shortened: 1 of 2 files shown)


jdetle:
Okay, shoot, I found some problems with the work I did.

src/notebook/epics/github-publish.js
```diff
@@ -156,14 +156,14 @@ export function handleGistError(store, error) {
...
-      message: githubError.message,
+      message: error.message,
       level: 'error',
     });
     return;
   }
   notificationSystem.addNotification({
     title: 'Unknown Publication Error',
-    message: error.toString(),
+    message: err.toString(),
```
rgbkrk: Why `err` here?

jdetle: Good question, I'm not sure what we would want to do if the error message cant be parsed. But this was due to the fact that I was passing a message that couldn't be parsed.

src/notebook/epics/github-publish.js
```diff
@@ -139,13 +139,13 @@ export function publishNotebookObservable(github, notebook, filepath,
  * @param  {Immutable.Record} store - Redux store containing current state.
  *
  */
-export function handleGistError(store, error) {
+export function handleGistError(store, err) {
   const state = store.getState();
   const notificationSystem = state.app.get('notificationSystem');
+  const error = JSON.parse(err);
```
rgbkrk: Since this comes from a remote, I'd wrap this in a try catch block.

PR #1071, opened 2016-10-26
Notebook testing

jdetle:
**Done**:
Extracted scroll to element function, made it provide int that we will set scrollTop to. Added tests.

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -74,9 +74,9 @@ export function getLanguageMode(notebook: any): string {
  * information.
  * @param  {HTMLElement} el - Element to be compared against window and body for
  * scrollTop value.
- * @return {Integer} - An integer for the new document scrollTop value. 
+ * @return {Integer} - An integer for the new document scrollTop value.
  */
-export function scrollToElement(el: HTMLElement): int {
+export function scrollToElement(el: HTMLElement): number {
   const viewportHeight = window.innerHeight;
   const viewportOffset = document.body.scrollTop;
 
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -70,7 +70,7 @@ export function getLanguageMode(notebook: any): string {
   return language;
 }
 
-export function scrollToElement(el) {
+export function scrollToElement(el: HTMLElement): int {
   const viewportHeight = window.innerHeight;
   const viewportOffset = document.body.scrollTop;
 
@@ -91,7 +91,7 @@ export function scrollToElement(el) {
     const offset = viewportHeight - cellHeight;
     return cellTop - offset;
   }
-  return undefined;
+  return document.body.scrollTop;
 }
 
 const mapStateToProps = (state: Object) => ({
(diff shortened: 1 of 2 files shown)


src/notebook/components/notebook.js
```diff
@@ -70,6 +70,30 @@ export function getLanguageMode(notebook: any): string {
   return language;
 }
 
+export function scrollToElement(el) {
```
rgbkrk: Since this component is flow typed, I'd make the arg here be `el: HTMLElement`

src/notebook/components/notebook.js
```diff
@@ -70,6 +70,30 @@ export function getLanguageMode(notebook: any): string {
...
+  }
+
+  if (belowFold) {
+    if (cellHeight > viewportHeight) {
+      return cellTop;
+    }
+    const offset = viewportHeight - cellHeight;
+    return cellTop - offset;
+  }
+  return undefined;
```
rgbkrk: This could return `document.body.scrollTop` instead of undefined.

src/notebook/components/notebook.js
rgbkrk: Also, since this is now out of context I'd put the JSDoc in that includes a description of the return.

src/notebook/components/notebook.js
```diff
@@ -174,30 +198,9 @@ export class Notebook extends React.Component {
...
-
-      if (belowFold) {
-        if (cellHeight > viewportHeight) {
-          document.body.scrollTop = cellTop;
-        } else {
-          const offset = viewportHeight - cellHeight;
-          document.body.scrollTop = cellTop - offset;
-        }
-      }
+      document.body.scrollTop = scrollToElement(focusedCell);
```
rgbkrk, line 209: This needs checking now that `scrollToElement` can be set to `undefined`. You don't want to set that on the DOM.

rgbkrk (changes requested):

src/notebook/components/notebook.js
jdetle: Totally broke this unintentionally.

src/notebook/components/notebook.js
jdetle: Sounds good.

src/notebook/components/notebook.js
```diff
@@ -69,6 +69,36 @@ export function getLanguageMode(notebook: any): string {
     'text')));
   return language;
 }
+/**
+ * Provide the appropriate position to scroll to when given cell position and size
+ * information.
+ * @param  {HTMLElement} el - Element to be compared against window and body for
+ * scrollTop value.
+ * @return {Integer} - An integer for the new document scrollTop value.
```
rgbkrk, line 77: This should be `number` in the type for return, not a big deal though - can be done in a follow up PR.

src/notebook/components/notebook.js
```diff
@@ -69,6 +69,36 @@ export function getLanguageMode(notebook: any): string {
...
+  }
+
+  if (belowFold) {
+    if (cellHeight > viewportHeight) {
+      return cellTop;
+    }
+    const offset = viewportHeight - cellHeight;
+    return cellTop - offset;
+  }
+  return document.body.scrollTop;
```
rgbkrk, line 100: It looks like we never get to this case in our tests. Is there a way to reach it?

rgbkrk (approved):

PR #1079, opened 2016-10-26
removed TODOs from nteract/src; created associated issues

LBKinson:


rgbkrk:
Incredible. Thank you!

too awesome

rgbkrk:
There is one additional thing we require of collaborators, which is to [format commit messages with a prefix](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md#how-should-i-write-my-commit-messages-and-pr-titles). In this case it would be `chore(cleanup): convert todo to issue`. However, I'll just merge this one and set the merge commit to show the `chore` prefix (they don't end up in the changelog anyways).

PR #1083, opened 2016-10-27
Epics testing

jdetle:
## Todo:
- [x] Cleanup github testing
- [x] Figure out how to change state of store, or create a state to pass to new function
- [x] Determine what args we want in createExecuteCellObservable
- [ ] Figure out how to mock a stream of actions.

### DIFF
--- a/src/notebook/epics/execute.js
+++ b/src/notebook/epics/execute.js
@@ -265,11 +265,35 @@ export function executeCell(id, source) {
   };
 }
 
+export function createExecuteCellObservable(action$, store, source, id) {
+  const state = store.getState();
+  const channels = state.app.channels;
+
+  const kernelConnected = channels &&
+    !(state.app.executionState === 'starting' ||
+      state.app.executionState === 'not connected');
+
+  if (!kernelConnected) {
+    // TODO: Switch this to dispatching an error
+    state.app.notificationSystem.addNotification({
+      title: 'Could not execute cell',
+      message: 'The cell could not be executed because the kernel is not connected.',
+      level: 'error',
+    });
+    return Rx.Observable.of(updateCellExecutionCount(id, undefined));
+  }
+
+  return executeCellObservable(channels, id, source)
+    .takeUntil(action$.filter(laterAction => laterAction.id === id)
+                      .ofType(ABORT_EXECUTION, REMOVE_CELL));
+};
+
 /**
  * the execute cell epic processes execute requests for all cells, creating
  * inner observable streams of the running execution responses
  */
 export function executeCellEpic(action$, store) {
+  const boundCreateExecuteCellObservable = createExecuteCellObservable.bind(null, action$, store);
   return action$.ofType('EXECUTE_CELL')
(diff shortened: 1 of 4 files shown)


jdetle:
@captainsafia @rgbkrk I'm not sure if you'll laugh or cringe at what I did. Probably some combination of both...

jdetle:
Ideally, I would like to test the returned stream to ensure that the actions are of the appropriate type. Let me know how I should go about approaching this. It would probably require the creation of an actual observable and testing that this behaves appropriately, but then we run the risk of testing that RxJS works, right?

src/notebook/epics/execute.js
```diff
@@ -309,7 +309,8 @@ export function executeCellEpic(action$, store) {
     .map(cellActionObservable =>
       cellActionObservable
         // When a new EXECUTE_CELL comes in with the current ID, we create a
-        .switchMap(({source, id}) => boundCreateExecuteCellObservable(source, id))
+        // a new observable and unsubscribe from the old one.
+        .switchMap(({ source, id }) => boundCreateExecuteCellObservable(source, id))
     )
     // Bring back all the inner Observables into one stream
     .mergeAll()
```
rgbkrk, line 316: A+ commit message here

rgbkrk:
It's ok - in other test we create an observable with actions like this:

```
const actions = Rx.Observable.from([{
      type: 'something'
      data: {},
    }]);
```

Now you have an Observable to work with for pumping through an epic.

rgbkrk (approved):

rgbkrk:
Here's a way to at the very least [call the function and test that it handles an ActionsObservable](https://github.com/nteract/nteract/blob/master/test/renderer/epics/kernel-launch-spec.js#L87-L93):

``` js
import { ActionsObservable } from 'redux-observable';

...

describe('watchExecutionStateEpic', () => {
  it('returns an Observable with an initial state of idle', () => {
    const action$ = new ActionsObservable();
    const obs = watchExecutionStateEpic(action$);
    expect(obs.subscribe).to.not.be.null;
  });
});
```

rgbkrk:
Here's an example out of our current code base of tests that properly pumps through actions and makes sure that it recieves a series of response actions:

https://github.com/nteract/nteract/blob/67f7964970541e12341b645f184c2b5667a812a8/test/renderer/epics/comm-spec.js#L161-L185

rgbkrk:
Basically, the pattern we want to aim for with testing observables from observables on the outside is:

``` js
const actionBuffer = [];

const responseActions = someEpic(mockActions, store);

const subscription = responseActions.subscribe(
  actionBuffer.push, // Every action that goes through should get stuck on an array
  (err) => expect.fail(err, null), // It should not error in the stream
  () => {
    expect(actionBuffer).to.have.whatever.structure.you.think(); // ;)
    done();
  }
```

rgbkrk:
I'll go ahead and merge it as it moves us forward with testing. Hopefully my comments help and that I'll be able to dive in with you again on this.

PR #1106, opened 2016-10-29
chore(ExecuteEpic): Try to have test expect error.

jdetle:
Refactor and add some work to ExecuteEpic. cc/ @rgbkrk @captainsafia, I don't think the error expectation here is actually getting evaluated. I'd really appreciate it if one of you could check this out and see if I'm handling things correctly. I might be creating meaningless tests.


rgbkrk:
This isn't meaningless - we certainly want the edge cases tested. Before this merely went through the `.do` without ever invoking the error. 👍

PR #1107, opened 2016-10-30
chore(install): Remove prepublish script

lgeiger:
It won't build after `npm install` and removes the following warning on Node 7:

```
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts will run only for `npm publish`.
npm WARN prepublish-on-install (In npm@4 and previous versions, it also runs for `npm install`.)
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.
```


rgbkrk (approved):
Since we run build on npm run start I suppose this isnt so bad to get rid of.

PR #1110, opened 2016-10-30
chore(tests): Cleanup kernel shutdown specs

lgeiger:
Use `mock-require`

### DIFF
--- a/test/renderer/api/kernel-spec.js
+++ b/test/renderer/api/kernel-spec.js
@@ -1,6 +1,7 @@
 const chai = require('chai');
 const sinon = require('sinon');
 const sinonChai = require("sinon-chai");
+const fs = require('fs');
 
 chai.use(sinonChai);
 const expect = chai.expect;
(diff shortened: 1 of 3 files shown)


test/setup.js
```diff
@@ -140,3 +141,7 @@ mock('react-notification-system', function () {
 mock('spawnteract', {
   'launch': function(kernelSpec, config) { return new Promise() },
 });
+
+mock('fs', {
+  unlinkSync: sinon.spy(),
```
rgbkrk, line 146: The only problem with this is that it's a global spy, which means that once anyone calls it "calledWith" will be true.

test/renderer/api/kernel-spec.js
```diff
@@ -68,19 +65,15 @@ describe('forceShutdownKernel', () => {
     expect(kernel.spawn.kill).to.have.been.calledWith('SIGKILL');
```
rgbkrk, line 66: I'd set up the spy in each of these functions that need it (still making the mock require in the other file) by doing

spy = sinon.spy(fs, 'unlinkSync')

You can then disattach the spy at the end of this test.

rgbkrk (changes requested):

rgbkrk:
Thanks for cleaning this one up.

test/renderer/api/kernel-spec.js
lgeiger: That's what I tried first but I can't get it to work.

rgbkrk: Ah ok. We can iterate after.

PR #1111, opened 2016-10-30
Saving epic tests

jdetle:
Put this up so someone can check it out and help me debug. Currently actions aren't getting pushed that should be getting emitted, please help if you can. Seriously confused as to why line 66 is working and line 80 isn't. Thanks for any help that people are willing to give :)

### DIFF
--- a/src/notebook/epics/saving.js
+++ b/src/notebook/epics/saving.js
@@ -37,7 +37,7 @@ export function saveEpic(action$) {
         // since SAVE effectively acts as the same as START_SAVING
         // you could just look for that in your reducers instead of START_SAVING
     );
-};
+}
 
 export function saveAsEpic(actions) {
   return actions.ofType(SAVE_AS)
--- a/src/notebook/epics/saving.js
+++ b/src/notebook/epics/saving.js
@@ -34,7 +34,7 @@ export function saveEpic(action$) {
           null,
           1))
         .catch(error => {
-          const input$ = Observable.of({ type: 'SAVING_ERROR' });
+          const input$ = Observable.of(error);
           return new ActionsObservable(input$);
         })
         .map(doneSaving)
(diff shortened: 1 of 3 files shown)


lgeiger:
@jdetle I think you are missing to call the `done()` callback on the in your tests. I suppose line 80 only works because it's synchronous.

I can push a fix, but it'll probably make both tests fail :wink:

jdetle:
So one issue was that in my observer, I wasn't actually passing any args to actionBuffer.push, which was definitely a problem. But I'll look into that too, thanks.

rgbkrk:
Looks like this is passing now - still need folks to check on stuff or is it ready to merge?

jdetle:
I want you to check what I added to the saving epic, but other than that I think its ready to ship.

src/notebook/epics/saving.js
```diff
@@ -32,15 +33,21 @@ export const saveEpic = actions =>
                 value.delete('inputHidden').delete('outputHidden').delete('status')))),
           null,
           1))
+        .catch(error => {
+          const input$ = Observable.of({ type: 'SAVING_ERROR' });
```
rgbkrk: You can just return `Observable.of` here. I would also include the error similar to our other payloads so this is easier to debug/deal with.

rgbkrk (approved):
One thing to change if you like, otherwise it's good to go.

src/notebook/epics/saving.js
jdetle: What would that error be? My first thought would be to create a new error action?

rgbkrk:
``` js
.catch(error => { // <-- This one
```

jdetle: Oh! Right. Silly me.

rgbkrk: You would make this:

``` js
Observable.of({ type: 'ERROR_SAVING', payload: error })
```

jdetle:
90% is so close I can almost taste it.

PR #1112, opened 2016-10-30
chore(test): Add missing done() callbacks

lgeiger:
During looking at #1111 I realized that in some of our test the `done()` callbacks are missing.

Unfortunately this makes some of our tests fail.


lgeiger:
I removed the failing test for now until we figure out how to properly test `publishEpic`.

jdetle:
Thanks @lgeiger!

PR #1115, opened 2016-10-31
chore(err): Provide an up to date guide to fixing an nteract dev install

rgbkrk:
While commenting on another repo to mention how we handle displaying errors for native module mismatches, I noticed this was out of date. Now it's more relevant.

### DIFF
--- a/app/static/index.html
+++ b/app/static/index.html
@@ -71,7 +71,7 @@
               break;
             default:
               if(/Module version mismatch/.test(err.message)) {
-                msgContainer.innerHTML = '<pre>nteract <a href="https://github.com/nteract/nteract#dependencies">requires node 6 for dev install</a></pre>';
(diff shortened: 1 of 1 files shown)


app/static/index.html
```diff
@@ -71,7 +71,7 @@
               break;
             default:
               if(/Module version mismatch/.test(err.message)) {
-                msgContainer.innerHTML = '<pre>nteract <a href="https://github.com/nteract/nteract#dependencies">requires node 6 for dev install</a></pre>';
+                msgContainer.innerHTML = `<p>nteract's native modules mismatch, try running <a href="https://github.com/nteract/nteract#troubleshooting"><pre>npm run reinstall</pre></a> from the root of the git clone</p>`;
```
lgeiger: `npm run reinstall` is a bit hard and it takes long. `npm install` should be enough.

rgbkrk: If they're in this fail state though, would an `npm install` really save them?

rgbkrk:
Switched to `npm install` @lgeiger.

PR #1119, opened 2016-11-01
Saving cleanup

jdetle:
Moved actions into `notebook/actions`, constants into `notebook/constants` for consistency.


rgbkrk:
Looking good.

PR #1121, opened 2016-11-01
chore(configEpic): Test and refactor configEpic.

jdetle:
More epic testing.

### DIFF
--- a/src/notebook/epics/config.js
+++ b/src/notebook/epics/config.js
@@ -51,7 +51,7 @@ export const saveConfigEpic = actions =>
     .mergeMap(action =>
       writeFileObservable(CONFIG_FILE_PATH, JSON.stringify(store.getState().config.toJS()))
       .map(doneSavingConfig)
-      .catch((err) =>
-        Observable.of({ type: 'ERROR', payload: err, error: true })
-      )
+    )
+    .catch((err) =>
+      Observable.of({ type: 'ERROR', payload: err, error: true })
     );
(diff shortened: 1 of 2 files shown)


src/notebook/epics/config.js
```diff
@@ -51,7 +51,7 @@ export const saveConfigEpic = actions =>
     .mergeMap(action =>
       writeFileObservable(CONFIG_FILE_PATH, JSON.stringify(store.getState().config.toJS()))
       .map(doneSavingConfig)
-      .catch((err) =>
-        Observable.of({ type: 'ERROR', payload: err, error: true })
-      )
+    )
+    .catch((err) =>
+      Observable.of({ type: 'ERROR', payload: err, error: true })
```
rgbkrk, line 56: Did you move the catch up for easier testing?

Theres a semantic reason to catch on the inside, which is specificity. It doesn't look like it matters here though. You could also test this simple observable error action creator

rgbkrk (approved):

src/notebook/epics/config.js
jdetle: There were errors that weren't getting caught by the catch statement, so I moved it out of mergemap. I think I understand what you're saying regarding specificity. Would it be overkill to have two separate catches?

rgbkrk: That would not be overkill, depending on how it's used.

Cool things you can do with Rx's catch (unrelated to this one):
- Return an observable that retries the entire observable
- Try something else entirely as an observable

Since this one is only reporting the error directly (and generically), it's fine to be moved out like it is now.

rgbkrk (approved):

PR #1123, opened 2016-11-01
chore(tests): Initial fs-observable testing

lgeiger:
Very rudimentary `fs-obselvable` testing.


rgbkrk:
Very. We'll definitely want to come back to these.

rgbkrk:
We are so close to 90%!

lgeiger:
> We are so close to 90%!

That was the motivation 😄

rgbkrk:
~14 more PRs like this

rgbkrk:
Though you'll be saddened to see the coverage drop if we start testing main. 😉

jdetle:
If we don't then we're just living a lie anyways.

PR #1124, opened 2016-11-01
chore(pathFallback): Test path fallback

lgeiger:
More simple tests

### DIFF
--- a/test/renderer/path-spec.js
+++ b/test/renderer/path-spec.js
@@ -1,0 +1,19 @@
+import { expect } from 'chai';
+import { remote } from 'electron';
+import {
+  defaultPathFallback,
+  cwdKernelFallback
+} from '../../src/notebook/path';
+
+describe('defaultPathFallback', () => {
+  it('returns a object with the defaultPath', () => {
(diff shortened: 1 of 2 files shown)


test/renderer/path-spec.js
```diff
@@ -11,9 +11,10 @@ describe('defaultPathFallback', () => {
     expect(path).to.deep.equal({ defaultPath: 'dummy-path' });
   });
   it('returns a object with the correct path', () => {
-    process.chdir('/')
-    const path = defaultPathFallback();
-    expect(path).to.deep.equal({ defaultPath: '/home/home/on/the/range' });
-
+    if (process.platform !== 'win32') {
```
rgbkrk, line 14: Should we do this check in the code for the path helpers?

rgbkrk (approved):

test/renderer/path-spec.js
lgeiger: We could, but I think the problem is cause by  the test and not by the function.

I suppose `process.chdir('/')` fails since it's invalid path windows.

PR #1127, opened 2016-11-02
fix(messaging): Store session id in state

### LINKED ISSUE #1087: move session ID from singleton to the state tree
within `src/notebook/kernel/messaging.js`

captainsafia:
Closes #1087.

### DIFF
--- a/src/notebook/components/cell/editor/complete.js
+++ b/src/notebook/components/cell/editor/complete.js
@@ -35,7 +35,7 @@ export function codeCompleteObservable(channels, editor, message) {
   });
 }
 
-export const completionRequest = (code, cursorPos) =>
+export const completionRequest = (code, cursorPos, session) =>
   createMessage(
     'complete_request',
     {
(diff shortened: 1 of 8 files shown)


rgbkrk:
Every single jupyter message is going to need the session ID.

captainsafia:
> Every single jupyter message is going to need the session ID.

Isn't that the case since it's in `createMessage`?

rgbkrk:
I'd aim at passing the session ID into create message (it doesn't and shouldn't have access to the store)

captainsafia:
> I'd aim at passing the session ID into create message (it doesn't and shouldn't have access to the store)

Great point! Code updated.

captainsafia:
Oops! Missed some stuff. Don't merge yet.

src/notebook/components/cell/editor/complete.js
```diff
@@ -35,23 +35,24 @@ export function codeCompleteObservable(channels, editor, message) {
...
         code,
         cursor_pos: cursorPos,
       }
-    }
+    },
+    session
   );
 
-export function codeComplete(channels, editor) {
+export function codeComplete(channels, editor, session) {
```
rgbkrk, line 50: This makes me wonder if our channels object should have the session. 🤔

captainsafia: I think this would actually be much easier. As I work on adding the session to the execution workflow I'm realizing it would be so much easier to have it couple with one of the existing parameters passed in.

And given what channels represents, it makes sense for the sessionId to go there.

PR #1129, opened 2016-11-02
chore(testPublishEpic): Add complete in observable, test epic.

jdetle:
Added a complete call at the end of the publishNotebookObservable, tested publishEpic.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -64,9 +64,8 @@ export function createGistCallback(firstTimePublish, observer, filename, notific
     notifyUser(filename, gistID, notificationSystem);
     if (firstTimePublish) {
       observer.next(overwriteMetadata('gist_id', gistID));
-    } else {
-      observer.next();
     }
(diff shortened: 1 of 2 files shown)


src/notebook/epics/github-publish.js
```diff
@@ -131,6 +131,7 @@ export function publishNotebookObservable(github, notebook, filepath,
       github.gists.create(gistRequest,
         createGistCallback(true, observer, filename, notificationSystem));
     }
+    observer.complete();
```
rgbkrk: This belongs in the `createGistCallback` doesn't it? You don't want the complete to run _before_ any `next`.

rgbkrk (changes requested):

src/notebook/epics/github-publish.js
jdetle: yeah your probably right.

jdetle:
Closing, this, will try to refactor the github publish code to something that is less difficult to test.

PR #1133, opened 2016-11-02
[DISCUSSION] chore(CLI): Unix: Use node script instead of nohup

### LINKED ISSUE #1132: command line `nteract` just shows splash screen
When running `nteract foo.ipynb` at the command line, I just get a splash screen and nothing else seems to happen. Running the actual app seems to work fine though.

PS:  This is a fantastic app and I am so happy that someone is working on it. Love the dark theme 👍

lgeiger:
I hoped this would fix #1132 but it doesn't.

Since it's already written, I let you decide if this is better than the old method or not.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
 dist/*.js
 test/
 scripts/
-bin/win
+bin/
--- a/bin/nteract.js
+++ b/bin/nteract.js
@@ -1,11 +1,12 @@
 var path  = require('path');
 var spawn = require('child_process').spawn;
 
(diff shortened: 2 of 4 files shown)


rgbkrk:
Do you want to just go ahead with this? I think you're the one most likely to know whether or not this is good to iterate with.

lgeiger:
Well, it doesn't really solve a problem. So I think we can close it.

I shared it because this code eliminates variables when trying to debug #1132.

rgbkrk:
Closing as advised

PR #1138, opened 2016-11-03
fix(zoomReset): changed reset level to 0

alexandercbooth:
I noticed when reseting zoom level to actual (CMD-0) that it wasn't the right size. It's my mistake, when I added this originally, I thought zoom level of 1 was 100%, but I see now it's actually zoom level 1. So I changed the actual reset level 0 to match the default and the corresponding test.


rgbkrk:
Awesome, thanks @alexandercbooth!

alexandercbooth:
My pleasure!

PR #1142, opened 2016-11-04
style(height): increase the default scroll height

rgbkrk:
To accomodate the default `df.head()`, plotly plot, etc. on a typical screen.


rgbkrk:
Going to fix up appveyor next.

Issue #1141, opened 2016-11-03
Add the Altair Notebook to the example notebooks menu


rgbkrk:
Title says it. The other example notebooks get loaded in by setting the Menu items directly. We likely should script our usage of `example-notebooks`. In the meantime, we should add the Altair notebook to the menu list.

rgbkrk:
Done in #1146

PR #1146, opened 2016-11-04
chore(exampleNotebooks): add altair notebook to menu

alexandercbooth:
Addresses #1141 and reorganizes notebooks so intro is at the top and similar notebooks near each other (all plotting notebooks first, then geo notebooks).

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -84,7 +84,7 @@ export const fileSubMenus = {
     label: '&Open Example Notebooks',
     submenu: [
       {
-        label: '&Geojson',
+        label: '&Intro',
         click: () => {
           const opts = {
             title: 'Open a notebook',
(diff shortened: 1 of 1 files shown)


src/main/menu.js
```diff
@@ -170,7 +170,22 @@ export const fileSubMenus = {
               'openFile',
             ],
           };
-          launch(path.join(exampleNotebooksDirectory, 'intro.ipynb'));
+          launch(path.join(exampleNotebooksDirectory, 'pandas-to-geojson.ipynb'));
+        },
+      },
+      {
+        label: '&Download nteract Stats',
```
rgbkrk: Could we label this one "Analyze nteract download metrics"?

rgbkrk (approved):

src/main/menu.js
alexandercbooth: Definitely!

PR #1147, opened 2016-11-04
feat(app): Ask user to save before closing

### LINKED ISSUE #1090: Ask user to save before global unload function
Use `beforeunload` to ask user about saving before closing.

captainsafia:
Closes #1090.

Shouldn't be merged yet. Although the `beforeUnload` function is triggered appropriately, the dialog doesn't display and I have no clue. Everything seems to be configured properly. Is there something obvious I am missing?

### DIFF
--- a/src/notebook/global-events.js
+++ b/src/notebook/global-events.js
@@ -2,10 +2,28 @@ import {
   forceShutdownKernel,
 } from './kernel/shutdown';
 
-/**
-export function beforeUnload(store, dispatch, e) {
+import { save } from './actions';
+
+import { remote } from 'electron';
+const dialog = remote.dialog;
+
+export function beforeUnload(store, e) {
+  const state = store.getState();
+  const filename = state.metadata.filename;
+  const notebook = state.document.notebook;
+  dialog.showMessageBox(remote.getCurrentWindow(), {
+    type: 'question',
+    button: ['Save', 'Don\'t Save'],
+    defaultId: 0,
+    title: 'Save Notebook',
+    message: 'Save notebook before closing?',
+    detail: 'Would you like to save this notebook before closing?',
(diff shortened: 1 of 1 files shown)


rgbkrk:
You need to set the e.returnValue to `false` - see the Electron docs for an example on the close event: https://github.com/electron/electron/blob/ef9d8fe9bd54e9c57568d7dd09d578034eebfb3d/docs/api/browser-window.md#event-close

captainsafia:
Got it! Debugging an issue where the dialog won't stay and wait for user input now....

rgbkrk:
This event happens synchronously (since it's handling SIGINT, quit, etc.). You'll need to unconditionally set `e.returnValue` to `false` and then do a proper quit based on the dialog after.

captainsafia:
> This event happens synchronously (since it's handling SIGINT, quit, etc.). You'll need to unconditionally set e.returnValue to false and then do a proper quit based on the dialog after.

Yeah. That's what I'm currently doing. There is no opportunity to do a proper quit (or to dispatch the save action) because the dialog immediately closes the moment it appears.

An interesting update — this problem (dialog window immediately closing) only happens on the _first_ window that is opened in a particular nteract session. It works alright for any new windows opened afterwards.

rgbkrk:
Yeah, this is why I think about us moving more to the main process closing down kernels, saving files, etc. (or at least access to it), since we can handle signal interrupts and more directly.

rgbkrk:
Want to re-tackle this one now that main does a bit more event handling?

rgbkrk:
I'm hoping we can get this one working in today's release, I'll see what I can do.

rgbkrk:
Opening a new PR, since the branch here is gone.

PR #1150, opened 2016-11-06
style(gangnam): handle loading a bit more cleanly

rgbkrk:
This PR makes the status bar show `loading... | not connected` instead of `| not connected` and helps webpack with the initial startup.


jdetle:
Awesome, thank you!

PR #1151, opened 2016-11-07
chore(deps): include testing dependencies in node side

rgbkrk:
Mostly to prevent scary warnings for new contributors, this adds React and React DnD to devDeps in the root of our electron builder setup.


rgbkrk:
Not going to press forward with this one.

PR #1155, opened 2016-11-07
chore(deps): bump electron-context-menu and eslint a11y

rgbkrk:
This bumps a couple dependencies. The extra package.json changes came from using `npm-check -u` (just formatting).


jdetle:
Thank you, oh human greenskeeper.

PR #1158, opened 2016-11-08
refactor(outputs): moved output appending to redux

ivanov:


rgbkrk:
Thanks @ivanov! This sets us up to be able to handle https://github.com/jupyter/jupyter_client/issues/209 more easily.

PR #1159, opened 2016-11-08
refactor(cell): Move cell status indicators to metadata key

### LINKED ISSUE #1148: outputExpanded should not be in the top-level cell data
Right now it's appearing as a sibling of `outputs`, `cell_type`, etc. when it should be under `metadata`. This is making our notebooks no longer conformed to nbformat.

captainsafia:
Closes #1148.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -86,7 +86,10 @@ export function publishNotebookObservable(github, notebook, filepath,
     const notebookString = JSON.stringify(
       commutable.toJS(notebook.update('cellMap', cells =>
         cells.map(value =>
-          value.delete('inputHidden').delete('outputHidden').delete('status')))),
+          value
+            .deleteIn(['metadata', 'inputHidden'])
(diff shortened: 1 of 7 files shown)


src/notebook/epics/github-publish.js
```diff
@@ -86,7 +86,10 @@ export function publishNotebookObservable(github, notebook, filepath,
     const notebookString = JSON.stringify(
       commutable.toJS(notebook.update('cellMap', cells =>
         cells.map(value =>
-          value.delete('inputHidden').delete('outputHidden').delete('status')))),
+          value
+            .deleteIn(['metadata', 'inputHidden'])
+            .deleteIn(['metadata', 'outputHidden'])
+            .delete(['metadata', 'status'])))),
```
rgbkrk, line 92: While not a big deal, we could switch these to `remove` (it's an alias in ImmutableJS).

rgbkrk:
Thank you for tackling this one and cleaning it all up!

PR #1160, opened 2016-11-08
dx(logger): remove predicate for COMM

rgbkrk:
To enable logging while working in dev mode:

```
DEBUG=true npm run start
```

/cc @peggyrayzis


peggyrayzis:
Thanks so much for fixing this! 👍

PR #1162, opened 2016-11-09
feat(display): allow updating displays

rgbkrk:
The nteract implementation of jupyter/jupyter_client#209

- `display(id=...)` always adds something to the document
- `update(id=...)` updates only existing things on the document

A single `display_id` can be used multiple times on the page. All instances are updated and kept in sync.

The motivation for this Jupyter protocol addition is to allow updating simple outputs like progress bars where previously users would need to write bidirectional widgets. This provides a very simple API to update the view of some data/output.

### Simple Progress Bar

progress

### Emoji animation

emoji

## TODO

* [x] shouldComponentUpdate logic for all transforms (`props.data` can now change)
  * [x] `text/html`
  * [x] `text/plain`
  * [x] `application/json` /cc @aleury 
  * [x] vega
  * [x] plotly
  * [x] markdown
  * [x] javascript
  * [x] latex
  * [x] image
  * [x] geojson
* [x] Tests
* [x] Cleanup cells when:
  * [x] removed
  * [x] executed on again
  * [x] output cleared (`output_clear` jupyter message, by user, executed on again)
  * [x] change cell type
* [x] Include a notebook that demonstrates how this is useful/used

## Beyond

The other thing we'll want to do is handle cut cells. I'm not sure what the behavior should be.


jdetle:
@rgbkrk @ivanov, Could you provide an explanation here of the additions made to message between the kernel and the notebook. As a part of the explanation, could you mention the problems this PR is meant to solve (and what it isn't meant to solve)? I know we've been over it, but I'd love to have something here for the public record :)

jdetle:
Posted something without reading... oops.

ivanov:
Good idea, @jdetle ! The idea here is to be able to created _named_ display outputs which can be updated later using that name... We will add some tests, and the longer explanation is in https://github.com/jupyter/jupyter_client/issues/209 with a jupyter notebook implementation in  https://github.com/jupyter/notebook/pull/1879 .

jdetle:
I guess it isn't a terrible idea to have a TL;DR, but admittedly the info is there to be found pretty easily, just didn't see it at first glance.

ivanov:
@rgbkrk is updating the summary to be more informative.

jdetle:
Awesome, thanks.

rgbkrk:
progress

rgbkrk:
Got the clean up working, even if a bit inefficient as a first pass (see bd28431).

rgbkrk:
To play with this PR, in your notebook you'll want:

``` python
from IPython.display import display

ip = get_ipython()
iopub = ip.kernel.iopub_socket
session = ip.kernel.session

def display_with_id(obj, display_id, update=False):
    """Create a new display with an id"""
    data, md = ip.display_formatter.format(obj)
    transient = {'display_id': display_id}
    content = {
        'data': data,
        'metadata': md,
        'transient': transient,
    }
    msg_type = 'update_display_data' if update else 'display_data'
    session.send(iopub, msg_type, content, parent=ip.parent_header)

def update_display(obj, display_id):
    """Update existing displays with id"""
    return display_with_id(obj, display_id, update=True)
```

### ProgressBar, traitlets based

``` python
import os
from binascii import hexlify
import time
import traitlets

from traitlets import observe, Int, HasTraits

class ProgressBar(HasTraits):
    capacity = Integer(100, help="capacity for progress")
    progress = Integer(0, help="value of progress")

    def __init__(self, capacity):
        self._display_id = hexlify(os.urandom(8)).decode('ascii')

    def _repr_html_(self):
        return "<progress max='{}' value='{}'></progress>".format(self.capacity, self.progress)

    def display(self):
        display_with_id(self, display_id=self._display_id)

    @observe('capacity', 'progress')
    def update(self, change):
        update_display(self, display_id=self._display_id)
```

Invoking:

``` python
bar = ProgressBar(100)
bar.display()

for i in range(101):
    bar.progress = i
    time.sleep(0.01)
```

### Emoji Updater, also traitlets based

``` python
import os
from binascii import hexlify

from traitlets import observe, HasTraits

class Emoji(HasTraits):
    emoji = traitlets.Unicode('😁', help="emoji to show")

    def __init__(self):
        self._display_id = hexlify(os.urandom(8)).decode('ascii')

    def _repr_html_(self):
        return '<h1 style="font-size: 100px">{}</h1>'.format(self.emoji)

    def display(self):
        display_with_id(self, display_id=self._display_id)

    @observe('emoji')
    def update(self, change):
        update_display(self, display_id=self._display_id)

em = Emoji()
em.display()
```

Invoking:

``` python
emojis = "😃😄😅😆😃"
for i in emojis:
    em.emoji = i
    time.sleep(0.5)
```

rgbkrk:
Animation, using some choices from my kids:

``` python
frames = [
    "😃 😺",
    "😍 😸",
    "😘 😻",
    "😍 😻",
    "😃 😺",
    "😎 😸",
    "🤗 😹",
    "🙂 😸",
]
em = Emoji()
em.display()

for i in frames*2:
    em.emoji = i
    time.sleep(0.5)
```

emoji-frames

test/renderer/epics/execute-spec.js
```diff
@@ -310,3 +311,70 @@ describe('executeCellEpic', () => {
     );
   });
 })
+
```
rgbkrk, line 316: This is my favorite test that I've written in this code base so far.

ivanov:
Ok, I committed my explanatory notebook, but there's a problem with the updating HTML displays. Run through the example notebook (now in the menu list) to see what I mean.

rgbkrk:
This notebook has a really great narrative Paul. Love it.

The last part of the notebook is where the update display doesn't seem to work, because 
There's only one spot in the notebook that doesn't update, yet I expected it -- we don't have support for `execute_result` nor a display id property.

See https://github.com/jupyter/jupyter_client/issues/209#issuecomment-260182391

rgbkrk:
Updated the notebook, love the 🏇 race.

progress-race

Also noticed something wrong in the notebook format related to fields that should be in metadata. I'll get that fixed on master.

rgbkrk:
@ivanov do you think it would be ok for us to introduce readers of the notebook to traitlets to make updates automatic on change of progress + capacity?

ivanov:
I actually had a similar question: should we have the ProgressBar class handle the updates right out of the gate, or should we have another example, so that they all build on each other.

I was thinking of doing it with `@property`  and calling `update` in the `@property.setter` - because at least properties come with the language, so if a user's never seen them before, they are being exposed to a core feature in the language.

rgbkrk:
Oooooh I like the idea of using `@property` and `@property.setter` here! Less cognitive burden (and dependencies) for folks.

rgbkrk:
> should we have the ProgressBar class handle the updates right out of the gate, or should we have another example, so that they all build on each other.

If we create a new ProgressBar later on that introduces the use of the property decorator that's cool.

Possible Flow:
- Outputs you can update by name
- Creating a Progress Bar
- Progress Bar that uses `@property` and `@property.setter`
- ThreadPool Executor linked up with the progress bar for jobs

ivanov:
cool, I'll add on to the end. 

Unrelated to this - but are you seeing the JSON key-order churn in our git history? Is this something we can (or want to) handle in commutable to keep diffs down to a minimum in the future?

rgbkrk:
I'm not sure if we can do much about the JSON key order churn. If it's simple enough to do while running a `JSON.stringify`, might as well add it on.

rgbkrk:
Two cells with a threadpool + progress bar:

``` python
import time
def progressive(bar, timeout=1):
    for ii in range(bar.capacity):
        bar.progress = ii
        bar.update()
        time.sleep(timeout)
    bar.progress = 100
    bar.update()

from concurrent import futures
tp = futures.ThreadPoolExecutor(max_workers=10)
```

``` python
bar1 = ProgressBar(100)
bar1.display()
tp.submit(progressive, bar1, 1)
```

Results:

threadpool

/cc @glyph

glyph:
Is there any example of how to do this without threads?

rgbkrk:
What do you have in mind?

ivanov:
Ok, I added a progress bar that calls update whenever its progress is modified:

``` python
class AutoupdatingProgressBar(ProgressBar):
    @property
    def progress(self):
        return self._progress

    @progress.setter
    def progress(self, value):
        self._progress = value
        self.update()
```

The horse race at the end now uses this version, so doesn't have to keep calling update.

rgbkrk:
@glyph - as long as the message can be passed through to indicate that you want to print to a named display (`display_id=...`) then it will get directed to the right cell. This has to be for structured displays/objects - it won't work for stdout/stderr/print out of the box (unless you override the `print` function). I used a thread pool because it was the easiest. If you show me another paradigm you want (you mentioned background callables), we can try that out.

rgbkrk:
Just read the update to the notebook

> This is good, but it would be awesome to have a progress bar that would automatically update whenever its progress was modified - that would be truly _progressive_.

hahahahahah

rgbkrk:
Time for me to tackle a few more mimetypes. I'll take images and plotly now.

rgbkrk:
Rebased to bring in the testing changes from @alexandercbooth.

rgbkrk:
Next up, javascript.

rgbkrk:
This now implements everything we set out to do in the TODOs above and is thoroughly ready for review.

jdetle:
👏 :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap: :clap:

PR #1171, opened 2016-11-11
refactor(saveDialog): Simplify remote electron require

lgeiger:
Remove unnecessary `require('electron')`

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -103,7 +103,7 @@ mock('electron', {
     },
     'dialog': {
       'showSaveDialog': function(config) { },
-    }
+    },
     'BrowserWindow': {
       'getFocusedWindow': function() {
         return {
--- a/test/setup.js
+++ b/test/setup.js
@@ -101,15 +101,9 @@ mock('electron', {
         throw Error('not mocked');
       }
     },
-    'require': function(module) {
-      if (module === 'electron') {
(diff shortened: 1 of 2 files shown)


test/setup.js
```diff
@@ -101,15 +101,9 @@ mock('electron', {
...
-        return {
-          'dialog': {
-            'showSaveDialog': function(config) { },
-          }
-        };
-      }
-    },
+    'dialog': {
+      'showSaveDialog': function(config) { },
+    }
```
rgbkrk: Missing a comma right here which causes tests to fail.

rgbkrk (changes requested):

test/setup.js
lgeiger: Ooops. Actually the mock is unnecessary right now but it could be helpful for future tests.

rgbkrk (approved):

test/setup.js
lgeiger: I should've added a `caution-late-night-PR` label 😂

rgbkrk: Heh.

rgbkrk:
Are there any more like this?

lgeiger:
> Are there any more like this?

I don't know. Discovered this at random.

PR #1172, opened 2016-11-11
fix(messaging): ofMessageType takes an array of message types

rgbkrk:
Noticed this while working on #1162. This adapts all our code that erroneously uses `ofMessageType(type)` to use `ofMessageType([type])`. We should make `ofMessageType` allow for either. For now though, this fixes what we have (as for why this worked sometimes - it would filter valid and invalid yet the plucks here would not find the right content keys).


rgbkrk:
@ivanov @jdetle - We can/should rebase our display PR after this is merged.

PR #1176, opened 2016-11-12
chore(packaging): Use one package.json structure

lgeiger:
Thanks to the excellent work from @develar we can now use only one `package.json` file.


src/main/menu.js
```diff
@@ -6,9 +6,14 @@ import { installShellCommand } from './cli';
 
 const kernelspecs = require('kernelspecs');
 
-const exampleNotebooksDirectory = path.resolve(
-  path.join(__dirname, '..', '..', '..', 'example-notebooks')
-);
+function getExampleNotebooksDir() {
+  if (process.env.NODE_ENV === 'development') {
```
lgeiger, line 10: It took me way too long to figure out that `process.env.NODE_ENV === 'production'` doesn't work, despite dev tools telling me `NODE_ENV` is set to `production`. 🙈

lgeiger:
I have no idea why CI fails with:

```
prebuild ERR! stack Error: 404 response downloading https://iojs.org/download/release/v1.4.6/iojs-v1.4.6.tar.gz
```

develar:
Hint — check install script in a package.json of dependency. I will check your case.

lgeiger:
It's strange, I can't reproduce this on my local machine.
The problem arises when `electron-builder` tries to rebuild `zeromq`.

To rebuild it runs 

```
npm run install -- --disturl=https://atom.io/download/electron --target=1.4.6 --runtime=electron --arch=x64 --build-from-source
```

since zeromq uses [`prebuild`](https://github.com/mafintosh/prebuild) this will be expanded to

```
prebuild --install --preinstall "npm run build:libzmq" "--disturl=https://atom.io/download/electron" "--target=1.4.6" "--runtime=electron" "--arch=x64" "--build-from-source"
```

The problem is `target` is a option for [`prebuilt`](https://github.com/mafintosh/prebuild#help) defining the Node.js version to build against. Hence we get a download error because there is now Node.js version 1.4.6.

@develar Would it be better to use

```
npm rebuild --disturl=https://atom.io/download/electron --target=1.4.6 --runtime=electron --arch=x64 --build-from-source
```

instead of `npm run install` for rebuilding?

develar:
`npm rebuild` works for the whole project, but we need to rebuild only and only one dependency. We specify `runtime`, so, error is not clear for me. I will investigate.

rgbkrk:
We technically need to skip the pre build entirely for electron anyways.

@ralphtheninja - are there any workarounds for prebuild with the style electron builder uses?

lgeiger:
@develar Thanks a lot for your help. `electron-builder` is working really great for us.

With `prebuild` one can provide prebuilt binaries for different node versions. For example `prebuilt --target 7.0.0` would build the native module against node 7.0.0. Though this doesn't work for electron since `prebuild` doesn't support the `--runtime` flag.

One can also rebuild only a specific node module by running `npm rebuild <module name>`.

I still don't know why the rebuilding works on my machine (OS X) just fine but doesn't work on Travis and AppVeyor. I'll investigate soon.

develar:
https://github.com/mapbox/node-pre-gyp should be used instead of prebuild. In 2 days I will investigate why npm rebuild works, but electron-builder doesn't.

Thanks for note that npm rebuild accepts module name — probably it should be used (yarn support is another question).

develar:
electron-builder 8.4.0 (will be published in 10 minutes) uses `npm rebuild m1, m2, mN...` way to rebuild native production dependencies. Thanks @lgeiger for tip. First approach is still used for `yarn`. In any case `yarn` is not recommend and only `npm` is proven. So, in case of build issues, electron-builder is not responsible any more :) (at least in the area of how to invoke rebuild). Build configuration is passed in the env as before (i.e. as in case of full install).

lgeiger:
Wow @develar you are amazing! Thanks!

ralphtheninja:
> We technically need to skip the pre build entirely for electron anyways.
> 
> @ralphtheninja - are there any workarounds for prebuild with the style electron builder uses?

I don't know. Maybe @mafintosh or @juliangruber can shed some light on this?

develar:
> are there any workarounds for prebuild with the style electron builder uses?

electron-builder follows electron recommendation and by default always pass `--build-from-source` arg. Prebuilt binaries are not used by default. You can set [npmSkipBuildFromSource](https://github.com/electron-userland/electron-builder/wiki/Options#BuildMetadata-npmSkipBuildFromSource) to `true`.

lgeiger:
Woo Travis is green 💚 

This is ready to test now 🎉

README.md
```diff
@@ -108,7 +108,7 @@ npm run build:renderer:watch
 and
 
 ```
-NODE_ENV=development ./node_modules/electron/cli.js ./app/
+npm run start
```
rgbkrk: This still runs the prestart (which invokes a build) so we want to keep this the same (except without the `app` line).

index.js
```diff
@@ -1,2 +0,0 @@
-// Necessary for bootstrapping electron, must be ES5
-require('./app/build/main/index.js');
```
rgbkrk, line 2: This file will still need to exist for the released copy, yet pointing to `./lib/main/index.js`.

package.json
```diff
@@ -63,6 +84,53 @@
...
+    "shell-env": "^0.2.0",
+    "source-map-support": "^0.4.5",
+    "spawn-rx": "^2.0.3",
+    "spawnteract": "^2.2.0",
+    "uuid": "^2.0.3",
+    "vega": "^2.6.3",
+    "vega-embed": "^2.2.0",
+    "vega-lite": "^1.3.1",
+    "yargs": "^6.3.0"
+  },
```
rgbkrk, line 134: Yay for back in one `package.json`!

rgbkrk (changes requested):

rgbkrk:
This is awesome. If it's working, I'm happy to just merge it and move forward. The only necessary change is bringing back `index.js` (I think).

index.js
lgeiger: The release build by `electron-builder` works just fine without this. For this I changed the `main` field in the `package.json` file.

Or am I missing something?

README.md
lgeiger: Oh I see.

package.json
lgeiger: No need for human greenkeeper 😉

index.js
rgbkrk: Great, if we don't need this anymore then sweet!

README.md
```diff
@@ -108,7 +108,7 @@ npm run build:renderer:watch
 and
 
 ```
-npm run start
+npm run spawn
```
rgbkrk, line 111: 👍

rgbkrk (approved):

rgbkrk:
Time for some rebasing on my own PRs.

Issue #771, opened 2016-09-15
Style scrollbars according to theme


rgbkrk:
[It can be done](https://css-tricks.com/custom-scrollbars-in-webkit/) and certainly needs to for the dark theme.

prafful002:
@rgbkrk may I have this assigned to me ?

rgbkrk:
@prafful002 Since you've posted the same across several issues - pick one, make a pull request, then we close the issue.

prafful002:
@rgbkrk sure !

kosmastsk:
Is this issue still open?

rgbkrk:
Yes

jkornblum:
Can be closed with pr #1184

rgbkrk:
Thanks again @jkornblum!

PR #1184, opened 2016-11-14
(style) styles scrollbars according to theme

jkornblum:
This address issue #771 and adds a custom style scroll bar to the light, dark, classic, and nteract themes.

I know this is a style and subjective so I'm open to any suggestions for changes. If no suggestions this should be an upgrade from the default chrome style scroll bars throughout the application.

It is a flat style, which shows a small track using css box-shadow. The scroll bar color is a translucent version of the --main-fg-color (for dark themes) or --cm-background (for light themes). The scrollbar styling only applies inside the Electron window, i.e. the "main" window itself keeps the default platform style. 

Here is a preview
scrollbars

### DIFF
--- a/static/styles/theme-classic.css
+++ b/static/styles/theme-classic.css
@@ -142,3 +142,24 @@
 .dropdown__content ul {
   box-shadow: -1px 1px 5px rgba(0,0,0,0.2);
 }
+
+/*custom scrollbar for notebook cells and hint box*/
+div::-webkit-scrollbar, ul::-webkit-scrollbar {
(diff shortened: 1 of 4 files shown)


rgbkrk:
This is looking excellent to me, certainly an improvement. Thank you. We can keep refining after so feel free to take the WIP out of the title here if it's ready to go!

rgbkrk (approved):

jkornblum:
Sure thing. PR ready to be merged.

Thanks.

PR #1186, opened 2016-11-14
[WIP] fix(messaging): Store session id in state

captainsafia:
Second attempt at closing #1087.


captainsafia:
This is different from the previous implementation because it now stores the `sessionId` in `kernel.channels`.

src/notebook/epics/kernel-launch.js
```diff
@@ -70,6 +70,7 @@ export function newKernelObservable(kernelSpecName, cwd) {
           iopub: createIOPubSubject(identity, config),
           control: createControlSubject(identity, config),
           stdin: createStdinSubject(identity, config),
+          sessionId,
```
rgbkrk, line 85: I would go ahead and generate the `sessionID` here so it's generally available with the channel rather than needing to propagate it with loading files in general. If we have a new kernel, we have a new session.

captainsafia:
OH WOW I REALLY SCREWED UP THIS REBASE. CLOSING. BYE. OH NO. NEVER AGAIN. 😨

PR #1193, opened 2016-11-15
chore: lint until the pain goes away

rgbkrk:
Fixing up the build so other PRs can rebase against master with this.


rgbkrk:
Nevermind, everything is fine now because we merged in the mega cell focus PR.

PR #1197, opened 2016-11-15
chore(test): Added more tests for editor focus actions

peggyrayzis:
I added some basic tests to cover the editor focusing actions. Please let me know if there are any other tests related to the editor focusing that you would like me to add!


rgbkrk:
Awesome, great to have more tests!

PR #1198, opened 2016-11-15
[WIP] Allow execute result

rgbkrk:
Update display data can also target `execute_result`. There's more to do here, as evidenced by the test I put in that fails (on purpose).


rgbkrk:
I'll go ahead and close this and open an issue about supporting this.

PR #1203, opened 2016-11-16
Fix execute error

jdetle:
Right now I'm really having trouble understanding why the test is failing in line 272 of execute-spec.

### DIFF
--- a/src/notebook/epics/execute.js
+++ b/src/notebook/epics/execute.js
@@ -231,13 +231,11 @@ export function createExecuteCellStream(action$, store, source, id) {
       state.app.executionState === 'not connected');
 
   if (!kernelConnected) {
-    // TODO: Switch this to dispatching an error
-    state.app.notificationSystem.addNotification({
-      title: 'Could not execute cell',
-      message: 'The cell could not be executed because the kernel is not connected.',
-      level: 'error',
+    return Rx.Observable.of({
+      type: ERROR_EXECUTING,
+      payload: 'Kernel not connected!',
+      error: true,
     });
-    return Rx.Observable.of(updateCellExecutionCount(id, undefined));
   }
 
   return executeCellStream(channels, id, source)
--- a/test/renderer/epics/execute-spec.js
+++ b/test/renderer/epics/execute-spec.js
@@ -228,7 +228,8 @@ describe('createExecuteCellObservable', () => {
             state: {
               app: {
                 executionState: 'starting',
-                channels: 'channelInfo',
+                // TODO must mock channels extensively here
+                channels: '',
                 notificationSystem: {
                   addNotification: sinon.spy(),
                 },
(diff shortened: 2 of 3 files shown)


jdetle:
Getting an error 

```
 const action$ = ActionsObservable.from([{type: 'EXECUTE_CELL', id: 'id' },
                                  {type: 'EXECUTE_CELL', id: 'id_2' },
                                  {type: 'ABORT_EXECUTION', id: 'id' },
                                  {type: 'EXECUTE_CELL', id: 'id' }]);
```

@rgbkrk, getting an error saying `action$.filter(...).ofType` is not a function, any idea whats going on here?

No operators for an `ActionsObservable` built in!

jdetle:
https://github.com/redux-observable/redux-observable/blob/master/docs/Troubleshooting.md

rgbkrk:
> @rgbkrk, getting an error saying action$.filter(...).ofType is not a function, any idea whats going on here?

That sounds like the observable prototype isn't being patched with the filter operator. This tends to happen if two different versions of Rx are installed locally (one for redux-observable and one for our base app). There are ways to work around it, I'd dig into what you have locally though.

jdetle:
Alright thanks, I'm having issues applying the fix that the redux-observable folks suggest so I'd appreciate any help.

jayphelps:
Discussed with @jdetle on gitter, but recording here for posterity. The issue is that redux-observable doesn't currently implement its own `ActionsObservable.from` so it inherits it from `Observable.from`, losing the `.ofType` prototype extension. tl;dr Static operators don't currently use the `lift` mechanism like prototype operators do, but we've [been discussing solving that in RxJS](https://github.com/ReactiveX/rxjs/issues/1829).

For this PR, you can just switch to using `ActionsObservable.of`:

``` js
const action$ = ActionsObservable.of(
  { type: 'EXECUTE_CELL', id: 'id' },
  { type: 'EXECUTE_CELL', id: 'id_2' },
  { type: 'ABORT_EXECUTION', id: 'id' },
  { type: 'EXECUTE_CELL', id: 'id' }
);
```

---

We will override `from` in redux-observable, if either of you want to PR it, I believe this works (but needs tests to prove):

``` js
class ActionsObservable extends Observable {
  static from(ish, scheduler) {
    const input$ = Observable.from(ish, scheduler);
    return new ActionsObservable(input$);
  }

  // ... other existing stuff
}
```

rgbkrk:
TIL `of` can take more than one argument.

jdetle:
And it works! Thanks so much @jayphelps.

jayphelps:
@jdetle you're welcome!

party

src/notebook/epics/execute.js
```diff
@@ -231,13 +231,11 @@ export function createExecuteCellStream(action$, store, source, id) {
...
   if (!kernelConnected) {
-    // TODO: Switch this to dispatching an error
-    state.app.notificationSystem.addNotification({
-      title: 'Could not execute cell',
-      message: 'The cell could not be executed because the kernel is not connected.',
-      level: 'error',
+    return Rx.Observable.of({
+      type: ERROR_EXECUTING,
+      payload: 'Kernel not connected!',
+      error: true,
```
rgbkrk, line 237: :+1:

test/renderer/epics/execute-spec.js
```diff
@@ -223,42 +227,58 @@ describe('createSourceUpdateAction', () => {
     });
   });
 });
-describe('createExecuteCellObservable', () => {
+describe('createExecuteCellStream', () => {
+  const frontendToShell = new Rx.Subject();
+  const shellToFrontend = new Rx.Subject();
+  const mockShell = Rx.Subject.create(frontendToShell, shellToFrontend);
```
rgbkrk: Good naming here.

However, I'd do this setup and tear down in each `it` block or use a `beforeEach`.

rgbkrk (changes requested):

rgbkrk (approved):

rgbkrk:
Thanks @captainsafia and @jdetle!

Issue #1165, opened 2016-11-10
File save indicator is backwards


lgeiger:
Take a look at the red close button:
![nteract](https://cloud.githubusercontent.com/assets/13285808/20195931/bc07fcc6-a798-11e6-869e-e44c9a65b75c.gif)

jdetle:
Hey @lgeiger did some things get cut off in this gif?

lgeiger:
Only the saving and kernel restart dialogs are cut off.
The save notifications are also correct. I triggered them with `cmd-s`.

rgbkrk:
Fixed in #1204.

PR #1204, opened 2016-11-17
Fix modified indicator

lgeiger:
This addresses part of #1165

But it still fails in a interesting way:
modified

### DIFF
--- a/src/notebook/native-window.js
+++ b/src/notebook/native-window.js
@@ -39,11 +39,18 @@ export function createTitleFeed(state$) {
       modified: false,
       notebook: state.document.get('notebook'),
     }))
+    .distinctUntilChanged(last => last.notebook)
     .scan((last, current) => ({
+      // We're missing logic for saved...
+      // All we know is if it was modified from last time
+      // The logic should be
+      //    modified: saved.notebook !== current.notebook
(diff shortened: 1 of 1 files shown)


src/notebook/native-window.js
```diff
@@ -36,7 +36,7 @@ export function createTitleFeed(state$) {
   const modified$ = state$
     .map(state => state.document.get('notebook'))
     .scan((prev, notebook) => ({
-      modified: prev.notebook === notebook,
+      modified: prev.notebook !== notebook,
```
lgeiger:
``` javascript
Immutable.is(prev.notebook, notebook) === false
```

will behave the same.

rgbkrk:
I just realized... We're missing logic for saved. All we know is if it was modified from last time.

The logic should be

``` js
  modified: saved.notebook !== current.notebook
```

Yet we don't have `saved.notebook` available here.

jdetle:
Is this still WIP?

rgbkrk:
It fixes a bug yet doesn't fix the modified feature. It can be merged, we've just got some work ahead of us. I think I know what to do for a follow up PR.

ivanov:
Merging, thank you!

Issue #963, opened 2016-10-09
configurable keybindings


tacaswell:
In particular, to ship with the default emacs keybindings :smiling_imp: .

https://github.com/rmcgibbo/jupyter-emacskeys maybe relevant?

rgbkrk:
That definitely helps. We have a config setup that we're planning to expose for CodeMirror. Part of the resulting config for you includes loading CodeMirror's built in emacs mode `'codemirror/keymap/emacs'` and then setting up the config.

In the meantime, here's a diff you can apply on master (probably to be out of date soon):

``` diff
diff --git a/src/notebook/components/cell/editor/index.js b/src/notebook/components/cell/editor/index.js
index 7742614..e82204f 100644
--- a/src/notebook/components/cell/editor/index.js
+++ b/src/notebook/components/cell/editor/index.js
@@ -17,6 +17,8 @@ import 'codemirror/addon/edit/closebrackets';

 import 'codemirror/addon/dialog/dialog';

+import 'codemirror/keymap/emacs';
+
 import 'codemirror/mode/javascript/javascript';
 import 'codemirror/mode/css/css';
 import 'codemirror/mode/julia/julia';
@@ -245,9 +247,15 @@ export default class Editor extends React.Component {
           Right: pick,
         },
       },
-      extraKeys: {
+      extraKeys: Object.assign({
         'Ctrl-Space': 'autocomplete',
-      },
+      }, CM.keyMap.emacs, {
+        'Ctrl-Y': (cm) => {
+          const originalBehavior = CM.keyMap.emacs['Ctrl-Y'];
+          originalBehavior(cm);
+          cm.setSelection(cm.getCursor(), cm.getCursor());
+        }
+      }),
       indentUnit: this.props.tabSize,
       tabSize: this.props.tabSize,
     };
```

(@tacaswell and I sat down at PyData DC to figure this out)

This definitely motivates that we need a js based config on the desktop app, similar to ~~hyperterm~~ hyper.app so that people can write custom functions for codemirror.

rgbkrk:
I guess if it's just a package plugin with namespace `nteract-codemirror-emacs`, then the user could add it to their plugins list... 🤔

rgbkrk:
Definitely need an expansion on how we do our config file, likely to take a raw javascript file.

rgbkrk:
[...]

PR #1205, opened 2016-11-17
feat: toggle the blinking of cursor editor

ivanov:
We will want to refactor this so passing around CodeMirror editors is less painful in the future. 

It would also be nice for the menu items to have access to state, so we could have a checkbox reflecting which theme is used, and whether or not the cursor blinking is enabled.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -434,6 +434,14 @@ export const view = {
       label: 'Theme',
       submenu: theme_menu,
     },
+    {
+      label: 'No Blink Editor Cursor',
+      click: createSender('menu:set-blink-rate', 0),
+    },
+    {
+      label: 'Blink Editor Cursor',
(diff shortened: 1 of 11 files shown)


rgbkrk:
Yeah, we should definitely figure out how to encode the entire codemirror config from within the nteract config file. I feel like we'll probably have to switch to `nteract.js` based on #963.

In the future I'd love to see us do the config from the main process, pushing changes to all the notebook/renderer processes. We'd then be able to handle the checkbox state appropriately.

rgbkrk:
This whole non-checkbox thing is so annoying. :-/

Also want this type of setting available to turn code completion typeahead off.

rgbkrk:
Thanks for adding tests in here that end up increasing coverage in general (the `componentDidUpdate` and `componentShouldUpdate` cases).

PR #1207, opened 2016-11-17
fix(bracketMatcher): Don't obstruct cursor

lgeiger:
master:
bildschirmfoto 2016-11-17 um 00 18 45
PR:
bildschirmfoto 2016-11-17 um 00 18 18


ivanov:
Thank you, this is lovely! 🍰

PR #1208, opened 2016-11-17
chore: bump deps, don't let greenkeeper upgrade eslint plugins

rgbkrk:
We need to handle upgrading eslint, eslint config, and eslint plugins directly and manually (while making sure they match the right versions from eslint airbnb config). I've made it so we don't get separate PRs from greenkeeper for this.


lgeiger:
Oops I just introduced conflicts by merging #1199. I'm going to rebase.

rgbkrk:
> Oops I just introduced conflicts by merging #1199. I'm going to rebase.

All good, thanks!

PR #1210, opened 2016-11-17
test(fsObs): Test different fs observable executions.

jdetle:
Need to add more mocks

### DIFF
--- a/test/renderer/api/kernel-spec.js
+++ b/test/renderer/api/kernel-spec.js
@@ -3,13 +3,13 @@ import chai, { expect } from 'chai';
 var mock = require('mock-require');
 const sinon = require('sinon');
 const sinonChai = require("sinon-chai");
-import * as fs from 'fs';
 chai.use(sinonChai);
 
 import {
   shutdownKernel,
   forceShutdownKernel,
   cleanupKernel,
+  filesystem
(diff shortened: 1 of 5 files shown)


jdetle:
@rgbkrk @lgeiger @captainsafia, ready for a revue

test/renderer/api/kernel-spec.js
```diff
@@ -31,26 +31,32 @@ const emptyKernel = Object.freeze({
...
 }
 
 describe('forceShutdownKernel', () => {
+  var sandbox;
+  beforeEach(function() {
+    sandbox = sinon.sandbox.create();
+  });
+  afterEach(function() {
+    sandbox.restore();
+  });
```
lgeiger, line 56: Would it work if we move `beforeEach` and `afterEach` to the outside of the `describe` function? This way we don't have to repeat them.

lgeiger (approved):
Nice 🎉

test/renderer/api/kernel-spec.js
rgbkrk: In this case, in tests, I think it's fine to repeat the setup code for the `beforeEach` and `afterEach`.

rgbkrk (approved):

PR #1212, opened 2016-11-18
[WIP][DISCUSSION] feat: "vdom" mimetype

rgbkrk:
Experimenting with a Virtual DOM-esque JSON mimetype for rendering that won't trash the DOM on updates. Not to be merged, only for discussion.

screen shot 2016-11-17 at 9 44 16 pm

/cc @captainsafia @ivanov

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "redux-actions": "^0.13.0",
     "redux-logger": "^2.7.4",
     "redux-observable": "^0.12.0",
+    "reon-core": "0.0.1",
     "rxjs": "^5.0.0-rc.1",
     "shell-env": "^0.2.0",
     "source-map-support": "^0.4.5",
--- a/src/notebook/components/transforms/index.js
+++ b/src/notebook/components/transforms/index.js
@@ -26,6 +26,8 @@ import {
   Vega,
 } from './vega';
 
+import REONTransform from './reon';
+
 type StandardTransforms = ImmutableMap<string, any>
 type StandardDisplayOrder = ImmutableList<string>
 
@@ -61,14 +63,16 @@ const transforms = standardTransforms
   .set(PlotlyTransform.MIMETYPE, PlotlyTransform)
   .set(GeoJSONTransform.MIMETYPE, GeoJSONTransform)
   .set(VegaLite.MIMETYPE, VegaLite)
-  .set(Vega.MIMETYPE, Vega);
+  .set(Vega.MIMETYPE, Vega)
+  .set(REONTransform.MIMETYPE, REONTransform);
 
 // Register our custom transforms as the most rich (front of List)
 const displayOrder = standardDisplayOrder
   .insert(0, PlotlyTransform.MIMETYPE)
   .insert(0, VegaLite.MIMETYPE)
   .insert(0, Vega.MIMETYPE)
-  .insert(0, GeoJSONTransform.MIMETYPE);
+  .insert(0, GeoJSONTransform.MIMETYPE)
+  .insert(0, REONTransform.MIMETYPE);
 
 /**
  * Choose the richest mimetype available based on the displayOrder and transforms
--- a/src/notebook/components/transforms/reon.js
+++ b/src/notebook/components/transforms/reon.js
@@ -1,0 +1,30 @@
+/* @flow */
+/* eslint class-methods-use-this: 0 */
+import React, { PropTypes } from 'react';
+
+type Props = {
+  data: Object,
+};
+
+const objectToReact = require('reon-core/lib/object-to-react');
+
+const MIMETYPE = 'application/vnd.nteract.react.v1+json';
+
+class REONTransform extends React.Component {
+  props: Props;
+
+  static MIMETYPE = MIMETYPE;
+
+  render(): ?React.Element<any> {
+    try {
+      return objectToReact(this.props.data.toJS());
+    } catch(err) {
+      console.error(err);
+      return (
+        <code>{JSON.stringify(err, 2, 2)}</code>
+      );
+    }
+  }
+}
+
+export default REONTransform;


gnestor:
@rgbkrk I think this could be really useful if a `path` property were provided that would accept a module path or if a `require` property that would accept a require path. This would allow users to render data using any React component vs. just standard DOM elements. 

Imagine a `React` display method like:

``` python
from ipython.display import React
React('react-json-tree', {
  'data': {
    'foo': 'bar'
  }
})
```

I implemented a "React transform" for magic console that accepts a `path` (path to module) property, hot reloads the component when the source changes, and evaluates props so that callbacks/functions can be passed as a prop. You can see the implementation here: https://github.com/gnestor/magic-console/blob/master/plugins/ReactComponent.js

rgbkrk:
> this could be really useful if a path property were provided that would accept a module path or if a require property that would accept a require path

That would be really useful! However, that would also be very brittle and tied to the environment:
- What does it mean for non-electron, does a webpack bundling strategy have to be defined like in JupyterLab when on the web?
- How do you show this with nbconvert, on nbviewer, etc.?

As for callbacks, I definitely want to explore that space... 🤔

gnestor:
I was reading about [Carbide](https://alpha.trycarbide.com/):

> Originally, we had written our own tiny implementation of NPM which would run in the browser (calling the official NPM API endpoint, locating tarballs, unpacking them in the browser, and loading the files appropriately).
> 
> But over the past few months we discovered SystemJS, and ended up using JSPM to handle just about everything. In fact, nowadays Carbide's "Current Window" mode uses the same SystemJS loader for both the editor itself and the code which gets executed on it. This means that it'll technically be able to self-host, hot-reload, and instrument various parts of itself. That's kinda neat.

I haven't been able to get SystemJS/jspm to dynamically import an npm module from within the browser (e.g. in a Jupyter notebook) yet, but if that's possible like they're implying it is, then we could potentially expose a `module` property and the value would be an npm module name!

Regarding callbacks, ideally you would define a callback in the native language (e.g. Python) so that the kernel can respond to events in React. I really have no idea how to implement that but I think [jupyter-react](https://github.com/timbr-io/jupyter-react) may have figured it out. jupyter-react involves:
1. A [notebook extension](https://github.com/timbr-io/jupyter-react-example/blob/master/js/src/index.js) for initializing/registering React component(s) on the front-end
2. [Python classes](https://github.com/timbr-io/jupyter-react-example/blob/master/example/thing.py) for each React component to handle comm with its component on the front-end
3. [webpack](https://github.com/timbr-io/jupyter-react-example/blob/master/js/webpack.config.js) for building the JS

So it's a lot of work to get just render some data in a React component. However, I imagine that:
1. `display_id` could be used to send comm messages from kernel to front-end
2. The display method could handle comm message from front-end to kernel
3. _Maybe_ SystemJS could allow React components to be imported from npm in the browser (no webpack necessary)

The dream:

``` python
from ipython.display import React
query = ''
def handleChange(event):
  # how to serialize `event` for JS -> Python?
  query = event.target.value
  data = search(query)
React('input', {
  'placeholder': 'query',
  # `handleChange` becomes a JS function that sends a comm message requesting that `handleChange` be called with `event` arg
  'onChange': handleChange
})
```

rgbkrk:
This is some great direction @gnestor -- I like the jupyter react ideas. I've seriously hoped not to tie things straight to React, yet it seems like it would be the easiest way to experiment with writing the layout and user event handling portion to match the models I wrote a bit on in [synchronized-object-diff-spec](https://github.com/rgbkrk/synchronized-object-diff-spec)

gnestor:
Ok, let's see if I'm understanding everything:

The synchronized object spec describes a general way to synchronize state between front-end and kernel. When implemented, this will effectively make the core of ipywidgets much more accessible to other _widgets_ (because ipywidgets is a beast and it only works with the ipython kernel). React components fit nicely into this picture because there are so many of them and they can easily be represented using JSON. Assuming that we can simply take an npm module name and dynamically import a React component instance into the front-end, then users of any kernel (that implements this sync object spec) could visualize AND interact with data using any React component available by doing something like:

``` py
# In[1]
from display_lib import React
data=[]
def handleChange(event):
  data = get_data(event.target.value)
React('input', {
  'onChange': handleChange
})

# In[2]
React('react-json-tree', {
  'data': data
})
```

rgbkrk:
I'm currently 👎  on auto-importing node modules notebook side. I like the idea of it, it's the likely implementation and maintenance burden that would worry me, not to mention the new security implications.

> The synchronized object spec describes a general way to synchronize state between front-end and kernel. When implemented, this will effectively make the core of ipywidgets much more accessible to other widgets

Yeah, largely speaking I was iterating on it to explore the space of only the models part of the widgets (no views). You've got the right idea on handling the change sets. I'm not sure what the frontend ---> backend changes look like, how you'd propagate from nested React components. It's at least clear to me that changes should bubble up as actions which change the overall model (while informing the backend).

gnestor:
Update: I got dynamic imports of npm modules working in a lab/notebook extension!

screencast

To clarify, the extension is actually fetching the npm module in the browser, so now you can render data in a Jupyter notebook using **any React component on npm** with only the name and valid props. The really interesting things are:

* Rendering data is no longer dependent on the client or client extensions (meaning that users need not install an extension in order to render a specific type of data)
* React components can be rendered as static markup (nbviewer)
* Beyond the scope of this conversation, I see a big opportunity with inferring JSON schema from arbitrary data + querying React components that can render that schema = inferring visualization from arbitrary data.

The catches are:

* React version conflicts: https://github.com/systemjs/systemjs/issues/1511
* React components that rely on CSS styles (vs. inline styles)

It's using SystemJS to do the imports. The gist of the implementation is:

```js
import React from 'react';
import ReactDOM from 'react-dom';

function render({ type, props }, node) {
  System.import(type).then(type => {
    ReactDOM.render(React.createElement(type, props, null), node);
  });
}
```

The usage inside the notebook looks like this:

```py
from IPython.display import display

def React(type, props):
    bundle = {
        'application/vnd.react.v1+json': {
            'type': type,
            'props': props
        },
        'application/json': props,
        'text/plain': '<IPython.core.display.React.object>'
    }
    display(bundle, raw=True)

React('react-json-tree', {
    'data': [
        {'month': 'September', 'profit': 35000, 'loss': 2000},
        {'month': 'October', 'profit': 42000, 'loss': 8000},
        {'month': 'November', 'profit': 55000, 'loss': 5000}
    ],
    'theme': 'google'
})
```

..or if you want to use a non-default member of an import:

```py
React({
    'module': 'victory',
    'member': 'VictoryBar'
}, {
    'data': [
        {'month': 'September', 'profit': 35000, 'loss': 2000},
        {'month': 'October', 'profit': 42000, 'loss': 8000},
        {'month': 'November', 'profit': 55000, 'loss': 5000}
    ],
    'x': 'month',
    'y': 'profit'
})
```

rgbkrk:
Fantastic!

It's going to be hard to warm me up to dynamic imports from a CDN or otherwise.

What I am a fan of here is that a notebook implementation/deployment can choose to provide _some_ modules, even if some implementations do the wildly dynamic thing.

rgbkrk:
Also, amusingly, I was literally sitting down to think about an updating models structure with some sort of nested `updateIn` like mechanic. I'm not sure how to convey what I'm thinking of yet so I'll keep working at a whiteboard and come back with more.

rgbkrk:
> Beyond the scope of this conversation, I see a big opportunity with inferring JSON schema from arbitrary data + querying React components that can render that schema = inferring visualization from arbitrary data.

Great note.

gnestor:
> Beyond the scope of this conversation, I see a big opportunity with inferring JSON schema from arbitrary data + querying React components that can render that schema = inferring visualization from arbitrary data.

* [x] Inferring JSON schema from arbitrary data
  * http://jsonschema.net/#/
  * npm: https://github.com/krg7880/json-schema-generator
* [ ] Querying React component that can render a schema
  * Typescript to JSON schema: https://github.com/lbovet/typson
  * Flow aliases to JSON schema?  React PropTypes?
  * npm metadata spec for providing `jsonSchema: {member: schema}` (e.g. `member` being `VictoryBar` in `victory`) and keywords for finding packages that provide them (e.g. 'typed-react')

nelix:
This is very cool,  in regards to dependencies (and I guess, inline images, event handlers, whatever else) could the object not be walked and everything not a string/arr/obj/null gets daturi`ed and then the same thing on the other end... maybe a ref system for anything the hash datauri gets used twice?

rgbkrk:
Hi again @nelix! Haven't seen you since... pipboylib. :wink:

> could the object not be walked and everything not a string/arr/obj/null gets daturi`ed and then the same thing on the other end... maybe a ref system for anything the hash datauri gets used twice?

As long as we can carve out a consistent spec, I'm game for experimenting. How would/should we serialize functions for this? When writing these components from Python, I imagine they'd be pure strings (with some helper to demarcate as a function).

rgbkrk:
Super healthy discussion here, closing this for now to focus my own efforts. Happy to revisit this later or to have someone else jump in with this mimetype + example usage in Python. This pairs well with #1313.

PR #1224, opened 2016-11-20
Conformance

### LINKED ISSUE #1168: Create conformance tests for nteract created notebooks
From a comment in #1148 (https://github.com/nteract/nteract/issues/1148#issuecomment-258590140), introduce conformance tests by creating a notebook through nteract then using the notebook JSON Schema to validate that it is still a proper notebook.

rgbkrk:
Closes #1168.

This will currently fail since several of our notebooks are invalid currently. I'm going to get them cleaned up as well.


rgbkrk:
I went ahead and opened all the notebooks we had after making sure they conformed first - the only that I'm seeing here that persist that shouldn't now is `status` as a top level key inside a cell.

lgeiger:
Nice!

PR #1232, opened 2016-11-22
[ISSUE] Evaluate validity

rgbkrk:
I'm opening this up to highlight issues in our conformance of the notebooks themselves.

I toyed with creating a new notebook and performed various operations to see if there was anything else that was non-conformant. Turns out there is! (the conformance tests will fail here).

* [ ] Converting a code cell to a markdown cell does not delete the `outputs` property, resulting in an invalid markdown cell

I'm curious if there are others. What I'd like to see in the future is to run a series of actions through our reducers and validate that resulting notebook.


rgbkrk:
We'll need this one fixed before we make another release.

rgbkrk:
Fixed in #1233

PR #1233, opened 2016-11-22
fix: delete code keys from converted markdown cells

rgbkrk:
Doing the cleanup that I noticed in #1232. I'm thinking the change cell type op should use an empty code or markdown cell so it ensures all the metadata is cleaned out. Not real sure what's best here or if we want to carry some metadata over. ¯\\\_(ツ)\_/¯

### DIFF
--- a/src/notebook/reducers/document.js
+++ b/src/notebook/reducers/document.js
@@ -350,8 +350,8 @@ export default handleActions({
     }
 
     return cleanCellTransient(state.setIn(['notebook', 'cellMap', id, 'cell_type'], to)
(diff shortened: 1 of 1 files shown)


jdetle (approved):

rgbkrk:
This makes me hopeful for when we do redux or more tests that address the actual functionality more directly. This one was passed over sadly, the test that is there only ensures the cell type has changed - not that the additional fields are gone. We could start using the JSON schema that is set up for whole notebook to verify cells, outputs, etc. too. 🤔

Issue #1065, opened 2016-10-25
Document Cell Focus UX Design


rgbkrk:
I'd :heart: to see us write documentation for what a notebook is in user experience terms. Right now it's a bit implicit ("it's the implementation!") -- I think we can capture different behaviors and user intents in some solid documentation/specifications. Let's tackle Cell Focus UX design first!

@captainsafia is currently solving the input focus on click in code cells problem in #1055 (our current setup is rather pesky). Hopefully we can solve all our focus problems with a clear declaration + implementation.
## Cell Focus UX

![cell focus user experience](https://cloud.githubusercontent.com/assets/836375/19669086/48b6f02c-9a11-11e6-875a-d330fdb0f900.png)

User intents, written:
### `shift enter`
- Focus on next cell (and create it if it doesn't exist)
- Scroll to next cell
- Focus Editor at top of next cell (first line) if a code cell or a markdown cell in edit mode

Note: this is the same set of changes expected with the down arrow (for focus, not execution).
### `up arrow`
- Focus on previous cell
- Scroll to previous cell
- Focus editor at bottom of previous cell (last line) if a code cell or a markdown cell in edit mode
### `down arrow`
- Focus on next cell (and create it if it doesn't exist)
- Scroll to next cell
- Focus Editor at top of next cell (first line) if a code cell or a markdown cell in edit mode
### Click editor area
- Focus clicked cell
- Don't scroll
- Don't _manually_ focus the editor, let CodeMirror do the handling
### Click display area
- Focus clicked cell
- Don't scroll
- Don't focus the editor
### Drag cell
- Focus clicked cell
- Don't scroll
- Don't focus the editor

Since I by no means think this is a decree, I've drawn my initial thoughts on cell focus UX with crayon. We should refine this as necessary and produce a document. Happy to turn this into a Sketch document when ready.

/cc @willingc and @peggyrayzis who expressed interest in this

rgbkrk:
Hopefully outlining this helps us fix #429, #803.

captainsafia:
I like these user intents. It's unclear to me from the diagram but is the drag cell intent directed at the gutter out of the output area or the input area or both? The current implementation does it along the entire gutter.

I closed the PR I was working on (#1055) to start off a clean slate with implementing a solution for this. It looks like we'll have to manage multiple focused states for the CodeMirror editor and the CodeCell/MarkdownCell so there is lots of fun state juggling to be done. 😛

captainsafia:
[...]

PR #1250, opened 2016-11-30
[WIP][discuss] feature(notebook): Added key binding for esc to notebook

peggyrayzis:
This PR is to address the issue that @jkornblum brought up in Slack.

Here's the user experience in the Jupyter notebook:
1. User presses the Esc key in a cell with the editor focused
2. The cell's editor loses focus while the cell's focus is maintained
3. User presses up/down arrow keys to navigate 
4. The cell's editor never focuses - in Esc mode, only the cell is focused
5. At the end of the notebook, a new cell is not created

I recreated this experience by creating key bindings at the notebook level to handle Esc, Up & Down. This is problematic because we currently handle Up & Down at the cell level so we can navigate through the notebook while maintaining focus on the editor as described by @rgbkrk's user intents in #1065.

Please check out the branch and let me know what you think. One caveat - right now, you can't move between rendered MD cells and code cells without focusing the cell's editor because of how it's handled here: https://github.com/nteract/nteract/blob/master/src/notebook/components/cell/markdown-cell.js#L110-L124

It's going to get messy splitting that out so I wanted to get feedback before proceeding.

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -197,7 +197,13 @@ export class Notebook extends React.Component {
         e.preventDefault();
 
         const shiftXORctrl = (shiftKey || ctrlKey) && !(shiftKey && ctrlKey);
-        if (!shiftXORctrl) return;
+
+        if (!shiftXORctrl) {
+          if (!editorFocused) {
+            store.dispatch(focusCellEditor(id))
+          }
+          break;
+        }
 
         if (shiftKey) {
           store.dispatch(focusNextCell(id, true));
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -213,7 +213,7 @@ export class Notebook extends React.Component {
       }
       case 'ArrowDown': {
         if (!editorFocused) {
-          store.dispatch(focusNextCell(id, false));
+          store.dispatch(focusNextCell(id, true));
         }
         break;
       }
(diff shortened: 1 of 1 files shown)


peggyrayzis:
Whatever we decide, I want to make that function cleaner. 😝 The control flow is kinda nasty right now.

rgbkrk:
Trying out the branch now, a couple things missing once this "command mode" is enabled - I'd like to be able to hit `enter` to get into edit mode in the cell as well as create new cells when at the bottom (like the current behavior).

On further evaluation, I notice that it does go back into edit mode style navigation when hopping over markdown cells.

rgbkrk:
To clarify, what was brought up in slack was that page down didn't work in the document when the codemirror editor was focused right?

peggyrayzis:
> Trying out the branch now, a couple things missing once this "command mode" is enabled - I'd like to be able to hit enter to get into edit mode in the cell as well as create new cells when at the bottom (like the current behavior).

Good call. Just pushed some changes to fix both of these issues 😀 

Rendered MD cells are definitely funky in command mode right now because [they handle their own key bindings](https://github.com/nteract/nteract/blob/master/src/notebook/components/cell/markdown-cell.js#L110-L124) at the cell level that conflict with the notebook level. This is going to take some work (possibly a new Redux action to toggle Command/Edit mode), so I wanted to get your opinion on the feature first before I started.

peggyrayzis:
Here's the original message in Slack:
franz

I interpreted the issue with page down to mean that he wanted a way to scroll through the notebook without passing through the cells' editors. @jkornblum Was there anything else you noticed with page down?

peggyrayzis:
We can revisit this later if necessary 👍

PR #1259, opened 2016-12-02
fix(notebook): Removed auto scroll to maintain viewport when executing cells

### LINKED ISSUE #803: scroll behaviour when you execute a cell
From what I understand, it seems like the notebook will auto-scroll to the next input cell on `shift-enter`.  

There are a couple common scenarios where I find this behaviour pretty frustrating:
- When the output takes more than 200ms to render. Jupyter will scroll to the next input cell, and then the output will render and the view will shift further downwards. This pushes the viewport in between cells which is pretty disorienting. 
- When I actually want to view the output of my cell. I press shift-enter and then I have to actively scroll upwards to see my output.
- When I'm happy with my viewport and I'm re-running the same cell over and over with different inputs. Running a cell isn't "idempotent" from a scroll point of view - I can't run a cell twice and preserve my viewport.

Here is an example:

scroll-behaviour

What if running a cell didn't adjust the scroll position?

peggyrayzis:
Resolves #803 

After playing around in the Jupyter notebook, it seems like ripping out our current scroll behavior provides a more consistent user experience. One thing I would like to improve is the transition from scrolling at the top/bottom of the cell output to the notebook scroll - right now, it seems a little janky. Maybe throttling the notebook scroll slightly so it's closer to the speed of cell scroll would help? What do you think?

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -72,36 +72,6 @@ export function getLanguageMode(notebook: any): string {
     'text')));
   return language;
 }
-/**
- * Provide the appropriate position to scroll to when given cell position and size
- * information.
- * @param  {HTMLElement} el - Element to be compared against window and body for
- * scrollTop value.
- * @return {Integer} - An integer for the new document scrollTop value.
- */
-export function scrollToElement(el: HTMLElement): number {
-  const viewportHeight = window.innerHeight;
-  const viewportOffset = document.body.scrollTop;
-
-  const cellTop = el.offsetTop;
-  const cellHeight = el.offsetHeight;
-
-  const belowFold = (cellTop + cellHeight) > (viewportOffset + viewportHeight);
-  const aboveFold = cellTop < viewportOffset;
-
-  if (aboveFold) {
-    return cellTop;
-  }
-
-  if (belowFold) {
-    if (cellHeight > viewportHeight) {
-      return cellTop;
-    }
-    const offset = viewportHeight - cellHeight;
-    return cellTop - offset;
-  }
-  return document.body.scrollTop;
-}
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
😢 I like _some_ of the scrolling, I wish there was a way to capture scrolling down the document on shift-enter in a way that is less obtrusive. People also use shift-enter as a way to go through a notebook and read their way through it.

rgbkrk:
> Maybe throttling the notebook scroll slightly so it's closer to the speed of cell scroll would help? 

That's an interesting idea!

peggyrayzis:
> 😢 I like some of the scrolling, I wish there was a way to capture scrolling down the document on shift-enter in a way that is less obtrusive. People also use shift-enter as a way to go through a notebook and read their way through it.

You can still use shift-enter to read through the notebook, it's just that the default scroll behavior takes over instead of auto scrolling every cell to the top. If you check out my branch, what don't you like about the default behavior?

rgbkrk:
Let me give it a try. I spoke too soon, it was general remorse about even what we have now being awkward. You're making terrific progress on iterating this part of our UX.

peggyrayzis:
Thanks! Yeah, I still think the default needs some work too. 😞 The transition is still a little awkward & could use an animation w/ some easing to make it appear more fluid.

rgbkrk:
Alright, I've tried the branch out side by side and I definitely agree that the default scroll behavior of the browser provides a more consistent user experience.

rgbkrk:
> One thing I would like to improve is the transition from scrolling at the top/bottom of the cell output to the notebook scroll 

Would love to see that.

peggyrayzis:
This is ready to go, I'll work on throttling soon :shipit:

PR #1260, opened 2016-12-02
style: singular "Open Example Notebook"

rgbkrk:
Every time I describe "File -> Open Example Notebooks -> Intro" I want to say "File -> Open Example Notebook -> Intro". I've switched it over to singular here.


jdetle:
👍

PR #1262, opened 2016-12-03
chore(publish): Cleanup github code

jdetle:
Posting this in an attempt to solicit help. Would like to be able to store the response of `createGistCallback` in `publishNotebookObservable`, but I have no idea to get the id and html_url from the response of `createGistCallback`. Please help me out if you have the time, thanks.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -41,8 +41,7 @@ export function notifyUser(filename, gistID, notificationSystem) {
   });
 }
 // give these module scope to allow overwriting of metadata
-let gistID;
-let gistURL;
+
 /**
  * Callback function to be used in publishNotebookObservable such that the
  * response from the github API can be used for user notification.
@@ -61,9 +60,9 @@ export function createGistCallback(observer, filename, notificationSystem) {
       observer.complete();
       return;
     }
-    gistID = response.id;
-    gistURL = response.html_url;
-
+    const gistID = response.id;
+    const gistURL = response.html_url;
+    observer.next(overwriteMetadata('gist_id', gistID));
     notifyUser(filename, gistID, notificationSystem);
   };
 }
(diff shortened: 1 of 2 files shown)


jdetle:
Gave gistID and gistURL module level scope to fix the problem, everything working now and its objectively cleaner code.

src/notebook/epics/github-publish.js
```diff
@@ -39,7 +40,9 @@ export function notifyUser(filename, gistID, notificationSystem) {
     },
   });
 }
-
+// give these module scope to allow overwriting of metadata
+let gistID;
+let gistURL;
```
rgbkrk: This needs to be in the state tree somewhere, we can't have module-level singletons. Reasons:

* For codebase style, this goes against our flux principles
* For testing, you can't assume that the module is in the same state. Every single function in here now has access to these variables as part of the module closure.
* For modularity reasons -- you can't use this module now to publish multiple gists with one single page app; every single notebook document would have the same gist ID and gist URL

As for what to do, rely on the state tree to set these, even if they're transient data (not in the notebook tree). What was the problem with the metadata field inside the notebook? It's also the one used with @minrk's gist extension.

rgbkrk (changes requested):

src/notebook/epics/github-publish.js
```diff
@@ -60,13 +62,8 @@ export function createGistCallback(firstTimePublish, observer, filename, notific
     }
     const gistID = response.id;
     const gistURL = response.html_url;
-
+    observer.next(overwriteMetadata('gist_id', gistID));
```
rgbkrk: Ok, this is better.

I'd walk through the user expectations with this, listing out different scenarios and the events you expect to see.

rgbkrk:
Cases for an existing gisted notebook:

## Authenticated gist

If the user "owns" the gist, they should be able to publish (and also update the metadata if necessary).

If the user doesn't own the gist then they can't publish, yet they could make a new gist (new metadata).

## Unauthenticated gist

Can't ever write back to an unauthenticated gist, have to make a new one (I think...)

src/notebook/epics/github-publish.js
```diff
@@ -39,6 +40,7 @@ export function notifyUser(filename, gistID, notificationSystem) {
     },
   });
 }
+// give these module scope to allow overwriting of metadata
```
rgbkrk: Should probably delete this comment.

rgbkrk (approved):

jdetle:
Unfortunately this is broken for overwriting gists. Will attempt to fix this after work tonight.

rgbkrk:
Oh whoop, I thought I merged this. Glad you're revisiting this now. When you get a chance, I'd try describing in Markdown the user and event flows, resulting states, and expectations.

jdetle:
Need to come to a decision regarding how we want to handle metadata from previous notebooks, but for now I think this is mergeable.

PR #1263, opened 2016-12-04
Update Rx links

rgbkrk:
Noticed we had a link to an RxJS 4 guide, updated it to link to RxJS 5 and a generic reactive tutorial.


lgeiger (approved):
Nice!

PR #1268, opened 2016-12-07
Use zeromq prebuilds for electron

lgeiger:
Burn those dev dependencies 🔥


rgbkrk:
So good.

rgbkrk:
This is the day I've been waiting for.

PR #1269, opened 2016-12-08
chore(install): Remove dependency check

lgeiger:
Follow up to #1268


scripts/check-deps.js
```diff
@@ -1,43 +0,0 @@
...
-  depedencies.forEach(function(dep) {
-    commandExists(dep, function(err, commandExists) {
-      if(!commandExists) {
-        console.error(`Missing dependency: ${dep.bold} Could not be found on your system.`.red);
-      }
-    });
-  });
-
-  if (process.platform === 'linux') {
-    var publishDeps = ['icnsutils', 'graphicsmagick', 'xz-utils'];
```
rgbkrk, line 34: We still need these though right?

lgeiger: Yes. But I noticed that `icnsutils`, `graphicsmagick` and `xz-utils` aren't even detected by `commandExists` hence this check always failed.

lgeiger: And they are only necessary for packaging a release.

rgbkrk: Sounds good, let's go with the trim down.

PR #1271, opened 2016-12-09
[WIP] Export as PDF

jdetle:
[print.pdf](https://github.com/nteract/nteract/files/641511/print.pdf)
This works!!! And it looks great. 

TODO:
------
- [x] Move logic to saving epic. (Wont)
- [x] Determine if I can do a Print Preview (Cant)

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -128,6 +128,27 @@ export function triggerSaveAs(store) {
       triggerKernelRefresh(store);
     });
 }
+export function triggerSaveAsPromise(store) {
+  showSaveAsDialog()
+    .then(filename => {
+      triggerWindowRefresh(store, filename);
+      return new Promise((resolve) => {
+        resolve(filename);
+      });
+    })
+    .then(filename => {
+      triggerKernelRefresh(store);
+      return new Promise((resolve) => {
+        resolve(filename);
+      })
+    })
+    .then(filename => {
+      const state = store.getState();
+      const notificationSystem = state.app.get('notificationSystem');
+      console.log('lololol')
+      setTimeout(exportPDF, 500, filename, notificationSystem);
+    });
+}
 
 export function dispatchSave(store) {
   const state = store.getState();
@@ -308,15 +329,36 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
   store.dispatch(newNotebook(kernelSpecName, cwdKernelFallback()));
 }
 
-export function exportPDF(store) {
-  const state = store.getState();
-  const filename = state.metadata.has('filename') ?
-    state.metadata.get('filename').split('.')[0] : '~/Untitled';
+export function storeToPDF(store) {
+  let state = store.getState();
+  let filename = path.basename(state.metadata.get('filename'), '.ipynb');
   const notificationSystem = state.app.get('notificationSystem');
+  if(filename == '') {
+    notificationSystem.addNotification({
+      title: 'File has not been saved!',
+      message: 'Click the button below to save, or it will overwrite `Untitled.pdf.`',
+      dismissible: true,
+      position: 'tr',
+      level: 'warning',
+      action: {
+        label: 'Save As',
+        callback: function cb() {
+          triggerSaveAsPromise(store);
+        },
+      },
+    });
+  }
+  else {
+    exportPDF(filename, notificationSystem);
+  }
+}
+
+
+
+export function exportPDF(filename, notificationSystem) {
   remote.getCurrentWindow().webContents.printToPDF({ printBackground: true }, (error, data) => {
     if (error) throw error;
     fs.writeFile(`${filename}.pdf`, data, (error_fs) => {
-      if (error_fs) throw error_fs;
       notificationSystem.addNotification({
         title: 'PDF exported',
         message: `File ${filename} has been exported to a pdf.`,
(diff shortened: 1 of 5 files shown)


rgbkrk:
It's so weird to me that [there is no print role](http://electron.atom.io/docs/api/menu-item/#class-menuitem) for menu items.

I'm guessing there's a print preview when using `contents.print`, perhaps we can provide both options in a reasonably simple way. Does this have to go to an epic or can this menu option only handle the event and take action, since it doesn't change state?

rgbkrk:
Also... Woohoo! Thank you for tackling this.

jdetle:
Yeah I guess you're right that it should just be a function in main since it really doesn't do anything with state. Let me play around with that tonight and get back to you. Honestly I think print preview is something to work towards and mess around with but maybe its own PR.

rgbkrk:
Seems like we still need the state to get the filename, it just doesn't have to dispatch an action.

src/notebook/menu.js
```diff
@@ -307,6 +308,31 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
   store.dispatch(newNotebook(kernelSpecName, cwdKernelFallback()));
 }
 
+export function exportPDF(store) {
+  const state = store.getState();
+  const filename = state.metadata.has('filename') ?
+    state.metadata.get('filename').split('.')[0] : '~/Untitled';
```
rgbkrk: I think we should show a notification if they haven't set the filename by saving. I'd rather they be able to choose where to save it than making a default.

Other related feedback for this filename extraction:

* `something.r.ipynb` would become `something.pdf` - use [`path.basename`](https://nodejs.org/api/path.html#path_path_basename_path_ext) to extract based on `.ipynb` at the end:

```js
> path.basename('something.r.ipynb', '.ipynb') + '.pdf'
'some.file.pdf'
```

* `~` as part of the filename is not a good default on Windows and may not always resolve correctly. If a user misses the notification, we're also overwriting the same file over and over

src/notebook/menu.js
```diff
@@ -307,6 +308,31 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
...
+        title: 'PDF exported',
+        message: `File ${filename} has been exported to a pdf.`,
+        dismissible: true,
+        position: 'tr',
+        level: 'success',
+        action: {
+          label: 'Open PDF',
+          callback: function openPDF() {
+            shell.openItem(`${filename}.pdf`);
+          },
```
rgbkrk, line 372: Love this extra touch with the notifications to let them open the PDF.

src/notebook/menu.js
```diff
@@ -307,6 +308,31 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
   store.dispatch(newNotebook(kernelSpecName, cwdKernelFallback()));
 }
 
+export function exportPDF(store) {
```
rgbkrk: I would probably write two functions, one that takes the store and one that takes only the filename.

This should be able to stand on its own as a function that prints-to-PDF the current file. The store adapter around it encodes our business/state logic rather than it being part of the `exportPDF` function.

src/notebook/menu.js
```diff
@@ -307,6 +308,31 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
...
+  const state = store.getState();
+  const filename = state.metadata.has('filename') ?
+    state.metadata.get('filename').split('.')[0] : '~/Untitled';
+  const notificationSystem = state.app.get('notificationSystem');
+  remote.getCurrentWindow().webContents.printToPDF({ printBackground: true }, (error, data) => {
+    if (error) throw error;
+    fs.writeFile(`${filename}.pdf`, data, (error_fs) => {
+      if (error_fs) throw error_fs;
```
rgbkrk: Since this is used by an event handler, this error will either not be seen or thrown at a global level. It would be better to show a notification.

rgbkrk (changes requested):

src/notebook/menu.js
```diff
@@ -307,6 +308,31 @@ export function dispatchNewNotebook(store, event, kernelSpecName) {
...
 }
 
+export function exportPDF(store) {
+  const state = store.getState();
+  const filename = state.metadata.has('filename') ?
+    state.metadata.get('filename').split('.')[0] : '~/Untitled';
+  const notificationSystem = state.app.get('notificationSystem');
+  remote.getCurrentWindow().webContents.printToPDF({ printBackground: true }, (error, data) => {
+    if (error) throw error;
+    fs.writeFile(`${filename}.pdf`, data, (error_fs) => {
```
rgbkrk, line 361: One thing you may want to do is to have this function accept a callback rather than do the notifications in here (to make it easier to test).


```js
function exportToPDF(filename, cb) {
  ...
  remote.getCurrentWindow().webContents.printToPDF({ printBackground: true }, (error, data) => {
    if (error) {
      cb(error);
    }    
    fs.writeFile(`${filename}.pdf`, data, cb);
  ...
}
```

jdetle:
Thanks for the very helpful feedback!

rgbkrk:
This needs a rebase and I think there are remaining TODOs you were looking to. Are you still coming back to this one?

jdetle:
Yes, I've been putting it off because I'm not sure if I can do a good job on it. I'll take a stab at it tonight and tomorrow and if I cant get it up to par I'll close it and come back to it later.

rgbkrk:
No problem, I'm just going through open PRs and some issues since we've been doing a fair bit of refactoring and I'm trying to carve out the module exporting for use in a web app version of the notebook.

Issue #670, opened 2016-09-02
 Registering custom mimetypes


rgbkrk:
This issue is twofold: determine a strategy for registering mimetypes and, at a higher level, register plugins in general.

Similar to how [HyperTerm](https://hyperterm.org/) expects particular keys on an `exports` object for a plugin, I'd like to do the same with custom `Transforms` (a la [`transformime-react`](https://github.com/nteract/transformime-react)). We have one custom transform right now and I'm expecting us to support some others right out of the box. Here's what I'm expecting for a common interface:

``` js
import React from 'react';

const MIMETYPE = 'application/myMimetype+json';

export class CustomTransform extends React.Component {
  shouldComponentUpdate() {
    return false;
  }

  render() {
    const data = this.props.data.toJS();
    return (
      <CustomThing data={data} />
    );
  }
}

CustomTransform.propTypes = {
  data: React.PropTypes.any,
};

CustomTransform.MIMETYPE = MIMETYPE;

exports.transform = CustomTransform;
```

The three main points:
- A CustomTransform is a `React.Component` that expects a payload in the `data` field of the props (which is Immutable.js).
- The mimetype expected to come through on the Jupyter messaging protocol (and in the notebook format) is a static `MIMETYPE` attribute on the transform.
- `exports` has a key called `transform` with the transformer.

Registering it is then a reduction on the store:

``` js
[REGISTER_NEW_MIMETYPE]: (state, action) => {
  const { transform } = action;
  const mimetype = transform.MIMETYPE;

  return state.update('displayOrder', displayOrder => displayOrder.splice(0, 0, mimetype)
       .update('transforms', transforms => transforms.set(mimetype, transform);
}
```

rgbkrk:
I've been thinking lately about passing comms in as a Subject for a display/transform:

``` jsx
<InteractiveMap
  data={data} // raw data from a mimebundle that targets the mimetype of this display
  comms={comms} // Two way binding for pure comms messages (with Observables)
/>
```

It would be a two-way binding to send and receive comm messages.

Perhaps the surface should be wider for what you can listen to/operate on. Perhaps it's context, not props. ¯\_(ツ)_/¯ Researching it over the coming months.

rgbkrk:
Related exploration: https://github.com/rgbkrk/synchronized-object-diff-spec

rgbkrk:
Also related: #840

rgbkrk:
[...]

Issue #481, opened 2016-05-19
[META] Extensions


jdfreder:
I'd like to start the discussion of extensions.  I think there are some cool things we can do with the npm API to find locally or globally installed extensions and load them dynamically.   Both would provide a stellar user experience.  I think and we could build an in app UI on top of it too.  

I haven't thought much about React extension points, but I'm sure that's a well worn trail.  Seems like as an MVP, extension writers would need to be able to A) provide custom reductions B) provide components or override components for given extension points in the nteract React DOM.

rgbkrk:
My favorite example of this is [hyperterm](https://hyperterm.org/), particularly in defining a structure for registering and relying on Redux as the plugin mechanism (actions, middlewares, etc.). I think we could build something pretty cool here.

rgbkrk:
Oh, and the quintessential [nylas - building plugins for react apps](https://nylas.com/blog/react-plugins) post!

rgbkrk:
Now that we've been at this for a bit, I think there are three main points of extensions we should aim for supporting:
- Themes
- Custom renderers #670
- Publication

rgbkrk:
We currently have a simple configuration file so in lieu of letting us expand our scope too much, I'll close this and we can open it again at a later date.

PR #1280, opened 2016-12-13
[WIP] feat: prototype a table component + mimetype for tabular data

rgbkrk:
A transform for [JSON Table Schema](http://specs.frictionlessdata.io/json-table-schema/) + data (record oriented). The nteract side of https://github.com/pandas-dev/pandas/issues/14386. 

Takes an overall structure like:

```json
{
  "schema": {
    "fields": [
      {
        "type": "any",
        "name": "Date"
      },
      {
        "type": "number",
        "name": "Open"
      },
      {
        "type": "number",
        "name": "High"
      },
      {
        "type": "number",
        "name": "Low"
      },
      {
        "type": "number",
        "name": "Close"
      },
      {
        "type": "integer",
        "name": "Volume"
      },
      {
        "type": "number",
        "name": "Adj Close"
      }
    ]
  },
  "data": [
    {
      "Date": "2000-03-01",
      "Adj Close": 33.68,
      "Open": 89.62,
      "Low": 88.94,
      "Volume": 106889800,
      "High": 94.09,
      "Close": 90.81
    },
    {
      "Date": "2000-03-02",
      "Adj Close": 34.63,
      "Open": 91.81,
      "Low": 91.12,
      "Volume": 106932600,
      "High": 95.37,
      "Close": 93.37
    }
  ]
}
```

and creates a table.

screen shot 2016-12-12 at 5 29 22 pm

Definitely a prototype, only putting this up to promote discussion. We have enough information here to align columns according to types, format tables accordingly, etc.

We should be able to map from `schema, data` to a reasonable table after iterating on this.

### DIFF
--- a/example-notebooks/only-table.ipynb
+++ b/example-notebooks/only-table.ipynb
@@ -1,0 +1,352 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {
+    "collapsed": false,
+    "outputExpanded": true
+   },
+   "source": [
+    "display({\n",
+    "    'application/vnd.tableschema.v1+json': payload\n",
+    "}, raw=True)"
+   ],
+   "outputs": [
+    {
+     "metadata": {},
+     "data": {
+      "application/vnd.tableschema.v1+json": {
+       "schema": {
+        "fields": [
+         {
+          "type": "any",
+          "name": "Date"
+         },
+         {
+          "type": "number",
+          "name": "Open"
+         },
+         {
+          "type": "number",
+          "name": "High"
+         },
+         {
+          "type": "number",
+          "name": "Low"
+         },
+         {
+          "type": "number",
+          "name": "Close"
+         },
+         {
+          "type": "integer",
+          "name": "Volume"
+         },
+         {
+          "type": "number",
+          "name": "Adj Close"
+         }
+        ]
+       },
+       "data": [
+        {
+         "Low": 88.94,
+         "High": 94.09,
+         "Close": 90.81,
+         "Adj Close": 33.68,
+         "Volume": 106889800,
+         "Open": 89.62,
+         "Date": "2000-03-01"
+        },
+        {
+         "Low": 91.12,
+         "High": 95.37,
+         "Close": 93.37,
+         "Adj Close": 34.63,
+         "Volume": 106932600,
+         "Open": 91.81,
+         "Date": "2000-03-02"
+        },
+        {
+         "Low": 93.87,
+         "High": 98.87,
+         "Close": 96.12,
+         "Adj Close": 35.65,
+         "Volume": 101435200,
+         "Open": 94.75,
+         "Date": "2000-03-03"
+        },
+        {
+         "Low": 90.12,
+         "High": 97.37,
+         "Close": 90.62,
+         "Adj Close": 33.61,
+         "Volume": 93609400,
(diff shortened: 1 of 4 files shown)


gnestor:
@rgbkrk I started work on this for jupyterlab and notebook: https://github.com/gnestor/jupyterlab_table/blob/labextension-build/component/index.js

I haven't fully adopted the JSON Table schema yet, I'm just rendering `data_frame.to_dict()` data.

Great to see that you're working the pandas side! I spoke to @ellisonbg about this earlier today!!

rgbkrk:
Went ahead and created the table directly, mirrored the current structure style of the pandas HTML we're used to seeing. Check the source out 👀. Result:

screen shot 2016-12-13 at 2 42 19 pm

I haven't used the typing information yet, this only sets us up for later allowance of theming and utility across other languages - we've broken up the presentational components and the data.

## What's missing?

The indexes on the data frame. They're not captured with this current `data, schema` approach.

gnestor:
@rgbkrk Nice! A couple things:

* Could you create a todo list in this issue so I can track what you've done and what you're working on? I'd like to help but I don't want to overlap.
* Why not use the `FixedDataTable` component?
* I updated the mimerender extensions today to provide their own display method, which would allow something like: 

```py
from jupyterlab_table import Table
Table(df)
```

See jupyterlab_plotly for a reference implementation: https://github.com/gnestor/jupyterlab_plotly#usage

I'd appreciate your feedback regarding this approach and I'd like to explore how our mimerender efforts can be better combined. For example, is it possible to register new transforms/mimetypes that are not bundled with nteract? If so, it would be nice if these mimerender extensions could also support nteract since they already support notebook and lab. I see some discussion at https://github.com/nteract/nteract/issues/670 and https://github.com/nteract/nteract/issues/481.

rgbkrk:
> Could you create a todo list in this issue so I can track what you've done and what you're working on? I'd like to help but I don't want to overlap.

Sure. I'm not working on this in any serious sense, it's only up as a prototype for the pandas discussion (this doesn't have to be the end component, just enough to rely on the JSON spec and messaging).

> Why not use the FixedDataTable component?

My first commit and comment used the fixed data table component. It required a lot more configuration to get right as opposed to creating the table directly. Feel free to iterate on a component that uses this `{ schema, data }` spec to match pandas while using the FixedDataTable. I don't have the time to fuss too much with it.

> I updated the mimerender extensions today to provide their own display method, which would allow something like...

Cool! I'm glad you're making Python side versions for exploring. I tend to defer to either providing these methods directly in `IPython.display.*` utilities or in the core libraries. In the case of Plotly, it would be cool if you picked up https://github.com/plotly/plotly.py/pull/562 from where @chriddyp left off. I'd rather the support be built into the libraries being supported and used by users.

>  is it possible to register new transforms/mimetypes that are not bundled with nteract?

Not currently, continue the discussion in #670. An externally facing API is not something I want to support until we finish up higher priority issues. For now, bundling many renderers is the approach I've been going with. The API in this case is the mimetype and the expected payload. We rely on specs, not the implementation.

rgbkrk:
Technicality:

`application/vnd.dataresource+json` is what we want and the structure is:

```js
{
  "resources": [
    {
      "schema": {...jsonTableSchemaHere},
      "data": [...rows],
    }
  ]
}
```

rgbkrk:
Closing for now to reimplement later instead of rebase.

PR #1284, opened 2016-12-20
chore(publish): Enable multi platform build

lgeiger:
We can now build all nteract releases on one platform if the [required dependencies](https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build) for a multi platform build are installed!

I didn't test windows but the linux release built on mac works great.


rgbkrk (approved):
Woohoo!

PR #1285, opened 2016-12-20
Stricter linting

lgeiger:


package.json
```diff
@@ -106,6 +106,7 @@
     "jupyter-paths": "^1.0.2",
     "kernelspecs": "^1.0.3",
     "leaflet": "^1.0.1",
+    "lodash.has": "^4.5.2",
```
lgeiger: We use `lodash` but didn't include it as a dep.

src/notebook/components/transforms/image.js
```diff
@@ -1,8 +1,6 @@
 /* @flow */
 import React from 'react';
 
-import _ from 'lodash';
```
lgeiger, line 4: 🔥

src/notebook/components/transforms/vega.js
```diff
@@ -27,7 +24,7 @@ function embed(el: HTMLElement, spec: Object, mode: string, cb: (err: any, resul
   };
 
   if (mode === 'vega-lite') {
-    embedSpec.spec.config = _.merge({
+    embedSpec.spec.config = Object.assign({
```
lgeiger: Unnecessary use of `lodash`.

src/notebook/reducers/document.js
```diff
@@ -76,9 +76,9 @@ export default handleActions({
     const output = action.output;
     const cellID = action.id;
 
-    if (output.output_type !== 'display_data' || !(_.has(output, 'transient.display_id'))) {
+    if (output.output_type !== 'display_data' || !(_has(output, 'transient.display_id'))) {
```
lgeiger: I don't know if there's a easy ES6 method for this.

src/notebook/components/transforms/geojson.js
```diff
@@ -63,7 +63,7 @@ export class GeoJSONTransform extends React.Component {
       this.map.removeLayer(this.tileLayer);
       this.tileLayer = L.tileLayer('https://api.tiles.mapbox.com/v4/{id}/{z}/{x}/{y}.png?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpandmbXliNDBjZWd2M2x6bDk3c2ZtOTkifQ._QA7i5Mpkd_m30IGElHziw', {
         attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
-        id: `mapbox.${this.props.theme}`,
+        id: `mapbox.${theme}`,
```
lgeiger, line 66: This may even fix a bug, since:
```javascript
const theme = this.props.theme === 'light' || this.props.theme === 'dark' ? this.props.theme : 'light';
```

lgeiger (commented):
I added a few comments to the important changes.

.eslintrc
```diff
@@ -44,7 +42,6 @@
         ]
       }
     ],
-    "import/no-extraneous-dependencies": "off"
   },
```
rgbkrk, line 44: Woohoo! Thanks for cleaning this up.

package.json
rgbkrk: Ah, we must have had a sneaky dep that had lodash as a dep. I'm ok with us including `lodash` in general, especially since we can use [`lodash-webpack-plugin`](https://www.npmjs.com/package/lodash-webpack-plugin).

src/main/menu.js
```diff
@@ -90,120 +90,48 @@ export const fileSubMenus = {
...
         click: () => {
-          const opts = {
-            title: 'Open a notebook',
-            filters: [
-              { name: 'Notebooks', extensions: ['ipynb'] },
-            ],
-            properties: [
-              'openFile',
-            ],
-          };
```
rgbkrk, line 101: Great to delete these since they're largely unnecessary copy paste.

If we wanted to trim this down more and not create new arrow functions these could also be:

```js
click: launch.bind(null, path.join(exampleNotebooksDirectory, 'intro.ipynb')),
```

That or use a [partial](https://lodash.com/docs/4.17.2#partial).

src/notebook/components/cell/draggable-cell.js
```diff
@@ -3,7 +3,6 @@
 import React from 'react';
 import { shouldComponentUpdate } from 'react-addons-pure-render-mixin';
 import { DragSource, DropTarget } from 'react-dnd';
-import { findDOMNode } from 'react-dom';
```
rgbkrk, line 6: Yay for getting rid of `findDOMNode`!

src/notebook/components/transforms/geojson.js
rgbkrk: Looks like it!

src/notebook/components/transforms/vega.js
rgbkrk: `_.merge` and `Object.assign` are not the same - `_.merge` does a deep merge of keys which we need here with this spec merge.

src/notebook/menu.js
```diff
@@ -57,7 +46,7 @@ export function dispatchSaveAs(store, evt, filename) {
 
 const dialog = remote.dialog;
 
-export function showSaveAsDialog(defaultPath) {
```
rgbkrk, line 60: Ah good a cleanup from an old refactor.

src/notebook/reducers/document.js
rgbkrk: Not really. This did make it a lot easier to work with. I'm a-ok with lodash because of how it's maintained and how I don't wish to maintain utility functions within our package (unless domain specific).

rgbkrk (changes requested):
This is so excellent! Thank you for the major cleanup.

There's only one part you'll want to change back which is the use of `_.merge` vs. `Object.assign`.

rgbkrk:
You'll need a rebase since I merged the other PR. 😄

rgbkrk:
It's annoying that comments in a review that are responses aren't grouped with the original message. :frowning_man:

lgeiger:
Alright rebased and ready to go

lgeiger:
@rgbkrk Thanks for the review 👍

src/notebook/components/transforms/vega.js
```diff
@@ -1,11 +1,9 @@
 /* @flow */
 import React from 'react';
-import ReactDOM from 'react-dom';
+import _ from 'lodash';
```
rgbkrk, line 3: Oooh much better form, great!

rgbkrk (approved):

PR #1286, opened 2016-12-20
Issue: chore(lint): undefined-var

lgeiger:
I don't know how to fix this [lint error](https://github.com/nteract/nteract/blob/master/src/notebook/epics/config.js#L70) because `store` is undefined.


rgbkrk:
change

```
export const saveConfigEpic = actions =>
```

to

```
export const saveConfigEpic = (actions, store) =>
```

Issue #1288, opened 2016-12-20
Refactor components to remove Redux logic


peggyrayzis:
In order to keep our components modular and reusable, @rgbkrk & I discussed removing all Redux logic (dispatching, accessing the state tree) from the components themselves and moving them into providers.

Providers are HOCs that act as middlemen between presentational components & the Redux store (or any other form of data hydration). They provide presentational components with access to application level data by passing down callbacks through props.

Currently, these components are connected to the store via context. To disconnect our components, we need to:
- Remove all references to the store through context
- Abstract dispatching actions into callbacks
- Move mapping state to props into providers

These components need to be refactored:
- [ ] Notebook (saving for last)
- [x] Cell Creator Buttons
- [x] Draggable Cell (@jdetle)
- [ ] Cell (@peggyrayzis)
- [x] Toolbar (@captainsafia) - #1323 
- [x] Editor (@peggyrayzis)

When finished, this will move us closer to our goal of completely reusable components. 🎉  I will be making a PR for Cell Creator Buttons soon so you can see an example of this in action.

jdetle:
I'll try to help out once I see your PR as a guide 👍

rgbkrk:
We're getting there!

rgbkrk:
Another one checked off!

rgbkrk:
I think we're on a pretty good level with this now with the packages in `@nteract/core` and it's getting better each day. One note for historical record -- at least at the time of writing this, we've kept providers and made the "views" just be called "components". 😄

Issue #549, opened 2016-07-18
Monorepo


rgbkrk:
In order to approach having more components available for an online version of the nteract components (in your friendly browser instead of Electron) we need to create a [monorepo](https://github.com/babel/babel/blob/master/doc/design/monorepo.md).

I'm in favor of exporting components _directly_ from nteract/nteract to assist with this so that's it's easy to work on the desktop app as a standalone in tandem with these core components. Especially since the API keeps evolving for actions, reducers, etc.

Motivation for "new wave modularity": http://www.macwright.org/2016/07/08/lerna-npm-organizations-new-wave-modularity.html

Sub-issues that make this easier for us:

* [ ] https://github.com/nteract/nteract/issues/1288

captainsafia:
I'm down for monorepo.

![](http://i.giphy.com/5ev3alRsskWA0.gif)

ivanov:
![monorail!](https://img.buzzfeed.com/buzzfeed-static/static/2014-08/4/8/enhanced/webdr04/anigif_enhanced-29437-1407154860-14.gif)

rgbkrk:
Another component I think we'll want to bring into here is `enchannel-zmq-backend` as well as `spawnteract` (I'm guessing we'd make it based off of `spawn-rx` in here). `jupyter-paths` can likely stay separate though.

I've updated the issue description with this great article too: http://www.macwright.org/2016/07/08/lerna-npm-organizations-new-wave-modularity.html

Soon I'll provide a better outline of what we can do collectively. We'll likely want to turn our redux / context focused components into components that accept callbacks (`onExecute`, etc.)

rgbkrk:
Here's how mapbox-gl-js organized it: https://github.com/mapbox/mapbox-gl-js/issues/3656

rgbkrk:
Wish I had noticed this before:

> You can use `lerna import <package>` to transfer an existing package into your Lerna repository; this command will preserve the commit history. `lerna import <package>` takes a local path rather than a URL. In this case you will need to have the repo you wish to link to on your file system.

Could have used that rather than the way we did it in the rush for time.

rgbkrk:
Closing this since we're now in a good setup with lerna.

PR #1290, opened 2016-12-21
refactor(provider): Removed Redux logic from cell creator buttons

peggyrayzis:
When finished, this will resolve the Cell Creator Buttons piece of #1288.

- Removed the Cell Creator Buttons component entirely. We didn't need it once we ripped the Redux logic out of the component
- Moved all Redux logic into the Cell Creator provider
- Removed the PureRenderMixin from Notebook & Cell Creator
- Added support for object rest/spread since it's stage 3 🎉 
- Throttled the function handling the visibility of the Cell Creator bar for better perf 🚀 (we could prob throttle this even more, lmk what you think)

I still need to fix the tests & resolve the merge conflict - will take care of that in the morning. I'm excited to hear everyone's feedback!


.babelrc
```diff
@@ -5,6 +5,7 @@
     ],
     "plugins": [
```
rgbkrk, line 6: Based on the [the usage docs](https://github.com/jhen0409/babel-preset-es2015-node6#usage) for `es2015-node6` preset, do we need to use `es2015-node6/object-rest` as the preset?

src/notebook/components/notebook.js
```diff
@@ -110,7 +107,6 @@ export class Notebook extends React.Component {
 
   constructor(): void {
     super();
-    this.shouldComponentUpdate = shouldComponentUpdate.bind(this);
```
rgbkrk: :smile:

src/notebook/views/cell-creator.js
```diff
@@ -0,0 +1,79 @@
...
+  hoverElement: HTMLElement;
+
+  constructor(): void {
+    super();
+
+    this.state = {
+      show: false,
+    };
+
+    this.updateVisibility = throttle(this.updateVisibility.bind(this), 200);
```
rgbkrk, line 48: Neat, didn't realize this would work so well with React synthetic events.

rgbkrk (approved):

rgbkrk:
This is looking pretty trim!

src/notebook/views/cell-creator.js
peggyrayzis: We're actually listening to a DOM mouse event here instead of a React synthetic event, which is why it works so well 😉

.babelrc
peggyrayzis: Hmm I'll look into this, thanks for pointing this out!

peggyrayzis:
Finally refactored those tests! 🎉 Everything should be good to go now, except for the DnD test which we will refactor at a later date.

rgbkrk:
I agree we should definitely get this in sooner than later, we can revisit the DnD tests later.

rgbkrk (approved):

rgbkrk:
Thanks for this mega refactor, driving us toward completing #549 and #1288. :smile:

captainsafia:
Thanks for the effort on this, @peggyrayzis! I'll review the changes you made and see if I can mimic it in the toolbar component working towards #1288. Will probably do that tomorrow or Wednesday.

peggyrayzis:
Awesome! Let me know if any questions pop up. 😀

PR #1293, opened 2016-12-22
fix(splash): Hide scrollbars

lgeiger:
Most platforms (all except macOS) don't hide scrollbars per default.


rgbkrk:
Thanks! I was seeing this intermittently, glad you fixed it.

PR #1294, opened 2016-12-22
chore(lodash): Import lodash partials instead of whole library

peggyrayzis:
This will enable us to take advantage of tree shaking when we eventually upgrade to Webpack 2. 🌳 

More info if you're curious: https://blog.engineyard.com/2016/tree-shaking


lgeiger (approved):
Thank you :tada:

rgbkrk:
Nice!

We should start doing the operator patching for Rx soon too.

PR #1296, opened 2016-12-23
chore(flow-typed): Update flow typed libdefs

peggyrayzis:
🎉


rgbkrk:
Based on reviewing the Travis logs and our current setup, we should take `mkdirp` out of our `package.json`'s `devDependencies`, leaving it only in `dependencies`.

rgbkrk:
Thank you!

PR #1302, opened 2016-12-26
chore(deps): bump vega to 3.0.0-beta.10

rgbkrk:
In an attempt to escape the whole `canvas` requirement, I bumped `vega` to the 3.0.0 beta. _However_, [vega-scenegraph](https://github.com/vega/vega-scenegraph) relies on canvas still so it's still a bit of a headache since we're both in node.js and don't actually need canvas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "spawn-rx": "^2.0.3",
     "spawnteract": "^2.2.0",
     "uuid": "^3.0.0",
-    "vega": "^2.6.3",
+    "vega": "^3.0.0-beta.10",
     "vega-embed": "^2.2.0",
     "vega-lite": "^1.3.1",
     "yargs": "^6.4.0"


rgbkrk:
Sure enough, that test fails for me locally too. Running the app was fine with this dep though. Closing this PR.

rgbkrk:
Nevermind, this had the same error after I nuked node modules and installed again.

Posted an issue in `vega-embed`: https://github.com/vega/vega-embed/issues/15

rgbkrk:
@mathisonian - would it be worth us asking if `canvas` can be made a peer dependency for `vega-scenegraph`?

mathisonian:
@rgbkrk yeah, this has been brought up, but hasn't really been a high priority. If you can explain how it would help improve compatibility, people should be open to it. 

I started doing a little work on this using a slightly different solution (see https://github.com/vega/vega-scenegraph/issues/29 and https://github.com/vega/vega-wordcloud/pull/1), it had a bug with the current vega builds because the build toolchain there doesn't use `browser` field in `package.json`. I believe it uses `jsnext:main`, but I haven't revisited this just because it hasn't been high priority.

rgbkrk:
I'm making a bit of a backwards ask, in that I'm hoping that I can have node-canvas not be installed in Electron environments and forcing other node users (that would not have a native canvas) to install it directly when they need it.

mathisonian:
Ahh, got it. That may be a bit of a bigger ask, but I'd see what @arvind or @jheer think

jheer:
We're open to making this easier if we can. For Vega 3, vega-scenegraph includes canvas only under `optionalDependencies`, and uses a `"browser": {"canvas": false}` directive to reduce problems for bundlers. If you can share a bit more about the nature of the errors you're running into we can take a closer look.

rgbkrk:
When canvas fails at building on folks machines, it doesn't always gracefully complete. See https://github.com/nteract/nteract/issues/1456

PR #1305, opened 2016-12-28
fix(editor): Indenting with tab no longer focuses the next cell

### LINKED ISSUE #1301: Tab key is switching between cells
On master (I'm at 187548811c1629cce5b08303f32394b0e57ecf01) the tab key is no longer indenting - it's switching to the next cell.

peggyrayzis:
Fixes #1301 

There's a slight issue with deleting - we need to write a spacesToTab function to convert the spaces back to tabs so you only have to press delete once to get rid of the tab. Right now, you have to press delete for as many spaces make up a tab.

### DIFF
--- a/src/notebook/components/cell/editor/index.js
+++ b/src/notebook/components/cell/editor/index.js
@@ -139,6 +139,7 @@ export default class Editor extends React.PureComponent {
   state: State;
   onChange: (text: string) => void;
   onFocusChange: (focused: boolean) => void;
+  tabToSpaces: (editor: Object) => void;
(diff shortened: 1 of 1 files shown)


rgbkrk:
Thank you for the fix.

> Right now, you have to press delete for as many spaces make up a tab.

Yuck.

I'm wondering if there's a built in way for codemirror to handle what we added on here (more bundled codemirror extensions) so we don't have to add on these functions in general. 
¯\\\_(ツ)\_/¯

rgbkrk (approved):

PR #1309, opened 2017-01-03
test: mock fs

rgbkrk:
I noticed when running the tests that 'filename' (literally) was being created
at the root while running tests. This forces our use of `fs` to be mocked within
tests.


lgeiger (approved):
Thanks 👍

PR #1313, opened 2017-01-08
Models 

rgbkrk:
At the end of the day Friday I spent time with @ivanov and @ian-r-rose to explore simple mechanics for modifying a model structure reliant on comms that are implementable in a variety of languages. The base concept for these models over the comms is that we employ the use of a function to go from `oldModel` to `newModel` based on a payload.

```js
model = reducer(model, data)
```

We experimented (on a whiteboard) with a few different reducers as a baseline:

* deepMerge
* replace
* shallowMerge
* setIn

After exploring these, we ended up thinking that `setIn` covered many of the uses of the merge operations while also encapsulating `replace` entirely\*. The payload for updating a model via `setIn` includes the `path` and the `value` (which can be arbitrary JSON).

\* caveat: `setIn` is overkill for single primitive values for a model

## Model change example

Starting model

```js
model = {
  fruits: [
    { color: 'red', name: 'apple' },
    { color: 'green', name: 'banana' }
  ]
}
```

```js
> setIn(model, ['fruits', 1, 'color'], 'yellow')
{
  fruits: [
    { color: 'red', name: 'apple' },
    { color: 'yellow', name: 'banana' }
  ]
}
```

## Displaying a model

This implementation expects that a transform/renderer accepts the `models` as well as a `modelID` to display a representation of a model.

Example component usage:

```js
<ModelDebugTransform
  data={data}
  models={models}
  modelID={modelID}
/>
```

This is similar in usage to the other transforms for html, images, plotly, vega, etc. with the addition of having access to all the models as an immutable Map as well as a particular model ID. The intent with the model ID is for this output to have an associated field in `transient`.

Example payload for `display_data`:

```js
content = {
  data: {
    'application/x-nteract-model-debug+json': 'pretend-data',
  }
  transient: {
    model_id: '1'
  },
}
```

This matches the format of the update display spec with the use of transient for something that is intended to match the lifetime of the kernel.

### DIFF
--- a/src/notebook/components/transforms/index.js
+++ b/src/notebook/components/transforms/index.js
@@ -22,6 +22,8 @@ import {
 import PlotlyTransform from './plotly';
 import GeoJSONTransform from './geojson';
 
+import ModelDebug from './model-debug';
+
 import {
   VegaLite,
   Vega,
@@ -58,20 +60,28 @@ export const standardDisplayOrder: StandardDisplayOrder = new ImmutableList([
   'text/plain',
 ]);
 
+function registerTransform({ transforms, displayOrder }, transform) {
+  return {
+    transforms: transforms.set(transform.MIMETYPE, transform),
+    displayOrder: displayOrder.insert(0, transform.MIMETYPE)
+  };
+}
 
-// Register custom transforms
-const transforms = standardTransforms
-  .set(PlotlyTransform.MIMETYPE, PlotlyTransform)
-  .set(GeoJSONTransform.MIMETYPE, GeoJSONTransform)
-  .set(VegaLite.MIMETYPE, VegaLite)
-  .set(Vega.MIMETYPE, Vega);
+const additionalTransforms = [
+  ModelDebug,
+  PlotlyTransform,
+  GeoJSONTransform,
+  VegaLite,
+  Vega,
+];
 
-// Register our custom transforms as the most rich (front of List)
-const displayOrder = standardDisplayOrder
-  .insert(0, PlotlyTransform.MIMETYPE)
-  .insert(0, VegaLite.MIMETYPE)
-  .insert(0, Vega.MIMETYPE)
-  .insert(0, GeoJSONTransform.MIMETYPE);
+const {
+  transforms,
+  displayOrder,
+} = additionalTransforms.reduce(registerTransform, {
+  transforms: standardTransforms,
+  displayOrder: standardDisplayOrder,
+});
 
 /**
  * Choose the richest mimetype available based on the displayOrder and transforms
--- a/src/notebook/components/transforms/model-debug.js
+++ b/src/notebook/components/transforms/model-debug.js
@@ -1,0 +1,40 @@
+/* @flow */
+import React from 'react';
+
+const Immutable = require('immutable');
+
+type Props = {
+  data: string,
+  models: Immutable.Map<string, any>,
+  modelID: string,
+}
+
+class ModelDebug extends React.Component {
+  props: Props;
+  static MIMETYPE = 'application/x-nteract-model-debug+json';
+
+  shouldComponentUpdate(): boolean {
+    return true;
+  }
+
+  render(): ?React.Element<any> {
+    const { models, data, modelID } = this.props;
+    // TODO: Provide model IDs on transient field
+    // For now, if modelID is not provided (or model does not exist),
+    // show all the models
+    const model = models.get(modelID, models);
+    return (
+      <div>
+        <h1>{JSON.stringify(data, null, 2)}</h1>
+        <pre>
+          {
+            model ? JSON.stringify(model, null, 2) : null
+          }
+        </pre>
+      </div>
+    );
+  }
+}
+
+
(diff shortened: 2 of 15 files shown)


rgbkrk:
As for updates client --> backend, I'm expecting for unidirectional data flow with a bit of a twist in that we'll both update our own version of the model while sending an update to the kernel backend and expecting to see the same update come across the wire as well. This makes for good reasoning why we want `setIn` and not allowing `push` as a mechanic.

rgbkrk:
Concrete example of updating the user's model after working through it with @peggyrayzis:

In the user's output is a slider that controls `r` and the view of an image called `image`. The model ID will be `my-image-model` for this little example.

```js
models = {
  'my-image-model': {
    'image': 'binaryData...',
    'r': 200,
  }
}
```

When user slides slider `r`, we emit a Redux action like:

```js
{
  type: "UPDATE_USER_MODEL",
  modelID: 'my-image-model',
  value: "200",
  path: ["r"]
}
```

Within reducers, modify `['models', modelID, "r"]`, updating the value for `r` in the model while also sending a `COMM_MESSAGE` to the backend to also update the model in the kernel.

The kernel responds by sending a `COMM_MESSAGE` with `{ "path": ["r"], value: "200" }`
and then also sends the a `COMM_MESSAGE` with a payload for the modified image:

```js
{ "path": ["image"], value: "somebinarydata" }
```

src/notebook/reducers/comms.js
```diff
@@ -1,10 +1,48 @@
...
+
+    const commInfo = {
+      target_module,
+      target_name,
+    };
+
+    return state.setIn(['info', comm_id], Immutable.fromJS(commInfo))
+                .setIn(['models', comm_id], Immutable.fromJS(data));
+  },
+  COMM_MESSAGE: (state, action) => {
```
peggyrayzis: I think you could split the message action into two types here: COMM_MESSAGE_SENT (from UI to kernel) and COMM_MESSAGE_RECEIVED (kernel echoing back message to UI with updated value) to enforce separation of concerns.

src/notebook/components/transforms/model-debug.js
```diff
@@ -0,0 +1,40 @@
...
+    return true;
+  }
+
+  render(): ?React.Element<any> {
+    const { models, data, modelID } = this.props;
+    // TODO: Provide model IDs on transient field
+    // For now, if modelID is not provided (or model does not exist),
+    // show all the models
+    const model = models.get(modelID, models);
+    return (
```
peggyrayzis, line 26: Is this component like our ModelProvider that will talk to Redux and eventually render the UIView?

rgbkrk: Yeah that's the intent with using this in tandem with a vdom style mimetype like we talked about over screenhero. This component is only here for debugging/utility.

src/notebook/reducers/comms.js
rgbkrk: That's a good idea.

src/notebook/components/transforms/model-debug.js
peggyrayzis: Ok awesome, it's all making sense now how this is fitting together. I'm going to draft up a quick gist today to nail down the component structure so we don't forget. Otherwise, this PR LGTM. 👍

rgbkrk:
I definitely still want this in so I'll be rebasing this again in hopes we can start experimenting with the VDOM views and `set` based model updates.

rgbkrk:
Rebased, now to see if I did that well... 😅

rgbkrk:
Update: tried this out using the notebook I included here - working quite well!

jdetle:
Ready for a review?

rgbkrk:
Extremely ready for another review here! Cross reference [Jupyter custom / comm messages](http://jupyter-client.readthedocs.io/en/latest/messaging.html#custom-messages) while you're at it.

src/notebook/components/transforms/index.js
```diff
@@ -58,20 +60,28 @@ export const standardDisplayOrder: StandardDisplayOrder = new ImmutableList([
...
-// Register our custom transforms as the most rich (front of List)
-const displayOrder = standardDisplayOrder
-  .insert(0, PlotlyTransform.MIMETYPE)
-  .insert(0, VegaLite.MIMETYPE)
-  .insert(0, Vega.MIMETYPE)
-  .insert(0, GeoJSONTransform.MIMETYPE);
+const {
+  transforms,
+  displayOrder,
+} = additionalTransforms.reduce(registerTransform, {
```
jdetle, line 81: Nifty little use of destructuring here.

jdetle:
Thanks for the awesome work, once again!

src/notebook/components/transforms/index.js
rgbkrk: I totally forgot I did this! 😄

rgbkrk:
Excellent! Thanks @jdetle. More to come now that this is merged, mostly via the VDOM mimetype and passing messages up from the components/outputs!

💃

PR #1316, opened 2017-01-09
Bump Eslint config and fix lint errors

lgeiger:


.eslintrc
```diff
@@ -20,6 +20,8 @@
     "react/require-extension": "off",
     "comma-dangle": [0],
     "react/jsx-filename-extension": "off",
+    "react/require-default-props": [0],
```
lgeiger, line 28: I didn't know how to fix this lint error: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md

src/notebook/components/cell/code-cell.js
```diff
@@ -76,14 +76,14 @@ class CodeCell extends React.PureComponent {
...
-            this.props.pagers.map((pager, key) =>
+            this.props.pagers.map(pager =>
               <Pager
                 className="pager"
                 displayOrder={this.props.displayOrder}
                 transforms={this.props.transforms}
                 bundle={pager.get('data')}
                 theme={this.props.theme}
-                key={key}
+                key={pager.id}
```
rgbkrk: Where do the id come from, Immutable.js?

rgbkrk:
I just realized that my "review" was in progress from this morning. Pushed the button now!

rgbkrk:
This needs a rebase, I'm going to try the web editor.

rgbkrk:
Ew, it creates a "Merger branch 'master' into X" commit when you rebase.

src/notebook/components/cell/code-cell.js
lgeiger: I'm not sure, I'm not very familiar with Immutable.js. I adapted it from the lint rule description: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

rgbkrk: Ah ok. `pager.id` is `undefined` here - it looks like they were using that for illustrative purposes. In our case here because of how outputs and pagers work (outputs do not get reordered), the array key is fine.

lgeiger: Oh I see. Thanks for the explaination.

lgeiger:
Rebased and fixed my mistake.

rgbkrk (approved):

src/notebook/components/cell/code-cell.js
rgbkrk: It's ok for cases like these if we only turn it off for this section or file, adding in a comment as to why.

PR #1318, opened 2017-01-10
Fix(Toolbar): removed hardwired shouldComponentUpdate

betatim:
Toolbar menu did not update when changing cell type. Is this the right way of fixing this or should we make `shouldComponentUpdate` smarter?

What kind of test could I add to prevent this from coming back in the future? The current test only checks that the correct action is being dispatched.

### DIFF
--- a/src/notebook/components/cell/toolbar.js
+++ b/src/notebook/components/cell/toolbar.js
@@ -47,10 +47,6 @@ export default class Toolbar extends React.PureComponent {
     this.toggleOutputExpansion = this.toggleOutputExpansion.bind(this);
   }
 
-  shouldComponentUpdate(): boolean {
-    return false;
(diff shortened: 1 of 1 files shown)


rgbkrk:
Good catch! One thing we should change is to make this component extend from React.PureComponent

rgbkrk (approved):

betatim:
[It already extends pure component?](https://github.com/nteract/nteract/blob/master/src/notebook/components/cell/toolbar.js#L23)

rgbkrk:
D'oh!

rgbkrk:
As for what test you could add you could write an enzyme test that updates the props and makes sure that the proper markdown menu is shown, etc. I can dig up an example shortly.

PR #1323, opened 2017-01-11
refactor(provider): remove Redux logic from Toolbar

captainsafia:
Parking this here for a minute. 🚗


src/notebook/providers/toolbar.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  executeCell(): void {
+    const { dispatch, id, cell} = this.props;
+    dispatch(executeCell(id,
+                        cell.get('source')));
+  }
+
+  clearOutputs(): void {
+    this.dropdown.hide();
```
peggyrayzis: Everything is looking really great so far! One thing I think we should move to the ToolbarView is the ability to hide the dropdown since it deals with view state vs. application state. The easiest way to accomplish this is to create a method on the view class that hides the dropdown and invokes the function on its props. Then, you can pass that method to the click handler.

src/notebook/providers/toolbar.js
```diff
@@ -0,0 +1,107 @@
...
+  cell: any,
+  id: string,
+  type: string,
+  dropdown: Dropdown,
+  dispatch: Dispatch,
+};
+
+class Toolbar extends Component {
+  props: Props;
+
```
peggyrayzis: Adding Flow typing here such as `removeCell: () => void` should get rid of all your covariant Flow errors. I realized that I never added these to my provider (or Flow to the top of my file 😞 ) so I apologize for being unclear. I'll submit a PR to my provider to fix that tonight.

src/notebook/views/toolbar.js
```diff
@@ -0,0 +1,89 @@
...
+      </Dropdown>
+    </div>
+  </div>
+);
+
+export default class Toolbar extends PureComponent {
+  props: ToolbarProps;
+
+  render(): React.Element<any> {
+    return renderActionButtons(this.props);
```
peggyrayzis: I think you might want renderToolbar here?

peggyrayzis (changes requested):
Looks awesome so far, thanks for tackling this! 😀 Just a couple minor tweaks that should help clear out some of the build errors.

src/notebook/providers/toolbar.js
peggyrayzis: I fixed my provider with #1335. We should also be able to type Dispatch properly now too 😀

src/notebook/views/toolbar.js
captainsafia: OMG THIS IS WHY IT WASN'T WORKING! OMG I AM THE WORST PROGRAMMER. OMG WHY?!!? OMG *THANK YOU*!!!

peggyrayzis: Haha you are definitely not the worst programmer 😝 sometimes a second pair of eyes makes all the difference!

src/notebook/providers/toolbar.js
captainsafia: Can you elaborate on this a little bit more?

peggyrayzis: Sure! 😀 Providers are only concerned with application level state and are connected to the Redux store. Views are unconnected to the store and should only be concerned with view level state isolated to that component (things like toggling a button or in our case, hiding a dropdown). This pattern helps us enforce separation of concerns.

Another reason we can't hide the dropdown on our provider is because we're using refs to invoke methods on the Dropdown component. Refs are destroyed when their children are destroyed, so we have to sync up establishing the ref with ToolbarView's render method instead of ToolbarProvider's.

Provider:

```js
clearOutputs() {
  const { dispatch, id } = this.props;
  dispatch(clearOutputs(id));
}
// pass down this method to the view through props
```

View:

```js
// define a method called clearOutputs on the class & bind it in the constructor
clearOutputs () {
  this.dropdown.hide()
  this.props.clearOutputs()
}

render () {
  <li onClick={() => this.clearOutputs()} />
}
```

You might want to move all of the content from renderToolbar into the ToolbarView's actual render method since there's no conditional logic. Here, you would want to leave out clearOutputs from the this.props destructuring since we want to invoke the clearOutputs function on the class.

captainsafia: Ah! I think I get it! Thanks for the explanation!

jdetle:
Hi @captainsafia, I made a change in #1403 that will cause a conflict for this PR.

jdetle:
I'm going to try to finish this off @captainsafia, hope thats okay

captainsafia:
Go ahead!

jdetle:
@peggy, This is ready for a review!

src/notebook/providers/toolbar.js
```diff
@@ -0,0 +1,113 @@
...
+  executeCell(): void {
+    const { dispatch } = this.props;
+    dispatch(executeCell(this.props.id,
+                        this.props.cell.get('source')));
+  }
+
+  clearOutputs(dropdown): void {
+    const { dispatch } = this.props;
+    dropdown.hide();
```
peggyrayzis: The same feedback that I gave to Safia still applies here. We want to control hiding the dropdown on the view, not the provider, because the view is where the ref is initialized. Passing the dropdown in here seems like we're introducing too much of a circular dependency.

Here's the feedback I gave Safia for your reference. Please let me know if anything is unclear, I'd be happy to jump on a quick call if it's confusing. 😀 
toolbar feedback 1
toolbar feedback 2

peggyrayzis (changes requested):
Thanks for taking this Toolbar refactor on! We still need to remove hiding the dropdown from the provider before I can merge. The tests look great though 👍

src/notebook/providers/toolbar.js
jdetle: OH. I get it. Let me see if I can take care of that right quick.

jdetle: Sorry that you had to repeat yourself.

peggyrayzis: No worries! I had to dig around to find my original feedback, GH hides it once you add new commits 😞

jdetle:
Think I took care of it, thanks @peggyrayzis

peggyrayzis (approved):
The changes look great! Thanks for making that fix. 🎉

rgbkrk:
Woo!

PR #1332, opened 2017-01-12
chore(flow): craft explicit flow types for notebook

rgbkrk:
I'm going to start getting our state to be a bit more exact to make typing of our reducers, actions, and epics much easier. This starts out by creating flow types for the overall notebook state.

### DIFF
--- a/src/notebook/records.js
+++ b/src/notebook/records.js
@@ -67,7 +67,32 @@ export type MarkdownCell = {
 
 export type Cell = MarkdownCell | CodeCell;
 
+export type KernelspecMetadata = {
+  name: string,
+  display_name: string,
+}
+
+export type LanguageInfoMetadata = {
+  name: string,
+  codemirror_mode?: string | ImmutableJSONMap,
+  file_extension?: string,
+  mimetype?: string,
+  pygments_lexer?: string,
+}
+
 export type NotebookMetadata = {
+  kernelspec: KernelspecMetadata,
+  language_info: LanguageInfoMetadata,
+  // We're not currently using orig_nbformat in nteract. Based on the comment
+  // in the schema, I'm not sure we should:
+  //
+  //   > Original notebook format (major number) before converting the notebook between versions. This should never be written to a file
+  //
+  //   from https://github.com/jupyter/nbformat/blob/62d6eb8803616d198eaa2024604d1fe923f2a7b3/nbformat/v4/nbformat.v4.schema.json#L58-L61
+  //
+  // It seems like an intermediate/in-memory representation that bled its way into the spec
+  //
+  // orig_nbformat?: number,
 }
 
 export type Notebook = {
@@ -91,6 +116,17 @@ export const AppRecord = Immutable.Record({
   configLastSaved: null,
   error: null,
 });
+
+export type Document = {
+  notebook: Notebook,
+  transient: Immutable.Map<string, any>,
+  cellPagers: any,
+  outputStatuses: Immutable.Map<string, any>,
+  stickyCells: Immutable.Set<any>,
+  editorFocused: any,
+  cellFocused: any,
+  copied: Immutable.Map<any, any>,
+}
 
 export const DocumentRecord = Immutable.Record({
   notebook: null,
--- a/src/notebook/records.js
+++ b/src/notebook/records.js
@@ -1,6 +1,84 @@
-import Immutable from 'immutable';
+/* @flow */
 
-export const AppRecord = new Immutable.Record({
+const Immutable = require('immutable');
+
+/*
+
+This is the definition of JSON that Flow provides
+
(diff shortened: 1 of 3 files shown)


rgbkrk:
This last commit is going to fail due to `flow`. 😢  Immutable's `Record` is missing a definition for the constructor call of the "class"es it returns.

rgbkrk:
Now for a commit that fails in the other direction:

```
src/notebook/index.js:33
 33:   document: DocumentRecord(),
                 ^^^^^^^^^^^^^^^^ function call. Callable signature not found in
 33:   document: DocumentRecord(),
                 ^^^^^^^^^^^^^^ Record
```

rgbkrk:
Since `Record`'s returned objects also don't have a Callable signature, also stuck there. 😢

rgbkrk:
and this last commit will pass.

rgbkrk:
If I set the container type for a Document Record as:

```
export const DocumentRecord: Immutable.Record<Document> = Immutable.Record({
  notebook: null,
  transient: new Immutable.Map({
    keyPathsForDisplays: new Immutable.Map(),
  }),
  cellPagers: new Immutable.Map(),
  outputStatuses: new Immutable.Map(),
  stickyCells: new Immutable.Set(),
  editorFocused: null,
  cellFocused: null,
  copied: new Immutable.Map(),
});
```

that's when it's unable to determine that it's a Callable. Any ideas for workarounds @peggyrayzis?

rgbkrk:
Solid SO answer here about why: http://stackoverflow.com/questions/37033339/how-to-describe-immutable-js-map-shape-with-flow

rgbkrk:
This makes me want to switch off of Immutable because I love `flow` so much, yet there are no other good libraries in this space that have a good `flow` story either (playing with `lodash/fp` taught me this).

rgbkrk:
However, with `lodash/fp` these type definitions would match really well.

rgbkrk:
Interesting way to do it that lets you use Records in the way we expect:

```
export type DocumentSpec = {
  notebook: Notebook,
  transient: Immutable.Map<string, any>,
  cellPagers: any,
  outputStatuses: Immutable.Map<string, any>,
  stickyCells: Immutable.Set<any>,
  editorFocused: any,
  cellFocused: any,
  copied: Immutable.Map<any, any>,
}

export type Document = DocumentSpec & Immutable.Record<DocumentSpec>

export const DocumentRecord: Document = Immutable.Record({
  notebook: null,
  transient: new Immutable.Map({
    keyPathsForDisplays: new Immutable.Map(),
  }),
  cellPagers: new Immutable.Map(),
  outputStatuses: new Immutable.Map(),
  stickyCells: new Immutable.Set(),
  editorFocused: null,
  cellFocused: null,
  copied: new Immutable.Map(),
});
```

Still can't have the Callable signature.

Also `Function cannot be called on any member of intersection type`

src/notebook/records.js
```diff
@@ -14,7 +117,18 @@ export const AppRecord = new Immutable.Record({
...
 });
 
-export const DocumentRecord = new Immutable.Record({
+export type Document = {
+  notebook: Notebook,
+  transient: Immutable.Map<string, any>,
+  cellPagers: any,
+  outputStatuses: Immutable.Map<string, any>,
+  stickyCells: Immutable.Set<any>,
+  editorFocused: any,
```
peggyrayzis, line 126: I think you can mark editorFocused & cellFocused as string|null here.

peggyrayzis (approved):
What you have now serves our purpose well, although it sucks that we can't define the object as a Callable. Honestly, I need to look into the typing of Callable objects closer to suggest any other possible workarounds. Do you think it would be worth working on a PR to the flow-typed immutable libdef to make records callable?

peggyrayzis:
Also, thank you for typing all of this. I like having the types in records, I think it makes the code more discoverable 😀

rgbkrk:
> Also, thank you for typing all of this. I like having the types in records, I think it makes the code more discoverable 😀

You make me feel more positive about including this since at the very least the types can be used with the reducers.

rgbkrk:
> Do you think it would be worth working on a PR to the flow-typed immutable libdef to make records callable?

The libdef for Immutable is in the Immutable.js repo. The number of open PRs attempting to fix the flow definitions as well as the unshipped commits make me worried about whether there's a good way to even go about this.

rgbkrk:
Alright, pushed a commit that at least passes and will be useful for people to see/use.

peggyrayzis:
Yeah, I think it will definitely be useful as we start to type the reducers. Is this still WIP?

rgbkrk:
Good to go!

rgbkrk:
I'm going to be trying out https://www.npmjs.com/package/flow-immutable-models with jscodeshift today, seeing how the code generation works out.

peggyrayzis:
Keep me updated, it looks awesome!

PR #1333, opened 2017-01-13
Remove Redux Logic from Draggable Cell 

jdetle:
Putting this up here to keep myself accountable :) 
More to come soon.


jdetle:
Ready for a review @peggyrayzis!

src/notebook/providers/draggable-cell.js
```diff
@@ -0,0 +1,38 @@
...
+import { focusCell } from '../actions';
+
+import DraggableCellView from '../views/draggable-cell';
+
+type Props = {|
+  dispatch: Dispatch<*>,
+  id: string,
+|};
+
+class DraggableCell extends React.PureComponent {
```
peggyrayzis: Here, we should probably extend from Component. Extending from PureComponent would return false for all shouldComponentUpdate calls, which could interfere with what react-redux is doing under the hood.

https://github.com/reactjs/react-redux/blob/master/src/components/connectAdvanced.js#L142

peggyrayzis (changes requested):
Looks good, thanks for completing the refactor for DraggableCell! 😀 One minor fix on the provider is the only issue I see. Eventually, we should look into refactoring the DraggableCellView so it matches the object destructuring style we have in the other refactored view components, but we don't need to worry about that immediately.

Also, we should try to write some tests for the provider in the near future. You can check out the tests for CellCreator as an example.

src/notebook/providers/draggable-cell.js
peggyrayzis: Also, could you please rebase some of the linting/flow commits and reformat your PR title? After those small changes, we should be good to go 👍

jdetle:
Done! Thanks.

peggyrayzis (approved):

peggyrayzis:
Thanks for making the changes @jdetle! One step closer to completing #1288 😀

PR #1335, opened 2017-01-14
chore(flow): Started adding Redux types, enabled typing on provider

peggyrayzis:
- Enabled Flow on the cell creator provider
- Added types lib for global types (@rgbkrk we can possibly put our generated model files here if we end up going that route)

We should consider dropping the eslint-plugin-react no-unused-prop-types rule. [It throws a false positive](https://github.com/yannickcr/eslint-plugin-react/issues/816) if you destructure your props in your component.

### DIFF
--- a/src/notebook/views/cell-creator.js
+++ b/src/notebook/views/cell-creator.js
@@ -1,19 +1,21 @@
 // @flow
+/* eslint-disable react/no-unused-prop-types */
+
 import React, { PureComponent } from 'react';
 import { throttle } from 'lodash';
 
-declare type CellCreatorProps = {|
+type Props = {
   above: boolean,
   id: string|null,
   createCell: (type: string) => void,
   mergeCell: () => void,
-|};
+};
 
 type State = {|
   show: boolean,
 |};
 
-const renderActionButtons = ({ above, createCell, mergeCell }: CellCreatorProps) => (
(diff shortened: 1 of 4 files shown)


src/notebook/views/cell-creator.js
```diff
@@ -1,19 +1,21 @@
 // @flow
+/* eslint-disable react/no-unused-prop-types */
+
 import React, { PureComponent } from 'react';
 import { throttle } from 'lodash';
 
-declare type CellCreatorProps = {|
+type Props = {
```
jdetle, line 7: Why the change from `declare type CellCreatorProps` to `type Props`?

jdetle (commented):
Thanks for the PR, ready for merge?

src/notebook/views/cell-creator.js
peggyrayzis: I think it was overkill for me to specify CellCreatorProps. We're checking the prop structure on the CellCreator view when the class is created so there's no reason for me to check that the {...props} being passed down in CellCreatorProvider's render method are also the same prop structure.

peggyrayzis:
Ready for merge 😀

rgbkrk:
> We should consider dropping the eslint-plugin-react no-unused-prop-types rule. It throws a false positive if you destructure your props in your component.

Sounds good to me!

PR #1336, opened 2017-01-15
chore(flow): fully type config.js

rgbkrk:
In order to fully type the reducers (with full flow coverage) I noticed I had to do several things:

* Switch away from using constants
  * Good news though! Flow detects and ensures the type names are used.
* Break away from `handleActions` since it can't have the same flow based typing
  * We get some great typing benefits from this and this will end up cutting out a dependency that we were kind of treating like magic.

### DIFF
--- a/src/notebook/reducers/config.js
+++ b/src/notebook/reducers/config.js
@@ -1,14 +1,31 @@
-import { handleActions } from 'redux-actions';
+/* @flow */
 
-import * as constants from '../constants';
+import { Map } from 'immutable';
 
-export default handleActions({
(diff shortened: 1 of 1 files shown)


rgbkrk:
If folks like this I'll happily tackle some other ones.

rgbkrk:
Since we're getting to the point of exporting all of the modules, we'll probably want to combine action creators and reducers into the same module file.

peggyrayzis (approved):
This PR is awesome! I think it's a really great structure for how we should type our Redux files. I can type some of the files too, just let me know how you'd like to split it up!

PR #1337, opened 2017-01-15
refactor(document): type annotate and refactor document reducers

rgbkrk:
Following in the foot steps of #1336 

Not ready yet.

### DIFF
--- a/src/notebook/reducers/document.js
+++ b/src/notebook/reducers/document.js
@@ -4,7 +4,6 @@ import * as Immutable from 'immutable';
 import { handleActions } from 'redux-actions';
 import * as uuid from 'uuid';
 import * as commutable from 'commutable';
-import { has } from 'lodash';
 
 import * as constants from '../constants';
 
@@ -13,6 +12,7 @@ import * as constants from '../constants';
 // We can swap these out later.
 
 type ImmutableOutput = Immutable.Map<string, any>;
+type ImmutableOutputs = Immutable.List<ImmutableOutput>;
 
 // Non-immutable output
 type JSON = | string | number | boolean | null | JSONObject | JSONArray; // eslint-disable-line no-use-before-define
@@ -34,6 +34,7 @@ export type DisplayData = {
   output_type: 'display_data',
   data: MimeBundle,
   metadata: JSONObject,
+  transient?: JSONObject,
 }
 
 export type StreamOutput = {
@@ -64,7 +65,7 @@ type DocumentState = Immutable.Map<string, any>; // & Document;
  * @param {Object} output - Outputted to be reduced into list of outputs
  * @return {Immutable.List<Object>} updated-outputs - Outputs + Output
  */
-export function reduceOutputs(outputs: Immutable.List<ImmutableOutput> = Immutable.List(), output: Output) { // eslint-disable-line max-len
+export function reduceOutputs(outputs: ImmutableOutputs = Immutable.List(), output: Output) { // eslint-disable-line max-len
   if (output.output_type !== 'stream' ||
      (outputs.size > 0 && outputs.last().get('output_type') !== 'stream')) {
     // If it's not a stream type, we just fold in the output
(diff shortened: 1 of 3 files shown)


rgbkrk:
I really need to write up a manifesto on our state, so I'll do a light treatment here:

## Notebook Document

* Everything that eventually ends up persisted to the notebook document

## Transient, Associated with Kernel

* Display ID
* Comms, Models (ipywidgets)

## Transient, Associated with Document

* Cell view output

## Transient app state (local to single user)

* Focused cell
* Copied cell (?)

rgbkrk:
For those following along, I'm amending to the same commit every time it's in a nice stable state rather than creating lots of commits. This is quite a heavy refactor to undertake.

rgbkrk:
Created new notebooks, loaded old notebooks, and ran the conformance tests -- looking good from a compliance standpoint!

src/notebook/reducers/document.js
```diff
@@ -47,325 +125,528 @@ export function cleanCellTransient(state, id) {
...
+function moveCell(state: DocumentState, action: MoveCellAction) {
+  return state.updateIn(['notebook', 'cellOrder'],
+    (cellOrder: CellOrder) => {
+      const oldIndex = cellOrder.findIndex(id => id === action.id);
+      const newIndex = cellOrder.findIndex(id => id === action.destinationId)
+                        + (action.above ? 0 : 1);
+      if (oldIndex === newIndex) {
+        return cellOrder;
+      }
+      return cellOrder
```
rgbkrk, line 332: This line is not covered by our current test suite.

src/notebook/reducers/document.js
```diff
@@ -15,30 +84,39 @@ import * as constants from '../constants';
...
         nextToLast.get('output_type') === 'stream' &&
-        nextToLast.get('name') === output.name) {
-      return outputs.updateIn([outputs.size - 2, 'text'], text => text + output.text);
+        nextToLast.get('name') === streamOutput.name) {
+      return outputs.updateIn([outputs.size - 2, 'text'], appendText);
     }
   }
 
-  return outputs.push(Immutable.fromJS(output));
+  return outputs.push(Immutable.fromJS(streamOutput));
```
rgbkrk, line 124: This line is not covered by our current test suite.

src/notebook/reducers/document.js
captainsafia: Yeah, this is that tricky one where we couldn't figure out exactly how it was behaving because it seemed like that line would never be hit based on the conditional and the way that `newIndex` and `oldIndex` were set.

src/notebook/reducers/document.js
```diff
@@ -47,325 +125,528 @@ export function cleanCellTransient(state, id) {
...
-    }
+function setNotebook(state: DocumentState, action: SetNotebookAction) {
+  const notebook = action.notebook
+    .update('cellMap', (cells: CellMap): CellMap =>
+      // TODO: Determine why setting the notebook is changing these values.
+      //       Should they be transient or part of an overall app state?
+      cells.map(value =>
+        value.setIn(['metadata', 'inputHidden'], false)
+              .setIn(['metadata', 'outputHidden'], false)
+              .setIn(['metadata', 'outputExpanded'], false)));
```
rgbkrk, line 150: @captainsafia do we want to be persisting `outputHidden` to the notebooks or is it only for at runtime? The fact that these get cleaned up on load seems like we're currently:


```
* Setting metadata on the notebooks
* Saving with the metadata (woo my hidden setting is acknowledged)
* Resetting them on load?
```

src/notebook/reducers/document.js
rgbkrk: When a user is using this, it must be getting hit because otherwise cells would stay fixed in the prior order. 🤔

src/notebook/reducers/document.js
captainsafia: These values are currently persisted to the notebook on save. The reason we didn't do this was because there was some discussion about persisting arbitrary (not defined in the notebook format) values. So as of now, this configuration is only used during runtime and forgotten once the user closes out of nteract.

rgbkrk: Ok, cool thanks for clarifying it! This helps me see why these are being set to false here. It isn't to wipe out what the user may have had set in the notebook, it's for the structure to be available within this cell when rendered. The colocation of the data here is for our internal use (it makes sense to have it as part of the cell), and not persisted to the document.

Next steps as we start refactoring state to scope is to figure out if we:

* Keep state like this colocated with the cell, initialize on load, delete/sanitize before save
* Create a separate state structure for transient-at-document-runtime (as opposed to the current transient which is transient-by-kernel-runtime)

rgbkrk: > there was some discussion about persisting arbitrary (not defined in the notebook format) values

Is this a feature where we should persist them and use our own prefix:

```
metadata: {
  "nteractOutputHidden": ...
}
```

We would definitely be on the hook for maintaining this as our own spec (including migrations in the future).

rgbkrk: Where we've run into issues before with not conforming to the spec was when keys were set above metadata, which [does allow for additional properties](https://github.com/jupyter/nbformat/blob/62d6eb8803616d198eaa2024604d1fe923f2a7b3/nbformat/v4/nbformat.v4.schema.json#L175).

src/notebook/reducers/document.js
rgbkrk: I've submitted a test that covers this now, it seemed important enough to make sure we reach it with enough edge cases.

rgbkrk:
This should now be fully covered after the next test + codecov run!

peggyrayzis:
Feels awesome to merge this PR! Thanks for getting the ball rolling w/ typing the reducers 👍

jdetle:
Very leet PR here.

rgbkrk:
> Thanks for getting the ball rolling w/ typing the reducers

My pleasure, I'm glad we took the time to do this. Helped me see a lot of the state we need to clean up.

PR #1338, opened 2017-01-15
chore(kernels): Store kernel specs on startup

### LINKED ISSUE #1324: Store kernel specs on startup
Currently `kernelspecs.findAll()` (**Edit:** and `kernelspecs.find()` are) called multiple times throughout the code base. (on startup, on menu init, on kernel launch via spawnteract).

Since it is a promise I think we should only call it once (on startup)  and store the kernel specs in a variable which then can be used throughout the code base.

This should improve performance everywhere kernel specs are used.
Also it would make it possible to very easily add additional kernel specs of bundled kernels in one central place.

E.g. [`ikernel`](https://github.com/rgbkrk/ikernel) could be simply bundled (as a npm dependency) by adding the following spec on startup:

```json
{
  "argv": [
    "path to executable, accessible with `process.execPath`",
    "path to node_modules/ikernel",
    "{connection_file}"
  ],
  "display_name": "ikernel (Node.js)",
  "language": "javascript",
  "env": {
    "ELECTRON_RUN_AS_NODE": 1
  }
}
```

lgeiger:
Before `kernelspecs.find()` and `kernelspecs.findAll()` were called multiple times throughout the code base. Kernel specs are now stored on startup.

This makes it possible to very easily add additional kernel specs of bundled kernels in one central place.

And it slightly improves performance 😄 :
bildschirmfoto 2017-01-15 um 14 36 45

Fix #1324

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -15,9 +15,10 @@ import {
   launchNewNotebook,
 } from './launch';
 
-import { defaultMenu, loadFullMenu } from './menu';
+import { loadFullMenu } from './menu';
 
 import prepareEnv from './prepare-env';
+import saveKernelspecs from './kernel-specs';
 
 const log = require('electron-log');
 
@@ -90,7 +91,8 @@ const prepJupyterObservable = prepareEnv
 
 const kernelSpecsPromise = prepJupyterObservable
   .toPromise()
-  .then(() => kernelspecs.findAll());
+  .then(() => kernelspecs.findAll())
+  .then(kernelSpecs => saveKernelspecs(kernelSpecs));
 
 /**
  * Creates an Rx.Subscriber that will create a splash page onNext and close the
@@ -203,10 +205,8 @@ fullAppReady$
   .subscribe(() => {
     kernelSpecsPromise.then((kernelSpecs) => {
       if (Object.keys(kernelSpecs).length !== 0) {
-        // Get the default menu first
-        Menu.setApplicationMenu(defaultMenu);
-        // Let the kernels/languages come in after
-        loadFullMenu().then(menu => Menu.setApplicationMenu(menu));
+        const menu = loadFullMenu(kernelSpecs);
+        Menu.setApplicationMenu(menu);
       } else {
         dialog.showMessageBox({
           type: 'warning',
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -91,9 +91,9 @@ const prepJupyterObservable = prepareEnv
 
 const kernelSpecsPromise = prepJupyterObservable
   .toPromise()
-  .then(() => kernelspecs.findAll())
-  .then(kernelSpecs => saveKernelspecs(kernelSpecs));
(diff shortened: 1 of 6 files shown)


src/main/index.js
```diff
@@ -90,7 +91,8 @@ const prepJupyterObservable = prepareEnv
 
 const kernelSpecsPromise = prepJupyterObservable
   .toPromise()
-  .then(() => kernelspecs.findAll());
+  .then(() => kernelspecs.findAll())
```
rgbkrk: I'd split this off in two:

```js
 const kernelSpecsPromise = prepJupyterObservable
    .toPromise()
    .then(() => kernelspecs.findAll());

kernelSpecsPromise.then(saveKernelspecs);
```

Then make it so `saveKernelspecs` doesn't have to return the kernelSpecs.

rgbkrk (approved):
I'm a little squeamish about letting people carve out a space for globals, we can give this a try though. With Electron, main <--> renderer calls are typically blocking calls which will have greater affect with us with more notebook windows open.

rgbkrk:
To clarify my squeamishness about the globals - I'd rather we manage the state on the main side with another `store` and have an explicit ipc structure.

src/main/index.js
lgeiger: 👍

lgeiger:
@rgbkrk Thanks for the feedback. I changed it to only rely on async ipc calls.

src/notebook/epics/kernel-launch.js
```diff
@@ -122,6 +120,20 @@ export const watchExecutionStateEpic = action$ =>
     );
 
 /**
+  * Get kernel specs from main process
+  *
+  * @returns  {Observable}  The reply from main process
+  */
+export const kernelSpecsObservable =
+  Rx.Observable.create((observer) => {
+    ipc.send('kernel_specs_request');
```
rgbkrk: The send should go after the `on` is registered, just in case the response happens before the registration.

test/renderer/epics/kernel-launch-spec.js
```diff
@@ -160,7 +160,7 @@ describe('newKernelByNameEpic', () => {
   it('creates a LAUNCH_KERNEL action in response to a LAUNCH_KERNEL_BY_NAME action', (done) => {
     const input$ = Rx.Observable.of({
       type: constants.LAUNCH_KERNEL_BY_NAME,
-      kernelSpecName: 'python2.8',
```
rgbkrk, line 163: Amusingly, this was `python2.8` here on purpose so that it would not exist to ensure we're testing the right thing and mocked properly.

rgbkrk:
Let's go with this!

test/renderer/epics/kernel-launch-spec.js
lgeiger: I can change it back to `python2.8` if you like 👍  😄

PR #1339, opened 2017-01-15
Bundle ijavascript

lgeiger:
Based on #1338 

This is mainly for demonstration purpose ~~and not tested with a packaged release~~:
ikernel

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,6 +132,7 @@
     "enchannel": "^1.1.3",
     "enchannel-zmq-backend": "^2.6.0",
     "github": "^8.0.0",
+    "ijavascript": "^5.0.17",
     "immutable": "^3.7.6",
     "jupyter-paths": "^1.1.0",
     "kernelspecs": "^1.1.0",
--- a/src/main/kernel-specs.js
+++ b/src/main/kernel-specs.js
@@ -1,9 +1,29 @@
 import { ipcMain as ipc } from 'electron';
+import { join } from 'path';
 
-const KERNEL_SPECS = {};
+const KERNEL_SPECS = {
+  node_nteract: {
+    name: 'node_nteract',
+    spec: {
+      argv: [
+        process.execPath,
+        join(require.resolve('ijavascript'), '..', 'lib', 'kernel.js'),
+        '{connection_file}',
+        '--protocol=5.0',
+        '--hide-undefined'
+      ],
+      display_name: 'Node.js (nteract)',
+      language: 'javascript',
+      env: {
+        ELECTRON_RUN_AS_NODE: '1'
+      }
+    }
+  }
+};
 
-export default function saveKernelspecs(kernelSpecs) {
+export default function initializeKernelSpecs(kernelSpecs) {
   Object.assign(KERNEL_SPECS, kernelSpecs);
+  return KERNEL_SPECS;
 }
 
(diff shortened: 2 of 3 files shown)


src/main/kernel-specs.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default function saveKernelspecs(kernelSpecs) {
+  global.KERNEL_SPECS = {
+    node: {
+      name: 'node',
+      spec: {
+        argv: [
+          process.execPath,
+          path.dirname(require.resolve('ikernel')),
+          '{connection_file}'
```
rgbkrk: Super cool!

rgbkrk:
Feedback from video chat this morning:

We should probably build on top of [`jp-kernel`](https://github.com/n-riesco/jp-kernel) for our embedded kernel or we'll have to build a lot more into `ikernel` (which was largely created as a demo js kernel by myself).

lgeiger:
Against my assumptions `ijavascript` doesn't install the kernel spec on `npm install` which is awesome and makes it usable for bundling.

Let's get this PR out of the demo stage 🚀  🔜

package.json
```diff
@@ -132,6 +132,7 @@
     "enchannel": "^1.1.3",
     "enchannel-zmq-backend": "^2.6.0",
     "github": "^8.0.0",
+    "ijavascript": "github:lgeiger/ijavascript",
```
lgeiger: Waiting for https://github.com/n-riesco/ijavascript/pull/92 to be merged.

lgeiger:
Ready to try out:
ijavascript

rgbkrk:
I'm a bit hesitant with bringing in something that relies on a github repo directly, want to publish a namespaced package of `ijavascript` as `@lgeiger/ijavascript`?

lgeiger:
> I'm a bit hesitant with bringing in something that relies on a github repo directly

Me too, if think we can wait till https://github.com/n-riesco/ijavascript/pull/92 gets merged

rgbkrk:
Should we label this as "node.js (nteract)" for the display name so it doesn't confust/conflict with an installed JavaScript kernel?

src/main/kernel-specs.js
```diff
@@ -1,9 +1,29 @@
...
+import { join } from 'path';
 
-const KERNEL_SPECS = {};
+const KERNEL_SPECS = {
+  node_nteract: {
+    name: 'node_nteract',
+    spec: {
+      argv: [
+        process.execPath,
+        join(require.resolve('ijavascript'), '..', 'lib', 'kernel.js'),
```
lgeiger, line 10: A bit hacky but it works as long as ijavascripts folder structure doesn't change.

For the long term we can make a custom kernel as outlined in https://github.com/n-riesco/ijavascript/pull/92.

rgbkrk: If we pin exactly we'll be alright.

rgbkrk (approved):

rgbkrk:
We should switch our status bar to showing the display name of the kernel I think:

screen shot 2017-01-18 at 9 39 17 am

captainsafia:
> We should switch our status bar to showing the display name of the kernel I think:

That makes sense. I believe I used to spec name when I first made the status bar just because it was what we stored in the state tree, so no real reason to have it this way in any case.

PR #1342, opened 2017-01-17
chore(lint): Remove no-unused-prop-types eslint rule

peggyrayzis:
The no-unused-prop-types rule throws false positives if you destructure your props in a component as I mentioned in #1335. We should turn it off until this issue is fixed.


jdetle:
Thanks @peggyrayzis

PR #1348, opened 2017-01-18
flow(app): flow type the app reducer

rgbkrk:
Type the `app` reducer this time.

### DIFF
--- a/src/notebook/kernel/shutdown.js
+++ b/src/notebook/kernel/shutdown.js
@@ -1,11 +1,28 @@
+/* @flow */
 import * as fs from 'fs';
+
+import type { Subject } from 'rxjs';
+// import type { ChildProcess } from 'child_process'; // eslint-disable-line no-unused-vars
 
 import { shutdownRequest } from 'enchannel';
 import { getUsername, session } from './messaging';
 
 export const filesystem = fs;
 
-export function cleanupKernel(kernel, closeChannels) {
+export type Channels = {
+  iopub: Subject,
+  stdin: Subject,
+  shell: Subject,
+  control: Subject,
+}
+
+export type Kernel = {
+  channels: Channels,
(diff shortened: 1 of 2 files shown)


jdetle (approved):
Are we eventually going to have no constants for our redux actions?

rgbkrk:
> Are we eventually going to have no constants for our redux actions?

Yup, `flow` determines these for us. We could still have constants if we want them as is though, because `flow` figures out `constants.BLAH` properly too.

rgbkrk:
Couple things this PR made me realize:

* The ZMQ/native local kernel portion of the state should be separate to make exporting easier (can't and won't be used by a web app)
* The github state should be in a separate module too

src/notebook/kernel/shutdown.js
```diff
@@ -1,11 +1,28 @@
...
+  control: Subject,
+}
+
+export type Kernel = {
+  channels: Channels,
+  spawn: any, // ChildProcess,
+  connectionFile: string,
+}
+
+export function cleanupKernel(kernel: Kernel, closeChannels: boolean) {
```
peggyrayzis: Should we type the returns of these functions? 🤔 Does it even matter?

rgbkrk: These are void returns with necessary side effects. 😉  I can add some on.

rgbkrk: Added on some return types. There was one with a `Promise<undefined>` which I switched to `Promise<boolean>` (it was only there to signal completion).

peggyrayzis: Awesome, this is helpful for understanding what's going on!

peggyrayzis:
Thanks again for another great Flow PR 🎉 !

PR #1350, opened 2017-01-18
chore(kernels): Add help menu item

lgeiger:
~~Since we now have our first bundled Node.js kernel, nteract can be used without user installed kernels.~~

This PR ~~refactors the error handling if no kernel specs are detected and~~ adds a menu item to provide more informations about installing additional kernels.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -407,6 +407,10 @@ const helpDraft = {
     {
       label: 'Learn More',
       click: () => { shell.openExternal('http://github.com/nteract/nteract'); }
+    },
+    {
+      label: 'Install Additional Kernels',
+      click: () => { shell.openExternal('https://ipython.readthedocs.io/en/latest/install/kernel_install.html'); }
     }
   ]
 };


src/main/index.js
```diff
@@ -204,25 +204,8 @@ openFile$
 fullAppReady$
   .subscribe(() => {
     kernelSpecsPromise.then((kernelSpecs) => {
-      if (Object.keys(kernelSpecs).length !== 0) {
```
lgeiger, line 207: `kernelSpecs` will now always contain the bundled Node.js kernel

rgbkrk (commented):
I think we want to keep the edge case for no kernels around even if our configuration defaults to having one.

rgbkrk:
Super happy this provides a link out for help with creating new kernels.

lgeiger:
> I think we want to keep the edge case for no kernels around even if our configuration defaults to having one.

Alright. I reverted 654f1ec6d3465c120374d1266d41ae68d13c7840

rgbkrk:
🙇 Thanks

PR #1355, opened 2017-01-20
refactor(cell): Add outputs for clarity.

jdetle:
Feel free to weigh in on this, I think its just more clear to include `Outputs` here.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -240,11 +240,11 @@ export const cell = {
       click: createSender('menu:run-all-below'),
     },
     {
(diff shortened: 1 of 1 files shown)


rgbkrk:
Great addition!

rgbkrk (approved):

rgbkrk:
We should probably find a better verb for changes like this other than refactor. `ui`? `ux`?

jdetle:
^ works for me regarding adding that into the lexicon.

PR #1358, opened 2017-01-20
Allow for exporting of the notebook as a PDF

jdetle:
Seemed to have deleted and recreated the wrong branch. Reopening here.

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -311,6 +311,13 @@ export function dispatchExportPDF(store) {
     position: 'tr',
     level: 'success',
   });
+  remote.getCurrentWindow().webContents.printToPDF({ printBackground: true }, (error, data) => {
+      if (error) throw error
+      fs.writeFile('/tmp/print.pdf', data, (error) => {
+        if (error) throw error
+        console.log('Write PDF successfully.')
+      });
+    });
   store.dispatch(exportPDF(filename))
 }
 
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -2,6 +2,7 @@ import {
   ipcRenderer as ipc,
   webFrame,
   remote,
+  shell,
 } from 'electron';
 
 import * as path from 'path';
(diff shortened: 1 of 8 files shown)


jdetle:
I need to add tests but if anyone wants to check this out its ready for a review. @rgbkrk, @peggyrayzis, @captainsafia, @lgeiger.

src/notebook/menu.js
```diff
@@ -295,6 +302,63 @@ export function dispatchNewNotebook(store, event, kernelSpec) {
...
+
+export function triggerSaveAsPDF(store) {
+  showSaveAsDialog()
+    .then(filename =>
+      Promise.all(
+        [triggerWindowRefresh(store, filename),
+          triggerKernelRefresh(store)]
+      )
+    )
+    .then(() => storeToPDF(store));
```
peggyrayzis: At the end of the promise chain, we should catch any errors here with `catch(e => console.error(e))`. You can substitute console.error for however else we handle errors throughout the app.

Maybe we could also display some kind of error message to the user via the notification system that their action was not completed? 🤔

peggyrayzis (changes requested):
Awesome work on this feature!! It looked really cool when you demoed it on Screenhero last night. 😎 One small change regarding error handling and then this should be ready to merge!

jdetle:
I have absolutely no idea how to go about testing 
```javascript
export function triggerSaveAsPDF(store) {
  showSaveAsDialog()
    .then(filename =>
      Promise.all(
        [triggerWindowRefresh(store, filename),
          triggerKernelRefresh(store)]
      )
    )
    .then(() => storeToPDF(store))
    .catch(e => store.dispatch({ type: 'error', payload: e.message, error: true }));
}
```
and given that TriggerSaveAs seems to be uncovered as well, I think other people have had the same problem. Any ideas?

jdetle:
I could make the trigger functions take the functions that call inside as arguments, but I'd rather be able to stub them in the tests. Unfortunately that's not possible to the best of my knowledge.

jdetle:
Hi @peggyrayzis, are the changes ok?

src/notebook/menu.js
```diff
@@ -331,7 +331,8 @@ export function triggerSaveAsPDF(store) {
           triggerKernelRefresh(store)]
       )
     )
-    .then(() => storeToPDF(store));
+    .then(() => storeToPDF(store))
+    .catch(e => store.dispatch({ type: 'error', payload: e.message, error: true }));
```
peggyrayzis: Nice, that's definitely how you catch an error at the bottom of a promise chain but I'm not sure if that object is what we want to dispatch to Redux. Usually, dispatch takes a function (action creator) that returns an action with a type & payload. Although, maybe redux-observable handles things differently with the epics and I'm missing something here.

Do we have an action creator that triggers an error action? @rgbkrk

jdetle: I dont think we do, you can search for other instances of the same pattern elsewhere in our codebase. I'd be happy to write up an error action creator and refactor all the instances of code that dispatch an error like this.

peggyrayzis: No need to refactor, I totally wasn't aware of the error handling middleware. Sorry about that! Since the middleware picks up uppercase 'ERROR', I think all you need to do here is change your action type to ERROR_PDF_SAVED and we should be good to go. 👍

jdetle:
@peggyrayzis, made that change you requested and squashed the commit :)

peggyrayzis (approved):

peggyrayzis:
Awesome! Thanks again for all your work on this feature. :shipit:

rgbkrk:
Woo!

jdetle:
Huzzah! Thanks so much for all the help. If any of you find things you'd like to iterate on and improve please let me know and I'll happily work on it.

PR #1363, opened 2017-01-20
docs(notebooks): javascript notebook on JSON.parse

rgbkrk:
This is a notebook that covers the use of `JSON.parse`, a little bit on how we load notebooks, and an attempt to optimize loading with `JSON.parse` revivers.

### DIFF
--- a/example-notebooks/immutable-revival.ipynb
+++ b/example-notebooks/immutable-revival.ipynb
@@ -1,0 +1,686 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false,
+    "outputExpanded": false
+   },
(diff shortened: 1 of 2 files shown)


lgeiger (approved):
Awesome!

Should we add a menu item for it?

rgbkrk:
If you think it's a good notebook for general demonstration. We should probably include a bit fat JS logo at the top then.

rgbkrk:
Updated with a JS logo, I'll add it into the menu too.

peggyrayzis:
Really cool seeing a notebook used as documentation! 👍 It's super clear and informative.

rgbkrk:
Menu in now, just have to wait on Travis. I'm at least glad I put this notebook together even though it seems I probably won't use a custom reviver for the loading of the notebook.

PR #1367, opened 2017-01-22
chore(webpack): Add lodash plugin

lgeiger:
This will create smaller Lodash builds


jdetle:
Awesome! Where'd you learn about this?

lgeiger:
> Where'd you learn about this?

@rgbkrk mentioned it some time ago.

I'm pretty sure we could also include source maps for easier debugging and do a more optimized build during packaging

jdetle:
I played around with it and unfortunately none of the source-map methods for debugging worked especially well, but hopefully you can find something that works.

PR #1368, opened 2017-01-24
ux: Switched 'sticky' and 'delete' button

arudhir:
Swapped "sticky" and "delete" buttons to reduce likelihood of people hitting "delete" instead of "run"


jdetle:
Welcome @arudhir! 🎉 🌮

PR #1373, opened 2017-01-26
flow(epics): lightly type annotate the saving epic

rgbkrk:
More flow!


rgbkrk:
Amusing that this increases coverage and yet hits 0% of the patch.

jdetle:
Thanks @rgbkrk!

PR #1374, opened 2017-01-26
flow(epics): type annotate kernel-launch epic

rgbkrk:
Even more type annotating. Ended up fixing the kernelspec metadata while I was at it. Hopefully soon I'll be able to simplify and organize these a bit better.


jdetle:
This will be helpful to look at when I'm flow typing in rx-jupyter. Thanks for all the hard work.

PR #1376, opened 2017-01-27
chore(pdf): Remove toolbar display.

### LINKED ISSUE #1371: [PDF] Hide Toolbar on Print
The print stylesheet should set `display: none;` for the toolbar.

screen shot 2017-01-25 at 11 14 26 am

jdetle:
Fixes #1371

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -13,6 +13,9 @@
   .status-bar {
     display: none!important;
   }
+  .cell-toolbar {
+    display: none!important;
+  }
 }
 
 #app {


static/styles/main.css
```diff
@@ -13,6 +13,9 @@
   .status-bar {
     display: none!important;
   }
+  .cell-toolbar {
+    display: none!important;
+  }
```
rgbkrk, line 18: Had to double check this was inside `@media print` first. :wink:

rgbkrk (approved):

static/styles/main.css
jdetle: Thank you sir!

PR #1378, opened 2017-01-28
chore(deps): Remove commutable (original).

rgbkrk:
Now that we have an internal commutable, remove dependence on the original commutable.


rgbkrk:
Next up is going to be me playing with lerna for our app and likely getting it wrong. :wink: Will use commuter and my sample lerna project as examples.

lgeiger (approved):

PR #1379, opened 2017-01-28
chore(lerna): relocate commutable to packages

rgbkrk:
Plan:

* Leave `src/` alone as the base Electron app
* Each time we extract a package we'll put it in `packages/`

## How I want this to end up

Recall our [vision and identity exploration](https://media.githubusercontent.com/media/nteract/logos/master/nteract_logo_cube_book/concepts/nteract_visual_id_exploration.png):

> Base package for people to build own notebook environments or _new_ explorations for interacting with compute and data.

In effect, developers should be able to install any of our packages to build their own notebook, dashboard (poster!). Some of the finer points of this is that it should be _enjoyable_ to work with our packages, including on the Electron app. We've got a ways to go to achieve that goal I think. We can do it!

### Deliverables

* Electron app
* Individual components for notebooks, output areas
* Reducers for the notebook state tree (amongst others)
* Flow types from all these packages!

Initial package organization:

* `src/` - Electron app, wholly distinct from lerna (for now)
* `packages/` - lerna tracked packages

## Unanswered questions in my mind

* ~~How do we export our flow types on transpile so they're available?~~ [`flow-copy-source`](https://medium.com/@ryyppy/shipping-flowtype-definitions-in-npm-packages-c987917efb65#.gvhlo1tqb)
* Should we use regular dependencies or peer dependencies for our base of `lodash`, `immutable`, `react`, and `react-dom`?
* ~~Will everything be fine with electron-builder and our binary dependencies?~~ Yes, our electron app and native dependencies will be encapsulated in the electron app itself.
* Would it be better for our electron app to be a lerna derived package as well?

## References

* [textlint](https://github.com/textlint/textlint-app) - uses electron-builder and relies on lerna
* [New wave modularity](http://www.macwright.org/2016/07/08/lerna-npm-organizations-new-wave-modularity.html)


rgbkrk:
Based on textlint, I'm already moving this around again to make packages go at the top level.

rgbkrk:
I made a tradeoff here to make `commutable` not be transpiled (including keeping the flow annotations). It's expected to be transpiled at time of webpack build (whether for the Electron app or otherwise).

jdetle:
🥇

rgbkrk:
Next up will be exporting all the transforms.

Stay tuned @gnestor :soon:. 😉

lerna.json
```diff
@@ -0,0 +1,7 @@
+{
+  "lerna": "2.0.0-beta.34",
+  "packages": [
+    "packages/*"
+  ],
+  "version": "1.0.0"
```
lgeiger, line 6: Shouldn't we use [independent mode](https://github.com/lerna/lerna#independent-mode---independent) instead?

src/notebook/epics/github-publish.js
```diff
@@ -12,7 +12,7 @@ import {
 import {
   toJS,
   stringifyNotebook,
-} from '../../commutable';
+} from '../../../packages/commutable';
```
lgeiger, line 15: Aren't we supposed to use `@nteract/commutable` here?

But this would mean we have to setup the correct babel config for all of the individual packages and use `lerna bootstrap`.

lgeiger (commented):
I have just a few questions which are probably beyond the scope of this PR.

With the current state we don't use lerna and instead rely on webpack to include the packages, right?

src/notebook/epics/github-publish.js
rgbkrk: Yeah I kind of skirted this for now.

lerna.json
rgbkrk: I suppose so?

cabhishek: @rgbkrk [commuter](https://github.com/nteract/commuter/blob/master/lerna.json) is using independent mode

src/notebook/epics/github-publish.js
rgbkrk: The part I was unsure of here was whether the Electron app itself would have to be a lerna package in order to use this namespacing or if we need to do a massive multi-package transpile before the webpack build.

lgeiger: Now, I get what you're doing here, thanks for the explaination!

We don't use lerna in the nteract electron application since we use webpack. Instead we'll use lerna only for publishing the packages. That makes sense!

rgbkrk: That's right! I also really have no idea what I'm doing so we should keep feeling this out in case we should do it a different way.

PR #1384, opened 2017-01-30
[WIP] feat: detect the default python setup

rgbkrk:
I don't think I'll be able to finish this the way I really want to, which involves refactoring how we do our initial kernel detection.

Changes I wish I could make:

* [ ] Rely on `jupyter` if the user has it installed, to pick up the default kernel. @minrk already made changes in `jupyter-paths` to do this, we never propagated it up through `kernelspecs`.
* [ ] Only inject the default kernel as is done here if not detected by one of the others (assume an overwrite in the `reduce` on the observable stream
* [ ] Organize our kernel / language menus with a separate subsection for conda environments
* [ ] Type annotate all of this, in hopes of simplifying how our kernelspec structures are laid out

As it stands, this _does_ detect if the user has `ipykernel` installed in the default `python`. I was certainly able to use my local Python.

### DIFF
--- a/src/main/kernels.js
+++ b/src/main/kernels.js
@@ -1,23 +1,56 @@
+/* @flow */
+
 import Rx from 'rxjs/Rx';
 
 import {
   spawn,
+  findActualExecutable,
 } from 'spawn-rx';
 
 const path = require('path');
+
+const kernelspecs = require('kernelspecs');
+
+type Environment = {
+  prefix: string,
+};
+
+function spawnIPyKernelObservable(executable = 'python') {
+  return spawn(executable, ['-m', 'ipykernel', '--version'], { split: true })
+    .filter(x => x.source && x.source === 'stdout');
+}
 
 /**
  * ipyKernelTryObservable checks for the existence of ipykernel in the environment.
  * @param  {Object} env - Current environment
  * @returns {Observable}  Source environment
  */
-export function ipyKernelTryObservable(env) {
+export function ipyKernelTryObservable(env: Environment) {
   const executable = path.join(env.prefix, 'bin', 'python');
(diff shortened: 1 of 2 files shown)


rgbkrk:
Closing, someone else can pick this up. I'm going to focus on some peskier bits as well as exporting more packages for use in the web version.

src/main/kernels.js
```diff
@@ -1,24 +1,57 @@
...
+    .filter(x => x.source && x.source === 'stdout');
+}
+
 /**
  * ipyKernelTryObservable checks for the existence of ipykernel in the environment.
  * @param  {Object} env - Current environment
  * @returns {Observable}  Source environment
  */
-export function ipyKernelTryObservable(env) {
+export function ipyKernelTryObservable(env: Environment) {
```
rgbkrk, line 28: Note: this function here is not used, and is a carry over from the conda kernel stuff.

src/main/kernels.js
```diff
@@ -53,22 +86,23 @@ export function condaEnvsObservable(condaInfo$) {
   * @param {Observable} envs - Environmental elements
   * @returns {Object}   Dictionary containing supported langauges paths.
   */
-export function createKernelSpecsFromEnvs(envs) {
+export function createKernelSpecsFromEnvs(envs: Object) {
   const displayPrefix = 'Python'; // Or R
   const languageKey = 'py'; // or r
 
   const languageExe = 'bin/python';
```
rgbkrk, line 93: Note that the `bin` here is for the conda environment (admittedly, this should be

```js
path.join('bin', 'python')
```

rgbkrk:
Looking back at this with @lgeiger, this would be much simpler and we could finish it off if we only do the default python here and _not_ worry about all conda environments.

PR #1386, opened 2017-01-30
transform flow into comments rather than stripping

rgbkrk:
To make our flow efforts more useful when consuming packages from the lerna package explosion, I've switched our configuration over to [transforming](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-flow-comments) to [flow comments](https://flowtype.org/blog/2015/02/20/Flow-Comments.html) rather than stripping them.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,7 +6,7 @@
     "plugins": [
       "lodash",
       "transform-class-properties",
-      "transform-flow-strip-types",
+      "babel-plugin-transform-flow-comments",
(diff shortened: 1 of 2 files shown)


rgbkrk:
Closing in favor of using `flow-copy-source`, see https://medium.com/@ryyppy/shipping-flowtype-definitions-in-npm-packages-c987917efb65#.gvhlo1tqb.

PR #1388, opened 2017-01-31
chore(docs): use aliased slack-in URL

rgbkrk:
To make this easier on myself, I'm just going with a now alias for our slack in URL: https://slackin-nteract.now.sh

Please update other projects and correspondence to use https://slackin-nteract.now.sh as the canonical URL for our slack inviter!


jdetle:
Thank you sir!

PR #1390, opened 2017-02-01
fix(output): rely on commutable's create output

### LINKED ISSUE #1387: plotly example notebook broken on master?
For me, with the latest master on win64/py35, The `plotize` command in the example notebook doesn't display the chart but also doesn't appear to throw any errors. The same notebook works fine for me with the last released nteract version.

Can anyone else see/repro this? It's highly likely there may have been some issue with my build, but in case its an actual bug I thought I'd report.

image

rgbkrk:
Fixes #1387 by keeping the output structures consistent between kernel runtime and notebook load.

### DIFF
--- a/packages/commutable/v4.js
+++ b/packages/commutable/v4.js
@@ -178,22 +178,27 @@ export function createImmutableMimeBundle(mimeBundle: MimeBundle): ImmutableMime
       .reduce(cleanMimeAtKey.bind(null, mimeBundle), Immutable.Map());
 }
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
~~More to do here to finish this up.~~ All set!

rgbkrk:
Ready for review here.

dhirschfeld:
I can confirm that plotly works again on this branch!

peggyrayzis:
Thanks for fixing this! Looks great 👍

PR #1399, opened 2017-02-03
refactor(provider): Removed Redux logic from Editor, made Codemirror a HOC

peggyrayzis:
When finished, this will resolve the Editor piece of #1288.
- Moved all Redux logic into the Editor provider
- Moved all Codemirror logic into a HOC. This will allow for greater customization as we move to standalone components. With this PR, users can choose to use all or part of our Codemirror implementation by wrapping an Editor view in the HOC and passing any custom options as the second argument of the function. They can also choose to bypass our Codemirror implementation by rendering their own Codemirror or Editor component as a child of the Editor view.

The chain of events can get a little confusing if you've never worked with a HOC before. To sum it up:

**EditorProvider** (injects props from Redux store & callbacks into WrappedEditor)
⬇️ 
**WrappedEditor** (injects Codemirror functionality into an EditorView)
⬇️ 
**EditorView** (dumb component, just renders its children)

I'm going to work on the tests today. Once that's complete, I'll remove the WIP flag.


rgbkrk:
I'm loving how much you're splitting this here and cleaning up in the process.

rgbkrk:
Woo, this puts us back in the 90%+ range!

peggyrayzis:
I'm going to add a couple more new tests to boost us even higher 🚀

peggyrayzis:
Is eslint taking forever to run locally for you guys? It hangs for 3+ minutes before I have to kill it.

rgbkrk:
It's taking forever for me locally too. I think it's a paired problem of eslint enabled for everywhere (not just src) and it is checking eslint within the node_modules of `packages/`.

rgbkrk:
From chat, Peggy is adding 

```
packages/**/node_modules
coverage
```

to our `.eslintignore`

src/notebook/components/cell/codemirror/excludedIntelliSenseKeys.js
```diff
@@ -0,0 +1,54 @@
...
+  '186': 'semicolon',
+  '187': 'equalsign',
+  '188': 'comma',
+  '189': 'dash',
+  '192': 'graveaccent',
+  '220': 'backslash',
+  '222': 'quote',
+};
+
+export default excludedIntelliSenseTriggerKeys;
```
rgbkrk, line 53: I'm happy this got moved out.

src/notebook/components/cell/codemirror/index.js
```diff
@@ -0,0 +1,206 @@
...
+        .switchMap(i => Rx.Observable.of(i))
+        .subscribe(({ editor, ev }) => {
+          const cursor = editor.getDoc().getCursor();
+          const token = editor.getTokenAt(cursor);
+
+          if (!editor.state.completionActive &&
+              !excludedIntelliSenseTriggerKeys[(ev.keyCode || ev.which).toString()] &&
+              (token.type === 'tag' || token.type === 'variable' || token.string === ' ' ||
+               token.string === '<' || token.string === '/') && executionState === 'idle') {
+            editor.execCommand('autocomplete', { completeSingle: false });
```
rgbkrk, line 97: This portion is one of our remaining untested parts, I should probably extract these to functions afterward to test. It would involve a lot of mocking I think though...

src/notebook/components/cell/codemirror/index.js
```diff
@@ -0,0 +1,206 @@
...
+          }
+        });
+    }
+
+    componentDidUpdate(prevProps: WrapperProps): void {
+      const cm = this.codemirror.getCodeMirror();
+      const { cursorBlinkRate, editorFocused, theme } = this.props;
+
+      if (prevProps.theme !== theme) {
+        cm.refresh();
```
rgbkrk, line 107: Probably a good one for us to test that is currently not covered.

src/notebook/components/cell/codemirror/index.js
```diff
@@ -0,0 +1,206 @@
...
+        },
+        extraKeys: {
+          'Ctrl-Space': 'autocomplete',
+          Tab: editor => editor.execCommand('insertSoftTab'),
+          Up: this.goLineUpOrEmit,
+          Down: this.goLineDownOrEmit,
+          'Cmd-/': 'toggleComment',
+          'Ctrl-/': 'toggleComment'
+        },
+        indentUnit: 4,
```
rgbkrk, line 158: Wow it's great to see that this is where it comes from. I've been annoyed with the 4 spaces when I'm working in a Javascript notebook. We should probably case this off based on language (or see if CodeMirror provides defaults per language).

src/notebook/components/cell/codemirror/index.js
```diff
@@ -0,0 +1,206 @@
...
+      const cursor = editor.getCursor();
+      const lastLineNumber = editor.lastLine();
+      const lastLine = editor.getLine(lastLineNumber);
+      if (cursor.line === lastLineNumber &&
+          cursor.ch === lastLine.length &&
+          !editor.somethingSelected()) {
+        CM.signal(editor, 'bottomBoundary');
+      } else {
+        editor.execCommand('goLineDown');
+      }
```
rgbkrk, line 174: Here's an untested section we'll want to extract (possibly).

rgbkrk (approved):
Thank you so much for this super solid refactor! We should probably take some time to figure out what we should do about the thorny cases - codemirror config, the channels used for autocomplete (that require the CodeMirror object), etc.

src/notebook/components/cell/codemirror/index.js
peggyrayzis: Agreed, I'm going to make an issue to write the remaining Editor tests.

src/notebook/components/cell/codemirror/index.js
peggyrayzis: I think the defaults per language comes from the smartIndent property, which defaults to true. Not sure if setting indentUnit prevents smartIndent from working properly, but I'm down to take it out.

rgbkrk: At least for Python we need to ensure it's at 4 spaces for PEP8 compliance within notebooks (our primary user base). At the same time, I'm largely more and more of a JS user within the notebooks, so we should have something sensible for those.

Issue #1177, opened 2016-11-12
Determine retry strategy for our epics


rgbkrk:
We're missing proper retrying on the observables in our epics - they currently catch errors yet don't restart the epics themselves (to keep handling the actions).

```js
export function someEpic(action$, store)
  action$.ofType(...)
    .catch() // 
```

rgbkrk:
Yay, I can close this. We still need to refine the UX. I can say we're good to go here with a sensible start though!

PR #1401, opened 2017-02-03
fix(epics): pass source observable through catch

rgbkrk:
Wow I'm about to fix a pile of our "strange" bugs... The originating observable is available to the callback used with the [`catch` operator](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-catch). Returning it will "restart" the observable.

Ref #1177.

I feel a bit sheepish in admitting I didn't notice this before, even with people commenting about it in `redux-observable`. Too much scanning from me.

### DIFF
--- a/src/notebook/epics/kernel-launch.js
+++ b/src/notebook/epics/kernel-launch.js
@@ -188,8 +188,11 @@ export const newKernelEpic = (action$: ActionsObservable) =>
     .mergeMap(action =>
       newKernelObservable(action.kernelSpec, action.cwd)
     )
-    .catch(error => Rx.Observable.of({
-      type: ERROR_KERNEL_LAUNCH_FAILED,
-      payload: error,
-      error: true,
-    }));
+    .catch((error, source) => Rx.Observable.merge(
+        Rx.Observable.of({
+          type: ERROR_KERNEL_LAUNCH_FAILED,
+          payload: error,
+          error: true,
+        }),
+        source,
+    ));
--- a/src/notebook/components/transforms/geojson.js
+++ b/src/notebook/components/transforms/geojson.js
@@ -40,12 +40,6 @@ export class GeoJSONTransform extends React.Component {
   static MIMETYPE = MIMETYPE;
 
   componentDidMount(): void {
-    // HACK: Work around for testing Leaflet in JSDOM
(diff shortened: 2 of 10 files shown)


src/notebook/epics/kernel-launch.js
```diff
@@ -188,8 +188,11 @@ export const newKernelEpic = (action$: ActionsObservable) =>
...
-      payload: error,
-      error: true,
-    }));
+    .catch((error, source) => Rx.Observable.merge(
+        Rx.Observable.of({
+          type: ERROR_KERNEL_LAUNCH_FAILED,
+          payload: error,
+          error: true,
+        }),
+        source,
```
jdetle, line 197: TIL. 👍

jdetle (approved):
Looks like coverage is the only issue, ready for merge?

lgeiger:
@rgbkrk Are the restarts limited?

It looks like it retries infinitely: https://travis-ci.org/nteract/nteract/jobs/198150130#L2953

rgbkrk:
@jdetle The name "coverage" in our test suite is also unit tests. If that fails, it's definitely failing. Make sure to look at the traceback in the runs themselves, we don't want to merge anything broken on master. When coverage is decreased we get a different message here on GitHub.

Looking at the results, @lgeiger is right. It retries infinitely (as we want it to for these). I'll need to adapt the tests to handle this.

rgbkrk:
> increase coverage by -0.09%

Ha.

rgbkrk:
I've confirmed that while a kernel is booting up you can _try_ to send when it's not ready and it will let you send after. We at least have something recoverable, even if not very informative.

rgbkrk:
Made this even better by wrapping all the epics with a default catch/retry mechanism.

lgeiger (approved):
Nice 🎉

rgbkrk:
Woohoo for stability!

From chat:

```
@lgeiger: One thing I’m unsure is: what happens if a epic keeps failing after the retry,
          will we just retry and retry again?

@rgbkrk:  Yup. It's not the epic itself that is retrying. We basically start listening
          for redux actions again from a clean slate.

@lgeiger: Oh I see, that makes totally sense!

@rgbkrk:  Before we were just letting the entire app die and then users had to restart
```

PR #1403, opened 2017-02-04
feature(toolbar): Add text complementing toolbar

jdetle:
This is a very small fix after seeing an improvement from some changes I think @peggyrayzis made in her cell-creator refactor. On hover, text now appears explaining the actions corresponding to each icon in the toolbar.

### DIFF
--- a/src/notebook/components/cell/toolbar.js
+++ b/src/notebook/components/cell/toolbar.js
@@ -95,7 +95,7 @@ export default class Toolbar extends React.PureComponent {
           <span>
             <button
               onClick={this.executeCell}
-              title= "execute cell"
+              title="execute cell"
               className="executeButton"
             >
               <span className="octicon octicon-triangle-right" />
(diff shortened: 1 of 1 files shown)


src/notebook/components/cell/toolbar.js
```diff
@@ -93,19 +93,31 @@ export default class Toolbar extends React.PureComponent {
         <div className="cell-toolbar">
           {showPlay &&
           <span>
-            <button onClick={this.executeCell} className="executeButton" >
+            <button
+              onClick={this.executeCell}
+              title= "execute cell"
```
lgeiger: Looks like eslint isn't happy with the space behind `=`.

jdetle: Yeah I couldn't run the linter because a GeoJSON test is failing on my machine :\

jdetle: Trying npm run reinstall right now

lgeiger: GeoJSON tests should be fixed by #1401.

You can also run `npm run test:lint` if you only want linting or try https://atom.io/packages/linter-eslint

jdetle:
Thanks for the catch @lgeiger. Made the fix.

lgeiger (approved):

Issue #1360, opened 2017-01-20
Convert to Markdown Cell should put Markdown cell in edit mode


rgbkrk:
When I start typing in a code cell then realize I want a markdown cell, I tend to

* Wish I was in a markdown cell
* Move the mouse to the cell dropdown menu
* Click "Convert to Markdown Cell"
* Try to get back in the flow of typing
* Realize I need to double click to get in edit mode

I wish the transition to a markdown cell was faster. While keyboard shortcuts would help, I'm likely to keep using our little cell menu to switch it over.

rgbkrk:
This was partially closed #1409. I realize now that we may want to switch up how we do cell components. The `<Editor>` component for a cell should remain the same when switching between markdown and code cells.

rgbkrk:
This was fixed after our refactor of the core cell!

PR #1406, opened 2017-02-04
feat(cell): keep cell in edit mode + cell type shortcut

jkornblum:
EDIT - This PR is now ready to be reviewed and merged.

I've removed all the keyboard shortcut functionality (see conversation below).
This commit will still place the markdown cell in edit mode from code cell and vice versa.

----------------
This commit addresses issue #1360
When a code cell is converted to a markdown cell (and there is content in the cell editor) the editor will be refocused and kept in edit mode.

The commit adds keyboard shortcuts for switching between code and markdown cells.
- When in a code cell "ALT+m" will switch to markdown.
- When in a markdown cell "ALT+c" will switch to a code cell.

Looking for feedback on where I should document the shortcut.

### DIFF
--- a/src/notebook/components/cell/code-cell.js
+++ b/src/notebook/components/cell/code-cell.js
@@ -29,7 +29,10 @@ type Props = {
 };
 
 class CodeCell extends React.PureComponent {
-  props: Props;
+
+  constructor(props: Props): void {
+    super(props);
+  }
 
   static defaultProps = {
     pagers: new ImmutableList(),
--- a/src/notebook/components/cell/markdown-cell.js
+++ b/src/notebook/components/cell/markdown-cell.js
@@ -74,6 +74,9 @@ export default class MarkdownCell extends React.PureComponent {
   updateFocus(): void {
     if (this.state && this.state.view && this.props.cellFocused) {
       this.rendered.focus();
+      if (this.props.editorFocused) {
+        this.openEditor();
+      }
     }
   }
 


src/notebook/components/cell/code-cell.js
```diff
@@ -29,7 +33,18 @@ type Props = {
 };
 
 class CodeCell extends React.PureComponent {
-  props: Props;
```
jdetle, line 32: was this intentional?

jdetle:
🤔  I would go for making the title of the "Convert to markdown/code cell"  in the dropdown menu be the text explaining the hotkey. If a user hovers on this while using the notebook, they would find the hotkey. That, or we could put it in parentheses in the same dropdown menu. This is definitely something I'd love input on from our other developers. 

-- edit, more thinking
For a while we have been fairly cautious about adding keybindings and config to nteract. Key concerns (pun intended) are to pick key combinations that are available to every system, don't conflict with other muscle-memory actions, align with hotkeys in Jupyter/RStudio (if they have them for the action in question), and to ensure its an action that is commonplace enough to warrant a hotkey. 

I think this issue is certainly something warranting a hotkey. #1360 mentioned the desire for a shortcut, but I think there is also an opportunity to enter the markdown editing upon changing the cell as well. My understanding is that this PR doesn't address this just yet, would you be interested in tackling it?

src/notebook/components/cell/code-cell.js
jkornblum: If you created (I added this) a 

```
constructor(props: Props): void {
    super(props);
    ...
  }
```

isn't that redundant with that type definition at the top of class?

jdetle: Ahh yes, didn't see that, sorry.

jdetle:
@jkornblum, I was thinking something similar while using the notebook today -- On a down arrow keypress event we create a new cell, I would like this cell to have the same type as the cell that was previously focused. What do you think of that as an idea?

jkornblum:
@jdetle I think that behavior would be the more intuitive than always creating a new code cell. I think Jupyter Notebook will always create a code cell by default (pressing + button on menu) and I know that this always bugs me because I tend to work using a couple markdown cells and then a few code cells in a row.

I can add (Alt+m) and (Alt+c) text to the dropdown menu for shortcut hint on this branch  before it is merged.

I will create another issue regarding creating cells of the same type.

jdetle:
@jkornblum note: edited my comment above to provide more thoughts.

jdetle:
@jkornblum, checked out your PR and ran into an unfortunate hotkey issue. Alt+m is a Mac hotkey to create the letter mu while in nteract. I think we may need to find another option.

jkornblum:
@jdetle (at above comment) I can leave this open for a little bit to get some input from group and then decide on hotkey if any.

Can you explain this a little clearer? What do you mean by enter markdown editing upon cell?

> #1360 mentioned the desire for a shortcut, but I think there is also an opportunity to enter the markdown editing upon cell as well.

Regarding staying in alignment with Jupyter and RStudio, I believe the only shortcut in Jupyter Notebook to toggle cell type is 'M'/'Y' when in command mode (press ESC) although I'll have to verify; nteract does not have such command mode.
Maybe we could enter a command mode like Jupyter when Document.editorFocused is null? 

I do not really use RStudio or RMarkdown so maybe someone else can comment on the thread?

jdetle:
Sorry I just edited it. I meant to say that we shouldn't have to double click the cell after changing its type with or without a hotkey. If we are editing the code cell, and switch to markdown, it should simply change to editing that text in markdown mode instead. Does that make more sense?

jdetle:
Oops, just checked it out and your work seems to address that issue perfectly. Thank you.

jarmokivekas:
In the Jupyter command mode, cells are added by pressing `A` (add above) or  `B` (below), you switch focus between cells with `J` and `K` (like in vim). The cell type is changed with `M` (markdown) and `Y` (code). Edit mode is entered with `enter` and command mode with `Esc`. Personally I have those quite engrained in memory by now, and regularly find myself trying to use them in nteract to no effect.

I would love to see those in nteract, not only because I'm used to them, but I genuinely think it's a good may to interact with the notebook.

Keyboard shorcuts are tricky, though, especially with multiple platforms. As a linux user I would by convention expect alt+c to open the *Cell* menu and alt+f to open the *File* menu and so on.

jarmokivekas:
I think something in this breaks `ctrl+enter` for running/exiting edit mode in markdown cells. It seems to work on 5cacb0fd7 (HEAD~1)

jkornblum:
Thanks @jarmokivekas.

I'm going to just close this PR and open a new cleaner one just addressing placing markdown back into edit mode. I'll comment on #1360 about how to handle keyboard shortcuts.

Issue #1404, opened 2017-02-04
border around hovered cell in exported PDF


jarmokivekas:
A border line is drawn in the PDF around the cell that the cursor is hovering over when exporting a pdf using *File > Export PDF*.

![screenshot 2017-02-04 23 01 22](https://cloud.githubusercontent.com/assets/3074453/22621533/9ddaabd4-eb2e-11e6-9f8a-12eb706fe1f2.png)

Clicking the menu item leaves the cursor hovering above the topmost visible cell. The border around the active cell is removed as expected, though.

![screenshot 2017-02-04 23 02 10](https://cloud.githubusercontent.com/assets/3074453/22621534/9f0d8eb8-eb2e-11e6-9783-a63b0c19b76d.png)

I wouldn't mind fixing this one myself. Probably something to do with `@media print` in [/static/style/main.css](https://github.com/nteract/nteract/blob/master/static/styles/main.css)?

PS: I think it's awesome that you can export pdfs where the cell input is hidden for plotting code. It looks nice :sparkles:

jdetle:
I can tackle this now or if you'd like to make the change it would be very appreciated. Thanks for the catch, I really appreciate you trying it out! Brings a tear of joy to this contributors eye :)

jarmokivekas:
I think I got it fixed a few seconds ago. I noticed that also the `cell-creator` boxes between cells were visible sometimes.

jdetle:
Fantastic! I will definitely merge any pull request I see addressing these issues. Thank you so much @jarmokivekas.

rgbkrk:
Closed by #1407.

Issue #1400, opened 2017-02-03
nightly .deb package archive


jarmokivekas:
I'm in the process of setting up a debian package archive that allows installing nteract with `apt-get install` on linux distros such as Ubuntu. Also automatic updates.

- [x] I'd like to make nightly packages from the master branch on git available. Does nteract have a culture of merging mostly working code to master (no point in nightly builds if not)? From personal experience, I think yes.

My current thinking:

- `git pull` latest master branch
- `git checkout -b new-package-branch`
- [x] ~~~`npm version patch`? The version number needs to change, but the git tags could be thrown away. This would mean that actual named releases here on github would need to be for example `npm version minor` instead of `patch` like now.~~~  Actually appending the date as -yyyymmdd to whatever the current version number is probably a better solution, since it will be correct semantic versioning, which is required by the packaging system. It would also help with tying the version number back to the version in git. 
- [x] `npm run dist` to create the .deb package
- [ ] copy the .deb to the archive. Debian packages archives are just any ftp/http hosts serving static files. I haven't quite figured out the best way to host all the files.
- [x] update archive metadata and sign it.
- [x] packages are available via regular system updates

---

There is launchpad.net, which is commonly used to host personal package archives, but as far as I'm aware the `electron-builder` process in its current is not compatible with the build process on launchpad. 

I'm trying to figure out if there is a way to use the GitHub release mechanism to host the `.deb`s for the archive (not for nightly builds, but proper builds maybe). I'm pretty sure the answer is yes. If github pages had http redirects you could even host the metadata and have a repo-contained package archive. That could be fun.

---

I've been playing around with this stuff at https://github.com/jarmokivekas/nteract-apt-archive

jarmokivekas:
I got a new update to  [GitKraken](https://www.gitkraken.com) today, and that process was quite nice as well, as far as self-updating applications on linux go.

On Linux their process is something like:

1. initial installation of the .deb
2. check for updates on application startup
3. notify user if there are updates available
4. the notification has a "download .deb" button, clicking it downloads a new .deb
5. when the download is complete, there is a "show in file manager" button.
6. double click the .deb to open the system package installer and replace the current install
7. back to 1.

This flow would probably be fairly straight-forward to implement using github releases mechanism to host the .deb, and I assume github has an API to check for new releases, since they seem to have most things. :thinking:

jarmokivekas:
[...]

PR #1407, opened 2017-02-04
fix(pdf): hide undesired UI elements in pdf print

jarmokivekas:
Fix to #1404 and some other related things I noticed.

- [x] hide notifications
- [x] hide cell-creator boxes
- [x] hide border around hovered cells. **wontfix/out-of-scoped:** This issue was a little deeper that some `@media print` css rules.

Is there a way to force the `@media` type in *View > Toggle Developer Tools*? Could be convenient.

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -10,12 +10,19 @@
   * {
     box-shadow: none!important;
   }
-  .status-bar {
+  .status-bar{
     display: none!important;
   }
-  .cell-toolbar {
+  .notifications-wrapper {
     display: none!important;
   }
+  .cell-toolbar{
+    display: none!important;
+  }
+  .cell-creator{
+    display: none!important;
+  }
+
 }
 
 #app {


lgeiger:
That's great 🎉 

> Is there a way to force the @media type in View > Toggle Developer Tools?

Yes this is possible: http://stackoverflow.com/questions/9540990/using-chromes-element-inspector-in-print-preview-mode?answertab=votes#tab-top

jarmokivekas:
Interestigly, the media type emulation does not show the border, while it is still in the pdf anyway. Media emulation hides notifications and cell-creators just fine though.

jdetle:
One potential fix for that issue is to unfocus the cell in question in the exportPDF function. That may be a somewhat hacky fix but I'd be okay with it personally.

jdetle:
@jarmokivekas Another thought is to hide scrollbars on outputs. The focus problem seems to be something that isn't as easily fixed. I'd be for hiding the scrollbars, merging this PR, and fixing that problem in a separate PR.

jarmokivekas:
I think the offending border is from a div like this:

```
<div class="cell-draggable" "style="opacity: 1; border-top: 3px solid transparent; border-bottom: 3px solid transparent;">
```

Editing the CSS doesn't seem to help, since the border is defined in the html, therefore overriding anything in `main.css`?

There is a workaround: "quickly move the cursor outside of the nteract window" :smile:

jdetle:
Oh man yeah I totally misunderstood the problem. If you move the style into `main.css` and get ride of the inline style, you could probably fix the problem, in addition, you could also override with `!important` as well. If you could get rid of the inline styling that's probably the more maintainable fix.

jarmokivekas:
Well, the problem wasn't actually the inline style, but a css transition that eases in the border on hover.

Seems if the transition is *in progress* when `exportPDF` calls `printToPDF` the cell's border will show up in the pdf, so sometimes it shows, sometimes not. It's a weird edge-case. Removing the transition altogether fixes the problem in the pdf a, but removing it in `@media print` is not enough.

static/styles/main.css
```diff
@@ -127,7 +137,7 @@ blockquote p {
 .cell {
     position: relative;
     background: var(--cell-bg);
-    transition: all .1s ease-in-out;
+    /*transition: all .1s ease-in-out; /* causes issue #1404 */
```
jarmokivekas: I don't love this solution, but I tried several other solutions, and can't find another other way.

jarmokivekas:
https://drive.google.com/file/d/0B4CRJznKbv1ATmh6d1owWGpBcnc/view?usp=sharing

The the box-shadow's ease-in transition is super subtle though. Old version to the left, and the new version to the right.

static/styles/main.css
rgbkrk: Hmmmm, I definitely don't want to get rid of our subtle animation cues within the notebook.

http://stackoverflow.com/questions/11131875/what-is-the-cleanest-way-to-disable-css-transition-effects-temporarily leads me to believe we have to do a bit more work to straddle this print CSS case.

Can we go with only the print media for now and work on a longer term solution to the in-between-states?

lgeiger: I agree with @rgbkrk.

I tried, but can't reproduce the problem on macOS. Could this be a Chrome related bug only happening on linux?

If so we could try out [Electron 1.5.0 beta](https://github.com/electron/electron/releases/tag/v1.5.0)  which is based on a newer version of Chrome.

rgbkrk:
This makes me wonder if there is a way to reflow the CSS prior to the printing.

rgbkrk (approved):

rgbkrk:
Thanks @jarmokivekas!

jdetle:
Thank you @jarmokivekas, is this your first merged pull request with
nteract?

static/styles/main.css
jarmokivekas: In macOS the menu bar is at the top of the screen, right? To trigger this, when you click the Export PDF menuitem, your cursor needs to end up hovering above a cell that is not focused. Even then there are times where it isn't triggered, if there is no mouse event happening before the PDF is exported, or the 0.1s transition is already completed vy the time the pdf is exported. At least that's it seems to me. It is kind of an edge case...

rgbkrk:
@jdetle - Jarmo [made nteract much better for Debian based Linux distros in December](https://github.com/nteract/nteract/commits?author=jarmokivekas).

jdetle:
Awesome!!! Thank you so much Jarmo and I'm very appreciative of you fixing
these problems. A very belated welcome!

jarmokivekas:
@jdetle It's my first pull request that touches the actual codebase. Earlier I did some stuff with package.json to make .deb archives for intalling nteract on linux. I've actually configured a way to autoupdate my nteract via the system updater on Ubuntu over this weekend ( issue #1400  << I have this actually working already on some level)

Also I should give kudos to the nteract project as a whole. As far as open source software communities go, I think this one does a lot of things right. People are nice, the feedback is quick, there is an active slack channel, documentation... I don't know, I've had many much worse experiences. :1st_place_medal:

jdetle:
That's very kind of you to say, and warmly appreciated.

PR #1409, opened 2017-02-05
feat(markdown-cell):  keep markdown cell in editing mode

jkornblum:
Addresses part of  #1360. This commit will keep the cell focused and in editing mode when changing from code cell type to markdown cell type. Previously the cell would come out of editing mode.

This is a cleaned up redo of PR #1406.


rgbkrk (approved):

rgbkrk:
Thanks!

PR #1411, opened 2017-02-06
chore(lint): Fix slow linter

peggyrayzis:
Eslint has been really slow locally! 🐢 This PR fixes that by changing the linting script to only lint src & test.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,3 +7,5 @@ flow-typed/
 types/
 build/
 example-notebooks/
+packages/**/node_modules
+coverage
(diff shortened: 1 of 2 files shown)


rgbkrk:
Mega 👍

rgbkrk:
With this eslintignore we no longer have to set the `npm run lint` to be specific here. I'm going to go ahead and merge this as is though!

peggyrayzis:
I actually tested setting `npm run lint` to `eslint .` as it was before, but it was still painfully slow for me. Maybe I'm just being impatient? 😆

PR #1420, opened 2017-02-07
Extract transforms for lerna

rgbkrk:
Extract our transforms from media type payloads ➡️  React Elements.


ivanov:
rebased, for starters.

packages/transforms-full/package.json
```diff
@@ -26,5 +26,12 @@
   },
   "devDependencies": {
     "react": "^15.4.2"
+  },
+  "dependencies": {
+    "@nteract/transform-geojson": "1.0.0",
+    "@nteract/transform-model-debug": "1.0.0",
+    "@nteract/transform-plotly": "1.0.0",
+    "@nteract/transform-vega": "1.0.0",
+    "@nteract/transforms":  "1.0.0"
```
rgbkrk, line 35: ❤️  this is definitely moving along

ivanov:
Tests pass for me locally, let's see what our CI pals have to say about it.

ivanov:
Oh, I think it's because I ran lerna locally, which CI doesn't yet know about. Let me add that.

ivanov:
I think the flow errors we're seeing are similar to some of the issues reported on this thread: https://github.com/facebook/immutable-js/issues/203 but I have not followed it all to figure out how to get around it.

ivanov:
I don't really know how flow works, but might it have something to do with the fact that we have 3 copies / version of immutable.js? 

```
$ find . -name immutable.js.flow
./node_modules/immutable/dist/immutable.js.flow
./packages/commutable/node_modules/immutable/dist/immutable.js.flow
./packages/transforms/node_modules/immutable/dist/immutable.js.flow
```

so the lerna-ed packages reference their own, but the main app references its own, and flow doesn't understand that they're the same?

rgbkrk:
I'm looking at the flow types for Immutable and the `getIn` definition is _very_ wrong:

```
getIn<T>(searchKeyPath: ESIterable<any>, notSetValue: T): T;
getIn<T>(searchKeyPath: ESIterable<any>): T;
```

A deeply nested value may not actually be `T`.

@peggyrayzis what's the best way for us to ignore bad parts of definitions that come as part of the library (not in our `flow-typed` directory)?

rgbkrk:
Then again, this was passing on master.

Have you run into anything like this @thejameskyle or @hzoo? We're trying to adapt to using lerna (amidst shipping an Electron app), running into trouble with the flow definitions for Immutable.

ivanov:
For what it's worth, I get the same flow error on master. :\

ivanov:
Ok, so here's how the errors can be eliminated: `lerna bootstrap --hoist immutable`... - This will not put `immutable` in the packages/*/node_modules, but only in the top level node_modules. 

except I think there's a bug in lerna, because running that with the `--hoist immutable` flag it just hangs there... 

symlinking the root `node_modules/immutable` to the `packages/commutable/node_modules/immutable` also fixes the flow stuff. As does removing the latter after a `lerna bootstrap` by hand.

peggyrayzis:
>@peggyrayzis what's the best way for us to ignore bad parts of definitions that come as part of the library (not in our flow-typed directory)?

It looks like Paul already fixed the issue, but I would put a line in the [ignore] section of `.flowconfig` like `.*/node_modules/immutable/dist/.*`. This is an OCaml regex, so it can can pick up on any subdirectories, like the directories in `packages/`, that also contain this file pattern.

rgbkrk:
> For what it's worth, I get the same flow error on master. :\

Ah, that must be because you and I have both done a `lerna bootstrap`, which means that it's resolving the Immutable in its `node_modules`.

rgbkrk:
> It looks like Paul already fixed the issue, but I would put a line in the [ignore] section of .flowconfig like .*/node_modules/immutable/dist/.*. This is an OCaml regex, so it can can pick up on any subdirectories, like the directories in packages/, that also contain this file pattern.

Tried it, didn't work. 😭

rgbkrk:
I wonder if this is because we listed it as a dev dependency:

https://github.com/lerna/lerna/issues/505

We could switch back to a normal dependency, the reason I didn't want to go that route was so that people could rely on it as a peer dependency.

rgbkrk:
I'm trying some things and will post a new commit when I think it's in a friendly state.

rgbkrk:
It passes! 🔨

package.json
```diff
@@ -31,7 +31,7 @@
     "lint": "eslint src/ test/",
     "lint:fix": "eslint . --fix",
     "prebuild": "rimraf lib",
-    "lerna": "lerna bootstrap --hoist immutable",
+    "lerna": "lerna bootstrap  && rimraf packages/commutable/node_modules/immutable",
```
rgbkrk: This will be a stopgap for us until the next lerna release (which will probably be today).

ivanov: yeap, I opened #1436 so we remember to do it in case lerna release gets held up.

rgbkrk:
Whoops, because of me there's a merge conflict with the geojson transform. I'll handle that now.

rgbkrk:
Rebased, brought in the background color -> GeoJSON tile support.

rgbkrk:
Yay, new lerna release fixes it for us! Thanks all.

peggyrayzis (approved):
Everything looks great to me! I'm excited to start working on exporting packages 🎉

Issue #1414, opened 2017-02-06
Migrate off eslint, switch to Prettier


peggyrayzis:
Amidst all the eslint slowness issues, @rgbkrk suggested that we switch to [Prettier](https://github.com/jlongster/prettier). This is a great idea and will hopefully make our lives easier! 😀 

To achieve this, we need to:

- [ ] Close all open PRs to avoid merge conflicts
- [ ] Remove eslint, its plugins, and config.
- [ ] Add Prettier, ensure everything still works as expected

rgbkrk:
Yay for the `gofmt` of JS!

PR #1421, opened 2017-02-07
Prettier

rgbkrk:
An almost complete version of #1414.


rgbkrk:
Something I noticed while doing repeated iterations of `prettier 'src/**/*.js'` - the output is not always a stable state. I eventually got it to one here.

One thing to be really careful of / look over in the changes here - the comments sometimes get mangled or end up fully deleted. We should comb through to see if they're all right.

rgbkrk:
On a separate note, I thought we would lose semi-colons with prettier. Drat!

P.S. We could totally do standard js instead, hint hint wink wink.

peggyrayzis:
I was hoping the output would be...prettier? 🙈 I don't like how strict the max line length is and I'm not a fan of double quotes or semicolons.

Standard is looking like a better option, which makes me sad because I really wanted to love Prettier.

rgbkrk:
Closing, not moving to prettier (for now). In other news, I've enabled my eslint plugin in Atom to auto-fix.

PR #1424, opened 2017-02-07
fix(codemirror): Selecting + pressing tab no longer deletes the line

### LINKED ISSUE #1422: Tab-indentation broken on master?
On master, highlighting text and pressing tab will delete the text rather than indenting it!
tab

Is anyone else seeing the same? On win64/py35.

peggyrayzis:
Fixes #1422.


jdetle:
Thanks @peggyrayzis!

rgbkrk:
Thank you!

dhirschfeld:
Thanks for this! I've tested on master and confirm that tab now works to correctly indent the selection. I did however notice that shift-tab doesn't work so I opened #1430 to track that.

PR #1426, opened 2017-02-07
feat(geojson): autodetect background to pick tiles

### LINKED ISSUE #1093: determine a strategy for picking tiles
in `componentDidMount` function within `src/notebook/components/transforms/geojson.js`

rgbkrk:
Fixes #1093.

Relies on `window.getComputedStyle` to get the background color then computes the [Luma value](https://en.wikipedia.org/wiki/Luma_(video)).

While I was playing with this I also noticed I really like the `streets` map tileset from Mapbox.

/cc @gnestor

### DIFF
--- a/src/notebook/components/transforms/geojson.js
+++ b/src/notebook/components/transforms/geojson.js
@@ -2,10 +2,7 @@
 /* eslint class-methods-use-this: 0 */
 import React from 'react';
 
-type Props = {
-  data: Object,
-  theme: string,
-};
+type Props = { data: Object, theme: string };
 
 const L = require('leaflet');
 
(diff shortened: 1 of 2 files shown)


gnestor (approved):

gnestor:
I tried this out! Nice! Streets is super nice. So jupyterlab_geojson is currently using OSM. Is it merely a matter of style preference or is there good reason to go with Mapbox?

rgbkrk:
No significant reason. Happy to switch to OSM, especially if that helps for packages we export from here.

gnestor:
Streets theme is nice though...

image

rgbkrk:
I ended up not using the `streets` tileset in this PR because it didn't quite match the style of the notebook, whereas light and dark drop in pretty well. It does look like it would be a much better map to use when doing urban work (note: there are tilesets better suited to topographic analysis too...)

Makes me think we should establish a metadata standard just like we have options for Images and JSON, then people could set the tile pattern directly if they like.

It's also tempting to make this component have some additional options that could be exposed with a menu.

We can wait to see what geo users think too, especially on releases coming this year.

gnestor:
I was thinking along the same lines. Brian and I are thinking through the VegaLite display class as it involves both a `spec` and `data`, so similarly GeoJSON could accept an optional `spec` or `options` for configuring tilelayer, etc.

rgbkrk:
That wouldn't be proper GeoJSON if you embed `spec` and `data` in the JSON structure itself.

I'm talking about the metadata key within a `display_data` message (i.e. `display(data, metadata=dict())`).  Check out the [payloads for metadata within the display_data message docs](http://jupyter-client.readthedocs.io/en/latest/messaging.html#display-data).

gnestor:
What I meant was that GeoJSON could accept additional arguments:

```py
GeoJSON(data, options={"theme": "streets"})
```

Like the existing VegaLite display method:

```py
VegaLite(spec={
  "mark": "point",
  "encoding": {
    "y": {"type": "quantitative","field": "Acceleration"},
    "x": {"type": "quantitative","field": "Horsepower"}
  }
}, df)
```

rgbkrk:
> What I meant was that GeoJSON could accept additional arguments:

Ah I see, yes that's what is done for `JSON`. It ends up in metadata on the protocol (where my brain space is for the frontend, since that's how we'll get it here).

```python
JSON(data, expanded=True)
```

rgbkrk:
At least for GeoJSON, I think I'd want to use:

```
GeoJSON(data, tileUrlTemplate, tileLayerOptions)
```

matching the Leaflet API as much as possible.

¯\\\_(ツ)\_/¯

gnestor:
Sounds good to me! I'm going to chime in on the ipython PR...

Issue #349, opened 2016-04-21
Handle ipynb local links directly


rgbkrk:
Clicking a link that resolves to `./foo/bar.ipynb` should open a new nteract window with `./foo/bar.ipynb`.

estevanjantsk:
Trying to figure out how to get the click event of a <a href=""> tag of the latex component ._.

rgbkrk:
Ah, there may be a neat way of [handling this within launch.js](https://github.com/nteract/nteract/blob/751bb0938252d08f9658d0bd0cf68cda5ba93bd7/src/main/launch.js#L31).

rgbkrk:
Right now our function for launching URLs has the native browser open it instead:

``` js
export function deferURL(event, url) {
  event.preventDefault();
  shell.openExternal(url);
}
```

That can do other logic on the passed in `url` to check if it's a local link that ends in `.ipynb`. No messy handling of the links on a page, letting Chromium/Electron just hand you the URL when ready.

estevanjantsk:
@rgbkrk Nice, that method that I was looking for :dagger: . Gonna try to help with this one. May you assign this issue to me, please? :+1: :+1:  enjoying help you, guys. =D

estevanjantsk:
@rgbkrk Do you know why this link './foo/bar.ipynb' reproduces this url? -> file:/Users/estevanj/Documents/Projects/nteract/static/foo/bar.ipynb .. Why goes to the static folder? 👍 But when i put '../foo/bar.ipynb' the url is ok.

It looks like the current folder when you do the ./ is the static folder, is this right?

The current solution I get is:

```
if (url.endsWith('.ipynb')) {
  url = url.replace('file://', '');
  launch(path.normalize(url));
}else {
  shell.openExternal(url);
}
```

Realize that I have to remove the file:// because when it goes to the win.webContents.send('main:load', filename); with the file:// it'll not load properly.

rgbkrk:
We've definitely got a bit of a problem in terms of how electron views the page - we have to track the user's `cwd` while keeping Electron's relative asset loading intact.

Separate from that issue, there is a helper path to finding the relative path: `__dirname`.

For instance, in the main window `__dirname` is `/Users/kyle/code/src/github.com/nteract/nteract/static` for me.

Another note: If we notice the prefix is not `file:` we must do `openExternal`, even if it ends in `.ipynb`.

rgbkrk:
[...]

PR #1432, opened 2017-02-07
Resolve relative URLs in markdown cells to notebook directory

gnestor:
Follow up from https://github.com/nteract/nteract/issues/349

To do:

* [ ] Use `<base>` html element: https://github.com/nteract/nteract/issues/349#issuecomment-278180910
* [ ] Fix broken paths to CSS/images
  * [ ] Webpack CSS/images

### DIFF
--- a/src/notebook/components/cell/markdown-cell.js
+++ b/src/notebook/components/cell/markdown-cell.js
@@ -2,6 +2,8 @@
 import React from 'react';
 import CommonMark from 'commonmark';
 import MarkdownRenderer from 'commonmark-react-renderer';
+import path from 'path';
+import url from 'url';
 
 import Editor from '../../providers/editor';
 import LatexRenderer from '../latex';
@@ -21,13 +23,6 @@ type State = {
   view: boolean,
   source: string,
 };
-
-type MDRender = (input: string) => string
-
-const parser = new CommonMark.Parser();
-const renderer = new MarkdownRenderer();
-
-const mdRender: MDRender = input => renderer.render(parser.parse(input));
 
 export default class MarkdownCell extends React.PureComponent {
   state: State;
@@ -123,6 +118,23 @@ export default class MarkdownCell extends React.PureComponent {
   }
 
   render(): ?React.Element<any> {
+    const { filename } = this.context.store.getState().metadata.toJS();
+    const cwd = `${path.dirname(filename)}/`;
+    const parser = new CommonMark.Parser();
+    const renderer = new MarkdownRenderer();
+    const parsed = parser.parse(
+      this.state.source ?
+      this.state.source :
+      '*Empty markdown cell, double click me to add content.*'
+    );
+    const walker = parsed.walker();
+    let event;
+    while ((event = walker.next())) { // eslint-disable-line no-cond-assign
+      const { node } = event;
+      if (event.entering && node.type === 'link') {
+        node.destination = url.resolve(cwd, node.destination);
+      }
+    }
     return (
        (this.state && this.state.view) ?
          <div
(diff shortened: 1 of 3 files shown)


src/notebook/components/cell/markdown-cell.js
```diff
@@ -123,6 +118,23 @@ export default class MarkdownCell extends React.PureComponent {
...
+      this.state.source :
+      '*Empty markdown cell, double click me to add content.*'
+    );
+    const walker = parsed.walker();
+    let event;
+    while ((event = walker.next())) {
+      const { node } = event;
+      if (event.entering && node.type === 'link') {
+        node.destination = url.resolve(cwd, node.destination);
+      }
```
rgbkrk: Neat!

src/notebook/components/cell/markdown-cell.js
```diff
@@ -123,6 +118,23 @@ export default class MarkdownCell extends React.PureComponent {
   }
 
   render(): ?React.Element<any> {
+    const { filename } = this.context.store.getState().metadata.toJS();
```
rgbkrk: We shouldn't be relying on `this.context` in these components. Perhaps we could pass down what the `cwd` is?

rgbkrk (changes requested):
This is so good to see the commonmark parser being used really well!

src/notebook/components/cell/markdown-cell.js
rgbkrk: As in from a top level.

gnestor: Ya, I couldn't figure out how to access it otherwise. Your `<base>` idea sounds better if I understand it correctly: it's a component that has access to `cwd` among other things and it passes it down to other components via props.

rgbkrk: [`<base>` is an HTML Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) that specifies the base URL to use for all relative URLs in a document, it's not a React component. With the power of React though, when the filename changes, so would base's `href` and thereby the interpretation of all relative paths.

gnestor: Ohhhh! Ya, that's perfect except that it messes up relative URLs to nteract assets 🤔

rgbkrk: Yeah I was thinking that since we're doing webpack anyways we could make sure the base assets are loaded first. The ones that are gunked up in the little prototype I showed was only the theme css (which can change).

gnestor: I just pushed another commit that breaks the context stuff out into a Base component. I did this yesterday before I realized that `<base>` is an HTML element, so for what it's worth...

rgbkrk: You had me a bit worried when I saw the commit before your comment.

The component you've added `Base` into is supposed to be a disconnected component, in line with #1288.

rgbkrk:
Great to see this solution! This component is supposed to be a presentational component and as such should not have access to context - props are intended to be the only API in accordance with #1288. I'd like to solve this either with the `<base>` (and required loading changes) or by passing the `cwd` down through the entire tree (yes, that would be a lot of boilerplate to pass it through).

The other issue with relying on context for this is that it will not necessarily re-render when the filename changes as opposed to when its a prop.

gnestor:
> This component is supposed to be a presentational component and as such should not have access to context

I get that, but why does MarkdownCell and others define `contextTypes` then? 

> I'd like to solve this either with the <base>

This is ideal.

rgbkrk:
> I get that, but why does MarkdownCell and others define contextTypes then?

The cell components are currently open check boxes on that issue. I'd rather not introduce more to refactor when we have other options for doing this in line with the architecture plan.

gnestor:
Ok I see. Props-only approach is vastly superior 👍

rgbkrk:
From direct messages, I'd love to see this done with `<base>` so that MarkdownCell can be used without knowledge of `cwd`. It should work like any other relative notebook asset.

rgbkrk:
https://github.com/nteract/nteract/issues/349#issuecomment-278180910

rgbkrk:
Closing in favor of #1720.

Issue #1410, opened 2017-02-05
Keep cell type when created with ArrowDown


jkornblum:
When the last notebook cell is in focus and the down arrow is pressed a new cell is created. Currently regardless if the last cell is a markdown type or code type a code type will be created.

Any thoughts on persisting the last cell type for newly created cell types? That is, a down arrow on a markdown cell will create another markdown cell and a down arrow on a code type will create another code type cell.

rgbkrk:
I like that idea.

jkornblum:
This is closed with #1435

PR #1435, opened 2017-02-08
feat(document): persist cell type on down arrow

jkornblum:
This addresses issue #1410.  This commit persists the cell type when it is created with the down arrow.

- When a code cell is the last cell down arrow will create a new code cell
- When a markdown cell is the last cell down arrow will create a new markdown cell

*Note this will persist the cell type with an execute + create new cell (shift+enter). If we want to only persist on down arrow I will refactor this.*


rgbkrk (approved):
Excellent, and it even has tests! Thank you.

jdetle:
Thanks @jkornblum!

PR #1440, opened 2017-02-08
Provide some basic READMEs for our new packages!

### LINKED ISSUE #1437: Write READMEs for each package in `packages/`
Now that we've extracted more components and libraries within `packages`, we need to follow it
up with READMEs for each so that we have documentation on npmjs.com after publishing as well
as when browsing in the repository.

* [x] commutable
* [x] transforms
* [x] transforms
* [x] transforms-full
* [x] transform-geojson
* [x] transform-model-debug
* [x] transform-plotly
* [x] transform-vega

rgbkrk:
📝  READMEs all over!

Fixes #1437

### DIFF
--- a/packages/README.md
+++ b/packages/README.md
@@ -2,5 +2,5 @@
 
 📦
 
-This directory contains all packages exported for use by the nteract desktop app.
-
+This directory contains all packages exported for use by the nteract desktop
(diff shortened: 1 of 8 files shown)


jdetle:
🥇

Issue #897, opened 2016-10-04
Provide better error when kernel doesn't spawn


jdetle:
Error:

``` js
events.js:160 
Uncaught Error: spawn /home/jdetlefs/github/nteract/bin/python3 ENOENT
```

On my machine I would think that I could run 

``` js
const spawnteract = require('spawnteract')

spawnteract.launch('python3').then(kernel => {
     console.log(kernel.config.ip + ':' + kernel.config.shell_port);
});
function cleanup(kernel) {
       kernel.spawn.kill();
       fs.unlink(kernel.connectionFile);
 }
```

But I get this error: 

``` js
127.0.0.1:9002
events.js:160
      throw er; // Unhandled 'error' event
      ^
Error: spawn /home/jdetlefs/github/nteract/bin/python3 ENOENT
    at exports._errnoException (util.js:1036:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
    at onErrorNT (internal/child_process.js:359:16)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

Any idea what's going on?

jdetle:
Contents of my `kernel.json` file

```
  1 {
  2  "language": "python",
  3  "display_name": "Python 3",
  4  "argv": [
  5   "/home/jdetlefs/github/nteract/bin/python3",
  6   "-m",
  7   "ipykernel",
  8   "-f",
  9   "{connection_file}"
 10  ]
 11 }      
```

rgbkrk:
That's a peculiar location for the python3 kernel. Was that installed in a `virtualenv`?

jdetle:
Very possible

rgbkrk:
Does the path to that python3 still exist?

jdetle:
No, but I installed more kernels 

```
 { name: 'python2',
     files: 
      [ '/usr/local/share/jupyter/kernels/python2/kernel.json',
        '/usr/local/share/jupyter/kernels/python2/logo-32x32.png',
        '/usr/local/share/jupyter/kernels/python2/logo-64x64.png' ],
     resources_dir: '/usr/local/share/jupyter/kernels/python2',
     spec: { display_name: 'Python 2', language: 'python', argv: [Object] } },
```

jdetle:
This still doesnt work,

```
  1 {
  2  "display_name": "Python 2",
  3  "language": "python",
[...]

PR #1442, opened 2017-02-09
Kernel error notification

jdetle:
Changed:
- New Kernel Observable does not complete unless there is an error
- Tests finish after both actions are pushed
- Emit an error that is caught by the notification system when spawning fails.

Issue:
When Spawnteract's child process spawning of a kernel fails, it fails asynchronously. `Child_process.spawn` in spawnteract emits an error, on this error we should emit an error action that is caught by the middleware. To test, change the path of a kernel in your kernelspec's kernel.json folder. This should help expose issues for new users.

### DIFF
--- a/src/notebook/epics/kernel-launch.js
+++ b/src/notebook/epics/kernel-launch.js
@@ -83,37 +83,37 @@ export function newKernelObservable(kernelSpec: KernelInfo, cwd: string) {
   const spec = kernelSpec.spec;
 
   return Rx.Observable.create((observer) => {
-      launchSpec(spec, { cwd })
-        .then((c) => {
-          const { config, spawn, connectionFile } = c;
-          const kernelSpecName = kernelSpec.name;
+    launchSpec(spec, { cwd })
+      .then((c) => {
+        const { config, spawn, connectionFile } = c;
+        const kernelSpecName = kernelSpec.name;
 
-          const identity = uuid.v4();
-          // TODO: I'm realizing that we could trigger on when the underlying sockets
-          //       are ready with these subjects to let us know when the kernels
-          //       are *really* ready
-          const channels = {
-            shell: createShellSubject(identity, config),
-            iopub: createIOPubSubject(identity, config),
-            control: createControlSubject(identity, config),
-            stdin: createStdinSubject(identity, config),
-          };
-          observer.next(setNotebookKernelInfo(kernelSpec));
+        const identity = uuid.v4();
+        // TODO: I'm realizing that we could trigger on when the underlying sockets
+        //       are ready with these subjects to let us know when the kernels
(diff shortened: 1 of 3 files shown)


src/notebook/epics/kernel-launch.js
```diff
@@ -108,7 +108,11 @@ export function newKernelObservable(kernelSpec: KernelInfo, cwd: string) {
           kernelSpecName,
           kernelSpec,
         });
-        observer.complete();
+
+        spawn.on('error', (error) => {
+          observer.error({ type: 'ERROR', payload: error, err: true });
```
rgbkrk, line 113: Excellent!

src/notebook/epics/kernel-launch.js
```diff
@@ -108,7 +108,11 @@ export function newKernelObservable(kernelSpec: KernelInfo, cwd: string) {
...
+          observer.complete();
```
rgbkrk, line 114: This makes me think we need to complete when the spawn also exits gracefully. Is there another event we should add a callback for?

rgbkrk: Looks like both `disconnect` and `exit` should be handled - https://nodejs.org/api/child_process.html#child_process_event_exit

jdetle:
Closes #897

src/notebook/epics/kernel-launch.js
```diff
@@ -108,7 +108,17 @@ export function newKernelObservable(kernelSpec: KernelInfo, cwd: string) {
...
+        });
+        spawn.on('exit', () => {
+          observer.complete();
+        });
+        spawn.on('exit', () => {
```
rgbkrk: I assume you meant `'disconnect'` here.

rgbkrk (changes requested):

src/notebook/epics/kernel-launch.js
jdetle: Ugh, yes. Too excited about my first PR in a while...

rgbkrk:
One added note, these callbacks receive more information that we can pump through if we like to inform users (and ourselves better). That can come later, just a note if you want to make this ✨ .

jdetle:
Fixed and rebased typo.

rgbkrk (approved):

rgbkrk:
This is certainly an improvement on what we had before. Happy to get this in before the next release and kick it around.

rgbkrk:
Code coverage will go down just a bit here, which I'm ok with for the purposes of this PR. Once this passes, we'll merge it. I'll be trying out the release lifecycle next for the main app, making sure our packaging is still ok.

jdetle:
Thank you thank you!

PR #1443, opened 2017-02-09
Publication

### LINKED ISSUE #1438: Publish our first lerna supported packages to npm
We now have a collection of `@nteract` packages. We need to export these so they can be used elsewhere. It would probably be good to start setting up our practices around this soon.

Depends on #1437.

rgbkrk:
Closes #1438.

We're live! We will still need to do transpilation of our packages :soon: though.


jdetle:
Thanks for this awesome maintenance work!

PR #1447, opened 2017-02-09
set nsis:perMachine to enable file associations

### LINKED ISSUE #1060: Windows Installer Configuration
The exe currently gives you no options on e.g. where to install nteract or whether to install for all users or just the current user.

It would be good to allow for such configuration in the installer. Alternatively, bringing back the windows zip file would let you unzip nteract to wherever you wanted

### LINKED ISSUE #1446: Building on Windows
For the 0.1.0 release I was unable to get it to package on Windows due to this error:

```
Building NSIS installer
Packaging NSIS installer for arch x64
Error: Please set perMachine to true — file associations works on Windows only if installed for all users
    at C:\Users\rgbkr\nteract\node_modules\electron-builder\src\targets\nsis.ts:326:15
    at next (native)
    at runCallback (timers.js:574:20)
    at tryOnImmediate (timers.js:554:5)
    at processImmediate [as _immediateCallback] (timers.js:533:5)
From previous event:
    at NsisTarget.executeMakensis (C:\Users\rgbkr\nteract\node_modules\electron-builder\out\targets\nsis.js:412:11)
    at C:\Users\rgbkr\nteract\node_modules\electron-builder\src\targets\nsis.ts:197:60
```

I got this both on my Windows 10 machine and when running an `npm run dist:all` on my mac.

rgbkrk:
fixes #1446
fixes #1060

I just ran the build with these options, I can also set `allowToChangeInstallationDirectory`

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,10 @@
         "zip"
       ]
     },
+    "nsis": {
+      "perMachine": true,
+      "oneClick": false
+    },
     "win": {
       "target": [
         "nsis",


lgeiger (approved):
👍  📦

rgbkrk:
I'm currently uploading with at least these settings so we have a windows release. Technically this means they've been snuck in after the fact for the Windows release. We'll be making another alpha release within the next two weeks though, so I'm not too worried.

PR #1454, opened 2017-02-11
Notebook Preview

rgbkrk:
As part of the package export, I'm setting up a new version of notebook-preview and a React app you can run called notebook-preview-demo.


rgbkrk:
This is now in a state that others can try it from this branch.

Current view:

screen shot 2017-02-15 at 5 48 53 pm

I've made a mess in creating stripped down components and may very well be missing all the dependencies that need to be interlinked. Other than that it's clear the styling is _not quite_ there.

rgbkrk:
To try this out yourself, I think this will be the right path. From the `nbp` branch off my remote (https://github.com/rgbkrk/nteract):

```
~/nteract/$ npm i # Install all dependencies, bootstrap lerna
~/nteract/$ cd packages/notebook-preview-demo
~/nteract/packages/notebook-preview-demo/$ npm run start 
```

The app should now be running on http://localhost:3000/

rgbkrk:
If you want to build/watch the `@nteract/notebook-preview` component, you'll want to run:

```
$(npm bin)/babel -d lib src --ignore '**/__tests__/**' --watch
```

within `packages/notebook-preview`

rgbkrk:
Compare to https://nbviewer.jupyter.org/gist/anonymous/8ea760c2e2a6bba2ebff27125a548508 for the moment.

rgbkrk:
Worked with @peggyrayzis on figuring out what to do next for our editor component and the future of all our components. In a separate PR I'll be creating a package for our current editor (without exporting the provider or view mode ones).

Went ahead and cleaned up the commits here after putting some finishing touches on styling:

screen shot 2017-02-15 at 8 01 08 pm

Syntax highlighting is not working yet, I'm not sure what's up with that. 
¯\\\_(ツ)\_/¯

packages/notebook-preview-demo/src/App.css
```diff
@@ -0,0 +1,28 @@
...
+  height: 80px;
+  padding: 20px;
+  color: white;
+}
+
+.App-intro {
+  font-size: large;
+}
+
+@keyframes wiggle {
```
jdetle, line 22: ^ 🤔  I wonder what this is used for...

rgbkrk: Ever used `create-react-app`? That's your hint.

jdetle: Oh, I was suspecting it was an easter egg a la the `untitled` theme

packages/notebook-preview-demo/yarn.lock
```diff
@@ -0,0 +1,5328 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
lgeiger: This would be the only package with a `yarn.lock` file. On purpose?

lgeiger (commented):
Wow this looks great!

packages/notebook-preview-demo/yarn.lock
rgbkrk: Yeah I think this happened because I created this package with `create-react-app`. I'm not intending to ship this app, it's only a demo.

rgbkrk: Worth noting that the reason `notebook-preview-demo` can use `yarn` is that it's an app that can be deployed vs. the libraries which must employ pure npm & semver.

The _real_ reason that the `yarn.lock` was created was because I used `create-react-app` to bootstrap this and it noticed I use `yarn`. I'm going to delete this `yarn.lock` file.

packages/notebook-preview-demo/src/App.css
rgbkrk: I just modified the animation they provide with the boilerplate.

rgbkrk:
Full styles are in _and_ we're using our _actual_ editor component from nteract here in the notebook preview. Probably want to figure out a way to cut out some of the Rx / completion stuff (providing it only when necessary). Not likely to happen for our first releases here.

rgbkrk:
I need to customize the editor component somehow to put it back in Read only mode, I'll try that out now.

rgbkrk:
screen shot 2017-02-16 at 10 04 45 pm

rgbkrk:
I feel like there is some missing inline style related to padding or margins. 🤔  The code cells butt up against each other.

rgbkrk:
screen shot 2017-02-16 at 10 10 49 pm

rgbkrk:
Rebased, now the editor is `readOnly`. I'm now a big fan of the wrapper with `customOptions`.

rgbkrk:
This is _almost_ done. A couple items for me to handle tomorrow morning:

* delete the yarn lock ✅ 
* decide on how to repackage styles (right now it's copy paste with some light modifications)

rgbkrk:
Big annoyance with our editor styling and CSS - we have a particular DOM+class structure that makes using a static `runMode` version of CodeMirror really tricky.

I got partway using this and I'd like to revisit it, so I think I'll post an issue about refactoring some of our CSS here.

```js
// Wish I could use this, we need to fix up 
export class StaticCodeMirror extends React.Component {
  props: Props;
  el: HTMLElement;

  componentDidMount(): void {
    console.log(this.props);
    CM.runMode(this.props.text, this.props.mode, this.el);
  }

  shouldComponentUpdate(nextProps: Props): boolean {
    return nextProps.text !== this.props.text || this.props.mode !== nextProps.mode;
  }

  componentDidUpdate(): void {
    console.log('update');
    CM.runMode(this.props.text, this.props.mode, this.el);
  }

  render(): ?React.Element<any> {
    return (
      <div className="input">
        <div className="cell_cm cm-s-composition">
          <pre className="CodeMirror-lines" ref={(el) => { this.el = el; }} />
        </div>
      </div>
    );
  }
}
```

rgbkrk:
I started going down the path of simplifying to runMode thinking that it would clear up several things and make it so I didn't have to pass dummy values through to satisfy flow.

jdetle:
Look forward to trying it out!!

rgbkrk:
Sweet, after you get it running it would be super cool to add a README that says how to start up the demo.

PR #1455, opened 2017-02-11
Transpile commutable and keep flow types

rgbkrk:
This transpiles commutable and keeps flow types.


rgbkrk:
This should now pass tests assuming everything gets built in the right order. Returning to this tomorrow (or Monday).

rgbkrk:
Note that this is a breaking change in terms of the published package (namespace changes), yet our package was not consumable for ES5 users yet.

packages/commutable/package.json
```diff
@@ -4,6 +4,11 @@
   "description": "library for immutable notebook operations",
   "main": "index.js",
   "scripts": {
+    "prepublish": "npm run build",
+    "build": "npm run build:clean && npm run build:lib && npm run build:flow",
+    "build:clean": "rimraf lib",
+    "build:flow": "flow-copy-source -v -i '**/__tests__/**' src lib",
+    "build:lib": "babel -d lib src --ignore '**/__tests__/**'",
```
lgeiger, line 11: Since we don't have `**/__tests__/**` directories, I think we don't need exclude them explicitly.

packages/commutable/package.json
```diff
@@ -4,6 +4,11 @@
   "description": "library for immutable notebook operations",
   "main": "index.js",
```
lgeiger, line 5: If we use `./lib/index.js` instead we shouldn't need a dedicated `index.js` file in the root directory.

src/notebook/reducers/document.js
```diff
@@ -32,16 +30,12 @@ import type {
...
-import {
-  createImmutableOutput,
-} from '../../../packages/commutable/v4';
+} from '../../../packages/commutable/lib/types';
 
 import type {
   Output,
   StreamOutput,
-} from '../../../packages/commutable/v4';
+} from '../../../packages/commutable/lib/v4';
```
lgeiger, line 38: Since we `webpack` the the electron app anyway, we may be able to directly use the `src` files.

lgeiger (commented):
Woo 🎉 

Why was the [ES5ification](https://github.com/nteract/nteract/pull/1455/commits/14cbd4c10caffa270dab73dd91cc91902be1c389) necessary? Shouldn't babel handle this for us.

src/notebook/reducers/document.js
rgbkrk: We totally can, I'm hoping we can transition to the built versions of all of these so that we straddle both the electron app publishing and npm publishing.

packages/commutable/package.json
rgbkrk: I think I'll counter that with adding some tests here. 😉

packages/commutable/package.json
rgbkrk: Might as well, I can switch that. The reason I usually tend to do it is so we're explicitly picking what we want to export/maintain, to ease semver.

lgeiger: I don't feel strongly about that. I'm totally with keeping it as is.

packages/commutable/package.json
lgeiger: Even better 👍

rgbkrk:
> Why was the ES5ification necessary? Shouldn't babel handle this for us.

Since we weren't using any classes here I figured I could trim down on what we [end up exporting](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&targets=&browsers=&builtIns=false&code=export%20function%20X()%20%7B%7D) vs. [a little more explicit](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&targets=&browsers=&builtIns=false&code=function%20X()%20%7B%7D%0A%0Amodule.exports%20%3D%20%7B%0A%20%20X%2C%0A%7D).

lgeiger:
> Since we weren't using any classes here I figured I could trim down on what we end up exporting.

👍 Thanks for the explaination. I was just curious.

lgeiger (approved):

rgbkrk:
Yeah it definitely goes against what's in other parts of our code base. If I didn't enjoy the declarative nature of JSX so much, I'd do the same with the transforms. ;)

packages/commutable/package.json
peggyrayzis: We're using babel-cli in our build:lib command but don't have it listed as a dependency - should we change this?

peggyrayzis (approved):
Everything looks great here! Just had one small question on the build:lib script.

Glad to see that flow-copy-source was an easy solution for us! 👍

rgbkrk:
Thanks @peggyrayzis! I'll add `babel-cli` to our packages here as well.

rgbkrk:
Thanks for finishing this off @peggyrayzis!

On to _even more_ maintenance work to get our monorepo packages in a good state.

PR #1475, opened 2017-02-16
Extract messaging and editor

rgbkrk:
Extracted the editor and the messaging.

The only part of the editor that is extracted is our base editor component and the view version (@peggyrayzis can explain better than I).


rgbkrk:
I'm going to leave the messaging/session refactor for another day as it would make this PR messier than intended. Mostly just want the editor component out.

packages/editor/__tests__/editor-spec.js
```diff
@@ -2,19 +2,14 @@ import React from 'react';
...
+import WrappedEditor from '../src/view';
 
-chai.use(sinonChai);
-
-const complete = require('../../../../src/notebook/components/cell/codemirror/complete');
+const complete = require('../src/complete');
 
 describe('WrappedEditor', () => {
-  it('reaches out for code completion', (done) => {
+  it.skip('reaches out for code completion', (done) => {
```
rgbkrk, line 12: I had trouble mocking one function from `complete` here so I've skipped this test for now.

rgbkrk:
Once this is in, I'll bring it into the notebook preview PR.

packages/editor/__tests__/editor-spec.js
```diff
@@ -2,19 +2,14 @@ import React from 'react';
 import Rx from 'rxjs/Rx';
 
 import { mount } from 'enzyme';
-import chai, { expect } from 'chai';
```
jdetle, line 5: Where is expect getting imported from now? Or is it no longer needed in jest?

jdetle (approved):

packages/editor/__tests__/editor-spec.js
jdetle: Ahh, it appears this is a benefit of Jest, awesome!

peggyrayzis:
Just getting around to looking at this, it looks great! Thanks for extracting the editor package 😀

PR #1477, opened 2017-02-16
fix(raw): Render raw as markdown.

jdetle:
Hot fix to prevent notebook from failing if it contains a raw cell type.


rgbkrk:
Works for me, we'll figure out what raw cells mean to us soon enough.

rgbkrk (approved):

captainsafia:
> Works for me, we'll figure out what raw cells mean to us soon enough.

As I work on the V3 converter, I think it might be a legacy carryover from older versions of the notebook.

PR #1484, opened 2017-02-18
all our packages should use the same babel config

rgbkrk:
Spreading babel config everywhere. 🤔  I wonder if I could just place a `.babelrc` inside of `packages/`. Let me try that, don't merge me.


rgbkrk:
I like the approach in #1485, it lets a base setup be picked up in `packages` so we don't have to cargo cult it everywhere and we can still modify it for any particular package.

PR #1486, opened 2017-02-18
style(css): scope CSS to allow embedding notebook

rgbkrk:
When using `notebook-preview` in commuter, it messed up the styling of other pages because of the global scope (that was intended for only outputs).


rgbkrk:
I look forward to when the display area has its own inline styles which can be overridden by a top level theme.

jdetle:
Thanks for this catch! Awesome work.

Issue #874, opened 2016-09-30
Custom Observable for our zmq and websocket subjects


rgbkrk:
# Creating custom Observables for our kernel subjects

Two weeks ago I had a conversation with @jayphelps and @SomeKittens which led to [wrapping some of our Observable creation methods with the originating `send`](https://github.com/nteract/nteract/blob/181b2f08337b25f62020c3de90267608e1395faf/src/notebook/epics/execute.js#L200), which roughly looks like this:

```js
function executeCodeObservable(code: string) {
  const executeRequest = createExecuteRequest(code);

  // All the message handling for responses gets wired up...
  const cellAction$ = Rx.Observable.merge(bunchOfCellAction$);

  // On subscription, send the message
  return Rx.Observable.create(observer => {
    const subscription = cellAction$.subscribe(observer);
    channels.shell.next(executeRequest);
    return subscription;
  });
}
```

That works pretty awesome.

In quite a few cases we want a message sent as soon as someone subscribes to a resulting observable. @jayphelps' suggestion was to create a custom observable that overrides `_subscribe` and handles this stuff in a custom subscriber a la [AjaxObservable's _subscribe](https://github.com/ReactiveX/rxjs/blob/e20cdb73e9a6255d704e39c2e867b3227207b4b0/src/observable/dom/AjaxObservable.ts#L171). WebSocketSubject is also a good example of that, which we should learn a bit about when we start using nteract with websocket based kernels.
## Background on nteract's channels

Our primary channels that we care about are `shell` and `iopub`. They're built on top of ZMQ sockets (though they could also be transported over websockets).
### `shell`

This is a `DEALER` socket, in ZMQ parlance, with the backend kernel/runtime being the ROUTER socket that accepts messages. Frontends send messages like `execute_request`.
### `iopub`

This is a `SUB` socket. Messages are never sent on this socket - it's only a receiver. Responses from the `execute_request` will end up on this channel.
### Next steps

Both of these channels stay alive up until they connect to a new backend. Roughly, this:

``` js
export const newKernelEpic = action$ =>
  action$.ofType(LAUNCH_KERNEL)
    .switchMap(action => Observable.of(newKernelAction(action.channels))
```

When we do execution, we rely on the channels being part of the state tree to use throughout the app (it also gets used for autocompletion, separately from redux but still with Rx and CodeMirror).

benlesh:
[...]

PR #1488, opened 2017-02-18
feat(editor): add tooltip for introspection

### LINKED ISSUE #1254: Tooltip for function signature/docstring
In the jupyter notebook `Shift-Tab` will bring up a tooltip showing the function signature/docsting whcih is incredibly handy.  It would be great if this were also available in nteract.

alexandercbooth:
Closes #1254. 

Following discussion in #1254, I think the basic parts are here.

### Current Status
---
- I'm currently sending the request message and reply bundle (thanks for the help on the reply, John!) to the console to confirm that it's operating as expected.
- There's also the addWidget that prints the bundle in the cell (you have to press enter). Of course it doesn't display since it hasn't been transformed. See picture:

screen shot 2017-02-18 at 5 08 56 pm
---
### Next steps
- I'd like to transform bundle and display using the widget (putting aside the actual UI for the moment). I can see the basic usage of the transforms package for the bundle from the readme, but I don't quite know how to implement this and use react to display it

Thanks in advance!

### DIFF
--- a/packages/editor/src/tooltip.js
+++ b/packages/editor/src/tooltip.js
@@ -1,0 +1,43 @@
+import Rx from 'rxjs/Rx';
+
+import {
+  createMessage,
+} from '@nteract/messaging';
+
+export function tooltipObservable(channels, editor, message) {
+  const tip$ = channels.shell
+    .childOf(message)
+    .ofMessageType(['inspect_reply'])
+    .pluck('content')
+    .first()
+    .map(results => ({
+      dict: results.data,
+    }));
+  // On subscription, send the message
+  return Rx.Observable.create((observer) => {
+    const subscription = tip$.subscribe(observer);
+    channels.shell.next(message);
+    return subscription;
+  });
+}
+
+export const tooltipRequest = (code, cursorPos) =>
+  createMessage(
+    'inspect_request',
+    {
+      content: {
+        code,
+        cursor_pos: cursorPos,
+        detail_level: 0,
+      }
+    }
+  );
+
+export function tool(channels, editor) {
+  const cursor = editor.getCursor();
+  const cursorPos = editor.indexFromPos(cursor);
+  const code = editor.getValue();
+
+  const message = tooltipRequest(code, cursorPos);
+  return tooltipObservable(channels, editor, message);
+}
--- a/packages/editor/src/wrapper.js
+++ b/packages/editor/src/wrapper.js
@@ -1,34 +1,40 @@
 // @flow
 /* eslint-disable class-methods-use-this */
-import React, { PureComponent } from "react";
-import Rx from "rxjs/Rx";
-import CodeMirror from "react-codemirror";
-import CM from "codemirror";
+import React, { PureComponent } from 'react';
+import ReactDOM from 'react-dom';
+import Rx from 'rxjs/Rx';
+import CodeMirror from 'react-codemirror';
+import CM from 'codemirror';
+import { Map as ImmutableMap } from 'immutable';
 
-import "codemirror/addon/hint/show-hint";
-import "codemirror/addon/hint/anyword-hint";
-import "codemirror/addon/search/search";
-import "codemirror/addon/search/searchcursor";
-import "codemirror/addon/edit/matchbrackets";
-import "codemirror/addon/edit/closebrackets";
-import "codemirror/addon/dialog/dialog";
-import "codemirror/addon/comment/comment.js";
+import {
+  transforms
+} from '@nteract/transforms';
 
-import "codemirror/mode/python/python";
-import "codemirror/mode/ruby/ruby";
-import "codemirror/mode/javascript/javascript";
-import "codemirror/mode/css/css";
-import "codemirror/mode/julia/julia";
-import "codemirror/mode/r/r";
-import "codemirror/mode/clike/clike";
-import "codemirror/mode/shell/shell";
-import "codemirror/mode/sql/sql";
-import "codemirror/mode/markdown/markdown";
-import "codemirror/mode/gfm/gfm";
+import 'codemirror/addon/hint/show-hint';
+import 'codemirror/addon/hint/anyword-hint';
+import 'codemirror/addon/search/search';
+import 'codemirror/addon/search/searchcursor';
+import 'codemirror/addon/edit/matchbrackets';
+import 'codemirror/addon/edit/closebrackets';
+import 'codemirror/addon/dialog/dialog';
+import 'codemirror/addon/comment/comment.js';
+import 'codemirror/mode/python/python';
+import 'codemirror/mode/ruby/ruby';
+import 'codemirror/mode/javascript/javascript';
+import 'codemirror/mode/css/css';
+import 'codemirror/mode/julia/julia';
+import 'codemirror/mode/r/r';
+import 'codemirror/mode/clike/clike';
+import 'codemirror/mode/shell/shell';
+import 'codemirror/mode/sql/sql';
+import 'codemirror/mode/markdown/markdown';
+import 'codemirror/mode/gfm/gfm';
 
-import "./codemirror-ipython";
-import excludedIntelliSenseTriggerKeys from "./excludedIntelliSenseKeys";
-import { codeComplete, pick } from "./complete";
+import './codemirror-ipython';
+import excludedIntelliSenseTriggerKeys from './excludedIntelliSenseKeys';
+import { codeComplete, pick } from './complete';
+import { tool } from './tooltip';
 
 type WrapperProps = {
   id: string,
@@ -36,6 +42,7 @@ type WrapperProps = {
   editorFocused: boolean,
   cellFocused: boolean,
   completion: boolean,
+  tip: boolean,
   focusAbove: () => void,
   focusBelow: () => void,
   theme: string,
@@ -63,11 +70,13 @@ const CodeMirrorWrapper: CodeMirrorHOC = (EditorView, customOptions = null) =>
     goLineDownOrEmit: (editor: Object) => void;
     executeTab: (editor: Object) => void;
     hint: (editor: Object, cb: Function) => void;
+    tips: (editor: Object) => void;
 
     constructor(): void {
       super();
 
       this.hint = this.completions.bind(this);
+      this.tips = this.tips.bind(this);
       this.hint.async = true;
     }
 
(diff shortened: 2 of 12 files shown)


alexandercbooth:
Forgot to say, the key binding is `Cmd-.`

packages/editor/src/tooltip.js
```diff
@@ -0,0 +1,47 @@
...
+    .ofMessageType(['inspect_reply'])
+    .pluck('content')
+    .first()
+    .map((results) => {
+      console.log(results.data);
+      return {
+        dict: results.data,
+      };
+    });
+  // On subscription, send the message
```
rgbkrk: 👍  That's our trick for now until #874. 😄

packages/editor/src/wrapper.js
```diff
@@ -132,6 +137,18 @@ const CodeMirrorWrapper: CodeMirrorHOC = (EditorView, customOptions = null) =>
...
 
+    tips(editor: Object, callback: Function): void {
+      const { tip, channels } = this.props;
+      const node = document.createElement('P');
+      const text = document.createTextNode(tool(channels, editor));
+      node.appendChild(text);
+      if (tip) {
+        tool(channels, editor)
+          .subscribe(callback,
+          editor.addWidget(editor.getCursor(), node, true));
```
rgbkrk: The order of the arguments looks a bit strange to me here, `subscribe`'s arguments should be:

```js
subscribe(onNext, onError, onClose);
```

For now you probably want to only use `onNext` to create the node on response.

```js
tool(channels, editor)
  .subscribe((resp) => {
    const node = document.createElement('pre');
    const text = document.createTextNode(JSON.stringify(resp));
    editor.addWidget(editor.getCursor(), node, true);
  });
```

Note that the response comes in async, we probably don't want to add a widget if they've moved the cursor. We can fix that later though.

As for rendering inline with React here -

```js
const node = document.createElement('div');
ReactDOM.render(<Something />, node);
editor.addWidget(editor.getCursor(), node, true);
```

packages/editor/src/tooltip.js
alexandercbooth: 😁

packages/editor/src/wrapper.js
alexandercbooth: Thanks @rgbkrk, you rock!

So now I think all the plumbing is in place. Here is the updated transformed output:
screen shot 2017-02-19 at 1 55 50 pm

If this looks good, I will start to think about UI.

Thanks as always!

alexandercbooth:
Also, looks like CI failing from `Cannot find module '@nteract/transforms'` (besides flow). Is there somewhere I need to list `@nteract/transforms` as a dependency or do something else with it?

rgbkrk:
> Also, looks like CI failing from Cannot find module '@nteract/transforms' (besides flow). Is there somewhere I need to list @nteract/transforms as a dependency or do something else with it?

`packages/editor/package.json` will need to have `@nteract/transforms`. After you add that in, run `npm run lerna` at the base of our repository and it will link up the packages for you.

packages/editor/src/wrapper.js
rgbkrk: Woo!

Looking forward to this, UI definitely seems a bit tricky here.

alexandercbooth:
Oops, I seem to have created a conflict.

packages/editor/src/wrapper.js
alexandercbooth: Me too! Now the real challenge begins..😅

rgbkrk:
For at least this flow error:

```
packages/editor/src/wrapper.js:155
155:             ReactDOM.render(<Transform data={bundle.get(mimetype)} />, node);
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ React element `Transform`. Expected React component instead of
153:             const Transform = transforms.get(mimetype);
                                   ^^^^^^^^^^^^^^^^^^^^^^^^ Transform
```

I think it's the `Transform` definition that is wrong, not yours. I was having trouble making that one work because of the `MIMETYPE` static property, so it's not in the best of states.

alexandercbooth:
# UI First Draft
ui1

---
Sorry for the poor quality gif, which looks a bit fatter than intended -- guess I am a bit rusty with ffmpeg 😅. 

This is just a first pass and meant to get the ball rolling on UI. Pressing the command launches it on the bottom of the notebook and closes it. I also included a button to close just to have the option to play with later.

### Formatting Issue
I didn't use the `addWidget` from codemirror since, if I understand it right, it would have restricted the display to within the cell and I had a hard time making that work. One issue, however, is that with `addWidget`, the display is formatted how it should be (with whitespace included), but with `insertbefore` (what I am using here) it seems to strip this whitespace.

Mainly just starting to throw some ideas around here to see what might be useful/attractive to the user. I'll continue adding some ideas as I get the chance. Feedback most welcome 😄.

alexandercbooth:
Same, now with new and improved gif 😄
---
ui1

jdetle:
I'm all for merging this as it is now and tinkering with it for a bit before we make another release. It looks like there are some issues with the flow typings that you're going to need to fix to get CI to pass.

jdetle:
Have you tried using the notification system? Curious to see how that would look too.

rgbkrk:
If we merge this in as is, I _must_ say that I'm not a fan of having this in a pager at the bottom. I much prefer the inline display shown in Jupyter for inspection.

Perhaps @Carreau has some good ideas here.

alexandercbooth:
Hi All,

Thanks for the feedback so far! This [W] is very much [IP] at the moment, so I think it might be safer to wait a little bit before merging.

> Have you tried using the notification system? Curious to see how that would look too.

I have not yet, but I will definitely look into it, thanks!

> If we merge this in as is, I must say that I'm not a fan of having this in a pager at the bottom. I much prefer the inline display shown in Jupyter for inspection.

To be honest, I am 100% with you on this @rgbkrk. Chucking a `?` or `??` at a function in Jupyter and having it pop up on the bottom was always too disruptive for me to find it comfortable using and I ultimately stopped doing it. Then when I discovered the tooltip, I was very much 😍. I started with the bottom version after hitting a small roadblock inline and since it was the easiest to implement (at least roughly), I figured it would be good way to get the ball rolling. I am working on an inline version and will share once I've got it working. Thanks again for all your help on this!

Thanks everyone for the feedback, very much appreciated!

alexandercbooth:
## Inline
Here was the inline iteration. The place where I got stuck was how to make the cell expand and retract as needed. To get this I manually expanded the cell by pressing return before the video.
inline1

Carreau:
> Perhaps @Carreau has some good ideas here.

Tooltip are definitively appreciated by users. I don't know enough about the nteract codebase, but one of the reason the tooltip was implemented outside of the cells in notebook was to be able to bleed outside of CP instance. 

We may be able to pop-out of clipped overflow: https://css-tricks.com/popping-hidden-overflow/ but I haven't really digged into that.

dhirschfeld:
I quite like the current notebook tooltip - for ref:
tooltip

In particular I like that it only shows the signature / first line initially as some docstrings can be huge and  so would be unwieldy in a tooltip. You can expand the tooltip to see the full docstring in a scrollable window.

You can also send it to the pager which gives you way to view two signatures/docstrings at the same time. IMHO, the ideal UI here would be to be able to "pin" the tooltip so that it became a movable/draggable  floating, window and you could pin as many tooltips as you wanted to your workspace.

Just some thoughts I've been mulling over, also in respect to the jupyterlab tooltip [discussion](https://github.com/jupyterlab/jupyterlab/issues/688)

alexandercbooth:
## UI Update
ui2

## Inline render
I've now got the inline widget expanding/retracting the cell as needed. 

>In particular I like that it only shows the signature / first line initially as some docstrings can be huge and so would be unwieldy in a tooltip. You can expand the tooltip to see the full docstring in a scrollable window.

Thanks very much for your thoughts, @dhirschfeld! I really like this functionality, too. The gif shows how it can be annoying to automatically expand a large docstring -- I will try to incorporate handling this going forward.

>You can also send it to the pager which gives you way to view two signatures/docstrings at the same time. IMHO, the ideal UI here would be to be able to "pin" the tooltip so that it became a movable/draggable floating, window and you could pin as many tooltips as you wanted to your workspace.

I like the idea of potentially making it draggable/sticky. The limitation for this at the moment is that it is rendered inline in the cell, but I am definitely open to exploring it.

Additionally, I like @Carreau's idea of using css to break outside the cell (instead of expanding it), but have had trouble making it happen with the css. I'll keep trying, though -- thanks @Carreau! 

@rgbkrk @jdetle Do you have thoughts on a particular direction you might like to see this go in?

Thanks again to everyone for all your input! 

PS I'm getting better at gifs 😉

Carreau:
Idea: look at how the completer is implemented. It does not expand the cell
IIRC.

alexandercbooth:
@Carreau, you are right! I don't know how I missed that before, but that is really helpful. I will try to iterate around that. Thanks so much!

alexandercbooth:
## Update
It's not perfect, but it's looking better now. Once the tests are fixed up (and assuming I didn't blow the rebase 😬), it might be in an OK place to merge (and continue iterating on design). Thoughts?

---

update

rgbkrk:
That's looking really good! The only things I'd worry about completing on this now is:

* Get tests passing
* (optionally) put an X in the close box

alexandercbooth:
Awesome, thanks Kyle!

lgeiger:
Great, Thanks for getting this started!

alexandercbooth:
I'll try and clean up the flow issues this weekend

Carreau:
Just out of information here is the current  behavior of classic notebook
 - first press of shift-tab: show tooltip with a limited height ~10 lines, 
 - second press : expand it and make it scrollable. (similar to clicking the `+` button)
 - third press: make it sticky (typing won't dismiss for 10sec, show a clock icon)
 - 4th press, send to pager and dismiss.

It's really rare to see people knowing about these behavior, and I definitely did a bad job of communicating this and having the UI for it. I believe the 3rd (sticky) behavior is not that useful, not is the 4rth. 

Though I'd like to have more change to investigate what multiple tooltip requests would do.

jdetle:
@alexandercbooth, I'm happy to either pair or make some commits to try to help fix the CI issues. Are you okay with collaborating?

rgbkrk:
```
packages/editor/lib/wrapper.js.flow:179
179:           ReactDOM.render(<Transform data={bundle.get(mimetype)} />, node);
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ React element `Transform`. Expected React component instead of
175:           const Transform = transforms.get(mimetype);
                                 ^^^^^^^^^^^^^^^^^^^^^^^^ Transform
packages/editor/src/wrapper.js:179
179:           ReactDOM.render(<Transform data={bundle.get(mimetype)} />, node);
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ React element `Transform`. Expected React component instead of
175:           const Transform = transforms.get(mimetype);
                                 ^^^^^^^^^^^^^^^^^^^^^^^^ Transform
```

This flow error may be hard to fix, since we're tacking on a static property to React Components. Our current transform interface is a module which has an export of a React Component with a static property called `MIMETYPE`. I'm not sure how to get around the flow typing of this, so it might be that we change the interface to an exported object `{ transform: ReactComponent, MIMETYPE: string}` then we'd be able to declare the type for it.

lgeiger:
> This flow error may be hard to fix

For now we could add a `$FlowFixMe` above this line to make the build pass.

alexandercbooth:
Thank you both!

I am happy to go either way, although I don't fully understand the implications of how changing the transform interface would affect the other parts of the codebase that utilize it.

rgbkrk:
> I don't fully understand the implications of how changing the transform interface would affect the other parts of the codebase that utilize it.

I'm effectively saying we would have to refactor all the transforms to conform to one new interface with the module exports, which would mean they would all be a major bump, semver wise. We could straddle the old API with the new one and make it a minor release, seems like it would be ok to just do the major release thing. We can make a separate PR that does this refactor.

rgbkrk:
I think for now the `$FlowFixMe` is fine. Add this line to the `.flowconfig` in the `[options]` section:

```
suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
```

Then where necessary to supress, comment with `// $FlowFixMe: Some Reason`.

For more on [this type of suppression, check out the Flow docs](https://flow.org/en/docs/config/options/#suppresscomment-regex-a-classtoc-idtoc-suppress-comment-regex-hreftoc-suppress-comment-regexa).

alexandercbooth:
Oops -- did prettier make all those extra changes automatically (I only changed the .flowconfig and wrapper.js)?

alexandercbooth:
Actually, just realized I think those extra changes were from when I pulled in the changes after the prettier adoption -- I think I accidentally just pulled in the changes and committed instead of rebasing.

rgbkrk:
This is looking real good. It's saying this needs a rebase -- can you do an interactive rebase here?

alexandercbooth:
Rats, it looks like the issue is still there. Do you think it's because I had forgotten to rebase after fetching in fb3050ca83846855ccdd34e5ab650122238f1ccf? Would I have wanted to cut that one out somehow and then fetch and rebase? I kept a local copy with that history in case I mix anything up here.

rgbkrk:
If it helps, I can rebase it for you. Otherwise, I'm more than happy to screenshare and pair to fix this up.

alexandercbooth:
Thanks, that would be great to screenshare, do you have time this afternoon?

rgbkrk:
I'll have to do tomorrow April 4 or Thursday April 6.

rgbkrk:
Paired with Alex on a call to get this rebased as well as fix the build (enzyme & React 15.5 changes). I think this is in a good state to iterate from here on. :smile:

PR #1489, opened 2017-02-20
feat(commutable): Add v3 to v4 converter

### LINKED ISSUE #1461: Write v3 notebook -> commutable converter
We need to write a `fromJS` that converts from v3 notebooks to our in-memory format.

captainsafia:
Closes #1461.

### DIFF
--- a/packages/commutable/src/index.js
+++ b/packages/commutable/src/index.js
@@ -13,6 +13,7 @@ type PlaceholderNotebook = {
 };
 
 const v4 = require('./v4');
+const v3 = require('./v3');
 
 const {
   emptyNotebook,
@@ -46,6 +47,8 @@ function fromJS(notebookJSON: Notebook): ImmutableNotebook {
     if (Array.isArray(notebookJSON.cells) && typeof notebookJSON.metadata === 'object') {
       return v4.fromJS(notebookJSON);
     }
+  } else if (notebookJSON.nbformat === 3 && notebookJSON.nbformat_minor >= 0) {
+    return v3.fromJS(notebookJSON);
   }
 
   if (notebookJSON.nbformat) {
--- a/packages/commutable/src/index.js
+++ b/packages/commutable/src/index.js
@@ -64,7 +64,7 @@ function toJS(immnb: ImmutableNotebook): v4Notebook {
   if (immnb.get('nbformat') === 4 && immnb.get('nbformat_minor') >= 0) {
     return v4.toJS(immnb);
   }
-  throw new TypeError('Only notebook format 4 is supported');
+  throw new TypeError('Only notebook formats 3 and 4 are supported!');
 }
 
 // Expected usage is stringifyNotebook(toJS(immutableNotebook))
--- a/packages/commutable/src/index.js
+++ b/packages/commutable/src/index.js
@@ -1,6 +1,7 @@
 /* @flow */
 
 import type { Notebook as v4Notebook } from './v4';
+import type { Notebook as v3Notebook } from './v3';
 
 import type {
   ImmutableNotebook,
@@ -29,7 +30,7 @@ const {
   removeCell,
 } = require('./structures');
 
-export type Notebook = PlaceholderNotebook & v4Notebook;
+export type Notebook = PlaceholderNotebook & v4Notebook & v3Notebook;
 
 function freezeReviver(k: string, v: JSONType): JSONType {
   return Object.freeze(v);
--- a/packages/commutable/src/v3.js
+++ b/packages/commutable/src/v3.js
@@ -14,8 +14,15 @@ import type {
 } from './types';
 
 import type {
-    CellStructure,
+  CellStructure,
 } from './structures';
+
+import type {
+  MultilineString,
+  ErrorOutput,
+  RawCell,
+  MarkdownCell,
+} from './v4';
 
 function demultiline(s: string | Array<string>) {
     if (Array.isArray(s)) {
@@ -23,8 +30,6 @@ function demultiline(s: string | Array<string>) {
         }
     return s;
 }
-
-type MultiLineString = string | Array<string>;
 
 export type Notebook = {|
   worksheets: Array<Worksheet>,
@@ -73,26 +78,7 @@ export type StreamOutput = {|
   text: MultiLineString,
 |};
 
-export type ErrorOutput = {|
-  output_type: 'pyerr',
-  ename: string,
-  evalue: string,
-  traceback: Array<string>,
-|};
-
 export type Output = ExecuteResult | DisplayData | StreamOutput | ErrorOutput;
-
-export type RawCell = {|
-  cell_type: 'raw',
-  metadata: JSONObject,
-  source: MultiLineString,
-|};
-
-export type MarkdownCell = {|
-  cell_type: 'markdown',
-  metadata: JSONObject,
-  source: MultiLineString,
-|};
 
 export type HeadingCell = {|
   cell_type: 'heading',
--- a/packages/commutable/src/v3.js
+++ b/packages/commutable/src/v3.js
@@ -4,6 +4,7 @@ const Immutable = require('immutable');
 const appendCell = require('./structures').appendCell;
 
 import type {
+  JSONObject,
   ImmutableNotebook,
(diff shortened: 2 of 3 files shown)


packages/commutable/src/index.js
```diff
@@ -61,7 +65,7 @@ function toJS(immnb: ImmutableNotebook): v4Notebook {
   if (immnb.get('nbformat') === 4 && immnb.get('nbformat_minor') >= 0) {
     return v4.toJS(immnb);
   }
-  throw new TypeError('Only notebook format 4 is supported');
+  throw new TypeError('Only notebook formats 3 and 4 are supported!');
 }
```
rgbkrk, line 70: This is looking pretty stellar so far.

packages/commutable/src/v3.js
```diff
@@ -0,0 +1,247 @@
...
+  ImmutableMarkdownCell,
+  ImmutableRawCell,
+  ImmutableCell,
+  ImmutableOutput,
+  ImmutableMimeBundle,
+} from './types';
+
+import type {
+  CellStructure,
+} from './structures';
```
rgbkrk, line 19: Since you're in here working with these, what do you think of the current `CellStructure` that has both `cellMap` and `cellOrder`? I've begun to wonder if we should solely have `cellOrder` and nest an ID into the `cell` (that would not persist to disk).

packages/commutable/src/v3.js
```diff
@@ -0,0 +1,247 @@
...
+  prompt_number?: number,
+  text?: MultiLineString,
+  latex?: MultiLineString,
+  png?: MultiLineString,
+  jpeg?: MultiLineString,
+  svg?: MultiLineString,
+  html?: MultiLineString,
+  javascript?: MultiLineString,
+  json?: MultiLineString,
+  pdf?: MultiLineString,
```
rgbkrk, line 80: Wow these are pretty pesky. I definitely like the current v4 nbschema better!

packages/commutable/src/v3.js
```diff
@@ -0,0 +1,247 @@
...
+    "text": "text/plain",
+    "latex": "text/latex",
+    "png": "image/png",
+    "jpeg": "image/jpeg",
+    "svg": "image/svg+xml",
+    "html": "text/html",
+    "javascript": "application/x-javascript",
+    "json": "application/javascript",
+    "pdf": "application/pdf",
+    "metadata": "",
```
rgbkrk, line 129: Perhaps we should call this `VALID_KEYS` since it's verifying all the keys.

Or wait, it looks like there's an explicit check for `metadata`, `output_type`, and `prompt_number`. Could take them out of this list then.

packages/commutable/src/v3.js
```diff
@@ -0,0 +1,247 @@
...
+      mimeBundle[VALID_MIMETYPES[key]] = output[key];
+    }
+  });
+  return Object.keys(mimeBundle)
+    .reduce(cleanMimeAtKey.bind(null, mimeBundle), Immutable.Map());
+}
+
+function sanitize(o: ExecuteResult | DisplayData) {
+  if (o.metadata) {
+    return { metadata: Immutable.fromJS(o.metadata) };
```
rgbkrk, line 145: Something I can't completely recall from how I did the v4 version - was it only display data that was kept as a frozen object rather than Immutable? (I'll hunt to find out after posting this).

packages/commutable/src/v4.js
```diff
@@ -419,11 +419,3 @@ function toJS(immnb: ImmutableNotebook): Notebook {
...
   };
 }
-
-module.exports = {
-  createImmutableMimeBundle,
-  createImmutableOutput,
-
-  toJS,
-  fromJS,
-};
```
rgbkrk, line 429: One nice effect of having the `module.exports` being explicit here was that we could see everything that was exported. Of course... I realize I shifted us away from our original style so :100: for bringing it back.

packages/commutable/src/v3.js
```diff
@@ -0,0 +1,247 @@
...
+  const starterCellStructure = {
+    cellOrder: Immutable.List().asMutable(),
+    cellMap: Immutable.Map().asMutable(),
+  }
+
+  const cellStructure = [].concat.apply([], notebook.worksheets.map((worksheet) => {
+    return worksheet.cells.reduce((cellStruct: CellStructure, cell: Cell) =>
+      appendCell(cellStruct, createImmutableCell(cell)),
+    starterCellStructure);
+  }))[0];
```
rgbkrk, line 241: Looking slick here. Should this do `notebook.worksheets[0]` since it's stripped off at the end here?

This also looks like a section that should have some bounds checking so we don't access `[0]` when it doesn't exist.

rgbkrk (approved):

packages/commutable/src/v3.js
rgbkrk: Turns out that doesn't matter because we didn't have JSON types in the display bundles before, so it wasn't a big deal.

packages/commutable/src/v3.js
captainsafia: This is me fidgeting with things to get Flow to be happy. Despite the if-statement being there, it doesn't like the fact that some keys in DisplayData and ExecuteResult are not present in VALID_MIMETYPES.

packages/commutable/src/v3.js
captainsafia: Mmmm. Yeah. This was me trying to work through the issue of whether or not it was possible for multiple worksheets to exist. I know there was no interface for it but is there any remote possibility that worksheets might actually be an array of length greater than 1?

packages/commutable/src/v3.js
captainsafia: It was both ExecuteResult and DisplayData.

packages/commutable/src/v4.js
captainsafia: You're absolutely right. Flow was angry about the mixed exports styles and I ended up tagging the export inline with the function because it was easier to consolidate.

packages/commutable/src/v3.js
rgbkrk: Mostly just uncertainty from me. The other way around this for safety sake is another flat map like you did with `[].concat.apply([], ...`

rgbkrk:
Looking at the flow output reminds me that we should set up flow to ignore the `lib/` directory amongst all the packages. 🤔  or is that the one we need since it's the one referenced by other libraries. 
¯\\\_(ツ)\_/¯

rgbkrk:
I added on another type annotation to your last commit, though I think there's one more flow issue to fix up. If this passes let's merge it!

rgbkrk:
Oh I see, what's failing for me locally is an artifact of `notebook-preview` coming after this commit. This should pass shortly.

packages/commutable/src/v3.js
```diff
@@ -0,0 +1,250 @@
...
+    "html": "text/html",
+    "javascript": "application/x-javascript",
+    "json": "application/javascript",
+    "pdf": "application/pdf",
+    "metadata": "",
+    "prompt_number": "",
+    "output_type": "",
+  };
+  const mimeBundle = {};
+  Object.keys(output).map(function(key: string) {
```
rgbkrk, line 134: This was the annotation I added to make this work, I'm not sure why it though this was a number as it's mapping over the object keys `Object.keys({'a': 'adsf', 'b': 'basdg'}) ---> ["a", "b"]`

rgbkrk:
Thanks again for doing the v3 to in-memory v4 notebook conversion here. 😄

rgbkrk:
Awww Appveyor, what's the issue?

rgbkrk:
Rekicking the appveyor build for now.

captainsafia:
Oh thanks for helping out with the flow errors here. You rock!

rgbkrk:
Looks like a hoisting problem, I think I better do some lerna package cleanup here (or make them depend on `jest: *`)

rgbkrk:
Everything passes!

captainsafia:
YYAAAAASSSSSS!

captainsafia:
Thanks a bunch for bringing this through the finish line.

PR #1500, opened 2017-02-24
style(css): don't center images

### LINKED ISSUE #1498: Matplotlib inline plots are centered
Matplotlib inline plots are centered:
image

Altair and plotly plots are all left-aligned.
Matplotlib plots are also (correctly) left aligned in jupyter notebooks
image

```
%matplotlib inline
import matplotlib.pyplot as plt
plt.plot([1,2,3,4])
plt.ylabel('some numbers')
plt.show()
```

rgbkrk:
Fixes #1498.


betatim:
:art:

PR #1511, opened 2017-03-01
chore(tests): Run package tests on CI

lgeiger:
This currently fails with:
```bash
FAIL  __tests__/text-spec.js
  > Text › renders plain text

    expect(received).toEqual(expected)
    
    Expected value to equal:
      "<code><span>hey</span></code>"
    Received:
      "<code><span><!-- react-text: 3 -->hey<!-- /react-text --></span></code>"
```


lgeiger:
I guess this error is caused by the `ansi-react` upgrade in #1427.

Any idea, how we should fix it?

rgbkrk:
The output looks alright, we should change the test up.

lgeiger:
👍 I fixed the failing test.

rgbkrk (approved):

rgbkrk:
Thanks! I think we'll want a check that's more agnostic of inner structure in the future -- works for now though!

PR #1514, opened 2017-03-01
chore(packages): Manage packages from the root dir

lgeiger:
This removes all dev dependencies from the packages since lerna links them already (inclusive `bin`).

This PR also removes `@nteract/dev-script` and `@nteract/mockument`.

Right now we're using `lerna bootstrap --hoist` which links everything.
In a future PR I'll add a glob pattern to only link the relevant dev dependencies. This *should* improve speed and file size a lot.


rgbkrk (approved):
If this works well then I'm certainly all for it. I'll make sure to deprecate the mockument and nteract-scripts packages if this does the job well.

lgeiger:
Oh wait, I'm unsure if the packages use the correct `.babelrc` since there's no `babel-preset-es2015` in the `package.json` which should be required. Yet the build doesn't fail.

rgbkrk:
Are the packages being built on initial install?

lgeiger:
I just checked, it's using the correct `.babelrc` with the `es2015`.

lgeiger:
Just checked the Electron app still works as expected 🎉

lgeiger:
Unfortunately [`lerna bootstrap --hoist`](https://github.com/lerna/lerna#--hoist-glob) is still **very** slow but with `lerna bootstrap` the Electron app won't work.

One could take a look at which packages need to be hoisted and manually add them with a glob pattern. Though I'm wondering if it would work without, if we move the Electron app inside the `packages/` folder.

rgbkrk:
I do want to move the Electron app inside the packages folder, wasn't sure when we would do it.

lgeiger:
There a two possibilities: Either move everything inside the packages folder or leave `electron-builder` in the root `package.json` and [point it to the right folder](https://github.com/electron-userland/electron-builder/wiki/Options#directories).

PR #1516, opened 2017-03-02
chore(lerna): Only hoist rxjs and immutable

lgeiger:
This will make bootstrapping **a lot** faster 🚀 

I looks like all packages must use the same `rxjs`. I suppose this makes sure that `@nteract/messaging` uses `node_modules/rxjs` instead of `packages/node_modules/rxjs`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "prebuild": "rimraf lib",
-    "lerna": "lerna bootstrap --hoist",
+    "lerna": "lerna bootstrap --hoist rxjs",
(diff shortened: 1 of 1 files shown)


rgbkrk:
It makes bootstrapping faster while filling your disk more. You'll have N copies of React and Immutable for instance.

lgeiger:
The packages (`notebook-preview-demo` may be a exception) all have React and Immutable as peerDependencies so they won't install separately.

lgeiger:
With this PR the size of `packages/` is over 230 MB though 78MB are only from `notebook-preview-demo`. 

That's definitely to much. I'm closing this one.

PR #1529, opened 2017-03-04
chore(cleanup): Remove unused packages

lgeiger:
Fixes https://github.com/nteract/nteract/pull/1528#issuecomment-284170765


rgbkrk:
Ah great, thanks for moving the dependencies out from the electron app that are actually in `packages/`.

PR #1533, opened 2017-03-05
chore(lerna): Add command to build packages

lgeiger:
I could't get a watch command with `lerna run --no-sort` to work. We would probably need to make our own top level build script, eg: https://github.com/facebook/jest/blob/master/scripts/build.js


rgbkrk (approved):

rgbkrk:
Thanks!

lgeiger:
I made a [PR to `babel-cli`](https://github.com/babel/babel/pull/5421) which would allow us to transpile our packages from one place with a simple option and add file watching without using gulp or maintaining a build script.

Issue #1526, opened 2017-03-04
Create a Table Transform for application/vnd.dataresource+json


rgbkrk:
Create a transform (within `packages/`) to accompany https://github.com/pandas-dev/pandas/pull/14904. Mimetype is `application/vnd.dataresource+json`.

Useful links:

* Top level: http://specs.frictionlessdata.io/data-resource/
  * Expectation of row oriented data and schema
* Table Schema: http://specs.frictionlessdata.io/table-schema/
* https://github.com/frictionlessdata/datapackage-js/issues/65

Round one for this component can be a simple table. Since we'll have the raw JSON we'll be able to do client side sorting (for the data available), filtering, formatting, and even light visualizations.

lgeiger:
Closed via #1534

PR #1534, opened 2017-03-05
Data resource renderer

rgbkrk:
An initial implementation of #1526 in support of the new data resource mimetype supported in pandas (https://github.com/pandas-dev/pandas/pull/14904), `application/vnd.dataresource+json`.

Scrollable Table

This is the table renderer from https://github.com/gnestor/jupyterlab_table, with a few modifications.

----------------------------

To try this out for yourself, you'll need to install pandas from their `master` branch and set this option in your running session:

```python
import pandas as pd
pd.options.display.html.table_schema = True
```

After that, all DataFrames will be returned with both `text/html` and `application/vnd.dataresource+json` representations.

----------------------------

## TODO

* [x] Render data resource
* [x] Margin between the last column and the scroll bar
* [x] Ability to scroll across many more columns rather than collapse them.
* [x] Synchronize scrolling across indices and columns like in https://bvaughn.github.io/react-virtualized/#/components/ScrollSync
* [x] ~Fix the sorting (strings do not sort properly here)~ (since been removed)

screen shot 2017-03-10 at 10 04 27 am

/cc @gnestor @TomAugspurger @captainsafia


rgbkrk:
Since I couldn't help myself, here's it being used in `notebook-preview` for a statically rendered table (like on commuter, nbviewer):

static-notebook

rgbkrk:
Now if only we had some built in table -> vega React component here as well.

gnestor:
It looks like the default max rows is set to 59, so users will need to set `pd.options.display.max_rows` in addition to `pd.options.display.html.table_schema = True`...

rgbkrk:
Rebased to get up to date dependencies.

rgbkrk:
Investigating https://bvaughn.github.io/react-virtualized/#/components/MultiGrid next.

gnestor:
I updated the `MultiGrid` but I'm having issue testing it in nteract... kernel is hanging. I might need to `git clean -xfd` and rebuild.

gnestor:
Ok, it's working now! I did my best to work with the existing styles but it's hacky because there are no row or column divs, just cell divs, so it's a mix of class names and inline styles. The `VirtualizedTable` will be much more straight-forward to style. I will save that for tomorrow...

The `CellMeasurer` doesn't work perfectly as you can see, and the `ScrollSync` performance leaves a lot to be desired, so maybe react-virtualized is not the best option.

rgbkrk:
Sweet! I tried it out locally and see what you mean quite well.

I have a feeling that we need to move the functions inside of CellMeasurer out and focus on making the (table) cell rendering be very fast functions. Right now there's a lot of computation performed that could be cached and updated only when the props have changed.

rgbkrk:
Oh ha, and I just noticed `CellMeasurerCache`.

gnestor:
Ya, we could try it without `CellMeasurer`...

rgbkrk:
An alternative that we use at work is ag-grid.

packages/transform-dataresource/src/virtualized-grid.js
```diff
@@ -0,0 +1,119 @@
...
+          ...this.props.theme === 'nteract' && rowIndex % 2 === 0 && !(rowIndex === 0 || columnIndex === 0)
+            ? { background: 'rgba(255,255,255,0.075)' }
+            : {},
+          borderCollapse: 'collapse',
+          boxSizing: 'border-box'
+        }}
+      >
+        {
+            [
+              this.schema.fields.reduce(
```
rgbkrk: I'm particularly interested in if this can be stored whenever `this.props.schema` changes (including the first one). This reduction is called quite a bit in my cursory look at this.

packages/transform-dataresource/src/virtualized-grid.js
```diff
@@ -0,0 +1,119 @@
...
+                (result, field) => ({ ...result, [field.name]: field.name }),
+                {}
+              ),
+              ...this.data
```
rgbkrk: In fact, this overall structure should become normalized and then used as:

```
virtualGridData[rowIndex][this.schema.fields[columnIndex].name]
```

gnestor:
@rgbkrk Good suggestion but `ScrollSync` is still pretty sluggish. I'll try removing `CellMeasurer` just to see...

gnestor:
`ScrollSync` perf is much better after removing `CellMeasurer` (which didn't really work anyway) 👍

I think we can implement our own cell measurer algorithm to determine the ideal column width for each column.

gnestor:
> * [ ] Figure out _something_ better with height

We can dynamically set the height based on row height, number of rows, etc.

rgbkrk:
This is definitely really slick now. A few notes on things we should either fix up or think about long term:

* Header should be bold or at least stand out compared to cells
* Indices should also stand out as distinct (either of these could use thicker borders instead)
* Sometimes the indices aren't lining up with the rows when I scroll (maybe the measurement is off?)

screen shot 2017-03-09 at 4 50 34 pm

Some optional things:

* Right align numbers
* Left align strings

rgbkrk:
I just noticed the styling is top notch in the "nteract" theme. I'm just missing theming in the light version I suppose.

screen shot 2017-03-09 at 4 53 01 pm

rgbkrk:
Super sweet stuff here, thanks @gnestor and @lgeiger!

gnestor:
> I think we can implement our own cell measurer algorithm to determine the ideal column width for each column.

Done!

rgbkrk:
Awesome!!!!

I modified the notebook just now to include when there are multiple primary keys so we can pin them to the left.

rgbkrk:
Let's talk about height now!

> We can dynamically set the height based on row height, number of rows, etc.

nteract, like the jupyter notebook, has a way to expand the size of the cell to fit its contents. What I'd love to see is that the virtualized table is able to grow to the size of the window or be collapsed. Not sure how to achieve that though. ¯\\\_(ツ)\_/¯

gnestor:
Can we pass the height as a prop to `DataResourceTransform`?

rgbkrk:
We're not storing or passing height through currently. We do have available if they've expanded the outputs or not. When it's not in expanded view we could make it N pixels high and if it's expanded make it `window.innerHeight`? I guess we'd need to be [tracking the change in height during a `window`'s `resize` event which means we'd need to pass size through on the state](http://stackoverflow.com/questions/36862334/get-viewport-window-height-in-reactjs)...

gnestor:
Ok, so we need an `expanded` prop and we need to decide upon sensible heights for expanded and collapsed. I think the notebook's collapsed height 336. Expanded height probably shouldn't be full height because then the user can't see the source at the same time? Or maybe that's the whole point of collapsed?

Regarding window size, we can [listen to window resize events from the component](https://github.com/gnestor/jupyterlab_plotly/blob/master/component/index.js#L46), but I assume that most transforms will also want to know about the window size, so it might make more sense to listen at the top and pass it down as props to transforms.

rgbkrk:
> it might make more sense to listen at the top and pass it down as props to transforms

I agree, we should make it top level somewhere, probably on the `Notebook` component.

rgbkrk:
> Expanded height probably shouldn't be full height because then the user can't see the source at the same time? Or maybe that's the whole point of collapsed?

That's the point of collapsed, though you're right about possibly not taking up the full height.

rgbkrk:
Since this passes tests and works pretty well, I think we can move the checklist to a new issue and merge this.

lgeiger (approved):
This looks really cool! 🎉 
@rgbkrk @gnestor Thanks for the work!

There are still a few unnecessary comments and a empty file (`packages/transform-dataresource/src/index.css`) which could be removed.

Other than that I think this is good to go. The rest we can fix in #1572.

gnestor:
@lgeiger Thanks! I just cleaned up that file and some comments. I'm going to leave the commented-out props in there for now as they sort of document the react-virtualized API. Once this transform is solid, we can remove...

lgeiger:
@gnestor Thanks! Do you think this is good to merge?

gnestor:
As long as it passes! 👍

rgbkrk:
Woo! What a big feature.

PR #1535, opened 2017-03-06
chore(lerna): Add watch command for lerna packages

lgeiger:
Yet another lerna PR 😄 

Watching for all packages still doesn't work reliably with this command, but it can be used to watch a single package which is enough for most cases:
```bash
npm run build:packages:watch -- --scope @nteract/transforms
```


jdetle:
Is it just me or are the comment and merge buttons a different color now?

lgeiger:
Yep, the whole color palette got more colorful.

PR #1540, opened 2017-03-06
chore(pkg): bump ansi-to-react

rgbkrk:
Linkify is breaking the rendering of many notebooks, so I want to upgrade and turn it off for us for now.


peggyrayzis (approved):
Thanks! 🎉

PR #1555, opened 2017-03-16
fix(save): Fix notification bug.

jdetle:
This fixes a bug where saving triggered a success notification on cancel. I _think_ this might need a refactor to just bring the notification system to the save epic. Consider this a temp fix.


jdetle:
Closing this in favor of alternate fixes in the middleware.

PR #1562, opened 2017-03-17
Save Notifications Change

### LINKED ISSUE #1548: "Save successfull" even though "cancel" or "esc" was pressed. Minor but still strange.
ENVIRONMENT DETAILS:
- Win7 v6.1.7601 SP1 Build 7601
- nteract v0.1.0

REPRODUCE:
1. open nteract 
2. do no more
3. click on the menu "file" --> "save" / or use ctrl+s
4. click on "cancel" or hit "esc"
5. "Save successfull!" message shown even though nothing was saved.
5a. after that the ".....new kernel needs to be launced" dialog comes up.

jdetle:
Closes #1548.

Added
---
- Reject promise if filename undefined
- Notify in save to couple with logic.
- Change error action such that it will be caught by middleware

### DIFF
--- a/src/notebook/epics/saving.js
+++ b/src/notebook/epics/saving.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import { ActionsObservable } from 'redux-observable'; /* eslint-disable no-unused-vars */
 import { writeFileObservable } from '../../utils/fs';
 import {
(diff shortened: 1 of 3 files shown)


rgbkrk:
Thank you!

PR #1587, opened 2017-03-24
fix: pin codemirror 5.24.x, workaround scroll prob

rgbkrk:
Pinning CodeMirror for now until we figure out what caused this, what we need to handle for codemirror 5.25.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     ]
   },
   "dependencies": {
-    "codemirror": "^5.23.0",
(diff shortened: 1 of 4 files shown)


rgbkrk:
Background: our editor component is causing the entire notebook to scroll to the bottom. This affected the desktop app and `notebook-preview` (which in turn affected commuter).

peggyrayzis (approved):
Good catch 👍 glad we're able to fix this!

PR #1597, opened 2017-03-28
feat: raw cell

rgbkrk:
Some initial raw cell support so that rendering doesn't break (currently raw cells completely break a notebook render).


jdetle (approved):

jdetle:
Looks good to me! I'm going to rebase against master now for the cell refactor.

PR #1599, opened 2017-03-29
fix(logger): Change syntax to fix breaking change.

jdetle:
Fixes a breaking change to redux-logger.


rgbkrk:
Thanks for finding and fixing this John!

Issue #1470, opened 2017-02-15
Which cells are scheduled for evaluation?


ajasja:
If I go to Run all cells in a jupyter notebook, the cells that are scheduled for evaluation get an asterix (*).
This is not completely true in nteract
![image](https://cloud.githubusercontent.com/assets/19926/22976142/be3e98b4-f389-11e6-89be-e38aac923732.png)

This also happens if I shift+Enter through the cells:
![image](https://cloud.githubusercontent.com/assets/19926/22976162/dd5c446c-f389-11e6-8858-4ea5bb1dbe69.png)

Here is the notebook demo 
[test-run-all-1.zip](https://github.com/nteract/nteract/files/777140/test-run-all-1.zip)

ajasja:
PS: hope I'm not submitting to many issues, I do try to check if these are duplicates, but finding the right keywords is sometimes tricky.

rgbkrk:
Keep submitting issues, it's good to have them outlined. I think this one was mentioned in an issue yet didn't have a good issue all on its own. I'm glad you've brought this up, I face this problem too.

dhirschfeld:
I have noticed this behaviour too and was going to open an issue but @ajasja beat me to it!

lgeiger:
This was fixed by #1608

PR #1608, opened 2017-04-03
fix(document): cell-eval-status

jkornblum:
This PR addresses issue #1470.

Previously when you ran multiple cells with one command (e.g. Run All, Run All Below) then the `[*]`status would not show correctly for all cells. This pr corrects that so that the `[*]` status will show for all cells to be evaluated.

The only update with this commit is to add the `id` in the `document.cleanCellTransient()` method where the cell in the transient.cellMap is reset. I believe this was the intended functionality, maybe someone who knows more can comment, thanks.

### DIFF
--- a/src/notebook/reducers/document.js
+++ b/src/notebook/reducers/document.js
@@ -1,16 +1,13 @@
 /* @flow */
 
-import * as Immutable from 'immutable';
-import * as uuid from 'uuid';
+import * as Immutable from "immutable";
+import * as uuid from "uuid";
 
-import * as constants from '../constants';
+import * as constants from "../constants";
 
-import type {
-  LanguageInfoMetadata,
-  KernelInfo,
-} from '../records';
+import type { LanguageInfoMetadata, KernelInfo } from "../records";
 
-import { DocumentRecord } from '../records';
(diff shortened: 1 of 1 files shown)


src/notebook/reducers/document.js
```diff
@@ -99,54 +102,61 @@ export function reduceOutputs(outputs: ImmutableOutputs = Immutable.List(), outp
...
-      keyPaths.filter((keyPath: KeyPath) => keyPath.get(2) !== id)
+  return state
+    .updateIn(
+      ["transient", "keyPathsForDisplays"],
+      (kpfd: Immutable.Map<string, KeyPaths>) =>
+        kpfd.map((keyPaths: KeyPaths) =>
+          keyPaths.filter((keyPath: KeyPath) => keyPath.get(2) !== id))
     )
-  ).setIn(['transient', 'cellMap'], new Immutable.Map());
+    .setIn(["transient", "cellMap", id], new Immutable.Map());
```
rgbkrk, line 112: Here's the line that actually changed for fellow reviewers.

I'll be making another PR to finish off our transition to prettier.

PR #1609, opened 2017-04-03
Much prettier

rgbkrk:
As noticed in #1608, not everything was run through prettier.


peggyrayzis (approved):
So pretty 😍 thanks for cleaning all this up!

PR #1613, opened 2017-04-05
chore(webpack): Upgrade current build to webpack 2

peggyrayzis:
Kicking off the first of what will be many more webpack 2 PRs to optimize our build. 🎉 @rgbkrk's suggestion to break this up into chunks instead of doing it all at once is definitely the way to go - it'll be much easier to review and debug.

This PR upgrades our current setup to webpack 2. No new plugins yet 😀


rgbkrk (approved):
Thank you!

PR #1615, opened 2017-04-05
ux(app): Make celltoolbar transparent to see-through text.

Carreau:
Unless hover'd in which case restaure 100% opacity.
Making it fully opaque hide text when editing markdown

see-through

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -177,6 +177,13 @@ body
 {
     display: inline-box;
     background-color: var(--toolbar-bg);
+    opacity: 0.4;
+    transition: opacity 0.4s;
+}
+
+.cell-toolbar:hover
+{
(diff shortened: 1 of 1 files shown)


rgbkrk:
Thank you!

rgbkrk:
The other part about the cell toolbar that irks me is when I can't click the text in the cells. Makes me think we should have it placed higher. What do you think?

Carreau:
i was thinking of smth like chrome hints when you hover on URL that move
away when you approach... But then you can't click right ? I think it's
fine as is, there is only a small area we can't click. And it's keyboard
accessible. One thing I'd like to see is investigating a global vertical
toolbar which is touchscreen friendly. Especially tablets friendly for
thumbs.

PR #1616, opened 2017-04-06
feat(webpack): Added vendor bundle for third party modules

peggyrayzis:
Added a vendor bundle for our third-party modules. Soon, I'm going to split these chunks out even more and add them dynamically w/ the HTMLWebpackPlugin. I also took out `moment` and replaced it with `date-fns`, which is a smaller library and better for tree shaking.

More improvements coming 🔜!


rgbkrk (approved):
Wahoo!

PR #1617, opened 2017-04-06
Consistency: All toolbar elements childrens are divs.

Carreau:
Easier for Css styling, you "just" need to target divs instead of
span/button/div.

See https://github.com/nteract/mocks/issues/15 as one example of use.


Carreau:
Note: review with [`?w=1`](https://github.com/nteract/nteract/pull/1617/files?w=1) make things easier to see

rgbkrk (approved):

rgbkrk:
Thanks!

PR #1627, opened 2017-04-07
fix(display-area): fix scrollbar still visible after expansion

jkornblum:
There was an issue when you expanded a cell output and the `overflowY` property was previously set to scroll that it wouldn't change the `overflowY` property correctly to remove the scrollbar.

This PR adds unit tests to verify the `overflowY` and `height` style change logic.

This PR fixes two existing unit tests in `display-spec.js` which were previously false passes.


rgbkrk (approved):

rgbkrk:
Thank you, this is looking good.

PR #1637, opened 2017-04-10
fix(toolbar): Inline divs.

jdetle:
Fixes a vertical toolbar issue.

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -177,6 +177,10 @@ body
     padding: 0px 0px 0px 50px;
 }
 
+.cell-toolbar > div {
+  display: inline-block;
+}
+
 .cell-toolbar
 {
     background-color: var(--toolbar-bg);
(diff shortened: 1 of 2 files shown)


static/styles/main.css
```diff
@@ -206,6 +205,10 @@ body
 
 }
 
+.inline-div {
```
rgbkrk: Instead of creating a new class, target this with:

```
.cell-toolbar > div
```

--------------------

If I had to guess, @carreau set this up so we could do media queries for touch screens to show a toolbar like in this mock: https://github.com/nteract/mocks/issues/15

jdetle: I love that mock, but until we figure out how to implement it I like the horizontal toolbar. I'm hoping you agree!

rgbkrk: I agree, I prefer the horizontal toolbar on my mac.

rgbkrk (approved):

Carreau:
Oops, yes. Sorry about this.

Carreau:
That's actually just a mistake when I uniformized the dom structure. didn't meant for it to change layout.

PR #1649, opened 2017-04-14
[WIP] feat: git init

alexandercbooth:
# Starting integration with git
The purpose of this PR is to begin integration of git within nteract. Specifically to display the current git branch in the status bar.
git mov

## Git Node Module
The two main choices I found were atom's [git-utils](https://github.com/atom/git-utils) and [nodegit](https://github.com/nodegit/nodegit). Nodegit initially seemed like the better option, but after struggling to make it work with electron I realized that Atom had stopped using it for the same reason and switched back to git-utils (so I did the same). 

## Issues
- When I run the tests, I get, 
```bash
Error: Module version mismatch. Expected 48, got 53.
```
I think this is related libtgit being compiled, but I don't quite know how to fix it.
- having trouble sizing the octicon (which makes the status bar slightly taller) ¯\_(ツ)_/¯

## Future
Next steps would be a follow up PR to
- display modified status
- make commit from menu

Any thoughts would be appreciated.

Thanks!


rgbkrk:
Normally electron builder should handle rebuilding the native module for you, @lgeiger would know best there. You may also need to add the module to our webpack config (look where zeromq is listed).

rgbkrk:
LOL, `amsters`, love those hamsters.

alexandercbooth:
>You may also need to add the module to our webpack config (look where zeromq is listed).

OK, I will play around with this.

>LOL, amsters, love those hamsters.

Haha, thought you would like that 😄.

rgbkrk:
oh, for tests, you'll probably want to add it to our current mocking setup -- I think we have mock-requires for a lot of those modules

lgeiger:
Oh no, a native module... Just kidding 😉 

> `Error: Module version mismatch. Expected 48, got 53.`

This error is due to the fact that `electron-builder` rebuilds the native module against Electron headers (`53`) to be able to use it. But the unit test run on your local Node.js version (`48`). 

Probably the easiest option is to avoid using `git-utils` in the tests and mock the behavior instead. Otherwise we would need to rebuild the module against the correct version.

Since this adds a new native dependency we would need to revert #1268 (except for the `package.json` changes). Or we ask the `git-utils` team if they would be interested in providing prebuilt binaries via [`prebuild`](https://github.com/mafintosh/prebuild) .

alexandercbooth:
haha I know, I don't like it either 😭. 

>Or we ask the git-utils team if they would be interested in providing prebuilt binaries via prebuild

I like this idea. If this is something they are not interested in doing, is it something we could do on our own (or is that even reasonable?). I want to avoid reverting #1268, so if not maybe this is not the right approach for git. 

Would it instead be appropriate (or even preferable), to write a custom script that accesses the user's own git client (and disables the functionality if they don't have git installed)? 

Thanks Lukas!

rgbkrk:
> Would it instead be appropriate (or even preferable), to write a custom script that accesses the user's own git client (and disables the functionality if they don't have git installed)?

I would not enjoy maintaining scripts cross platform (both running and parsing), I'd rather work with `git-utils`.

> we ask the git-utils team if they would be interested in providing prebuilt binaries via prebuild

certainly worth an ask, we could show them how the zeromq one works 😄

rgbkrk:
Ready to come back to this @alexandercbooth?

rgbkrk:
I'm going to go ahead and close this, we can pair on it later or feel free to reopen when you're ready.

PR #1665, opened 2017-04-23
bug(tooltip): stop tooltip from showing if introspection request is empty

alexandercbooth:
I noticed that even if the tooltip has nothing to show, a blank box is still created. This prevents that from happening.


jdetle:
Thanks for the catch @alexandercbooth!

alexandercbooth:
My pleasure!

PR #1668, opened 2017-04-25
chore(package): Update electronVersion to match dependency

lgeiger:
Follow up on #1667 and #1654


alexandercbooth:
Thanks Lukas!

PR #1669, opened 2017-04-28
fix(wrapper): adds support for tooltip on windows

jkornblum:
Adds shortcut `ctrl-.` to show the tooltip

(it's showing a ton of changes again, this is prettier?, my only edit is add line 234)


alexandercbooth:
I think you're right, that is prettier.

Good catch, thanks @jkornblum!

PR #1670, opened 2017-04-29
Use Yarn

lgeiger:
This cuts the initial install times in half:
## `npm`
Fresh `npm i`: 
```
real    6m28.978s
user    5m51.561s
sys     0m46.525s
```
Second `npm i`:
```
real    1m59.847s
user    2m48.043s
sys     0m12.689s
```
## `yarn`
Fresh `yarn`:
```
real    2m55.251s
user    3m20.890s
sys     0m46.197s
```

Second `yarn`:
```
real    1m54.571s
user    2m30.263s
sys     0m27.417s
```

Note that the metrics include the lerna, webpack and babel build steps so the improvement on the second build is quite small.


rgbkrk:
Sweet stuff. How should we hack on individual packages while working on the main app? Same build steps as before?

rgbkrk:
At least two travis failures right now:

* https://travis-ci.org/nteract/nteract/jobs/227225050
* https://travis-ci.org/nteract/nteract/jobs/227225049

Looks like the same one, while building native modules.

lgeiger:
> How should we hack on individual packages while working on the main app? Same build steps as before?

In theory lerna only uses a different client to install dependencies. I should test it though.

I'll fix up the CI failures tomorrow.

lgeiger:
> How should we hack on individual packages while working on the main app? Same build steps as before?

I just checked this works the exactly the same.

PR #1636, opened 2017-04-09
[WIP] feat(menu): expand cells on export pdf

jkornblum:
Addresses issue #1610.

It will expand collapsed (default state) cells before export to PDF via modifying CSS. After export is
complete it will restore the cells to as before the export PDF.

Print to PDF still respects if you have the output hidden.

Issue #1610, opened 2017-04-03
feature: expand all cells for PDF print


jkornblum:
I see for each cell there is a cell menu option called `Toggle Expanded Ouput`. When I have a notebook and I'd like to print to PDF, I generally want to expand all my cells before printing. I've gone ahead and wired the functionality up, but before I pull request

What would the preferred option/behavior be?
1. Dialog box asking if the user would like to expand all cells before print (Yes|No). Cells are returned to their original state after print.
2. Menu option, Cell > Expand All Cells, Cell > Collapse All Cells.
3. Other suggestions?

I prefer 1 because I like to work with collapsed cells but just want to expand them for printing.
What are everyone's thoughts?

rgbkrk:
I think that expanded by default on printing makes complete sense. We do the same thing for notebook viewer-like things. I realize this means it's not as easy as a stylesheet with that change.

rgbkrk:
Side / separate note -- the hidden inputs/outputs settings should remain consistent while printing though (then you can have reports backed by code that don't show the code).

rgbkrk:
Is this one done now with the current PDF code?

willingc:
The hidden cells are left hidden when creating a PDF. If `Cell -> Unhide Input and Output in all Cells` is selected before saving a PDF the previously hidden cells are displayed and output in the PDF.

Closing as resolved

PR #1671, opened 2017-04-30
feat(menu): expand collapsed cells on export PDF

jkornblum:
Opening new PR, git branch on the last one #1636 blew up.

Addresses issue #1610.

It will expand collapsed (default state) cells before export to PDF via modifying CSS. After export is
complete it will restore the cells to as before the export PDF.

Print to PDF still respects if you have the output hidden.

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -10,21 +10,22 @@ import { load, newNotebook } from "./epics/loading";
 import { loadConfig } from "./epics/config";
 
 import {
-  executeCell,
-  clearOutputs,
-  newKernel,
-  killKernel,
-  interruptKernel,
-  copyCell,
-  cutCell,
-  pasteCell,
-  createCellAfter,
-  setGithubToken,
   changeInputVisibility,
-  setTheme,
-  setCursorBlink,
+  clearOutputs,
+  copyCell,
+  createCellAfter,
+  cutCell,
+  executeCell,
+  interruptKernel,
+  killKernel,
+  newKernel,
+  pasteCell,
   save,
-  saveAs
+  saveAs,
+  setCursorBlink,
+  setGithubToken,
+  setTheme,
+  toggleOutputExpansion
 } from "./actions";
 
 import { defaultPathFallback, cwdKernelFallback } from "./path";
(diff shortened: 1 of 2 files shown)


jdetle:
Looking forward to checking this out! Thanks @jkornblum.

src/notebook/menu.js
```diff
@@ -281,27 +282,56 @@ export function dispatchNewNotebook(store, event, kernelSpec) {
...
+    cellID => cellMap.getIn([cellID, "metadata", "outputHidden"]) === false
+  );
+
+  // Expand unexpanded cells
+  unexpandedCells.map(cellID => store.dispatch(toggleOutputExpansion(cellID)));
+
+  remote.getCurrentWindow().webContents.printToPDF({
+    printBackground: true
+  }, (error, data) => {
+    if (error) throw error;
```
rgbkrk, line 312: There's nothing available to catch this error -- it should likely be dispatched so that the user can see it failed. Otherwise you're getting into undefined handling by electron.

rgbkrk: Then again, I see that's the default from their docs. For our case though, we'd still want to log it, which means dispatching an error action.

src/notebook/menu.js
```diff
@@ -281,27 +282,56 @@ export function dispatchNewNotebook(store, event, kernelSpec) {
...
+  const state = store.getState();
+  const notebook = state.document.get("notebook");
+  const cellMap = notebook.get("cellMap");
+  const cellOrder = notebook.get("cellOrder");
+  const unexpandedCells = cellOrder.filter(
+    cellID => cellMap.getIn([cellID, "metadata", "outputHidden"]) === false
+  );
+
+  // Expand unexpanded cells
+  unexpandedCells.map(cellID => store.dispatch(toggleOutputExpansion(cellID)));
```
rgbkrk, line 307: The fact that we're changing the document and putting it back asynchronously (depending on how long it takes to printToPDF) has the smell of race conditions. The only way I can think of to make this better is to make a separate document that gets printed (using the same state tree). 
¯\\\_(ツ)\_/¯

At any rate, what you've done here is great. Thanks for doing this.

jkornblum: Yeah I agree, I looked into do that but didn't know a quick solution. I can try to revisit in the future. There are a few other print to PDF things I'd like to work on.

PR #1672, opened 2017-04-30
fix github publish

jdetle:
Will get tests to pass, this is a start.

### DIFF
--- a/src/notebook/epics/github-publish.js
+++ b/src/notebook/epics/github-publish.js
@@ -55,7 +55,7 @@ export function createGistCallback(observer, filename, notificationSystem) {
       observer.complete();
       return;
     }
-    const gistID = response.id;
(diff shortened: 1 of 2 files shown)


rgbkrk:
The `data` key came from an update to `github`?

rgbkrk (approved):

jdetle:
Yes: https://github.com/mikedeboer/node-github/commit/ceeefb2c478ca4302e1fe60dece10bc40075eaaf

rgbkrk:
Perhaps we should pin this more tightly (only patch releases with `~`) since we don't touch this section of the code all that often. Then again, I guess it forces us to keep moving it forward.

jdetle:
I wasn't aware of that as an option, lets do it.

PR #1676, opened 2017-05-02
chore(📦): squash react deprecations all over

rgbkrk:
Trying to bring us out of deprecation warnings. I think some of the packages we rely on are still triggering warnings in the dev console. I'd like to get those fixed up, this works for getting us inline though. :smile:

As part of this, I brought in codemirror react since it's not being maintained all that actively and we already override a bunch of the events it registers. This should give us more control over something that's fairly central to us.

👨🏻‍🏭


rgbkrk:
This will likely conflict with your PR @carreau, yet should provide much easier ways of overriding behavior here.

packages/editor/src/codemirror.js
```diff
@@ -0,0 +1,137 @@
...
+        <textarea
+          ref="textarea"
+          name={this.props.path}
+          defaultValue={this.props.value}
+          autoComplete="off"
+        />
+      </div>
+    );
+  }
+});
```
rgbkrk, line 135: I'm really happy we're bringing this in because I'll be able to do readonly mode of codemirror for notebook preview :soon: after.

There's a lot we can clean up from this original source.

peggyrayzis: That's an awesome use case for bringing it in! 🎉

packages/editor/src/wrapper.js
```diff
@@ -126,169 +123,165 @@ const CodeMirrorWrapper: CodeMirrorHOC = (EditorView, customOptions = null) =>
...
-          );
-          const x = document.getElementById("cl");
-          if (x != null && body != null) {
-            const pos = x.getBoundingClientRect();
-            body.appendChild(x);
-            x.style.top = pos.top + "px";
-          }
-        });
-      }
+  tips(editor: Object): void {
```
peggyrayzis, line 159: We should try to refactor this function so it's more in line with React paradigms.

peggyrayzis (approved):
Thanks for tackling this! Looks like we'll be in good shape when Fiber is officially released 👍

packages/editor/src/wrapper.js
rgbkrk: Yeahhhhhh definitely.

PR #1689, opened 2017-05-08
chore(build): Fix lerna warnings:

lgeiger:
* remove unnecessary devDependencies
* upgrade normalize.css to match root version

Get rid of the following warnings:
```
lerna WARN EHOIST_ROOT_VERSION The repository root depends on jest@^20.0.0, which differs from the more common jest@^19.0.2.
lerna WARN EHOIST_PKG_VERSION "@nteract/transform-dataresource" package depends on jest@^19.0.2, which differs from the hoisted jest@^20.0.0.
lerna WARN EHOIST_ROOT_VERSION The repository root depends on normalize.css@^7.0.0, which differs from the more common normalize.css@^6.0.0.
lerna WARN EHOIST_PKG_VERSION "@nteract/notebook-preview-demo" package depends on normalize.css@^6.0.0, which differs from the hoisted normalize.css@^7.0.0.
```

### DIFF
--- a/packages/notebook-preview-demo/package.json
+++ b/packages/notebook-preview-demo/package.json
@@ -14,7 +14,7 @@
     "@nteract/transform-plotly": "^1.2.0",
     "@nteract/transforms": "^1.0.11",
     "codemirror": "~5.24.0",
-    "normalize.css": "^6.0.0",
+    "normalize.css": "^7.0.0",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
     "react-router": "^4.0.0-beta.5",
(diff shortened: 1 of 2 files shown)


rgbkrk (approved):

rgbkrk:
The tests still get run on the data resource package if you take out the test line? When rebuilding a single package can we still do it from within the package for? Do we have to have jest, etc. all installed globally?

lgeiger:
> The tests still get run on the data resource package if you take out the test line? When rebuilding a single package can we still do it from within the package for? 

Yes, all other packages don't have `devDependencies` either.

> Do we have to have jest, etc. all installed globally?

No, lerna should take care of that.

PR #1690, opened 2017-05-10
fix(flow): Silence flow warnings on newer flow versions

lgeiger:
Noticed while experimenting with commutable in Hydrogen which uses a newer flow version.

Unfortunately Immutable doesn't allow us to properly type everything 😞


rgbkrk (approved):
Thanks.

PR #1692, opened 2017-05-11
Greenkeeper: Don't ignore flow

lgeiger:


rgbkrk:
Thanks for bringing this back, when we were waiting on fbjs we had to hold back for a bit.

PR #1518, opened 2017-03-02
[WIP] start creating notebook metadata schema

rgbkrk:
As stated in #1465, I want to validate our own metadata and provide the schemas for anyone who wants to use them. This is a work in progress. Need to figure out what will be the most idiomatic API here and tighten up the initial JSON schema.

PR #1697, opened 2017-05-17
feat(nbp): respect hide_input from jupyter-contrib

rgbkrk:
nteract desktop uses `inputHidden` while [hide input](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/tree/master/src/jupyter_contrib_nbextensions/nbextensions/hide_input) uses `hide_input`. We'll at least acknowledge it within notebook preview though we should likely align desktop and nbextensions where they make sense.

/cc @jcb91

### DIFF
--- a/packages/notebook-preview/src/code-cell.js
+++ b/packages/notebook-preview/src/code-cell.js
@@ -35,7 +35,10 @@ class CodeCell extends React.PureComponent {
   }
 
   isInputHidden(): any {
-    return this.props.cell.getIn(["metadata", "inputHidden"]);
+    return (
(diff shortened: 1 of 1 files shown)


jcb91:
Thanks for the heads up @rgbkrk. I think your 'hidden' wording is a better expression, so perhaps we should migrate to that as well...

lgeiger (approved):
Thanks @rgbkrk

rgbkrk:
> I think your 'hidden' wording is a better expression, so perhaps we should migrate to that as well...

As long as we're consistent I'm happy. I've been hoping to standardize the metadata we're using in #1518, but I haven't been able to come back to that PR again for a bit.

PR #1699, opened 2017-05-18
feat(stream): Escape carriage return in output reducer

lgeiger:
- This reduces notebook file sizes if it contains a lot of streaming output. Eg. a notebook with the output of `!npm i electron` now only is 2kB big. Before it had 49kB.
- This should also improve rendering performance of big text output

Note: `escapeCarriageReturnSafe` won't fully escape the string. It only escapes the parts of the string that don't depend on future output.

### DIFF
--- a/src/notebook/reducers/document.js
+++ b/src/notebook/reducers/document.js
@@ -3,6 +3,8 @@
 import * as Immutable from "immutable";
 import * as uuid from "uuid";
 
+import { escapeCarriageReturnSafe } from "escape-carriage";
+
 import * as constants from "../constants";
 
 import type { LanguageInfoMetadata, KernelInfo } from "../records";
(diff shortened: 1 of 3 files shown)


src/notebook/reducers/document.js
```diff
@@ -73,7 +75,7 @@ export function reduceOutputs(
 
   function appendText(text: string): string {
     if (typeof streamOutput.text === "string") {
-      return text + streamOutput.text;
+      return escapeCarriageReturnSafe(text + streamOutput.text);
```
rgbkrk, line 78: excellent, this is exactly what I was hoping for.

We'll also set up ansi-to-react to not handle carriage returns inline I assume?

lgeiger: I'm not 100% sure about how to do this. `escapeCarriageReturnSafe` only escapes part of the string and possibly leaves a few carriage returns at the end where it can't escape without knowing the future output (see my note).

rgbkrk: If you view it as a terminal, the terminal keeps track of its internal state and can get carriage returns without a newline while waiting for the newline character to show up.

lgeiger: That's what escapeCarriageReturnSafe tries to emulate. But without storing the cursor position it may not fully escape the last line since otherwise we would loose information.

PR #1710, opened 2017-05-23
[DON'T MERGE] Prototype on how to possibly store parsed ANSI

lgeiger:
This is intended to outline a fairly easy way to improve ANSI rendering performance without implementing a terminal like described [here](https://github.com/nteract/meeting-minutes/blob/master/2017/05-22.md#ansi-handling-in-nteract--hydrogen). The advantage over a terminal is mainly that it's easier to implement since we only need to copy paste existing code 😄 .

This would be the next steps:
- Add `@nteract/ansi-parse` which exports [`ansiToInlineStyle`](https://github.com/nteract/ansi-to-react/blob/master/src/index.js#L7-L40)
- Add new transform using [`inlineBundleToReact`](https://github.com/nteract/ansi-to-react/blob/master/src/index.js#L65-L70) to render the json from `@nteract/ansi-parse`. This would accept a new mime type like `application/ansi+json` or `application/nteract.ansi+json`.
- Update [`reduceOutput`](https://github.com/nteract/nteract/blob/master/src/notebook/reducers/document.js#L61) to always use `createImmutableOutput`.

### DIFF
--- a/packages/commutable/src/v4.js
+++ b/packages/commutable/src/v4.js
@@ -32,6 +32,7 @@ import type { CellStructure } from "./structures";
 
 const Immutable = require("immutable");
 const appendCell = require("./structures").appendCell;
+const parse = require("@nteract/ansi-parse");
 
 export type ExecutionCount = number | null;
 
@@ -160,6 +161,12 @@ export function cleanMimeAtKey(
   previous: ImmutableMimeBundle,
   key: string
 ): ImmutableMimeBundle {
+  if (key === "text/plain") {
+    previous = previous.set(
+      "application/ansi+json",
+      parse(cleanMimeData(key, mimeBundle[key]))
+    );
+  }
   return previous.set(key, cleanMimeData(key, mimeBundle[key]));
 }
 
@@ -224,7 +231,8 @@ export function createImmutableOutput(output: Output): ImmutableOutput {
       return Immutable.Map({
         output_type: output.output_type,
         name: output.name,
-        text: demultiline(output.text)
+        text: demultiline(output.text),
+        parsed_ansi: parse(demultiline(output.text))
       });
     case "error":
       return Immutable.Map({
@@ -233,7 +241,12 @@ export function createImmutableOutput(output: Output): ImmutableOutput {
         evalue: output.evalue,
         // Note: this is one of the cases where the Array of strings (for traceback)
         // is part of the format, not a multiline string
-        traceback: Immutable.List(output.traceback)
+        traceback: Immutable.List(output.traceback),
+        parsed_ansi: parse(
+          output.traceback
+            ? output.traceback.join("\n")
+            : `${output.ename}: ${output.evalue}`
+        )
       });
     default:
       throw new TypeError(`Output type ${output.output_type} not recognized`);


packages/commutable/src/v4.js
```diff
@@ -160,6 +161,12 @@ export function cleanMimeAtKey(
   previous: ImmutableMimeBundle,
   key: string
 ): ImmutableMimeBundle {
+  if (key === "text/plain") {
+    previous = previous.set(
+      "application/ansi+json",
```
rgbkrk, line 166: This is only intended for at-runtime right? We still want to end up with `text/plain` for the notebook format.

Also... text/plain won't get additional text. New display_data are wholly distinct.

packages/commutable/src/v4.js
```diff
@@ -224,7 +231,8 @@ export function createImmutableOutput(output: Output): ImmutableOutput {
       return Immutable.Map({
         output_type: output.output_type,
         name: output.name,
-        text: demultiline(output.text)
+        text: demultiline(output.text),
+        parsed_ansi: parse(demultiline(output.text))
```
rgbkrk, line 235: Smart.

packages/commutable/src/v4.js
```diff
@@ -233,7 +241,12 @@ export function createImmutableOutput(output: Output): ImmutableOutput {
         evalue: output.evalue,
         // Note: this is one of the cases where the Array of strings (for traceback)
         // is part of the format, not a multiline string
-        traceback: Immutable.List(output.traceback)
+        traceback: Immutable.List(output.traceback),
+        parsed_ansi: parse(
+          output.traceback
+            ? output.traceback.join("\n")
+            : `${output.ename}: ${output.evalue}`
+        )
```
rgbkrk, line 249: Yeah I'm definitely liking this approach, it's handling it as it gets it.

rgbkrk (approved):
I like this approach of including it in the commutable data structure.

We'll probably want to bring `reduceOutputs` into commutable itself too.

rgbkrk:
> Update reduceOutput to always use createImmutableOutput.

Oh right, you already said here that we'd move reduceOutput into here. Silly me.

rgbkrk:
By the way, my approval here means that I think the approach is :cool:.

lgeiger:
@rgbkrk Thanks a lot for the review!

I did some further testing and unfortunately the ANSI parsing plays only a small part of the performance problem 😕 .

I recorded a performance profile of different Ansi components while executing
```python
print("This is tooooooo long :(" * 1000000)
```

- Current **master** took **9.07 s** to render
- A functional component with no ANSI escaping renders in **8.34 s**:
```js
const Ansi = props => <code className={props.className}>{props.children}</code>;
```
- A pure component with no ANSI escaping renders in **8.24 s**:
```js
class Ansi extends React.PureComponent {
  shouldComponentUpdate(nextProps) {
    return (
      this.props.children != nextProps.children ||
      this.props.className != nextProps.className
    );
  }
  render() {
    return <code className={this.props.className}>{this.props.children}</code>;
  }
}
```

The performance gain of precomputing the ANSI escape will increase when handling streaming output but this doesn't solve the rendering problem 😕 .

Please note, I only recorded the profiles once so I don't claim any statistical significance but it's clear to see that removing the ANSI parsing won't fix the rendering performance issues.

I'm happy to upload the recorded profiles if someone wants to investigate further.

rgbkrk:
Now that I'm thinking about this some more, I think we should handle this at the level of the message processing from the stream. We'll be able to handle backpressure further up the stack and only present to React what's ready in a reasonable time.

rgbkrk:
Did you have any more thoughts on this?

lgeiger:
Since this doesn't solve the problem, I think we can close it for now.

PR #1716, opened 2017-05-26
chore(displa-area): Simplify height calculation

lgeiger:
While looking into #1710 I noticed this could be simplified.


rgbkrk:
Just tested this out and it works beautifully. Much thanks for this simplification.

PR #1718, opened 2017-05-26
fix(text-transform): Disable linkify to fix performance issue

lgeiger:
Fixes rendering performance for large text as mentioned in https://github.com/nteract/hydrogen/issues/817

This disables the linkification until we figure out a performant way of doing this.

### DIFF
--- a/packages/transforms/src/text.js
+++ b/packages/transforms/src/text.js
@@ -15,6 +15,6 @@ export default class TextDisplay extends React.Component {
   }
 
   render(): ?React.Element<any> {
-    return <Ansi linkify>{this.props.data}</Ansi>;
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):

rgbkrk:
Definitely in favor here.

lgeiger:
Should we ship a new major version, since this changes behavior? Or would you consider this a patch release?

rgbkrk:
I'd just call this a minor release

lgeiger:
👍  Sounds good to me.

rgbkrk:
If someone wants this explicit behavior they can override the text transform.

PR #1719, opened 2017-05-28
fix(flow): richestMimetype can return undefined

lgeiger:
`richestMimetype` can return `undefined`. This can lead to nasty React errors.

See https://github.com/nteract/hydrogen/pull/746


rgbkrk:
Certainly patch for this one, though since it's part of the other change you'll have the overall minor release.

Issue #1242, opened 2016-11-24
Markdown cells don't render local images


lgeiger:
![bildschirmfoto 2016-11-24 um 22 48 59](https://cloud.githubusercontent.com/assets/13285808/20610330/39c77dd0-b298-11e6-980b-9b61396baf99.png)

~~The same happens with local images~~

rgbkrk:
As for the absolute URL you're using there -- it needs to be without quotes.

For local images, there is a bigger problem. The page is relative to the install location, not the notebook file. We have to change our architecture to handle relative assets.

lgeiger:
Oops You're right. But the issue with local images is still there.

rgbkrk:
Suggestion by @smashwilson: use [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) and set the `href` to the notebook path. This would likely have to occur after we load our other assets.

rgbkrk:
We definitely have to solve this one before the release candidate and have it kicked around during the beta or alpha stages.

jkornblum:
I had looked at this probably a few weeks ago now and never really finished working on it.
I found this library [react-helmet](https://github.com/nfl/react-helmet), MIT licensed. 

> This reusable React component will manage all of your changes to the document head with support for document title, meta, link, style, script, noscript, and base tags.

Which made it simple to edit `<base href="">` dynamically.

rgbkrk:
Definitely still needs a fix and we're in a good spot for where desktop is for this to be tackled.

PR #1720, opened 2017-05-29
[WIP] Base resolution

### LINKED ISSUE #349: Handle ipynb local links directly

rgbkrk:
This makes relative assets load properly on the page. At the same time, it's also made our elements destined for `<head />` go to one `<Head />` component, managed by [react-helmet](https://github.com/nfl/react-helmet).

I've only done some initial testing, and while images and links are displaying properly there are a few things to fix up:

* [x] Our main handler for launching notebooks doesn't handle `file://` links for notebooks right (they get processed in a really weird way...)
* [ ] All assets are shown right after the second React render of the element containing it. Basically, since the text hasn't changed in the source of the markdown, there's no reason to flush a new element out.

I might see if there's a better way to handle this where the literal URL for every notebook is what we load with `loadURL`.

At any rate, demo with a notebook from my desktop:

screen shot 2017-05-29 at 11 50 44 am

The link text at the bottom is a result of me copying the link output and pasting the URL (success!).

When finished, this closes #349, #1242.

### DIFF
--- a/src/notebook/head.js
+++ b/src/notebook/head.js
@@ -40,6 +40,11 @@ export class Head extends React.PureComponent {
   };
 
   render(): ?React.Element<any> {
+    if (true) {
+      // silly me
+      return null;
+    }
+
     return (
       <Helmet>
         <base href={this.props.dirname} />
--- a/src/notebook/head.js
+++ b/src/notebook/head.js
@@ -1,0 +1,62 @@
+/* eslint-disable no-return-assign */
+/* @flow */
+import React from "react";
+import { connect } from "react-redux";
+import { Helmet } from "react-helmet";
+
+const remote = require("electron").remote;
+const BASE_PATH = remote.app.getAppPath();
+
+const path = require("path");
+const url = require("url");
+
+// TODO: Remove after provider refactor finished
+const PropTypes = require("prop-types");
+
+type Props = {
+  dirname: string,
+  theme: string
+};
+
+const mapStateToProps = (state: Object) => ({
+  dirname: url.format({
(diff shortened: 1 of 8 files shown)


rgbkrk:
Woo, local links are working now!

relative-links

There's a bit of pain in initial refresh while the `<base>` tag gets changed up.

lgeiger:
Wow cool stuff 💯

rgbkrk:
Coming back to this one now, rebasing. I'm going to see about closing down a bunch of my open PRs while I get cranking on the web nteract again.

src/notebook/head.js
```diff
@@ -0,0 +1,62 @@
...
+          href={path.join(BASE_PATH, "static/styles/main.css")}
+        />
+        <link
+          rel="stylesheet"
+          href={path.join(
+            BASE_PATH,
+            `static/styles/theme-${this.props.theme}.css`
+          )}
+        />
+      </Helmet>
```
rgbkrk, line 62: Given what I've figured out in https://github.com/rgbkrk/electron-protocol-play, I'm going to move this to the main process, creating a template instead of our current index.html. It's going to be nice and clean. :smile:

rgbkrk: I lied and am making a terrible mess of things. Since I need to clean this up I'll have to visit it again later.

rgbkrk:
I think I'd rather use this PR as a reference if/when I take on implementing it again, so I'm going to close it for now.

PR #1723, opened 2017-05-30
fix(CI): Don't update yarn.lock a second time

lgeiger:
This should prevent repeated updates as happening in #1721 by being more explicit.


rgbkrk:
Thanks.

PR #1724, opened 2017-05-31
chore(pkg): handle upgrades

rgbkrk:
Since we had a bit of a snafu with greenkeeper upgrades and yarn, I went ahead and upgraded a bunch. While I was at it I noticed some glaring warnings from babel preset env about us not using a string for the electron version and ended up noticing we were targeting an older version of electron.


rgbkrk:
Let me get the yarn update in here too.

package.json
```diff
@@ -181,33 +181,33 @@
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-flow-strip-types": "^6.18.0",
     "babel-plugin-transform-object-rest-spread": "^6.20.2",
-    "babel-preset-env": "^1.3.3",
+    "babel-preset-env": "^1.5.1",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
-    "chai": "^3.4.1",
+    "chai": "^4.0.0",
```
lgeiger, line 187: Chai probably needs some extra care

rgbkrk: Fixed it!

lgeiger (approved):

PR #1726, opened 2017-06-03
BREAKING CHANGE: Remove immutable from display-area and transforms

lgeiger:
As mentioned in https://github.com/nteract/hydrogen/pull/772#issuecomment-301544392 it would be really cool for Hydrogen and other frontends not using Redux/Immutable.js if our packages would  use plain objects instead of immutable.js datastructures.

It would also help extremely with proper typings since it's very hard to properly type Immutable.js objects.

I totally understand if we don't accept this PR but I figured I should give it a try anyway 😄

**TODO:**
- [x] Fix flow error in `transforms`
- [x] Document changes in README

### DIFF
--- a/packages/transforms/src/index.js
+++ b/packages/transforms/src/index.js
@@ -1,5 +1,4 @@
 /* @flow */
-import { List as ImmutableList, Map as ImmutableMap } from "immutable";
 
 // Needed for flow
 /* eslint-disable no-unused-vars */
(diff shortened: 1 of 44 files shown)


packages/transforms/src/index.js
```diff
@@ -12,82 +11,76 @@ import JavaScriptDisplay from "./javascript";
...
+  GIFDisplay,
+  PNGDisplay,
+  JPEGDisplay,
+  TextDisplay
+];
+
+export const standardTransforms: Transforms = {};
+
+tfs.forEach(transform => {
+  standardTransforms[transform.MIMETYPE] = transform;
```
lgeiger, line 40: Flow doesn't recognizes our transforms to be of type [`Transform`](https://github.com/nteract/nteract/pull/1726/files#diff-f5db81f8e359881e75a24b529c3c4131R17).

I tried to fix this by adding the missing `MIMETYPE` property in e02d843d49987cac3a03aab7665a4d67efd351c9 but unfortunately this doesn't fix the problem.

I think it's the same problem that causes [this error](https://github.com/nteract/nteract/pull/1488/files#diff-15529ce2dbbf99b78203bc542ae41888R175).

rgbkrk:
Hurrah! I'll help work this through with you. I'm in favor of switching _all_ of commutable over to plain objects (frozen objects too pretty please).

packages/transforms/src/index.js
```diff
@@ -12,82 +11,76 @@ import JavaScriptDisplay from "./javascript";
...
+ * @param  {Array} ordered list of mimetypes - ['text/html', 'text/plain']
+ * @param  {Map}   mimetype -> React Component - Map({'text/plain': TextTransform})
  * @return {string}          Richest mimetype
  */
 
 export function richestMimetype(
-  bundle: ImmutableMap<string, any>,
-  order: ImmutableList<string> = standardDisplayOrder,
-  tf: ImmutableMap<string, any> = standardTransforms
+  bundle: Object,
```
lgeiger, line 69: I though about using ES6 Maps here but decided it would be best to use objects since the kernels also send objects instead of Maps. This way we can get rid of one extra conversion/parsing step in commutable.

rgbkrk:
Once we tightly specify these, we'll be able to have exact definitions of the metadata we support for each transform too. 😄

packages/transforms/src/index.js
rgbkrk: Yeah I'm thinking we should make the interface for transform modules to effectively be:

```js
module.exports = {
  component: ActualReactComponent,
  MIMETYPE: 'actual/mimetype',
}
```

rgbkrk: Interface wise then I'd assume it would be like this:

```
import * as JSONTransfrom from '@nteract/transform-json';

// JSONTransform.component
// JSONTransform.MIMETYPE
```

Unless there's some better way to put static properties on React classes and declare those as a type. I suppose we _could_ extend `React.Component` even though people don't recommend it...

rgbkrk: Perhaps we keep the default export actually and do this instead for the import interface:

```js
import JSONTransform, { MIMETYPE as JSONMimetype } from '@nteract/transform-json'
```

lgeiger: That's a good idea

lgeiger: I tried this, but couldn't satisfy flow.

packages/transforms/src/index.js
rgbkrk: Objects are good here -- we'll be able to tightly declare the structure.

packages/transforms/src/index.js
rgbkrk: Ok, I can take a gander at this.

rgbkrk:
This highlights to me that we need tests for `notebook-preview`, even if a super simple one. We have to adapt that package to the new API too.

lgeiger:
> This highlights to me that we need tests for notebook-preview, even if a super simple one. We have to adapt that package to the new API too.

We definitely need tests for that.

I adapted it already in d5955d358b7e5a480d79bcb2e8bc61fe8ab65585 but haven't tested it yet.

lgeiger:
I verified that `nodebook-preview-demo` still works.

src/notebook/components/cell/code-cell.js
```diff
@@ -4,6 +4,7 @@ import { List as ImmutableList, Map as ImmutableMap } from "immutable";
 
 import Inputs from "./inputs";
 import { Display } from "../../../../packages/display-area";
+import { outputToJS } from "../../../../packages/commutable";
```
rgbkrk: You no longer need this call since you use `toJS` below.

rgbkrk:
Now that I'm looking at this again I'm surprised you didn't modify commutable to no longer have the outputs in immutable form.

It just dawned on me that display updates might not work now because of how we use keypaths. Try out the example notebook at `example-notebooks/display-updates.ipynb`

lgeiger:
> Now that I'm looking at this again I'm surprised you didn't modify commutable to no longer have the outputs in immutable form.

Yeah, I though this would keep the diff somewhat manageable.

>It just dawned on me that display updates might not work now because of how we use keypaths. Try out the example notebook at example-notebooks/display-updates.ipynb

Just checked, I didn't experienced any problems with the display update notebook.

rgbkrk:
> Just checked, I didn't experienced any problems with the display update notebook.

Try running through the display update notebook, saving it, then reopening it.

I'm suggesting this because of [how the keypaths work](https://github.com/nteract/nteract/blob/548fa45734b7655aad711195619d3fd407f06d08/src/notebook/reducers/document.js#L179-L181).

rgbkrk:
Ok, I've tried this out too. I think we should push this out for the sake of Hydrogen. We can adjust things for the desktop app and figure out any edge cases as we move completely off of Immutable.

PR #1727, opened 2017-06-04
[WIP] prettier everything

rgbkrk:
I was looking over the prettier release notes and noticed that they handle CSS too. I ran prettier over _everything_. However, I'd rather bring in #1726 first. I can revisit this.

### DIFF
--- a/flow-typed/npm/ansi-to-react_vx.x.x.js
+++ b/flow-typed/npm/ansi-to-react_vx.x.x.js
@@ -13,7 +13,7 @@
  * https://github.com/flowtype/flow-typed
  */
 
-declare module 'ansi-to-react' {
+declare module "ansi-to-react" {
   declare module.exports: any;
 }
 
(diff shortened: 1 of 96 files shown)


rgbkrk:
I'm running into a build error locally that I think is related to the `prepublish` mechanism being different in npm5. I'll be back with some other fixes shortly. :wink:

flow-typed/npm/ansi-to-react_vx.x.x.js
```diff
@@ -13,7 +13,7 @@
  * https://github.com/flowtype/flow-typed
```
lgeiger, line 13: I don't think we should prettify all the `flow-type` definitions.

In Hydrogen explicitly state the folders to prettify [here](https://github.com/nteract/hydrogen/blob/master/package.json#L36) and [here](https://github.com/nteract/hydrogen/blob/master/package.json#L47).

rgbkrk: ooooook

The reason I did it is because I updated a definition and the prettier plugin automatically formatted it. After that I went "eh, why not do all of them".

rgbkrk:
I'll go ahead and close this in light of other refactors and major changes going through. When we touch a file it will get prettified anyways.

PR #1732, opened 2017-06-08
chore(CI): Use greenkeeper-lockfile instead of custom script

lgeiger:
Greenkeeper now has support for yarn and npm 5 lockfiles via [`greenkeeper-lockfile`](https://github.com/greenkeeperio/greenkeeper-lockfile).

We can now move away from our custom shell script.


rgbkrk:
Woo!

PR #1728, opened 2017-06-04
chore: switch to npm@5, test node@8

rgbkrk:
Test across the [current long term stable (LTS) and the next LTS of node.js](https://github.com/nodejs/LTS).

PR #1738, opened 2017-06-12
chore: test and update for node8 (keep yarn)

rgbkrk:
This is an alternative to #1728.

If it's good, I'll rebase this down to the one commit it should be since I did it on top of #1728.


rgbkrk:
Everything passes here too!

rgbkrk:
Oh wait, I have some more to make this use all the `yarn` steps.

rgbkrk:
I'll go ahead and merge this since it keeps thing as they were with `yarn` while allowing npm 5 to work.

lgeiger (commented):
Thanks for bringing us up-to-date!

PR #1747, opened 2017-06-17
feat(notebook): add nteract version to the notebook metadata

Madhu94:

### DIFF
--- a/src/notebook/epics/saving.js
+++ b/src/notebook/epics/saving.js
@@ -1,14 +1,14 @@
 /* @flow */
 
-import {
-  ActionsObservable
-} from "redux-observable"; /* eslint-disable no-unused-vars */
+import { ActionsObservable } from "redux-observable"; /* eslint-disable no-unused-vars */
 import { writeFileObservable } from "../../utils/fs";
 import { SAVE, SAVE_AS } from "../constants";
 
 import { changeFilename, save, doneSaving } from "../actions";
 
 import { toJS, stringifyNotebook } from "../../../packages/commutable";
+
+import { remote } from "electron";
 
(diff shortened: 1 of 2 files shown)


src/notebook/epics/saving.js
```diff
@@ -32,7 +32,14 @@ export function saveEpic(action$: ActionsObservable, store: Store<any, any>) {
       action =>
         writeFileObservable(
           action.filename,
-          stringifyNotebook(toJS(action.notebook))
+          stringifyNotebook(
+            toJS(
+              action.notebook.setIn(
+                ["metadata", "nteract", "version"],
+                remote.app.getVersion()
```
rgbkrk, line 39: Sweet!

One caveat though -- sometimes the `remote` module disappears so you'll want to `require` this dynamically and wrap it in a try catch.

rgbkrk:
```
require('electron').remote.app.getVersion()
```

should be good to go for now

Then again... maybe it's because of our webpacking that the weird behavior of `remote` intermittently not being available.

rgbkrk:
I left some notes, I think this is fantastic to bring in for now though. :smile:

The approach I originally wanted to take was to make this part of commutable -- I now see that it pretty much has to be done by the originating app on save.

rgbkrk:
annnnd welcome to the nteract party! :tada:

Madhu94:
@rgbkrk I did think of putting it in commutable, since that's where all the other "notebook-manipulating" functions go. But I'd already made the PR by then so I thought I'd wait for some review comments. 

Thank you! :)

rgbkrk:
Since commutable works outside the electron app, it's perfect where you put it.

PR #1756, opened 2017-06-25
feat(notebook): add one-dark theme colors

Madhu94:


static/styles/theme-one-dark.css
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/* -------------------- */
+/* Miscellaneous Styles */
+/* -------------------- */
+
+body
+{
+  line-height: 1.3 !important;
+}
```
rgbkrk, line 57: Where does this one come from?

rgbkrk:
Thanks for doing this!

static/styles/theme-one-dark.css
Madhu94: @rgbkrk Uhm..The other themes had this property set, so I just let it stay here too.

rgbkrk: Oh hmmmm, perhaps we need that in main.css then?

Madhu94: @rgbkrk That's right, I should have done that.  Guess I should learn to step back and look at the big picture, instead of focusing narrowly on the specific problem I was trying to solve :-(

The line-height and the webkit scrollbar style properties are in all themes except the halloween, christmas and the "undefined" theme, which are not even exposed through the UI right now. I will move them into main.css very soon. 

What is this undefined theme, anyway?

rgbkrk: Those are all easter eggs. :smile:

:warning: the undefined theme may not be safe for people sensitive to color flickering. It was put in place to notice a certain class of theme related bugs for humor.

Madhu94: :-) Easter eggs should be exposed to the end-user somehow, no? These themes are exposed to the developer but...

rgbkrk: The Christmas and Halloween ones only come up during certain times of the year (and it would be ok if we had more). undefined has only one way of coming up, though you can enable it using the `~/.jupyter/nteract.json` file. :wink:

Madhu94: "during certain times of the year " -> Gonna hunt for this part of the code.

" it would be ok if we had more" -> I'll add an Indian holiday soon if I can come up with the colors :smiley:

rgbkrk: > I'll add an Indian holiday soon if I can come up with the colors 

That's awesome!

By the way, I'm trying out packaging of the app again to prep for another release.

BenRussert: > I'll add an Indian holiday soon if I can come up with the colors

@Madhu94 that would be awesome!

> By the way, I'm trying out packaging of the app again to prep for another release.

@rgbkrk  I messed around with nteract some over the weekend. I wasnt able to build the deb file on ubuntu, but figured it was something I needed on my system. (I can get dev mode to work fine)

PR #1759, opened 2017-06-28
feat(app): prompt before closing window

rgbkrk:
This alleviates #1090 by unconditionally asking users if they really want to close.

However it does not close it off fully -- that will require bringing in #1147 for the saving piece.


lgeiger (approved):
Cool! 👍 

Agree we should definitely bring in #1147

Issue #1733, opened 2017-06-09
Allow scrolling past the end of the document


adamsiembida:
I believe that ntercat should allow scrolling past the end of the document. The way it is now, I have to work at the bottom of my screen any time a document is longer than the window. Scrolling past the end of the screen would allow me to position the cell I am editing in the middle of my monitor.

rgbkrk:
Done in #1781. About to be shipped in 0.2.0 too!

PR #1761, opened 2017-06-29
ux(scroll-buffer): Add buffer div so user can scroll cell to top

jkornblum:
This commit addresses issue #1733

An empty div will be added below the last cell if the last cell has no output. It will allow the user to scroll to the last cell to the top of the screen.

Happy to get feedback on whether `calc(100vh - 120px)` is appropriate. I think this should work.
I've removed some CSS from the `classic` theme. I could not see it helping anything, but it changes the height of the cell compared to other themes.

Also I tried to run Flow but I had some issues, Flow is wonky on Windows.

### DIFF
--- a/src/notebook/components/notebook.js
+++ b/src/notebook/components/notebook.js
@@ -12,6 +12,7 @@ import Cell from "./cell/cell";
 import DraggableCell from "../providers/draggable-cell";
 import CellCreator from "../providers/cell-creator";
 import StatusBar from "./status-bar";
+import ScrollBuffer from "./scroll-buffer";
 import {
   focusNextCell,
   focusNextCellEditor,
@@ -19,6 +20,7 @@ import {
   executeCell
 } from "../actions";
 import type { CellProps } from "./cell/cell";
+import type { ImmutableCell } from "../../../packages/commutable/";
 
 // TODO: Remove after provider refactor finished
 const PropTypes = require("prop-types");
@@ -71,6 +73,7 @@ export class Notebook extends React.PureComponent {
   props: Props;
   createCellElement: (s: string) => ?React.Element<any>;
   createStickyCellElement: (s: string) => ?React.Element<any>;
+  findLastCell: () => ImmutableCell;
   keyDown: (e: KeyboardEvent) => void;
   moveCell: (source: string, dest: string, above: boolean) => void;
   stickyCellsPlaceholder: HTMLElement;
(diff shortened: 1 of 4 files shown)


src/notebook/components/notebook.js
```diff
@@ -177,6 +182,21 @@ export class Notebook extends React.PureComponent {
...
+   * @returns {ImmutableCell} 
+   * @memberof Notebook
+   */
+  findLastCell(): ImmutableCell {
+    const cellOrder = this.props.notebook.get("cellOrder");
+    const cellMap = this.props.notebook.get("cellMap");
+    const cellId = cellOrder.get(-1);
+    const cell = cellMap.get(cellId);
+
+    return cell;
```
rgbkrk, line 197: Note: you can have notebooks with zero cells. The edge case should be handled, possibly by the caller of this function.

jkornblum: So how is this typically handled in JavaScript? In Python one would return None or raise an exception. Here the function returns undefined if `.get(-1)` does not exist. Is it bad practice to return undefined in situations like this, is returning null more typical?

In my my react component I added the `if (this.props.lastCell)` to handle that situation.

static/styles/main.css
```diff
@@ -182,7 +182,7 @@ body
 }
 
 .notebook .cell-container:last-child {
-    padding-bottom: 200px;
+    padding-bottom: 20px;
```
rgbkrk, line 185: What if we bumped this amount up higher? It does currently allow for a pretty big buffer at the moment without having to have another element.

src/notebook/components/notebook.js
rgbkrk: Oh I see you did handle it. Never mind me.

static/styles/main.css
jkornblum: Yeah that's a styling/ux decision. I don't have a strong opinion.

Pro: It's a much simpler to just increase this padding value.
       Can provide the same buffer amount if desired `(100vh - 110px)`

Con: Will leave that spacing even when the last cell is not a "new" cell.
        Will not allow user to scroll cell to top (if px value is not large enough)
        
I find myself trying to scroll my last cell to the top when I work.

rgbkrk: I just tried using the style you used in the special element above as the padding here `calc(100vh - 110px)` and it works quite well.

jkornblum: The issue with the static CSS is that when you go to export PDF or print the notebook it will have this large empty space. Another option could be to add a class to the last cell (if empty) just apply CSS styling to that class.

rgbkrk: ohhhhhhh, in that case there's one more thing to add to the top level:

```css
@media not print {
  .notebook .cell-container:last-child {
    padding-bottom: calc(100vh - 110px)
  }
}
```

That way we're only targeting the non-print view.

jkornblum:
I've kept other pr much simpler (no special logic) but still allow use to scroll last cell to the bottom.
Closing this in favor of #1781

PR #1765, opened 2017-07-05
fix(vega): Add version key to Vega MIMETYPE

lgeiger:
The next version of altair will return JSON with MIMETYPES `application/vnd.vega.v2+json` and `application/vnd.vegalite.v1+json`:  https://github.com/altair-viz/altair/pull/216


rgbkrk:
Great, thanks! I've been looking forward to this change. :smile:

PR #1767, opened 2017-07-06
SSR readiness

### LINKED ISSUE #1766: Server Side Renderable Components
I've been working with next.js and our components (again) and am running into issues with the Editor component + CodeMirror. CodeMirror wants certain globals on the page, which can be "polyfilled" but breaks things for other components. Instead of doing that I'd like to focus on making all the components server-renderable. I'll track these as individual items here:

* [x] Conditionally load CodeMirror in editor component
* [x] LaTeX component should not use `window` server side
* [x] When server side, the HTML transform "_could_" just use `dangerouslySetInnerHTML`
* [x] Fix up the plotly build
* [x] Inference for the data resource transform

rgbkrk:
This readies parts of our app for server side rendering.

Closes #1766 

* [x] Conditionally load CodeMirror in editor component
* [x] LaTeX component should not use `window` server side
* [x] Provide easy ways to use our CSS, somehow scoped...


rgbkrk:
One major drawback in my PR here is that I copied over CSS to make it available for next.js to serve from `/static` when I really wish I was loading from `/node_modules`. In reality, I don't want to use any css in this way -- I'd rather just embed it with styled-jsx.  ¯\\\_(ツ)\_/¯

rgbkrk:
We also should tweak the editor component some more so we have the static version of the codemirror cells here. Otherwise there's a little bit of a flicker waiting for codemirror to instantiate (I think).

renderit

However, as you can see in the GIF it loads super ffffast (I did not modify the video there).

rgbkrk:
Ok, I've ported over some of the styles to the default `<textarea>` so it's a bit less jarring. We need some adjustments to how we do the css to make this maintainable. Works for now though!

getting_better

I'm about ready to change up the HTML transform so that it `dangerouslySetInnerHTML`s when on the server while doing the proper thing on the client side (the current document node creation).

rgbkrk:
From Travis build logs:

```
npm ERR! peer dep missing: ajv@>=5.0.0, required by ajv-keywords@2.1.0
```

rgbkrk:
Uh oh, that's master failing: https://travis-ci.org/nteract/nteract/builds/251355182

rgbkrk:
All is set now after #1773

lgeiger (approved):
Excellent stuff 👍 💯

rgbkrk:
🙇 Thank you!!!

PR #1769, opened 2017-07-07
chore(npm): Switch to npm@5

lgeiger:
Switch back to npm to prevent strange build issues with `yarn`.


rgbkrk:
Thank you so much, I'm super happy about npm 5. One less default tool required.

PR #1774, opened 2017-07-08
dx: make desktop app resolve using the direct src

rgbkrk:
This brings us back to the golden age of hacking on nteract.

You can now run two terminals for ease of hacking:

## Terminal 1

```
npm run build:renderer:watch
```

## Terminal 2

```
npm run build:main && npm run spawn
```

No more having to build the subpackages. Initial `npm i` is still a bit painstaking.


jdetle:
Awesome!

rgbkrk:
Went ahead and shipped patch releases since it's a nice thing to add on.

PR #1777, opened 2017-07-08
fix(notebook-preview-demo): Fix placeholder CSS to load smooth

lgeiger:
## PR
pr
## master
master

Now the only thing that flickers is the code mirror wrapping which cannot be fixed with simple CSS.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1799,17 +1799,6 @@
 			"integrity": "sha1-El+wBgS2P7sYjblqZnziki3NbN0=",
 			"dev": true
 		},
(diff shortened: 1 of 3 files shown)


rgbkrk:
> Now the only thing that flickers is the code mirror wrapping which cannot be fixed with simple CSS.

Sweet. Yeah that's since we (or CodeMirror) relies on `CodeMirror-lines`. If this wasn't a `<textarea>` it would be easy to split them (and it has to be a textarea for instantiating after).

rgbkrk (approved):

PR #1778, opened 2017-07-09
set inner HTML for server side render

rgbkrk:
Following on to #1777, now with even less flicker since the HTML portion can be server side rendered.

less-flicker

Props to @lgeiger for making me want to perfect it even more.

### DIFF
--- a/packages/transforms/src/html.js
+++ b/packages/transforms/src/html.js
@@ -8,6 +8,11 @@ type Props = {
 // Note: createRange and Range must be polyfilled on older browsers with
 //       https://github.com/timdown/rangy
 export function createFragment(html: string): Node {
+  /**
+   * createFragment takes in an HTML string and outputs a DOM element that is
+   * treated as if it originated on the page "like normal".
+   * @type {Node} - https://developer.mozilla.org/en-US/docs/Web/API/Node
+   */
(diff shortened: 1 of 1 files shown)


packages/transforms/src/html.js
```diff
@@ -20,6 +20,11 @@ export default class HTMLDisplay extends React.Component {
   static MIMETYPE = "text/html";
 
   componentDidMount(): void {
+    // clear out all DOM element children
+    // This matters on server side render
+    while (this.el.firstChild) {
+      this.el.removeChild(this.el.firstChild);
+    }
     this.el.appendChild(createFragment(this.props.data));
```
lgeiger: Is this still necessary since we now use `dangerouslySetInnerHTML`?

lgeiger (approved):

packages/transforms/src/html.js
rgbkrk: If we do `dangerouslySetInnerHTML`, it will create those DOM nodes on mounting. They have to be removed before we create the document fragment.

rgbkrk:
Added some comments to hopefully make what should otherwise be a really simple component very clear.

lgeiger (approved):
Thanks for the comments 👍

PR #1780, opened 2017-07-10
Rx lightweight

rgbkrk:
I've begun trimming down our Rx imports, mostly to make our bundles smaller and our loading faster.

Running into a few issues though that I'll need to track down.


rgbkrk:
Trim down finished, and followed on with https://github.com/nteract/enchannel-zmq-backend/issues/55.

lgeiger (approved):
Excellent! Thanks for cleaning this up 👍

PR #1781, opened 2017-07-10
ux(notebook): change padding after last cell so user can scroll cell to top

jkornblum:
This commit addresses issue #1733
This commit incorporates feedback from pr #1761 

This simply changes CSS to add dynamic padding so that the last cell may be scrolled to the top.

I've removed some CSS from the classic theme. I could not see it helping anything, but it gives a different cell spacing compared to other themes.

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -181,10 +181,11 @@ body
     padding-right: 10px;
 }
 
-.notebook .cell-container:last-child {
-    padding-bottom: 200px;
+@media not print {
+    .notebook .cell-container:last-child {
+        padding-bottom: calc(100vh - 110px);
+    }
 }
-
 /*
   Cell
  */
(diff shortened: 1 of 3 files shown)


static/styles/main.css
```diff
@@ -181,10 +181,11 @@ body
     padding-right: 10px;
 }
 
-.notebook .cell-container:last-child {
-    padding-bottom: 200px;
+@media not print {
+    .notebook .cell-container:last-child {
+        padding-bottom: calc(100vh - 110px);
```
rgbkrk, line 186: 👍

static/styles/theme-classic.css
```diff
@@ -77,13 +77,6 @@
   padding: 0px 0px 0px calc(var(--prompt-width) + 10px);
 }
 
-.draggable-cell {
-  padding: 0px !important;
-  border-top-width: 1px !important;
-  border-bottom-width: 1px !important;
-  margin-bottom: -1px;
-}
```
rgbkrk, line 85: 👍

rgbkrk:
Thanks for coming back to this and incorporating feedback. 😄

PR #1783, opened 2017-07-10
chore: upgrade enchannel-zmq and package lock

rgbkrk:
Closing off the Rx shakedown.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,13 +21,13 @@
 			"integrity": "sha512-jgBTCcJ0gQedE9o8Jw+H/Gyq//EnQxmVpha7CdprIwzRSC81Uj37inHvPzv6jaZgZwkCcfho52rAaIFBrdbO7w==",
 			"dev": true,
 			"requires": {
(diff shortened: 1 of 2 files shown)


rgbkrk:
In reality, plotly is the dominant part of our bundle (and doesn't get tree-shaken since we're using the dist version). It's good to have this trimmed down though.

rgbkrk:
Found a potential "offendor" keeping our bundle from being small:

```
./node_modules/stream-to-observable/rxjs-all.js:var Observable = require('rxjs/Rx').Observable;
```

rgbkrk:
It looks like it _should_ detect using `any-observable` so that's probably not a problem. 🤔

rgbkrk:
Ok, that's a red herring.

```
$ npm ls stream-to-observable
nteract@0.1.0 /Users/kylek/code/src/github.com/nteract/nteract
└─┬ lint-staged@4.0.1
  └─┬ listr@0.12.0
    └── stream-to-observable@0.1.0
```

Looks like we are where we need to be. 😄

PR #1788, opened 2017-07-15
ux(print): Update style to ignore focus cell on print for all themes

### LINKED ISSUE #1734: Add way to deselct a cell for printing
Right now I am using the classing theme. This means that when a cell is selected, it has a green border around it. Other themes also have some indication of which cell is selected.

During printing (to PDF) it seems that even the cell selection indication also prints. I don't believe that cell selection has any relevance to a printed document.

I suggest adding a way to deselect all cells. For example, pressing the escape key might be a way to do that. I also find the cleaner look without having anything selected would be nice for non-interactive presentations using nteract directly (i.e., without printing).

Additionally, it might make sense to automatically deselect everything during printing as well.

rpalo:
Hi!  I think I fixed #1734 for all of the themes that show up right out of the box (light, dark, classic, nteract, and one-dark).  This is essentially my first contribution to an Open Source project, so let me know if I could do/could have done anything better 😄 

### DIFF
--- a/static/styles/main.css
+++ b/static/styles/main.css
@@ -22,6 +22,14 @@
   .cell-creator{
     display: none!important;
   }
+  .cell.focused {
+    border: none;
+    background: var(--cell-bg)!important;
+  }
+  .cell:focus .prompt,
+  .cell.focused .prompt {
+      background: var(--pager-bg)!important;
+  }
 
 }
 
--- a/static/styles/theme-nteract.css
+++ b/static/styles/theme-nteract.css
@@ -367,6 +367,17 @@ ul li::before
 }
 
 /* -------------------- */
+/*   Printing Styles    */
+/* -------------------- */
+@media print {
+  .cell:focus .cm-s-composition.CodeMirror,
+  .cell.focused .cm-s-composition.CodeMirror,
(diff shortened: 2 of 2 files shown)


static/styles/main.css
```diff
@@ -22,6 +22,14 @@
...
     display: none!important;
   }
+  .cell.focused {
+    border: none;
+    background: var(--cell-bg)!important;
+  }
+  .cell:focus .prompt,
+  .cell.focused .prompt {
+      background: var(--pager-bg)!important;
+  }
```
rgbkrk, line 32: It took me a sec to recognize these were under the print stylesheet segment. 😄

rgbkrk:
Thanks for the contribution, I think this is great. I don't have a chance to test, certainly in favor of CSS based changes for printing. :smile:

rgbkrk:
@jkornblum I added you as a reviewer since you've worked with these quite a bit. 😄

rgbkrk (approved):

rpalo:
Just in case this helps:
## Before (in print/PDF)
before
## After (in print/PDF)
after

rgbkrk:
Thanks and welcome to the project! Great to have your help.

PR #1791, opened 2017-07-18
Discuss: enable plotly transform to be "created"

rgbkrk:
I want to put Plotly in the `<head>` of nteract desktop and commuter to ease the webpack build and make loading consistent. An approach that I'd like to take is to make the plotly transform module be a function that accepts the `Plotly` object which will create the class.

With this setup, the current Plotly transform would be setup like this:

```js
const PlotlyTransform = createPlotlyTransform(require("@nteract/plotly"));
```

With Plotly part of a `<script>` in the `<head>`, creating the PlotlyTransform is a matter of:

```js
const PlotlyTransform = createPlotlyTransform(global.Plotly);
```

No matter what, I'd intend for this to be a major bump in this transform (as well as transforms-full).

### DIFF
--- a/packages/transform-plotly/src/index.js
+++ b/packages/transform-plotly/src/index.js
@@ -15,66 +15,74 @@ declare class PlotlyHTMLElement extends HTMLElement {
 
 const MIMETYPE = "application/vnd.plotly.v1+json";
 
-export class PlotlyTransform extends React.Component {
-  props: Props;
-  getFigure: () => Object;
-  el: PlotlyHTMLElement;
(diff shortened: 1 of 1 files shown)


rgbkrk:
I like what @theengineear did in https://github.com/nteract/commuter/pull/210 and think we should do the same here. Closing, in hopes we'll follow suit with the commuter implementation (except using the bundled Plotly instead of the CDN).

rgbkrk:
Closing in favor of nteract/commuter#210 to be brought into here.

PR #1799, opened 2017-07-25
fix: correct the mimetype compare function

### LINKED ISSUE #1798: richestMimetype is returning the wrong value
```js
> richestMimetype({'image/png': '999', 'text/plain': '123', 'application/vnd.vegalite.v1+json': {}})
text/plain
```

rgbkrk:
Fixes #1798.

/cc @lgeiger

### DIFF
--- a/packages/transforms/src/index.js
+++ b/packages/transforms/src/index.js
@@ -75,7 +75,7 @@ export function richestMimetype(
       // we can only use those we have a transform for
       .filter(mimetype => tf[mimetype] && order.includes(mimetype))
       // the richest is based on the order in displayOrder
-      .sort(mimetype => order.indexOf(mimetype))[0]
+      .sort((a, b) => order.indexOf(a) - order.indexOf(b))[0]
   );
 }
 export const transforms = standardTransforms;


rgbkrk:
I'm going to ship after this because I noticed that it was only showing `text/plain` on notebooks on commuter.

packages/transforms/src/index.js
```diff
@@ -75,7 +75,7 @@ export function richestMimetype(
       // we can only use those we have a transform for
       .filter(mimetype => tf[mimetype] && order.includes(mimetype))
       // the richest is based on the order in displayOrder
-      .sort(mimetype => order.indexOf(mimetype))[0]
```
theengineear, line 78: Is this a breaking change? If you're depending on whatever the sort order currently is, won't this new sort break your code?

theengineear: Nvm, @rgbkrk and I chatted about it.

theengineear (approved):

packages/transforms/src/index.js
rgbkrk: It was using just fine logic when this was using Immutable.js. `sort` now needs to return -1, 0, or 1 because it's a regular js sort. This was making all outputs end up as `text/plain`.

lgeiger:
Thanks for fixing this!

packages/transforms/src/index.js
```diff
@@ -75,7 +75,7 @@ export function richestMimetype(
...
+      .sort((a, b) => order.indexOf(a) - order.indexOf(b))[0]
```
rgbkrk, line 78: It was really hard not to precompute these, which makes me want to create a real class for the displayOrder and transforms (which would likely be mutative for adding in new mimetypes or replacing old ones).

PR #1801, opened 2017-07-25
Graphviz Transform

rgbkrk:
Spurred by [a twitter conversation](https://twitter.com/nteractio/status/889894412095938560) with @Jaykul and others, I whipped up a quick graphviz transform.

screen shot 2017-07-25 at 1 23 06 pm

### DIFF
--- a/example-notebooks/graphviz.ipynb
+++ b/example-notebooks/graphviz.ipynb
@@ -1,0 +1,128 @@
+{
+  "cells": [
+    {
+      "cell_type": "code",
+      "source": [
+        "from graphviz import Digraph\n",
+        "\n",
+        "def dotsplay(self):\n",
+        "    display({\n",
+        "        'text/vnd.graphviz': self.source\n",
+        "    }, raw=True)\n",
+        "\nDigraph._ipython_display_ = dotsplay"
+      ],
+      "outputs": [],
+      "execution_count": 71,
+      "metadata": {
+        "collapsed": false,
+        "outputHidden": false,
+        "inputHidden": false
+      }
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "dot = Digraph(comment='Jupyter Architecture')\n",
+        "\n",
+        "dot.node('kernel', 'Kernel (Python)')\n",
+        "dot.node('frontend', 'Frontend (Notebook)')\n",
+        "\n",
+        "dot.edge('kernel', 'frontend')\n",
+        "\n",
+        "dot.node('user', 'User (you!)')\n",
+        "dot.edge('frontend', 'user')\n",
+        "\ndot"
+      ],
+      "outputs": [
+        {
+          "output_type": "display_data",
+          "data": {
+            "text/vnd.graphviz": [
+              "// Jupyter Architecture\n",
+              "digraph {\n",
+              "\tkernel [label=\"Kernel (Python)\"]\n",
+              "\tfrontend [label=\"Frontend (Notebook)\"]\n",
+              "\tkernel -> frontend\n",
+              "\tuser [label=\"User (you!)\"]\n",
+              "\tfrontend -> user\n",
+              "}"
+            ]
+          },
+          "metadata": {}
+        }
+      ],
+      "execution_count": 72,
+      "metadata": {
+        "collapsed": false,
+        "outputHidden": false,
+        "inputHidden": false
+      }
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "print(dot.source)"
+      ],
+      "outputs": [
+        {
+          "output_type": "stream",
+          "name": "stdout",
+          "text": [
+            "// Jupyter Architecture\n",
+            "digraph {\n",
+            "\tkernel [label=\"Kernel (Python)\"]\n",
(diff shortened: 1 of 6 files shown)


gnestor (approved):

rgbkrk:
Oh look a `%%dot` magic:

screen shot 2017-07-25 at 2 32 48 pm

😈

Jaykul:
This one's cooler http://js.cytoscape.org/ ;-)

rgbkrk:
> This one's cooler http://js.cytoscape.org/ ;-)

cytoscape is cool! Is the cytoscape JSON format pretty well standardized? That's the main requirement for a supported transform (mimetype & spec).

rgbkrk:
My chief concern on this one is if it increases the overall app size considerably (since it's effectively graphviz run through emscripten).

rgbkrk:
Closing this for now, if someone wants it I'd love to measure the bundle size to make sure this is appropriate later.

mpacer:
@bollwyvl, would your solution that you talked about today for lazy loading a graphviz emscripten bundle apply to nteract as well?

rgbkrk:
I've unlocked this since @mpacer wants to open this up for discussion. I'm a fan of having a graphviz transform. Now that nteract is about to support dynamically `import()`ing transforms I'd be fine having this one loading that way since it's pretty mega.

bollwyvl:
Graphviz all the things!

- some of the blogs i read
  - https://templecoding.com/blog/2016/02/17/async-await-with-es6-babel-and-typescript/
  - https://blog.mariusschulz.com/2018/01/14/code-splitting-with-import-typescript-and-webpack
- the wrapper
  - https://github.com/PhE/jupyterlab_graphviz/blob/master/src/viz-wrap.ts
- the tsconfig that says to use esnext
  - https://github.com/PhE/jupyterlab_graphviz/blob/master/src/tsconfig.json#L7
- a notebook of all the things I could dig up for a demo
  - https://github.com/PhE/jupyterlab_graphviz/blob/master/samples/Graphviz%20Rich%20Display%20Cookbook.ipynb

It all seems to work, but I don't know how really kosher it all is.

rgbkrk:
As long as we agree on a mimetype and what the raw format is, I'm happy. 😄 

In your demo you used `application/vnd.graphviz` and in this PR I used `text/vnd.graphviz`. Doing a bit of digging again, it appears [`text/vnd.graphviz` is the officially registered media type](https://www.iana.org/assignments/media-types/text/vnd.graphviz). I'd ask to switch them out, it looks like @PhE's repo has been going for about a year though so I only want to slightly nudge that it should align with the registered media type.

bollwyvl:
https://github.com/PhE/jupyterlab_graphviz/pull/5#issue-185587410

bollwyvl:
Oh, and I'm happy to split out the CodeMirror mode into a separate package if that would be useful. I've been hammering on it a fair amount, and it seems pretty robust, but probably hasn't had too many eyes on it, and there's no [official place to land it anymore](https://github.com/codemirror/CodeMirror/blob/master/CONTRIBUTING.md#contributing-code) :crying_cat_face: .

rgbkrk:
I'm cool with using a node module that we can require. We [keep a separate module for the ipython mode](https://github.com/nteract/nteract/blob/master/packages/editor/src/mode/ipython.js) and could easily move that one to a separate module too.

bollwyvl:
Hm, in light of the official CM channel spinning down, would a `jupyter(lab)/editor-modes` monorepo (and maybe npm `@jupyter-editor-modes` scope) make sense? Even though jlab is (maybe) [considering](https://github.com/jupyterlab/jupyterlab/issues/135) [adopting](https://github.com/jupyterlab/jupyterlab-monaco) [monaco](https://microsoft.github.io/monaco-editor/), it might be good to have a place to work on things independent of the frontend and maybe the underlying library, and make things like [jupyter markdown](https://github.com/jupyter/nbformat/issues/80) more real.

:phone: @jasongrout @gnestor

rgbkrk:
Amusingly, we just [adopted monaco for the file editor on the nteract jupyter extension](https://github.com/nteract/nteract/pull/2779). It's still codemirror in the notebook app though.

rgbkrk:
It's still codemirror in the notebook app though.

bollwyvl:
Indeed! And lo, there is an [official repo](https://github.com/Microsoft/monaco-languages/tree/master/src), which also doesn't have graphviz yet, and they would likely accept PRs!

There is also [this repo](https://github.com/Stephanvs/vscode-graphviz), which looks pretty quiet, 

Either way, CM is here right now, so maybe there's something in the near term. Seems like a good thing to raise for the Dev Meeting agenda.

rgbkrk:
Yup, I don't want to breakaway from CodeMirror within the notebook yet as I understand it deeply and I'd have to do a lot of :new: hand tuning to get Monaco to work well in our notebooks (mostly because of what's enabled by default and the sheer number of web workers that would be on the page per cell...)

jasongrout:
In my very initial testing in JLab, it seems that the memory use of Monaco is far greater than codemirror, so it may not be appropriate for a notebook.

PR #1826, opened 2017-08-10
Optional for vega

rgbkrk:
This makes optional dependencies in vega get ignored (no more building of `canvas`, since we don't need it).

Way faster builds now!


lgeiger (approved):
Awesome!

PR #1827, opened 2017-08-10
chore: update flow types for codemirror and lodash

rgbkrk:
This brings in some codemirror definitons from @dperetti in https://github.com/flowtype/flow-typed/issues/357. Went ahead and addressed an issue with the lodash definition (that I've fixed in flow-typed) that causes prettier to fail (use of the `default` keyword as a parameter).


flow-typed/npm/codemirror_v5.x.x.js
```diff
@@ -0,0 +1,197 @@
...
+    ) => void
+  };
+
+  declare class CodeMirror {
+    static fromTextArea(
+      textArea: HTMLElement,
+      config?: CodeMirrorOptions
+    ): CodeMirror,
+
+    static defineMode(mimetype: string, modefactory: any): void,
```
rgbkrk, line 101: These are the extra definitions I snuck in on top of what was in https://gist.github.com/dperetti/d44379562fb656a5cf39f57ab6cc28f4 (as of https://gist.github.com/dperetti/d44379562fb656a5cf39f57ab6cc28f4/085e30aed7ed055cf2ae5d00f75342a44c64aa05)

PR #1837, opened 2017-08-15
Monoslurp enchannel

rgbkrk:
Bringing in `enchannel` as a package within the monorepo. It should probably get folded into `@nteract/messaging` next.


rgbkrk:
Coverage goes down a little bit here, though it's code we're already using (since it came from a separate repo / package). Ready to ship though!

PR #1838, opened 2017-08-15
transform-dataresource: Fix infer call signature

gnestor:
Demo using it in commuter with https://github.com/nteract/commuter/pull/211


rgbkrk:
Thank you so much!

rgbkrk:
Woo, thank you again!

PR #1840, opened 2017-08-16
fix: handle surrogate pairs

### LINKED ISSUE #1706: handle surrogate pairs in character offsets
CodeMirror indices must be converted from UTF-16 units to actual character offsets:

> I think it makes the most sense for the protocol spec to be 'characters', so if CodeMirror indices are returning UTF-16 units, it ought to be the responsibility of the Jupyter javascript to deal with surrogate pairs and turn that into actual character offsets. - @minrk

See:

* https://github.com/jupyter/notebook/pull/2509
* https://github.com/jupyter/jupyter_client/issues/259

rgbkrk:
Fixes #1706

* [x] completion
* [x] introspection

/cc @minrk @Carreau how's this look for a start?

Finally got to this @stevengj, hoping to get it done to make a release in the next few days.

### DIFF
--- a/packages/editor/src/complete.js
+++ b/packages/editor/src/complete.js
@@ -83,6 +83,7 @@ export const expand_completions = editor => results => {
     // handle surrogate pairs
     // HACK: This seems susceptible to timing issues, we could verify changes in
     //       what's in the editor, as we'll be able to correlate across events
+    //       Suggestions and background in https://github.com/nteract/nteract/pull/1840#discussion_r133380430
     const text = editor.getValue();
     end = char_idx_to_js_idx(end, text);
     start = char_idx_to_js_idx(start, text);
--- a/packages/editor/src/complete.js
+++ b/packages/editor/src/complete.js
@@ -6,6 +6,8 @@ import "rxjs/add/operator/timeout";
 
 import { createMessage } from "@nteract/messaging";
 
+import { js_idx_to_char_idx, char_idx_to_js_idx } from "./surrogate";
+
 // Hint picker
 export const pick = (cm, handle) => handle.pick();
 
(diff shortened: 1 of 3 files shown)


packages/editor/src/complete.js
```diff
@@ -63,14 +65,37 @@ export const expand_completions = editor => results => {
...
+    // and negative start relative to that
+    end = editor.indexFromPos(editor.getCursor());
+    if (start === null) {
+      start = end;
+    } else if (start < 0) {
+      start = end + start;
+    }
+  } else {
+    // handle surrogate pairs
+    // HACK: This seems susceptible to timing issues, we could verify changes in
```
minrk, line 84: Re: timing issue, you could cache the value keyed by the request msg_id and use that for transforming the value back. It might also be worthwhile to just dismiss completions if input changed since completion was requested, even excluding the potential (unlikely) surrogate measurement issue.

I'm not sure how often `import m<tab>atpl<completion arrives with matplotlib>` happens or how well you handle it.

minrk (approved):

packages/editor/src/complete.js
rgbkrk: Typically with network latency (remote deployments, crappy wifi) and with slow kernels (Toree, other JVM based kernels) this comes up. What makes matters a little bit worse with nteract is that we do typeahead instead of tab complete. CodeMirror seems smart enough to ignore it if the cursor is not where the suggestion is though, with my limited testing.

rgbkrk:
I recommitted as you by the way @minrk, since I yoinked this code out of jupyter/notebook while I was offline.

Carreau (approved):
The potential timing issue also exist in classic notebook.

rgbkrk:
Thank you both!

PR #1850, opened 2017-08-18
Scroll cell into view when active

rgbkrk:
Well this is neat, I figured out how to do scrolling in a Chrome & WebKit specific way `scrollIntoViewIfNeeded`. It works pretty well, tending to focus the next cell into the middle.


rgbkrk:
~~One glaring issue I'm running into with this now though -- if your code cell is big enough, partially in view, and you click into a cell, codemirror (sometimes) selects a bunch of text on scroll.~~ Fixed by not letting scroll happen when the mouse is already hovered on the cell!

rgbkrk:
OMG I am so happy. We already had hover calculation, which I was able to use to exclude when we would do scrolling into the viewport. :D :D :D

Issue #1849, opened 2017-08-17
Optimizing the desktop app size


rgbkrk:
As an experiment, I told electron-builder to not bundle in `node_modules` to save size (put an entry in `build.files` that is `!**/node_modules`). Immediately after building I ran into the following issue:

* the main script needs to be webpack bundled _and_ use the vendored RxJS that the frontend also uses
* some modules may need to be `require`able, based on our webpack config -- native modules and their dependencies

lgeiger:
We want to have the following modules requireable:

```bash
> npm-remote-ls --development false jmp
└─ jmp@0.7.5
   ├─ uuid@3.1.0
   └─ zeromq@4.5.0
      └─ nan@2.6.2

> npm-remote-ls --development false github
└─ github@9.2.0
   ├─ netrc@0.1.4
   ├─ follow-redirects@0.0.7
   │  ├─ stream-consume@0.1.0
   │  └─ debug@2.6.8
   │     └─ ms@2.0.0
   ├─ https-proxy-agent@1.0.0
   │  ├─ agent-base@2.1.1
   │  │  ├─ extend@3.0.1
   │  │  └─ semver@5.0.3
   │  ├─ extend@3.0.1
   │  └─ debug@2.6.8
   └─ mime@1.3.6
```

Unfortunately `electron-builder` doesn't accept my settings 😢

rgbkrk:
I can work on the files handler and the webpack setup for that. I can't recall why we made github separated at the top of the webpack commonjs ignores.

lgeiger:
One thing we might want to check out is: https://github.com/electron-userland/electron-webpack

rgbkrk:
On commit d7c0fc23e9a8948d34f40cf19d8c31d3d5c99a3d: 

* DMG size: 55M
* Unpacked Mac App: 140M

v0.2.0:

* DMG size: 75M
* Unpacked Mac App: 198M

lgeiger:
I'll look into tree-shaking and minification after #1934

PR #1852, opened 2017-08-18
Webpack main script

lgeiger:
and remove the package build since we use the `nteractDesktop` field with webpack.

This adresses part of #1849


src/main/index.js
```diff
@@ -34,8 +34,9 @@ const kernelspecs = require("kernelspecs");
 const jupyterPaths = require("jupyter-paths");
 const path = require("path");
 
-const argv = require("yargs")
-  .version()
+const yargs = require("yargs/yargs");
+const argv = yargs()
+  .version(() => require("./../../package.json").version)
```
rgbkrk, line 39: I hope webpack bundles this as a constant so it's not a pain point on the built app.

Actually... wait a second... I think this is available via an Electron API. 🤔

src/main/index.js
```diff
@@ -118,7 +119,7 @@ export function createSplashSubscriber() {
         show: false
       });
 
-      const index = join(__dirname, "..", "..", "static", "splash.html");
+      const index = join(__dirname, "..", "static", "splash.html");
```
rgbkrk, line 122: Why did this change? Oh! Because the built file is now at `lib/webpack-main`... right. 🤔  I wonder if there's a way to reach these assets in an idiomatic way in electron.

src/main/kernel-specs.js
```diff
@@ -7,7 +7,14 @@ const KERNEL_SPECS = {
     spec: {
       argv: [
         process.execPath,
-        join(require.resolve("ijavascript"), "..", "lib", "kernel.js"),
+        join(
+          __dirname,
+          "..",
+          "node_modules",
+          "ijavascript",
```
rgbkrk, line 14: This reminds me that we need to make sure to include this specific module in our `files` array for electron builder.

webpack.config.js
```diff
@@ -62,4 +84,6 @@ module.exports = {
       exclude: ["vendor.js"]
     })
   ]
-};
+});
+
+module.exports = [mainConfig, rendererConfig];
```
rgbkrk, line 89: I had no idea you could do this with webpack!

I wonder if they could/should share the vendor bundle (mostly to deduplicate RxJS).

rgbkrk:
Are these Travis failures unrelated? This is weird, looks like lerna didn't bootstrap right.

src/main/index.js
lgeiger: I just checked, this also works with the bundled app.

src/main/index.js
lgeiger: I'm pretty sure there's a better way to do this with webpack but I don't know one from the top of my mind :D

webpack.config.js
lgeiger: We could do this with [webpack entry points](https://webpack.js.org/configuration/entry-context/) but I don't know if we can define different targets/configs then. If they would share the same config I don't think we would gain a lot.

lgeiger:
> Are these Travis failures unrelated? This is weird, looks like lerna didn't bootstrap right.

Looks like we still need to build the packages for the tests to run since our testing setup doesn't recognize the `nteractDesktop` main field.

It would be great if the tests would also use the `nteractDesktop` main field since we could remove one build step to speed things up.

package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "postinstall": "electron-builder install-app-deps",
     "reinstall": "npm run clean && rimraf node_modules && npm install",
-    "prestart": "npm run build",
+    "prestart": "npm run build:webpack",
```
rgbkrk, line 20: 👍

package.json
```diff
@@ -27,7 +27,7 @@
     "prebuild": "rimraf lib",
     "lerna": "lerna bootstrap --hoist",
     "install": "npm run lerna && npm run build",
-    "build": "npm run build:webpack",
+    "build": "npm run build:packages && npm run build:webpack",
```
rgbkrk: 👍

webpack.config.js
rgbkrk: Yeah not a big deal, we can come back to optimizing that part.

src/main/index.js
rgbkrk: Ok, great. That addresses a comment I was going to make on another one too (the splash).

rgbkrk:
Awesome!!!

src/main/kernel-specs.js
lgeiger: I think I should revert this change and add `ijavascript` to the externals in the webpack config

PR #1854, opened 2017-08-19
Move deps bundled by webpack to dev-deps

lgeiger:
This removes unnecessary node modules from our bundled app.

nteract.app size:
- 0.2.0: 207.3 MB
- PR: 159.5 MB

nteract.dmg size:

- 0.2.0: 74.9 MB
- PR: 63 MB

Now we can start optimizing our webpack config to bundle more assets and further improve package size.


package.json
```diff
@@ -94,7 +94,7 @@
       "category": "science",
       "packageCategory": "editors"
     },
-    "files": ["lib", "node_modules", "static"],
+    "files": ["lib/*.js", "static"],
```
lgeiger, line 97: This ignores the quite big source maps.

rgbkrk:
This is a super smart idea. Wowsers. Too bad we can't comment inside package.json to make it clear why. I have a feeling that more node_modules are being included from the @nteract packages, I'd love to look at what's still in the bundled app.

lgeiger:
> I have a feeling that more node_modules are being included from the @nteract packages, I'd love to look at what's still in the bundled app.

If you want to look into the bundled app just disable `asar` in the `electron-builder` config.

rgbkrk:
Woooow this is really trimmed down:

```
dist/mac/nteract.app/Contents/Resources/yapp$ du -sh node_modules/* | gsort -h
8.0K	node_modules/netrc
8.0K	node_modules/stream-consume
 12K	node_modules/ms
 16K	node_modules/agent-base
 16K	node_modules/https-proxy-agent
 16K	node_modules/normalize.css
 24K	node_modules/extend
 24K	node_modules/follow-redirects
 36K	node_modules/jmp
 48K	node_modules/debug
 48K	node_modules/semver
 56K	node_modules/jp-kernel
 56K	node_modules/mime
 60K	node_modules/uuid
312K	node_modules/nel
748K	node_modules/ijavascript
1.1M	node_modules/zeromq
1.5M	node_modules/github
3.0M	node_modules/mathjax-electron
3.1M	node_modules/codemirror
 16M	node_modules/nteract-assets
```

rgbkrk:
Admittedly, codemirror only needs the codemirror.css (which I know I've bundled all over the damn place for commuter and notebook preview). We only use a couple of the font types in nteract assets IIRC. I should still see if we really have to separately build `github`. 

At any rate, this is awesome!

rgbkrk:
Updated the size comparisons in the top to include the compressed dmg.

package.json
lgeiger: I'm thinking about this again. Should we include the sourcemaps in our bundled app? @rgbkrk What do you think?

rgbkrk: It does help with debugging. Then again we can always add it into a released copy into the packaged app after we want to some debugging. Doesn't help in the moment though. I'm on the fence here, leaning towards no source map in production release.

PR #1855, opened 2017-08-19
Webpack: Respect jsnext:main field from node modules

lgeiger:
Some of our dependencies ship ES2015/ES2016 versions of their package
via the `jsnext:main` field. We should make use of it.

I wonder if we should expose this in our `@nteract` modules as well.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -17,7 +17,7 @@ const options = {
     ]
   },
   resolve: {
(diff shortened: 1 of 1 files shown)


rgbkrk:
> I wonder if we should expose this in our @nteract modules as well.

We'd have to ship another folder, our current src has flow types.

PR #1860, opened 2017-08-22
[WIP] Outline what we'll do for saving

rgbkrk:
This brings over @captainsafia's work in #1147 and hopefully finishes up our document, kernel, and app lifecycle.

I've put in the same hacky thing I did in the main thread to set a module level global variable `totallyExit`. This misses quite a few bits though:

* kernels need to be properly cleaned up
  * connection file
  * the underlying kernel process and its own children
  * all connections
* we must wait for the save and give feedback for each of the edge cases
  * file hasn't been saved yet (have to do saveAs)
  * file can't be saved (permissions), need to try to save somewhere else or generally give the user a way out

I'm thinking about having specific `SAVE_AND_QUIT` and `FULLY_QUIT` actions with associated epics.

### DIFF
--- a/src/main/launch.js
+++ b/src/main/launch.js
@@ -39,6 +39,7 @@ export function launch(filename) {
   let actuallyExit = false;
 
   win.on("close", e => {
+    actuallyExit = true;
     if (!actuallyExit) {
       e.preventDefault();
       dialog.showMessageBox(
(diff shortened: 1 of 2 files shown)


rgbkrk:
@lgeiger did you already finish off parts of this?

lgeiger:
> @lgeiger did you already finish off parts of this?

I haven't touched any of the specific points listed here, but I did implement a reliable "unsaved indicator". If there is unsaved content, users are now prompted if they really want to quit.

It solves the same problem, but in the most basic/simple way possible.

rgbkrk:
Oh ok, I'll close this for now. We do need to revisit kernel cleanup as it isn't done quite properly yet.

PR #1866, opened 2017-08-27
VDOM

rgbkrk:
Bringing back the vdom mimetype from #1212.

This is (mostly) good to go. Quick demo:

gametime

Problems to fix:

* [x] ~On save, for some reason the `children` array is empty. This is completely confusing behavior considering that it's getting into the component when displayed like normal. Initial digging shows this in the state tree, at least while being deserialized.~ objectToReact is mutative
* [x] ~Invalid attributes will completely break the whole notebook, not at the `createElement` stage but at `ReactDOM.render`, which can't be caught in the component (we can use a safelist though).~ Turns out they're just warnings, I broke something else

### DIFF
--- a/packages/transform-vdom/src/index.js
+++ b/packages/transform-vdom/src/index.js
@@ -3,6 +3,8 @@ import React from "react";
 
 import { objectToReactElement } from "./object-to-react";
 
+const cloneDeep = require("lodash").cloneDeep;
+
 type Props = {
   data: Object
 };
@@ -17,7 +19,9 @@ export default class VDOM extends React.Component {
 
   render(): ?React.Element<any> {
     try {
-      return objectToReactElement(this.props.data);
+      // objectToReactElement is mutatitve so we'll clone our object
+      var obj = cloneDeep(this.props.data);
+      return objectToReactElement(obj);
     } catch (err) {
       return (
         <div>
--- a/packages/transform-vdom/src/index.js
+++ b/packages/transform-vdom/src/index.js
@@ -2,8 +2,7 @@
 import React from "react";
 
 import { objectToReactElement } from "./object-to-react";
-
-const cloneDeep = require("lodash").cloneDeep;
+import { cloneDeep } from "lodash";
 
 type Props = {
   data: Object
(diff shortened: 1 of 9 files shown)


rgbkrk:
Branch is worth trying out at least!

rgbkrk:
Figured it out, object-to-react is mutative.

rgbkrk:
I've updated the spec here to be:

```js
{
  tagName: string, // Must be an HTML tag
  attributes: Object, // attributes must be valid for setting on the element,
  children: Array<VDOMEl>,
}
```

This is to me a little more agnostic of React (the implementation here) and match more closely the language within the DOM.

packages/transform-vdom/src/index.js
```diff
@@ -0,0 +1,38 @@
+/* @flow */
+import React from "react";
+
+import { objectToReactElement } from "./object-to-react";
+
+const cloneDeep = require("lodash").cloneDeep;
```
peggyrayzis: `import { cloneDeep } from 'lodash'` to keep it consistent

packages/transform-vdom/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+type Props = {
+  data: Object
+};
+
+export default class VDOM extends React.Component {
+  props: Props;
+  static MIMETYPE = "application/vdom.v1+json";
+
+  shouldComponentUpdate(nextProps: Props): boolean {
```
peggyrayzis, line 15: do we need the sCU check since there's only one prop?

packages/transform-vdom/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+    return nextProps.data !== this.props.data;
+  }
+
+  render(): ?React.Element<any> {
```
peggyrayzis: I dont think this will ever be null so we can do `React.Element<*>`

willingc (approved):

peggyrayzis (changes requested):
some minor changes but this looks awesome 🎉

packages/transform-vdom/src/index.js
rgbkrk: We don't want this to call the `objectToReact` function repetitively (since I know that our components are not very perf optimized). Seemed like a simple quick check.

peggyrayzis: 👍

peggyrayzis (approved):
thanks for making those changes, I think this is ready to go! 😀

PR #1868, opened 2017-08-28
fix(notebook): Throttle the save event to prevent too many notifications

### LINKED ISSUE #1865: Debounce "Save" keyboard shortcut
I think we may need to debounce the Ctrl+S/Cmd+S shortcuts ? I hold the key down a second longer and I get this -

image

Also, where can I see the notebook name?

Madhu94:
Closes #1865.

I set a 2s interval because that's what we use for ``autoDismiss`` property passed to react-notification-system.

### DIFF
--- a/src/notebook/menu.js
+++ b/src/notebook/menu.js
@@ -5,6 +5,8 @@ import * as path from "path";
 
 import * as fs from "fs";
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
:cool:, let's go with it.

PR #1872, opened 2017-09-03
Add prettier npm script

lgeiger:
This adds a npm script to make it easier to run prettier on all relevant files.

If we want to ensure to have the latest version of prettier, we could also use [`prettier-greenkeeper`](https://github.com/lgeiger/prettier-greenkeeper) to run prettier on CI whenever a new version is released.

### DIFF
--- a/esdoc.json
+++ b/esdoc.json
@@ -1,5 +1,5 @@
 {
-    "source": "./src",
-    "destination": "./docs/",
-    "lint": true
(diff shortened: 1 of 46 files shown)


rgbkrk:
Thanks! ✨

rgbkrk:
I'd be fine with prettier greenkeeper.

PR #1876, opened 2017-09-06
list basic papermill metadata in notebook preview

rgbkrk:
Preview:

screen shot 2017-09-06 at 11 46 49 pm

/cc @mpacer @ewmassey

### DIFF
--- a/packages/notebook-preview/src/code-cell.js
+++ b/packages/notebook-preview/src/code-cell.js
@@ -78,8 +78,11 @@ class CodeCell extends React.PureComponent<Props> {
   }
 
   render(): ?React$Element<any> {
+    const running =
+      this.props.running ||
+      this.props.cell.getIn(["metadata", "papermill", "status"]) === "running";
     return (
-      <div className={this.props && this.props.running ? "cell-running" : ""}>
+      <div className={running ? "cell-running" : ""}>
         <PapermillView
           {...this.props.cell
             .getIn(["metadata", "papermill"], ImmutableMap())
@@ -89,7 +92,7 @@ class CodeCell extends React.PureComponent<Props> {
           <div className="input-container">
             <Inputs
               executionCount={this.props.cell.get("execution_count")}
-              running={this.props.running}
+              running={running}
             />
             <Editor
               completion
--- a/packages/notebook-preview/src/code-cell.js
+++ b/packages/notebook-preview/src/code-cell.js
@@ -22,6 +22,38 @@ type Props = {
   sourceHidden: boolean
 };
 
+type PapermillMetadata = {
+  status?: "pending" | "running" | "completed"
+  // TODO: Acknowledge / use other papermill metadata
+};
+
+const PapermillView = (props: PapermillMetadata) => {
+  if (!props.status) {
+    return null;
+  }
+
+  if (props.status === "running") {
+    return (
+      <div
+        style={{
+          width: "100%",
+          backgroundColor: "#e8f2ff",
+          paddingLeft: "10px",
+          paddingTop: "1em",
+          paddingBottom: "1em",
+          paddingRight: "0",
+          marginRight: "0",
+          boxSizing: "border-box"
+        }}
+      >
+        Executing with Papermill...
+      </div>
+    );
+  }
+
+  return null;
+};
+
 class CodeCell extends React.PureComponent<Props> {
   static defaultProps = {
     running: false,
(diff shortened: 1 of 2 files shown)


mpacer:
I don't see the Papermill metadata or is it just "Executing with Papermill" bit?

packages/notebook-preview/src/code-cell.js
```diff
@@ -48,6 +80,11 @@ class CodeCell extends React.PureComponent<Props> {
   render(): ?React$Element<any> {
     return (
       <div className={this.props && this.props.running ? "cell-running" : ""}>
+        <PapermillView
+          {...this.props.cell
+            .getIn(["metadata", "papermill"], ImmutableMap())
```
rgbkrk: Here's where the papermill metadata gets passed through (if it exists in the notebook).

packages/notebook-preview/src/code-cell.js
```diff
@@ -22,6 +22,38 @@ type Props = {
...
+          paddingLeft: "10px",
+          paddingTop: "1em",
+          paddingBottom: "1em",
+          paddingRight: "0",
+          marginRight: "0",
+          boxSizing: "border-box"
+        }}
+      >
+        Executing with Papermill...
+      </div>
```
rgbkrk, line 50: When a cell's `metadata.papermill.status` is set to `running`, we display this "banner" above a cell.

rgbkrk:
I commented inline, though you'll probably have to read them in the order I posted them rather than right at the file with the github comments.

I used the metadata from https://github.com/nteract/papermill/pull/32 to prototype this.

mpacer:
In a slightly different prototype, could you give an example of what the banner would look like in the case where one cell has been completely run, one of the middle cells with code has not yet completed execution and a third cell has not had its message dispatched? Those seem to be the three kinds of state that need to be expressed in order to give people an idea of the execution sequence (as exhibited by the difference between the status metadata values "pending", "running" and "completed").

The current prototype suggests that all of the cells have been run and so it's not clear what "Executing with Papermill…" entails. 

Is it the first cell that is currently being executed? Then why haven't the outputs from the later cells been cleared and why do the later cells have execution counts? 

If it is indicating that some of the cells are being executed… which cells are in the process of execution? 

If all the cells have already been executed, why does it still display the banner at all? Is it merely stating that the execution was organised by papermill rather than via a standard notebook UI?

rgbkrk:
One terrible thing with my prototype above is that I faked the notebook. I wasn't actually running papermill, I crafted the metadata that would say which cell was running and uploaded it to a gist. This notebook, if run through papermill would look like:

```
[1]: CODE
     OUTPUT

--- Executing with Papermill ---
[*]: CODE RUNNING
     OUTPUT

[ ]: CODE WE RUN LATER
     OUTPUT
```

papermill writes to disk (or straight to S3) after each cell finishes, which is why we wanted this intermediate state. 😉

rgbkrk:
I felt like "pending" and "completed" were duplicative information, considering `null` entries for execution count are "pending" and numbered execution counts are "completed". These fields are only supposed to exist at runtime IIRC @ewmassey.

rgbkrk:
Let me fix the prototype notebook since I chose metadata setting arbitrarily.

rgbkrk:
Alright, I changed the graphic which is more representative. It's not perfect, it's certainly workable though. This does tell me I should make the `*` appear when `papermill.status === "running"`

rgbkrk:
Oh! One thing that may not be clear here @mpacer! The component being modified here is _only_ the notebook preview -- it doesn't affect the desktop app whatsoever -- this is for use in [commuter](https://github.com/nteract/commuter) and other web apps that use this component for rendering a notebook from some storage location.

mpacer:
Shall I guess that the `*` is added by other machinery not touched by this PR (keyed to replace the execution count with the class "cell-running")? Because I don't see that being in the code you modified. But yes… this makes much more sense… my only remaining question is why the "Executing with Papermill…" is necessary once you have the ability to reinsert the `*` to the execution count without relying on its existence in the notebook file.

rgbkrk:
The `<inputs>` component has a boolean flag that I updated to use the papermill status here

```js
             <Inputs
                executionCount={this.props.cell.get("execution_count")}
                running={running}
              />
```

Inside of that component, it puts a nice little star (or could be modified for any UI really).

rgbkrk:
> My only remaining question is why the "Executing with Papermill…" is necessary once you have the ability to reinsert the * to the execution count without relying on its existence in the notebook file.

The users I work with have hundreds of notebooks each, any of which can be currently running. This gives it a slight more amount of lineage about where it's running (or who/what is running it).

PR #1878, opened 2017-09-08
flow type all the epics (minimally)

rgbkrk:
Started some minimal flow typing of the epics. I welcome any changes that seem easy to add on here. Note that `actions` is still not typed, so once these all get wrapped up hopefully we could make a proper `NStore` and `NAction` type to use across our reducers and epics.

I'm really curious about having multiple stores or stitching these together in different ways so we can cross the divide of Desktop app and Web app. There were some ideas that @peggyrayzis and @kkemple had about making our own namespace, something like what Apollo does.

I swear I'm getting closer to exporting our epics and reducers for use in other apps, I swear.


src/notebook/epics/execute.js
```diff
@@ -134,11 +142,17 @@ export function createCellAfterAction(id, setInputStream) {
...
-    .pluck("content", "execution_state")
-    .map(status => updateCellStatus(id, status));
+export function createCellStatusAction(
+  id: string,
+  cellMessages: Observable<*>
+) {
+  return (
+    cellMessages
+      // $FlowFixMe: We patched this onto the prototype
+      .ofMessageType(["status"])
```
rgbkrk, line 152: @jayphelps is there any way to pick up addons to the prototype for Observables? maybe I should handle these differently?

jayphelps: Not that I know of. I once tried to make flow types for rxjs's operator patching but I couldn't find a way to do this. It will always complain about "Covariant property incompatible with contravariant use".

Looking again, I found someone who says that this prototype-patching style intentionally not supported--but they might not speak for the real official position.

https://github.com/facebook/flow/issues/4643#issuecomment-323551894

> Flow is intentionally not focused on typing pre-class, prototypical inheritance code

If the rxjs community brought this use case up directly they might reconsider, but it really depends on their reasons and limitations. Even with the differences I'm aware of between flow and TypeScript, it's not immediately clear what they mean by "unsafe" as it seems to be relatively safe in TS. Maybe they just mean philosophically the act of modifying an existing type in-place is more unsafe than not allowing it; which is true and isn't ideal, but this is JS so.. 😜

rgbkrk: :smile:

We could probably just make `ofMessageType` be a function that takes an observable and the message type, returning an observable.

jayphelps: Yeah, dunno if you've heard about the [whole lettable operators work](https://github.com/ReactiveX/rxjs/pull/2667) that is coming shortly but you could follow the same approach.

PR #1879, opened 2017-09-10
fs observable

rgbkrk:
This extracts our `utils/fs` to a module we can generally use as `fs-observable`.

### DIFF
--- a/packages/fs-observable/package.json
+++ b/packages/fs-observable/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "fs-observable",
+  "version": "1.0.0",
+  "description": "Observables for node fs",
+  "main": "lib/index.js",
+  "nteractDesktop": "src/index.js",
+  "scripts": {
+    "prepare": "npm run build",
+    "build": "npm run build:clean && npm run build:lib && npm run build:flow",
+    "build:clean": "rimraf lib",
+    "build:flow": "flow-copy-source -v -i '**/__tests__/**' src lib",
+    "build:lib": "babel -d lib src --ignore '**/__tests__/**'",
+    "build:lib:watch": "npm run build:lib -- --watch"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nteract/nteract.git"
+  },
+  "keywords": ["rxjs", "fs", "observable"],
+  "author": "Kyle Kelley <rgbkrk@gmail.com>",
+  "license": "BSD-3-Clause",
+  "bugs": {
+    "url": "https://github.com/nteract/nteract/issues"
+  },
+  "homepage":
+    "https://github.com/nteract/nteract/tree/master/packages/fs-observable",
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "rxjs": "^5.4.1"
+  },
+  "peerDependencies": {
+    "rxjs": "^5.4.1"
+  }
+}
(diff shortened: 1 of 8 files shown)


rgbkrk:
Looks like introducing `flow` here highlighted an issue (possibly):

```
Error: src/notebook/epics/config.js:69
 69:         JSON.stringify(store.getState().config.toJS())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unused function argument
   62:   static bindNodeCallback<T, U>(callbackFunc: (v1: T, callback: (err: any, result: U) => any) => any, selector?: void, scheduler?: rxjs$SchedulerClass): (v1: T) => rxjs$Observable<U>;
                                                                                                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type expects no more than 1 argument. See lib: flow-typed/npm/rxjs_v5.0.x.js:62

Error: src/notebook/epics/saving.js:43
               v-----------------
 43:           stringifyNotebook(
 44:             toJS(
 45:               action.notebook.setIn(
...:
 50:           )
               ^ unused function argument
   62:   static bindNodeCallback<T, U>(callbackFunc: (v1: T, callback: (err: any, result: U) => any) => any, selector?: void, scheduler?: rxjs$SchedulerClass): (v1: T) => rxjs$Observable<U>;
                                                                                                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type expects no more than 1 argument. See lib: flow-typed/npm/rxjs_v5.0.x.js:62
```

rgbkrk:
Seems that the signature of `fs.writeFile` doesn't get propagated (is that the word I should use here) into the `bindNodeCallback` return.

packages/fs-observable/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "fs-observable",
```
jayphelps, line 2: @trxcllnt is the owner of [rxjs-fs](https://www.npmjs.com/package/rxjs-fs), which appears mostly defunct (last release 4 years ago). So if you plan to release this package maybe he would be willing to part with that package name?

rgbkrk: rxjs-fs covers a much larger surface area (that we wish was available here), while also having some nice helpers. I'm not sure we'd be ready to take over with `rxjs-fs`, especially since it really needs to be brought up to RxJS 5 (https://github.com/trxcllnt/rxjs-fs/issues/2). Happy to coordinate on that!

trxcllnt: @rgbkrk I have a branch of an RxJS 5 rewrite already started: https://github.com/trxcllnt/rxjs-fs/tree/rxjs5. I'd made more progress (some notes [here](https://github.com/trxcllnt/rxjs-fs/issues/2)), but unfortunately I hadn't pushed to gh yet and lost it last week when my laptop was stolen. Lmk if you're interested in collaborating. I'm also fine transferring the name if it's going to a good home.

rgbkrk: At this point in time I can't focus much on our desktop app which extends to not needing to do much on this module yet. I think I'd rather collaborate on yours, as well as generally learn from you. An initial glance at the rxjs5 branch makes me pretty hopeful.

rgbkrk: For the external reader, "can't focus much on our desktop app" means "desktop app doesn't get as much love as the current web app bits, I'll be back to doing more for desktop :soon:"

PR #1880, opened 2017-09-11
Implement top level state management for the web app

rgbkrk:
This PR is going to serve as a guide for how we'll break off our records, epics, and reducers. As I carve things out, I'll end up making PRs that simplify our current redux architecture as much as possible. The main goal is to keep the desktop and web app working in tandem with each other.

---------------

Update!

This now carves out the basics of state management in tandem with the jupyter config data.

* epic for loading from the contents API
* easy-mode reducer with contents API response
* provider and view components for contents


rgbkrk:
This is now coming along swimmingly. I think I know how we'll carve more of these out, which means I could probably group this work for others to tackle if they'd like.

rgbkrk:
This now does the full round trip with the contents API, putting data back into the store. I've been running on a subpath to make sure that we handle JupyterHub deployments successfully. Hooking up all the other APIs is fairly trivial here since `rx-jupyter` already has all of these implemented. Most of the work is creating the epics and figuring out how our state tree will look.

rgbkrk:
I've got a critical decision about how we do routing next -- do I adopt the same router from next's router since we use `next/link` in commuter?

rgbkrk:
Looks like there is some flow to address in `rx-jupyter`. I'll fix that over there.

rgbkrk:
What do you think of the epic structure here @jayphelps?

packages/nbextension/nteract_on_jupyter/epics/contents.js
```diff
@@ -0,0 +1,121 @@
...
+      if (xhr.status === 200) {
+        return {
+          type: "LOADED",
+          payload: xhr.response
+        };
+      } else {
+        return {
+          type: "LOAD_FAILED",
+          payload: xhr.response
+        };
```
jayphelps: this will sometimes map a LOAD_FAILED action into a nearly identical LOAD_FAILED action, if an exception was caught inside the above switchMap where you `Observable.of(loadFailed(xhrError))`. This feels a bit brittle as someone might come in here later and not realize that "xhr" does not always mean xhr (it might be that action).

How I usually handle things like this by checking inside a `do` and throwing, basically the same way you did it above for the `!action.path` check. Then just use a single catch

```js
return contents
  .get(serverConfig, action.path, {})
  .do(xhr => {
    if (xhr.status !== 200) {
      throw new Error(xhr.response);
    }
  })
  .map(xhr => ({
    type: "LOADED",
    payload: xhr.response
  })
  .catch((xhrError: any) => Observable.of(loadFailed(xhrError)));
```

packages/nbextension/nteract_on_jupyter/store.js
```diff
@@ -0,0 +1,26 @@
...
+    case "LOADED":
+      return Object.assign({}, state, { contents: action.payload });
+  }
+  return state;
+};
+
+type AppState = {};
+
+export default function configureStore(initialState: AppState = {}) {
+  const rootEpic = combineEpics(loadEpic);
```
jayphelps: looks like you exported an array of all epics in `packages/nbextension/nteract_on_jupyter/epics/index.js` but it's not used here. FYI

jayphelps:
> What do you think of the epic structure here @jayphelps?

If you mean like directory/file structure, looks good to me 👍

packages/nbextension/nteract_on_jupyter/epics/contents.js
rgbkrk: Oh whoa, thanks for catching this.

rgbkrk:
> If you mean like directory/file structure, looks good to me 👍

I meant the structure of the epic, and you found issues in there that I hoped you'd uncover. 🙇

packages/nbextension/nteract_on_jupyter/store.js
rgbkrk: That was totally me spelunking to figure out why things weren't configured right. I'll put that back together.

lgeiger:
This looks great!

One minor thing I found: `react-redux` is missing in the dependencies.

rgbkrk:
Yet another thing I've cheated on by having it loaded from the base desktop app!

jayphelps:
@rgbkrk Looks like these are missing too: redux, rxjs, react-notification-system.

I think the best way to prevent these sorts of things is to move everything that isn't a common build dependency into its own package, instead of having stuff in src/

rgbkrk:
We ran into rebuild problems with the desktop app when I moved it into a lerna package, which is why it's at the top level still. :(

I'll get those deps in.

PR #1885, opened 2017-09-13
Don't run build script on install

lgeiger:
Running the entire build script on install is often unnecessary and takes a lot of time.

This should speedup the CI builds since we don't need the full build for all jobs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
-    "postinstall": "electron-builder install-app-deps",
+    "postinstall": "npm run lerna && electron-builder install-app-deps",
     "reinstall": "npm run clean && rimraf node_modules && npm install",
     "prestart": "npm run build:webpack",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,11 +26,11 @@
     "lint:fix": "eslint . --fix",
     "prebuild": "rimraf lib",
     "lerna": "lerna bootstrap --hoist",
-    "install": "npm run lerna && npm run build",
     "build": "npm run build:packages && npm run build:webpack",
     "build:webpack": "webpack --progress --colors",
     "build:webpack:watch": "npm run build:webpack -- --watch",
-    "build:packages": "lerna run build --ignore @nteract/notebook-preview-demo",
+    "build:packages":
+      "lerna run build:lib --ignore @nteract/notebook-preview-demo",
```
rgbkrk, line 33: Now that we're on next.js, it's probably not too bad to build notebook-preview-demo. No big deal here though.

lgeiger: 👍 Happy to change that.

Issue #1892, opened 2017-09-14
Not all package tests are run on CI 


lgeiger:
While going through #1891 I discovered that some tests of recently brought in packages aren't running on CI.
We should add a script to run them or move them over to `jest` ([`jest-codemods`](https://github.com/skovhus/jest-codemods) are very helpful)

- [x] `ansi-to-react` (done in #1891)
- [x] `enchannel`
- [ ] `enchannel-zmq-backend`
- [x] `fs-observable`

PR #1895, opened 2017-09-14
Move fs-observable tests into right package and convert to jest

lgeiger:
Adress parts of #1892


rgbkrk:
You're on a tear today!

PR #1897, opened 2017-09-14
Convert enchannel-zmq-backend specs to jest

### LINKED ISSUE #1892: Not all package tests are run on CI

lgeiger:
Now all test properly run on CI. Fixes #1892 

I wasn't able to properly mock jmp/zeromq so a small amount of rebuilding is needed. Luckily the Electron app still works after running the tests without rebuilding (probably because we're not hoisting so it requires zeromq from a different location).


rgbkrk:
Oh yeah, our old mock setup used mock-require and was doing it with relative paths.

rgbkrk:
I'm not sure what's up with codecov saying CI failed. Does it seem like a fluke and that we can just merge this in?

package.json
```diff
@@ -35,7 +35,8 @@
     "test:coverage": "npm run coverage",
     "test:lint": "npm run lint",
     "test:flow": "npm run flow",
-    "test:packages": "jest",
+    "test:packages":
+      "lerna exec --scope enchannel-zmq-backend -- npm rebuild && jest",
```
rgbkrk, line 39: At least it's explicit here for now. :smile:

lgeiger:
Reseted the Appveyor cache. This should now be good to go.

For the desktop app, I think it's best to wait for [sinon support in jest-codemods](https://github.com/skovhus/jest-codemods/issues/68), before converting fully moving to jest.

lgeiger:
> Oh yeah, our old mock setup used mock-require and was doing it with relative paths.

Which mock setup are you talking about?
`enchannel-zmq-backend` tests did never run after we brought them into the monorepo and at the `nteract/enchannel-zmq-backend` they just used regular zeromq.

rgbkrk:
Oh interesting, oops.

Within the desktop app's tests: https://github.com/nteract/nteract/blob/master/packages/desktop/test/setup.js#L82-L87

lgeiger:
Oh I see, thanks

PR #1898, opened 2017-09-15
Use synchronous close dialog to fix cmd+q

lgeiger:
Fully closing nteract with `cmd+q` doesn't work right now because `win.close();`won't fully quit the app. This adds two steps for fully closing nteract.

This PR prevents that by using a synchronous close dialog which will block the process. I don't think blocking here is too bad, but if you like I could search for another solution.


rgbkrk:
Yeah, this is definitely the right behavior now. Especially since we do similar handling in the notebook process.

PR #1899, opened 2017-09-15
Kernelspecs! (web app)

rgbkrk:
Time to get down to brass tax and bring in compute!

### DIFF
--- a/packages/nbextension/nteract_on_jupyter/actions/contents.js
+++ b/packages/nbextension/nteract_on_jupyter/actions/contents.js
@@ -1,0 +1,29 @@
+// @flow
+
+export type SAVE_ACTION = {
+  type: "SAVE",
+  path: string,
+  model: any
+};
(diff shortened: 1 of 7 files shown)


rgbkrk:
Hurrah, flow is passing! I'm adding something else on now to make install work again since I rebased against master and `webpack` was missing as a dependency here.

rgbkrk:
Hmmmm, seems the rx hoisting is now breaking things instead for this app. 🤔

rgbkrk:
> `rx_jupyter_lib.contents.get(...).do is not a function`

I am sad.

rgbkrk:
Wow looking over this after we've done all the pipelining, I kind of want to just grab the pieces I want  here and make a new PR instead of fixing this up.

rgbkrk:
🌟 Now updated to match the lettable style 🌟

Issue #1886, opened 2017-09-13
Overeager check for unsaved changes before quit


pgcudahy:
Even if I've just saved and haven't made any changes before quitting I still get the "Unsaved data will be lost. Are you sure you want to quit?" dialog. Can the application check to see if there are unsaved changes before prompting?

rgbkrk:
There is some work in progress in #1860. It's definitely a bit annoying right now, as you always have to dismiss it. I put the current version in as a safeguard until we have the proper UX.

PR #1901, opened 2017-09-15
Fix "modified indicator" on mac

lgeiger:
This PR fixes the "modified indicator" (don't know how to call the dot in the close button) on macOS and paves the way towards fixing #1886.

save

Main changes in this PR are:
- Store the last saved Notebook in the document store. I don't know if this is the right thing to do in terms of performance and memory consumption, but it was the easiest to implement.
- Don't display saving notification on Mac since we now have a working saving indicator.

### DIFF
--- a/packages/desktop/src/notebook/actions.js
+++ b/packages/desktop/src/notebook/actions.js
@@ -327,8 +327,9 @@ export function saveAs(filename: string, notebook: any) {
   };
 }
 
-export function doneSaving() {
+export function doneSaving(notebook: any) {
   return {
-    type: constants.DONE_SAVING
+    type: constants.DONE_SAVING,
+    notebook
   };
 }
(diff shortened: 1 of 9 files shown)


packages/desktop/src/notebook/actions.js
```diff
@@ -327,8 +327,9 @@ export function saveAs(filename: string, notebook: any) {
   };
 }
 
-export function doneSaving() {
+export function doneSaving(notebook: any) {
   return {
-    type: constants.DONE_SAVING
+    type: constants.DONE_SAVING,
+    notebook
```
rgbkrk, line 333: Wow this came out way simpler than I expected, if a little indirect.

rgbkrk: I was thinking it would be better to do this in the reducer for `SAVE`, then I realized that by doing it as a result of successful save we know if it _actually_ saved. Brilliant!

rgbkrk (approved):
Hero!

PR #1902, opened 2017-09-15
Don't show close warning when file is already saved

### LINKED ISSUE #1090: Ask user to save before global unload function

### LINKED ISSUE #1445: Are You Sure Dialog for destructive commands
I just learnt the hard way  that `Ctrl-R` will reload nteract which will nuke any unsaved work :sob: 

To save fat-fingered users from themselves it might be nice to have a pop-up Ok/Cancel dialog before reloading the page. To not annoy people who do want to reload the page, the Ok could be highlighted by default so that you could reload with `Ctrl-R`, `Enter`.

Similarly, for `Ctrl-W` or clicking on the window close button - both of which will kill nteract whether or not you have saved your work.

Potentially the dialog could only be shown if there were unsaved changes.

### LINKED ISSUE #1886: Overeager check for unsaved changes before quit

lgeiger:
This is based on #1901 so only the last commit is relevant for review. It will prevent the dialog to be shown if file is already saved.

This is the new workflow:
close dialog

Fixes #1886 
Fixes #1090
Fixes #1445

### DIFF
--- a/packages/desktop/src/main/launch.js
+++ b/packages/desktop/src/main/launch.js
@@ -36,14 +36,14 @@ export function launch(filename) {
   const index = path.join(__dirname, "..", "static", "index.html");
   win.loadURL(`file://${index}`);
 
-  win.on("close", e => {
+  win.webContents.on("will-prevent-unload", e => {
     const response = dialog.showMessageBox({
       type: "question",
       buttons: ["Yes", "No"],
       title: "Confirm",
       message: "Unsaved data will be lost. Are you sure you want to quit?"
     });
-    if (response == 1) {
+    if (response == 0) {
       e.preventDefault();
     }
   });
--- a/packages/desktop/src/notebook/global-events.js
+++ b/packages/desktop/src/notebook/global-events.js
@@ -3,6 +3,8 @@
 import type { Store } from "redux";
 import type { AppState } from "./records";
 
+import { dialog } from "electron";
+import { is } from "immutable";
 import { forceShutdownKernel } from "./kernel/shutdown";
 
 export function unload(store: Store<AppState, Action>) {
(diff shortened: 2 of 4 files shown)


rgbkrk:
I'm beginning to regret my decision to rebase merge rather than do a merge commit. You'll have to rebase this against master now since I merged the other.

rgbkrk:
We can hyper optimize later by only storing a hash of the notebook object. Since our Immutable.js structure is impure (the mutable data objects within display_data), I'm not sure if the `iterable.hashCode()` will be valid.

rgbkrk:
Sure enough, can't use `hashCode` for fast comparisons because of the mutable leaf nodes:

```js
> m = Immutable.Map({'a': {'b': 2}}) // the top level is immutable here
Map { "a": [object Object] }
> m.get('a') // The inner object is mutable
{ b: 2 }
> m.get('a')['b']
2
> m.hashCode()
-356504055
> m.get('a')['b'] = 3
3
> m.hashCode()
-356504055
```

Maybe it's not a big deal since display updates actively change that entire data payload.

lgeiger:
> I'm beginning to regret my decision to rebase merge rather than do a merge commit. You'll have to rebase this against master now since I merged the other.

Rebased it 🚀 

> We can hyper optimize later

Yes, storing the whole notebook two times seem kind of redundant. I don't know if there is a canonical for these kind of tasks in redux.

rgbkrk:
> I don't know if there is a canonical for these kind of tasks in redux.

## Non-redux, had we stuck with full immutable

If the full structure was Immutable.js, then you'd only store the `notebook.hashCode()` for the saved notebook and compare it to the `notebook.hashCode()` for the current notebook. I'm only noting this to show what the benefit of truly immutable data structures are (when deeply immutable).

## Redux way (proposal)

I _think_ the way to do this is to register certain kinds of changes as `dirty` for the notebook. Then, everytime someone has a successful save we set `dirty = false` (in the state tree).

lgeiger:
> I _think_ the way to do this is to register certain kinds of changes as `dirty` for the notebook. Then, everytime someone has a successful save we set `dirty = false` (in the state tree).

This sounds like a good approach to me.

Issue #620, opened 2016-08-21
Setup a default kernel based on user


rgbkrk:
/cc @glyph who has a bunch of custom kernels and not the magical Python3 default kernel we expect.

glyph:
My custom kernels are mostly just different virtualenvs for different projects, with different things installed into them.

rgbkrk:
Very legitimate.

On a first usage, would you say it's fair to pick the first kernel we see and from there on users can pick a default kernel?

glyph:
@rgbkrk That would be totally fine with me. Although it would be nice to have a defined "first" - lexicographical order maybe? :)

rgbkrk:
This was "done" in #763 as a useful side effect of our app startup. Still want to think on configuring the default kernel as well as the order (we're currently at the mercy of `Object.keys`, which means we're at the mercy of property order).

rgbkrk:
@captainsafia - here's a good one to add to the new config setup - default kernel (by name)

AlJohri:
@rgbkrk one really nice feature of the jupyter web interface is that using the CLI, if I run: `jupyter notebook analysis.ipynb` within a virtualenv, it will automatically use the virtualenv's kernel without having to run `python -m ipykernel install`. I find that running `ipykernel install` can quickly end up with a ton of kernels in nteract's drop down menu since I use a virtualenv for each project (related #1041, https://github.com/nteract/nteract/issues/1576#issuecomment-331669619)

is there any way that running `nteract analysis.ipynb` within a virtualenv could automatically detect that its being run from within a virtualenv and start with the virtualenv's kernel if it exists? I know that running `jupyter-kernelspec list` within the virtualenv is able to find it:

```
$ pipenv run jupyter-kernelspec list
Available kernels:
  python3    /Users/johria/.local/share/virtualenvs/hss-gPwjJtJU/share/jupyter/kernels/python3
```

I do understand that this is a little more of an advanced user edgecase but, it would be great to sort of implicitly support this kind of usage when triggered from the CLI but don't surface it in the GUI in anyway. so in the end, I can run something like:

```
envdir envs/local pipenv run nteract analysis.ipynb
```

[...]

PR #1906, opened 2017-09-24
[MRG] Don't use shellEnv when launched from terminal

betatim:
Related to #620 

This is pretty hacky but does mean you get a kernel that corresponds to the conda environment in which you typed `npm run start`. There are two things that puzzle me:

1. the `process.env` print out I inserted shows various custom env variables I set in my terminal, as well as `CONDA_blah` variables that belong to the current conda env, but it shows a different PATH. So it seems to inherit some variables but does something special to PATH. Any ideas where that is happening?

1. only seems to work with `ipykernel~=4.6.x`, at least in some of my envs where the version is 4.3.x executing a cell never completes. Just seems to hang and I can't find an error message anywhere.

If we can track down when/where PATH gets modified maybe we can prevent it from removing (or re-add) the path for the current conda environment. Maybe the real way to do this is to add a kernel spec like this if we detect that we are in either a virtualenv or a conda env and do "the right thing" for either to work.

WDYT?

### DIFF
--- a/packages/desktop/src/main/prepare-env.js
+++ b/packages/desktop/src/main/prepare-env.js
@@ -10,7 +10,10 @@ import "rxjs/add/operator/publishReplay";
 const env$ = Observable.fromPromise(shellEnv())
   .first()
   .do(env => {
-    Object.assign(process.env, env);
+    // no need to change the env if started from the terminal on Mac
+    if (process.platform !== "darwin" || !process.env.TERM) {
+      Object.assign(process.env, env);
+    }
   })
   .publishReplay(1);
 


rgbkrk:
> only seems to work with ipykernel~=4.6.x, at least in some of my envs where the version is 4.3.x executing a cell never completes. Just seems to hang and I can't find an error message anywhere.

~~That might be related to the whole `ipykernel_launcher` switchover.~~ yes yes indeed

packages/desktop/src/main/kernel-specs.js
```diff
@@ -25,11 +25,24 @@ const KERNEL_SPECS = {
...
     }
+  },
+  python_nteract: {
+    name: "python_nteract",
+    spec: {
+      argv: ["python", "-m", "ipykernel_launcher", "-f", "{connection_file}"],
+      display_name: "python (nteract)",
+      language: "python",
+      env: {
+        ELECTRON_RUN_AS_NODE: "1",
```
rgbkrk: You won't need this here since this was just something we passed through to enable our embedded JS kernel.

packages/desktop/src/main/kernel-specs.js
```diff
@@ -25,11 +25,24 @@ const KERNEL_SPECS = {
         ELECTRON_RUN_AS_NODE: "1"
       }
     }
+  },
+  python_nteract: {
+    name: "python_nteract",
+    spec: {
+      argv: ["python", "-m", "ipykernel_launcher", "-f", "{connection_file}"],
+      display_name: "python (nteract)",
+      language: "python",
```
rgbkrk: We really need this to get merged successfully in with the user's default python if they happen to be the same path as you'll want the notebook to point to the right kernelspec in the document (here it would identify as `python_nteract`)

rgbkrk:
> Maybe the real way to do this is to add a kernel spec like this if we detect that we are in either a virtualenv or a conda env and do "the right thing" for either to work.

Sounds about right.

rgbkrk:
For a random user with a standalone install on a mac who is using anaconda, their flow might end up looking like this:

* Install desktop app
* Launch by double-clicking a notebook (yes outside your use case)
* This kernelspec will (possibly) be their system python (ew)
* "It's named nteract, is this the right one?"
* Be confused
* Learn about the "right way"
* Complain that this didn't make things simpler

I'm not actually sure so I think this _could_ be destructive at the moment. I think there's a decent way to do this that is going to involve checking some environment variables as well as seeing which version of ipython is installed (I did some of this in a branch I closed a while ago). However... I think there's a way to do this for today in a way that makes the `nteract` shell script work ok in a way that the naming (hopefully) makes sense:

## Special case CLI launches

When the nteract app is started from the CLI, set a special environment variable that says `NTERACT_SOURCE_FROM_ENV_AT_CLI=1` (I'm being a little verbose). Then, make this kernel's display name be "python (current env)". It can also be a little more aware of whether it's in a virtual environment or a conda environment, showing a bit more information.

Generally that would make me a little more comfortable for one of these semi-generated kernelspecs.

packages/desktop/src/main/kernel-specs.js
betatim: Double checking: You mean we should set the language to something else aka `python_nteract`?

rgbkrk: I'm trying to say that if this kernelspec is exactly the same as their `python` kernelspec (by name -- folder on disk, key in the overall kernelspecs object here).

However, I'm thinking that naming it by env and only populating it when we know we're launched from the terminal may be a better approach (see below).

betatim:
Wouldn't checking for `CONDA_PREFIX` in the environment take care of the `NTERACT_SOURCE_FROM_ENV_AT_CLI=1`? It is only set if you `source activate some-env` so if `nteract` is started by double clicking it won't be set. So I'd propose to only add this kernel spec if we detect/are convinced we are being started "inside" a conda environment. Should also work out what `virtualenv` does which is the other popular way of having virtual environments.

rgbkrk:
> So I'd propose to only add this kernel spec if we detect/are convinced we are being started "inside" a conda environment.

👍 

> Wouldn't checking for CONDA_PREFIX in the environment take care of the NTERACT_SOURCE_FROM_ENV_AT_CLI=1?
> Should also work out what virtualenv does which is the other popular way of having virtual environments.

Combine these two questions/statements and this is why I'm suggesting you set a var to communicate the intent of launching from the CLI. You can make the "Python (current env)" spec use whatever Python is on the path here.

rgbkrk:
> If we can track down when/where PATH gets modified maybe we can prevent it from removing (or re-add) the path for the current conda environment.

Right here: https://github.com/nteract/nteract/blob/d813188524733b825b3150b7be833912246c3d22/packages/desktop/src/main/prepare-env.js

rgbkrk:
I'm going to add some notes to that file to state _why_ we're doing certain things before they're lost to time. 😛

rgbkrk:
Damn it, I just committed straight to master in the GitHub UI instead of making a PR like I meant to for the the comments on the env prep. See the updated version here: https://github.com/nteract/nteract/blob/d84fcc3d2ac3b16987ba3b6548fdeda76feda91d/packages/desktop/src/main/prepare-env.js

betatim:
Some notes to (mainly) self: when launching the app by double click `shellEnv` gives me the following for `PATH`:
```
/Users/thead/.nvm/versions/node/v5.11.0/bin:./bin:/usr/local/bin:/usr/local/sbin:/Users/thead/.dotfiles/bin:/Developer/NVIDIA/CUDA-8.0/bin:/Users/thead/anaconda/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/opt/X11/bin:/Library/TeX/texbin:/node_modules/.bin:/Users/thead/git/nteract/nteract/packages/desktop/dist/mac/nteract.app/Contents/MacOS:/usr/texbin:/Users/thead/.rvm/bin
```
This is also what I get if I start a new terminal. So far so good.

Start from terminal, shellEnv gives me:
```
/Users/thead/.nvm/versions/node/v5.11.0/bin:./bin:/usr/local/bin:/usr/local/sbin:/Users/thead/.dotfiles/bin:/Developer/NVIDIA/CUDA-8.0/bin:/Users/thead/anaconda/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/opt/X11/bin:/Library/TeX/texbin:/Users/thead/git/nteract/nteract/packages/desktop/node_modules/.bin:/Users/thead/git/nteract/nteract/packages/node_modules/.bin:/Users/thead/git/nteract/nteract/node_modules/.bin:/Users/thead/git/nteract/node_modules/.bin:/Users/thead/git/node_modules/.bin:/Users/thead/node_modules/.bin:/Users/node_modules/.bin:/node_modules/.bin:/Users/thead/git/nteract/nteract/packages/desktop/node_modules/electron/dist/Electron.app/Contents/MacOS:/Users/thead/.nvm/versions/node/v8.1.3/lib/node_modules/npm/bin/node-gyp-bin:/Users/thead/anaconda/envs/skopt/bin:./bin:/usr/local/sbin:/Users/thead/.dotfiles/bin:/Developer/NVIDIA/CUDA-8.0/bin:/Users/thead/anaconda/bin:/usr/texbin:/Users/thead/.rvm/bin:/usr/texbin:/Users/thead/.rvm/bin
```
this somehow duplicates stuff. As a result the conda env path is towards the back and get's shadowed by the `/usr/bin` etc from the first set of paths.

So I (currently) think that all that needs to happen is to not use shellEnv when we detect that nteract was started from a terminal. If I by hand comment out the relevant bit in `prepare-env.sh` I indeed get a "python 3" kernel that matches my conda env. That is without the kernelspec hackery in this PR!

betatim:
Checking if `TERM` is set is the best idea I have for detecting if we were launched from a terminal or not. Apparently you can also check if stdin is a file/tty??

The detection works for me on Mac. 🎉

lgeiger:
> Checking if `TERM` is set is the best idea I have for detecting if we were launched from a terminal or not.

I think `process.argv` *should* be different in the main process if it's started via the CLI.

packages/desktop/src/main/prepare-env.js
```diff
@@ -10,7 +10,9 @@ import "rxjs/add/operator/publishReplay";
 const env$ = Observable.fromPromise(shellEnv())
   .first()
   .do(env => {
-    Object.assign(process.env, env);
+    if (!process.env.TERM) {
```
rgbkrk: I'm ok with merging this if it's working on your mac if you make the conditional `process.platform !== 'darwin' || !process.env.TERM` (or you can DeMorgan's Law this and do `!(process.platform === 'darwin' && process.env.TERM)`)

betatim:
Waiting for travis, then good to go.

betatim:
I got the following three values for `process.argv`: launch with `npm run star`: ` [ '/Users/thead/git/nteract/nteract/packages/desktop/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron',
  '.' ]`

launch with `open packages/desktop/dist/mac/nteract.app`:
`[ '/Users/thead/git/nteract/nteract/packages/desktop/dist/mac/nteract.app/Contents/MacOS/nteract' ]`

and when double clicking `nteract.app` from Finder:
`['/Users/thead/git/nteract/nteract/packages/desktop/dist/mac/nteract.app/Contents/MacOS/nteract' ]
`

The last two cases are the ones we'd want to tell apart, so sad panda.

packages/desktop/src/main/prepare-env.js
```diff
@@ -10,7 +10,10 @@ import "rxjs/add/operator/publishReplay";
 const env$ = Observable.fromPromise(shellEnv())
   .first()
   .do(env => {
-    Object.assign(process.env, env);
+    // no need to change the env if started from the terminal on Mac
+    if (process.platform !== "darwin" || !process.env.TERM) {
```
rgbkrk, line 14: 👍

rgbkrk (approved):

rgbkrk:
Thanks for tackling this and going back+forth on it Tim to find the simplest way to solve this.

betatim:
Years of physics training make me be unhappy if the solution is too complex :p

PR #1907, opened 2017-09-24
[WIP] VDOM + models

rgbkrk:
I've been experimenting with something along with @mpacer and @willingc, just posting this PR to get something up. I'll fill this out to be more rigorous and tested since it's going to be a pretty impactful change, in terms of the supported API.

# Goal

Create a simple way to create interactive forms and layouts that will work across all Jupyter frontends that is compatible with real-time (consistent serializable state tree + diffs that can be propagated).

# Related PRs:

* Models #1313

# Core concepts

* When a VDOM output contains `data-path-for-ATTRIBUTE`, it replaces the actual `ATTRIBUTE` with a value from a model (as in #1313)
* When the value is bound, we can set an implicit `onChange` handler (need to spec this out a bit)

This is fairly similar to how [React's (deprecated) `LinkedStateMixin` worked](https://facebook.github.io/react/docs/two-way-binding-helpers.html)

### `onChange` implicit bindings

Status: 🤔 

To start things off, we're going to allow changing form fields that use `name` and `value`, working in tandem with the `setIn` based models from #1313.

Rough sketch is that when an `<input>` is changed, we bubble up the event as an action that will both set the local model as well as send it to the kernel (to change the "server side" model).

### `onClick` implicit bindings

Not scoped for this PR (or any :soon: release), I want only actions that can be viewed if there were more than one viewer at the same time.

My current thinking on this is that we possibly bubble the `onClick` event up for a given `name` (which means only supporting `<button>`s for the moment).

### Specifying the model to use

Status: 🤔 

Potential options:

* Within the transient field of the vdom output, set a `model_id`

### DIFF
--- a/packages/desktop/example-notebooks/model-debug.ipynb
+++ b/packages/desktop/example-notebooks/model-debug.ipynb
@@ -1,116 +1,145 @@
 {
- "cells": [
-  {
-   "cell_type": "code",
-   "execution_count": 1,
-   "metadata": {
-    "collapsed": false,
-    "outputExpanded": false
-   },
-   "source": [
-    "import IPython"
-   ],
-   "outputs": []
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 2,
-   "metadata": {
-    "collapsed": false,
-    "outputExpanded": false
-   },
-   "source": [
-    "import ipykernel"
-   ],
-   "outputs": []
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 3,
-   "metadata": {
-    "collapsed": false,
-    "outputExpanded": false
-   },
-   "source": [
-    "IPython.display.display({\n",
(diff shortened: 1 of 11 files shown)


mpacer:
Shouldn't we be able to use the React.children.map trick to bubble up onClick events on arbitrary element types? Something like that which @rgbkrk and I stumbled upon at the react workshop…
I think it was a code pattern along the lines of:

```javascript
React.children.map(this.props.children, (child) => 
  {React.cloneElement(child,{
    onClick: () => {…}
    })
  })
```

That would mean not restricting to buttons.

rgbkrk:
Did you mean to respond in https://github.com/nteract/vdom/pull/44 @mpacer?

rgbkrk:
Closing for now, let's revisit this next week during the nteract winter sprint if we make time.

PR #1909, opened 2017-09-26
[wip] enforce dependencies

rgbkrk:
Testing out RxJS 5.5 beta to see if the lettable operators work out well for us. While I'm at it, trying to fix up install from scratch.


rgbkrk:
I'm going to come back to this one since I went ahead with React upgrades.

PR #1914, opened 2017-09-27
Setup auto update

lgeiger:
On Windows and Mac new releases will now be automatically downloaded and installed on quit.
[RELEASING.md](https://github.com/lgeiger/nteract/blob/082fec44919bc8d3ed6a4261be978b1d1cd3ea53/packages/desktop/RELEASING.md) documents the new automated release process to make this work.

On macOS the auto-update will look like this: https://youtu.be/dsmb-Piocs0

On Windows it works too (the virtual machine is a bit slow though): https://youtu.be/xUPapWgq4kU
Note: The Windows release was actually built on macOS.

To try it out yourself [download and install `0.2.4`](https://github.com/lgeiger/nteract/releases/tag/v0.2.4) and wait for `0.2.5` to be installed automatically on launch.

### DIFF
--- a/packages/desktop/RELEASING.md
+++ b/packages/desktop/RELEASING.md
@@ -1,29 +1,40 @@
-Let's package a release!
+# :package: Let's package a release!
 
-# All platform precursor
+## All platform precursor
 
-Cut a release using the `npm` workflow:
-
+To be able to publish a release you'll need to generate a GitHub access token by going to <https://github.com/settings/tokens/new>.  The access token should have the `repo` scope/permission.  Once you have the token, assign it to an environment variable (on macOS/linux):
+```bash
+export GH_TOKEN="<YOUR_TOKEN_HERE>"
 ```
+
+## macOS
+
+In order to build a signed copy with working auto-update, you will need to join the Apple developer program and get a certificate. The [Electron docs have a document on submitting your app to the app store](https://github.com/electron/electron/blob/master/docs/tutorial/mac-app-store-submission-guide.md), you only have to get through to the certificate step.
+
+## Release Process
+
+1. Make sure the release is working by running `npm run dist` and testing the built app inside the `./packages/desktop/dist/` folder. You can build for all platforms using `npm run dist:all`.
+
+2. If everything works as expected, bump the version number using `npm`:
+```bash
 npm version {major, minor, patch}
 npm publish
 git push
 git push --tags
 ```
+or manually by editing the version in `./packages/desktop/package.json` and pushing the changes to GitHub.
 
-Now, from GitHub go to [nteract's releases](https://github.com/nteract/nteract/releases) and make a release from the version you just published above. The name should follow our [naming guidelines](https://github.com/nteract/naming), namely that we use the last name of the next scientist in the list with an adjective in front.
+3. Run `npm run publish` on macOS, Windows and Linux or run `npm run publish:all` to build everything on a single machine. This will draft a new release on GitHub and will upload all necessary assets.
 
+4. From GitHub go to [nteract's releases](https://github.com/nteract/nteract/releases), verify everything works and edit the release notes. The name should follow our [naming guidelines](https://github.com/nteract/naming), namely that we use the last name of the next scientist in the list with an adjective in front.
 Example:
-
-```
+```bash
 Last release: Avowed Avogadro
 Next Scientist: Babbage
 Next release: Babbling Babbage
 ```
-
 My favorite way to pick the alliterative adjectives is using the local dictionary and our friend `grep`:
-
-```
+```bash
 $ cat /usr/share/dict/words | grep "^babb"
 babbitt
 babbitter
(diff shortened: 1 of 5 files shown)


packages/desktop/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@nteract/desktop",
+  "name": "nteract",
```
lgeiger, line 2: The auto-update didn't like `@nteract/desktop`. I'm sure it could be configured to work but renaming is the easiest solution.

packages/desktop/package.json
```diff
@@ -42,14 +42,19 @@
...
+      "npm run clean && cross-env NODE_ENV=production webpack && electron-builder --dir",
+    "dist":
+      "npm run clean && cross-env NODE_ENV=production webpack && electron-builder",
+    "publish":
+      "npm run clean && cross-env NODE_ENV=production webpack && electron-builder -p always",
+    "dist:all": "npm run dist -- -mlw",
+    "publish:all": "npm run publish -- -mlw"
   },
   "build": {
-    "electronVersion": "1.7.6",
```
lgeiger, line 50: `electron-builder` will use the version defined in the devDependencies.

packages/desktop/RELEASING.md
```diff
@@ -39,13 +50,6 @@ babblishly
 babbly
 babby
 ```
+*To ensure that auto-update works, it is best to not modify the name of the release directly, instead just add the chosen one to the release notes.*
```
lgeiger, line 53: I don't know if this is really necessary but I guess there could be problems otherwise.

packages/desktop/package.json
rgbkrk: Seems alright to me

packages/desktop/package.json
rgbkrk: Nice.

packages/desktop/RELEASING.md
rgbkrk: I may not totally understand this part.

rgbkrk:
This is so awesome! I've been looking forward to this since day one!

packages/desktop/RELEASING.md
```diff
@@ -1,29 +1,40 @@
...
+# :package: Let's package a release!
 
-# All platform precursor
-
-Cut a release using the `npm` workflow:
+## All platform precursor
 
+To be able to publish a release you'll need to generate a GitHub access token by going to <https://github.com/settings/tokens/new>.  The access token should have the `repo` scope/permission.  Once you have the token, assign it to an environment variable (on macOS/linux):
+```bash
+export GH_TOKEN="<YOUR_TOKEN_HERE>"
```
rgbkrk, line 7: I'll have to remember do that. 😉

packages/desktop/RELEASING.md
```diff
@@ -1,29 +1,40 @@
...
+In order to build a signed copy with working auto-update, you will need to join the Apple developer program and get a certificate. The [Electron docs have a document on submitting your app to the app store](https://github.com/electron/electron/blob/master/docs/tutorial/mac-app-store-submission-guide.md), you only have to get through to the certificate step.
+
+## Release Process
+
+1. Make sure the release is working by running `npm run dist` and testing the built app inside the `./packages/desktop/dist/` folder. You can build for all platforms using `npm run dist:all`.
+
+2. If everything works as expected, bump the version number using `npm`:
+```bash
 npm version {major, minor, patch}
 npm publish
```
rgbkrk, line 21: Is this at the base level or into what's now the `nteract` package that's deep inside our tree?

packages/desktop/RELEASING.md
rgbkrk: Doesn't it look it up by tag? I feel like it's ok for the name to get updated.

rgbkrk (approved):

rgbkrk:
Let's do this!

packages/desktop/RELEASING.md
lgeiger: Good point. It's the package inside the tree.
We could also remove this section and just manually bump the version number which is arguably easier. We don't need the desktop app published on npm anyway.

packages/desktop/RELEASING.md
lgeiger: I'll check if it works with arbitrary names too.

packages/desktop/RELEASING.md
lgeiger: I hope electron-builder has a nice error message. If not I can always wrap the publish script so it reminds you ;)

packages/desktop/RELEASING.md
rgbkrk: True true. We do need the tag though right?

rgbkrk:
The only thing I want to answer now is if we should ship a release or not. I was hoping to finish off the package upgrades first, make sure things were in a good spot, and potentially fix a few things. Could be fine to ship now though. Anything glaring?

packages/desktop/RELEASING.md
lgeiger: No, electron builder will create the tag for us if it's not there.

lgeiger:
I'd be OK to shipping a new release very soon. Now we can start shipping more often since users don't need to upgrade manually.

PR #1016, opened 2016-10-15
perf(load): bundle and tree shake assets with webpack

rgbkrk:
Closes #1014 by slowly introducing webpack.

To try it out, fetch this branch and then: `npm install`. It does the build automatically.

PR #1925, opened 2017-10-02
Use lettable operators

lgeiger:
- Upgrade to `rxjs@5.5.0-beta.2`
- Use lettable operators everywhere in order to remove the need for modifying the Observable prototype
- Enable all skipped rxjs tests
- Upgrade jest


flow-typed/npm/rxjs_v5.0.x.js
```diff
@@ -12,73 +12,342 @@ interface rxjs$IObserver<-T> {
...
+    op2: rxjs$OperatorFunction<A, B>,
+    op3: rxjs$OperatorFunction<B, C>,
+    op4: rxjs$OperatorFunction<C, D>,
+    op5: rxjs$OperatorFunction<D, E>,
+    op6: rxjs$OperatorFunction<E, F>,
+    op7: rxjs$OperatorFunction<F, G>,
+    op8: rxjs$OperatorFunction<G, H>,
+    op9: rxjs$OperatorFunction<H, I>
+  ): rxjs$Observable<I>;
+  pipe<R>(...operations: rxjs$OperatorFunction<T, R>[]): rxjs$Observable<R>;
```
lgeiger, line 117: Blindly copied this from the RxJS typescript defs

rgbkrk:
Oh my gosh, thank you for taking this on.

packages/messaging/src/index.js
```diff
@@ -61,24 +61,21 @@ export function childOf(parentMessage: Object) {
...
-export function ofMessageType(messageTypes: Array<string>) {
-  return Observable.create(subscriber => {
-    // since we're in an arrow function `this` is from the outer scope.
-    // save our inner subscription
-    const subscription = this.subscribe(
+export const ofMessageType = (messageTypes: Array<string>) => (
+  source: rxjs$Observable<*>
+) =>
+  new Observable(subscriber =>
+    source.subscribe(
```
rgbkrk, line 78: This is looking so good, thank you.

packages/desktop/src/notebook/epics/comm.js
```diff
@@ -126,15 +123,17 @@ export function commMessageAction(message: any) {
  * @return {ActionsObservable}          all actions resulting from comm messages on this kernel
  */
 export function commActionObservable(newKernelAction: any) {
-  const commOpenAction$ = newKernelAction.channels.iopub
-    .ofMessageType(["comm_open"])
-    .map(commOpenAction);
+  const commOpenAction$ = newKernelAction.channels.iopub.pipe(
+    ofMessageType(["comm_open"]),
```
rgbkrk, line 127: These are so good now.

packages/desktop/src/notebook/epics/execute.js
```diff
@@ -221,43 +222,42 @@ export function executeCellStream(
...
-    .pluck("content", "payload")
-    .filter(Boolean)
-    .mergeMap(payloads => Observable.from(payloads));
+  const payloadStream = shell.pipe(
+    childOf(executeRequest),
+    ofMessageType(["execute_reply"]),
+    pluck("content", "payload"),
+    filter(Boolean),
+    mergeMap(payloads => Observable.from(payloads))
+  );
```
rgbkrk, line 239: Wow do I love the lettable operators, these come out so much cleaner and look way easier to refactor for the future.

rgbkrk (approved):

rgbkrk:
All notebooks I tested, including some rather obscure messaging payloads, are working quite well.

rgbkrk:
I think this is good to go and I'm more than happy to rely on the RxJS 5.5 beta

lgeiger:
@rgbkrk thanks for testing

jayphelps:
Just to confirm, this imports _all_ operators by using `from "rxjs/operators";` whereas previously operators were individually imported. These new lettable operators are tree-shakeable, but I did a bit of digging at the various webpack configs and it doesn't appear it's configured to remove the dead code? ([webpack itself only makes the exports unused](https://webpack.js.org/guides/tree-shaking/#minify-the-output), something else needs to then remove the deadcode, like uglify). Then again #1016 was done specifically to do shaking, so perhaps I'm mistaken? 😆  

Someone might want to confirm my theory. I think the bundle size likely increased, but it might be negligible since nteract already imports a good percent of the operators.

rgbkrk:
We could probably run this through uglify. Luckily the size of the js bundle isn't a big issue considering what users currently deal with -- this still loads faster than the classic jupyter notebook.

lgeiger:
I tried to enable minification in the past, but it failed due to memory overflow on my system (at least when minifying the renderer bundle).

PR #1926, opened 2017-10-02
docs(commutable): Add a link to the commuter docs in the README

Madhu94:
I didn't know these docs existed! I was going through the code to figure out how commutable worked, I just want to make sure no one else misses this..

### DIFF
--- a/packages/commutable/README.md
+++ b/packages/commutable/README.md
@@ -22,3 +22,8 @@ Credits to [Tom MacWright](http://www.macwright.org/2015/05/18/practical-undo.ht
 ```
 npm install --save @nteract/commutable
 ```
+
+## Docs
+
+Check out our [docs](https://nteract.github.io/docs/commutable/) for getting started.
+


rgbkrk:
🤔  Those are probably from the "old" https://github.com/nteract/commutable, we should probably figure out a way to document all our packages in one handy spot.

Madhu94:
Yeah, I should remove the typescript reference and put flow there, but otherwise the tutorial still holds good.

Do you have any docs about how nteract transforms work ? I'm guessing they are the nteract equivalent of the notebook's custom mime-types, but don't know how they work..(Not that figuring out stuff from code isn't fun :) )

rgbkrk:
nteract transforms are similar to the notebook's custom mimetype. They take mimetyped data and turn it into a react element. The nteract transforms library has some helpers for picking which of the transforms to use given a bundle of output data from a notebook file.

Madhu94:
> The nteract transforms library has some helpers for picking which of the transforms to use given a bundle of output data from a notebook file.

Thanks, will try it soon..! I'm trying out some nteract packages these days and would love more unified docs.

edit:- Instead of saying I could use some docs, I should add more docs myself. :)

rgbkrk:
Yeah I've been really hoping we could feature them on the main site in a nice way.

PR #1928, opened 2017-10-03
pin rx-jupyter to the prerelease version

rgbkrk:
I made a prerelease of `rx-jupyter` that relies on the rxjs beta release for us to use here.


rgbkrk:
Ok, great this is working locally. I'll rebase #1899 and get that up to par to bring in right after this.

PR #1931, opened 2017-10-04
only run updateAndNotify on windows + macOS

rgbkrk:
Fixing up

image

https://twitter.com/chendaniely/status/915402770655608834

### DIFF
--- a/packages/desktop/src/main/auto-updater.js
+++ b/packages/desktop/src/main/auto-updater.js
@@ -1,8 +1,11 @@
 import { autoUpdater } from "electron-updater";
 
 export function initAutoUpdater() {
-  const log = require("electron-log");
-  log.transports.file.level = "info";
-  autoUpdater.logger = log;
-  autoUpdater.checkForUpdatesAndNotify();
+  // Autoupdate should only get called on macOS and Windows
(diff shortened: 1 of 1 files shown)


chendaniely (approved):
I've confirmed that this PR fixes the issue and runs on my arch machine (:

rgbkrk:
Alright, I'll get this shipped then. It will be the first version that the macOS and Windows versions get an auto-update for amusingly. 😉

chendaniely:
does it support marquees?

rgbkrk:
Full blown enterprise level `<marquee>` support coming your way soon!

rgbkrk:
(build running...)

lgeiger:
I don't know why this is necessary. `updateAndNotify` should just be a [noop on linux](https://github.com/electron-userland/electron-builder/blob/c14a19300e426e6d99d20fdabb27ccbac529bb19/packages/electron-updater/src/AppUpdater.ts#L175).

I'm wondering if it's because it's build on mac instead of on Linux.

rgbkrk:
> I'm wondering if it's because it's build on mac instead of on Linux.

Yeah that could be. I would hope webpack doesn't take a selective path while building the JS with an electron target.

PR #1934, opened 2017-10-04
Webpack codemirror and fonts

lgeiger:
This strips 3MB from our bundle compared to 0.3.1.


packages/editor/src/wrapper.js
```diff
@@ -12,6 +12,10 @@ import { Map as ImmutableMap } from "immutable";
 
 import { transforms } from "@nteract/transforms";
 
+import "codemirror/lib/codemirror.css";
+import "codemirror/addon/hint/show-hint.css";
+import "codemirror/addon/dialog/dialog.css";
```
lgeiger: Could this break notebook-preview?

packages/desktop/webpack.config.js
```diff
@@ -56,7 +60,6 @@ const rendererConfig = Object.assign({}, options, {
       "redux-observable",
       "immutable",
       "rxjs",
-      "codemirror",
```
lgeiger: Is the code splitting into a vendor bundle even useful in the Electron context?

packages/editor/src/wrapper.js
lgeiger: We could also import it in `main.css` if the css import in the `@nteract/editor` package breaks other stuff.

rgbkrk: This very well could... It won't affect `commuter` since we load this css explicitly later.

packages/desktop/webpack.config.js
rgbkrk: Yeah it makes it useful when working on the desktop app, because webpack only has to rebuild the source we're changing here (instead of Rx + React + everything else).

rgbkrk: It's also useful on the web app too, and is primarily here for developer experience.

packages/desktop/package.json
```diff
@@ -34,11 +34,11 @@
...
     "test:coverage": "npm run coverage",
     "test:unit":
-      "mocha -r test/setup.js --compilers js:babel-core/register \"test/renderer/**/*.js\"",
+      "mocha -r test/setup.js --compilers js:babel-core/register --require ignore-styles \"test/renderer/**/*.js\"",
     "test:debug":
-      "mocha --debug-brk --inspect -r test/setup.js --compilers js:babel-core/register \"test/renderer/**/*.js\"",
+      "mocha --debug-brk --inspect -r test/setup.js --compilers js:babel-core/register --require ignore-styles \"test/renderer/**/*.js\"",
     "test:unit:individual":
-      "mocha -r test/setup.js --compilers js:babel-core/register ",
+      "mocha -r test/setup.js --compilers js:babel-core/register --require ignore-styles ",
```
rgbkrk: 👍

packages/desktop/src/notebook/main.css
```diff
@@ -0,0 +1,730 @@
+/*
+  App
+ */
+@import "../../node_modules/normalize.css/normalize.css";
+@import "../../node_modules/nteract-assets/fonts/source-sans-pro/source-sans-pro.css";
+@import "../../node_modules/nteract-assets/fonts/source-code-pro/source-code-pro.css";
+@import "../../node_modules/nteract-assets/fonts/octicons/octicons.css";
```
rgbkrk: Nice, glad CSS imports can work here and that webpack will handle them.

packages/editor/src/wrapper.js
rgbkrk: Oh wait, now I see. You're _introducing_ the css here. This does enforce that users of this component are using the webpack css loader.

packages/desktop/webpack.config.js
```diff
@@ -13,7 +14,25 @@ const options = {
   module: {
     rules: [
       { test: /\.js$/, exclude: /node_modules/, loader: "babel-loader" },
-      { test: /\.json$/, loader: "json-loader" }
+      { test: /\.json$/, loader: "json-loader" },
+      {
+        test: /\.css$/,
+        use: ExtractTextPlugin.extract({
+          fallback: "style-loader",
+          use: "css-loader"
```
rgbkrk, line 22: Oh cool, this takes all the css imports and creates `styles.css`.

packages/editor/src/wrapper.js
lgeiger: Should I just import it in the `main.css` style to be save?

packages/desktop/webpack.config.js
lgeiger: Yep, otherwise it would inline everything

packages/editor/src/wrapper.js
rgbkrk: Yeah that sounds good to me.

packages/desktop/webpack.config.js
rgbkrk: Super slick, gotta love webpack. Thanks @TheLarkInn and pals.

lgeiger:
Rebased it, so we don't modify `@nteract/editor`

packages/desktop/webpack.config.js
lgeiger: > Yeah it makes it useful when working on the desktop app, because webpack only has to rebuild the source we're changing here (instead of Rx + React + everything else).

Right now `rimraf lib` runs before `build`, so webpack doesn't cache it.

PR #1935, opened 2017-10-04
Remove unused dependencies

lgeiger:
This removes some leftovers from the lerna restructuring.


packages/transform-dataresource/package.json
```diff
@@ -23,7 +23,6 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "d3-time-format": "^2.0.5",
-    "jsontableschema": "^0.2.2",
```
rgbkrk, line 26: Is this because I ended up copying in what we needed for tableschema into here?

packages/desktop/package.json
```diff
@@ -129,23 +128,17 @@
...
     "jsdom": "^11.0.0",
-    "jsontableschema": "^0.2.2",
     "jupyter-paths": "^2.0.0",
     "kernelspecs": "^2.0.0",
-    "leaflet": "^1.0.3",
     "lodash": "^4.17.4",
     "prop-types": "^15.5.10",
     "react-dnd": "^2.4.0",
     "react-dnd-html5-backend": "^2.4.1",
-    "react-json-tree": "^0.10.9",
```
rgbkrk, line 143: Rockin'

packages/desktop/package.json
```diff
@@ -129,23 +128,17 @@
     "enchannel": "^4.0.0",
     "enchannel-zmq-backend": "^5.0.0",
     "escape-carriage": "^1.2.0",
-    "flow-bin": "^0.55.0",
```
rgbkrk, line 132: We do need `flow-bin` here for when we install new dependencies into this package. It has to be installed at this level in order to run the `flow-typed` install (though the flow definitions do end up in the right spot).

packages/transform-dataresource/package.json
lgeiger: I can't find any mention of `jsontableschema` anywhere in the code so I think we're good to remove it.

rgbkrk: Ok, cool!

PR #1937, opened 2017-10-04
Tree shake and uglify production build

### LINKED ISSUE #1849: Optimizing the desktop app size

lgeiger:
Overall this strips 13.3MB from our `nteract.app` compared to `0.3.1`. Closes #1849

file | size master | size PR
--- | --- | ---
`vendor.js` | 3.1 MB | 0.8 MB
`webpacked-main.js` | 2.9 MB | 1.1 MB
`wegpacked-notebook.js` | 11.8 MB | 5.5 MB

For further runtime improvements we can finetune our babel setup and take a look at:
- https://babeljs.io/docs/plugins/transform-react-inline-elements/
- https://babeljs.io/docs/plugins/transform-react-constant-elements/
- https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,9 @@
     "sinon": "^3.2.0",
     "sinon-chai": "^2.10.0",
     "style-loader": "^0.19.0",
+    "uglifyjs-webpack-plugin": "^1.0.0-beta.2",
     "url-loader": "^0.6.1",
-    "webpack": "^3.5.5"
+    "webpack": "^3.5.5",
+    "webpack-merge": "^4.1.0"
   }
 }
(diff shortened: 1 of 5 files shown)


rgbkrk:
Rockin'

rgbkrk:
I wonder if we could/should move plotly to vendor. Or wait, we run it as a top level global don't we? I feel like that's what we do for commuter, I'm not sure if that carried over here...

PR #1938, opened 2017-10-04
Smoother startup

### LINKED ISSUE #1933: Remove splash screen
It seams like Electrons startup is now faster than it used to be. Now the DOM takes longer to load than the splash screen itself. What do you think, should we remove it entirely and focus our efforts on providing a smoother initial startup?

This is 0.3.1 (the splash screen doesn't even appear):
kapture 2017-10-04 at 12 43 56

lgeiger:
- Only show the window once it's ready
- Close the splash screen once 

Closes #1933

### DIFF
--- a/packages/desktop/src/notebook/index.js
+++ b/packages/desktop/src/notebook/index.js
@@ -1,6 +1,7 @@
 // @flow
 import React from "react";
 import ReactDOM from "react-dom";
+import { ipcRenderer as ipc } from "electron";
 
 import { Provider } from "react-redux";
 
@@ -48,6 +49,7 @@ class App extends React.PureComponent<Object, Object> {
 
   componentDidMount(): void {
     store.dispatch(setNotificationSystem(this.notificationSystem));
+    ipc.send("react-ready");
   }
 
   render(): ?React$Element<any> {
--- a/packages/desktop/static/index.html
+++ b/packages/desktop/static/index.html
@@ -6,47 +6,8 @@
     <script type="text/javascript" src="../node_modules/mathjax-electron/resources/MathJax/MathJax.js?config=electron"></script>
   </head>
   <body>
-    <div id="app">
-      Loading...
-    </div>
-    <script>
-      window.onload = function deferredLoad() {
-        if (process.env.NODE_ENV === 'development') {
-          // TODO:
-          // Provide some helpers to instantiate following the process in
-          // https://github.com/electron/electron/blob/master/docs/tutorial/devtools-extension.md
-        } else {
-            // Force production by default for the sake of packaging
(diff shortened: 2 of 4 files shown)


packages/desktop/src/notebook/index.js
```diff
@@ -48,6 +49,7 @@ class App extends React.PureComponent<Object, Object> {
 
   componentDidMount(): void {
     store.dispatch(setNotificationSystem(this.notificationSystem));
+    ipc.send("react-ready");
```
rgbkrk, line 52: What a superb idea.

packages/desktop/static/index.html
```diff
@@ -6,47 +6,8 @@
...
-          }
-          el.appendChild(headerEl);
-          el.appendChild(msgContainer);
-          console.dir(err);
-        }
-      };
-    </script>
+    <div id="app"></div>
+    <script src="../lib/vendor.js"></script>
+    <script src="../lib/webpacked-notebook.js"></script>
```
rgbkrk, line 11: I've got a funny feeling about this not being a require statement, though I can't recall why...

rgbkrk: Maybe it was pre-webpack days that this not being `require`'d was odd.

rgbkrk:
With all these sweet changes you're putting in so quickly along with autoupdate, I'm definitely ready to start shipping regularly.

rgbkrk:
smooth

lgeiger:
Looking great. The release build will be even faster ;)

I wonder if we should move back to the static splash screen since the video still takes quite a bit to initialize.

lgeiger:
Also it would be great to initialize the browser window with the correct background color to get rid of the flickering with dark themes.

Do you know if we really need to wait for the `did-finish-load` and `will-navigate` events before we can load the notebook?

rgbkrk:
> I wonder if we should move back to the static splash screen since the video still takes quite a bit to initialize.

I'm ok with the video for the moment. I think it's the electron app readiness that takes the longest here, not the video.

rgbkrk:
> Also it would be great to initialize the browser window with the correct background color to get rid of the flickering with dark themes.

That would be nice. We should probably do some "templating" of our `index.html` for this, unless you're saying a setting on the `BrowserWindow`.

rgbkrk:
> Do you know if we really need to wait for the `did-finish-load` and `will-navigate` events before we can load the notebook?

Not sure.

lgeiger:
> We should probably do some "templating" of our index.html for this, unless you're saying a setting on the BrowserWindow.

Yes it can be configured in the [BrowserWindow](https://electron.atom.io/docs/api/browser-window/#new-browserwindowoptions).

PR #1941, opened 2017-10-04
RxJS: Use es2015 build in webpack

lgeiger:
This should improve tree-shaking of RxJS.


packages/desktop/src/notebook/epics/execute.js
```diff
@@ -235,7 +235,7 @@ export function executeCellStream(
     ofMessageType(["execute_reply"]),
     pluck("content", "payload"),
     filter(Boolean),
-    mergeMap(payloads => Observable.from(payloads))
+    mergeMap(payloads => from(payloads))
```
rgbkrk, line 238: Good catch

rgbkrk (approved):

PR #1943, opened 2017-10-05
[NO MERGE] let's do it for Science!

rgbkrk:
Ref #1942 

Apparently this `science` field was ending up in the `.desktop` file as the improper lowercase version. ¯\\\_(ツ)\_/¯

I don't think it totally fixes the aforementioned issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   - GROUP=lint
   - GROUP=flow
   - GROUP=conformance
+  - GROUP=dist
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


rgbkrk:
🤔 I wonder if I should have travis run `npm run dist:all` as a sanity check...

lgeiger:
For Travis we'll need a few more dependencies: https://www.electron.build/multi-platform-build

rgbkrk:
I don't _really_ want to enable shipping from Travis, this was more of a sanity check.

rgbkrk:
Closing since I think you fixed this right up @lgeiger

PR #1946, opened 2017-10-05
Pin electron-builder to 19.33.0

lgeiger:
Adresses #1942 

Looks like there was a bug introduced int 19.34.0 and still present in 19.34.1. We should open a issue on the electron-builder repo.


rgbkrk:
Looks like I'll need to re-kick the appveyor build when I get a chance.

PR #1950, opened 2017-10-05
[VDOM] Further validation

rgbkrk:
I've managed to break the app doing some misguided things from Python land, so I'm putting more validation into `objectToReact`.


rgbkrk:
Looks like my trying to turn off codecov being a requirement didn't work too well.

PR #1952, opened 2017-10-05
[WIP] Add tagging to cells

### LINKED ISSUE #1951: Notebook cells should allow "tagging"
Each notebook cell should have an editable `tags` field which allows the user to input tags to be saved in the cell's metadata block.

We need to be able to set the `parameters` tag on a cell in order to use `papermill` with `nteract`

cc @ewmassey

klebba:
Fixes #1951

I took at stab at integrating the [react-tagsinput](https://github.com/olahol/react-tagsinput) component in the cell header so that users can add tags.

screen shot 2017-10-05 at 1 34 01 pm

There is still open work remaining to ensure the feature looks acceptable when each of the default stylesheets is loaded. A more holistic solution could be to redesign the "cell configuration" settings (such as cell type switching, input & output visibility, and now tagging) but this will be done in a separate effort.

### DIFF
--- a/packages/desktop/src/notebook/actions.js
+++ b/packages/desktop/src/notebook/actions.js
@@ -259,6 +259,14 @@ export function changeCellType(id: CellID, to: CellType) {
   };
 }
 
+export function updateCellTags(id: CellID, tags: any) {
+  return {
+    type: constants.UPDATE_CELL_TAGS,
+    id,
+    tags
+  };
+}
+
 export function setGithubToken(githubToken: string) {
   return {
     type: constants.SET_GITHUB_TOKEN,
--- a/packages/desktop/src/notebook/reducers/document.js
+++ b/packages/desktop/src/notebook/reducers/document.js
@@ -646,6 +646,16 @@ function changeCellType(state: DocumentState, action: ChangeCellTypeAction) {
   );
 }
 
+type UpdateCellTagsAction = {
+  type: "UPDATE_CELL_TAGS",
+  id: CellID,
+  tags: Array<string>
+};
+function updateCellTags(state: DocumentState, action: UpdateCellTagsAction) {
+  const { id, tags } = action;
+  return state.setIn(["notebook", "cellMap", id, "metadata", "tags"], tags);
+}
+
 type ToggleCellExpansionAction = {
   type: "TOGGLE_OUTPUT_EXPANSION",
   id: CellID
@@ -771,6 +781,8 @@ function handleDocument(
       return pasteCell(state);
     case constants.CHANGE_CELL_TYPE:
       return changeCellType(state, action);
+    case constants.UPDATE_CELL_TAGS:
+      return updateCellTags(state, action);
     case constants.TOGGLE_OUTPUT_EXPANSION:
       return toggleOutputExpansion(state, action);
     default:
--- a/packages/desktop/src/notebook/views/toolbar.js
+++ b/packages/desktop/src/notebook/views/toolbar.js
@@ -12,6 +12,8 @@ import Dropdown, {
   DropdownContent
 } from "react-simple-dropdown";
 
+import TagsInput from "react-tagsinput";
+
 declare type ToolbarProps = {|
   cell: any,
   id: string,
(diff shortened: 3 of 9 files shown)


packages/desktop/static/styles/theme-nteract.css
```diff
@@ -399,12 +399,12 @@ tr
...
       height: 11px;
-} 
+}
 div::-webkit-scrollbar-track, ul::-webkit-scrollbar-track {
       -webkit-box-shadow: inset 0 0 10px rgba(0,0,0,0.3);
-} 
+}
 div::-webkit-scrollbar-thumb, ul::-webkit-scrollbar-thumb {
-      background-color: rgba(237, 243, 247, .6) 
+      background-color: rgba(237, 243, 247, .6)
```
klebba: Fix missing semicolon

packages/desktop/src/notebook/views/toolbar.js
```diff
@@ -149,6 +160,12 @@ export default class Toolbar extends PureComponent<ToolbarProps> {
             </DropdownContent>
           </Dropdown>
         </div>
+        <TagsInput
+          value={this.props.cell.getIn(["metadata", "tags"], [])}
+          onlyUnique
+          inputProps={this.tagsInputProps.inputProps}
```
klebba: Maybe use inline object notation here

rgbkrk:
Thanks for tackling this! I know some users have wondered about this for both jupyter + nteract ("how do I get to tags?") -- https://github.com/nteract/papermill/issues/20

I'm really looking forward to having a cell configuration "area", which I've been mentally seeing as if you could flip the "card" that is the cell over. Great to explore.

I'll go ahead and comment inline.

packages/desktop/src/notebook/views/toolbar.js
rgbkrk: Yeah, this would also fix up the current `flow` error.

```js
  inputProps={{ className: "react-tagsinput-input", placeholder: "Add tag" }}
```

packages/desktop/src/notebook/actions.js
```diff
@@ -259,6 +259,14 @@ export function changeCellType(id: CellID, to: CellType) {
   };
 }
 
+export function updateCellTags(id: CellID, to: any) {
+  return {
+    type: constants.UPDATE_CELL_TAGS,
+    id,
+    to
```
rgbkrk: I would probably make this be `tags` instead of `to`, to be explicit.

packages/desktop/src/notebook/reducers/document.js
```diff
@@ -646,6 +646,16 @@ function changeCellType(state: DocumentState, action: ChangeCellTypeAction) {
...
 
+type UpdateCellTagsAction = {
+  type: "UPDATE_CELL_TAGS",
+  id: CellID,
+  to: Array<string>
+};
+function updateCellTags(state: DocumentState, action: UpdateCellTagsAction) {
+  const { id, to } = action;
+  return state.setIn(["notebook", "cellMap", id, "metadata", "tags"], to);
+}
```
rgbkrk: Since pretty much all the `metadata` setting reducers are exactly the same, it's really tempting to make a `SET_IN_CELL_METADATA` action/reducer pair, where the utility functions like `updateCellTags` only returns the metadata data change update:

```js
export function updateCellTags(id: CellID, tags: any) {
  return {
    type: constants.SET_IN_CELL_METADATA,
    id,
    path: ["metadata", "tags"],
    value: tags
  }
}
```

rgbkrk: You don't have to worry about it in this PR, I'm thinking we can trim a lot of boilerplate in this repository now.

rgbkrk: Amusingly, I think we could probably make a reducer that's literally `setInState`. It would miss a lot of error checking though, so maybe the highest granularity we could do is:

```js
{
  type: 'SET_IN_CELL',
  id: string,
  path: Array<string>
}
```

That loses type information, though I suppose at the point of that reducer, perhaps we have enough context to know if what we're doing is "safe".

klebba: Agreed -- as discussed offline we will defer refactoring to future work

packages/desktop/src/notebook/actions.js
klebba: Done

rgbkrk:
I'm going to go ahead and close this PR since I've talked to Casey in person about this one fairly recently. We'd like to wait on tackling this until we refactor the cell context menu's design.

PR #1958, opened 2017-10-06
switch npm t to the new unit test setup

rgbkrk:
Since the testing setup has changed, `npm test` needed to be updated. If there's more that should be in here, I'd be happy for us to have it. 😄


rgbkrk:
Bringing this in for now, I really need to figure out what's going on with the test setup. It seems like bb5d387a9372f4b28a5f52fc2e4bb70ccbffa286 should have fixed this.

PR #1959, opened 2017-10-06
remove lodash plugin from babel config

rgbkrk:
seeing if this fixes tests

### DIFF
--- a/packages/.babelrc
+++ b/packages/.babelrc
@@ -11,7 +11,6 @@
   },
   "presets": ["es2015", "react"],
   "plugins": [
-    "lodash",
     "transform-class-properties",
     "transform-flow-strip-types",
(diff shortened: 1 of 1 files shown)


rgbkrk:
I'm going to err on the side of going with this so tests pass for contributors. I'm surprised the env config isn't being respected here.

rgbkrk:
Sorry I made the bundle bigger @lgeiger, I wanted to get things passing for now -- we can fix before the next release. I'd be ok with switching to writing explicit lodash imports if you'd prefer here. I wonder if it's a dependency that's actually "introducing" the issue (enzyme, jest, etc.).

lgeiger:
Thanks for fixing the tests :+1:

PR #1963, opened 2017-10-07
prevent more vdom errors from reaching ReactDOM

rgbkrk:
* `style` is checked to ensure it's a proper object (and not an array), as well as allowing it to be undefined or null
* show better error messages


rgbkrk:
I've found these while doing some crazy generated things, figured I should lock this down better until we have a json schema (https://github.com/nteract/vdom/issues/3).

Issue #1939, opened 2017-10-04
Provide the About Menu Item in the Help Menu on Windows


dhirschfeld:
There should be an about section in the Help menu which lets the user know what version they're running.

I got a notification pop up to say nteract would be updated but I've no idea if it worked?!

I'm just running the Windows app so there is no terminal where I can look for debug output. The notification popped up then... nothing. I restarted nteract, but have no idea what version I have now :thinking:

rgbkrk:
Hmmmmm... I wonder if that's because the menu item on Mac has `nteract -> About` and that `nteract` menu item doesn't exist on Windows...

One place you can check for the time being:

```
 %USERPROFILE%\AppData\Roaming\nteract\log.log
```

dhirschfeld:
Looks like the update did work!
```
[2017-02-08 09:12:10:0244] [info] args { _: [],
  version: false,
  v: false,
  help: false,
  h: false,
  kernel: 'python3',
  k: 'python3',
  '$0': 'nteract.exe',
  verbose: undefined }
[2017-02-08 09:12:10:0352] [info] launching an empty notebook by default
[2017-10-04 11:36:00.696] [info] launching an empty notebook by default
[2017-10-04 11:36:00.707] [info] Checking for update
[2017-10-04 11:36:01.894] [info] Generated new staging user ID: 0f559096-13be-5c46-b969-25890077c07f
[2017-10-04 11:36:05.085] [info] Update for version 0.3.0 is not available (latest version: 0.3.0, downgrade is disallowed.
[2017-10-05 09:00:25.918] [info] launching an empty notebook by default
[2017-10-05 09:00:25.924] [info] Checking for update
[2017-10-05 09:00:30.654] [info] Found version 0.3.1 (url: https://github.com/nteract/nteract/releases/download/v0.3.1/nteract-setup-0.3.1.exe)
[2017-10-05 09:00:30.656] [info] Downloading update from https://github.com/nteract/nteract/releases/download/v0.3.1/nteract-setup-0.3.1.exe
[2017-10-05 09:01:55.993] [info] New version 0.3.1 has been downloaded to C:\Users\dhirschf\AppData\Local\Temp\up-oo5dUY\nteract-setup-0.3.1.exe
[2017-10-05 09:04:01.585] [info] Auto install update on quit
[2017-10-05 09:04:19.117] [info] launching an empty notebook by default
[2017-10-05 09:04:19.124] [info] Checking for update
[...]

Issue #1964, opened 2017-10-09
v0.3.3


rgbkrk:
* [x] Fix the pager / introspection handling https://github.com/nteract/nteract/issues/1804 (fixed in #1965)
* [x] Provide the about menu on Windows #1939

rgbkrk:
All set, enjoy the patches!

dhirschfeld:
Thanks for the fixes. Loving the auto-update too!
```
[2017-10-10 14:27:44.619] [info] Checking for update
[2017-10-10 14:27:48.879] [info] Update for version 0.3.3 is not available (latest version: 0.3.3, downgrade is disallowed.
```

A couple of things I've noticed though...

Whilst there is an "About nteract" menu now, it doesn't bring up any other menu or dialog or anything. There's also nothing shown in the developer tools console when you click on it.

Also, the "Interrupt Running Kernel" menu item doesn't appear to do anything. Again, there's no error in the JS console. Tested with:
```python
x = 1
while True:
    x += 1
```
...I had to restart the kernel to get out of that one.

So, sorry to be the bearer of bad news  :disappointed: 
On the plus side, nothing new is broken and the help does now work great! :smile:

rgbkrk:
Alright then! At least that was the quickest way for us to try out those changes for windows. We'll have to make a different dialog for the versioning.

dhirschfeld:
Always happy to test on Windows... it would probably be more useful if I tested *before* release  though :thinking:

rgbkrk:
Yeahhhhh, I was really hoping that would work. I figured it wouldn't _explicitly_ break anything.

rgbkrk:
Lukas has something in progress for everyone :soon:.

PR #1966, opened 2017-10-10
Windows fixes

### LINKED ISSUE #1611: Implement Kernel Interruption for Windows
It would be very nice to have this feature which is available in the classic notebook.

I've just submitted a calculation with a mistake in it which will take ~15mins to finish :grimacing: so figured I would use my time productively and open a feature request issue!

rgbkrk:
Closes #1611, #1939

Guess I'm going after all of @dhirschfeld's Windows bugs. :smile:

If you're available @jkornblum, it would be good to have you look this over.

### DIFF
--- a/packages/desktop/src/main/menu.js
+++ b/packages/desktop/src/main/menu.js
@@ -472,6 +472,10 @@ const helpDraft = {
   role: "help",
   submenu: [
     {
+      label: "About nteract",
+      role: "about"
+    },
+    {
       label: "Learn More",
       click: () => {
         shell.openExternal("http://github.com/nteract/nteract");
--- a/packages/desktop/src/notebook/menu.js
+++ b/packages/desktop/src/notebook/menu.js
@@ -212,17 +212,7 @@ export function dispatchKillKernel(store) {
 }
 
 export function dispatchInterruptKernel(store) {
-  const state = store.getState();
-  const notificationSystem = state.app.get("notificationSystem");
-  if (process.platform === "win32") {
-    notificationSystem.addNotification({
-      title: "Not supported in Windows",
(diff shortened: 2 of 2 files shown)


rgbkrk:
I'm going to go ahead and get this out the door so we can ship it along with the fix for help (`?`).

packages/desktop/src/main/menu.js
```diff
@@ -472,6 +472,10 @@ const helpDraft = {
   role: "help",
   submenu: [
     {
+      label: "About nteract",
+      role: "about"
```
lgeiger, line 476: According to the electron docs, the `about` menu is only a macOS thing: https://electron.atom.io/docs/api/menu-item/#roles

packages/desktop/src/notebook/menu.js
```diff
@@ -212,17 +212,7 @@ export function dispatchKillKernel(store) {
...
-  if (process.platform === "win32") {
-    notificationSystem.addNotification({
-      title: "Not supported in Windows",
-      message: "Kernel interruption is currently not supported in Windows.",
-      level: "error"
-    });
-  } else {
-    store.dispatch(interruptKernel);
-  }
+  store.dispatch(interruptKernel);
```
lgeiger, line 215: When was this fixed?

I remember https://github.com/nteract/hydrogen/issues/597 and the [issue on `spawn-rx`](https://github.com/paulcbetts/spawn-rx/issues/11
) is still open.

lgeiger (commented):
I'm a bit worried that this won't work 😞

dhirschfeld:
@lgeiger - see also my follow-up testing in #1964

PR #1967, opened 2017-10-10
Revert "Windows fixes"

lgeiger:
Reverts #1966 to address https://github.com/nteract/nteract/issues/1964#issuecomment-335359714

I'm sorry 😜

### DIFF
--- a/packages/desktop/src/main/menu.js
+++ b/packages/desktop/src/main/menu.js
@@ -472,10 +472,6 @@ const helpDraft = {
   role: "help",
   submenu: [
     {
-      label: "About nteract",
-      role: "about"
-    },
-    {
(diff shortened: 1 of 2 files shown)


packages/desktop/src/main/menu.js
```diff
@@ -472,10 +472,6 @@ const helpDraft = {
   role: "help",
   submenu: [
     {
-      label: "About nteract",
-      role: "about"
-    },
-    {
```
rgbkrk, line 478: No, I'm sorry. Also I think it's cool if we stick in a static version thing here.

rgbkrk: 😄

PR #1968, opened 2017-10-10
Help menu: Add link to release notes

### LINKED ISSUE #1939: Provide the About Menu Item in the Help Menu on Windows

lgeiger:
This adds a link to the release notes together with version information to fix #1939 in the easies way possible:
bildschirmfoto 2017-10-10 um 18 45 24

### DIFF
--- a/packages/desktop/src/main/menu.js
+++ b/packages/desktop/src/main/menu.js
@@ -478,6 +478,14 @@ const helpDraft = {
       }
     },
     {
+      label: `Release Notes (${app.getVersion()})`,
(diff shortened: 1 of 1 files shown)


rgbkrk:
Nice, I love it.

PR #1969, opened 2017-10-11
Issue 1789

### LINKED ISSUE #1789: Screen jumps on "shift+enter" for pinned cells
When you have cell pined and run the cell through "shift+enter" the screen will jump to where that cell is located. I'm not sure what "shift+enter" should do with a pinned cell, should it create a new cell / focus next or nothing? It feels like bad UX that the screen jumps even with "shift+enter" when the cell is pinned.

Me running cell through button and then with "shift+enter"
nteract-pinned-cell-jump

captainsafia:
Closes #1789.

While working on this I noticed that we sometimes refer to sitckyCells as a Map when it seems to actually be a Set. That should probably be changed...

### DIFF
--- a/packages/desktop/src/notebook/components/notebook.js
+++ b/packages/desktop/src/notebook/components/notebook.js
@@ -144,7 +144,7 @@ export class Notebook extends React.PureComponent<Props> {
     const id = this.props.cellFocused;
     const cell = cellMap.get(id);
 
-    if (e.shiftKey) {
+    if (e.shiftKey && !this.props.stickyCells.has(id)) {
       this.context.store.dispatch(focusNextCell(this.props.cellFocused, true));
       this.context.store.dispatch(focusNextCellEditor(id));
     }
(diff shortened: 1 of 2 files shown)


packages/desktop/src/notebook/components/notebook.js
```diff
@@ -144,7 +144,7 @@ export class Notebook extends React.PureComponent<Props> {
     const id = this.props.cellFocused;
     const cell = cellMap.get(id);
 
-    if (e.shiftKey) {
+    if (e.shiftKey && !this.props.stickyCells.has(id)) {
```
rgbkrk, line 147: Makes sense to me here.

rgbkrk:
🎉 Thanks!

I was tempted recently to take the pinned cells out entirely because there were a bunch of undiagnosed issues I'd run into if demoing with them. I suppose if we fix them up it would all work out but wow is it tempting to ✂️  snip out some code. 😄

Issue #1970, opened 2017-10-11
Maintenance Wishlist


rgbkrk:
Hey all, there have been some old parts of desktop that I'd love to revisit. I keep thinking about them as I touch various parts of the app. It's about time I outline what some of them are, either to document what I'm going to go after or see if others want to tackle some. Without further ado:

* [x] Collapse all the (notebook cell) reducers into one that are practically the same `setIn` operation on a cell with a different path. All the action creators will end up creating one kind of action for reduction, and it will simplify any additions to metadata changing. (#1971)
* [x] Merge the `metadata` reducer into the document reducer (which will involve changing the usage)
* [x] Delete the constants.js file, switch to literals. Since we're using flow, we can use the literal type names and last I checked was a requirement for proper flow typing.
* [x] Unify all the action flow types. We've duplicated them across `actions.js` and the reducers
* [x] Convert hover logic for the cell creator to CSS (#1972)
* [x] Convert hover logic for the cell / cell-toolbar to CSS (done by @aaronraff in #1975)

rgbkrk:
~I'm going to tackle the first one I listed there, collapsing reducers.~ Done.

aaronraff:
I would like to take up the last bullet! Is it still open?

rgbkrk:
I'm started on it in #1974, if you want to work on it today that's cool with me! My next step was going to be setting up the hover region in CSS that would affect if the toolbar is shown and remove our current JS solution.

aaronraff:
Okay, I changed the JS logic over to CSS for cell.js. Going to commit it later tonight, does that sound like what you are looking for?

rgbkrk:
Sounds good to me! I'm looking forward to it. I'm flying today and need to do some writing anyways so I'm happy for you to take it over. Feel free to open the PR early and seek feedback or ask questions as soon as you're ready.

rgbkrk:
Oh hey we finished all this.

PR #1971, opened 2017-10-11
create a simplified SET_IN_CELL action

rgbkrk:
Starting to tick off the boxes of current sprawl that I think we can control with a `SET_IN_CELL` action. Since I realized how much boilerplate was happening to implement #1952 (and any others that use metadata), I figured I could turn this into a `setIn` setup.

I'd love to see what people think here. Validation will happen at the level of the action creator.

### DIFF
--- a/packages/desktop/src/notebook/actions.js
+++ b/packages/desktop/src/notebook/actions.js
@@ -43,14 +43,6 @@ export function setExecutionState(executionState: string) {
   };
 }
 
-export function updateCellSource(id: string, source: string) {
-  return {
-    type: constants.UPDATE_CELL_SOURCE,
-    id,
-    source
-  };
-}
-
 export function clearOutputs(id: string) {
   return {
     type: constants.CLEAR_OUTPUTS,
@@ -110,6 +102,27 @@ export function mergeCellAfter(id: string) {
   };
 }
 
+/**
+ * setInCell can generically be used to set any attribute on a cell, including
+ * and especially for changing metadata per cell.
+ * @param {CellID} id    cell ID
+ * @param {Array<string>} path  path within a cell to set
+ * @param {any} value what to set it to
+ *
+ * Example:
+ *
+ * > action = setInCell('123', ['metadata', 'cool'], true)
+ * > documentReducer(state, action)
+ * {
+ *   ...
+ *   '123': {
+ *     'metadata': {
+ *       'cool': true
+ *     }
+ *   }
+ * }
+ *
+ */
 export function setInCell(id: CellID, path: Array<string>, value: any) {
   return {
     type: "SET_IN_CELL",
@@ -117,6 +130,10 @@ export function setInCell(id: CellID, path: Array<string>, value: any) {
     path,
     value
   };
+}
+
+export function updateCellSource(id: string, source: string) {
+  return setInCell(id, ["source"], source);
 }
 
 export function updateCellExecutionCount(id: string, count: number) {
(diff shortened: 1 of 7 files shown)


rgbkrk:
Almost did `CHANGE_OUTPUT_VISIBILITY` and `CHANGE_INPUT_VISIBILITY` until I realized they'd have to be a `TOGGLE_IN_CELL` action.

lgeiger (approved):
That looks great!

It's lot simpler now 👍

packages/desktop/src/notebook/reducers/document.js
```diff
@@ -163,6 +163,8 @@ function appendOutput(state: DocumentState, action: AppendOutputAction) {
   const output = action.output;
   const cellID = action.id;
 
+  // If it's display data and it doesn't have a display id, fold it in like non
+  // display data
```
rgbkrk, line 167: As I was scanning through the code I did a double take here, only to realize that if I would have commented about this in the first place it would have been quicker to read through.

packages/desktop/src/notebook/reducers/document.js
```diff
@@ -486,6 +490,7 @@ function changeOutputVisibility(
   );
 }
 
+// TODO: This should be called TOGGLE_INPUT_VISIBILITY
```
rgbkrk, line 493: 🤔  Should we make a `TOGGLE_IN_CELL` action? It would always be a bit flip for boolean fields.

packages/desktop/src/notebook/actions.js
```diff
@@ -110,14 +102,44 @@ export function mergeCellAfter(id: string) {
...
+ *       'cool': true
+ *     }
+ *   }
+ * }
+ *
+ */
+export function setInCell(id: CellID, path: Array<string>, value: any) {
   return {
-    type: constants.UPDATE_CELL_EXECUTION_COUNT,
+    type: "SET_IN_CELL",
```
captainsafia: We can probably set this aside as a Hacktoberfest ticket but it would be good to have this specified in the constants file.

packages/desktop/src/notebook/reducers/document.js
captainsafia: That makes sense. It could be a wrapper around the `SET_IN_CELL` action that sets `value` to `!value`.

captainsafia (approved):
Good stuff! I posted some follow-ups but I think it is good to merge.

packages/desktop/src/notebook/actions.js
rgbkrk: I'll fix this up, thanks for noting it.

I coupled a bit of my concerns from #1970 here, as I've been hoping to get rid of the constants due to a flow limitation. Last I checked, flow interprets the `type` of `type` here as a `string` rather than the explicit literal. When I was doing a bunch of the flow typing for the reducers and action creators I ran into this issue and was hoping to fix it. The other reason I wanted to break away from a constants file was so that it would be easier for us to export the reducers for the web app (some are electron specific, some are not).

In the interest of consistent readable code, I'll get this into the constants.

packages/desktop/src/notebook/reducers/document.js
rgbkrk: Since we need to know the `value` at action creation time, I think it won't be able to use the `SET_IN_CELL` machinery. I can make a follow up PR for that tonight!

rgbkrk:
Addressed the (lack of) use of `constants`, all set now (other than Travis).

PR #1972, opened 2017-10-11
rely on css for tracking hover state

rgbkrk:
This cleans up an excess of mouse events (and handlers for that matter) with some simple CSS.


rgbkrk:
Oh wait, hold on, I need to bring the "above" part back.

rgbkrk:
Wow, this is going to increase coverage by removing untested code? That's fantastic!

lgeiger (approved):
That's a lot better now :+1:

PR #1973, opened 2017-10-11
cut down on boilerplate for Toolbar View/Provider

rgbkrk:
* Pass the exact props into `<ToolbarView />`
* Write component declaration in `render` instead of a `renderToolbar` helper function


lgeiger (approved):
:+1: Nice

PR #1974, opened 2017-10-12
[WIP] remove all the mouse handlers on the Cell

rgbkrk:
This takes out all the hover logic that was on the `Cell` component before. A little bit more to do though as we want the cell toolbar to still show when hovering over it. Similar to #1972, I'll set up a hover region and let CSS do all the work.

* [ ] create css hover region for the cell toolbar
* [ ] only show the toolbar when in the hover region

### DIFF
--- a/packages/desktop/src/notebook/components/cell/cell.js
+++ b/packages/desktop/src/notebook/components/cell/cell.js
@@ -36,11 +36,7 @@ export type CellProps = {
   models: ImmutableMap<string, any>
 };
 
-type State = {
-  hoverCell: boolean
-};
-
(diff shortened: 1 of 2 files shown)


rgbkrk:
@aaronraff -- I just pushed one more commit I had sitting locally since I was partway. Feel free to ignore / update this with how you're expecting it to work. 😄  I'll happily close this PR in favor of yours.

aaronraff:
@rgbkrk Awesome, I looked over your code and used some of it in my PR. Just submitted it a few minutes ago. Please let me know if there is anything I can do to improve it and I will be happy to do so!

rgbkrk:
Closing, fixed in #1975

PR #1975, opened 2017-10-12
Change hovering logic for cell from JS to CSS

aaronraff:

In response to #1970

### DIFF
--- a/packages/desktop/src/notebook/components/cell/cell.js
+++ b/packages/desktop/src/notebook/components/cell/cell.js
@@ -36,16 +36,11 @@ export type CellProps = {
   models: ImmutableMap<string, any>
 };
 
-type State = {
-  hoverCell: boolean
-};
-
 export class Cell extends React.PureComponent<CellProps, State> {
   selectCell: () => void;
   focusAboveCell: () => void;
   focusBelowCell: () => void;
   focusCellEditor: () => void;
-  setCellHoverState: (mouseEvent: MouseEvent) => void;
   cellDiv: ?HTMLElement;
   scrollIntoViewIfNeeded: Function;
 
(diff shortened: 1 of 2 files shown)


packages/desktop/src/notebook/components/cell/cell.js
```diff
@@ -155,9 +125,7 @@ export class Cell extends React.PureComponent<CellProps, State> {
           this.cellDiv = el;
         }}
       >
-        {cellFocused || hoverCell ? (
-          <Toolbar type={type} cell={cell} id={this.props.id} />
-        ) : null}
+        <Toolbar type={type} cell={cell} id={this.props.id} />
```
rgbkrk, line 128: 👍

packages/desktop/src/notebook/main.css
```diff
@@ -268,6 +268,13 @@ body {
   background-color: var(--cell-bg-focus);
 }
 
+/* Show the cell toolbar if hovering on cell,
+   or cell was the last clicked (has .focused class). */
+.cell:hover .cell-toolbar,
+.cell.focused .cell-toolbar {
+  display: block;
```
rgbkrk, line 275: There we go, this is great!

rgbkrk (approved):
Awesome, thank you.

packages/desktop/src/notebook/components/cell/cell.js
```diff
@@ -36,16 +36,11 @@ export type CellProps = {
   models: ImmutableMap<string, any>
 };
 
-type State = {
-  hoverCell: boolean
-};
-
 export class Cell extends React.PureComponent<CellProps, State> {
```
rgbkrk: There's a current flow error here, you'll want to set this to `React.PureComponent<CellProps, *>`

rgbkrk (changes requested):

rgbkrk:
~Tried this out locally, it _also_ fixes an outstanding bug where the dropdown menu would go transparent and the user wouldn't be able to click some items in the dropdown.~ 🙇 

Edit: Actually, the bug is still there, it's just a bit lessened.

packages/desktop/src/notebook/components/cell/cell.js
aaronraff: Just committed this change, thanks for pointing that out! I'm still pretty new to Flow.

rgbkrk (approved):

rgbkrk:
This is fantastic, I can't wait for this to go in the next release!

aaronraff:
I'm happy that I could help out, I'm hoping to contribute more in the future!

rgbkrk:

rgbkrk:
Great, we have much ahead. If you want a next issue to tackle, the dropdown menu could use a fix:

oopsy-under-menu

It's a little better than before, where it was completely unclickable because of the cells underneath hijacking it via that mouse handler. Now at least I think there's just a bit more CSS to get it right.

aaronraff:
I will take a look at that! I'm also thinking that a little CSS will do the trick.

aaronraff:
Took a look at the dropdown menu bug here: #1978

PR #1978, opened 2017-10-13
Fix dropdown menu bug (hovering over toolbar under menu) 

aaronraff:
This makes changes to the `.cell-toolbar-mask` CSS. The problem before was that the cell toolbar mask was interfering with the dropdown menu.


rgbkrk (approved):
Perfect!

lgeiger (approved):
Great :+1:

CI is now happy too.

PR #1979, opened 2017-10-13
Some quick chores

rgbkrk:
Since I was on a plane without internet for a bit, I went digging in our code to see what things I could fix up. They're small enough to me that I thought they'd be ok to just group a few. I can break this out into separate PRs if people like. They're at least separate commits!

* Sticky cell wasn't using the "passed in" `Cell` component
* The plaintext mimetype transform should be a `PureComponent` and not have a shouldComponentUpdate (`PureComponent` will do the handling)

### DIFF
--- a/packages/desktop/src/notebook/components/notebook.js
+++ b/packages/desktop/src/notebook/components/notebook.js
@@ -8,7 +8,6 @@ import { List as ImmutableList, Map as ImmutableMap } from "immutable";
 
 import { displayOrder, transforms } from "@nteract/transforms-full";
 
-import Cell from "./cell/cell";
 import DraggableCell from "../providers/draggable-cell";
 import CellCreator from "../providers/cell-creator";
 import StatusBar from "./status-bar";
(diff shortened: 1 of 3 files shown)


packages/desktop/src/notebook/components/notebook.js
```diff
@@ -219,7 +220,7 @@ export class Notebook extends React.PureComponent<Props> {
     const cell = cellMap.get(id);
     return (
       <div key={`cell-container-${id}`}>
-        <Cell {...this.createCellProps(id, cell, transient)} />
+        <CellComponent {...this.createCellProps(id, cell, transient)} />
```
lgeiger, line 222: The `Cell` import is now unnecessary, isn't it?

lgeiger (approved):
🆒

packages/desktop/src/notebook/components/notebook.js
rgbkrk: Yeah

PR #1980, opened 2017-10-13
Enable auto-updates on Linux AppImage

lgeiger:
This enables auto updates on Linux with AppImage, which is super new for Electron apps.

For multi-platform builds on mac I had to run `brew install glib` to make it work. I'll open a issue on the `electron-builder` repo, since `electron-builder` shouldn't require external dependencies.

Note: I haven't tested the AppImage build on Linux yet.


rgbkrk:
That's fine I'll install glib.

Issue #1927, opened 2017-10-02
Scroll output only when `ctrl` is pressed


betatim:
How about changing the behaviour of the output area so that you have to press `ctrl` or `cmd` to scroll it. Similar to how several map widgets do it. I frequently try to scroll through a notebook and then get "stuck" at a cell with a lot of output and scroll through that instead of the notebook.

GIF of the behaviour on google maps:
![scroll](https://user-images.githubusercontent.com/1448859/31096553-b12ac034-a7bc-11e7-88bb-0e4ff79af964.gif)

rgbkrk:
I'd also love a version where clicking on the output makes it "focus" and allows scrolling within. Otherwise it's a container that doesn't "hijack" scroll.

aaronraff:
Is anyone currently working on this? I'm interested in looking into this tonight or this weekend.

rgbkrk:
This one is definitely open!

rgbkrk:
This is still a good idea, though I couldn't imagine this being touched for a bit. As we're now more able to affect the core experience (more maintainers and contributors, out from under the muck we ended up with in flow), we might be able to revisit this one.

captainsafia:
Commenting on this to remove the stale tag.

I'm gonna go ahead and ref this to https://github.com/nteract/nteract/issues/4448 since they are related.

I'll also tag this as new contributor friendly in case people are interested.

PR #1985, opened 2017-10-14
[WIP] Prevent output from hijacking scroll

aaronraff:
This PR is not completely finished, just wanted to get some thoughts on this. I changed the output div so that the user must click on it before it can be scrolled. This way, you won't accidentally scroll on an output box while trying to scroll through your notebook. I'm also looking at adding a shortcut for this feature as well.  

outputscrolldemo  

This is what I currently have, any suggestions?


aaronraff:
This is in response to #1927 by the way.

rgbkrk:
Behavior wise this looks great! We do have a focused prop for the editor and cell, nothing about tracking output focus state. I'm looking at this from my phone so I can't tell if any of that logic is useful to incorporate from here.

When I get a free chance I would love to take this for a spin.

Couple questions I have for the future:

* Could we make outputs scroll to the bottom automatically as new outputs come in, provided that the user isn't actively scrolling on it?
* Would it look good if we showed something visual for when the outputs are focused or unfocused (yet larger than the minimum)?

aaronraff:
I saw that there was a prop for focused cell and editor so I was thinking maybe I should create a prop to track output focus as well. As you said, I'm not sure if that would be something useful in the future or not. As far as your questions, I will look into them and see what I can come up with!

rgbkrk:
Ok, cool. By the way, it would be great if your GIF had some more cells to drive home the scrolling a bit more. It's usually a notebook somewhere in the middle of the document with large outputs that "hijacks the scrolling".

rgbkrk:
You'll need to rebase if you go after this again, though you'll probably want to start again from the current components. Closing for now, feel free to re-open or start a new PR. Thanks @aaronraff!

Issue #1988, opened 2017-10-16
Horizontal scrollbar in cell output not click & draggable


dhirschfeld:
As it says in the title - hopefully that's clear?

You can still scroll the output left and right with the arrow keys but you can't click, hold and drag the scrollbar.

The horizontal scrollbar for the *input* does seem to work though.

rgbkrk:
@aaronraff want to check this one out?

dhirschfeld:
Code for repro:
```python
from numpy.random import randn
import pandas as pd

dates = pd.date_range('01-Jan-2018', '01-Feb-2018')
df = pd.DataFrame(randn(dates.size, 3), dates, list('ABC'))
df.T
```
I'd create a gif  but it would just show my mouse cursor moving left and right :|

aaronraff:
I'll take a look at this one!

captainsafia:
Closing as it has been resolved.

PR #1992, opened 2017-10-16
Make horizontal scrollbar draggable

aaronraff:
This references issue #1988. Basically the `.creator-hover-region` was covering up the scrollbar and not allowing the user the click on the scroll bar since it was below the hover region. To fix this, I moved the hover region below the cell. One downfall to this is that you have to hover below the cell (not just towards the bottom) in order to bring up the option to create a new cell. Let me know if this can be improved!

### DIFF
--- a/packages/desktop/src/notebook/main.css
+++ b/packages/desktop/src/notebook/main.css
@@ -411,22 +411,20 @@ body {
 }
 
 .creator-hover-region {
-  display: flex;
   position: relative;
   overflow: visible;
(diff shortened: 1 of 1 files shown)


rgbkrk:
What's this look like now?

aaronraff:
Here is what it looks like now!

horizontalscrollbar

rgbkrk:
Oh I see, yeah that's perfect.

PR #1994, opened 2017-10-19
Add state store to electron main process

### LINKED ISSUE #1930: saveAs dialog is able to be triggered when no notebook is open
Steps to reproduce:

* Run nteract on macOS
* Close all nteract windows
* Click "File -> Save As"

captainsafia:
Closes #1930 (and a couple of others).

`store` is stored in a global variable (*gasp*). This should be moved to a closure.

We might want to move other things into the main process store. For example, the GitHub token.

Also I should add tests.

But this is a start.

### DIFF
--- a/packages/desktop/src/main/index.js
+++ b/packages/desktop/src/main/index.js
@@ -28,6 +28,11 @@ import { loadFullMenu } from "./menu";
 
 import prepareEnv from "./prepare-env";
 import initializeKernelSpecs from "./kernel-specs";
+import { setKernelSpecs } from "./actions";
+
+import configureStore from "./store";
+const store = configureStore();
+global.store = store;
 
 const log = require("electron-log");
 
@@ -217,7 +222,8 @@ fullAppReady$.subscribe(() => {
   kernelSpecsPromise
     .then(kernelSpecs => {
       if (Object.keys(kernelSpecs).length !== 0) {
-        const menu = loadFullMenu(kernelSpecs);
+        store.dispatch(setKernelSpecs(kernelSpecs));
+        const menu = loadFullMenu();
         Menu.setApplicationMenu(menu);
       } else {
         dialog.showMessageBox(
--- a/packages/desktop/src/main/index.js
+++ b/packages/desktop/src/main/index.js
@@ -32,6 +32,7 @@ import { setKernelSpecs } from "./actions";
 
 import configureStore from "./store";
 const store = configureStore();
+// HACK: The main process store should not be stored in a global.
 global.store = store;
 
 const log = require("electron-log");
--- a/packages/desktop/src/main/menu.js
+++ b/packages/desktop/src/main/menu.js
@@ -61,260 +61,6 @@ export function authAndPublish(item, focusedWindow) {
   win.loadURL("https://oauth.nteract.io/github");
 }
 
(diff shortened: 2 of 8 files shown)


packages/desktop/src/main/menu.js
```diff
@@ -585,23 +249,371 @@ export function loadFullMenu(kernelSpecs) {
...
         click: createSender("menu:interrupt-kernel")
       },
       {
         label: "Restart Running Kernel",
+        enabled: BrowserWindow.getFocusedWindow() !== null,
         click: createSender("menu:restart-kernel")
       },
       {
         label: "Restart and Clear All Cells",
+        enabled: BrowserWindow.getFocusedWindow() !== null,
```
rgbkrk, line 616: Sharp! I see why this has to be redone on `focus`, `show`, etc.

rgbkrk (approved):
The cool thing about this at least is that at some point we'll be able to use this setup to update the kernelspecs so that users don't have to restart to get their available kernelspecs.

rgbkrk:
> store is stored in a global variable (gasp). This should be moved to a closure.

That's definitely ok for now, we can leave it for a refactor. We may want to leave a `TODO: ` or `HACK: ` comment for now.

packages/desktop/src/main/index.js
```diff
@@ -217,7 +222,8 @@ fullAppReady$.subscribe(() => {
   kernelSpecsPromise
     .then(kernelSpecs => {
       if (Object.keys(kernelSpecs).length !== 0) {
-        const menu = loadFullMenu(kernelSpecs);
+        store.dispatch(setKernelSpecs(kernelSpecs));
```
rgbkrk, line 226: I look forward to us reacting to new kernelspecs and changing the Menu as a result. I wish our API (internally) for this was:

```
<Menu>
  {bunchaMenus}
  <ViewMenu />
  <KernelMenu kernelspecs={kernelspecs} />
</Menu>
```

packages/desktop/src/main/reducers.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+
```
rgbkrk, line 2: 👍

packages/desktop/src/main/menu.js
```diff
@@ -552,25 +231,10 @@ export const named = {
   ]
 };
 
-export function generateDefaultTemplate() {
-  const template = [];
-
-  if (process.platform === "darwin") {
-    template.push(named);
-  }
```
rgbkrk, line 560: One way you could make this testable is by accepting the store as a variable and letting it default to using the global store:

```js
export function loadFullMenu(store=global.store) {
  ...
}
```

Or exporting a bound and unbound version.

rgbkrk:
Thanks for tackling this one, it's definitely a yucky one!

packages/desktop/src/main/menu.js
captainsafia: Sounds good! I'll add this change in the PR with the test cases.

captainsafia:
I'm gonna merge this for now and add a testing setup for main process reducers in an upcoming PR.

rgbkrk:
Awesome, thanks!

PR #2006, opened 2017-10-25
Create a README for transform-vdom

rgbkrk:
* Show off some usage
* List how to install, with the caveat that it's already part of our default transforms
* Link out to bindings in Python (sure would be nice if there were ones for R too 😉 )


rgbkrk:
I wish I would have put `skip ci` in my commit message 😄

rgbkrk:
Merging since it's a README addition that will be way more transparent than our current empty view.

PR #2017, opened 2017-10-27
bootstrap a core package

rgbkrk:
Goal: Get more of the desktop app exported for use in the web apps. I'm going to start this off with a new package to make merging later easy. After this PR I'm going to bring in all the actions, reducers, epics, and components.

-----------

Further detail:

Since exporting at a granular level from the desktop application is proving to feel more constraining (to me) than using our actions and reducers as a whole, I'm going to try out making a `core` package here that uses them directly as is here. We can keep this package free of electron specific bits and export the actions / reducers / epics as modules to incorporate within both desktop and the web app(s).

### DIFF
--- a/packages/core/.npmrc
+++ b/packages/core/.npmrc
@@ -1,0 +1,1 @@
+package-lock = false
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@nteract/core",
+  "version": "0.0.1-beta",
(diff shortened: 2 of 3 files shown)


mpacer:
I have a few clarifying questions. I'll have them written by tomorrow.

rgbkrk:
Feel free to submit feedback later, I'll go ahead and build on top of this. The plan here is to bulk export portions of the desktop app that feel stable.

PR #2020, opened 2017-10-28
[WIP][webapp] set up basics for execution

rgbkrk:
This begins the setup for executing on the nteract nbextension.

Missing pieces:

* [ ] `rx-jupyter` doesn't seem to handle tokens / the XSRF token (requires investigation)
* [ ] `channels` are not in line with `enchannel` (they're multiplexed websockets)

I'm tempted to step away from this PR for a sec and bring `rx-jupyter` into the monorepo


rgbkrk:
@captainsafia -- here's where we can start from for execution, hooking up to the jupyter API. Changes need to happen on `rx-jupyter` to handle tokens though

rgbkrk:
I'm going to let @captainsafia pick this up (even if as a new branch) while I'm doing other `@nteract/core` cleaning and integration with the original reducers.

PR #2021, opened 2017-10-28
Import rx jupyter

rgbkrk:
Bring [`rx-jupyter`](https://github.com/nteract/rx-jupyter) into the monorepo. I assume we'll want to adapt this to use jest as well. These are the pure commits, so if they come in cleanly then I think we can start adapting after.

/cc @captainsafia


rgbkrk:
🤔 Oh yeah I should delete the tag commits probably. Did I do that for the other imports? I'll come back to this after family things.

captainsafia:
Very nice! I will merge.

rgbkrk:
Oops I meant to delete the old release commits / reword them to match the lerna style like I did for other PRs.

captainsafia:
Oops. 🤦🏽‍♀️

rgbkrk:
All good, I'm going to see if I can fix it real quick. I will either rebase master or just revert this PR. I'm really sorry, I should have put WIP in the title after I initially noticed the tag commits.

rgbkrk:
Since the revert is a single commit and all the commits are interwoven, I think we'll just go with this. 😬  Time to make my clean up PR next.

PR #2022, opened 2017-10-28
remove redundant setup for rx-jupyter package

rgbkrk:
Little bits of cleanup to more fully incorporate rx-jupyter into the monorepo.

/cc @captainsafia


captainsafia:
That's the one! Thanks!

PR #2024, opened 2017-10-29
match repository link style across lerna packages

rgbkrk:
Tiny little cleanup to make sure that the links on npmjs.com resolve to the right spot in our `packages`.


lgeiger (approved):
Thanks for the cleanup 👍

PR #2029, opened 2017-10-31
move error handling into componentDidCatch to trigger error boundaries

### LINKED ISSUE #2026: Use componentDidCatch within transforms
Especially those that execute code or create react elements directly. If `transform-vdom` creates React code that can cause React to crash, it brings down the whole app. Implement [`componentDidCatch`](https://reactjs.org/docs/react-component.html#componentdidcatch) for the transforms, especially `transform-vdom`.

* [ ] `@nteract/transform-vdom`
* [ ] `@nteract/transforms`
* [ ] `@nteract/display-area`

mpacer:
Closes #2026 

I'm implementing the solution suggested for catching errors. 

Right now the error catching was "already happening" inside the vdom element, so it was easy enough to work around. 

However, as I'm trying to deal with other mimetypes. I'm having trouble figuring out how to trigger the error in a way that componentDidCatch will be tripped. It seems that the method needs to be defined in a higher order component. I'm going to commit the non-working LaTeX approach in a bit so that people can weigh in.

### DIFF
--- a/packages/transform-vdom/src/index.js
+++ b/packages/transform-vdom/src/index.js
@@ -18,29 +18,33 @@ export default class VDOM extends React.Component<Props> {
     return nextProps.data !== this.props.data;
   }
 
+  componentDidCatch(error, info) {
+    return (
+      <div>
+        <pre
+          style={{
+            backgroundColor: "ghostwhite",
+            color: "black",
+            fontWeight: "600",
+            display: "block",
+            padding: "10px",
+            marginBottom: "20px"
+          }}
+        >
+          There was an error rendering VDOM data from the kernel or notebook
+        </pre>
+        <code>{error.toString()}</code>
+      </div>
+    );
+  }
+
   render(): React$Element<any> {
     try {
       // objectToReactElement is mutatitve so we'll clone our object
       var obj = cloneDeep(this.props.data);
       return objectToReactElement(obj);
     } catch (err) {
-      return (
-        <div>
-          <pre
-            style={{
-              backgroundColor: "ghostwhite",
-              color: "black",
-              fontWeight: "600",
-              display: "block",
-              padding: "10px",
-              marginBottom: "20px"
-            }}
-          >
-            There was an error rendering VDOM data from the kernel or notebook
-          </pre>
-          <code>{err.toString()}</code>
-        </div>
-      );
+      return componentDidCatch(err, "");
     }
   }
 }
--- a/packages/transform-vdom/src/index.js
+++ b/packages/transform-vdom/src/index.js
@@ -18,7 +18,7 @@ export default class VDOM extends React.Component<Props> {
     return nextProps.data !== this.props.data;
   }
 
-  componentDidCatch(error, info) {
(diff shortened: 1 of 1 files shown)


packages/transform-vdom/src/index.js
```diff
@@ -18,29 +18,33 @@ export default class VDOM extends React.Component<Props> {
...
-              padding: "10px",
-              marginBottom: "20px"
-            }}
-          >
-            There was an error rendering VDOM data from the kernel or notebook
-          </pre>
-          <code>{err.toString()}</code>
-        </div>
-      );
+      return componentDidCatch(err, "");
```
rgbkrk: I'm glad you kept the original `catch` in here too, since the case I wanted to make sure was covered was during the `ReactDOM.render` which is called at the root level while this local catch here makes sure that `React.createElement` goes properly.

packages/transforms/src/latex.js
```diff
@@ -27,12 +28,16 @@ export default class LaTeXDisplay extends React.Component<Props> {
...
-        }}
-      />
+      <MimeWrapper>
+        <div
+          ref={el => {
+            throw new Error("anything");
+            this.el = el;
+          }}
+        />
+      </MimeWrapper>
```
rgbkrk: I'd be happy if you made this PR only the vdom stuff, since we can handle the individual cases separately.

mpacer: oh ok… well the vdom was the easiest to do… i'm having trouble even raising an error on the mimerender side i'll make a different branch for that

rgbkrk: I'd put a `componentDidCatch` here: https://github.com/nteract/nteract/blob/master/packages/display-area/src/richest-mime.js

Since that's the rendering spot for any generic transform. I'm _guessing_ you can trigger it by putting in junk data -- `json` when it's expecting a string for example with the image transform.

packages/transform-vdom/src/index.js
```diff
@@ -18,29 +18,33 @@ export default class VDOM extends React.Component<Props> {
     return nextProps.data !== this.props.data;
   }
 
+  componentDidCatch(error, info) {
```
rgbkrk: To get this to pass `flow`, you'll want to set the function signature to:

```
componentDidCatch(error: Error, info: React.ErrorInfo) {
```

packages/transform-vdom/src/index.js
rgbkrk: This should be:

```
return this.componentDidCatch(err, "");
```

packages/transform-vdom/src/index.js
mpacer: grazie! wasn't sure what type of thing the info would be.

also i'm wondering if we should just have a componentDidCatch method at the top level of the application…

mpacer:
I feel like this approach to using didComponentCatch is missing the point of it somewhat given all the ways people have talked about using it. In no cases I've found are people explicitly passing in an error.

mpacer:
closing in favour of #2031

PR #2030, opened 2017-10-31
Mimetype wrapper

mpacer:
This is part of closing #2026. 

This attempts to deal with the more general case of adding error catching with didComponentCatch for all Mimetype rendering. 

Unfortunately, the approach I took required changing many files. I'm sure there's a way inside  `index.js` to apply this wrapper in a format agnostic fashion. However, I'm a little stumped as to how to do that since there are no actual components defined inside index.js.

### DIFF
--- a/packages/transforms/src/html.js
+++ b/packages/transforms/src/html.js
@@ -1,5 +1,6 @@
 /* @flow */
 import React from "react";
+import MimeWrapper from "./mimewrapper";
 
 type Props = {
   data: string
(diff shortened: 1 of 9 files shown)


mpacer:
I took this approach because the `didComponentCatch` method needs to be defined one layer above the component that is throwing the error, so wrapping everything in this manner seemed like the best way to go.

rgbkrk:
Yeah instead of doing these at each individually, I go for the top layer which is at https://github.com/nteract/nteract/blob/master/packages/display-area/src/richest-mime.js instead of building in the `MimeWrapper` into each of these.

mpacer:
When i've tried doing it at the level of RichestMime it catches it but then stops rendering everything. When i do it at the lower levels it catches it and still renders everything else.

mpacer:
closing in favour of #2031

PR #2031, opened 2017-10-31
Display error catch

### LINKED ISSUE #2026: Use componentDidCatch within transforms

mpacer:
OK…

So this should take care of everything I was trying to do in #2030 #2029 and then close #2026.

Happy to work on the formatting for the error, but this seems much more of an idiomatic approach than either #2030 or #2029.

### DIFF
--- a/packages/display-area/src/richest-mime.js
+++ b/packages/display-area/src/richest-mime.js
@@ -13,10 +13,6 @@ type Props = {
   models?: Object
 };
 
-type ErrorProps = {
-  children?: React$Node
-};
-
 type ErrorInfo = {
   componentStack: string
 };
@@ -50,8 +46,8 @@ const Fallback = ({ componentStack, error }: FallbackProps) => (
   </div>
 );
 
-class ErrorBoundary extends React.Component<ErrorProps, State> {
-  constructor(props: ErrorProps) {
+export default class RichestMime extends React.Component<Props, State> {
+  constructor(props: Props) {
     super(props);
     this.state = {
       error: null,
@@ -63,20 +59,6 @@ class ErrorBoundary extends React.Component<ErrorProps, State> {
     this.setState({ error, info });
   }
 
-  render() {
-    if (this.state.error) {
-      return (
-        <Fallback
-          componentStack={this.state.info ? this.state.info.componentStack : ""}
-          error={this.state.error}
-        />
-      );
-    }
-    return this.props.children;
-  }
-}
-
-export default class RichestMime extends React.Component<Props> {
   static defaultProps = {
     transforms,
     displayOrder,
@@ -101,6 +83,15 @@ export default class RichestMime extends React.Component<Props> {
   }
 
   render(): ?React$Element<any> | null {
+    if (this.state.error) {
+      return (
+        <Fallback
+          componentStack={this.state.info ? this.state.info.componentStack : ""}
+          error={this.state.error}
+        />
+      );
+    }
+
     const mimetype = richestMimetype(
       this.props.bundle,
       this.props.displayOrder,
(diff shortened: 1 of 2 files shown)


mpacer:
I'm really confused, when I run this locally returning the childrens' props… 

any thoughts would be appreciated.

rgbkrk:
This is looking good, I’ll take it for a spin locally later today. Thanks for finding the cleanest way to go about this.

rgbkrk:
Looking good.

screen shot 2017-10-31 at 10 29 39 am

I'll get the tests updated and put some formatting in for the message.

rgbkrk:
screen shot 2017-10-31 at 11 00 26 am

rgbkrk:
Expanded:

screen shot 2017-10-31 at 11 02 08 am

packages/display-area/src/richest-mime.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import * as React from "react";
+import React from "react";
```
rgbkrk: Running through why I made some changes...

The default export on React should be preferred here.

packages/display-area/src/richest-mime.js
```diff
@@ -14,37 +14,62 @@ type Props = {
 };
 
 type ErrorProps = {
-  children?: React.Node
+  children?: any
```
rgbkrk: Actually maybe I shouldn't have changed this one...

packages/display-area/src/richest-mime.js
```diff
@@ -14,37 +14,62 @@ type Props = {
...
+};
+
+type ErrorInfo = {
+  componentStack: string
```
rgbkrk: I wanted to have a consistent type here, I imagine we'll move these into a package we use across the overall app.

I wonder if we could/should use these types directly: https://github.com/facebook/react/blob/2b3592331efd4b41f19de521528f664b870393e5/packages/react-reconciler/src/ReactFiberScheduler.js#L70

packages/display-area/src/richest-mime.js
```diff
@@ -13,6 +13,69 @@ type Props = {
...
+  error: ?Error,
+  info: ?ErrorInfo
+};
+
+type FallbackProps = {
+  componentStack: string,
+  error: Error
+};
+
+const Fallback = ({ componentStack, error }: FallbackProps) => (
```
mpacer, line 30: Great idea to make this a standalone component. This is making me think the ErrorBoundary should exist as a separate module so anywhere in the codebase can access it.

mpacer (commented):
I think this should be in its own module what should I read to figure out how to manage that infrastructure change?

packages/display-area/src/richest-mime.js
rgbkrk: Amusingly my next commit completely counters that by making `RichestMime` be an error boundary itself.

Motivation:

* Simplified testing around the one component
* Semantically, we care about a single output that doesn't render properly in that we can take other actions based on this (scrubbing it from the notebook, reporting errors specific to this, etc.)

mpacer: Yep I was thinking errorBoundary should be a module that takes a renderprop and as we need it we can customise the rendering (in this case using the Fallback Component).

mpacer: To clarify, I think semantically it makes more sense to have an ErrorBoundary wherever we discover the need for  it, customizing the formatting to be specific to that case.

And with that consideration… I would normally hesitate from making code architecture decisions on the sole basis of it being a test more akin to what was already present. 

Basically I think we'll eventually want error handling elsewhere and we should deal with these problems (incl. the testing issue) in a way that makes it easy for people in the future to add error handling without needing to work through these issues again.

rgbkrk: Since React 16 is pretty new, I think they wrote out `ErrorBoundary` as a concept when really there's only one new lifecycle method -- `componentDidCatch`. The way I'd probably shape this after is to list the offending mimetype, data, etc. since we can give more context right here.

rgbkrk:
Thanks for tackling this @mpacer, glad we have it in!

PR #2036, opened 2017-11-01
fix: add PlotlyTransform and PlotlyNullTransform.

theengineear:
The PlotlyNullTransform is our sneaky way of ignoring a Plotly HTML
MIMETYPE. When in offline mode, the plotly.py library attempts to do
two things:

1. On init_notebook_mode, it attempts to load in PlotlyJS. Instead we
   want to do nothing in this case.
2. On `iplot`, it attempts to run a <script> that makes a plot. Instead
   we want to handle the raw JSON string.

For the second case, we have an `application/*` MIMETYPE that we can
use, however, for the first case, we basically register the MIMETYPE so
that we can later ignore it.


theengineear:
@rgbkrk sorry for the ridiculous delay 😿

rgbkrk:
You still came around to it. 😄

rgbkrk (approved):
Yeah this is awesome!

PR #2042, opened 2017-11-03
create a starter storybook setup

rgbkrk:
This starts off a storybook setup for us, starting with the input component. Since our styles are still coupled to CSS files that exist only in the desktop app, it's time to start migrating them into here.

My primary thinking for this is that we'll use inline styling via `styled-jsx`, accepting CSS variables to control what it looks like.

/cc @klebba

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
     "postinstall": "npm run lerna",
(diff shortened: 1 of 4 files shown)


rgbkrk:
After cloning this branch and installing it (since you'll need the new dependencies), cd into `packages/core` and run `npm run storybook`. The server should start on `127.0.0.1:9001`.

rgbkrk:
Closing in favor of #2057.

PR #2055, opened 2017-11-09
Add Authorization header to requests in rx-jupyter

captainsafia:
Follow up on #2020.

### DIFF
--- a/packages/rx-jupyter/src/base.js
+++ b/packages/rx-jupyter/src/base.js
@@ -11,12 +11,16 @@ export function createAJAXSettings(
 ): Object {
   const baseURL = normalizeBaseURL(serverConfig.endpoint || serverConfig.url);
   const url = `${baseURL}${uri}`;
+  const headers = {
+    Authorization: `token ${serverConfig.token ? serverConfig.token : ""}`
+  };
   // Merge in our typical settings for responseType, allow setting additional options
   // like the method
   const settings = Object.assign(
     {
       url,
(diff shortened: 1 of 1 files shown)


packages/rx-jupyter/src/base.js
```diff
@@ -11,12 +11,16 @@ export function createAJAXSettings(
 ): Object {
   const baseURL = normalizeBaseURL(serverConfig.endpoint || serverConfig.url);
   const url = `${baseURL}${uri}`;
+  const headers = {
+    Authorization: `token ${serverConfig.token ? serverConfig.token : ""}`
```
rgbkrk, line 15: When the user doesn't use a `token` (by setting `token=""` on the Python side), what does this end up doing? Does the Jupyter server still take it?

rgbkrk: I ask because I absolutely don't know. Thanks for diving into uncharted territory for us here. 😄

rgbkrk:
Going ahead and merging because this is definitely the simplest usage possible with the security recommendations we ask of users ("use tokens, don't disable them").

packages/rx-jupyter/src/base.js
captainsafia: If there is no token set, this will just be ignored. In the Jupyter server, there is a little handler that checks to see if the token is enabled in the config. If it is, it checks to see if this `Authorization` header is passed or if a cookie was given in the request.

Issue #2039, opened 2017-11-03
yarn workspaces


rgbkrk:
It did not occur to me that yarn had "workspaces" to provide the ability to do monorepo. https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/ Have you looked at this @lgeiger or @captainsafia? Does it offer benefits that would make us want to move off of lerna?

rgbkrk:
Oooh, looks like the `yarn.lock` is only at the base and that you can `cd` into a directory to run commands as is... 🤔

rgbkrk:
> Do Yarn Workspaces make Lerna obsolete?

> Not at all. Yarn Workspaces are easily integrated with Lerna.

rgbkrk:
So I guess we could be using it today by adding it to our lerna setup...

```
{
  "npmClient": "yarn",
  "useWorkspaces": true
}
```

captainsafia:
I did some research into this. I particularly like the fact that Workspaces removes duplicated dependencies. The lerna bootstrapping process takes quite some time and even gets my CPU going quite a bit.

> For avid Lerna users this is similar to bootstrapping code via the --hoist flag.

Could we add this flag to our bootstrap process?

rgbkrk:
> Could we add this flag to our bootstrap process?

Yeah, we used to use it. @lgeiger swapped it out due to an issue with electron builder -- he told me that it's since been fixed so we can switch back. I'll see if I can make our repo switch over to this today.

rgbkrk:
The only major annoyance I'm seeing for us + adoption is that we have to make contributors install yarn and set this:

```
yarn config set workspaces-experimental true
```

rgbkrk:
I've changed everything locally that is necessary, time to push a PR and see what folks think. 😬

rgbkrk:
<details>
<summary>Current run</summary>

```
$ yarn
yarn install v1.3.2
info No lockfile found.
[1/4] 🔍  Resolving packages...
warning babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update!
warning sinon > build > wrench@1.3.9: wrench.js is deprecated! You should check out fs-extra (https://github.com/jprichardson/node-fs-extra) for any operations you were using wrench for. Thanks for all the usage over the years.
[...]

PR #2058, opened 2017-11-09
clean up core, extract styling into next gen components

rgbkrk:
Some of the chief complaints I've gotten about nteract is that you have to "weave" props through to change inner components, like the `<Editor />`. Another chief irritation I've had (and has made it harder for others) is that our styles don't come with our components.

This PR takes another look at how we do components while making sure that:

* [x] components can be server side rendered
* [x] styles come with the components (we should not have N versions of `main.css`
* [x] there are less nteract desktopisms within the components
* [x] components are "compound" and declarative
* [x] we have a showcase to make it easy to work on UI aspects without a full desktop or jupyter app running

```jsx
<Cell>
  <Input>
    <Prompt counter={11} />
    <Editor>{"import wizardry
wizardry.birthday()"}</Editor>
  </Input>
  <Outputs>
    <img
    src="https://i.ytimg.com/vi/YFNsRogBqb0/maxresdefault.jpg"
    width="200"
    />
    <h2>yer a wizard harry</h2>
  </Outputs>
</Cell>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,8 +78,7 @@
   "dependencies": {
     "immutable": "^3.8.1",
     "react": "^16.0.0",
-    "react-dom": "^16.0.0",
-    "styled-jsx": "^2.1.3"
+    "react-dom": "^16.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
   "dependencies": {
     "immutable": "^3.8.1",
     "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react-dom": "^16.0.0",
+    "styled-jsx": "^2.1.3"
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -24,6 +24,7 @@
     "react-dnd": "^2.4.0",
     "react-dnd-html5-backend": "^2.4.1",
     "react-redux": "^5.0.5",
+    "redux": "^3.7.0",
     "react-simple-dropdown": "^3.0.0",
     "react-syntax-highlighter": "^5.8.0",
     "rxjs": "^5.5.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -16,7 +16,7 @@
     "@nteract/commutable": "^2.1.2",
     "@nteract/display-area": "^3.0.2",
     "@nteract/editor": "^3.0.2",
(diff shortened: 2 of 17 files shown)


packages/core/package.json
```diff
@@ -25,13 +25,18 @@
...
+    "styled-jsx": "^2.1.3",
     "uuid": "^3.1.0"
   },
   "publishConfig": {
     "access": "public"
   },
-  "keywords": ["nteract", "ohgodamonomoduleinamonorepo"],
+  "keywords": [
+    "nteract",
+    "ohgodamonomoduleinamonorepo"
```
captainsafia, line 39: LOL.

packages/showcase/pages/core.js
```diff
@@ -0,0 +1,293 @@
...
+
+const {
+  Cell,
+  Input,
+  Prompt,
+  PromptBuffer,
+  Editor,
+  Outputs,
+  Notebook
+} = _nextgen;
```
captainsafia, line 12: I'm curious: why not import each of these directly from `@nteract/core/components` instead of having `_nextgen` be an intermediary?

As in, could this be:

`import { Cell, Input } from '@nteract/core/components';`

package.json
```diff
@@ -78,7 +78,8 @@
   "dependencies": {
     "immutable": "^3.8.1",
     "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "react-dom": "^16.0.0",
+    "styled-jsx": "^2.1.3"
```
captainsafia: Should this `styled-jsx` dependency exist at the top-level? We only use it in the `core` submodule right?

rgbkrk: Also in `notebook-preview-demo`, `showcase` (and `commuter` externally).

I should make sure this can work without the top-level babelrc. It was just easier to place it at the top and ensure our babel setup was consistent across components.

packages/showcase/pages/core.js
rgbkrk: I was extremely scared of the fact that they're not _really_ the full `Cell` and `Notebook` components yet, when right now desktop is using the "old" `notebook.js`.

package.json
rgbkrk: At some point we need to put `styled-jsx` back at the top level, after we have hoisting back in place (yay for yarn workspaces #2039). The main reason is that there is flicker on load when there is more than one version of `styled-jsx` on the page (even if the same version, as with non-hoisted lerna they have an independent version).

flicker


I could also make the lerna app do hoisting as is for styled-jsx...

captainsafia:
Great stuff! Merging it!

PR #2060, opened 2017-11-11
Introduce @nteract/core into nbextension

rgbkrk:
This begins bringing in `@nteract/core` (I actually started this a few days ago), though we still need a nice clean way to work across our various reducers and epics. I'll get that cleaned up as I keep trying to bring the nbextension into fruition.


rgbkrk:
Bringing this in as its iterative on the nbextension. Keep on moving forward.

Issue #2004, opened 2017-10-25
Menu items / actions not showing sometimes


rgbkrk:
The Menu context updates work most of the time yet sometimes I've been able to get it into a state where the notebook document is focused but the menu no longer has the relevant options available.

<img width="860" alt="screen shot 2017-10-25 at 9 29 19 am" src="https://user-images.githubusercontent.com/836375/32010547-006d1528-b967-11e7-82c4-4ddafb4ea6bf.png">

I can do this reproducibly by exporting my notebook as a PDF, clicking to open my PDF, then command-tabbing back to the notebook app.

captainsafia:
Oops. Yeah. I noticed this but couldn't consistently reproduce. I wonder if it is related to the keyboard action for focus that can't be picked up.

rgbkrk:
Oh man, I've made this come up by trying to use the devtools too. :( I'll look into this on the weekend if I get a free chance.

rgbkrk:
Done!

PR #2067, opened 2017-11-16
enable menu items if _any_ windows exist for app

rgbkrk:
Since the original issue (#1930) being fixed was to prevent the saveAs dialog from being able to work when there are no windows at all and the current behavior (intermittently working) is preventing release, I've switched the `enabled` toggle to just check that there are _any_ windows. For the cases reported in #2004 this fixes the issue. We probably still need to track down the right events. I don't want it to hold us back from releasing though.


captainsafia:
This is very good. Thank you! 🙇‍♀️

PR #2069, opened 2017-11-16
Cell creator and toolbar next gen

rgbkrk:
Switched over cell creator and toolbar to using styled-jsx. While I was at it, I included Reactified SVG for the octicons.


rgbkrk:
I think I'll bring the toolbar into this one as well.

rgbkrk:
Woo, everything passes since I re-kicked AppVeyor. 😄

captainsafia:
Nice! Digging the SVG octagons.

PR #2074, opened 2017-11-17
turn draggable cell into an opinionated styled component

rgbkrk:
This converts draggable cell into a simpler component that takes a cell as a `child` element and accepts two callbacks related to drag movement: `selectCell` and `moveCell`. It also includes the CSS as has been done with the rest of the `styled-jsx` introductions.


rgbkrk:
~~Time to fix tests and possibly flow.~~ Done

rgbkrk:
This was not nearly as bad as I thought and I think I've got a little more I can clean up in the big notebook component. I can leave it in a new PR though. 😄

PR #2076, opened 2017-11-19
Cleanup core/components/notebook

rgbkrk:
This turned out to be a major refactor in some parts, ending up with nice styles for the pinned cell (vs. our last release) while I was at it.

`v0.3.4`:

screen shot 2017-11-18 at 8 09 55 pm

This PR:

screen shot 2017-11-18 at 8 11 58 pm

### DIFF
--- a/packages/core/src/components/notebook.js
+++ b/packages/core/src/components/notebook.js
@@ -308,7 +308,21 @@ export class Notebook extends React.PureComponent<Props> {
             padding-top: 10px;
             padding-left: 10px;
             padding-right: 10px;
-            padding-bottom: calc(100vh - 110px);
+          }
+
+          @media not print {
+            .cells {
+              padding-bottom: calc(100vh - 110px);
+            }
+          }
+        `}</style>
+        <style jsx global>{`
+          /* Show the cell-toolbar-mask if hovering on cell,
+             or cell was the last clicked (has .focused class). */
+
+          .cell:hover .cell-toolbar-mask,
+          .cell.focused .cell-toolbar-mask {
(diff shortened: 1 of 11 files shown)


rgbkrk:
Also fixed something else I noticed, which was that codemirror hints were appearing below sticky cells. I made CodeMirror hints' `z-index` high to counter this.

packages/core/src/components/notebook.js
```diff
@@ -29,7 +29,9 @@ const PropTypes = require("prop-types");
 
 type Props = {
   displayOrder: Array<string>,
-  notebook: any,
+  // TODO: Make types stricter, we have definitions _somewhere_
+  cellOrder: ImmutableList<any>,
+  cellMap: ImmutableMap<string, any>,
```
betatim, line 34: for people from the future: what is the thinking behind switching from a `notebook` to having these two unpacked variables?

betatim:
This is too big for me to seriously review, but travis seems happy and coverage didn't go down that much so ... what could go wrong?

packages/core/src/components/notebook.js
rgbkrk: Inside this component we only ever use `cellMap` and `cellOrder` so this reduced the amount of `this.props.notebook.get('cellMap')` that was going on. At some point this file should probably be re-named since the nextgen components have their own `<Notebook />` container while this one is more like `<NotebookApp />`

rgbkrk:
Thanks @betatim. Luckily this PR is more shuffling than additions.

PR #2077, opened 2017-11-19
extract sticky cell as separate component

rgbkrk:
This takes some lifecycle that was previously managed in the ol' notebook component and brings it into it's own special container for the sticky cells. Usage:

```
<StickyCellContainer>
  <Cell />
  <Cell />
</StickyCellContainer>
```

### DIFF
--- a/packages/core/src/components/notebook.js
+++ b/packages/core/src/components/notebook.js
@@ -46,6 +46,69 @@ type Props = {
   language: string
 };
 
+class StickyCellContainer extends React.Component<*, *> {
+  stickyCellsPlaceholder: ?HTMLElement;
+  stickyCellContainer: ?HTMLElement;
+
+  componentDidUpdate(prevProps: Props): void {
+    if (this.stickyCellsPlaceholder && this.stickyCellContainer) {
+      // Make sure the document is vertically shifted so the top non-stickied
+      // cell is always visible.
+      this.stickyCellsPlaceholder.style.height = `${
+        this.stickyCellContainer.clientHeight
+      }px`;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


packages/core/src/components/notebook.js
```diff
@@ -46,6 +46,69 @@ type Props = {
...
 };
 
+class StickyCellContainer extends React.Component<*, *> {
+  stickyCellsPlaceholder: ?HTMLElement;
+  stickyCellContainer: ?HTMLElement;
+
+  componentDidUpdate(prevProps: Props): void {
+    if (this.stickyCellsPlaceholder && this.stickyCellContainer) {
+      // Make sure the document is vertically shifted so the top non-stickied
+      // cell is always visible.
```
rgbkrk, line 56: This is the original code so I left it as is. _However_, I think that this likely misses some of the resizing it would need. I'm not fussing with it though. I think the real solution is some proper CSS that would adapt to the changing size.

packages/core/src/components/notebook.js
```diff
@@ -193,62 +244,21 @@ export class Notebook extends React.PureComponent<Props> {
...
 
   renderCell(id: string): React$Element<any> {
     const cell = this.props.cellMap.get(id);
-    const transient = this.props.transient.getIn(
-      ["cellMap", id],
-      new ImmutableMap()
-    );
+
+    const running =
+      this.props.transient.getIn(["cellMap", id, "status"]) === "busy";
```
rgbkrk, line 273: Oh whoop, I snuck this one in. I can make it another PR, it's just cleaning up what ends up being only one field in use.

PR #2078, opened 2017-11-20
Cleanup editor

rgbkrk:
In the process of trying to make it easier to work with the components we have here, I've started cleaning up the CodeMirror editor, moving some of the jupyter protocol bits into a `jupyter` directory and starting to place modes in a `mode` directory (singularity comes from matching up with codemirror import styles). In the process I've converted the old codemirror component we forked from [react-codemirror](https://github.com/JedWatson/react-codemirror) to using an ES6 class.


rgbkrk:
This doesn't go the whole way for unifying the codemirror component and our little wrapper -- it gets us started though and I'm happy to say that we now have some basic flow types here.

packages/editor/src/highlighter.js
```diff
@@ -1,50 +0,0 @@
-// @flow
-// To be used for server side rendering
-// Not finished yet
```
rgbkrk, line 3: I'm the one that put this in as a partially complete implementation (that isn't even referenced locally here). I'm also the one that's happy to remove it. 😛

rgbkrk:
A little more work to do here and I think we'll have a nice clean surface for our editor.

Issue #2040, opened 2017-11-03
Suggestions: Softwrap text and drag and drop between notebooks


MasonProtter:
* There should be an option to disallow text from continuing offscreen and make it wrap onto a new line in cell. 

* It would be very nice to be able to use the drag and drop gesture to copy a code cell to another notebook

rgbkrk:
Thanks for reporting.

Related issues (in the opposite direction): #1940, #1921 

/cc @betatim

MasonProtter:
I can see and understand the desire for horizontal scrolling which is why I suggested making soft-wrap optional, possibly togglable per cell?

betatim:
I'd be cool with togglable (per cell). And because I can't remember: do we currently soft wrap for mardkown cells?

MasonProtter:
The rendered markdown is soft wrapped but the code is not.

betatim:
+ 💯 for softwrap in markdown cells :)

PR #2081, opened 2017-11-20
Create full codemirror component

rgbkrk:
In an attempt to get a handle on one of the murkiest parts of our code base I merged together our wrapper editor and our codemirror react code. Some highlights that should make maintainability easier and extensibility easier:

## There is now only _one_ name for the source code input: `value`

This makes the component line up better with `<textarea />` and other `<input />` elements.

## CodeMirror `options` can passed in directly

One criticism we have had is that it's hard to override our editor options. This surfaces it a bit more, even though it's not core to the app itself. We'll want to think on a good design here as well as make sure we're managing it well.

## Reduce dependencies

Just a little bit of cleanup

# TODO

* [x] Fix up flow and tests
* [x] Clean up the top level divs (there should likely only be _one_ around the `textarea`)
* [x] Introduce the massive amount of styles that belong with the component
* [x] Ensure CSS variables have all the default values from the light theme


packages/editor/src/index.js
```diff
@@ -257,25 +355,33 @@ class CodeMirrorEditor extends React.Component<WrapperProps> {
...
-          className="cell_cm"
-          options={options}
-          onChange={onChange}
-          onClick={() => {
-            if (this.codemirror) this.codemirror.focus();
-          }}
-          onFocusChange={onFocusChange}
-        />
+        <div className={editorClassName}>
+          <div className="CodeMirror cm-s-composition CodeMirror-wrap">
```
betatim: do you know if you really need `CodeMirror-wrap` as class? In my digging around yesterday I think I decided you don't want to hard this code (as well as removing something  in `main.css`:
```
 .CodeMirror pre {
   padding: 0;
-  white-space: pre;
-  border: 0;
-  border-radius: 0;
 }
```
to get the `lineWrapping` option to have any effect.

rgbkrk: Yeah I don't think we need this class anymore, I'm looking to remove all the unused divs here and end up with the same styling.

For the line wrapping, I'm curious if codemirror itself has an option to enable it instead, which we'd override per Markdown and Code Editors.

betatim: Should have been clearer: `lineWrapping` is the name of the codemirror option to turn line wrapping on and off. So once we can pass `options` from the code and markdown cell parts down into code mirror we will be golden 😁

rgbkrk: ohhhhhhh yayyyyyyyyy

well that's about to be 8000x times easier to deal with now

rgbkrk: I _think_ I can delete this whole defintion here. Yay for soon to be sane styles.

rgbkrk:
Now that I know this is working, I'm tempted to do the style extraction in a separate PR.

packages/core/src/components/cell/cell.js
```diff
@@ -213,6 +213,10 @@ export class ConnectedCell extends React.PureComponent<CellProps, *> {
               focusBelow={this.focusBelowCell}
               cellFocused={cellFocused}
               editorFocused={editorFocused}
+              options={{
+                // Markdown should always be line wrapped
+                lineWrapping: true
+              }}
```
rgbkrk, line 219: This one is for you @betatim!

betatim: Fixes half of #2040 !

rgbkrk:
Fixed the tooltips! (Howdy @alexandercbooth! I didn't realize we broke it -- oops).

packages/editor/src/index.js
```diff
@@ -291,33 +292,54 @@ class CodeMirrorEditor extends React.Component<
...
+              .tip {
+                padding: 20px 20px 50px 20px;
+                margin: 30px 20px 50px 20px;
+                box-shadow: 2px 2px 50px rgba(0, 0, 0, 0.2);
+                white-space: pre-wrap;
+                background-color: var(--main-bg-color);
+                z-index: 9999999;
+              }
+            `}</style>
+          </div>,
```
rgbkrk, line 333: /cc @alexandercbooth -- I made a bit more of this inline React. I think we can do even more with a [React Portal](https://reactjs.org/docs/portals.html). I aimed for "just work" for now. Would love to make the styles nice and easy here while also having clean DOM ops. 😄

rgbkrk:
Fixed up the rest of the tests, I'll come back to put the default values on CSS variables and then this will be all set. Super ready for review, I seriously hope I'm done with these shuffling PRs soon. 😉

packages/desktop/src/notebook/index.js
```diff
@@ -12,7 +12,7 @@ import NotificationSystem from "react-notification-system";
 import configureStore from "./store";
 // TODO: This should pull from a regular import that webpack and jest will use to
 //       load the right copy
-import Notebook from "@nteract/core/lib/components/notebook";
+import Notebook from "@nteract/core/lib/components/notebook.js";
```
captainsafia, line 15: Why is there a `.js` here?

captainsafia (commented):
Looks great!

packages/desktop/src/notebook/index.js
rgbkrk: I kept getting an error for this import after a clean rebuild. Definitely need to finalize this refactor so it’s not doing a deep grab of an unsupported API.

PR #2085, opened 2017-11-21
Fix windows CLI

### LINKED ISSUE #2084: Install Shell Commands produces Javascript Error
When I click ```Help > Install Shell Commands``` in a freshly installed nteract (Windows 10, x64 release 0.3.4). I get following Javascript Error message:  

nteract_javascript_error

lgeiger:
This should fix #2084

Could anybody test this on Windows?

Side Note: I think refactoring this file to use `async` / `await` would probably make this more readable.

### DIFF
--- a/packages/desktop/src/main/cli.js
+++ b/packages/desktop/src/main/cli.js
@@ -1,5 +1,5 @@
-import { Observable } from "rxjs/Observable";
-import { merge, catchError, mergeMap } from "rxjs/operators";
+import { catchError, mergeMap } from "rxjs/operators";
+import { merge } from "rxjs/observable/merge";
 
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):
This looks right to me!

rgbkrk:
I wonder if there’s a way to test the script from appveyor. I really wish there was some magical way to only test the code that only affects windows (path handling). We end up mocking all the things that touch the system when we probably should have a few integration or functional tests.

rgbkrk:
I'm going to operate under the assumption that this is in a good state and make it part of the shipped app tonight.

lgeiger:
:+1:

PR #2091, opened 2017-11-22
vendorize normalize.css

rgbkrk:
Bring the parts of normalize.css in that we're actually using with our Chrome based app into the `index.html`. I am on a CSS import rampage, likely tackling #1720 after I finish all these. 😄

### DIFF
--- a/packages/desktop/src/notebook/main.css
+++ b/packages/desktop/src/notebook/main.css
@@ -1,6 +1,5 @@
 /*
   App
  */
-@import "../../node_modules/normalize.css/normalize.css";
 @import "../../node_modules/nteract-assets/fonts/source-sans-pro/source-sans-pro.css";
 @import "../../node_modules/nteract-assets/fonts/source-code-pro/source-code-pro.css";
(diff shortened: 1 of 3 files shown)


packages/desktop/src/notebook/main.css
```diff
@@ -1,7 +1,7 @@
 /*
   App
  */
-@import "../../node_modules/normalize.css/normalize.css";
+/*@import "../../node_modules/normalize.css/normalize.css";*/
```
lgeiger: We can now also remove `normalize.css` as a dependency from the `desktop/package.json`.

lgeiger (commented):
👍

packages/desktop/src/notebook/main.css
```diff
@@ -1,6 +1,5 @@
 /*
   App
  */
-@import "../../node_modules/normalize.css/normalize.css";
 @import "../../node_modules/nteract-assets/fonts/source-sans-pro/source-sans-pro.css";
 @import "../../node_modules/nteract-assets/fonts/source-code-pro/source-code-pro.css";
```
rgbkrk, line 5: The fonts themselves I'm unsure of what to do about when I do the `<base>` changing. Guess we'll see here!

PR #2095, opened 2017-11-25
pagers should have same css as outputs

### LINKED ISSUE #2093: Docstrings broken in 0.4.0
In the latest version docstrings are no longer rendered correctly - e.g.

image

...it appears that the line-breaks aren't being respected.

rgbkrk:
Fixes #2093.

/cc @dhirschfeld

screen shot 2017-11-25 at 9 53 33 am


rgbkrk:
Erring on the side of getting these regression fixes shipped quickly.

PR #2104, opened 2017-11-27
stick in missing style for editor on markdown cell

### LINKED ISSUE #2100: css problem with markdown cells (0.4.0)
nteract

Note: As well as showing the narrow width of the markdown cell it also demonstrates the error in #2099 which I stumbled upon accidentally. Specifically, after I change the cell to be markdown type then click in  it the cursor disappears and I can't select it. I got out of that state by using `Ctrl-Shift-N` to create a new cell which I could then select after which I could then click back into the markdown cell.

rgbkrk:
If there's a sure sign that this refactor isn't finished and me used to the components, it's that this is what broke the markdown cells in the last release.

Closes #2100

### DIFF
--- a/packages/core/src/components/cell/cell.js
+++ b/packages/core/src/components/cell/cell.js
@@ -204,20 +204,22 @@ export class ConnectedCell extends React.PureComponent<CellProps, *> {
             editorFocused={editorFocused}
             cell={cell}
           >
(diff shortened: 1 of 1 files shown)


rgbkrk:
I am _definitely_ prioritizing a release after this is merged.

PR #2109, opened 2017-11-28
specify our builtin eventsource as a ponyfill

rgbkrk:
🐴

Ensure the default `EventSource` is the ponyfill.


betatim:
Now you are just horsing around ...

PR #2111, opened 2017-11-28
Clear execution count when clearing cell output

### LINKED ISSUE #2107: Clear All Outputs doesn't clear execution counts
Similar for Restart and Clear All Cells.

After clearing all outputs the execution count remains:
image

I would expect the execution count to also be cleared - i.e. it should be a blank `[ ]` after clearing the cells. This is useful to clearly distinguish which cells have been run and which haven't after a restart.

betatim:
Fixes #2107 

Is `null` the right thing to set the execution count to? It works but maybe there is something that I should be using.

### DIFF
--- a/packages/core/src/reducers/document.js
+++ b/packages/core/src/reducers/document.js
@@ -151,7 +151,9 @@ function clearOutputs(state: DocumentState, action: ClearOutputsAction) {
 
   if (type === "code") {
     return cleanCellTransient(
-      state.setIn(["notebook", "cellMap", id, "outputs"], new Immutable.List()),
+      state
+        .setIn(["notebook", "cellMap", id, "outputs"], new Immutable.List())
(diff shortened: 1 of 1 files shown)


betatim:
I can't find any tests for the reducers, worth getting started now with adding one? A pointer to an example to copy would be good.

rgbkrk (approved):
At this point the execution count should get cleared by a different action but I think this is fine.

betatim:
Could also add a second action, though you could argue that "clear outputs" always means "remove the outputs and reset the exec counts". At least as a user I'd expect that always to happen together.

rgbkrk:
There are tests for the reducers, it looks like they need to be reorganized though since they're not in a `reducers` directory. Here's the document one for instance: https://github.com/nteract/nteract/blob/master/packages/core/__tests__/document-spec.js

PR #2112, opened 2017-11-28
include token and session_id with kernel ws URL

rgbkrk:
The token needs to be included on authenticated servers, this adapts our current websocket URL creator to include it.

While I was at it I included a way to identify the session as well.

/cc @jameswomack @captainsafia


captainsafia (approved):
Looks good!

rgbkrk:
Thanks! I'm going to ship these now so I can keep using them without a local fork.

PR #2119, opened 2017-11-30
create another incarnation of play.nteract.io

rgbkrk:
This relies on [mybinder](mybinder.org) to launch a server, then a kernel, and connect to it.

~~I'm running into an issue with `rx-jupyter`, particularly in making sure that cross domain works just fine. Tokens are good, trouble is now CORS.~~ Update: Fixed!

Motivations:

* Kickstart channel socket adapations (following up with @captainsafia's work)
* Provide a `https://play.golang.org/` like site where people can share snippets, run code
* Run a service that uses our components on the web
* Play with binder!


rgbkrk:
To run this, after doing the initial dev setup of the repo:

```
./node_modules/.bin/lerna run dev --scope @nteract/play --stream
```

(or plain `lerna` at the front if you have it installed globally)

rgbkrk:
screen shot 2017-11-30 at 2 54 06 pm
screen shot 2017-11-30 at 2 54 10 pm

packages/rx-jupyter/src/base.js
```diff
@@ -19,13 +19,17 @@ export function createAJAXSettings(
   const settings = Object.assign(
     {
       url,
-      responseType: "json"
+      responseType: "json",
+      createXHR: function() {
+        return new XMLHttpRequest();
+      }
```
rgbkrk, line 25: This overrides the defaults that RxJS sets up for an `XMLHTTPRequest`, including setting some headers that for CORS Jupyter does not accept.

packages/rx-jupyter/src/base.js
```diff
@@ -19,13 +19,17 @@ export function createAJAXSettings(
...
     },
     serverConfig,
     opts,
     {
       // Make sure we merge in the auth headers with user given headers
-      headers: Object.assign({}, headers, opts.headers)
+      headers: Object.assign({}, headers, opts.headers),
+      crossDomain: true
```
rgbkrk: Oh wait, I should remove this since it should get set from the call / setup in our `serverConfig`.

packages/rx-jupyter/src/kernels.js
```diff
@@ -5,8 +5,9 @@ import { webSocket } from "rxjs/observable/dom/webSocket";
 import { Observable } from "rxjs/Observable";
 
 import { createAJAXSettings } from "./base";
+import { omit } from "lodash";
```
rgbkrk: This should be deleted.

packages/rx-jupyter/src/kernels.js
```diff
@@ -5,8 +5,9 @@ import { webSocket } from "rxjs/observable/dom/webSocket";
...
 
-import { URLSearchParams } from "url";
+const URLSearchParams = require("url-search-params");
```
rgbkrk: The node based library for `url` was not getting properly imported with webpack, so I used this "polyfill".

rgbkrk:
Once this is brought in, I'm going to deploy it. Heck, I think I'll deploy it now.

Oh wait, I'd need the changes to `rx-jupyter` deployed too.

packages/rx-jupyter/src/base.js
betatim: Interesting that RxJS changes around `XMLHTTPRequest`.

packages/rx-binder/index.js
```diff
@@ -40,13 +41,14 @@ function formBinderURL(
...
+  EventSourceDI /* :* */ = EventSourcePolyfill
 ) /*: Observable<*> */ {
   const url = formBinderURL(options);
+
   return Observable.create(observer => {
-    const es = new EventSource(url);
+    const es = new EventSourceDI(url);
 
-    es.onmessage = evt => {
+    es.onmessage = function(evt) {
```
betatim, line 51: why can't we keep the fat arrow?

rgbkrk: Oh yeah, let me dig into that. It seemed like I was running into an issue with EventSource not liking this be bound with a different `this`.

packages/play.nteract.io/pages/index.js
```diff
@@ -0,0 +1,230 @@
...
+    this.state = {
+      binderMessages: [],
+      kernelMessages: [],
+      serverConfig: null,
+      error: null
+    };
+  }
+
+  async kernelLifecycle(kernel) {
+    await new Promise(resolve =>
```
betatim, line 39: this is a pattern to do "promise where i don't care if you call `resolve`"?

betatim: wrap something that doesn't return a promise so that you can await it?

betatim:
LGTM, my comments are mostly for my education.

packages/play.nteract.io/pages/index.js
rgbkrk: That’s exactly it. setState takes a callback for when the state is really updated and rendered. Since I wanted to stay within this async await setup I used a promise to wrap around the callback.

PR #2124, opened 2017-12-02
Add dockerfile just for playground

yuvipanda:

### DIFF
--- a/packages/play.nteract.io/.dockerignore
+++ b/packages/play.nteract.io/.dockerignore
@@ -1,0 +1,1 @@
+node_modules/*
--- a/packages/play.nteract.io/Dockerfile
+++ b/packages/play.nteract.io/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:8
+ADD . /playground
+WORKDIR /playground
+RUN npm install && npm run build
+
+EXPOSE 3000
+
+CMD ["/playground/node_modules/.bin/next", "start", "--host", "0.0.0.0"]


betatim:
Prodded travis to see if the failure is random or not. Don't really see why this should be failing.

rgbkrk:
It looks like the build is failing on `master`, which is my fault in hastily fixing certain things for packaging so we could ship the app. I'm getting this locally on `master` now too... `O_____O`

rgbkrk:
Hmmmm only with a fresh install, one dep must have changed that affects the flow definitions. It's bombing on all:

```
export type ...
```

declarations.

betatim:
@yuvipanda 🔄🎸 (rebase) now that this is fixed, then we can merge this :)

Should we setup building and pushing for releases?

rgbkrk:
We'd need all the packages it relies on to be released too. It would be nice to have a travis group that only runs on tagged releases and it would build & ship the docker images for apps that are webapps.

rgbkrk:
Seeing if the close and reopen works, since it should be testing the merge commit rather than as is.

betatim:
> Seeing if the close and reopen works, since it should be testing the merge commit rather than as is.

Smart!

betatim:
🎉

betatim:
This has been built and pushed to dockerhub at least once: https://hub.docker.com/r/yuvipanda/play.nteract.io/

betatim:
> It would be nice to have a travis group that only runs on tagged releases

I had assumed you were  doing releases by tagging on GH and letting travis doing all the work.

rgbkrk:
The JS package releases are pretty simple. I think things are just changing for the likely proliferation of apps built on the packages here.

yuvipanda:
This is awesome!

Ideally, I'd love to have automated docker build + push whenever anything under here changes to something like `nteract/playground` (or `nteract/play` as you wish).

PR #2130, opened 2017-12-03
[play] expand display-area by default

rgbkrk:
Without this PR:

screen shot 2017-12-02 at 5 43 46 pm

With this PR:

screen shot 2017-12-02 at 5 43 37 pm

/cc @betatim @yuvipanda

### DIFF
--- a/packages/play.nteract.io/pages/index.js
+++ b/packages/play.nteract.io/pages/index.js
@@ -295,7 +295,7 @@ display(
 
         <div className="play-outputs">
           <Outputs>
-            <Display outputs={this.state.outputs} />
+            <Display outputs={this.state.outputs} expanded />
           </Outputs>
         </div>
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
am testing this locally now

yuvipanda:
So the output area expands and I can scroll down, but the code are a does not seem to expand. This causes a color difference on the left side as you scroll down. Is that intentional?

yuvipanda:
so it looks like https://screenshots.firefox.com/Dpnh9FSK4KjWvbfU/localhost (after pressing 'enter' to trigger scrollbar on left). Am ok to merge as is though, since it's clearly an improvement over current setup.

rgbkrk:
> This causes a color difference on the left side as you scroll down. Is that intentional?

Nope. Definitely need to make that full height with both. Not sure how to fix that at the moment. 🤔

yuvipanda:
Do you wanna do that in a separate PR?

rgbkrk:
Yeah it’s ok to call this an iterative improvement and create an issue for the color on the left down below.

yuvipanda:
w00t, my first merge!

PR #2132, opened 2017-12-03
[play] load up the jupyter-stacks image by default

rgbkrk:
This sets the default image for `@nteract/play` to use [binder-examples/jupyter-stacks](https://github.com/binder-examples/jupyter-stacks) which uses [`all-spark-notebook`](https://github.com/jupyter/docker-stacks/tree/master/all-spark-notebook) underneath.

At some point we'll want to expose a way for the users to set the repo and ref they want. For now (and in general), we'll set a reasonable default here that has a good collection of packages.

/cc @yuvipanda

### DIFF
--- a/packages/play.nteract.io/pages/index.js
+++ b/packages/play.nteract.io/pages/index.js
@@ -192,7 +192,7 @@ display(
 
   async getServer() {
     const serverConfig = await binder(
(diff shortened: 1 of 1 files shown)


yuvipanda:
Happy to merge when tests pass. The default should be configurable too...

PR #2135, opened 2017-12-03
provide command for running play app from the root

rgbkrk:
To run `@nteract/play` in dev mode from this repo you can now run:

```
npm run app:play
```


yuvipanda:
Tested locally, works!

PR #2136, opened 2017-12-04
flow type enchannel-zmq-backend

rgbkrk:
Set up flow types for `enchannel-zmq-backend`. This is a precursor to some sweeping changes to our zmq backed channels since we'll be trying to get `channels` to conform to the same interface across both remote kernels that use Jupyter's websockets and local ZMQ kernels.

I'm going to leave some commentary inline about why I deleted some things (since I can't leave inline comments in code for something taken away).

### DIFF
--- a/flow-typed/npm/rxjs_v5.0.x.js
+++ b/flow-typed/npm/rxjs_v5.0.x.js
@@ -1011,9 +1011,16 @@ declare class rxjs$Observer<T> {
   complete(): mixed;
 }
 
-// FIXME(samgoldman) should be `mixins rxjs$Observable<T>, rxjs$Observer<T>`
-// once Babel parsing support exists: https://phabricator.babeljs.io/T6821
-declare class rxjs$Subject<T> extends rxjs$Observable<T> {
+// NOTE: This is the nteract definition, for our subjects -- not that of flow-typed
+// Bring in the latest after https://github.com/flowtype/flow-typed/pull/1616 is
+// reviewed and merged.
+// prettier-ignore
+declare class rxjs$Subject<T> extends rxjs$Observable<T> mixins rxjs$Observer<T> {
+  static create<T>(
+    destination: rxjs$Observer<T>,
+    source: rxjs$Observable<T>
+  ): rxjs$AnonymousSubject<T>;
+
   asObservable(): rxjs$Observable<T>;
 
   observers: Array<rxjs$Observer<T>>;
@@ -1028,6 +1035,11 @@ declare class rxjs$Subject<T> extends rxjs$Observable<T> {
   // For use in subclasses only:
   _next(value: T): void;
   _subscribe(observer: rxjs$PartialObserver<T>): rxjs$Subscription;
+}
+
+declare class rxjs$AnonymousSubject<T> extends rxjs$Subject<T> {
+  destination: rxjs$Observer<T>;
+  source: rxjs$Observable<T>;
 }
 
 declare class rxjs$BehaviorSubject<T> extends rxjs$Subject<T> {
--- a/packages/enchannel-zmq-backend/src/index.js
+++ b/packages/enchannel-zmq-backend/src/index.js
@@ -1,8 +1,16 @@
+// @flow
 import { SHELL, STDIN, IOPUB, CONTROL } from "./constants";
 import { Subject } from "rxjs/Subject";
 import { Subscriber } from "rxjs/Subscriber";
-import { Observable } from "rxjs";
+import { Observable } from "rxjs/Observable";
+
+import { merge } from "rxjs/observable/merge";
+
+import { map } from "rxjs/operators";
+
 import { createSubject, createSocket } from "./subjection";
+
+import type { JUPYTER_CONNECTION_INFO } from "./subjection";
 
 /**
  * convertToMultiplex converts an enchannel multiplexed message to a
@@ -11,7 +19,7 @@ import { createSubject, createSocket } from "./subjection";
  * @param {Object}  message The enchannel message to convert
  * @return  {Object}  Converted message
  */
-export function convertToMultiplex(message) {
+export function convertToMultiplex(message: any) {
   return Object.assign({}, message.body, { channel: message.type });
 }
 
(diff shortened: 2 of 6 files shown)


flow-typed/npm/rxjs_v5.0.x.js
```diff
@@ -1011,9 +1011,15 @@ declare class rxjs$Observer<T> {
   complete(): mixed;
 }
 
-// FIXME(samgoldman) should be `mixins rxjs$Observable<T>, rxjs$Observer<T>`
-// once Babel parsing support exists: https://phabricator.babeljs.io/T6821
+// NOTE: This is the nteract definition, for our subjects -- not that of flow-typed
+// Bring in the latest after https://github.com/flowtype/flow-typed/pull/1616 is
+// reviewed and merged.
 declare class rxjs$Subject<T> extends rxjs$Observable<T> {
```
rgbkrk: Gah, our prettier hook got rid of `mixins rxjs$Observer<T> `

rgbkrk: ref: https://github.com/prettier/prettier/issues/3390

packages/enchannel-zmq-backend/src/index.js
```diff
@@ -58,7 +73,7 @@ export function createMainChannelFromChannels(shell, control, stdin, iopub) {
         }
       }
     }),
-    Observable.merge(
```
rgbkrk, line 61: This was unintentionally increasing the size of the bundle since it was bringing in all of Rx.

packages/enchannel-zmq-backend/src/index.js
```diff
@@ -96,87 +115,34 @@ export function createChannels(identity, config, subscription = "") {
...
- * @param  {Object} config                  Jupyter connection information
- * @param  {string} config.ip               IP address of the kernel
- * @param  {string} config.transport        Transport, e.g. TCP
- * @param  {string} config.signature_scheme Hashing scheme, e.g. hmac-sha256
- * @param  {number} config.shell_port       Port for shell channel
- * @return {Rx.Subject} subject for sending and receiving messages on the shell
- *                      channel
- */
-export function createChannelSubject(channel, identity, config) {
-  return createSubject(createSocket(channel, identity, config));
```
rgbkrk, line 113: I deleted `createChannelSubject` as it was not providing type safety given that `createSocket` takes an enum for `CHANNEL_NAME`. All the usage switches to direct calls of `createSubject(createSocket(...))` now.

packages/enchannel-zmq-backend/src/index.js
```diff
@@ -96,87 +115,34 @@ export function createChannels(identity, config, subscription = "") {
...
- * @param  {Object} config                  Jupyter connection information
- * @param  {string} config.ip               IP address of the kernel
- * @param  {string} config.transport        Transport, e.g. TCP
- * @param  {string} config.signature_scheme Hashing scheme, e.g. hmac-sha256
- * @param  {number} config.iopub_port       Port for iopub channel
- * @param  {string} subscription            subscribed topic; defaults to all
- * @return {Rx.Subject} subject for receiving messages on the shell_port
- *                      channel
- */
-export function createIOPubSubject(identity, config, subscription = "") {
```
rgbkrk, line 177: All these comments for internal functions for creating subjects are overkill with the flowtypes in place.

rgbkrk:
Once everything is adapted to the one multiplexed interface, a lot of the boilerplate in this code base can be cleaned up. We'll have one clean `Subject.create` that handles the `session` ID and `username` directly, like is now done in `rx-jupyter`.

rgbkrk:
```
 FAIL  packages/enchannel-zmq-backend/__tests__/index_spec.js (9.984s)
  ● createMainChannel › pipes messages from socket appropriately
    TypeError: shell.source.map is not a function
      
      at createMainChannelFromChannels (packages/enchannel-zmq-backend/src/index.js:155:2605)
      at Object.<anonymous> (packages/enchannel-zmq-backend/__tests__/index_spec.js:33:58)
          at new Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)

```

Odd, this passes locally. `source` should exist on an `AnonymousSubject`, which is the return from a `Subject.create`.

rgbkrk:
The error is quite legitimate, apparently with a fresh install the desktop app is now broken. I'll figure it out!

rgbkrk:
Oh of course! This was implicitly using `map` when we need to bring in the operator. Fix underway.

packages/enchannel-zmq-backend/src/index.js
```diff
@@ -1,17 +1,25 @@
+// @flow
 import { SHELL, STDIN, IOPUB, CONTROL } from "./constants";
 import { Subject } from "rxjs/Subject";
 import { Subscriber } from "rxjs/Subscriber";
-import { Observable } from "rxjs";
+import { Observable } from "rxjs/Observable";
+
+import { merge } from "rxjs/observable/merge";
```
theengineear, line 7: I think `merge` exists in `rxjs/operators` as well, fwiw. I think that might be the better way to import it.

packages/enchannel-zmq-backend/src/index.js
```diff
@@ -96,87 +125,34 @@ export function createChannels(identity, config, subscription = "") {
...
- * kernel's control channel
- * @param  {string} identity                UUID
- * @param  {Object} config                  Jupyter connection information
- * @param  {string} config.ip               IP address of the kernel
- * @param  {string} config.transport        Transport, e.g. TCP
- * @param  {string} config.signature_scheme Hashing scheme, e.g. hmac-sha256
- * @param  {number} config.control_port     Port for control channel
- * @return {Rx.Subject} subject for sending and receiving messages on the control
- *                      channel
- */
```
theengineear, line 143: nice clean up 😺

theengineear (approved):
I probably don't have all the context on this, but your changes make sense to me and overall, seems like a great cleanup. I had one comment that you might want to address regarding an RxJS import, but that's about it!

packages/enchannel-zmq-backend/src/index.js
rgbkrk: The `merge` creation method is different from the operator I think, but I'll check it out.

rgbkrk: It is ever so slightly different, though it does use the operator for `mergeAll` under the hood: https://github.com/ReactiveX/rxjs/blob/master/src/observable/merge.ts#L83

rgbkrk:
I think I'm going to preemptively merge this instead of waiting on AppVeyor. Thanks for checking it out @theengineear!

PR #2141, opened 2017-12-12
UI for specifying a specific binder repo & redirecting to it

mpacer:
This creates simple UI elements for linking to a specific binder repository (as opposed to just the main binder page with the binder image).

image

On submitting the form it will take you to the appropriate URL (in that example: https://mybinder.org/v2/gh/binder-examples/python2_with_3/master )as though you had filled out binder's github repo and branch/tag/commit fields where it will build the appropriate image and launch when it is ready.

The style needs a lot of work (right now the text box doesn't even expand to accommodate the GH repo name, but this functionality is stateless and straightforward. 

I also plan to factor out this form into it's own component to make it easier to rearrange going forward.

### DIFF
--- a/packages/play.nteract.io/components/consoles.js
+++ b/packages/play.nteract.io/components/consoles.js
@@ -1,11 +1,8 @@
 import * as React from "react";
 
-// TODO: Make a generic little console for some of the styled container pieces,
-//       then make this component inject the binder specific bits
-export class BinderConsole extends React.Component {
+class BinderUI extends React.Component {
   render() {
     const {
-      logs,
       onFormSubmit,
       onGitrefChange,
       onRepoChange,
@@ -13,55 +10,39 @@ export class BinderConsole extends React.Component {
       gitref
     } = this.props;
     return (
-      <div className="binder-console">
-        <div className="binder-ui-wrapper">
-          <a className="anchor" href="https://mybinder.org" target="_blank">
-            <img
-              src="https://mybinder.org/static/logo.svg?v=f9f0d927b67cc9dc99d788c822ca21c0"
-              alt="binder logo"
-              height="20px"
+      <div className="binder-ui-wrapper">
+        <a className="anchor" href="https://mybinder.org" target="_blank">
+          <img
+            src="https://mybinder.org/static/logo.svg?v=f9f0d927b67cc9dc99d788c822ca21c0"
+            alt="binder logo"
+            height="20px"
+          />
+        </a>
+        <form onSubmit={onFormSubmit} className="form">
+          <fieldset>
+            <label htmlFor="repoInput">GitHub Repo: </label>
+            <input
+              id="repoInput"
+              onChange={onRepoChange}
+              type="text"
+              name="repo"
+              value={repo}
             />
-          </a>
-          <form onSubmit={onFormSubmit} className="form">
-            <fieldset>
-              <label htmlFor="repoInput">GitHub Repo: </label>
-              <input
-                id="repoInput"
-                onChange={onRepoChange}
-                type="text"
-                name="repo"
-                value={repo}
-              />
-            </fieldset>
-            <fieldset>
-              <label htmlFor="gitrefInput">Branch/commit: </label>
-              <input
-                id="gitrefInput"
-                onChange={onGitrefChange}
-                type="text"
-                name="gitref"
-                value={gitref}
-              />
-            </fieldset>
-            <fieldset>
-              <button type="submit"> Submit</button>
-            </fieldset>
-          </form>
-        </div>
-        {logs.length > 0
-          ? logs.map((log, index) => {
-              return (
-                <span className="log" key={index}>
-                  <span className="sidebar" />
-                  <span className="phase">{log.phase}</span>
-                  <span className="content">
-                    <span className="message">{log.message}</span>
-                  </span>
-                </span>
-              );
-            })
-          : null}
-
+          </fieldset>
+          <fieldset>
+            <label htmlFor="gitrefInput">Branch/commit: </label>
+            <input
+              id="gitrefInput"
+              onChange={onGitrefChange}
+              type="text"
+              name="gitref"
+              value={gitref}
+            />
+          </fieldset>
+          <fieldset>
+            <button type="submit"> Submit</button>
+          </fieldset>
+        </form>
         <style jsx>{`
           img {
             vertical-align: middle;
@@ -94,6 +75,39 @@ export class BinderConsole extends React.Component {
             float: left;
           }
 
+          .binder-ui-wrapper {
+            width: 100%;
+            display: inline-block;
+            padding: 0 0 0 10px;
+          }
+        `}</style>
+      </div>
+    );
+  }
+}
+// TODO: Make a generic little console for some of the styled container pieces,
+//       then make this component inject the binder specific bits
+export class BinderConsole extends React.Component {
+  render() {
+    const { logs, ...otherprops } = this.props;
+    return (
+      <div className="binder-console">
+        <BinderUI {...otherprops} />
+        {logs.length > 0
+          ? logs.map((log, index) => {
+              return (
+                <span className="log" key={index}>
+                  <span className="sidebar" />
+                  <span className="phase">{log.phase}</span>
+                  <span className="content">
+                    <span className="message">{log.message}</span>
+                  </span>
+                </span>
+              );
+            })
+          : null}
+
+        <style jsx>{`
           .log {
             padding: 0 15px 0 32px;
             margin: 0;
@@ -137,12 +151,6 @@ export class BinderConsole extends React.Component {
             background-color: #1a1a1a;
             counter-reset: line-numbering;
             margin-top: 0;
-          }
-
-          .binder-ui-wrapper {
-            width: 100%;
-            display: inline-block;
-            padding: 0 0 0 10px;
           }
 
           .log:last-child {
--- a/packages/play.nteract.io/components/consoles.js
+++ b/packages/play.nteract.io/components/consoles.js
@@ -26,11 +26,15 @@ export class BinderConsole extends React.Component {
           <form onSubmit={this.handleSubmit} className="form">
             <p className="para">
               GitHub Repo:
-              <input type="text" defaultValue="hello" name="ghrepo" />
+              <input
+                type="text"
+                defaultValue="binder-examples/python2_with_3"
+                name="ghrepo"
+              />
             </p>
             <p className="para">
               Branch/commit:
-              <input type="text" defaultValue="hello" name="branch" />
+              <input type="text" defaultValue="master" name="branch" />
             </p>
             <button type="submit"> Submit</button>
           </form>
--- a/packages/play.nteract.io/components/consoles.js
+++ b/packages/play.nteract.io/components/consoles.js
@@ -3,6 +3,14 @@ import * as React from "react";
 // TODO: Make a generic little console for some of the styled container pieces,
 //       then make this component inject the binder specific bits
 export class BinderConsole extends React.Component {
+  handleSubmit = event => {
+    event.preventDefault();
+    let ghrepo = event.target.ghrepo.value;
+    let branch = event.target.branch.value;
+    const url = "https://mybinder.org/v2/gh/" + ghrepo + "/" + branch;
+    console.log(url);
+  };
+
   render() {
     const { logs } = this.props;
     return (
@@ -15,8 +23,16 @@ export class BinderConsole extends React.Component {
               height="20px"
             />
           </a>
-          <form className="form">
-            <input defaultValue="hello" />
+          <form onSubmit={this.handleSubmit} className="form">
+            <p className="para">
+              GitHub Repo:
+              <input type="text" defaultValue="hello" name="ghrepo" />
+            </p>
+            <p className="para">
+              Branch/commit:
+              <input type="text" defaultValue="hello" name="branch" />
+            </p>
+            <button type="submit"> Submit</button>
           </form>
         </div>
         {logs.length > 0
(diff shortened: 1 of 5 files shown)


packages/play.nteract.io/components/consoles.js
```diff
@@ -3,17 +3,42 @@ import * as React from "react";
 // TODO: Make a generic little console for some of the styled container pieces,
 //       then make this component inject the binder specific bits
 export class BinderConsole extends React.Component {
+  handleSubmit = event => {
+    event.preventDefault();
+    let ghrepo = event.target.ghrepo.value;
+    let branch = event.target.branch.value;
+    const url = "https://mybinder.org/v2/gh/" + ghrepo + "/" + branch;
+    window.location = url;
```
betatim: For my education: why do we set this? I was expecting us to use this information to create a new/update a rx-binder observable.

mpacer: Could you say more about what you mean?

I assumed that we'd need to redirect to the standard binder interface since it provides a file browser. Given that repos can have more than one notebook I wasn't sure how the current play interface would accommodate that & introducing such a UI would have a much larger scope.

But I recognise that I may not be following you entirely, & in any case I'm definitely open to discussion.

rgbkrk: The intention for play is to provide a http://jsbin.com/ like experience -- code on the left, outputs on the right. The binder image is the dependency behind the scenes (as is the kernel that gets launched). We don't want to do the full binder experience, as people can go to mybinder.org for that.

packages/play.nteract.io/components/consoles.js
```diff
@@ -3,17 +3,42 @@ import * as React from "react";
...
-        </a>
+        <div className="binder-ui-wrapper">
+          <a className="anchor" href="https://mybinder.org" target="_blank">
+            <img
+              src="https://mybinder.org/static/logo.svg?v=f9f0d927b67cc9dc99d788c822ca21c0"
+              alt="binder logo"
+              height="20px"
+            />
+          </a>
+          <form onSubmit={this.handleSubmit} className="form">
```
rgbkrk: I'd bubble the `onSubmit` up to a passed in prop:

```
<form onSubmit={this.props.onSubmit} />
```

You may still end up having a handleSubmit if you want to do any cleaning or normalizing before passing it up to the parent component via the callback prop, so it's also fine to keep `handleSubmit`. I just wouldn't make `handleSubmit` _do_ anything to the page.

packages/play.nteract.io/components/consoles.js
```diff
@@ -3,17 +3,42 @@ import * as React from "react";
...
+            <p className="para">
```
rgbkrk: I'd use a `<fieldset>` and a `<label>` here for each of these in keeping with the accessibility of plain HTML forms. I can help style them if they feel hairy to work with, CSS wise.

packages/play.nteract.io/components/consoles.js
```diff
@@ -3,17 +3,42 @@ import * as React from "react";
...
+              GitHub Repo:
+              <input
+                type="text"
+                defaultValue="binder-examples/python2_with_3"
+                name="ghrepo"
+              />
+            </p>
+            <p className="para">
+              Branch/commit:
+              <input type="text" defaultValue="master" name="branch" />
```
rgbkrk: We're going to want to keep this as a "controlled" component, so I'd use `value={this.props.ref}` and not use `defaultValue`, while also setting the default prop for `value`.

rgbkrk:
Thanks for going after this, it's going to be awesome to be able to change the image backing the play instance on demand. Once we get this situated, we should figure out how we want to represent a particular fiddle / pen.

Right now I'm thinking query parameters would be fine. We shouldn't make them part of this PR though, as this one could just be for setting up the initial form, state handling, and server/kernel lifecycle. I'll post a new issue for the query parameter & persistence layer.

packages/play.nteract.io/components/consoles.js
mpacer: I didn't address this in the most recent push. Will start on that now.

packages/play.nteract.io/components/consoles.js
```diff
@@ -4,16 +4,46 @@ import * as React from "react";
 //       then make this component inject the binder specific bits
 export class BinderConsole extends React.Component {
   render() {
-    const { logs } = this.props;
+    const {
+      logs,
+      handleFormSubmit,
+      handleGitrefChange,
+      handleRepoChange,
```
rgbkrk: When exposed as a prop use a leading `on` -- `onGitRefChange`, `onRepoChange`, `onSubmit`.

packages/play.nteract.io/pages/index.js
```diff
@@ -275,7 +295,14 @@ display(
         </header>
 
         {this.state.showPanel ? (
-          <BinderConsole logs={this.state.binderMessages} />
+          <BinderConsole
+            handleGitrefChange={this.handleGitrefChange}
+            handleRepoChange={this.handleRepoChange}
+            handleFormSubmit={this.handleFormSubmit}
```
rgbkrk: Follow up to above comment about the naming patterns. Here it will look like:

```js
<BinderConsole
  onGitRefChange={this.handleGitRefChange}
  onRepoChange={this.handleRepoChange}
```

You can still use `handle` locally. The common practice on props is that you'll lead with `on`. This makes them match the interface for pure DOM components too.

mpacer: +1

packages/play.nteract.io/components/consoles.js
mpacer: ✓

packages/play.nteract.io/pages/index.js
```diff
@@ -110,6 +115,23 @@ display(
...
+
+  handleGitrefChange(event) {
+    this.setState({
+      gitref: event.target.value
+    });
+  }
+
+  async handleFormSubmit(event) {
+    event.preventDefault();
+    await this.getServer();
```
rgbkrk: I think you're going to want to make this run `this.initialize()` so it does the kernel lifecycle too.

mpacer: Sounds good.

mpacer:
Is there a good way to get the list of available kernels inside an environment directly from binder or do we use the `kernels` from rx-jupyter to accomplish that?

rgbkrk:
> Is there a good way to get the list of available kernels inside an environment directly from binder or do we use the kernels from rx-jupyter to accomplish that?

We'd have to use the jupyter kernelspecs API to find out.

mpacer:
> We'd have to use the jupyter kernelspecs API to find out.

~~Do you mean these: https://github.com/jupyter/notebook/blob/master/notebook/kernelspecs/handlers.py~~

~~because if so, both of the handlers assume that you know the name of the kernel and so this may require a notebook server change to be possible…~~

Ah! so `/api/kernelspecs` should do what we need to do. My guess is that should be done in rx-jupyter. Is that out of scope for this PR or should i work on that.

rgbkrk:
A `GET` on `/api/kernelspecs` with no kernel name in the path returns the full list of kernels as JSON

screen shot 2017-12-12 at 2 52 26 pm

Let's mark switching the kernel for another PR though, as it's an additional complication. We should just assume our current default is fine for iterating on only the form field for the repo.

rgbkrk:
By the way, this is the notebook source file you'd want to look at for that API: https://github.com/jupyter/notebook/blob/master/notebook/services/kernelspecs/handlers.py#L49

packages/play.nteract.io/components/consoles.js
mpacer: When you say accessibility of plain HTML forms, can you point me to a resource as to what you mean by that? Is there something special I need to do to ensure that (e.g., am I not supposed to have a callback)?

rgbkrk: To be a bit more explicit, don't use `<p>` here, use `<label>`. Then group the `<label>` and the `<input>` in a [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) like this:

```js
<fieldset>
    <label htmlFor="repoInput">GitHub Repo: </label>
    <input id="repoInput" type="text" name="ghrepo" defaultValue="binder-examples/python2_with_3" />
</fieldset>
```

My general rule of thumb, for the sake of making sure to accomodate accessibility, is to always use elements that describe their intent by being those elements. Screen readers know how to treat those, the tab behavior gets naturally handled by the browser, other accessibility tools know how to work with them, and quite a bit more that I have no idea about as a non-expert in a11y. While I'm not super worried about accessibility, I prefer to do the light easy steps when I can.

rgbkrk:
New review comments, since I've taken it for a spin -- when the form is submitted, I'd expect the logs to be cleared out, ready for the new logs.

mpacer:
So, if we only clear the logs (see recent commit)… this results slightly bizarre behaviour if you submit multiple times in rapid succession:

image

Thoughts?

rgbkrk:
> So, if we only clear the logs (see recent commit)… this results slightly bizarre behaviour if you submit multiple times in rapid succession:

That's totally fine for now, we'll want to interrupt the previous call to binder once another submission is hit (in Rx parlance, we'd use a `switchMap` and we'd also want to shutdown the previous server to be kind). The way `play` is written right now, it's fairly unavoidable (without more state management around events).

mpacer:
Can we shut down the previous server today? or is there no API for that?

rgbkrk:
There's an API, see https://github.com/nteract/nteract/issues/2126

mpacer:
why is eslint failing? I'm confused because both of the labels have an htmlFor attribute that matches the input ids…

mpacer:
K, got the shutdown working before switching to a new image (with no delay unlike suggested in #2126 because we aren't shutting down the page… so we can wait until we receive the response rather than aborting after some period of time). I may not have implemented that 100% correctly, but I think it's doing its job.

packages/rx-jupyter/src/index.js
```diff
@@ -15,4 +15,25 @@ function apiVersion(serverConfig: Object) {
...
+ *
+ * @param {Object} serverConfig  - The server configuration
+ *
+ * @return  {Object}  An Observable with the request/response
+ */
+function shutdown(serverConfig: Object): Observable<*> {
+  return ajax(
+    createAJAXSettings(serverConfig, "/api/shutdown", { method: "POST" })
+  );
+}
```
rgbkrk, line 29: Perfect!

packages/rx-jupyter/test/index-spec.js
```diff
@@ -20,4 +20,15 @@ describe("rx-jupyter", () => {
...
+  describe("shutdown", () => {
+    it(" Creates an AjaxObservable for shutting down a notebook server", () => {
+      const shutdown$ = jupyter.shutdown({
+        endpoint: "https://somewhere.com",
+        crossDomain: true
+      });
+      const request = shutdown$.request;
+      expect(request.url).to.equal("https://somewhere.com/api/shutdown");
+      expect(request.method).to.equal("POST");
+    });
```
rgbkrk, line 32: With a test too! You're awesome!

rgbkrk:
This is working pretty darn well!

rgbkrk:
I'll check out the lint thing now.

mpacer:
I've got to call it a night, but I did track down and fix another bug in a somewhat hacky way. 

If you were to go to an invalid address and then tried to go back to a valid address, it would fail because it no longer had a serverConfig.

Rather than doing that I'm now storing the old ServerConfig and messages and then setting them if it catches an error. I'm sure this is not the right way to do this… but at least it makes this substantially more functional than if you only implemented the killing of the server.

NB: I would have moved the state setting until after the initialize function, but whenever I did that the state setting simply never occurred. I wasn't able to figure out why that was.

betatim:
Just here to say: sweeet! This is looking nice

rgbkrk:
Based on the [docs for `jsx-a11y/label-has-for`](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md#succeed) sticking the input inside the label will make this succeed, even though they also say to use `htmlFor` here. I'll push up a commit fixing it then come back to review again.

I ended up getting rather busy last night so I wasn't able to touch this until now. Apologies.

rgbkrk:
Oops, I just realized that fixed it while also breaking our previously good CSS. I'll fix that next.

rgbkrk:
Good stack exchange post on why it’s good to put inputs inside their respective labels: https://ux.stackexchange.com/a/47920

rgbkrk:
screen shot 2017-12-13 at 10 52 10 am

packages/play.nteract.io/components/consoles.js
```diff
@@ -1,19 +1,115 @@
...
+      </div>
+    );
+  }
+}
 // TODO: Make a generic little console for some of the styled container pieces,
 //       then make this component inject the binder specific bits
 export class BinderConsole extends React.Component {
   render() {
-    const { logs } = this.props;
+    const { logs, ...otherprops } = this.props;
```
rgbkrk, line 108: As soon as you see a pattern like this, you begin to see this should be a different set of components altogether.

I think we're going to end up wanting to make this:

```js
<Drawer>
  <BinderLogo />
  <BinderForm onGitRefChange={} />
  {/* Note that I'm not reciting the exact props, just that these should be split up at some point */}
  <BinderLogs logs={} />
</Drawer>
```

I think it's ok to bring this in, but we should probably change this soon so we're not doing too much in this one component.

rgbkrk:
As long as you're cool with it (and don't want to do the component refactoring suggested above in this PR), I think we can merge it.

mpacer:
LGTM i think this is ready

PR #2142, opened 2017-12-12
delete the metadata reducer setup for nbextension

rgbkrk:
It doesn't exist anymore.

I broke master when I merged #2060. This fixes it back up.


rgbkrk:
For the sake of fixing master, I'm merging this. Thanks y'all.

(and sorry for the earlier breakage).

PR #2150, opened 2017-12-13
create individual BinderLogs and BinderLogo components

mpacer:
This follows up on a comment from #2141 to refactor the structure. I didn't actually take out the `...other props` bit because the problem isn't that it needs to be refactored further but that at the level that it's being passed down there there are just too many props for it to be easily legible.

We could break it up into a Higher-level component and subcomponents, but that would seem to isolate some of the functionality, though we could do that if it were desired by using `this.props.children`.


rgbkrk:
That's ok, we can do the next part of the refactor to compose them from `pages/index.js` in another PR.

PR #2151, opened 2017-12-13
Binder refactor

mpacer:
This continues to refactor the Binder components to be more transparent at the top level.


rgbkrk (approved):
Awesome, thanks

PR #2152, opened 2017-12-14
Npm build only

mpacer:
These `build:watch` commands are not exactly the same as `build` because they won't run flow until after the rest of the process is completed (which will never occur). I probably should just remove those `&& npm run build:flow` bits…

Right now there doesn't seem to be a good way to have type checking running as part of what is triggered with the build. 

If anyone has any ideas I'm all ears!


rgbkrk:
The build:flow part isn’t actually doing any type checking, it’s just copying over type files for us so the types are up to date inline.

rgbkrk (approved):

rgbkrk (approved):
Thanks for doing maintenance work on the monorepo and diving in to understand everything here!

PR #2153, opened 2017-12-14
[jupyter-ext] use our builtin Notebook package

rgbkrk:
This brings out nbextension back to a working state with the current core (though the store is not fully hooked up). This also cleans up a flow/jest error that will intermittently happen with the desktop app.


mpacer:
Doesn't actually make a file browser or anything, just dumps a payload to the console, but it doesn't error! And that's a great improvement!

PR #2154, opened 2017-12-14
Fix npm run build:watch to use --parallel

mpacer:


packages/ansi-to-react/package.json
```diff
@@ -14,7 +14,7 @@
     "build": "npm run build:es5 && npm run build:flow",
     "build:flow": "flow-copy-source -v -i '**/tests/**' src lib",
     "build:es5": "babel src --out-dir lib/ --source-maps",
-    "build:watch": "npm run build:es5 --watch && npm run build:flow",
+    "build:watch": "npm run build:es5 -- --watch",
```
rgbkrk: ah this is definitely one that could be adapted to the singular `lib:watch` like we've done. The es5 is an old carryover. No special things here.

rgbkrk (approved):

packages/ansi-to-react/package.json
```diff
@@ -11,11 +11,12 @@
     "test:flow": "flow",
```
rgbkrk, line 11: Note that above here `prebuild` (which runs before `build`) does an `npm run clean`. Since you changed `clean` to `build:clean` you'll have to change up the `prebuild` above here (that's why the build is failing).

mpacer: does the prebuild step need to exist? it looks like it's primary purpose is to keep the `lib/` directory around, which none of the other packages need. Either I need to change `npm run clean` to `rimraf lib/*` or make `build:clean` depart from the system used elsewhere (otherwise `npm run prebuild` just creates a directory if it didn't exist and then deletes it).

rgbkrk: Odd yeah, that's the only one doing it. Seems like you could remove it. I think this predates when babel wasn't creating the `lib` folder for us IIRC. I think it can be taken out.

mpacer: So then it should be good to go now (since now it just runs `npm run build:clean` like almost everything else).

mpacer: Or if there's other cleanup you'd like me to do I can do that more broadly — I just don't know what is and isn't vital.

package.json
```diff
@@ -37,7 +37,8 @@
     "build:packages:watch": "lerna run build:lib:watch --parallel --ignore @nteract/play --ignore @nteract/showcase --ignore nteract",
     "build:icon": "./scripts/make_icons.sh && cd static/icons && iconutil -c icns nteract.iconset && mv nteract.icns ../icon.icns",
     "build:only": "lerna run build --scope",
-    "build:only:watch": "lerna run build:watch --scope",
+    "build:only:watch": "lerna run build:watch --stream --scope",
```
rgbkrk, line 40: total nit, but I'd prefer to have this as `build:watch:only`, noting that I'd read it as "build watch, only for packageName":

```
npm run build:watch:only -- packageName
```

rgbkrk (approved):

PR #2157, opened 2017-12-15
Binder kernels — trying to make a UI so that it's configurable

mpacer:
This is really a WIP, but at the point of submitting this PR we're getting the kernelspecs and then dumping the kernel names to the left of the Kernel status area. 

My current plan is to put a selector there from which we can choose different kernels and on switching kernels it handles the kernel lifecycle startup and closure.

Unfortunately I might not have a chance to work on this again until next week at the sprint, but it seems like it's a good start.

### DIFF
--- a/applications/play/components/kernelUI.js
+++ b/applications/play/components/kernelUI.js
@@ -12,14 +12,11 @@ class KernelSelector extends React.Component {
     const { kernelspecs, currentKernel, onChange } = this.props;
     return (
       <form>
-        <label>
-          Current Kernel:
-          <select value={currentKernel} onChange={onChange}>
-            {Object.keys(kernelspecs).map((kernelName, index) => {
-              return <KernelOption kernelName={kernelName} key={kernelName} />;
-            })}
-          </select>
-        </label>
+        <select value={currentKernel} onBlur={onChange} onChange={onChange}>
+          {Object.keys(kernelspecs).map((kernelName, index) => {
+            return <KernelOption kernelName={kernelName} key={kernelName} />;
+          })}
+        </select>
         <style jsx>
           {`
             form label,
--- a/applications/play/components/kernelUI.js
+++ b/applications/play/components/kernelUI.js
@@ -1,0 +1,33 @@
+import * as React from "react";
+
+// TODO: Make a generic little console for some of the styled container pieces,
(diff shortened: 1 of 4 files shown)


applications/play/components/kernelUI.js
```diff
@@ -0,0 +1,78 @@
...
+        </select>
+        <style jsx>
+          {`
+            form label,
+            form select,
+            form option {
+              font-family: inherit;
+              font-size: inherit;
+            }
+            :global(form span.kernelOption) {
```
rgbkrk: I'm not seeing this `className` anywhere above. Is this supposed to be for the `<option>` in `KernelOption`?

applications/play/pages/index.js
```diff
@@ -23,6 +23,8 @@ const {
 } = _nextgen;
 
 import { BinderConsole } from "../components/consoles";
+import { KernelUI } from "../components/kernelUI";
+import { kernelspecs } from "rx-jupyter";
```
rgbkrk: You can move the `kernelspecs` import down into the other `rx-jupyter` import (and it would be ok to switch the below to `import` instead of `const`.

applications/play/pages/index.js
```diff
@@ -123,6 +128,40 @@ div(
...
+
+    const kernel = await this.getKernel(
+      this.state.serverConfig,
+      this.state.kernelName
+    );
+    await new Promise((resolve, reject) => {
+      this.setState({ kernel }, resolve);
+    });
+    await this.kernelLifecycle(kernel);
+  }
```
rgbkrk, line 163: 👍 I think this is good for now.

rgbkrk (approved):

packages/rx-jupyter/__tests__/kernel-spec.js
```diff
@@ -38,7 +38,7 @@ describe("kernels", () => {
       });
       expect(request.method).toBe("POST");
       expect(request.body.path).toBe("/tmp");
-      expect(request.body.kernel_name).toBe("python3000");
+      expect(request.body.name).toBe("python3000");
```
rgbkrk, line 41: 👍

mpacer:
No longer WIP!

PR #2159, opened 2017-12-15
switch build:only:watch to build:watch:only

mpacer:
The emoji in the commit message is throroughly accidental… but it makes me wonder if we're going to have difficulty conveying this command to people in github contexts?

Edit: Maybe we just need to always put it in backticks: `build:watch:only`.

I also want to update the docs where appropriate.


rgbkrk:
The emoji is awesome and yeah we will have to use backticks. :smile:

rgbkrk (approved):

PR #2162, opened 2017-12-15
Bring commuter into the nteract monorepo

rgbkrk:
This moves commuter into our nteract monorepo (and is a do-over of #2155). I had trouble using lerna import to bring in commuter (it's a lerna repo itself), so I did a subtree merge & squash. The server is in `applications/commuter` while the frontend is in `packages/commuter-frontend`. This work will lead us towards sharing more across commuter, especially as we use components destined for the general nteract web app. In particular, I want to get our directory navigator available.

* [x] Use `git subtree` to bring in commuter
* [x] Update commuter to bring it in line with current package versions
* [x] Correct issues reported by linter, flow
* [x] Fork the timeago component
* [x] Fix the flow types for the timeago component
* [x] Switch the tests from the timeago component to jest

PR #2163, opened 2017-12-17
create an octicons package from core + commuter

rgbkrk:
Follow on from #2162, unifying our octicon SVG React components.

Note that I switched it to inline styles here. We may want to go back to styled-jsx at some point, I just didn't want to add in the babel setup for this little package and for so few styles.


rgbkrk:
One omission on this PR — I think commuter front end is still depending on the official octicons package and it should get deleted out of the package.json

captainsafia:
🍨

PR #2164, opened 2017-12-17
Jest all the things

### LINKED ISSUE #1911: Switch all tests to jest
Now that we're completely a monorepo and the desktop package is no longer a special case, I'd love to see all the tests switch over to jest. We'd then be able to run one big global jest run that reports coverage for us.

Bonus points: Make sure our tests are nice idiomatic jest tests instead of the old enzyme cruft we are likely carrying around

Recommended path: Do this slowly, for each package, then finally finish off the ones for `packages/desktop`.

lgeiger:
Closes #1911 to unify our testing setup

- [x] Convert desktop tests to jest
- [x] 🔥unused desktop integration tests
- [x] Convert `commuter/routesUtil` tests to jest
- [x] Convert `commuter/s3` tests to jest. This needs some mocking since `proxyquire` doesn't work with `jest`
- [x] Convert `rx-jupyter` tests to jest
- [x] Fix `rx-jupyter` tests. I think 96f83b920d86d622151d30ac791b49106822183f broke them since they weren't running on CI (@rgbkrk )
- [x] 🔥`sinon`
- [x] 🔥`sinon-chai`
- [x] 🔥`electron-mocha`
- [x] 🔥 `chai-immutable`
- [x] 🔥 `mocha`
- [x] 🔥 `chai`


rgbkrk:
```
    Expected value to equal:

      {"Content-Type": "application/json"}

    Received:

      {"Authorization": "token ", "Content-Type": "application/json"}
```

Sure enough, that needs a fix since the token is not set.

It doesn't appear to be 96f83b920d86d622151d30ac791b49106822183f though as that's adding it to query params, not the headers. I can investigate and submit a fix.

applications/commuter/__tests__/s3test.js
```diff
@@ -0,0 +1,160 @@
+// @flow
+const s3Service = require("./../src/content-providers/s3/s3");
+jest.mock("aws-sdk/clients/s3");
```
rgbkrk, line 3: Whoa didn’t realize you could mock in-line in a test

applications/commuter/package.json
```diff
@@ -48,15 +47,6 @@
...
   },
-  "devDependencies": {
-    "babel-cli": "^6.26.0",
-    "babel-preset-flow": "^6.23.0",
-    "chai": "^4.1.2",
-    "flow-bin": "^0.54.0",
-    "mocha": "^3.5.0",
-    "proxyquire": "^1.7.10",
-    "supertest": "^3.0.0"
-  },
```
rgbkrk, line 59: Goodbye dependencies!

rgbkrk (approved):
A+ cleanup!

applications/commuter/__tests__/s3test.js
lgeiger: We could probably simplify our other mocks as well

PR #2171, opened 2017-12-18
[MRG] Add URL handlers to edit/view notebooks

betatim:
Add URL handlers to edit/view a notebook fetched from a GIST.


betatim:
Looks ugly but works!

rgbkrk:
Oops, looks like we need a `.npmrc` with `package-lock=false` at the base of this app as this PR created a `package-lock.json`.

PR #2178, opened 2017-12-19
Cache all node_modules on CI

lgeiger:
CI has a hard time with our monorepo since we're note using hoisting or yarn workspaces at the moment.

Let's try a more aggressive caching strategy to hopefully speed things up.


rgbkrk (approved):
That’s a start I guess.

PR #2204, opened 2017-12-19
refactor lerna packages

willingc:


package.json
```diff
@@ -21,17 +21,19 @@
...
     "postinstall": "npm run lerna",
-    "app:desktop": "lerna run start --scope nteract --stream",
+    "//lerna bootstrap": "builds packages",
+    "lerna": "lerna bootstrap",
+    "build:apps": "lerna run build --scope nteract --scope @nteract/play --scope @nteract/commuter --scope @nteract/notebook-on-next --scope @nteract/showcase",
+    "app:desktop": "lerna build --scope nteract --stream",
     "app:play": "lerna run dev --scope @nteract/play --stream",
     "app:showcase": "lerna run dev --scope @nteract/showcase --stream",
-    "app:commuter": "lerna run dev --scope @nteract/commuter --stream",
+    "app:commuter": "lerna build --scope @nteract/commuter --stream",
```
rgbkrk: We still need `app:commuter` to do the `lerna run dev` command, as it should run the app locally for us.

applications/showcase/package.json
```diff
@@ -4,7 +4,9 @@
   "description": "showcase of nteract components",
   "main": "index.js",
   "scripts": {
-    "dev": "next"
+    "dev": "next",
+    "build": "next build",
+    "start": "next start"
```
rgbkrk, line 9: 👍

package.json
```diff
@@ -21,6 +21,9 @@
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
     "postinstall": "npm run lerna",
+    "//lerna bootstrap": "builds packages",
+    "lerna": "lerna bootstrap",
+    "build:apps": "lerna run build --scope nteract --scope @nteract/play --scope @nteract/commuter --scope @nteract/notebook-on-next --scope @nteract/showcase",
```
mpacer: I don't think you need to have --scope repeated each time.

package.json
```diff
@@ -21,6 +21,9 @@
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
     "postinstall": "npm run lerna",
+    "//lerna bootstrap": "builds packages",
```
rgbkrk, line 24: The `lerna bootstrap` symlinks and installs all packages.

package.json
rgbkrk: Oh really? I didn't realize that.

PR #2211, opened 2017-12-19
Disable uglify on next.js apps

lgeiger:
Disable uglify on next.js apps until we find out why it's failing. This should fix our build issues for now.

Probably some dependency is not correctly transpiled.


rgbkrk:
It's mostly that uglify does not handle ES6+, I believe most folks have been switching to uglify-es and we'd need next to do the same for us.

PR #2218, opened 2017-12-20
Use next-gen components for notebook

captainsafia:
* [x] Move Cell creation into the `<NotebookApp />`
* [x] Remove old Cell tests


rgbkrk:
This is looking great! We can probably extract the pinned cell and other component to nextgen or otherwise here.

rgbkrk:
I forget, was this superseded by another PR @lgeiger & @captainsafia ?

captainsafia:
Nope! It's WIP still. I think the thing that was holding us up was wanting to waiting for your new DropDown component so we could use that....

rgbkrk:
I’m so close on the drop down! After I’m done Ice skating with the kiddos I’ll finish it up.

rgbkrk:
Oooooook the dropdown menu, and by extension the toolbar, is ready to go so you can rebase and charge ahead here.

captainsafia:
@rgbkrk: Thanks! Let me update this.

rgbkrk:
I've taken over this branch while @captainsafia is doing the valiant effort of upgrading our immutable to 4.0.0-rc9 and the latest in flow technology.

Some initial TODOs / steps ahead to take note of where I think this needs to go:

* [x] Create this current notebook as a `provider` component
* [ ] See if there is a "view" level of component we can have as our base component
* [x] Move StickyCellContainer and PinnedCellPlaceholder to their own component files

I'll keep adding to this list (and moving it to the top of this PR).

rgbkrk:
Ok this is _mostly_ in order here. Seems to work well with desktop, I'll do some more checking of behavior here and review this thoroughly. Next up is tests. I'll add on to the PR description for the TODO.

rgbkrk:
Finished up each of those, now for the exercising of the app in case this touched parts that it shouldn't have.

rgbkrk:
Lookin' good on the apps side!

rgbkrk:
:shipit:

PR #2226, opened 2017-12-21
Only require codemirror from @nteract/editor

lgeiger:
Closes #2221 and updates venorized css.

### DIFF
--- a/packages/editor/src/vendored/codemirror.js
+++ b/packages/editor/src/vendored/codemirror.js
@@ -83,7 +83,6 @@ export default css`
     animation: blink 1.06s steps(1) infinite;
     background-color: #7e7;
   }
-
   @-moz-keyframes blink {
     0% {
     }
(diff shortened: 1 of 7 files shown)


packages/editor/src/vendored/codemirror.js
```diff
@@ -471,6 +470,15 @@ export default css`
     padding-right: 0.1px;
   }
 
+  /**
+  style jsx doens't work well with media queries
+  @media print {
```
rgbkrk, line 475: Ah yeah I think I just deleted the print media queries for our use here when I first vendorized.

packages/notebook-preview/package.json
```diff
@@ -23,7 +23,6 @@
     "@nteract/display-area": "^3.1.0",
     "@nteract/editor": "^4.2.2",
     "@nteract/transforms": "^3.0.3",
-    "codemirror": "^5.28.0",
```
rgbkrk, line 26: 👍

packages/editor/src/vendored/codemirror.js
```diff
@@ -236,10 +235,10 @@ export default css`
   /* Default styles for common addons */
 
   :global(div.CodeMirror span.CodeMirror-matchingbracket) {
-    color: #0f0;
+    color: #0b0;
```
rgbkrk, line 238: I like that they lightened this a bit. 😄

PR #2227, opened 2017-12-21
Remove packages/enchannel since it's not used anywhere

lgeiger:


rgbkrk:
🎉 

One thing we could/should do with enchannel is move parts of the README to a top level docs directory. _However_, it would need to match the two current enchannel implementations -- the one in rx-jupyter and the one in enchannel-zmq-backend. We could/should switch `enchannel-zmq-backend` to be named `@nteract/zmq-backend` (or similar).

PR #2228, opened 2017-12-21
[DONT MERGE] Showcase at root

rgbkrk:
This starts moving showcase to the root to make it simpler to resolve straight from `src` for next.js. _However_, this also makes it so the packages' `node_modules` aren't resolved. I'm looking into aliasing with webpack next.


rgbkrk:
Ok this now works for the webpack build yet not the server side bit because node doesn't resolve it. _However_, using [babel-plugin-module-resolve](https://github.com/tleunen/babel-plugin-module-resolver) is completely an option and matches up pretty well:

```
{
  "plugins": [
    ["module-resolver", {
      "alias": {
        "^@nteract/(.+)": "packages/\\1"
      }
    }]
  ]
}
```

rgbkrk:
At the _very least_ I've figured out a great way to handle aliasing styled-jsx so that there's no initial flicker. Also, I think we can probably keep the old showcase now and have it resolve the appropriate `src` directories too. 😄

rgbkrk:
Probably the best PR to watch for current ease with lerna: https://github.com/zeit/next.js/pull/3319 (it's a PR for transpiling certain modules as part of the build)

rgbkrk:
I'm going to close this and do over small parts of this as separate PRs.

PR #2236, opened 2017-12-22
Improve play

theengineear:
This change moves us from using `setState` to using `redux` / `redux-observable` to handle our state management in the `play` application.

TODO:

- [x] Clear outputs each time source is run
- [x] Finish flow typing

For another PR:

* Tie source to UI state rather than kernel #2243
* Ability to change servers #2244 
* Test suites #2245
* Add functionality to interrupt and kill kernels #2246


theengineear:
@mpacer ^^ there's a short list of open issues in the description for this. Want to (a) add anything that you think might need adding to that list and (b) pick one you'd like to start out on?

@rgbkrk anything else you'd like to add to the list?

Feel free to just edit the description.

I think the first check-box up there might warrant some more discussion. I think you two have more knowledge around it than I do, but... I'm sort-of thinking that it'd be a *good* user experience to have the source/outputs linked directly to each active kernel and then just display that information given the currently-selected kernel.

mpacer:
> Do we want source/outputs in ui state? Or, should they belong per-kernel?

We had talked about not requiring a 1 kernel at a time model for everything… I'm not sure if my intuition is that you'd keep the code run in them around as well. 
My intuition is that we should start the kernels clean and therefore start the source clean as well.

mpacer:
Clear outputs each time source is run — that would seem to be done in an epic…how would we ensure that it is cleared before the output messages are received and not after (which would be a fairly unpleasant UX)?

rgbkrk:
> Clear outputs each time source is run — that would seem to be done in an epic…how would we ensure that it is cleared before the output messages are received and not after (which would be a fairly unpleasant UX)?

This seems like we'd have to do something "cell" like, which is tracking the originating message id to a collection of outputs.

theengineear:
>Clear outputs each time source is run...

@mpacer i think that you'd want to do a couple things:

1. you'll now want to return _mulltiple_ actions from the `runSourceEpic`, so you'll need to change the `map` to `mergeMap` so that you can return an observable on this line https://github.com/nteract/nteract/blob/improve-play/applications/play/redux/epics.js#L192

2. I think if you add a new `CLEAR_KERNEL_OUTPUT` (S?) actionType and a related action, you can basically return `Observable.of(actions.clearKernelOutput({ serverId, kernelName }), actions.setSource(source));`

Is that what you're asking? I just saw @rgbkrk 's comment and now I'm not sure if I understood correctly.

rgbkrk:
> My intuition is that we should start the kernels clean and therefore start the source clean as well.

I agree with this in spirit. Here's a user workflow that flies in the face of this:

```
User is writing python code, connected with a python2 kernel
They realize they meant to connect to python3 (and a different binder)
They expect to be able to finish what they were working on (and share it)
```

theengineear:
>My intuition is that we should start the kernels clean and therefore start the source clean as well.

Yah, this is my intuition as well. Personally, I think we should move `source` from `ui.source` to `serversById[id].server.activeKernelsByName[kernelName].kernel.source`. I think Kyle had some interesting counter-arguments to this point though.

mpacer:
> Is that what you're asking? I just saw @rgbkrk 's comment and now I'm not sure if I understood correctly.

The order of events matters (if I submit code & it comes back quickly and it only then clears the outputs we'll have a nondeterministic UI).  

I'm asking how we can ensure that we don't clear output from the most recent submitted source.

theengineear:
>The order of events matters (if I submit code & it comes back quickly and it only then clears the outputs we'll have a nondeterministic UI).

When you click `run code`, are we expecting that:

1. the output is immediately cleared
2. when the kernel responds, the output is populated?

If that's the case, since synchronous things will remain synchronous, there's no way the kernel will beat us to it. Do I have that right @rgbkrk

rgbkrk:
The kernel can respond with more than one output message (which is currently handled properly here).

rgbkrk:
The output should be immediately cleared after running code.

rgbkrk:
The kernel can respond with more than one output message (which is currently handled properly here). This is inherently asynchronous (and subject to the halting problem 😉 ).

rgbkrk:
The easiest way to do things "too quickly" is to press execute code really fast (at least with the old UI). Old outputs will come in slower than the clear outputs action does.

theengineear:
ah, i think i see the issue:

1. click run source
2. kernel response with output 0-0
3. click run source
4. clear output 0-0
5. oops, kernel responds with output 0-1
6. kernel responds with output 1-0
7. kernel responds with output 1-1

the user sees old output in the outputs panel

rgbkrk:
I'm totally reading this thread upside down so bear with me here. 😄 

> anything else you'd like to add to the list?

I think it's ok to table a few of these, in particular:

* Add functionality to interrupt kernels
* Add functionality to kill kernels

Saving them for another PR

theengineear:
well, one solution to that is to revise how the ui works. we could (instead of clearing output), assign an id each time we submit source? is there a way to map a kernel output to an input @rgbkrk ? then the `kernel` state should keep an `inputId` hunk of state as well, right?

mpacer:
If we give each source that is submitted an id, then mapping outputs to input should be straightforward, but would be more state to manage inside the kernel.

> is there a way to map a kernel output to an input @rgbkrk ? then the kernel state should keep an inputId hunk of state as well, right?

yes, but we can just create the inputId as a uuid on the fly, i think?

rgbkrk:
When messages come in, they have two headers to pay attention to: `header` and `parent_header`. Within each there's a `msg_id`. For cells in a notebook (for reference), we effectively associate a cell id and their last parent message (for an `execute_request`).

rgbkrk:
We created one nice little operator for this in `@nteract/messaging` too -- `childOf`.

```
channel.pipe(
  // filters all those that have parent_header.msg_id === req.header.msg_id
  childOf(req)  
)
```

rgbkrk:
However, I think with this renewed epic setup and messages that pump through the central store -- I think you could just group messages that have the same `parent_header.msg_id` in the store. The outputs that show are those that go with the last `execute_request`'s `msg_id`.

mpacer:
am i correct that RUN_SOURCE doesn’t trigger any reducers currently? In which case creating a reducer that clears outputs would be the right way to go? (since reducers are always run before epics?)

theengineear:
>RUN_SOURCE doesn’t trigger any reducers currently?

correct, it's just handled in an epic.

>In which case creating a reducer that clears outputs would be the right way to go?

yup, and that was my first suggestion. however, you could also just go ahead and have the `activeKernelOutputs` handle the `RUN_SOURCE` type by just returning `[]` empty.

In general, we shouldn't be afraid of reducers listening to multiple action types. Totally OK to do that.

mpacer:
> yup, and that was my first suggestion. however, you could also just go ahead and have the activeKernelOutputs handle the RUN_SOURCE type by just clearing itself.

So I think this won't deal with the delayed messages coming back from a different execution request. which is what kyle was pointing toward above.

mpacer:
Would it be cleaner to have a `CLEAR_OUTPUTS` actionType in case other parts of the code eventually need to do this as well?

theengineear:
>Would it be cleaner to have a CLEAR_OUTPUTS actionType in case other parts of the code eventually need to do this as well?

Yah, I guess that is clearer (no pun intended). It _is_ the case here that we can just clear on `RUN_SOURCE` though. I'm fine either way. Whichever feels better to you. I can imagine other parts of the code wanting to clear the output, so I guess I'm more in favor of the `CLEAR_KERNEL_OUTPUTS` approach :+1:

rgbkrk:
One excellent reason to have a `CLEAR_KERNEL_OUTPUTS` action is that the Jupyter kernels themselves can send a message called `clear_output`. You can do this via IPython with `IPython.display.clear_output()`.

rgbkrk:
This whole discussion is making me think that the reducer from inside `@nteract/core` for document could be made to be a reducer for cells without a notebook (or we just manage this as a single cell notebook).

Almost always in most of the reducers they navigate through the notebook document state into the cells by ID, like this:

```js
state.updateIn(
      ["notebook", "cellMap", cellID, "outputs"],
      (outputs: ImmutableOutputs): ImmutableOutputs =>
        reduceOutputs(outputs, output)
);
```

When I think we could make a reducer/action pair that would use the state of just a single cell (so a reducer per cell in the notebook app) and return the state of that single cell. The update display stuff makes this a bit more complicated, we can figure it out though.

theengineear:
^^ @rgbkrk , not sure i can 100% 👍 / 👎  your last suggestion, but if you want to try doing that in a subsequent PR, I'm certainly for it 😸

theengineear:
Re: keeping source/outputs alongside the kernel state vs alone in the ui...

```
User is writing python code, connected with a python2 kernel
They realize they meant to connect to python3 (and a different binder)
They expect to be able to finish what they were working on (and share it)
```

^^ CMD+C / CMD+P

However, I do have another compelling argument for keeping it alone in the UI. It's how notebooks currently work, right? If this was just a single cell in a notebook, we wouldn't change the cell/outputs when the user changes the kernel right?

Ha, I'm a little torn on this one. I feel like keeping source/outputs per kernel makes the most sense. BUT, I'm more than willing to believe that I'm a newcomer to this ecosystem and my opinions may be naive.

theengineear:
@rgbkrk , it's the convention here to *not* commit `package-lock.json` files, yah? should we remove this file from the PR?

rgbkrk:
Yeah we don’t keep the package lock files in the repo. They’d probably be fine within application directories, it’s better for us not to at the moment.

rgbkrk:
I'm going to add one commit fixing up the flow and I'll comment inline as to why I went with particular changes.

applications/play/redux/actions.js
```diff
@@ -42,14 +42,14 @@ export const activateServer = (payload: {
 });
 export const activateServerFulfilled = (payload: {
   serverId: string,
-  config: object
+  config: *
```
rgbkrk: The correct flow type for an Object is `Object`. Instead of changing it to that though, I made these inferred types. When we use Object, it's basically a catch all for any object with any properties. Using the `*` is also a catchall, however it will be informed based on our usage. We can still make a more explicit type here too. Not particularly pressing though.

applications/play/redux/epics.js
```diff
@@ -2,7 +2,6 @@
 import * as actionTypes from "./actionTypes";
 import * as actions from "./actions";
 
-// $FlowFixMe: ofType should now be a lettable operator and is used on master
```
rgbkrk: Since we rebased (hurrah :tada:!), we can rely on the proper flow type for `ofType` here.

applications/play/components/Main.js
```diff
@@ -130,9 +131,39 @@ class Main extends React.Component {
...
+            // TODO: This property needs to be excised from the editor, as it belongs
+            //       in codemirror options below
+            cursorBlinkRate={0}
+            // TODO: This is the notebook implementation leaking into the editor
+            //       component. It shouldn't be here, I won't refactor it as part
+            //       of the current play PR though.
+            id="not-really-a-cell"
+            onFocusChange={() => {}}
+            focusAbove={() => {}}
+            focusBelow={() => {}}
```
rgbkrk, line 154: I didn't dig far into these, as I was primarily just making sure that the types were all set here. The `editor` implementation needs to have more defaultProps set, as well as allowing these functions to not have to be passed. For now, I made it work as is and didn't muck with editor (setting up defaultProps and figuring out how these should be used should be a different PR).

applications/play/components/Main.js
```diff
@@ -130,9 +131,39 @@ class Main extends React.Component {
...
         <div className="play-editor">
           <CodeMirrorEditor
+            // TODO: these should have defaultProps on the codemirror editor
+            cellFocused
+            editorFocused
+            channels={
+              currentKernel && currentKernel.channel
+                ? currentKernel.channel
+                : null
+            }
```
rgbkrk, line 141: Since our `channels` format is now the same between desktop and web apps, enabling completion is a "simple" matter of passing the `currentKernel.channel` to the editor and setting `ctrl-space` to autocomplete.

rgbkrk:
Based on where this is at and how we need to keep evolving inner packages, my personal opinion is that we should merge this in and leave the TODOs as new issues to tackle after.

mpacer:
So I’ve been thinking about it and I think on the similarity to notebook grounds we should keep the source in the ui independent of the kernel. But that we shouldn’t alter the source currently in the cell when switching kernels to address the “oh I meant python 3” case since it then would be a matter of switching kernels and that wouldn’t change what they’ve written.

Re:merging, What more needs to be addressed testing-wise? In terms of new tests and shouldn’t we make the current tests pass before merging?

rgbkrk:
Yeah, I would aim for "make things pass", then we merge and move forward. There are a few improvements related to packages (namely `@nteract/editor`) that I want to do for the sake of all the apps, and I'd like to keep them separate from this one.

rgbkrk:
I'm definitely in agreement with @mpacer that we should keep the source code in the ui independent of the kernel.

rgbkrk:
All the tests pass locally for me, we're still running into issues with timeouts with jest on Travis and we're unsure as to why. As long as folks are ok with going ahead with this, I'd like to merge it.

rgbkrk:
I'm going to merge this now and create new issues for each of the todos with a play label + title tag.

applications/play/components/Main.js
mpacer: Did you make an issue describing more about this?

rgbkrk: Good point, I’ll make an issue.

applications/play/redux/actions.js
mpacer: Ah, I didn't know the role of `*` ( the [existential type](https://flow.org/en/docs/types/utilities/#toc-existential-type)) thank you for pointing that out. I put the link here for future reference of anyone stumbling across this note.

applications/play/components/Main.js
mpacer: As in, we get a feature "for free" now? Or so do we need to do something more?

rgbkrk: Free!

Try hitting `ctrl-space` inside the editor on play (when it makes sense for an autocomplete).

PR #2237, opened 2017-12-23
add output clearing action, type & reducer; add to runSourceEpic

mpacer:
@theengineear this clears the outputs on each RUN_SOURCE call.


applications/play/redux/epics.js
```diff
@@ -200,11 +200,15 @@ const runSourceEpic = (action$, store) =>
...
+      const actionsArray = [
+        actions.clearKernelOutputs({ serverId, kernelName })
+      ];
       const channel = objectPath.get(store.getState(), channelPath);
       if (channel) {
         channel.next(executeRequest(source));
       }
-      return actions.setSource(source);
+      actionsArray.push(actions.setSource(source));
+      return of(...actionsArray);
```
theengineear: Since we're always returning both actions from this, it might be clearer to just have:

```
return of(
    actions.clearKernelOutputs({ serverId, kernelName }),
    actions.setSource(source)
);
```

theengineear (approved):
Just one 🐄 (tip), but otherwise --> looks good!

applications/play/redux/epics.js
theengineear: ^^ oops, edited, i had some bad copy pasta in there

theengineear: :+1: cool, thanks for updating that!

PR #2238, opened 2017-12-23
add types to action payloads so that they don't need to be inferred

mpacer:
I might have some of these slightly off but I think I got most of them right.


mpacer:
@theengineear would love to have your review.

rgbkrk:
Since this is going to the branch, I'll just go ahead and merge it and review the PR in whole.

PR #2239, opened 2017-12-23
Create @nteract/types package

captainsafia:
In progress...


packages/core/src/dummy/index.js
```diff
@@ -23,7 +23,7 @@ const rootReducer = combineReducers({
   config
 });
 
-import type { AppState } from "@nteract/types/records";
+import type { AppState } from "@nteract/types/core/records";
```
rgbkrk: Ah yeah, we're going to have an interesting set of choices to make for the app records since the desktop and web versions will probably need different app states. Something to figure out in the future though. I really like the `entities` approach that is on the current play PR. The sprints were really good for shaking out some new approaches to our overall app handling. 😄

packages/messaging/src/index.js
```diff
@@ -25,7 +25,7 @@ import {
 
 import { message, executeRequest } from "./messages";
 
-import type { JupyterMessage, ExecuteRequest } from "@nteract/types";
+import type { JupyterMessage, ExecuteRequest } from "@nteract/types/messaging";
```
rgbkrk: 😄

captainsafia:
This is ready for a final look through and merge.

Ditto your commend on the `core/records` thing. Making a `@nteract/types/core/web/records` and a `@nteract/types/core/desktop/records` and a `@nteract/types/core/shared/records` seems a little verbose but maybe appropriate?

packages/types/src/core/records.js
```diff
@@ -0,0 +1,180 @@
...
+  nbformat: 4,
+  nbformat_minor: 0 | 1 | 2 | 3 | 4,
+  metadata: NotebookMetadata
+};
+
+// Parts of AppRecord should become
+// ElectronAppRecord
+// Basically, anything that's only for desktop should have its own record & reducers
+
+export const AppRecord = Immutable.Record({
```
mpacer, line 124: why are these exported as `const`s and not as `type`s?

rgbkrk: Oh right, we may need these to be separate from here, kept in core.

rgbkrk: At some point, when we're able to get on Immutable 4.0.0-RC*, we'll be able to [use the `RecordOf<T>` type](https://github.com/facebook/immutable-js/issues/1209#issuecomment-334041276).

captainsafia: I moved them here in anticipation of the upgrade to the 4.X.X stream. With the `RecordOf<T>` I think we'll be exporting a Record factory out of our `@nteract/types` package.

packages/commutable/src/index.js
```diff
@@ -3,7 +3,7 @@
 import type { Notebook as v4Notebook } from "./v4";
 import type { Notebook as v3Notebook } from "./v3";
 
-import type { ImmutableNotebook, JSONType } from "./types";
+import type { ImmutableNotebook, JSONType } from "@nteract/types/commutable";
```
rgbkrk, line 6: Since it's within the types directory, perhaps we should call this `@nteract/types/notebook` or `@nteract/types/immutableNotebook`? I'm not quite sure. 🤷‍♂️

packages/editor/src/index.js
```diff
@@ -18,7 +18,12 @@ import { tool } from "./jupyter/tooltip";
 
 import { debounce, merge } from "lodash";
 
-import type { EditorChange, ScrollInfo, CMI, CMDoc } from "./types";
+import type {
+  EditorChange,
+  ScrollInfo,
+  CMI,
+  CMDoc
+} from "@nteract/types/editor";
```
rgbkrk: Unless we're using these elsewhere (by more than one package, other than using the editor), I don't think these have to be extracted from inside of the editor package as they're really just here for codemirror usage.

packages/types/src/messaging.js
```diff
@@ -0,0 +1,29 @@
...
+  store_history: boolean,
+  user_expressions: Object,
+  allow_stdin: boolean,
+  stop_on_error: boolean
+};
+
+export type ExecuteRequest = JupyterMessage<
+  "execute_request",
+  ExecuteMessageContent
+>;
```
rgbkrk, line 29: 👏

rgbkrk:
I'm personally in favor of merging these in. If we need/want to move the record types up somewhere else later, that's totally ok. @captainsafia knows best as the one who has been working with the immutable 4.x upgrade.

packages/editor/src/index.js
captainsafia: Good point! Let me revert this really quick.

packages/commutable/src/index.js
captainsafia: I dunno how I feel about changing the name of the types to not match up with the name of the package they are used...

Then again, calling it `notebook` does make it more friendly for external devs.

packages/editor/src/index.js
rgbkrk: Cool, works for me!

packages/commutable/src/index.js
rgbkrk: Yeah my purpose was getting away from cryptic names since we use an @nteract namespace now. Back when commutable was created we didn’t have much for options other than creating crafty names. Not that I would name it this but if we went by purpose we’d call it immutable-model-of-an-in-memory-notebook-and-related-data-structures

PR #2251, opened 2017-12-26
allow kernel_info_request to change codemirror highlighting language

### LINKED ISSUE #2250: [play] editor's highlighting language should match kernel language after switching
Right now we don't have the language being extracted and propagated up to the CodeMirrorEditor component; it's just hardcoded: 
https://github.com/nteract/nteract/blob/cef76af981463d34758abec82c8cc913d444a4d4/applications/play/components/Main.js#L172

If I'm thinking about this correctly, we probably want to store this in the `currentKernel` entity.

I'm not sure what pattern is best to accomplish this though (since the relevant info is deeply nested inside `currentServer.kernelSpecs.kernelSpecByKernelName.[currentKernelName].spec.language`)… I suppose I can look at the pattern used for `currentKernelName`?

mpacer:
closes #2250


applications/play/redux/epics.js
```diff
@@ -126,6 +126,12 @@ const initializeKernelMessaging = (action$, store) =>
                 })
               );
               break;
+            case "kernel_info_reply":
+              actionsArray.push(
+                actions.setCurrentKernelLanguage(
+                  message.content.language_info.name
```
rgbkrk: I think it would be ok to surface the overall language info, cached on the kernel object located on that `entities.serversById[serverId].server.activeKernelsByName[kernelName].kernel` path.

applications/play/redux/reducer.js
```diff
@@ -82,6 +82,17 @@ const currentKernelName = (state = "", action) => {
   }
 };
 
+const currentKernelLanguage = (state = "", action) => {
+  switch (action.type) {
+    case actionTypes.SET_CURRENT_KERNEL_LANGUAGE: {
+      return action.payload;
```
rgbkrk: It just dawned on me that this was doing a `combineReducers` for each and every one of these so that their name exactly matches what they are on the state tree here and that they only have the local `currentKernelLanguage` state.

applications/play/redux/epics.js
mpacer: Are you saying I should create a new `language_info` field inside the kernel object created in `activeKernelsByName`?

rgbkrk: Yeah, as long as that makes sense.

mpacer: I don't really understand what it would mean to "make sense" or "not make sense" in this case… I'm also not sure if you were saying to do this in lieu of adding a top level ui field (which I only am guessing based on https://github.com/nteract/nteract/issues/2250#issuecomment-354018740) or in addition to it. I think I'm going to put this down for the day and I'll come back to it later as I'm not making much progress.

applications/play/redux/reducer.js
mpacer: I think I follow but I'm not sure if/how this is supposed to affect how I've written this.

rgbkrk: This is more of a comment on me realizing how the architecture was setup here by you and @theengineear -- it's super clean. There's a 1:1 mapping between the function names and the state it's resolving.

rgbkrk:
I would change this to using an action creator that takes:

* the server ID
* the kernelName
* the language info

Which would mean that when the `kernel_info_reply` message comes through, the action is formed like this:

```js
actionsArray.push(
  actions.setLanguageInfoForKernel(serverId, kernelName, message.content.language_info)
)
```

I'm not sure how the reducers would need to be situated to make use of this though, I'd have to understand the `utils.createObjectReducer` used here.

mpacer:
> ```js
> actionsArray.push(
>   actions.setLanguageInfoForKernel(serverId, kernelName, message.content.language_info)
> )
> ```

Ah! That is beautifully clear! Thank you for addressing all of my questions so clearly!

I'm going to keep the current mechanism because it's a simpler api for someone to implement in another part of the code (in case that is ever needed). Rather than creating an entire language_info if they only want to use that to set the  `SET_CURRENT_KERNEL_LANGUAGE` action (e.g., if the kernel_info doesn't match what the person was looking for & they want to change it). 

But I will implement another for the `mode` as well (I see why you linked to that now), since that is a field inside `kernel_info`.

mpacer:
Realised that the `currentKernelLanguage` should be `highlightedLanguage` due to its different purpose while I was trying to figure out how to do the `languageInfo` bit which has turned out to be substantially more tricky than I expected.

rgbkrk:
Okee dokee. Want me to merge this as is then or wait for the `languageInfo` bits?

mpacer:
Ok I think these should work now, including setting the Codemirror mode info (at least as retrieved from the kernel).

mpacer:
I should note that to actually use that CodeMirror mode info would require a more substantial restructuring of the way that the current editor handles setting the mode, it is currently works through setting the language string: 
https://github.com/nteract/nteract/blob/82611f8e92949da8e145ecb38b19b6e828e13a1c/packages/editor/src/index.js#L98
https://github.com/nteract/nteract/blob/82611f8e92949da8e145ecb38b19b6e828e13a1c/packages/editor/src/index.js#L230-L233

I'll try to include a refactor for there, but I'm somewhat unsure about how to handle that most safely without disrupting other parts of the codebase that are using only the language prop.

mpacer:
Not actually how to debug whether the way that codemirror is being set is appropriate because it's not a matter of setting the react prop (which I know works) but rather how codemirror is working on the js side.

packages/editor/src/index.js
```diff
@@ -227,9 +228,13 @@ class CodeMirrorEditor extends React.Component<
...
     }
 
-    if (prevProps.language !== this.props.language) {
+    if (prevProps.language !== this.props.language && !prevProps.mode) {
       this.cm.setOption("mode", this.props.language);
     }
+
+    if (prevProps.mode !== this.props.mode) {
+      this.cm.setOption("mode", this.props.mode);
+    }
```
rgbkrk, line 237: Fantastic, this is a good first pass at closing that codemirror issue about using the mode.

applications/play/redux/reducer.js
```diff
@@ -214,14 +225,24 @@ const activeKernelMessages = (state = [], action) => {
...
 const activeKernel = combineReducers({
   name: activeKernelName,
   id: activeKernelId,
   lastActivity: activeKernelLastActivity,
   channel: activeKernelChannel,
   outputs: activeKernelOutputs,
   status: activeKernelStatus,
-  messages: activeKernelMessages
+  messages: activeKernelMessages,
+  languageInfo: activeKernelLanguageInfo
```
rgbkrk, line 245: superb

applications/play/components/Main.js
```diff
@@ -321,6 +327,7 @@ const mapStateToProps = state => ({
   showPanel: state.ui.showPanel,
   currentServerId: state.ui.currentServerId,
   currentKernelName: state.ui.currentKernelName,
+  highlightedLanguage: state.ui.highlightedLanguage,
```
rgbkrk: Since you now have languageInfo in the tree, you could switch this function up a bit and trim out the highlighted language action.

```js
const mapStateToProps = state => {
  const currentKernel = objectPath.get(state, [
    'entities',
    'serversById',
    state.ui.currentServerId,
    'server',
    'activeKernelsByName',
    state.ui.currentKernelName,
    'kernel',
  ]);

  const highlightedLanguage = objectPath.get(currentKernel, [
    'languageInfo',
    'name',
  ]);

  return {
    repo: state.ui.repo,
    gitref: state.ui.gitref,
    source: state.ui.source,
    platform: state.ui.platform,
    showPanel: state.ui.showPanel,
    currentServerId: state.ui.currentServerId,
    currentKernelName: state.ui.currentKernelName,
    highlightedLanguage,
    currentKernel,
    currentServer: objectPath.get(state, [
      'entities',
      'serversById',
      state.ui.currentServerId,
      'server',
    ]),
  };
};
```

Then you can do away with the action set up for the highlighted language.

mpacer: I specifically was leaving the action to set up the highlighted language because we might want that in the ui without actually changing anything about the kernel (). It's a simpler API for doing that kind of setting. Maybe that's not a useful feature (& I don't have a ui element for it yet), but I feel like I had a desire for that feature on a couple occasions. 

*NB*: I was also considering a separate epic to handle the setting as a consequence of setting the language info, but that seemed like a bit too much in that direction.

rgbkrk: The main reason we don't want to track the same state in two different sections of the state tree is that they can become out of sync. We want one source of truth, otherwise all developers have to know that a change in that field has to also effect it in another location.

rgbkrk:
By the way, I'm not holding this PR back because of the single state comment above, there's a typing issue with the Editor and the current code here:

```
Error: applications/play/components/Main.js:134
               v----------------
134:           <CodeMirrorEditor
135:             // TODO: these should have defaultProps on the codemirror editor
136:             cellFocused
...:
180:           />
               -^ React element `CodeMirrorEditor`
159:                 : null
                       ^^^^ null. This type is incompatible with
 46:   mode?: string | Object,
              ^^^^^^^^^^^^^^^ union: string | object type. See: packages/editor/src/index.js:46
  Member 1:
   46:   mode?: string | Object,
                ^^^^^^ string. See: packages/editor/src/index.js:46
  Error:
  159:                 : null
                         ^^^^ null. This type is incompatible with
   46:   mode?: string | Object,
                ^^^^^^ string. See: packages/editor/src/index.js:46
  Member 2:
   46:   mode?: string | Object,
                         ^^^^^^ object type. See: packages/editor/src/index.js:46
  Error:
  159:                 : null
                         ^^^^ null. This type is incompatible with
   46:   mode?: string | Object,
                         ^^^^^^ object type. See: packages/editor/src/index.js:46
```

applications/play/components/Main.js
mpacer: I agree about having a single source of truth, I just want that truth to be overwritable without needing to change the kernel. I guess I'm thinking of the highlighting mode as something with a default determined by the kernel but with that default being potentially overwritten. I'll give this yet another refactor with this in mind to treat the mode as the single source of truth for the play component.

rgbkrk:
:rotating_light: I rebased this branch :rotating_light:

rgbkrk:
Ok, flow should be fixed. Time to let Travis do its thing.

rgbkrk:
Passes! Merging.

PR #2256, opened 2017-12-27
ensure sockets are connected before usage

### LINKED ISSUE #2156: Kernel sometimes isn't really starting
With the desktop app, on `master`, it seems that sometimes the app acts like its connected to the kernel even though it isn't responding to messages. It _seems_ like it's a timing issue of the kernel not quite being ready. Ever since we optimized things in #2137 I've noticed this behavior. That leads me to believe that there's a readiness issue that before was a non-issue due to timing.

One other thing I noticed while doing #2137 is that it seems like registering callbacks (other than `message`) on the jmp sockets is not giving us access to the zeromq events.

/cc @captainsafia @lgeiger

rgbkrk:
closes #2156 (thankfully)

I've made the interface to `enchannel-zmq-backend`'s channel creation to be a promise that resolves when all the sockets have been properly connected. This does use the monitoring that was suggested by @lgeiger.

With this, I think we've finished off the last release blocker (that I can recall).

### DIFF
--- a/applications/desktop/__mocks__/enchannel-zmq-backend.js
+++ b/applications/desktop/__mocks__/enchannel-zmq-backend.js
@@ -12,7 +12,7 @@ class HokeySocket extends EventEmitter {
 }
 
 module.exports = {
-  createMainChannel: () => {
+  createMainChannel: async function() {
     const shellSocket = new HokeySocket();
     const iopubSocket = new HokeySocket();
     const sockets = {
@@ -20,7 +20,7 @@ module.exports = {
       iopub: iopubSocket
     };
 
-    const channels = createMainChannelFromSockets(sockets, {
+    const channels = await createMainChannelFromSockets(sockets, {
       session: "spinning",
       username: "dj"
     });
--- a/applications/desktop/__mocks__/spawnteract.js
+++ b/applications/desktop/__mocks__/spawnteract.js
@@ -11,7 +11,11 @@ module.exports = {
     }
     return writeConnectionFile("config").then(c => {
       return {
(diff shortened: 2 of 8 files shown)


rgbkrk:
Working on fixing up the tests and adding new ones.

rgbkrk:
This will soon be good to go and we'll have a much more stable app as a result. 🎉

rgbkrk:
Readyyyyy

rgbkrk:
Luck of the draw, I've still gotten the app to boot with a kernel not really quite attached / ready.

I'll debug now.

rgbkrk:
Ok, things are more stable now. If I go after anything similar again, I think I'll bring the message encoding and decoding in from jmp, handling it inline with enchannel-zmq while having more control / insight of what's going on with zeromq.

rgbkrk:
@captainsafia check this out when you get a chance.

rgbkrk:
Now I fully understand how I fixed then broke this. The monitoring did patch around the sockets to make sure they're ready. Where I made a fatal mistake was when I did a refactor I connected to zeromq twice with the same identity (two socket connections). This completely broke all the `ROUTER` sockets (`shell` showing the most glaring of symptoms). After noticing this and cleaning it up, the sockets connect properly and stay connected.

All that's to say -- 🎉  this is now working and stable and I learned a bit about how to break zeromq connections.

Madhu94:
@rgbkrk @captainsafia Thank  you so much! My build works again!

rgbkrk:
Sweet! We'll get a new release out soon.

PR #2257, opened 2017-12-28
reduce what we do in travis

rgbkrk:
Just looking to see if this is cruft we can toss.


rgbkrk:
It is!

rgbkrk:
Now saving you a whopping 20 seconds.

PR #2258, opened 2017-12-28
Upgrade immutable and flow!

captainsafia:
* [x] Upgrade Immutable
* [x] Upgrade Flow
* [x] Fix tests
* [x] Upgrade library definitions
* [x] Apply `$FlowFixMe`s in all the corners that are difficult yet working


rgbkrk:
Looking forward to this!

captainsafia:
OK. I got this down to 5 errors with some `$FlowFixMe`s. The remaining errors are related to types on the `CellCreator`. I avoided touching this for now as #2263 is open. The other error is a result of some of the typings in the RxJS library (I think that can be fixed with a PR that @rgbkrk  sent me a link to.)

rgbkrk:
Woohoo!

Would you want us to bring in #2263 first then or do you want me to tackle the last few types on `CellCreator`?

captainsafia:
I think it makes sense to bring #2263 and see if that fixes in anything...

rgbkrk:
Ok I'll merge it in and then `git rebase -i origin/master` this branch.

rgbkrk:
Rebased, time to see what issues I can address.

rgbkrk:
Danger mode time, I'm nuking the old Rx definitions and letting `flow-typed` generate RxJS 5.5.x defs.

rgbkrk:
Everything is about to pass!

rgbkrk:
Can't nuke Rx without nuking other definitions:

```
Library type error:
flow-typed/npm/redux-observable_vx.x.x.js:31
 31:       scheduler: rxjs$SchedulerClass
                      ^^^^^^^^^^^^^^^^^^^ rxjs$SchedulerClass. Could not resolve name
Error: applications/desktop/src/notebook/kernel/shutdown.js:13
 13:   channels: rxjs$Subject<*>,
                 ^^^^^^^^^^^^^^^ rxjs$Subject. Could not resolve name
Error: packages/messaging/src/index.js:48
 48:   source: rxjs$Observable<JupyterMessage<*, *>>
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ rxjs$Observable. Could not resolve name
Error: packages/messaging/src/index.js:86
 86:     source: rxjs$Observable<JupyterMessage<*, *>>
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ rxjs$Observable. Could not resolve name

```

It's decision time. Either:

* Create some really good definitions that match well for RxJS
* Nuke more types!

captainsafia:
I think creating some definitions for RxJS will create the smallest footprint.

rgbkrk:
Oooook, I'll bring back the pretty good definitions here and we can fix them up.

rgbkrk:
Alright, back to me working through Rx + flow. 🤞

rgbkrk:
We are off to the races, time to see what Travis thinks.

rgbkrk:
Everything passes! 🎉

packages/core/src/dummy/index.js
```diff
@@ -17,24 +17,26 @@ export { dummyCommutable, dummy, dummyJSON } from "./dummy-nb";
 
 const rootReducer = combineReducers({
   // Fake out app, since it comes from
-  app: (state = AppRecord(), action) => state,
+  app: (state = makeAppRecord(), action) => state,
```
lgeiger, line 20: Where is `makeAppRecord` coming from?

It's not imported above. On the other hand `AppRecord` wasn't imported above either.

lgeiger (approved):
Thank you so much for doing this @captainsafia @rgbkrk 💯

This is really great! We can fix the FixMe's bit by bit down the road.

packages/core/src/dummy/index.js
captainsafia: `makeAppRecord` is an initial step into the conversion of the new Records-with-types support in Immutable 4.0. `makeAppRecord` is a Record factory that creates an AppRecord. It is defined in `pakcage/types/src/core/records`.

lgeiger:
I took the desktop app for a quick spin and everything seems to be working great. Let's go with this 🎉

packages/core/src/dummy/index.js
captainsafia: In the future, we will have factories like `makeDocumentRecord`, `makeImmutableNotebook` for all the typed Records in the app.

lgeiger: Thanks for the explanation.

I was just wondering about the import. It is used on line 20 but it's imported on line 27 (same as `AppRecord` before). I would expect flow or babel to fail here.

rgbkrk: It should be coming from `@nteract/types/core/records`, though it looks like this is never being called and not causing an issue?

Issue #2244, opened 2017-12-26
[play] Ability to switch servers


rgbkrk:
ref: #2236 

Maintainers: feel free to edit/update the description here.

/cc @mpacer @theengineear

mpacer:
It does seem like this feature already exists via 

https://github.com/nteract/nteract/blob/82611f8e92949da8e145ecb38b19b6e828e13a1c/applications/play/components/Main.js#L54-L60

What needs to be made is the server killing feature, which currently has no epic (see the comments):

https://github.com/nteract/nteract/blob/82611f8e92949da8e145ecb38b19b6e828e13a1c/applications/play/redux/epics.js#L44-L48

So should we change this issue to be "kill oldServer after changing servers"?

Or is there a way in which this functionality is missing that I'm misunderstanding?

rgbkrk:
That sounds right. I just copied over the text that was in the top of the originating PR.

PR #2260, opened 2017-12-28
[play] Kill old server

mpacer:
Adds a new epic to kill the oldServer if you change to a new server.

Also needed to adjust the handler to properly destructure the props. 

I used mergeMap because it's possible that you do this repeatedly if you keep changing the server… but map might be sufficient (wasn't sure if it was needed but this seemed safer).

### DIFF
--- a/applications/play/redux/epics.js
+++ b/applications/play/redux/epics.js
@@ -56,6 +56,20 @@ const activateServerEpic = action$ =>
     })
   );
 
+const killServerEpic = (action$, store) =>
+  action$.pipe(
+    ofType(actionTypes.KILL_SERVER),
+    mergeMap(({ payload: { serverId } }) => {
+      const { config } = store.getState().entities.serversById[serverId].server;
+      return shutdown(config).pipe(
+        mergeMap(data => {
+          return of(actions.killServerFulfilled({ serverId }));
+        }),
+        catchError(error => of(actions.killServerFailed({ serverId, error })))
+      );
+    })
+  );
+
 const fetchKernelSpecsEpic = (action$, store) =>
   action$.pipe(
     ofType(actionTypes.FETCH_KERNEL_SPECS),
--- a/applications/play/redux/epics.js
+++ b/applications/play/redux/epics.js
@@ -234,6 +234,7 @@ const runSourceEpic = (action$, store) =>
 
 const epics = combineEpics(
   activateServerEpic,
+  killServerEpic,
   fetchKernelSpecsEpic,
   setActiveKernelEpic,
   activateKernelEpic,
--- a/applications/play/redux/epics.js
+++ b/applications/play/redux/epics.js
@@ -60,7 +60,15 @@ const killServerEpic = (action$, store) =>
   action$.pipe(
     ofType(actionTypes.KILL_SERVER),
     mergeMap(({ payload: { serverId } }) => {
-      const { config } = store.getState().entities.serversById[serverId].server;
+      const oldServer = store.getState().entities.serversById[serverId];
+      if (!oldServer)
+        return of(
+          actions.killServerFailed({
+            serverId,
+            error: `server with id ${serverId} does not exist in store.`
+          })
+        );
+      const { config } = oldServer.server;
       return shutdown(config).pipe(
         mergeMap(data => {
           return of(actions.killServerFulfilled({ serverId }));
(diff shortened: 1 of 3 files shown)


mpacer:
I think this may close #2244 but I could be wrong since that's based on my reinterpretation of that issue.

rgbkrk:
Yeah, `mergeMap` is right here since you want to make sure to handle multiple servers being killed in succession. What's not quite answered here is what would happen if the same server was requested to be deleted in a row (and there are Rx ways of handling that).

I'll finish reviewing though.

applications/play/redux/epics.js
```diff
@@ -56,6 +56,20 @@ const activateServerEpic = action$ =>
...
+const killServerEpic = (action$, store) =>
+  action$.pipe(
+    ofType(actionTypes.KILL_SERVER),
+    mergeMap(({ payload: { serverId } }) => {
+      const { config } = store.getState().entities.serversById[serverId].server;
+      return shutdown(config).pipe(
+        mergeMap(data => {
+          return of(actions.killServerFulfilled({ serverId }));
+        }),
+        catchError(error => of(actions.killServerFailed({ serverId, error })))
```
rgbkrk, line 76: Smart!

applications/play/redux/epics.js
```diff
@@ -56,6 +56,20 @@ const activateServerEpic = action$ =>
     })
   );
 
+const killServerEpic = (action$, store) =>
+  action$.pipe(
+    ofType(actionTypes.KILL_SERVER),
+    mergeMap(({ payload: { serverId } }) => {
+      const { config } = store.getState().entities.serversById[serverId].server;
```
rgbkrk: I would check to make sure that `serverId` is in the `serversById` here and if it's not, either return:

```js
empty() // this is an observable creation method, like "of"
```

or

```js
of(actions.killServerFailed({ serverId, error: "server not known" }))
```

with a sufficient error message

rgbkrk (approved):

applications/play/redux/epics.js
mpacer: Sounds good!

rgbkrk: Let me know if/when you're tackling this one. Otherwise I can merge this as is and make it part of the overall deployment of _everything_ tonight.

mpacer: I’ll get to this in a bit; need to run a few errands and then it’s first on the queue

mpacer:
Not quite sure how to force serverId to be removed from the store…which in light of your comment seems like the right thing to do.

rgbkrk:
> Not quite sure how to force serverId to be removed from the store…which in light of your comment seems like the right thing to do.

Yeah, after successful shutdown (with that killServerFulfilled) we'd have to update the reducers to use that action to ~delete~ omit that server. Normally, if you want to delete a key out of an object, you use `delete`. With our store though, we shouldn't mutate state directly:

```js
> x = {'a': 2}
{a: 2}
> delete x['a'] // Let's not do this
true
> x
{}
```

I tend to use `omit` from lodash for this:

```js
> import { omit } from 'lodash'
> x = {'a': 2}
{ a: 2 }
> omit(x, 'a')
{}
> x
{ a: 2 }
```

mpacer:
Thank you for the comment about `omit` I wouldn't have thought of that! 

One of the things that I'm running into is the way that `createObjectReducer` works in combination with the way that reducers' keys map to the state inside `combineReducers`. 

Any pointers?

rgbkrk:
Since we just hashed this out a bit in chat, I'll port over the convo:

* We'll likely wrap the call to `createObjectReducer` so we can delete the key on some actions or change up how `createObjectReducer` works
* We keep having to recall how `createObjectReducer` is working, so we need to figure out a good balance for readability and maintainability around using it
* It's ok to punt on deleting the in-memory references to the servers after they've been deleted by binder. It's pretty small and we can fix up the above later.

mpacer:
ok i made some progress on this but now I'm running into a really strange issue:

the shutdown api is returning a 405 Method not allowed.

I'll push my changes so far.

rgbkrk:
> ok i made some progress on this but now I'm running into a really strange issue:
> the shutdown api is returning a 405 Method not allowed.

That sounds like the token isn't being done properly for the shutdown request. Do you have the full traceback? Sometimes it's a matter of a preflight being rejected (`OPTIONS` method).

rgbkrk:
Talked with M in chat, since the app still works yet doesn't shutdown properly I think we're ok to merge this in and file a bug. If I had to guess, it's on the `rx-jupyter` side.

mpacer:
It is a preflight being rejected in the console…I wasn't able to trace down
what that exactly meant… and it's getting the same config that all the
other rx-jupyter calls were getting, but it definitely could be on the
rx-jupyter side of things.

PR #2263, opened 2017-12-28
pre-factor the notebook provider component

rgbkrk:
While working on #2218, I noticed some bits that really needed to move either to their own component files or within providers. Let's see if I can get this situated then rebase #2218 on top of it.

* [x] extract the pinned cell and related components
* [x] move ye ole notebook.js to providers (for now, since it is a connected redux component)
* [x] fix tests
* [x] split cell into view layer and provider components
* [x] fix tests again


rgbkrk:
Yuck. The cell isn't _really_ disconnected from the store in this component now as the toolbar is still a provider component...

rgbkrk:
Getting closer though. 😄

rgbkrk:
All set now for the first round.

PR #2269, opened 2017-12-29
Finish off the nextgen and provider exports

rgbkrk:
This is going to be the final sweep over the components, making sure that "provider" components are in the right spot and that by default we export `@nteract/core/components` as purely view layer components and `@nteract/core/providers` as the opinionated redux store backed ones.

* [x] markdown cell --> markdown preview component
* [x] create provider components
* [x] clean up all the "next gen" usage
* [x] declare cell as a provider component for now
* [x] fix up the flow path for `@nteract/core`
* [x] fix tests
* [x] `<Notebook />` (from `@nteract/core/components`) --> `<Cells />`
* [x] `<Notebook />` from `@nteract/core/providers` --> `<NotebookApp />`

Follow up PR:

* `<Cell />` from provider will get broken apart as it's one of those components that get too much threaded through. That's what #2218 is seeking to accomplish and will be what I go after next.


rgbkrk:
Oooook, this makes the core package a bit more reasonable to read, with regular (no redux) components in `@nteract/core/component` and the connected ones in `@nteract/core/providers`. I also cleaned up all the `_nextgen` stuff. Next gen is _here_ now and we won't have to differentiate. 

We still probably want to do #2218, this at least makes a clean distinction here.

packages/core/__tests__/components/markdown-preview-spec.js
```diff
@@ -1,8 +1,9 @@
+// @flow
 import React from "react";
 
 import { shallow, mount } from "enzyme";
 
-import MarkdownCell from "../../../src/components/cell/markdown-cell";
+import MarkdownCell from "../../src/components/markdown-preview";
```
mpacer: Why is MarkdownCell in markdown-preview (as opposed to markdown-cell)?

packages/core/__tests__/providers/cell-spec.js
```diff
@@ -3,9 +3,9 @@ import Immutable from "immutable";
 
 import { mount } from "enzyme";
 
-import Cell from "../../../src/components/cell/cell";
+import Cell from "../../src/providers/cell";
```
mpacer, line 6: Much more reasonable! W00t!

packages/core/__tests__/components/markdown-preview-spec.js
rgbkrk: oh I should change this name, thanks

applications/showcase/pages/index.js
```diff
@@ -1,15 +1,13 @@
...
+import {
   Cell,
   Input,
   Prompt,
   PromptBuffer,
   Editor,
   Outputs,
   Notebook
-} = _nextgen;
+} from "@nteract/core/components";
```
mpacer: shouldn't this be `@nteract/core/providers`?

rgbkrk: Nope! This one (in showcase) was demonstrating how to use the next gen components.

applications/play/components/Main.js
```diff
@@ -4,15 +4,12 @@ import CodeMirrorEditor from "@nteract/editor";
 import { BinderConsole } from "./consoles";
 import { Display } from "@nteract/display-area";
 import { KernelUI } from "./kernelUI";
-import { _nextgen } from "@nteract/core/components";
+import { Outputs } from "@nteract/core/components";
```
mpacer, line 7: Shouldn't this be `@nteract/core/providers`?

packages/core/__tests__/providers/notebook-spec.js
```diff
@@ -7,7 +7,6 @@ import { shallow, mount } from "enzyme";
 import renderer from "react-test-renderer";
 
 import { displayOrder, transforms } from "@nteract/transforms";
-import Cell from "../../src/components/cell/cell";
```
mpacer, line 10: Why don't we need cells in the notebook tests any longer?

packages/core/providers.js
```diff
@@ -0,0 +1,2 @@
+// @flow
+module.exports = require("./").providers;
```
mpacer, line 2: 😄 👍

applications/play/components/Main.js
rgbkrk: Nope! We only use the presentational `Outputs` components in play, it's a totally separate redux store setup in play.

mpacer: Oh nope… actually it shouldn't be!

applications/showcase/pages/index.js
mpacer: Actually… I think this needs to be split between those components that are in `@nteract/core/providers` and `@nteract/core/components`

packages/commuter-frontend/components/contents/zeppelin.js
```diff
@@ -3,9 +3,7 @@ import * as React from "react";
 
 import { JSONTransform, HTMLTransform } from "@nteract/transforms";
 
-import { _nextgen } from "@nteract/core/components";
-
-const { Editor } = _nextgen;
+import { Editor } from "@nteract/core/components";
```
mpacer, line 6: I'd have imagined that this would require heavier logic and so would be inside `@nteract/core/providers`

applications/play/components/Main.js
mpacer: lovely!

applications/showcase/pages/index.js
mpacer: oh… i'm really confused now… above Notebook is a provider not a component…

packages/core/__tests__/providers/notebook-spec.js
rgbkrk: The notebook provider uses Cell directly and we _used_ to have a `CellComponent` prop where we passed the `Cell` implementation. That's completely gone (and has been for a bit), so I cleaned up this part of the test.

mpacer: sweet! well done!

lgeiger (approved):
Great stuff 👍

Can't wait for Travis to go green 😄

packages/commuter-frontend/components/contents/zeppelin.js
rgbkrk: This is probably a bad sign for the `<Editor>` component's naming. 🤔  It's designed to be used like this:


```js
<Editor>
  import this
</Editor>
```

which by default just shows a syntax highlighted representation of the text and has some attached styling. When you want an editable one, you pass in a child component that is actually editable:

```js
<Editor>
  <CodeMirror>
    import this
  </CodeMirror>
</Editor>
```

packages/core/src/components/markdown-preview.js
```diff
@@ -25,31 +24,43 @@ type State = {
...
 const renderer = new MarkdownRenderer();
-
 const mdRender: MDRender = input => renderer.render(parser.parse(input));
 
+// TODO: Consider whether this component is really something like two components:
+//
+//       * a behavioral component that tracks focus (possibly already covered elsewhere)
+//       * the actual markdown previewer
+//
+//       Since I'm really unsure and don't want to write a silly abstraction that
```
mpacer, line 40: I'm actually with you on separating this, previewing markdown is it's own concept with its own (display) decisions & it's not a silly abstraction. 

What is handling the keyboard shortcuts that move focus between different cells? Is that separate from "selecting" multiple cells? How does it currently handle multi-cell selection (&c.)

applications/showcase/pages/index.js
rgbkrk: yeahhhhhh... We should probably call it `<NotebookApp />` or something for the provider version. It's way different than the presentational `<Notebook />` which just does styling for between the (non-provider) `<Cell />`s.

lgeiger: This uses the [notebook component](https://github.com/nteract/nteract/pull/2269/files#diff-1c8f2160fb8b611b4ec84056347f1b0bR380) from `core/src/components/index.js` and not the provider

**Edit:** Not relevant anymore since Kyle answered it already above.

mpacer:
So this is great! Much simpler! 

Unfortunately, it introduces some really confusing nomenclature things — we seem to have duplicate names for providers and components. Presumably these have different purposes and "signatures"… maybe we could postfix "View" onto all of the view only components with little logic inside them?

packages/core/src/components/markdown-preview.js
rgbkrk: We don't have any multi-cell select logic.

The keyboard shortcuts are handling of when you have the cell focused so that you can flip back and forth between view mode. I've got to run, for the time being though I'd suggest playing with up/down on the current app. I can do a GIF and explanation later.

rgbkrk:
Thanks for the super earnest review @mpacer, I definitely want the nomenclature here to make sense, be approachable, and very useful.

packages/core/src/components/markdown-preview.js
mpacer: I don't think this split needs to occur yet… but if it doesn't occur I don't think this process of separating the providers from components is done. There's a lot of key handling logic in here when what matters is displaying the render from a view only perspective (in analogy with what you have above).

mpacer:
At the least, with them having the same names, there would be no way for someone to import both providers and components without renaming some of them or using the module itself as a namespace (which seems potentially annoying). 

I'm not really sure which should be the "default" though, the provider (implying that we postfix `View` or `Component` something to the component names) or the component (implying that we postfix `App` or `Container` or `Provider` or something to the provider names). 

Maybe to avoid confusion we could add both postfixes?

rgbkrk:
> At the least, with them having the same names, there would be no way for someone to import both providers and components without renaming some of them or using the module itself as a namespace (which seems potentially annoying).

Yeah that's a very good point.

Normally, as you see with the `Toolbar` component and the provider version of the `Toolbar` component, they _do_ map between the view component and the provider. The provider version merely augments the view to have callbacks hooked up to dispatch.

The two that don't follow this pattern are `Cell` and `Notebook`. The plan in #2218 was to do away with the `Cell` as a messy component you have to pass things through and make `Notebook` act more like `NotebookApp`. On some level, this just goes to show that we're bearing the legacy of learning React while building the app.

Right now I think I'm in favor of calling all the provider versions `Connected`, so `ConnectedToolbar`, `ConnectedNotebook`, etc. What do you think?

mpacer:
> Normally, as you see with the Toolbar component and the provider version of the Toolbar component, they do map between the view component and the provider. The provider version merely augments the view to have callbacks hooked up to dispatch.

AH — so in that case the expectation is that you'd only want one or the other so they can easily be swapped out for each other.

Ok in that case, I think my comments on the `MarkdownPreview` class need clarification. I thought the primary distinction was between "functional" pieces (callbacks, event management &c.) vs. "display" pieces (styling, layout, class assignment, passing props, mapping props to children). That is pretty different from this distinction between the redux store connected components which I now understand this to be focusing on. It just happened to be the case that most of the components in components weren't too js heavy and focused on what I was calling display. 

I still think that "functional" vs. "display" split should also occur. But perhaps that should occur later…

> Right now I think I'm in favor of calling all the provider versions Connected, so ConnectedToolbar, ConnectedNotebook, etc. What do you think?

My only concerned is that it doesn't actually specify that it's specifically intended to be connected to a redux store (not any generic state manager), but I think that can be mentioned in the docs and so should be fine.

rgbkrk:
I've done some thinking while away for a bit and I'm thinking the following for a few components:

* `<Notebook />` (from `@nteract/core/components`) --> `<Cells />`
* `<Notebook />` from `@nteract/core/providers` --> `<NotebookApp />`
* `<Cell />` from provider will get broken apart as it's one of those components that get too much threaded through. That's what #2218 was seeking to accomplish.

Part of my rubric here is from the [Presentational and Container Components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) post from Dan Abramov.

rgbkrk:
> I still think that "functional" vs. "display" split should also occur. But perhaps that should occur later…

In some cases, they're intertwined -- especially as it relates to "active", "selected", etc. We'll figure it out together over time here.

mpacer:
> In some cases, they're intertwined -- especially as it relates to "active", "selected", etc. We'll figure it out together over time here.

Makes sense! Happy to be on the journey with ye. 

> <Notebook /> (from @nteract/core/components) --> <Cells />

That seems to fit its purpose more cleanly too. 

> <Notebook /> from @nteract/core/providers --> <NotebookApp />

You technically don't need to change this since you got rid of the old `Notebook` component.

> <Cell /> from provider will get broken apart as it's one of those components that get too much threaded through. That's what #2218 was seeking to accomplish.

Hrm… I see what you mean about the dethreading in #2218. But when you say broken apart do you also mean creating separate components rather than working over one gigantic switch statement?

rgbkrk:
> do you also mean creating separate components rather than working over one gigantic switch statement?

As much as possible I want to be able to pass the props for display area, editor, etc. directly, without having to pass it through a component above it (as currently works with `NotebookApp`'s `<Cell />`). Right now that likely means a large component for the `<NotebookApp />`.

rgbkrk:
By the way, what I mean by threading is that, for example, with the `<Editor />` component, if I wanted to change things on (the old version of) the component, I had to thread (read: sew) all properties through each component above it. The hierarchy used to look something like this:

```
Notebook
  --> DraggableCell
    --> Cell
      --> CodeCell
        --> Editor
          --> CodeMirror
```

As you can see now, most of those components are gone and we've been favoring:

```
<Notebook>
  <Draggable>
    <Cell>
      <Editor>
        <CodeMirror>
```

The distinction being that we use `children` in the same way as you would with HTML.

rgbkrk:
I'm going to go ahead and merge this, let's keep refining our interfaces with new PRs. Luckily, unlike Python, we're only exporting _some_ of the interfaces here so we can keep changing the underneath bits. As it stands though, we'll be making a major release of `@nteract/core` (which is currently at 0.0.8).

PR #2272, opened 2017-12-30
clarify use of SyntaxHighlighter in our <Editor />

rgbkrk:
@mpacer asked a bit ago about why we use react syntax highligther rather than `CodeMirror` inside of the `<Editor />` component. The reason is performance for non-interactive views. It's also nice for server rendering. This is a just a comment update to clarify things.

### DIFF
--- a/packages/core/src/components/index.js
+++ b/packages/core/src/components/index.js
@@ -241,6 +241,9 @@ export class Editor extends React.Component<EditorProps> {
 
   render() {
     // Build in a default renderer when they pass a plain string
+    // This is primarily for use with non-editable contexts (notebook-preview)
+    // to make rendering much faster (compared to codemirror)
+    // Ref: https://github.com/nteract/notebook-preview/issues/20
     if (typeof this.props.children === "string") {
       return (
         <SyntaxHighlighter


mpacer:
If it's read-only, it seems a little bit odd to call it  `<Editor/>` wouldn't it be more accurately `<CodeRenderer/>`, `<SourceDisplay/>`, or `<SourceView/>` or something?

rgbkrk:
It's the component you'd wrap around the `<CodeMirror />` instance like this:

```js
<Editor>
  <CodeMirror value={source}/>
</Editor>
```

Whereas if you were merely showing it as a read-only display you'd use:

```js
<Editor>
  {source}
</Editor>
```

The component primarily exists to make sure that the style stays consistent between static rendering and codemirror rendering, while also working in tandem with the `<Prompt />` and overall `<Input />` component:

```js
<Input hidden={sourceHidden}>
  <Prompt />
  <Editor>
    <CodeMirror
      tip
      completion
      id={cellID}
      value={source}
      language={this.props.language}
      cellFocused={this.props.cellFocused}
      editorFocused={this.props.editorFocused}
      theme={this.props.theme}
      focusAbove={this.focusAboveCell}
      focusBelow={this.focusBelowCell}
    />
  </Editor>
</Input>
```

mpacer:
So then wouldn't it be more accurate to call it `<SourceContainer/>` (or even `<Code/>` or `<Source/>`) so that it doesn't imply that you can edit the rendered content?

rgbkrk:
I _want_ to use the postfix `Container` because I personally think of these as containers, _however_ the current standard nomenclature in the React community is that a

> [A container does data fetching and then renders its corresponding sub-component. That’s it.](https://medium.com/@learnreact/container-components-c0e67432e005)

I have to wonder though... Maybe I could make `<Input />` work with only `<Prompt />` and "whatever other child" is within `<Input />`... Then people have to make a choice between using CodeMirror and the ReactSyntaxHighlighter. That does mean that `@nteract/notebook-preview` has to explicitly switch to using `ReactSyntaxHighlighter`. That or we detect that they're passing a string as one of the children and assume that it should be a `ReactSyntaxHighlighter` block.

mpacer:
Is there a reason why `<Code/>` or `<Source/>` wouldn't work?

rgbkrk:
`Source` seems ok. Usage would be:

```js
<Cell>
  <Input>
    <Prompt />
    <Source>
      <CodeMirror />
    </Source>
  </Input>
  <Outputs />
</Cell>
```

I'm not particularly keen on it. Maybe what we expose for static viewing vs. editing is a different component, but they both use the same component underneath for slotting themselves for `<Input />`.

rgbkrk:
I'm going to merge this in since it provides an explanation as to what the component is doing at the moment. Feel free to make PRs updating the text or making an issue for the further discussion.

PR #2274, opened 2017-12-31
Detach more functionality from the Cell provider

rgbkrk:
More in efforts towards #2218. This PR:

* [x] creates a `HijackScroll` component that does exactly what cells were doing previously with an onClick handler and a `focused` prop. It's more of a `ScrollIntoViewIfNeeded` component.
* [x] clean up unused props from core's `<Cell />` component

This makes the current Cell component into something small and simple to bring into the `NotebookApp` provider component

### DIFF
--- a/packages/core/src/components/hijack-scroll.js
+++ b/packages/core/src/components/hijack-scroll.js
@@ -1,0 +1,58 @@
+// @flow
+
+import * as React from "react";
+
+type HijackScrollProps = {
+  focused: boolean,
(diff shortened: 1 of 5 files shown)


rgbkrk:
This is currently failing because I didn't run the jest snapshots after running a local build of `@nteract/core`. Since `@nteract/notebook-preview` is resolving modules within lib, it's missing the current version. Commit coming shortly.

PR #2283, opened 2017-12-31
Bandaid for a commuter

rgbkrk:
Enable uglify-es for commuter-frontend, up until next.js has a release that includes this.


lgeiger:
👍 next.js 4.3.0 will include this: https://github.com/zeit/next.js/pull/3150

PR #2285, opened 2018-01-01
switch to hoisted lerna

rgbkrk:
* [x] switch us back to hoisted (symlink _all_ the packages)

This cuts install time down by half from fresh. Also the "right" way to start up from scratch here is:

```
lerna clean --yes && npm install
```

All the next.js apps boot up much quicker now (no repeated deps), there are no styled-jsx strangenesses when starting, it's beautiful.

While doing this I ended up needing to fix quite a bit more so I've also fixed something with these:

* [x] `jest` now resolves straight from src for `@nteract/core`, which was a necessity for our internal packages that had a special `@nteract/core/NAME` path to ease usage.
* [x] `@nteract/core/themes` is a special path like `@nteract/core/components` (see above)
* [x] `enchannel-zmq-backend` allows for injecting the `jmp` dependency for zeromq sake
  * NOTE: We should probably make a cleaner setup here in the future. I put it this way to make it backwards compatible.


lgeiger:
> HOWEVER. I disabled the `electron-builder install-app-deps` just to make this work. 😢

This may work if you move `electron-builder` to the root `package.json` and adjust the app directory for the desktop app but it'll probably still fail when bundling the app.

rgbkrk:
Now that I've gone down this path with a freshened look (and the fact that I can troubleshoot it in between things), I think I may get this one wrapped up.

rgbkrk:
My most recent step was to make a way to inject the `jmp` dependency into `enchannel-zmq-backend`, so that the non-hoisted zeromq is the one that `enchannel-zmq-backend` uses. It's tempting to instead move this into the desktop app and let it carry its own copy of zeromq. We'd still extract what's effectively EventEmitter style hookup to multiplexed channels, which could be nice on its own since we use a similar setup in testing.

rgbkrk:
This... works!

rgbkrk:
Wow, we could even do yarn workspaces with this setup now. 🎉

rgbkrk:
No longer WIP. Since we're in a bit of a pickle with regards to:

* Current `master` exceedingly slow
* Travis having a large backlog due to failures at the start of the year

I'd love to see if you think this should me merged in @lgeiger and click that green button. I will of course update anything that goes wrong and keep this moving well forward. Afterward I'd like to make a patch release of the packages because they weren't all linked up to the right versions.

applications/desktop/package.json
```diff
@@ -137,6 +137,11 @@
     "kernelspecs": "^2.0.0",
     "lodash": "^4.17.4",
     "nteract-assets": "^4.0.0",
+    "netrc": "^0.1.4",
+    "mime": "^1.4.1",
+    "zeromq": "^4.0.0",
+    "https-proxy-agent": "^1.0.0",
+    "follow-redirects": "0.0.7",
```
lgeiger, line 144: These are coming mostly from the `node-github`, right?

applications/desktop/webpack.common.js
```diff
@@ -4,8 +4,6 @@ const path = require("path");
 
 const nodeModules = {
   jmp: "commonjs jmp",
-  uuid: "commonjs uuid",
-  github: "commonjs github",
```
lgeiger, line 8: Unfortunately this breaks github publishing since github@9 is incompatible with webpack: https://github.com/octokit/node-github/pull/593

We need to upgrade `node-github` for this to work and should move `uuid` and `github` to the `devDependencies` in desktop since it's now included twice in the packaged app.

package.json
```diff
@@ -21,8 +21,7 @@
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
     "postinstall": "npm run lerna",
-    "//lerna bootstrap": "links and builds packages",
-    "lerna": "lerna bootstrap",
+    "lerna": "lerna bootstrap --hoist --loglevel verbose --nohoist='{electron-builder,electron-updater,lodash,ijavascript,zeromq,netrc,mime,https-proxy-agent,follow-redirects,nteract-assets,mathjax-electron,jmp}'",
```
lgeiger, line 24: Do we still need to ignore `netrc,mime,https-proxy-agent,follow-redirects` when `github` is bundled by webpack?

Why are we ignoring `lodash`?

lgeiger (commented):
Thanks for making this work. 🎉 

I have a few comments that should be addressed to fix some problems with the desktop app.

applications/desktop/webpack.common.js
rgbkrk: Ahhhh, ok. Thanks.

package.json
rgbkrk: `lodash` was one of the ones that `electron-builder` was saying it couldn't resolve.

PR #2291, opened 2018-01-02
ensure styles are encoded properly by force

rgbkrk:
This fixes up an issue where the `"` in the style were being turned into `&quot;`. There's one other-ssr related issue here where an extra (or missing) `<div />` is causing an error on client side render.

### DIFF
--- a/applications/notebook-on-next/pages/edit.js
+++ b/applications/notebook-on-next/pages/edit.js
@@ -69,7 +69,9 @@ export default class Edit extends React.Component<*> {
     if (!this.props.serverNotebook) return <Error />;
     return (
       <div>
-        <style>{`@media print {
+        <style
+          dangerouslySetInnerHTML={{
+            __html: `@media print {
           * {
             box-shadow: none !important;
           }
@@ -117,7 +119,9 @@ export default class Edit extends React.Component<*> {
             "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji",
             "Segoe UI Emoji", "Segoe UI Symbol";
           font-size: 1em;
-        }`}</style>
+        }`
+          }}
+        />
         <Provider store={store}>
           <div>
             <NotebookApp />;
--- a/applications/notebook-on-next/pages/index.js
+++ b/applications/notebook-on-next/pages/index.js
@@ -25,7 +25,16 @@ class Input extends React.Component {
 
   render() {
     return (
-      <div className="centering">
+      <div>
+        <div className="centering">
+          <input
+            placeholder="Gist ID"
+            value={this.state.gistid}
+            onChange={this.handleChange}
+            onKeyDown={this.handleKeydown}
+          />
+          <button onClick={this.handleConfirm}>Launch Notebook</button>
+        </div>
         <style jsx>{`
           .centering {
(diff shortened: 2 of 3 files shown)


applications/notebook-on-next/pages/edit.js
```diff
@@ -69,7 +69,9 @@ export default class Edit extends React.Component<*> {
     if (!this.props.serverNotebook) return <Error />;
     return (
       <div>
-        <style>{`@media print {
+        <style
+          dangerouslySetInnerHTML={{
```
mpacer, line 73: Could we use `<style jsx-global>` to avoid needing to use dangerouslySetInnerHtml?

applications/notebook-on-next/pages/index.js
```diff
@@ -25,7 +25,16 @@ class Input extends React.Component {
 
   render() {
     return (
-      <div className="centering">
+      <div>
```
mpacer, line 28: Can’t we return an array of elements with react 16? Unless there’s a css rule expecting this… what do we gain from a blank wrapper `div`?

packages/core/src/providers/notebook-app.js
```diff
@@ -392,7 +392,16 @@ export class NotebookApp extends React.PureComponent<Props> {
             padding-right: 10px;
           }
         `}</style>
-        <style>{`:root { ${themes[this.props.theme]} } `}</style>
+        <style
+          dangerouslySetInnerHTML={{
```
mpacer, line 396: Same comment as above re: style jsx

applications/notebook-on-next/pages/edit.js
rgbkrk: `style-jsx` can't do nested, like with `@media print`

applications/notebook-on-next/pages/index.js
rgbkrk: I tried to use a [fragment](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html) here, but the webpack build was failing with it. Not sure if it was a next, babel, or flow issue.

packages/core/src/providers/notebook-app.js
rgbkrk: styled-jsx can't handle the css variable `:root` appropriately, the build will just fail.

applications/notebook-on-next/pages/index.js
lgeiger: Looks like we need to wait for babel 7 to support fragments: https://github.com/zeit/next.js/issues/3356

lgeiger (approved):
Thanks!

applications/notebook-on-next/pages/index.js
rgbkrk: Oh hey I didn’t realize I could use React.Fragment. I can change that here.

lgeiger: Sorry, merged too fast. I should've read the whole issue.

rgbkrk: Eh, all good. Leaves it as an opportunity for someone else. 😄

PR #2292, opened 2018-01-02
speed up action creators with fake(r) data

rgbkrk:
Invoking commutable to create the dummy notebook for each of these tests makes the actions tests much slower than they need to be, relative to a real test.


lgeiger (approved):
LGTM

PR #2293, opened 2018-01-02
Remove styled-jsx workarounds

lgeiger:
https://github.com/zeit/styled-jsx/issues/329 is fixed and we're hoisting our dependencies so we can remove a few workarounds.


rgbkrk:
Sweet!

PR #2302, opened 2018-01-03
set URL query on gitref/repo update for [play]

rgbkrk:
This is going to make it so that when a user submits a change to the gitref and repo, the URL will change.

Major problem though: the props aren't _actually_ being updated for `gitref` and `repo`. It looks like they're only local state to the component...

### DIFF
--- a/applications/play/components/Main.js
+++ b/applications/play/components/Main.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from "react";
 import Head from "next/head";
+import Router from "next/router";
 
 import CodeMirrorEditor from "@nteract/editor";
 import { BinderConsole } from "./consoles";
@@ -14,6 +15,7 @@ import * as utils from "../utils";
 
 const NTERACT_LOGO_URL =
   "https://media.githubusercontent.com/media/nteract/logos/master/nteract_logo_cube_book/exports/images/svg/nteract_logo_wide_purple_inverted.svg";
+
 class Main extends React.Component<*, *> {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


rgbkrk:
Sure enough, those are still doing ye ole `setState`, no going through redux.

rgbkrk:
Ah, we're not _using_ the `submitBinderForm` action yet.

rgbkrk:
Ok, all set! Now we've got eminently sharable URLs (sans content publishing).

rgbkrk:
Fair warning on the PR though -- while I was hacking on it I was able to crash my devtools (in Chrome) several times. I'm not sure if/what caused this and if it related to the router updates. I _really_ only care about the shown location.

rgbkrk:
No more WIP tag, ready for review (especially the actual routing bit).

applications/play/components/Main.js
```diff
@@ -273,34 +311,37 @@ class Main extends React.Component<*, *> {
...
-    "entities",
-    "serversById",
-    state.ui.currentServerId,
-    "server"
-  ])
-});
+const mapStateToProps = state => {
+  return {
+    repo: state.ui.repo,
+    gitref: state.ui.gitref,
```
theengineear, line 332: If we're going to store the `repo` and `gitref` in the url, we should remove it from the `ui` state. I.e., we want a single source of truth.

To handle this, I think we have access to a `url` prop for our `Page` component as documented in https://github.com/zeit/next.js/#shallow-routing.

I think the appropriate flow would be the following (not thinking about server failures yet...):

1. User submits binder form
2. We push a change to the url (shallow route)
3. `componentWillReceiveProps` notices a change to the `repo` and/or `gitref` (might even want to store these together as a single query param for clarity) and dispatches an action to create a new server connection.

On the initial page load, we should be able to more-or-less kick off the same action from `componentDidMount`, right?

theengineear: @rgbkrk is there any reason that we'd want to have this information live in our ui state and in the url?

rgbkrk: > On the initial page load, we should be able to more-or-less kick off the same action from componentDidMount, right?

Ah yeah we could do that.

rgbkrk: > is there any reason that we'd want to have this information live in our ui state and in the url?

The way I think of it is that there are two views (that are stateful) that show the current true state of `gitref` and `repo`:

* URL
* form

The way I triggered this right now is that the URL only shows the current state of actual submitted binder so that's it's copy-pastable by user. I don't have much opinion on this other than making sure that the user has a nice URL to copy that reflects what they'd like to share with others.

rgbkrk:
Since I think we've got some more to fix up here, I'm going to ship the current packages as is and we can refine this before shipping play again.

rgbkrk:
I rebased this, though I haven't addressed the concerns about location state. Still need to tackle that at some point.

theengineear (approved):

PR #2303, opened 2018-01-03
perform prefetching of mybinder DNS + logo

rgbkrk:
This helps prevent flicker of the binder logo and is a bit more useful if the drawer is closed on start (which we don't currently do). The amount of savings is probably not much, it's an easy thing to do this for DNS and images so the user experience is ✨ .

### DIFF
--- a/applications/play/components/Main.js
+++ b/applications/play/components/Main.js
@@ -1,5 +1,7 @@
 // @flow
 import * as React from "react";
+import Head from "next/head";
+
 import CodeMirrorEditor from "@nteract/editor";
 import { BinderConsole } from "./consoles";
 import { Display } from "@nteract/display-area";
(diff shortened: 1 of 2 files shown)


applications/play/components/Main.js
```diff
@@ -77,6 +79,21 @@ class Main extends React.Component<*, *> {
...
+            href="https://mybinder.org/static/logo.svg?v=f9f0d927b67cc9dc99d788c822ca21c0"
+          />
+          {/*
+            prefetch our little sample graphic for an extra touch of ✨
+          */}
+          <link rel="prefetch" href="https://bit.ly/storybot-vdom" />
+          <link
+            rel="prefetch"
+            href="https://media.giphy.com/media/xUPGcguWZHRC2HyBRS/giphy.gif"
+          />
```
lgeiger, line 95: Why do we prefetch both https://media.giphy.com/media/xUPGcguWZHRC2HyBRS/giphy.gif and https://bit.ly/storybot-vdom?

lgeiger (approved):
Great ✨
Also 👍 for hiding the drawer on initial load.

applications/play/components/Main.js
rgbkrk: Since prefetch usually happens in parallel, I did this so that the browser fires off requests for:

* The actual image on giphy
* The bitly link

Which would typically mean that once the bit.ly link is resolved the browser already has (or knows about) the actual image asset.

An unstated part of this is that I used the bit.ly link just so the python code is smaller within the play app. 😄

lgeiger: 👍

PR #2305, opened 2018-01-03
[Play] Initially hide console on startup

lgeiger:
This makes the initial loaded view a bit cleaner.


rgbkrk (approved):

rgbkrk:
The way this used to work was that it would show the drawer until the build was finished. We've got the actions in place to notice that, the behavior didn't port over.

We're also missing the `provisioning` display we used to show in the upper right too huh?

PR #2306, opened 2018-01-03
[Play] Initial code: Add unique key to vdom

lgeiger:
This prevents
```
warning.js?a041e4b:33 Warning: Each child in an array or iterator should
have a unique "key" prop.
```

### DIFF
--- a/applications/play/redux/getInitialState.js
+++ b/applications/play/redux/getInitialState.js
@@ -10,7 +10,7 @@ def fancy(word):
     colors = ['#FA8AAE', '#8AE7FA', '#FAFA8A', '#8AFA8A']
 
     return [
-        span(letter, style=dict(color=colors[idx % len(colors)]))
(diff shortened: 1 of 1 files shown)


applications/play/redux/getInitialState.js
```diff
@@ -10,7 +10,7 @@ def fancy(word):
     colors = ['#FA8AAE', '#8AE7FA', '#FAFA8A', '#8AFA8A']
 
     return [
-        span(letter, style=dict(color=colors[idx % len(colors)]))
+        span(letter, style=dict(color=colors[idx % len(colors)]), key=idx)
```
rgbkrk, line 13: oh snap!

rgbkrk: I originally made this a `span([span(...) ...])` then switched to an array to make it smaller.

PR #2315, opened 2018-01-03
Remove useless top-level elements

rgbkrk:
Since we're on React 16.2, I've taken the liberty of switching a bunch of our app to using either `React.Fragment`, raw arrays, or strings where possible.

I was trying to figure out some styling differences and it was much easier to reason about without all the extra divs (or spans). I'm going to keep at this and verify all the apps. A finetooth comb should probably be applied.

* [x] switch from wrapper divs to `<React.Fragment>`s
* [x] rely on raw strings and arrays where appropriate
* [x] update jest snapshots to match


lgeiger (approved):
🆒This is great!

PR #2316, opened 2018-01-04
remove unused hljs styles from showcase

rgbkrk:
Working on more CSS sync up between our light codemirror theme and our react-syntaxhighlighter and noticed these weren't really used.


lgeiger (approved):
👍

PR #2317, opened 2018-01-04
set the default values from the light theme inline

### LINKED ISSUE #2217: Provide fallbacks for the CSS variables in syntax highlighter theme
Just a few values missing from https://github.com/nteract/nteract/pull/2215, comments inline say what they are.

rgbkrk:
These are only used as fallbacks if someone doesn't include the dark or light theme on the page. I know this seems silly (shouldn't these be defined?!) -- it just makes our components stand alone without extra fuss around including themes on the page.

Closes #2217

### DIFF
--- a/packages/core/src/themes/syntax-highlighting.js
+++ b/packages/core/src/themes/syntax-highlighting.js
@@ -3,8 +3,8 @@ module.exports = {
     display: "block",
     overflowX: "auto",
     padding: "0.5em",
-    background: "var(--cm-background)",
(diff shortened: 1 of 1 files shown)


rgbkrk:
On a separate note, what if our default was `DeepPink`. Then it would be like a glaring

> :warning: :warning: "You haven't included the theme on the page, so there are undefined CSS variables lurking about"

😝

lgeiger (approved):
LGTM 👍

PR #2320, opened 2018-01-04
match our codemirror setings on rsh a bit better

rgbkrk:
I'm going to do something a bit more significant with the react syntax highlighter shortly, for now this should be workable.
 
rsh == React Syntax Highlighter


lgeiger (approved):
LGTM

PR #2327, opened 2018-01-05
allow unfocusing the editor for markdown preview

### LINKED ISSUE #1735: Ctrl + Enter should hide the code for a Markdown cell (not just Shift + Enter)
If you are editing a Markdown cell, you see both the text that you are typing, as well as the Markdown rendering of that text. When you press `Shift + Enter`, the text entry part of the cell closes, and all you are left with is the rendered markdown.

However, `Shift + Enter` also creates another cell below, which is not always desirable. This is especially noticeable when you have a Markdown cell in the middle of the document. The only way to hide the edit box is to use `Shift + Enter`, and then manually delete the extra cell that was created.

I believe that `Ctrl + Enter` should have a similar behavior, as far as closing the edit box in a markdown cell, and leaving only the rendered Markdown... without the extra line being inserted.

rgbkrk:
This closes #1735 by allowing the markdown cell to trigger an unfocusCellEditor callback. For the connected notebook app, this uses the `focusCellEditor` action, setting the id to `null` so no editor has focus.

I went the path of creating an `unfocusCellEditor` action then I realized how silly that was when it could be done with `focusCellEditor` and an `id`. I can still go back to that if people would prefer separate actions. LMK.

### DIFF
--- a/packages/core/src/components/markdown-preview.js
+++ b/packages/core/src/components/markdown-preview.js
@@ -1,6 +1,7 @@
 // @flow
 
 /* eslint jsx-a11y/no-static-element-interactions: 0 */
+/* eslint jsx-a11y/no-noninteractive-tabindex: 0 */
 
 import React from "react";
 import CommonMark from "commonmark";
@@ -12,10 +13,10 @@ import { Outputs, PromptBuffer, Input } from "./";
 
 type Props = {
   source: string,
-  focusAbove: () => void,
-  focusBelow: () => void,
   focusEditor: () => void,
   unfocusEditor: () => void,
+  focusAbove: () => void,
+  focusBelow: () => void,
   cellFocused: boolean,
   editorFocused: boolean,
   children: React$Element<*>
@@ -33,6 +34,8 @@ const parser = new CommonMark.Parser();
 const renderer = new MarkdownRenderer();
 const mdRender: MDRender = input => renderer.render(parser.parse(input));
 
+const noop = function() {};
+
 // TODO: Consider whether this component is really something like two components:
 //
 //       * a behavioral component that tracks focus (possibly already covered elsewhere)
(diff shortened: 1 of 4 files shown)


packages/core/src/components/markdown-preview.js
```diff
@@ -119,8 +127,7 @@ export default class MarkdownCell extends React.PureComponent<any, State> {
     const shift = e.shiftKey;
     const ctrl = e.ctrlKey;
     if ((shift || ctrl) && e.key === "Enter") {
-      this.setState({ view: true });
-      return false;
+      this.closeEditor();
```
lgeiger: Shouldn't we still return here?

lgeiger (approved):
👍I like having a single action for this.

packages/core/src/components/markdown-preview.js
rgbkrk: Whoops, yes!

rgbkrk:
Fixed the above and noticed something else. Since the div for the markdown cell has no tabindex, the keydown and keyup events aren't doing _anything_. It's entirely handled by a higher level keydown handler we have in the notebook and in the cells. I'm going to tear it out since it's unused. I see what the behavior was for in the past (handling inter-cell navigation from the markdown cell), this has to be on the outside now.

rgbkrk:
I both tore it out just now and brought it back. We have fully navigable Markdown cells within a document again. 🎉

rgbkrk:
Well this fixes a bunch of regressions. Time to get this shipped. 😄

rgbkrk:
I'd go ahead and merge @lgeiger, however I've made a bunch of new changes since your first approval. Check it out when you've got a chance.

lgeiger (approved):

PR #2330, opened 2018-01-05
[nomerge] switch to one big describe to check speed

rgbkrk:
The actions test takes like 30s. I'm curious if having one describe is better or even much larger tests within. Just seeing what Travis thinks.


rgbkrk:
Mostly looking at https://github.com/jsdf/posts/blob/master/content/Make-Jest-tests-run-fast.md and the jest monorepo to figure out how to fine tune this. I think some of these changes we can keep, others not so much.

rgbkrk:
This is worth a gander @lgeiger -- let me know what you like and don't like. 😄

applications/desktop/__tests__/renderer/index-spec.js
```diff
@@ -1,20 +0,0 @@
...
-describe("App", () => {
-  it("renders app", () => {
-    ipc.send = jest.fn();
-    jest.spyOn(actions, "setNotificationSystem");
-    const component = shallow(<App />);
-    expect(toJson(component)).toMatchSnapshot();
-    expect(ipc.send).toHaveBeenCalledWith("react-ready");
-    expect(actions.setNotificationSystem).toHaveBeenCalled();
-  });
-});
```
lgeiger, line 20: If the performance gain is noticeable, I'm good with removing this test.

package.json
```diff
@@ -66,13 +66,40 @@
       "./scripts/test-shim",
       "./scripts/test-setup"
     ],
+    "testPathIgnorePatterns": [
```
lgeiger, line 72: 👍

package.json
```diff
@@ -66,13 +66,40 @@
...
+      "/packages/.*/node_modules",
+      "/packages/.*/static",
+      "/packages/.*/build",
+      "/packages/.*/lib",
+      "/packages/.*/.next",
+      "/applications/.*/node_modules",
+      "/applications/.*/static",
+      "/applications/.*/build",
+      "/applications/.*/lib",
+      "/applications/.*/.next"
```
lgeiger: Should we add `<rootDir>/` in front of these, just to be sure?
http://facebook.github.io/jest/docs/en/configuration.html#testpathignorepatterns-array-string

package.json
```diff
@@ -66,13 +66,40 @@
...
+    ],
     "moduleNameMapper": {
+      "^nteract/(.*)$": "<PROJECT_ROOT>/packages/$1/src",
       "^@nteract/core/(.*)$": "<rootDir>/packages/core/src/$1",
+      "^enchannel-zmq-backend$": "<PROJECT_ROOT>/packages/enchannel-zmq-backend/src",
+      "^rx-binder$": "<PROJECT_ROOT>/packages/rx-binder/src",
+      "^rx-jupyter$": "<PROJECT_ROOT>/packages/rx-jupyter/src",
```
lgeiger, line 90: 👍

package.json
```diff
@@ -66,13 +66,40 @@
...
       "\\.css$": "<rootDir>/scripts/noop-module.js"
     },
     "coveragePathIgnorePatterns": [
```
lgeiger, line 93: Should we do the opposite and use [`collectCoverageFrom`](http://facebook.github.io/jest/docs/en/configuration.html#collectcoveragefrom-array) for more reliable coverage reports?

packages/core/__tests__/actions-spec.js
```diff
@@ -2,36 +2,30 @@
 import * as actions from "../actions";
 import * as constants from "../constants";
 
-describe("setExecutionState", () => {
-  test("creates a SET_EXECUTION_STATE action", () => {
+describe("actions", () => {
+  test("SET_EXECUTION_STATE action", () => {
```
lgeiger: If it improves performance I'd say go for it.

lgeiger (commented):
Thanks for looking into this! 💯

package.json
rgbkrk: Ah yeah definitely

package.json
rgbkrk: I suppose we could

packages/core/__tests__/actions-spec.js
rgbkrk: I think it only will if we group the ‘test’s.

package.json
rgbkrk: Done.

package.json
```diff
@@ -61,18 +60,49 @@
     "publish:all": "lerna run publish:all --scope nteract --stream"
   },
   "jest": {
+    "collectCoverageFrom": [
+      "packages/*/src/**/*.{js,jsx}",
+      "!**/node_modules/**"
```
lgeiger, line 65: What about the `applications/` folder?

rgbkrk:
Whoops! I broke coverage testing for a second here. 😄  I'll have this back to normal shortly.

package.json
rgbkrk: Yeah...

I suppose we'd need to be explicit for each type of app since desktop, the jupyter extension, and the next apps are pretty different.

rgbkrk:
I think I can reasonably say that this hasn't really changed the speed much, if at all.

rgbkrk:
Locally I'm sitting at 5.524s which is a good improvement. I don't know about Travis though.

lgeiger:
> Locally I'm sitting at 5.524s which is a good improvement. I don't know about Travis though.

Is it because we're now [actually running coverage](https://github.com/nteract/nteract/pull/2330/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R48) on Travis?

rgbkrk:
Ah, that may be true. I'll make a smaller PR for just some cleanup of the top level testing commands.

rgbkrk:
Now that I've run `npm t -- --coverage` locally, I can see that this _did_ help with the speed quite a bit.

rgbkrk:
The tests that could use the most speedup include:

* `packages/core/__tests__/providers/notebook-spec.js`
* `packages/core/__tests__/actions-spec.js`
* `packages/core/__tests__/providers/cell-creator-spec.js`
* `applications/desktop/__tests__/renderer/index-spec.js`

PR #2331, opened 2018-01-06
Token switch out

rgbkrk:
Since the name token appears all over our code base (primarily in jupyter), having a top level `token` in our core app record was unclear. I've changed it to `githubToken`. While I was at it I mocked `github` because I noticed that it was making requests to github.com in our tests, causing a slow down and is possibly why we were getting stalled out sometimes.


lgeiger (approved):
:+1: thanks

PR #2333, opened 2018-01-07
clean up top level test commands

rgbkrk:
Ripping just one thing from #2330, the cleanup of `test` commands in our top level package.json. No more `npm run test:desktop`, since we're on jest and should be using that only.


lgeiger (approved):
👍

PR #2337, opened 2018-01-07
Create new types, for Hosts and Kernels

rgbkrk:
This will be for both local "hosts" (the desktop app) and for connection to remote hosts:

* Jupyter Notebook Server
* JupyterHub
* Binders

This PR is the `HostRecord` and `KernelRecord` portion of #2338.

### DIFF
--- a/applications/desktop/src/notebook/reducers/app.js
+++ b/applications/desktop/src/notebook/reducers/app.js
@@ -4,7 +4,10 @@ import type { ChildProcess } from "child_process"; // eslint-disable-line no-unu
 
 import { shutdownKernel } from "../kernel/shutdown";
 
-import { makeAppRecord } from "@nteract/types/core/records";
+import {
+  makeAppRecord,
+  makeLocalKernelRecord
+} from "@nteract/types/core/records";
 
 import type { Channels } from "@nteract/types/channels";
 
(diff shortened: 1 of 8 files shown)


rgbkrk:
Since it seems like the kernel setup is pretty vital here, I'm going to carve those bits out too following on from #2338.

rgbkrk:
handy reference: https://facebook.github.io/immutable-js/docs/#/Record.Factory

rgbkrk:
I would ❤️  a look-over of this PR.

applications/desktop/src/notebook/reducers/app.js
```diff
@@ -24,18 +27,11 @@ declare class AppState {
 }
 
 function cleanupKernel(state: AppState): AppState {
-  const kernel = {
-    channels: state.channels,
-    spawn: state.spawn,
-    connectionFile: state.connectionFile
-  };
-  shutdownKernel(kernel);
+  shutdownKernel(state.kernel);
```
captainsafia: Flow gets mad at this line because there is no `kernel` attribute defined in the app state. I think what we can do here is. I think you might need to add a `kernel: KernelRecord` attribute to the AppState definition in the lines above.

captainsafia (commented):
Looks great! I posted one comment.

applications/desktop/src/notebook/reducers/app.js
rgbkrk: Thanks! I've swapped this out to use the proper `AppRecord` now.

rgbkrk:
Since this all passes, @captainsafia said it was great, and I've fixed the flow issue by adopting the official `AppRecord`, I'm going to go ahead and merge this. 😄

PR #2338, opened 2018-01-08
create plan document for new composable app state

rgbkrk:
In the course of getting all our apps to have unified state, actions, and epics that they can be composed from, I spent some time with @theengineear figuring out what our core state objects would be. Since this is a pretty big overhaul, it's likely pieces will be implemented in their own distinct chunks (I started on "hosts" in #2337). We'll have to compare and contrast this with our current state setup to ensure that we've gotten all the pieces we might have missed.

### DIFF
--- a/packages/types/core/plan.md
+++ b/packages/types/core/plan.md
@@ -1,0 +1,203 @@
+This document attempts to specify the goal for an nteract core package that can
+provide generic/core state management to all notebook-y applications. Right now
+this is focused on these apps:
+
+* Desktop
+* nteract on jupyter
+* `@nteract/play`
+* notebook on next
+
+This describes a pretty far and wide overhaul of our core logic, including the
+notebook itself.
+
+## Core concepts
+
+* `ref` - an internal _reference_ to an entity upon _recognition_, e.g. kernels, hosts, kernelspec collections, etc.
+* `id` - likely an external identifier, e.g. with /api/kernels/9092, 9092 is the id
+
+We use the term _recognition_ over _creation_ because we need to have a uid to
+reference _before_ we get a response from some api. A good example is having a uid
+for an active kernel before the kernel has been launched with a jupyter notebook server.
+Since there will be a proliferation of uids, the internal ones are refs while
+the external ones are called ids.
+
+---
+
+Side note -- In flow, these would be typed as:
+
+```js
+opaque type Id = string;
+opaque type Ref = string;
+```
+
+Which can also help us enforce we are using the right ids and refs amongst disparate entities:
+
+```js
+opaque type KernelId = Id;
+```
+
+---
+
+## Flattened Structure, Database Like Feeling
+
+Stemming from the Redux docs' [Normalizing State Shape](https://redux.js.org/docs/recipes/reducers/NormalizingStateShape.html), we setup our application to be collections of entries built in a relational fashion. We were doing something similar with `cellMap` (map of cell id to cell) and `cellOrder` (list of cell ids). We're taking it to the next level here.
+
+## The Proposed Structure
+
+```js
+{
+  // The ref for the current host
+  // On desktop we'll have the one built-in local host that connects to
+  // zeromq directly. On jupyterhub backed apps, you'll be able to switch to
+  // different hosts.
+  selectedHostRef: Ref,
+  hostRefs: List<Ref>,
+
+  configLastSaved: Date,
+
+  // The piece of state that allows the ui to show loading/error indicators.
+  // This is split apart from the entities definitions because the two parts of
+  // state serve very different purposes.
+  communication: {
+    notebook: {
+      isSaving: boolean,
+      lastSaved: Date,
+      error: ?Object
+    },
+    hosts: {
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          loading: boolean,
+          error: ?Object
+        }
+      }
+    },
+    kernels: {
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          loading: boolean,
+          error: ?Object
+        }
+      }
+    },
+    kernelSpecs: {
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          loading: boolean,
+          error: ?Object
+        }
+      }
+    }
+  },
+
+  // These are the actual data that we get back from
+  //   * API Calls
+  //   * User input
+  //   * Kernel output
+  entities: {    
+    outputs: {
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          data: Object,
+          metadata: Object,
+          transient: Object,
+          outputType: string // TODO: should be enummed
+        }
+      },
+
+      // For capturing the display ID mappings
+      displayIdToOutputRefs: {
+        [id: Id]: List<Ref>
+      }
+    },
+
+    cells: {
+      byRef: {
+        [ref: Ref]: {
+          type: "code" | "markdown" | "raw",
+          ref: Ref,
+          source: string,
+          metadata: Object,
+
+          // NOTE: the following fields are only on code cells
+          outputRefs: List<Ref>,
+          executionState: "finished" | "executing" | "queued" | "dirty",
+          executionCount: ?number,
+          lastExecuteMessage: JupyterMessage
+        }
+      },
+      order: List<Ref>
+    },
+
+    // Each host implementation has a set of kernels which may be activated.
+    kernelSpecs: {
+      byRef: {
+        [ref: Ref]: {
+          name: string,
+          ref: Ref,
+          hostRef: Ref,
+          resources: Object, // TODO: What was this?
+          spec: Object, // TODO: Type me!
+        }
+      }
+    },
+
+    // Is there an analogy for `kernelSpecs` that needs to exist for hosts? The
+    // `hosts` hunk of state indicates *active* hosts.
+    // hostSpecs?
+    // availableHosts?
+    hosts: {
+      byRef: {
+        [ref: Ref]: {
+          id: string,
+          ref: Ref,
+          type: "local"|"jupyter",
+          selectedKernelRef: Ref,
+	        kernelRefs: List<Ref>,
+          kernelSpecsRef: Ref,
+          defaultKernelName: string,
+          token: string,
+          serverUrl: string,
+          crossDomain: boolean,
+          messages: List<string> // binder only
+        }
+      }
+      refs: List<Ref>
+    },
+
+    // A host may have one active kernel (but we allow multiple to allow smooth
+    // transitions between switching kernels).
+    kernels: {
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          type: "local" | "jupyter", // same as server, unchanging
+          hostRef: Ref,
+          name: string,
+          lastActivity: Date,
+          channel: rxjs$Subject,
+          status: string,
+
+          id: string, // jupyter only
+
+          spawn: ChildProcess, // local only
+          connectionFile: string, // local only
+        }
+      }
+    },
+    notifications: {
+      byRef: {
+        [ref: Ref]: {
+          message: string,
+          // TODO: Figure out our structure here
+        },
+      refs: List<Ref>
+      }
+    }
+  }
+}
+```
--- a/packages/types/core/plan.md
+++ b/packages/types/core/plan.md
@@ -182,7 +182,7 @@ and `cellOrder` (list of cell ids). We're taking it to the next level here.
     hosts: {
       byRef: {
         [ref: Ref]: {
-          id: string,
+          id: Id,
           ref: Ref,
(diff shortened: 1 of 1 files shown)


packages/types/core/plan.md
```diff
@@ -0,0 +1,221 @@
...
+fashion. We were doing something similar with `cellMap` (map of cell id to cell)
+and `cellOrder` (list of cell ids). We're taking it to the next level here.
+
+## The Proposed Structure
+
+```js
+{
+  // The top level of core state can be considered a notebook, but we formalize
+  // that by nesting it under `notebook`.
+  notebook: {
```
theengineear: @rgbkrk I think it's clearer to have a `notebook` hunk of state here. What do you think? One place where it relieves some awkwardness is:

```
notebookLastSaved: Date,
configLastSaved: Date
```

to

```
notebook: {
    lastSaved: Date
},
preferences: {
  lastSaved: Date
}
```

It also makes the `communication` portion clearer, IMO.

packages/types/core/plan.md
```diff
@@ -0,0 +1,221 @@
...
+    },
+
+    // Each host implementation has a set of kernels which may be activated.
+    kernelSpecs: {
+      byRef: {
+        [ref: Ref]: {
+          name: string,
+          ref: Ref,
+          hostRef: Ref, // TODO: explain why kernelSpecs needs a reference to
+                        // its host?
```
theengineear: @rgbkrk I know we talked about this, but I'm forgetting why this is? Something to do with sending a shutdown message to a kernel? Is it different depending on the host? In this case, I wonder if giving the kernel/kernelSpec a reference to it's host is an anti-pattern. Perhaps we should just have a `hostType` instead? These are the only two places where a child knows about its parent. I feel like we should limit that knowledge to just what these children _actually_ need to know.

packages/types/core/plan.md
```diff
@@ -0,0 +1,221 @@
...
+          resources: Object, // TODO: What was this?
```
theengineear: I don't recall what resources was, we should probably remove it until we remember :)

packages/types/core/plan.md
```diff
@@ -0,0 +1,221 @@
...
+
+    // A host may have one active kernel (but we allow multiple to allow smooth
+    // transitions between switching kernels).
+    kernels: {
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          type: "local" | "jupyter", // same as server, unchanging
+          hostRef: Ref,  // TODO: explain why kernelSpecs needs a reference to
+                         // its host?
```
theengineear: Ditto here on this hostRef. Perhaps just `hostType` would do?

theengineear:
@rgbkrk I updated this thread based on some of the most recent changes I had made to that doc we'd been working with. I pointed out a couple open questions I have.

packages/types/core/plan.md
rgbkrk: That’s cool with me!

packages/types/core/plan.md
rgbkrk: For this one I’m not sure, I copied it direct from the google doc. I think it’s probably because we’d need to know which host to launch a kernel from given a collection of them. We do need all the server connection info for the binder launching.

packages/types/core/plan.md
rgbkrk: Resources is part of the response from the kernelspecs API (which I should probably note is different from local mode as we just see these as files). Example `GET` on `/api/kernelspecs`:

```json
{
  "default": "python3",
  "kernelspecs": {
    "python3": {
      "name": "python3",
      "resources": {
        "logo-32x32":
          "/user/nteract-vdom-6m4yabcq/kernelspecs/python3/logo-32x32.png",
        "logo-64x64":
          "/user/nteract-vdom-6m4yabcq/kernelspecs/python3/logo-64x64.png"
      },
      "spec": {
        "display_name": "Python 3",
        "language": "python",
        "argv": [
          "python",
          "-m",
          "ipykernel_launcher",
          "-f",
          "{connection_file}"
        ],
        "env": {}
      }
    }
  }
}
```

packages/types/core/plan.md
rgbkrk: `hostType` seems sensible for kernels. We could also call it `socketType`. However, I'm curious what we'd need available when we want to restart or interrupt a kernel on a jupyter notebook server, as we'd need access to the host URL and token.

packages/types/core/plan.md
```diff
@@ -62,6 +62,8 @@ and `cellOrder` (list of cell ids). We're taking it to the next level here.
       selectedHostRef: Ref,
       hostRefs: List<Ref>,
       lastSaved: Date
+      
+      // TODO: language information? Does this belong at the notebook level?
```
rgbkrk: It certainly belongs in the notebook document.

theengineear: I guess I'm just looking for guidance on what sort of keys/values are appropriate. Is it just a single key `language` or something?

theengineear:
@rgbkrk having `kernel` and `kernelSpecs` know about their `host` just feels oddly circular to me. I sorta feel like just having some component have `host` and `kernel` in its props will do the trick? I'm just wary of adding more state keys than are really necessary.

packages/types/core/plan.md
rgbkrk: It's shown in the start here: https://nbformat.readthedocs.io/en/latest/format_description.html#top-level-structure

```python
    "language_info": {
        # if language_info is defined, its name field is required.
        "name" : "the programming language of the kernel",
        "version": "the version of the language",
        "codemirror_mode": "The name of the codemirror mode to use [optional]"
    }
```

so:

```
languageInfo: {
  name: string,
  version: string,
  codemirror_mode: string
}
```

rgbkrk:
> having kernel and kernelSpecs know about their host just feels oddly circular to me. I sorta feel like just having some component have host and kernel in its props will do the trick? I'm just wary of adding more state keys than are really necessary.

That's fine let's just set the type here instead of a reference to the host.

packages/types/core/plan.md
```diff
@@ -0,0 +1,233 @@
...
+    },
+
+    hosts: {
+      byRef: {
+        [ref: Ref]: {
+          id: string,
+          ref: Ref,
+          type: "local"|"jupyter",
+          selectedKernelRef: Ref,
+	      kernelRefs: List<Ref>,
```
rgbkrk: The only way we'll have `kernelRefs` is if an app uses multiple kernels in one page. I think we can delete this one.

rgbkrk: Oh I see, we could have multiple. I have them called `activeKernelRefs` in my other PR. Is that ok wording choice?

packages/types/core/plan.md
```diff
@@ -0,0 +1,233 @@
...
+          ref: Ref,
+          type: "local" | "jupyter", // same as server, unchanging
+          hostRef: Ref,  // TODO: explain why kernelSpecs needs a reference to
+                         // its host?
+          name: string,
+          lastActivity: Date,
+          channels: rxjs$Subject,
+          status: string,
+
+          id: string, // jupyter only
```
rgbkrk: Oh, should this be an `Id` type here?

rgbkrk:
I think we're ok to merge this and keep evolving this with new PRs. I'll go ahead and rebase this.

rgbkrk:
@captainsafia -- would love your thoughts on this one, including if we need to document more in here

packages/types/core/plan.md
captainsafia: Adding on to this, does it makes sense to break down the notebook state to different pieces. Like so:

```
notebook: {
   host: { the host details described above }
   metadata: { top-level metadata here (for example, the language) }
   document: the actual cells
}
```

packages/types/core/plan.md
captainsafia: I think `kernelRefs` is a fine wording. `activeKernelRefs` implies that there are inactive kernels. What would an inactive kernel be? On that is running but you are not connected to, one that isn't running, etc.

packages/types/core/plan.md
captainsafia: I think yes. Since this is referencing the kernel id that the kernel and front-end agree upon for communication?

captainsafia (commented):
Cool spec! Added some comments...

packages/types/core/plan.md
rgbkrk: The reason that we've been aiming at putting the `hostRefs` and other `*Refs` at the top level is to create a structure that can be re-used in non-notebook apps (like in play). I also think that, nomenclature wise, we should probably call any structure detailing the notebook app as `notebookApp`.

packages/types/core/plan.md
rgbkrk: An inactive kernel would be any that might still be shutting down, or, in the case of play (currently), be a kernel that just isn't the currently selected one by the user.

packages/types/core/plan.md
rgbkrk: Yeah, I'll update it now.

packages/types/core/plan.md
```diff
@@ -0,0 +1,233 @@
...
+      byRef: {
+        [ref: Ref]: {
+          ref: Ref,
+          type: "local" | "jupyter", // same as server, unchanging
+          hostRef: Ref,  // TODO: explain why kernelSpecs needs a reference to
+                         // its host?
+          name: string,
+          lastActivity: Date,
+          channels: rxjs$Subject,
+          status: string,
```
theengineear: We should change this to `executionState` since that's what's been used previously https://github.com/nteract/nteract/pull/2385/commits/aeb4b0b39aa003512046624a3a7426ccf5d18c31#diff-cefd1a3d3e94e5d5e684286dd7a56c21R37

rgbkrk: In #2385 I ended up switching (mostly) everything to `kernel.status`, since we also use this to depict the other states around the kernel launch I think it fits better as the status of the kernel.

rgbkrk:
Something I learned a bit more fully via this plan doc and #2385 -- having old kernels that have to be cleaned up be able to be shutdown cleanly in epics is highly important, both for ensuring we're leaving the users' system in a good state and for giving us opportunity to see if/when the lifecycle isn't going well by inspecting the store.

rgbkrk:
In particular, these hokey little reducers should be in epics that can clean up old kernels rather than only being able to see current kernels: https://github.com/nteract/nteract/pull/2385/files#diff-cefd1a3d3e94e5d5e684286dd7a56c21R41

rgbkrk:
We should probably rebase this so that it gets the updated circle ci setup.

theengineear:
huh... I thought I did rebase it. maybe I still missed some changes. @rgbkrk i'll update it now

EDIT:

oop, right you were :)

rgbkrk:
Another addition we'll probably want to add to this plan is how we'll represent the contents API entries.

Generally, it's a map from `path` (`/reporting/constructive-model.ipynb`) to the contents API responses. I'll use https://mybinder.org/v2/gh/nteract/vdom/master?urlpath=%2Fapi%2Fcontents%2Fexample-notebooks
 for this example.

```json
GET /api/contents/example-notebooks

{
  "mimetype": null,
  "writable": true,
  "name": "example-notebooks",
  "last_modified": "2018-01-02T21:44:05Z",
  "format": "json",
  "path": "example-notebooks",
  "created": "2018-01-24T17:47:24.918265Z",
  "type": "directory",
  "content": [
    {
      "mimetype": "text/markdown",
      "writable": true,
      "name": "README.md",
      "last_modified": "2018-01-02T21:44:05Z",
      "format": null,
      "path": "example-notebooks/README.md",
      "content": null,
      "created": "2018-01-24T17:47:24.914265Z",
      "type": "file"
    }
  ]
}
```

Here's https://mybinder.org/v2/gh/nteract/vdom/master?urlpath=%2Fapi%2Fcontents%2Fexample-notebooks%2FREADME.md

```json
GET /api/contents/example-notebooks/README.md

{
  "created": "2018-01-24T17:44:14.577890Z",
  "path": "example-notebooks/README.md",
  "last_modified": "2018-01-02T21:44:05Z",
  "writable": true,
  "type": "file",
  "mimetype": "text/markdown",
  "name": "README.md",
  "content":
    "Please submit your own notebooks built with vdom here!\n\n[We need your notebooks and we need you, join us!](https://github.com/nteract/vdom/issues/4)\n",
  "format": "text"
}
```

Honestly we're really going to be mapping `(HostRef, Path)` to a Contents entry. This one needs some thought. For now, since I'm working on the jupyter extension I'll only work on the one host case.

rgbkrk:
image

rgbkrk:
Opting for a refs approach here too, as it will allow us to do renames successfully from the tree view.

rgbkrk:
I'm going to go ahead and call this on the path to where we want, let's go ahead and merge it. We can follow up with new PRs against the plan document.

Issue #2277, opened 2017-12-31
Post release woes


rgbkrk:
Just some scattered notes while I fix this mega release

```
⸨░░░░░░░░░░░░░░░░░░⸩ ⠴ npmPublish: verb execScript No prepublish script found at /Users/kyle/code/src/github.com/nteract/nteract/applications/jupyter-extension/nteract_on_jupyter
```

```
lerna ERR! publish Retrying failed publish: nteract
lerna ERR! publish Retrying failed publish: nteract
⸨         ░░░░░░░░░⸩ ⠦ npmPublish: ERR! publish Retrying failed publish: nteract
```

```
⸨         ░░░░░░░░░⸩ ⠙ npmPublish: ERR! publish Retrying failed publish: nteract
lerna ERR! publish Retrying failed publish: nteract
lerna ERR! publish Ran out of retries while publishing nteract Error: Command failed: npm publish --tag latest
lerna ERR! publish npm ERR! "toString()" failed
lerna ERR! publish
lerna ERR! publish npm ERR! A complete log of this run can be found in:
lerna ERR! publish npm ERR!     /Users/kyle/.npm/_logs/2017-12-31T17_29_06_478Z-debug.log
lerna ERR! publish
lerna ERR! publish     at Promise.all.then.arr (/Users/kyle/.nvm/versions/node/v8.0.0/lib/node_modules/lerna/node_modules/execa/index.js:236:11)
lerna ERR! publish     at <anonymous>
lerna ERR! publish     at process._tickCallback (internal/process/next_tick.js:169:7)
lerna ERR! execute callback with error
lerna ERR! Error: Command failed: npm publish --tag latest
lerna ERR! npm ERR! "toString()" failed
lerna ERR!
lerna ERR! npm ERR! A complete log of this run can be found in:
lerna ERR! npm ERR!     /Users/kyle/.npm/_logs/2017-12-31T17_29_06_478Z-debug.log
lerna ERR!
lerna ERR!     at Promise.all.then.arr (/Users/kyle/.nvm/versions/node/v8.0.0/lib/node_modules/lerna/node_modules/execa/index.js:236:11)
lerna ERR!     at <anonymous>
(node:14331) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 16): [object Object]
[...]

PR #2339, opened 2018-01-08
create some guided steps to ease release

rgbkrk:
Ever since #2277, I've been hoping to make the release process a bit more staged and with a few more verification steps. The new steps (and as documented in the releasing notes):

```
npm run publish
npm run publish:commuter
npm run publish:desktop:all
```

I think with it set up in this way I'd probably add in the two web apps deployed on now. No worries for now though.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -14,7 +14,9 @@ In order to build a signed copy with working auto-update, you will need to join 
 
 ## Release Process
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
Heh, apparently `lerna publish` with `--ignore` will not do the publishing of all the other packages. Cleaning this up some more. ¯\\\_(ツ)\_/¯

PR #2341, opened 2018-01-08
begin migrating last vega release to new package

rgbkrk:
Sets up a new package handle the old version of vega-embed separately.

/cc @nikitakit

### DIFF
--- a/packages/vega-embed2/package.json
+++ b/packages/vega-embed2/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "@nteract/transform-vega",
-  "version": "3.0.6",
-  "description": "Vega Transform",
+  "name": "@nteract/vega-embed2",
+  "version": "1.0.0",
+  "description": "Direct vega2 and vega-lite 1 embed",
   "main": "lib/",
   "nteractDesktop": "src/index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


packages/vega-embed2/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@nteract/vega-embed2",
+  "version": "1.0.0",
+  "description": "Direct vega2 and vega-lite 1 embed",
+  "main": "lib/",
+  "nteractDesktop": "src/index.js",
```
nikitakit, line 6: not sure what this field is

packages/vega-embed2/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "build:flow": "flow-copy-source -v -i '**/__tests__/**' src lib",
+    "build:lib": "babel -d lib src --ignore '**/__tests__/**'",
+    "build:lib:watch": "npm run build:lib -- --watch",
+    "build:watch": "npm run build:clean && npm run build:lib:watch && npm run build:flow"
+  },
+  "author": "",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": "https://github.com/nteract/nteract/tree/master/packages/transform-vega",
```
nikitakit: repo url will need to be changed, too

packages/vega-embed2/package.json
rgbkrk: this allows us to not have to transpile packages when building apps -- it should probably be called `"nteractAppEntry"` or something. It's the equivalent to the `main` field.

packages/vega-embed2/package.json
rgbkrk: Thank you!

rgbkrk:
By the way, I realize how silly the babel setup is for this package -- I left it consistent with the other packages just so they're all the same. It's been problematic when I've started (solely) ES5 packages in here in the past that turned out to need transpilation after new contributions.

rgbkrk:
Merging this so @nikitakit can head forward with this.

PR #2342, opened 2018-01-08
Add transforms for vega 3 and vega-lite 2

nikitakit:

This adds transforms for the latest version of Vega.

(I couldn't get tests to run locally, but hopefully there's some CI bot in place to catch errors)

### DIFF
--- a/applications/desktop/webpack.common.js
+++ b/applications/desktop/webpack.common.js
@@ -4,7 +4,8 @@ const path = require("path");
 
 const nodeModules = {
   jmp: "commonjs jmp",
-  canvas: "commonjs canvas"
+  canvas: "commonjs canvas",
+  "canvas-prebuilt": "commonjs canvas-prebuilt"
 };
 
 const mainConfig = {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
     "postinstall": "npm run lerna",
-    "lerna": "lerna bootstrap --hoist --loglevel verbose --nohoist='{electron-builder,electron-updater,lodash,ijavascript,zeromq,nteract-assets,mathjax-electron,github,jmp}'",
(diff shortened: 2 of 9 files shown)


applications/desktop/webpack.common.js
```diff
@@ -4,7 +4,8 @@ const path = require("path");
 
 const nodeModules = {
   jmp: "commonjs jmp",
-  canvas: "commonjs canvas"
+  canvas: "commonjs canvas",
+  "canvas-prebuilt": "commonjs canvas-prebuilt"
```
nikitakit, line 8: I actually have no idea what this syntax does.

The situation here is that the latest vega has optional dependencies on both canvas and canvas-prebuilt. Webpack has issues resolving canvas-prebuilt, so I just pattern-matched the code here to get things to work.

I have no idea if there are other places that I need to patch.

rgbkrk:
One other thing you'll want to do is add canvas to the no-hoist section of the root level package.json, right here:

https://github.com/nteract/nteract/blob/464e17003b4ba745776cbd386a354c094473cb3a/package.json#L24

package.json
```diff
@@ -21,7 +21,7 @@
   "homepage": "https://github.com/nteract/nteract",
   "scripts": {
     "postinstall": "npm run lerna",
-    "lerna": "lerna bootstrap --hoist --loglevel verbose --nohoist='{electron-builder,electron-updater,lodash,ijavascript,zeromq,nteract-assets,mathjax-electron,github,jmp}'",
+    "lerna": "lerna bootstrap --hoist --loglevel verbose --nohoist='{electron-builder,electron-updater,lodash,ijavascript,zeromq,nteract-assets,mathjax-electron,github,jmp,canvas,canvas-prebuilt}'",
```
nikitakit, line 24: Did I do this right?

nikitakit:
The CI test reports have some canvas-related warnings showing up, but I don't know if this is new or not. But it looks like tests pass.

package.json
rgbkrk: Yup.

rgbkrk:
Yeah I definitely wish that vega didn't have canvas as an optional dependency, it forces us to have it try to install even though we don't want it.

rgbkrk:
This is working well for me, though I don't have data to test with for the newer vega.

rgbkrk:
Let's do this!

PR #2343, opened 2018-01-09
Bring execution epic into core

rgbkrk:
This begins the migration of epics into the `@nteract/core` package, starting with execute. Only non-desktop specific epics will end up in core.


rgbkrk:
I'm just shuffling files so the code coverage change should be a non-issue.

PR #2350, opened 2018-01-09
Create a unified @nteract/markdown package to use all over nteract

### LINKED ISSUE #1037: "\\" does not give newline in Markdown cell LaTeX
In Markdown cell LaTeX blocks, "\\" is not parsed as a newline (as it is in Jupyter notebook); one is required to give four consecutive backslashes ("\\\\"). When the notebook is subsequently opened in the Jupyter environment, LaTeX is then rendered differently again (which makes matrices etc. look awkward).

### LINKED ISSUE #1330: Tables not rendering in markdown cell
The following simple table displays properly in Jupyter, but not nteract.

| H1 | H2 |
| --- | --- |
| v1 | 0 |

### LINKED ISSUE #1557: markdown table not rendered
```
| Age | Number of Cars owned | Owns House | Number of Children| Marital status (s/m/d/w) | Owns a dog | Bought a Boat|
|-|-|-|-|-|-|-|
|66|1|y|2|w|n|y|
|52|2|y|3|m|n|y|
|22|0|n|0|m|y|n|
|25|1|n|1|s|n|n|
|44|0|n|2|d|y|n|
|39|1|y|2|m|y|n|
|26|1|n|2|s|n|n|
|40|3|y|1|m|y|n|
|53|2|y|2|d|n|y|
|64|2|y|3|d|n|n|
|58|2|y|2|m|y|y|
|33|1|n|1|s|n|n|
```
image

### LINKED ISSUE #1663: Underscores in inline LaTeX should not make the source italic
The markdown editor doesn't recognize underscores in inline LaTeX as subscripts, but formats the source as though the rendered text will be italic. The rendered text is correct though, so this issue is just a minor annoyance. Example below.

image

### LINKED ISSUE #1875: strikethrough markdown doesn't seem to work
I've tried both `~~strikethrough~~` and `<s>strikethrough</s>`. The later works in a regular notebook setup, haven't tried the former.

screenshot 2017-09-06 09 24 48

### LINKED ISSUE #2012: Mathjax \begin{align} environment does not render properly.
The following markdown code

```
$$
\begin{align}
a &=1 \\
b &=2
\end{align}
$$
```

should render two equations in two separate lines, with the equal signs aligned. It works this way under jupyter notebook.
Hoever in nteract it renders everything in one line.
untitled

### LINKED ISSUE #2037: LaTeX Issue with \left\{  \right\}
In a Jupiter notebook, the markdown editor knows how to correctly handle 
```
\begin{equation}
\left\{ ... \right\}
\end{equation}
```

ie. it knows to make appropriately sized curly braces around whatever is between the `\left` and `\right`. However, it seems that nteract doesn't know how to handle it for me and the editor refuses to render the LaTeX. Furthermore, sometimes when exploring this issue it'd make my entire editor go blank and I would need to close and open the notebook to be able to see anything again.

### LINKED ISSUE #2115: Evaluate react-markdown + remark-math plugin
https://github.com/rexxars/react-markdown/issues/10#issuecomment-347763068

```typescript
import * as ReactMarkdown from "react-markdown";
import MathJax from "@matejmazur/react-mathjax";
import RemarkMathPlugin = require("remark-math");

export const MarkdownRender = (props: ReactMarkdown.ReactMarkdownProps) => {
  const newProps = {
    ...props,
    plugins: [
      RemarkMathPlugin,
    ],
    renderers: {
      ...props.renderers,
      math: (props: {value: string}) =>
        <MathJax.Node>{props.value}</MathJax.Node>,
      inlineMath: (props: {value: string}) =>
        <MathJax.Node inline>{props.value}</MathJax.Node>,
    }
  };
  return (
    <MathJax.Context input="tex">
      <ReactMarkdown {...newProps} />
    </MathJax.Context>
  );
};
```

### LINKED ISSUE #2434: Issue with rendering a Latex equation
Hello,

I'm trying to render the following piece of Latex code in a Markdown cell:
```
$\hat{y}_{i}=\hat\beta_{0}+\hat\beta_{1}x_{i}$
```

It looks like the parser is getting confused by the first `\hat` - removing it produces the expected results. The preview widget on mathjax.org renders it fine, but in my notebook it's not displayed as rendered Latex, just my original code.

I'm using 0.6.2.

Thanks,

Clément

tlyon3:
The CommonMark markdown renderer currently being used escapes certain LaTeX commands (specifically, those with a single punctuation mark following the slash: `\,`, `\;`, etc...) causing the punctuation mark to be left in. 

Switching to `react-markdown` package fixes this issue and doesn't escape anything between the MathJax delimiters (default `$` and `$$`).

Closes #2434 
Closes #2037
Closes #2012
Closes #1037
Closes #1663 
Closes #1557
Closes #1875
Closes #2115
Closes #1663 
Closes #1330

This also makes it so that markdown tables and strikethrough render properly. 🎉 

--------

This completely switches us over to a new package called `@nteract/markdown` to control both our mathjax and markdown rendering. This change affects:

* `@nteract/transforms` - markdown and latex transforms
* `@nteract/core` - markdown cells
* `@nteract/notebook-preview`  it sets up it's own top level `<MathJax.Context>` so there's only one for all children
* `nteract` desktop - for now, until nteract/mathjax-electron#30 is resolved, we're relying on MathJax v2.7.1 from cdnjs.
* `@nteract/commuter` - rendering of markdown documents, zeppelin notebooks, and all jupyter notebook rendering

By default the markdown renderer and the latex output will create their own `<MathJax.Context>` if there's not already one. For any apps where you want to control the MathJax setup, wrap the highest level component that should have context (for a notebook, the cells).

```
<MathJax.Context>
  <App />
</MathJax.Context>
```

These are the props for `<MathJax.Context>`:

```js
Props = {
  children: React.Node,
  didFinishTypeset: ?() => void,
  script: string | false,
  input: "ascii" | "tex",
  delay: number,
  options: Object,
  loading: React.Node,
  noGate: boolean,
  onError: (err: Error) => void,
  onLoad: ?Function
};
```


packages/notebook-preview/src/index.js
```diff
@@ -146,7 +140,9 @@ export class NotebookPreview extends React.PureComponent<Props, State> {
                       {/* TODO: embed mathjax on the page here, otherwise LaTeX
                               rendering doesn't actually happen
                     */}
-                      <LatexRenderer>{mdRender(source)}</LatexRenderer>
+                      <LatexRenderer>
+                        <Markdown source={source} />
```
rgbkrk: Nice, does this work as is without their `remark-math` plugin? xref: https://github.com/nteract/nteract/issues/2115

tlyon3: I believe so. I didn't have to use that custom MarkdownRenderer. Just imported the package and used it. Looking through its source code, I couldn't find a reference to the `remark-math` plugin

rgbkrk:
Would you be willing to standardize the markdown setup across the entire monorepo?

If not, I can likely do that -- then we could just make one `@nteract/markdown` package we use internally that uses the LaTeX and markdown setup here, in all markdown contexts.

mpacer:
Sweet! A few questions:

Is `import MarkdownRenderer from "commonmark-react-renderer";` still needed with this change?

Are there any known quirks about `react-markdown` that we should look out for?

Can you give a rough idea of how it avoids the problems with MathJax?  I ask both because I'm generally curious and because the classic notebook's solution (yank out the mathjax, run markdown and replace the mathjax) isn't my favorite and I'd love to see a better solution.

Do our test files include commands with 
> a single punctuation mark following the slash: `\,` , `\;`
? The example snapshots don't seem to handle that. 

@rgbkrk Am I correct in thinking that nteract doesn't and won't support the double slash LaTeX delimiters (`\\(…\\)` and `\\[…\\]`; n.b.: the dollar signs are technically TeX delimiters and so are dispreferred by some)?

How does this handle multiline LaTeX blocks defined with `\begin{.*}…\end{.*}`?

rgbkrk:
> Am I correct in thinking that nteract doesn't and won't support the double slash LaTeX delimiters (`\\(…\\)` and `\\[…\\]`; n.b.: the dollar signs are technically TeX delimiters and so are dispreferred by some)?

Dollar signs are the simplest to support, in terms of not conflicting the with markdown grammar whereas `\\[` is completely plausible in markdown and should be able to be written without being treated as `LaTeX`. That being said, the mathjax config we have for mathjax-electron does allow for `\\[` -- https://github.com/nteract/mathjax-electron/blob/master/resources/MathJax/config/electron.js

> How does this handle multiline LaTeX blocks defined with \begin{.*}…\end{.*}?

I don't think this does. This swaps out the markdown parser and renderer to not be commonmark anymore so that the entities we want, particularly `$$`. Everything else is up to the mathjax config from [mathjax-electron](https://github.com/nteract/mathjax-electron/blob/master/resources/MathJax/config/electron.js)

tlyon3:
> Would you be willing to standardize the markdown setup across the entire monorepo?

@rgbkrk yeah I can do that. I've got a few other things I need to get done first, but I'll work on it

tlyon3:
> Is import MarkdownRenderer from "commonmark-react-renderer"; still needed with this change?

@mpacer No it isn't. This is replacing the `commonmark-react-renderer` with the `react-markdown` package.

> Are there any known quirks about react-markdown that we should look out for?

I haven't run into any quirks in my usage of it.

> Can you give a rough idea of how it avoids the problems with MathJax?

If it finds a block of MathJax, then it won't escape single characters (`\,`, `\;`, etc) This issue doesn't come up with other MathJax that aren't single characters (eg: `\thinspace`) which is why the issue was a bit obscure.

rgbkrk:
I'll be taking a look at this next and rebasing against master. Stay tuned!

rgbkrk:
Fully rebased. I've put basic snapshots in even though they're kind of useless since they're not deep. Next up is switching our `<Markdown />` transform to use this and getting rid of our old `LaTeX` handler.

rgbkrk:
Notes from me for things to fix:

```
lerna WARN EHOIST_PKG_VERSION "@nteract/markdown" package depends on react-markdown@^3.1.4, which differs from the hoisted react-markdown@^3.0.0.
```

Whereas this can only be fixed by making a totally separate package:

```
lerna WARN EHOIST_PKG_VERSION "@nteract/transform-vega" package depends on vega-embed@^3.0.0-rc7, which differs from the hoisted vega-embed@^2.2.0.
```

packages/markdown/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "repository": "https://github.com/nteract/nteract/tree/master/packages/markdown",
+  "keywords": [
+    "markdown",
+    "mathjax"
+  ],
+  "author": "Trevor Lyon <github.com/tlyon3>",
+  "license": "BSD-3-Clause",
+  "dependencies": {
+    "react-markdown": "^3.1.4",
+    "react-mathjax": "git+https://github.com/wko27/react-mathjax.git",
```
rgbkrk: What's the reasoning behind pointing to this particular fork?

rgbkrk: I think I'll bring this one in directly into the `@nteract/markdown` package.

rgbkrk:
OMG, this is looking so good and editing LaTeX has _zero_ flicker.

screen shot 2018-01-30 at 12 48 35 pm

Thank you so much for kicking this off @tlyon3!

/cc @jeffwong -- LaTeX is about to be way better now in the notebooks, both in the coming nteract web ui and in commuter!

rgbkrk:
Ok I've forked @wko27's most excellent handling of `react-mathjax` into `@nteract/markdown`, included flow types and caught a few edge cases as a result.

Hi @wko27! Love what you've put into the component. I ended up forking it so we didn't have to rely on a particular git commit for our use in nteract projects.

mpacer:
@nteract What are the LaTeX and Markdown transforms? Do you mean the transforms for the output mimetypes?

rgbkrk:
Yeah, those are the transforms for the output mimetypes for LaTeX and markdown. This PR unifies what markdown and latex setup is used for:

* Markdown Cells
* LaTeX mimetype
* Markdown mimetype
* Markdown (file) rendering in commuter
* Markdown rendering for Zeppelin notebooks in commuter

While also fixing a crap ton of outstanding LaTeX bugs we've had for a while.

mpacer:
What is the Remark-math plugin doing now?

rgbkrk:
Wow, `Co-authored-by` actually works in the UI.

rgbkrk:
> What is the Remark-math plugin doing now?

It's only use in here is for demarcating what denotes an inline math or math block, which are then rendered using that mathjax component.

rgbkrk:
`git shortlog -sne | cut -f2 | sed "s'^'Co-authored-by: '"` is my new yoink+give_attribution

rgbkrk:
Uglify!!!

```
lerna ERR! execute callback with error
lerna ERR! Error: Command failed: npm run build
lerna ERR! > Failed to build on /var/folders/kd/cylz4mhs1_9cpsjh0_c_gzfr0000gn/T/1a73dcd6-b736-4d17-9692-2303d0643105
lerna ERR! { Error: commons.js from UglifyJs
lerna ERR! Unexpected token operator «=», expected punc «,» [commons.js:72809,43]
lerna ERR!     at /Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/dist/server/build/index.js:183:21
lerna ERR!     at emitRecords.err (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/Compiler.js:269:13)
lerna ERR!     at Compiler.emitRecords (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/Compiler.js:375:38)
lerna ERR!     at emitAssets.err (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/Compiler.js:262:10)
lerna ERR!     at applyPluginsAsyncSeries1.err (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/Compiler.js:368:12)
lerna ERR!     at next (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/tapable/lib/Tapable.js:218:11)
lerna ERR!     at Compiler.compiler.plugin (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/performance/SizeLimitsPlugin.js:99:4)
lerna ERR!     at Compiler.applyPluginsAsyncSeries1 (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/tapable/lib/Tapable.js:222:13)
lerna ERR!     at Compiler.afterEmit (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/Compiler.js:365:9)
lerna ERR!     at require.forEach.err (/Users/kylek/code/src/github.com/nteract/nteract/node_modules/next/node_modules/webpack/lib/Compiler.js:354:15)
lerna ERR!   errors:
lerna ERR!    [ 'commons.js from UglifyJs\nUnexpected token operator «=», expected punc «,» [commons.js:72809,43]' ],
lerna ERR!   warnings: [] }
lerna ERR! npm ERR! code ELIFECYCLE
lerna ERR! npm ERR! errno 1
lerna ERR! npm ERR! @nteract/play@1.3.2 build: `next build`
lerna ERR! npm ERR! Exit status 1
lerna ERR! npm ERR!
lerna ERR! npm ERR! Failed at the @nteract/play@1.3.2 build script.
lerna ERR! npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

rgbkrk:
If I disable builds for `play` and `notebook-on-next` then all is well. I think I'll just upgrade the next apps to use a new version of next + styled-jsx in another PR.

mpacer:
~~I'm going to see if I can get this working without remark-math since I don't get why that extra parsing should be necessary if react-markdown is doing its job.~~

I understand now why it needs to be there, but it has some issues.

mpacer:
There are still some parsing issues (e.g., it shouldn't matter whether I put new lines after the $$)
Contrast:
With no newlines it renders this as a inline math element
image

With newlines it renders this as a equation math element
image

mpacer:
It appears that this is a feature of Remark-math (though I disagree):

> ## Double dollars in inline
> Some TeX packages and Markdown processors use double dollars, $$, as a inline token. Remark Math will parse it also properly.
> 
> $$\alpha$$

https://github.com/Rokt33r/remark-math#double-dollars-in-inline

Based on some of the content at this [math-in-markdown wiki](https://github.com/cben/mathdown/wiki/math-in-markdown), it's mostly git related applications based on kramdown (e.g., [GitBook](https://github.com/cben/mathdown/wiki/math-in-markdown#gitbook), [gitter](https://github.com/cben/mathdown/wiki/math-in-markdown#gitter))  that have this convention (where `$$…$$` is treated as inline and not a display equation). 

This is not what either notebook or lab uses for its interpretation of math syntax. It also is not, by default, what mathjax itself expects (where `$$…$$` is to be treated as a display equation). 

Remark math has a feature that enables something that is [_somewhat_ like a display equation](https://github.com/Rokt33r/remark-math#inlinemathdouble-experimental-use-with-caution) if you add the right css, but that seems super fragile.

So this is going to be a huge improvement, but is going to move away from consistency (in this small regard) with the rest of the jupyter ecosystem.

rgbkrk:
Thank you for the outline and edge cases @mpacer! I take it what we could do for now is go with this current setup, following it up with adjusting the inline and block plugins from `remark-math`?

rgbkrk:
As for the ugify bits, `master` is not failing when running the build of the apps. If I had to guess, either the code here or one of the dependencies has untranspiled bits. Options:

* Disable the build for apps (for now)
* See if bumping the next.js + styled-jsx versions works out
* Adjust the webpack config for play + others to use the uglify-es (again...)

rgbkrk:
annnnnd it's the remark-math plugin that has untranspiled bits

screen shot 2018-01-30 at 6 36 02 pm

I might just fork that into here too since we are inevitably going to be changing how the parser works anyways...

rgbkrk:
Woo! Ok now to fix up desktop so that it includes our bundled mathjax still because it should work offline. I commented out the include from the `index.html` here.

rgbkrk:
Hmmmm... Now that I see how this all works, it's apparent that most apps should have `<MathJax.Context>` at a much higher level, since it's literally for setting MathJax via the context API. Any component that wants to opt in to the configured mathjax can use contextTypes.

rgbkrk:
I tried an experiment where a `<MathJax.Context>` can opt in to a parent `<MathJax.Context>`. It seems impossible to do that, based on when the context is set. 🤔  Maybe though if I have a different context variable to set.

rgbkrk:
Took another stab at this, in order to have `<MathJax.Context />` be a child of another `<MathJax.Context />`, it would somehow have to trigger once the parent is loaded.

packages/transforms/src/markdown.js
```diff
@@ -1,19 +1,12 @@
 /* @flow */
-import React from "react";
-import CommonMark from "commonmark";
-import MarkdownRenderer from "commonmark-react-renderer";
+import * as React from "react";
```
captainsafia, line 2: Is there a particular reason this import style is used instead of `import React from 'react'`.

packages/transforms/src/latex.js
```diff
@@ -1,38 +1,24 @@
...
   render(): ?React$Element<any> {
     return (
-      <div
-        ref={el => {
-          this.el = el;
-        }}
-      />
+      <MathJax.Context input="tex">
+        <MathJax.Node>{this.props.data}</MathJax.Node>
+      </MathJax.Context>
```
captainsafia: Super elegant!

captainsafia (commented):
Started looking through this but I'll give it a more thorough look through when it's not midnight! 🤣

packages/transforms/src/markdown.js
rgbkrk: From the [flowtype docs for React](https://flow.org/en/docs/react/components/#toc-class-components):

> **Note:** We import `React` as a namespace here with `import * as React from 'react'` instead of as a default with `import React from 'react'`. When importing React as an ES module you may use either style, but importing as a namespace gives you access to React’s [utility types](https://flow.org/en/docs/react/types).

rgbkrk:
The biggest nit I have with this setup is that in notebooks with N Markdown cells there will end up being N many `<MathJax.Context>` which each load a _separate_ MathJax. The only way to get around that, since my multiple contexts approach didn't work, is to enforce that users of `@nteract/markdown` declare their own `<MathJax.Context>` at a high level.

The other reason it's annoying to have separate contexts is that they're deeply nested. For the desktop app I want to set it to use mathjax-electron because it's a local copy. For `<NotebookPreview />` a CDN is fine.

rgbkrk:
OMG. Actually, I think I _might_ have had the context stuff working right. I instead was running into an issue where it just wasn't rendering the mathjax on initial load.

rgbkrk:
:boom: We can now define custom MathJax config much higher up in an nteract app, and all the children will use the same MathJax setup, whether they're a markdown cell, a LaTeX transform, or a Markdown transform. 🎉

rgbkrk:
I had to pin react-virtualized in the midst of this, as they had a bad release in the middle of doing successive builds here. We can unpin it later.

applications/desktop/static/index.html
```diff
@@ -3,7 +3,6 @@
   <head>
     <meta charset="UTF-8">
     <link rel="stylesheet" href="../lib/styles.css"/>
-    <script type="text/javascript" src="../node_modules/mathjax-electron/resources/MathJax/MathJax.js?config=electron"></script>
```
rgbkrk, line 6: We can bring this back once https://github.com/nteract/mathjax-electron/issues/30 is closed. The config within the desktop app will then be:

```js
<MathJax.Context script={false}>
  {appStuff}
</MathJax.Context>
```

rgbkrk:
Tests pass, let's get some reviews in (because this is a bit of a fundamental change) and then let's merge it!

packages/markdown/src/mathjax/context.js
```diff
@@ -0,0 +1,125 @@
...
+
+  constructor(props: Props) {
+    super(props);
+    this.state = { loaded: false };
+    (this: any).onLoad = this.onLoad.bind(this);
+  }
+
+  getChildContext() {
+    return {
+      MathJax: typeof MathJax === "undefined" ? undefined : MathJax,
```
theengineear: 🤔 Was this supposed to supply a different default?

packages/transforms/__tests__/__snapshots__/latex-spec.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`LaTeXDisplay processes basic LaTeX 1`] = `null`;
```
theengineear: Sorta not testing much, right? Should this really be `null` here?

theengineear (approved):
Wow. Amazing work, folks! This is a really clean PR. I just had a couple comments. Really awesome work!

packages/markdown/src/mathjax/context.js
rgbkrk: Here's where it's weird (and I clearly should leave a comment). This code is checking if `MathJax` is defined globally by running a typeof on a potentially not set value then _explicitly_ setting the `MathJax` context to undefined.

packages/transforms/__tests__/__snapshots__/latex-spec.js.snap
rgbkrk: I should sneak some more snapshots in here. :) Because of how `React.Node` doesn't render until mount, it's only a `null`. I'll get some better tests here. I was over reliant on flow for helping me figure my way through this.

rgbkrk:
Tests updated, we're more covered now!

packages/transform-dataresource/package.json
```diff
@@ -34,7 +34,7 @@
     "d3-time-format": "^2.0.5",
     "lodash": "^4.17.4",
     "moment": "^2.18.1",
-    "react-virtualized": "^9.7.5",
+    "react-virtualized": "9.18.0",
```
rgbkrk, line 37: This change could go in a separate PR, it was in solely to deal with a regression in react-virtualized that broke our tests.

packages/markdown/src/mathjax/context.js
theengineear: ah. gotcha, thanks for adding a note!

packages/markdown/__tests__/remark-math-spec.js
```diff
@@ -0,0 +1,268 @@
...
+                }
+              }
+            },
+            "\\alpha"
+          )
+        ])
+      ])
+    );
+  });
+});
```
theengineear, line 268: ^^ nice, that's a good call

packages/transforms/__tests__/__snapshots__/latex-spec.js.snap
```diff
@@ -1,3 +1,22 @@
...
+  noGate={false}
+  onError={[Function]}
+  onLoad={null}
+  options={Object {}}
+  script="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-MML-AM_CHTML"
+>
+  <Node
+    inline={false}
+    onRender={null}
+  >
```
theengineear, line 18: 🙇

theengineear:
Awesome, thanks for beefing up the tests @rgbkrk!

rgbkrk:
As is customary with nteract contributions that demonstrate a spirit of positivity and continued collaboration, I'm inviting you to the org @tlyon3. Welcome to the nteract party! 🎉

PR #2352, opened 2018-01-10
rely solely on core for action type constants

rgbkrk:
Cleaned up all the spread out constants. I just noticed a few more, I'll get those cleaned up too.

* [x] Finish finding epics that declare their own constants

### DIFF
--- a/applications/desktop/src/notebook/epics/kernel-launch.js
+++ b/applications/desktop/src/notebook/epics/kernel-launch.js
@@ -46,7 +46,9 @@ import {
   LAUNCH_KERNEL,
   LAUNCH_KERNEL_BY_NAME,
   SET_LANGUAGE_INFO,
-  ERROR_KERNEL_LAUNCH_FAILED
+  ERROR_KERNEL_LAUNCH_FAILED,
+  KERNEL_RAW_STDOUT,
+  KERNEL_RAW_STDERR
 } from "@nteract/core/constants";
 
 export function setLanguageInfo(langInfo: LanguageInfoMetadata) {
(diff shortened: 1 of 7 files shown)


rgbkrk:
_Done_. Next PR will make sure that epics don't have their action creators next to them (this is entirely my fault...)

applications/desktop/src/notebook/epics/kernel-launch.js
```diff
@@ -95,11 +97,11 @@ export function newKernelObservable(kernelSpec: KernelInfo, cwd: string) {
       const kernelSpecName = kernelSpec.name;
 
       spawn.stdout.on("data", data => {
-        const action = { type: "RAW_STDOUT", payload: data.toString() };
+        const action = { type: KERNEL_RAW_STDOUT, payload: data.toString() };
         observer.next(action);
       });
       spawn.stderr.on("data", data => {
-        const action = { type: "RAW_STDERR", payload: data.toString() };
```
theengineear, line 102: 🚨  These are actually changing the string. Is that OK? `"RAW_STDERR"` --> `"KERNEL_RAW_STDERR"`.

rgbkrk: Good catch! It's not a problem here as we have no reducers or anything else that relies on it. This output currently goes _nowhere_, we just emit it through redux so that we can see it with debugging tools. I figured the name could be more specific so I took the liberty to change it.

theengineear: :+1: cool, then no biggie :tada:

theengineear (approved):

PR #2353, opened 2018-01-10
Message logic to core reducers

rgbkrk:
Ooooook, this was done mostly in light of #2338. More of the logic behind messaging has moved from the epics into the reducers. This should have been done all along and I'd like to continue moving all the stateful parts of the message handling from the epics into our reducers. I think a lot is going to simplify. As a result of this work, I was able to set the queued status that I set up for display some time ago (check out the ellipses):

queued-cells

TODO:

* [x] Add tests for `acceptPayloadMessage`
* [x] Add tests for `sendExecuteRequest` reducer (and probably change its name)


rgbkrk:
This is ready for review!

applications/desktop/__tests__/renderer/menu-spec.js
```diff
@@ -248,12 +248,15 @@ describe("menu", () => {
         .getState()
         .document.getIn(["notebook", "cellOrder"])
         .first();
+
+      const source = store
```
betatim: where is this used later on?

rgbkrk: Ah, it's not -- I ended up faking the `message` below.

packages/core/src/reducers/document.js
```diff
@@ -451,17 +453,60 @@ function newCellAppend(state: DocumentState, action: NewCellAppendAction) {
...
+  const { id, payload } = action;
+
+  if (payload.source === "page") {
+    // append pager
+    return state.updateIn(["cellPagers", id], Immutable.List(), l =>
+      l.push(payload.data)
+    );
+  } else if (payload.source === "set_next_input") {
+    if (payload.replace) {
+      // replace cell source here
```
betatim: could you extend this comment to explain when this happens? Is this the tab completer? That would be my guess. I can tell the cell source is being replaced from the code but not why or what with ;)

betatim:
LGTM. I will stir my pasta sauce while travis does the testing, you can extend the comment now or make a mini PR, as you prefer.

packages/core/src/reducers/document.js
rgbkrk: This happens when you use `%load` in IPython. I'll add a comment about this here.

Example:

percent-load

rgbkrk:
I extended the comment so we can wait on Travis for this one.

packages/core/src/reducers/document.js
betatim: Ha, guessed wrong :)

PR #2354, opened 2018-01-10
remove old records setup

rgbkrk:
While working on the records for local / remote host & kernel, I noticed this file was still around. I verified that it wasn't being used anymore (in favor of `@nteract/types/core/records`).


betatim:
My favourite kind of PR: deleting more code than it is adding!

betatim:
Tests still pass so it must be correct :)

PR #2357, opened 2018-01-11
Migrate DocumentRecord to RecordFactory

captainsafia:
Gonna carve away at this bit by bit.


rgbkrk:
This is fantastic, thanks.

Issue #2270, opened 2017-12-29
Housekeeping


rgbkrk:
A few things we need to cleanup, which are free for folks to tackle:

* [x] Update the flow config to resolve the path for the dropdown menu
* [x] Create a Markdown renderer package `@nteract/markdown` to become consistent (#2350)
*  ~~Switch _all_ `PureComponent` to `Component` because we don't _really_ have full immutable structures~~ I was wrong about this one.

I'm just coming up with these while finishing some other bits of cleanup and wanted to at least track them in an issue.

PR #2358, opened 2018-01-11
Use Component instead of PureComponent

captainsafia:
This addresses part of #2270.


rgbkrk:
Nice! We can then switch to proper `shouldComponentUpdate`s here.

PR #2359, opened 2018-01-11
instead of erroring the stream, log bad messages

### LINKED ISSUE #2356: No Syntax-Highlighting in 0.5.5
Everything is black, white or gray :(

image

rgbkrk:
An attempt at figuring out how the parent header is missing from some messages. Doesn't completely fix #2356, worth a cleanup though I think.


captainsafia:
Oh these pesky bugs! 🐛

Issue #2360, opened 2018-01-11
flowconfig issue


rgbkrk:
It just dawned on me that our flow config is wrong.

```
module.name_mapper='^@nteract/(.*)$' -> '<PROJECT_ROOT>/packages/$1/src/'
```

will make `@nteract/play`, `@nteract/nbextension`, etc. all point to the wrong spot. I'm not sure if this effects things. Seems like a bit of funk though.

rgbkrk:
Oh hey, this is fixed, thanks stale bot.

PR #2361, opened 2018-01-11
Fix flow and check more

rgbkrk:
This likely won't pass because of #2360, namely that packages aren't resolving properly from applications that have their package name as `@nteract`...

### DIFF
--- a/applications/notebook-on-next/pages/_document.js
+++ b/applications/notebook-on-next/pages/_document.js
@@ -1,8 +1,9 @@
+// @flow
 import React from "react";
 import Document, { Head, Main, NextScript } from "next/document";
 
 export default class MyDocument extends Document {
-  static getInitialProps({ renderPage }) {
+  static getInitialProps({ renderPage: Function }) {
     const { html, head, errorHtml, chunks } = renderPage();
     return { html, head, errorHtml, chunks };
   }
--- a/applications/notebook-on-next/pages/_document.js
+++ b/applications/notebook-on-next/pages/_document.js
@@ -3,7 +3,7 @@ import React from "react";
 import Document, { Head, Main, NextScript } from "next/document";
 
 export default class MyDocument extends Document {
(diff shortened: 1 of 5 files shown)


betatim:
What do we conclude from the fact that it did pass?

rgbkrk:
I'm thinking that _somehow_ locally my setup wasn't fully bootstrapped (each of our packages weren't linked in these apps for some reason).

rgbkrk:
I was able to fix it locally (and quite sadly) by doing the `git clean -xfd && npm i` trick.

applications/notebook-on-next/pages/_document.js
```diff
@@ -1,8 +1,9 @@
+// @flow
 import React from "react";
 import Document, { Head, Main, NextScript } from "next/document";
 
 export default class MyDocument extends Document {
-  static getInitialProps({ renderPage }) {
+  static getInitialProps({ renderPage }: { renderPage: Function }) {
```
betatim, line 6: What does this do? Not sure I've seen this notation before. `renderPage` is of type `Function` but `renderPage` is also of type hash with key renderPage ?

rgbkrk: I'll expand this a bit locally to explain --

```js
// getInitialProps is a function unique to next.js, not normal to React
static getInitialProps
// Apparently our flow setup doesn't have this definition already situated
// The first (and only?) argument is an object
// which has a property called renderPage
({ renderPage }:
// Everything after the ":" is the actual type definition for this
// so we're saying the type of this argument is an object with property renderPage,
// which is a Function (which I could have typed out the type of more explicitly --
// I aimed for passing for now)
{ renderPage: Function })
```

betatim: ah! I had forgotten about the unpacking via `{ keyHere }`! Makes sense.

betatim:
merges in return for eduction. I think this is what they call a win-win :)

PR #2366, opened 2018-01-12
impose host record setup in core AppRecord

rgbkrk:
This introduces the `Host` record into the `AppRecord`, intended for switching between Desktop and Jupyter Hosts (including binder & jupyterhub in general).

* [x] Create a `host` record for the `AppRecord`
* [x] Set the desktop app up with a default "Desktop" host record


rgbkrk:
Now that I've gotten used to `Immutable.Record`s, I look forward to switching from `Immutable.Map` in all the places we have consistent keys.

screen shot 2018-01-11 at 10 31 10 pm

PR #2370, opened 2018-01-12
Move CodeMirror options to options prop

captainsafia:


packages/core/src/providers/editor.js
```diff
@@ -18,14 +18,15 @@ type Props = {
...
-  language: string
+  options: Object
 };
 
 function mapStateToProps(state: Object): Object {
   return {
-    cursorBlinkRate: state.config.get("cursorBlinkRate"),
+    options: {
+      cursorBlinkRate: state.config.get("cursorBlinkRate")
+    },
```
rgbkrk: This will need to merge in `ownProps` (the [second argument to `mapStateToProps`](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options))

packages/editor/src/index.js
```diff
@@ -73,8 +69,9 @@ class CodeMirrorEditor extends React.Component<
   static defaultProps = {
     // Workaround a flow limitation
     onScroll: () => {},
-    preserveScrollPosition: false,
-    options: {}
+    options: {
+      preserveScrollPosition: false
```
rgbkrk: We could probably move this into the `defaultOptions` below and just make the `defaultProps` here be an empty object, then instead of just `super(props)` we could do something like:

```
super({ ...props, {options: { ...props.options, defaultOptions }}})
```

Maybe we can't though because of the bound functions there. 🤔

rgbkrk (commented):
Woohoo! Thanks for tackling this, you went right after the biggest sore spot of this component!

packages/editor/src/index.js
captainsafia: OK. I moved the `preserveScrollOption` setting into default options.

I'm not sure what to do about the `onScroll` default prop.

For me, personally, the `super` thing looks a little bit hard to understand. It took me a while to figure out what was going on so maybe best to keep them as is for readability?

rgbkrk: ok, that's fine

We're not actually using onScroll (?) so maybe it's best to remove it completely, which I guess would remove preserveScrollPosition too. IIRC this came from the original component we forked.

packages/editor/src/index.js
```diff
@@ -83,45 +77,36 @@ class CodeMirrorEditor extends React.Component<
...
-      autoCloseBrackets: true,
-      mode: props.mode || props.language || "python",
-      lineNumbers: false,
-      matchBrackets: true,
-      theme: "composition",
-      autofocus: false,
-      hintOptions: {
-        hint: this.hint,
-        completeSingle: false, // In automatic autocomplete mode we don't want override
+    defaultOptions = Object.assign(
```
rgbkrk: This needs to be `this.defaultOptions` since it's used in other functions on this component.

rgbkrk: That or later usage of it needs to be trimmed out. Either way, this is making the tests fail at the moment.

captainsafia: Welp! This is my bad. I totally just pushed to the branch without running all the Travis tests locally. Lemme fix.

packages/core/src/providers/editor.js
```diff
@@ -24,9 +24,11 @@ type Props = {
...
 function mapStateToProps(state: Object, ownProps: Props): Object {
   return {
-    options: Object.assign(ownProps.options, {
-      cursorBlinkRate: state.config.get("cursorBlinkRate")
-    }),
+    options: ownProps.options
+      ? Object.assign(ownProps.options, {
+          cursorBlinkRate: state.config.get("cursorBlinkRate")
+        })
+      : { cursorBlinkRate: state.config.get("cursorBlinkRate") },
```
rgbkrk, line 29: I look forward to when we can have a generic codeMirrorConfig as part of our state.config so people can opt in to exactly the changes they'd like.

rgbkrk (approved):

packages/core/src/providers/editor.js
captainsafia: Yes!

Also there are a few problems with this now that we have the `ownProps` parameter on. Specifically, it freaks out because the props passed to the component from a a code cell are different from the ones passed to the component on a markdown cell and reconciling those differences gets annoying fast.

Working on separating the `Prop` type into a `MarkdownProp` and a `CodeProp` with the distinct requirements.

rgbkrk:
This is some pretty interesting flow error stuff and is fairly illustrative of an issue some of our provider components have. Flow is erroring on props that are intended for the unconnected component which we pass via `{...props}`. Many of these have `defaultProps` which should make them not have issues.

rgbkrk:
If I switch `ownProps` to the exported `CodeMirrorEditorProps`, all the errors drop away. _however_, they're completely missing the _actual_ props. Waaaaaait actually that _is_ the solution. The `ownProps` are the ones intended for the raw component. Commit incoming!

rgbkrk:
Ruh roh, looks like Travis is stuck on the app builds.

rgbkrk:
🎉

PR #2367, opened 2018-01-12
WIP: (Remote) Kernel launch epic

rgbkrk:
Working on setting up jupyter hosts as a kernel target -- this relies on both #2365 and #2366 being merged first. Still a WIP.

TODOs that I see so far:

* [x] Create an epic that launches jupyter server hosted kernels
* [x] Unify the action for when a kernel is launched to allow for desktop or jupyter websocket kernels (currently called `NEW_KERNEL`, should probably have a few action creators and one unifying action type)
* [ ] Hook this up to the jupyter extension
  * [ ] Configure the epic
  * [ ] Hardcode launching the kernel after the server starts (for now)
  * [ ] Initial Jupyter Config Data should get set in the `AppRecord`
* [ ] Test this

PR #2378, opened 2018-01-15
switch to launchKernel over newKernel naming

rgbkrk:
Making sure our naming is consistent for kernel launching, a re-visitation of #2367.


packages/core/__tests__/epics/comm-spec.js
```diff
@@ -87,11 +87,11 @@ describe("commActionObservable", () => {
       buffers: new Uint8Array()
     };
 
-    const newKernelAction = {
+    const launchKernelAction = {
```
rgbkrk: rename as `hokeyLaunchKernelAction` 😉

PR #2380, opened 2018-01-16
pull kernelDisplayName from the notebook metadata

rgbkrk:
Part of the ongoing remote kernel work. I was trying to figure out why our `AppRecord` needed kernel info since it was part of the document. This was the only place `app.kernelSpecDisplayName` was used.

* [x] Fix tests
* [x] Rebase against #2370, since that will use the explicit codemirror mode prop

### DIFF
--- a/packages/core/__tests__/providers/notebook-spec.js
+++ b/packages/core/__tests__/providers/notebook-spec.js
@@ -7,7 +7,10 @@ import { shallow, mount } from "enzyme";
 import renderer from "react-test-renderer";
 
 import { displayOrder, transforms } from "@nteract/transforms";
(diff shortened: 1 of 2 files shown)


rgbkrk:
Nice, this also got rid of a `$FlowFixMe` since we're _actually_ using the `ownProps` properly here.

rgbkrk:
Travis is currently in outage / backlog mode -- https://www.traviscistatus.com/incidents/c1s4dlyxd4lf

rgbkrk:
I'm going to err on the side of merging here since I've run tests and apps locally, as well as the fact that appveyor is passing.

PR #2385, opened 2018-01-17
switch to kernel record props

rgbkrk:
This is the kernel type unification of #2367. Primary benefits here are that the `LAUNCH_KERNEL_SUCCESSFUL` action emits either the jupyter websocket kernels ("remote") or local kernels ("zeromq"). I'm definitely seeing more bits to clean up in our app record, including what/where we have kernel status.

### DIFF
--- a/applications/desktop/src/notebook/reducers/app.js
+++ b/applications/desktop/src/notebook/reducers/app.js
@@ -52,7 +52,7 @@ function startSaving(state: AppRecord) {
 }
 
 function setExecutionState(state: AppRecord, action: SetExecutionStateAction) {
-  return state.setIn(["kernel", "status"], action.executionState);
+  return state.setIn(["kernel", "status"], action.kernelStatus);
 }
 
 function doneSaving(state: AppRecord) {
--- a/applications/desktop/src/notebook/reducers/app.js
+++ b/applications/desktop/src/notebook/reducers/app.js
@@ -25,29 +25,17 @@ function cleanupKernel(state: AppRecord): AppRecord {
   shutdownKernel(state.kernel);
 
   return state.withMutations((ctx: AppRecord) =>
-    ctx
-      .set("kernel", null)
-      .set("kernelSpecName", null)
-      .set("kernelSpecDisplayName", null)
-      .set("kernelSpec", null)
-      .set("executionState", "not connected")
+    ctx.set("kernel", null).set("executionState", "not connected")
   );
 }
 
 function launchKernel(state: AppRecord, action: NewKernelAction) {
-  const kernel = makeLocalKernelRecord({
-    channels: action.channels,
-    spawn: action.spawn,
-    connectionFile: action.connectionFile
-  });
+  const kernel = makeLocalKernelRecord(action.kernel);
 
   return cleanupKernel(state).withMutations((ctx: AppRecord) =>
-    ctx
-      .set("kernel", kernel)
-      .set("kernelSpecName", action.kernelSpecName)
-      .set("kernelSpecDisplayName", action.kernelSpec.spec.display_name)
-      .set("kernelSpec", action.kernelSpec)
-      .set("executionState", "starting")
+    // TODO: set the executionState inside the kernel (?)
+    //       was that what the status field was for?
+    ctx.set("kernel", kernel).set("executionState", "starting")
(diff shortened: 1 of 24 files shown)


rgbkrk:
Now that we've got CircleCI (and damn do our builds finish faster now), I think it's ok for us to just disable Travis until they're back up.

rgbkrk:
I've marked Travis CI to be inactive within the GitHub integrations for now.

rgbkrk:
Oh my. So much to adapt with `executionState`.

applications/desktop/src/notebook/reducers/app.js
```diff
@@ -25,37 +25,30 @@ function cleanupKernel(state: AppRecord): AppRecord {
...
 
   return cleanupKernel(state).withMutations((ctx: AppRecord) =>
-    ctx
-      .set("kernel", kernel)
-      .set("kernelSpecName", action.kernelSpecName)
-      .set("kernelSpecDisplayName", action.kernelSpec.spec.display_name)
-      .set("kernelSpec", action.kernelSpec)
-      .set("executionState", "starting")
+    // TODO: set the executionState inside the kernel (?)
+    //       was that what the status field was for?
```
theengineear: 👍 yep, that's what status was meant to be used for. i'll make a note to change that to `executionState`

packages/core/src/actionTypes.js
```diff
@@ -255,11 +257,7 @@ export type ChangeCellTypeAction = {
 export const LAUNCH_KERNEL_SUCCESSFUL = "LAUNCH_KERNEL_SUCCESSFUL";
 export type NewKernelAction = {
   type: "LAUNCH_KERNEL_SUCCESSFUL",
-  channels: Channels,
-  connectionFile: string,
-  spawn: ChildProcess,
-  kernelSpecName: string,
-  kernelSpec: Object
+  kernel: LocalKernelProps | RemoteKernelProps
```
theengineear, line 260: 👍

packages/types/src/core/hosts.js
```diff
@@ -78,6 +78,7 @@ export type BaseKernelProps = {
   kernelSpecName: ?string,
   lastActivity: ?Date,
   channels: ?rxjs$Subject<*, *>,
+  // TODO: Determine what stauses we'll set here
```
theengineear: See note about `executionState` above.

packages/core/src/actions.js
```diff
@@ -78,11 +80,13 @@ import type {
 
 import { createExecuteRequest } from "@nteract/messaging";
 
-export function launchKernelSuccessful(payload: *): NewKernelAction {
+export function launchKernelSuccessful(
+  kernel: LocalKernelProps | RemoteKernelProps
+): NewKernelAction {
   // TODO: Use our new kernel types instead of only matching the old setup
```
rgbkrk: I just noticed this TODO can be deleted.

packages/types/src/core/hosts.js
rgbkrk: We should leave a comment here outlining some of our states for this, we'll want to make proper constants at some point I think...

theengineear: Yah, definitely. I figured that was coming at some point.

rgbkrk:
Do you think I should switch all use of `executionState` to `kernelStatus` where components are using it?

theengineear:
>Do you think I should switch all use of executionState to kernelStatus where components are using it?

like the components are using the *variable name* `executionState`? If so, ya, we should switch to `kernelStatus` in this PR.

rgbkrk:
> like the components are using the *variable name* `executionState`? If so, ya, we should switch to `kernelStatus` in this PR.

Yeah and as one of their `props`.

rgbkrk:
Kernel restarting is failing, I've got a bit more to fix up here.

rgbkrk:
Switching kernels is a-ok though.

rgbkrk:
Restarting kernels should definitely be an action + epic pair in the future -- right now it's hidden away in our menu as two subsequent actions `killKernel` and `launchKernel`.

rgbkrk:
Ooook, this is good to go.

theengineear (approved):

PR #2391, opened 2018-01-17
create epic for launching a jupyter ws kernel

rgbkrk:
This sets up a kernel launching epic that launches kernels on jupyter hosts.

🎉  Connections to remote kernels, ready for desktop, play, and the nteract web app. Connecting this up to Desktop is not a focus for this PR, nor is the UI/config for doing so. The first target is the jupyter extension.

TODO:

* [x] Match semantics of the zeromq kernel launch
* [x] Adapt either the jupyter extension or `rx-jupyter` to adapt the URL based on `location.origin` as necessary
* [x] Integrate into the jupyter extension app
* [x] Tests

### DIFF
--- a/applications/desktop/src/notebook/reducers/app.js
+++ b/applications/desktop/src/notebook/reducers/app.js
@@ -33,6 +33,7 @@ function launchKernel(state: AppRecord, action: NewKernelAction) {
   }
   return cleanupKernel(state).set("kernel", kernel);
 }
+
 function exit(state: AppRecord) {
   return cleanupKernel(state);
 }
--- a/applications/jupyter-extension/nteract_on_jupyter/epics/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/epics/index.js
@@ -3,6 +3,22 @@ import { loadEpic } from "./contents";
 import { listKernelSpecsEpic } from "./kernelspecs";
 import { setNotebookEpic } from "./notebook";
 
-const epics = [loadEpic, listKernelSpecsEpic, setNotebookEpic];
+import {
+  executeCellEpic,
+  updateDisplayEpic,
+  commListenEpic,
+  launchWebSocketKernelEpic
+} from "@nteract/core/epics";
+
+// TODO: Bring desktop's wrapEpic over to @nteract/core so we can use it here
+const epics = [
+  executeCellEpic,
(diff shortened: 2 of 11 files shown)


rgbkrk:
One thing that `rx-jupyter` doesn't do that it probably should -- if it's not passed the full URL for the endpoint it should construct the full URL (especially for the websockets) using `location.origin` and the `baseURI`. Another slight annoyance is the different names we have for some of the same bits:

* `baseUrl` - our current jupyter extension base _`URI`_ (e.g. `/user/fvzhyd`)
* `serverUrl` - what we've set up for host records
* `serverConfig.endpoint || serverConfig.url` same thing in rx-jupyter, which is used as the full base URL `https://hub.example.org/user/fvzhyd` yet _can_ also function as the pure base path on a localhost server `/user/fvzhyd` but not in all cases (notably websockets)

Thoughts @theengineear? We mostly just need consistency among these while being able to straddle localhost (HTTP) and remote authenticated servers on HTTPS.

rgbkrk:
Notes from the jupyter notebook server as to why the jupyter extension used `baseURL` -- the server itself uses `baseURL` for that base URI.

theengineear:
@rgbkrk 

>if it's not passed the full URL for the endpoint it should construct the full URL

Is there any reason for us not to required that users of the lib just pass in the full URL, else we just bail in rx-jupyter?

rgbkrk:
> Is there any reason for us not to required that users of the lib just pass in the full URL, else we just bail in rx-jupyter?

I think that's a good standard.

rgbkrk:
Rebased this one to stay up to date, going to town on finishing it off now.

rgbkrk:
The way I'm running this right now (since we don't have the contents API and `SET_NOTEBOOK` integrations setup) is firing off this action with the redux devtools:

```js
{
  type: 'LAUNCH_KERNEL_BY_NAME',
  kernelSpecName: 'python3',
  cwd: '/'
}
```

rgbkrk:
:boom:

rgbkrk:
This _now_ also sets the kernel into the app record. Execution works as does autocomplete. 😄

rgbkrk:
execution

theengineear (approved):
🎉

PR #2392, opened 2018-01-18
Immutable redux

theengineear:
This PR does two things:

1. brings in `redux-immutable`, which recently allowed users to pass in a `getDefaultState` function to it's `combineReducers` that we can use as a `RecordFactory`
2. defines a new reducer-creation-function to create immutable maps of `ids: values`

This is sort-of just for discussion atm, but it sets us up to become fully immutable with our state.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -34,6 +34,7 @@
     "react-redux": "^5.0.6",
     "react-syntax-highlighter": "^6.0.0",
     "redux": "^3.7.2",
+    "redux-immutable": "^4.0.0",
     "rxjs": "^5.5.6",
     "uuid": "^3.1.0"
   },
--- a/packages/core/__tests__/reducers/createImmutableMapReducer-spec.js
+++ b/packages/core/__tests__/reducers/createImmutableMapReducer-spec.js
@@ -1,0 +1,74 @@
+import createImmutableMapReducer from "../../src/reducers/createImmutableMapReducer";
+import { Map, Record } from "immutable";
+import { combineReducers } from "redux-immutable";
+
+const appRecordFactory = Record({ resourcesById: Map() });
+const resourceRecordFactory = Record({ exists: false, name: "" });
+
+describe("createImmutableMapReducer", () => {
+  it("can be used with redux-immutable's combineReducers w/ Records", () => {
+    const exists = (state, action) => {
+      if (action.type === "RESOURCE/CREATE") {
+        return true;
+      }
+      return state;
+    };
+
+    const name = (state, action) => {
+      if (action.type === "RESOURCE/NAME") {
+        return action.name;
+      }
+      return state;
+    };
+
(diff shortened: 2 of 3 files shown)


theengineear:
@rgbkrk thoughts? I started trying to implement some of these `byRef` functions and realized that I would have some really awkward places where the `thingByRef` object would keep coming in as a Plain Javascript object and the underlying `thing` would be a `Record` 😿.

theengineear:
I wrote out the spec file as a sort-of POC that this should all work as expected.

packages/core/package.json
```diff
@@ -34,6 +34,7 @@
     "react-redux": "^5.0.6",
     "react-syntax-highlighter": "^6.0.0",
     "redux": "^3.7.2",
+    "redux-immutable": "^4.0.0",
```
rgbkrk, line 37: One minor issue I see is that redux-immutable's peerDep is not set to handle our range on immutable. You might have found it to be a non-issue though.

rgbkrk (approved):

rgbkrk:
I'm not really sure why you're thinking the byRef objects would sometimes be non immutable JS objects wouldn't we just create the records when they're first created and then work with them as immutable objects from there on?

rgbkrk:
Going ahead and merging as this can be started to be out into place on all the root reducers for each app.

theengineear:
>I'm not really sure why you're thinking the byRef objects would sometimes be non immutable JS objects wouldn't we just create the records when they're first created and then work with them as immutable objects from there on?

I could have been clearer above. My comment was just pointing out:

```
const resourceFactory = Immutable.Record({ name: "default" });
const resource = resourceFactory({ name: "some-name" });
const root = { byId: { someId: resource } };

console.log(Immutable.isImmutable(root))
// false
console.log(Immutable.isImmutable(root.byId));
// false
console.log(Immutable.isImmutable(root.byId.someId));
// true
```

I think we're on the same page. I just want to have the whole state tree be immutable. I think it's not actually so bad now that we've flattened the state. However, I was thinking about how `play` nests state, something like this:

```
{
    serversById: {
        serverId: {  // server envelope POJO
          server: {  // server object Record
              config: { .. }, // server config Record
              kernelsById: {  // ew, POJO
                  kernelId: {  // kernel envelope POJO
                      kernel: { .. } // Record
                  }
              }
          }
        }
    }
}
```

Again, since we flattened the state, we don't have any `byId`'s that are nested inside `Record`'s like this, so it's less of an issue. However, I'll still find it less confusing if our whole state tree is really in Immutable-land.

theengineear:
@rgbkrk ^^

theengineear:
Also, :tada: for `Immutable.isImmutable`!

rgbkrk:
Ah, cool yeah. Since we have the flattened structures we can use well typed `Record`s and `Ref`s liberally.

PR #2399, opened 2018-01-19
Add kernelspecs record

theengineear:
Please excuse any shortcomings wrt `flow` / `immutable`, I'm far from an expert at either technology. This is a stab at getting our typings set based on our plan doc in https://github.com/nteract/nteract/pull/2338

Also, just wondering about the pattern in general here. Should we have these types in a separate package? Or, should the `core` stuff be in core? What's the common convention?

### DIFF
--- a/packages/types/src/core/hosts.js
+++ b/packages/types/src/core/hosts.js
@@ -1,24 +1,20 @@
 // @flow
 import type { ChildProcess } from "child_process";
+import type { Id } from "./id";
+import type { RecordFactory, RecordOf } from "immutable";
+import type { Ref } from "./ref";
+import type { KernelspecsRef } from "./kernelspecs";
 
-// The Id type is for external resources, e.g. with /api/kernels/9092-7679-9978-8822,
-// 9092-7679-9978-8822 is the Id
-export opaque type Id = string;
-// A ref is an internal Id, used for kernels, hosts, kernelspec collections, etc.
-export opaque type Ref = string;
+import { Record, List } from "immutable";
 
 export opaque type KernelRef = Ref;
 export opaque type HostRef = Ref;
-export opaque type KernelSpecsRef = Ref;
-
-import type { RecordFactory, RecordOf } from "immutable";
-import { Record, List } from "immutable";
 
 export type BaseHostProps = {
   id: ?Id,
   ref: ?HostRef,
   selectedKernelRef: ?KernelRef,
-  kernelSpecsRef: ?KernelSpecsRef, // reference to a collection of kernelspecs
+  kernelspecsRef: ?KernelspecsRef, // reference to a collection of kernelspecs
   defaultKernelName: ?string,
   // In the desktop case, this _should_ be only one, pending
   // kernel cleanup
@@ -39,7 +35,7 @@ export const makeJupyterHostRecord: RecordFactory<
   ref: null,
   id: null,
   selectedKernelRef: null,
-  kernelSpecsRef: null,
+  kernelspecsRef: null,
   defaultKernelName: null,
   activeKernelRefs: List(),
   token: null,
@@ -65,7 +61,7 @@ export const makeDesktopHostRecord: RecordFactory<
   ref: null,
   id: null,
   selectedKernelRef: null,
-  kernelSpecsRef: null,
+  kernelspecsRef: null,
   defaultKernelName: null,
   activeKernelRefs: List()
 });
--- a/packages/types/src/core/id.js
+++ b/packages/types/src/core/id.js
@@ -1,0 +1,3 @@
+// The Id type is for external resources, e.g. with /api/kernels/9092-7679-9978-8822,
+// 9092-7679-9978-8822 is the Id
+export opaque type Id = string;
--- a/packages/types/src/core/kernelspecs.js
+++ b/packages/types/src/core/kernelspecs.js
@@ -1,3 +1,55 @@
+import type { RecordFactory, RecordOf } from "immutable";
 import type { Ref } from "./ref";
+import { List, Map, Record } from "immutable";
 
 export opaque type KernelspecsRef = Ref;
+
+type CommunicationKernelspecsProps = {
+  loading: boolean,
+  error: ?Object
+};
+
+export type CommunicationKernelspecsRecord = RecordOf<
+  CommunicationKernelspecsProps
+>;
+
+export const communicationKernelspecsFactory: RecordFactory<
+  CommunicationKernelspecsProps
+> = Record({
+  loading: false,
+  error: null
+});
+
+type EntitiesKernelspecsSpecProps = {
+  language: string,
+  argv: List<string>,
+  env: Map
+};
+
+export type EntitiesKernelspecsSpecRecord = RecordOf<
+  EntitiesKernelspecsSpecProps
+>;
+
+export const entitiesKernelspecsSpecFactory: RecordFactory<
+  EntitiesKernelspecsSpecProps
+> = Record({
+  language: "", // TODO: what's a reasonable default?
+  argv: List(),
+  env: Map()
+});
+
+type EntitiesKernelspecsProps = {
+  name: string,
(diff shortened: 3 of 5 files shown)


packages/types/src/core/kernelspecs.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export type EntitiesKernelspecsSpecRecord = RecordOf<
+  EntitiesKernelspecsSpecProps
+>;
+
+export const entitiesKernelspecsSpecFactory: RecordFactory<
+  EntitiesKernelspecsSpecProps
+> = Record({
+  language: "", // TODO: what's a reasonable default?
```
theengineear, line 36: Ping >> thoughts?

packages/types/src/core/kernelspecs.js
```diff
@@ -0,0 +1,55 @@
...
+  resources: Map,
+  spec: RecordOf<EntitiesKernelspecsSpecProps>
+};
+
+export type EntitiesKernelspecsRecord = RecordOf<EntitiesKernelspecsProps>;
+
+export const entitiesKernelspecsFactory: RecordFactory<
+  EntitiesKernelspecsProps
+> = Record({
+  name: "", // TODO: what's a reasonable default?
```
theengineear, line 52: Ping >> thoughts?

theengineear:
Whoops. Broke some stuff. I'll ping when it's fixed.

theengineear:
Fixed.

rgbkrk:
> Should we have these types in a separate package? Or, should the core stuff be in core? What's the common convention?

My personal take on this is that the records should be in core. The `@nteract/types` package was created for us to unify a lot of the types that were being used all over the place where there's a general type (jupyter messages, notebook format bits, ...). I think it makes sense to actually move the records bits from `@nteract/types` to `@nteract/core`. The raw kernelspecs structure (as exists on disk) is probably wise inside of `@nteract/types`.

packages/types/src/core/kernelspecs.js
rgbkrk: `null`

This is twofold as to why:

1. It gets loaded from disk or API, so we can't guarantee that they set this key
2. This one has a big impact since the `name` is supposed to be unique

packages/types/src/core/kernelspecs.js
rgbkrk: `null` here too as the default

packages/types/src/core/id.js
```diff
@@ -0,0 +1,3 @@
+// The Id type is for external resources, e.g. with /api/kernels/9092-7679-9978-8822,
+// 9092-7679-9978-8822 is the Id
+export opaque type Id = string;
```
rgbkrk, line 3: 👍

packages/types/src/core/kernelspecs.js
```diff
@@ -0,0 +1,55 @@
+import type { RecordFactory, RecordOf } from "immutable";
+import type { Ref } from "./ref";
+import { List, Map, Record } from "immutable";
+
+export opaque type KernelspecsRef = Ref;
```
rgbkrk, line 5: 👍

packages/types/src/core/hosts.js
```diff
@@ -39,7 +35,7 @@ export const makeJupyterHostRecord: RecordFactory<
   ref: null,
   id: null,
   selectedKernelRef: null,
-  kernelSpecsRef: null,
+  kernelspecsRef: null,
```
rgbkrk, line 38: thanks for making the case consistent

packages/types/src/core/kernelspecs.js
rgbkrk: `null`

packages/types/src/core/kernelspecs.js
rgbkrk: `null`

theengineear:
>I think it makes sense to actually move the records bits from @nteract/types to @nteract/core

👍

packages/types/src/core/kernelspecs.js
rgbkrk: I mean, unless we do validation before creating a "real" record.

packages/types/src/core/hosts.js
theengineear: Yup. I did the same in the plan.md doc ;)

I'm all about consistency 😺 https://github.com/nteract/nteract/issues/2400

packages/types/src/core/kernelspecs.js
```diff
@@ -0,0 +1,55 @@
...
+type CommunicationKernelspecsProps = {
+  loading: boolean,
+  error: ?Object
+};
+
+export type CommunicationKernelspecsRecord = RecordOf<
+  CommunicationKernelspecsProps
+>;
+
+export const communicationKernelspecsFactory: RecordFactory<
```
theengineear, line 16: It's just occurring to me know that we should probably be capitalizing these? `CommunicationKernelspecsFactory`?

theengineear: Additionally, we have the pattern `make*` for factories in other places here. I think that that is a little confusing. The Immutable pattern seems to be `RecordFactory`, so  think `*Factory` makes sense? Then, we end up using just like `Map`:

```
const map = Map();
const record = CommunicationKernelspecsFactory();
```

Ya?

theengineear: @rgbkrk ^^

packages/types/src/core/kernelspecs.js
```diff
@@ -0,0 +1,55 @@
...
+> = Record({
+  loading: false,
+  error: null
+});
+
+type EntitiesKernelspecsSpecProps = {
+  language: string,
+  argv: List<string>,
+  env: Map
+};
```
rgbkrk, line 27: Could we just call this one `KernelspecProps`?

packages/types/src/core/kernelspecs.js
rgbkrk: The previous casing standard came from the [Record.Factory docs](http://facebook.github.io/immutable-js/docs/#/Record.Factory), particularly the Flow typing records section.

rgbkrk:
I'm going to go ahead and merge so you can keep on heading forward, though I'd love simpler names for `EntitiesKernelspecsSpecProps`, especially considering it exactly represents an actual kernelspec (of the jupyter sense).

packages/types/src/core/kernelspecs.js
theengineear: Gotcha! Cool. If that's what the docs say, then we should go with that. Thanks!

packages/types/src/core/kernelspecs.js
theengineear: Yah, that's fine by me 😄. What does the `s` vs no `s` mean to you? Like, why do we call the 
object:

```
{
  name: " .. ",
  resources: { .. },
  spec: { .. }
}
```

`kernelspecs` (with an `s`)?

theengineear: This sorta feels like it should be `KernelspecsProps` (with that `s`) for consistency?

theengineear: Oop, nvm, I get it, the loss of the `s` is that you've got `kernelspecs.spec`, so `kernelspecs` is the whole object and `kernelspecs.spec` == `kernelspec`.

theengineear: That's honestly a little confusing to me, but I'll go for it since it's a much shorter name :+1:

rgbkrk: kernelspecs is an array of `kernelspec` which sadly includes a key called `spec`. This comes entirely from the [Jupyter nomenclature for kernelspec](https://ipython.org/ipython-doc/3/development/kernels.html#kernel-specs). As long as we're consistent and it's easy for us to read, I'm happy.

PR #2401, opened 2018-01-20
Add core kernelspecs

theengineear:
Adds redux stuff for kernelspecs.

TODO:

- [x] action type strings
- [x] action creator types
- [x] action creators
- [x] reducers
- [x] epic
- [x] tests
- [x] rewire jupyter extension to point to new kernelspecs
- [x] delete old kernelspecs code


theengineear:
Definitely just a work in progress at this point, but I wanted to throw this partial PR up here to make sure this is all looking like we *want* it to look.

theengineear:
@rgbkrk mind doing a quick peruse? Anything set off 🚨s for you?

theengineear:
The only new commit is 4b7c9e0329491f15bc9f625d567e09f7714132f9

packages/core/src/actionTypes.js
```diff
@@ -24,6 +26,43 @@ import type { ExecuteRequest } from "@nteract/types/messaging";
 
 import type { Output, StreamOutput } from "@nteract/commutable/src/v4";
 
+export const FETCH_KERNELSPECS = "CORE/FETCH_KERNELSPECS";
```
rgbkrk, line 31: A+ to new style being introduced as part of new actions.

rgbkrk:
Merged #2399, so you can rebase this.

theengineear:
Rebased on top of some edits based off of your previous PR review. Going to start adding epics in now.

theengineear:
@rgbkrk Can you point me to a place in the codebase where we're mocking out rx-jupyter for testing epics? We do have to mock that out for testing, right? I've set up my own ways to mock out things like `ajax` and provide mock schedulers (so you can test things with debounceTime, etc), but I'm wondering how that's handled in here?

rgbkrk:
> I've set up my own ways to mock out things like ajax and provide mock schedulers (so you can test things with debounceTime, etc), but I'm wondering how that's handled in here?

In https://github.com/nteract/nteract/pull/2391/files#diff-bf4f11387916efdb1656d04751fe4e5a, I created `__mocks__` for `rx-jupyter`, only for `@nteract/core`. I only "implemented" `kernels` so you'll have to add on `kernelspecs`. I'd be cool with mocking `ajax` too, I figured that can be tested inside of `rx-jupyter` though.

theengineear:
@rgbkrk just missing tests at this point. Want to do a review?

rgbkrk:
Rockin', will do.

applications/jupyter-extension/nteract_on_jupyter/index.js
```diff
@@ -21,6 +23,7 @@ function createApp(jupyterConfigData: JupyterConfigData) {
 
     componentDidMount(): void {
       store.dispatch({ type: "LOAD", path: jupyterConfigData.contentsPath });
+      store.dispatch(fetchKernelspecs({ kernelspecsRef: "single-server" }));
```
rgbkrk: I wonder if we should leave commentary in these places about why we have this ref and what the function of the current "single-server" literal is for.

theengineear: Oh, yah, good idea. 👍

packages/core/src/reducers/kernelspecs.js
```diff
@@ -0,0 +1,62 @@
...
+
+// Entities state.
+const entitiesKernelspecs = (state = Immutable.Map(), action) => {
+  switch (action.type) {
+    case actionTypes.FETCH_KERNELSPECS_FULFILLED:
+      let newState = Immutable.Map();
+      Object.keys(action.payload.kernelspecs).forEach(key => {
+        const value = action.payload.kernelspecs[key];
+        newState = newState.set(key, makeKernelspec(value));
+      });
```
rgbkrk: Since you're using `.set` anyways here, I'd probably just do this:


```js
return Immutable.Map(action.payload.kernelspecs).map(
  (value: KernelspecProps) => makeKernelspec(value)
);
```

rgbkrk: Whoops, value comes first. Make sure to switch that up if you read that before I edited it.

theengineear: Almost! I forgot about how Immutable switches the expected order of `key, value` to `value, key`! Helpful in this case though:

```
const newState = Immutable.Map(action.payload.kernelspecs).map(
  (value: KernelspecProps) => makeKernelspec(value)
);
```

theengineear:
Hmm. Any opinions on how much of the state-management chain to test at once? I sorta feel like core's contract is:

    Give us actions, we update state.

But, when you test epics, you only test:

    Give us actions, we give you actions.

And when you test reducers, you only test:

    Give us partial-state and actions, we give you partial-state.

I.e., I feel like it's not that much more work to just configure a complete store, setup reducers/epics, dispatch actions and look for the resulting state? I'm not going to do this now, but maybe food for thought? I don't want us to get in the habit of writing tests that don't actually prevent breaking changes.

theengineear:
The test I just wrote really only tests actions in >> actions out, but that's not really the thing consumers care about at the end of the day.

rgbkrk:
I would say that something which does the full setup should be part of a functional testing suite rather than part of our unit tests.

rgbkrk:
And then... if we really wanted to we could be doing integration tests backed by binder or a launched server. 😉

theengineear:
@rgbkrk hey, sorry, i ended up rebasing again to update this pr in response to some comments.

rgbkrk:
No probs, rebases are my jam.

packages/core/__mocks__/rx-jupyter.js
```diff
@@ -1,9 +1,55 @@
 import { of } from "rxjs/observable/of";
 import { Subject } from "rxjs/Subject";
 
+// Mock response generated 01/24/2018, may need updating.
+const mockListReponse = {
```
rgbkrk, line 5: Looking good to me.

rgbkrk (approved):

Issue #2394, opened 2018-01-19
Notebooks become unresponsive


falbarelli:
I am using nteract 0.5.5 on macOS High Sierra.
I noticed that very often notebooks become unresponsive, i.e. I scroll but the notebook window is almost frozen. I use notebooks with some figures, but not big at all...
Does it happen to anybody else?

rgbkrk:
We had a regression in the last release that caused performance issues on notebooks, even on smaller sizes. We're looking to get a new release out soon.

rgbkrk:
We did get a release out and we've got another coming up that has optimized rendering a bit more.

Issue #2362, opened 2018-01-11
Optimize performance of our components


rgbkrk:
This is especially important for big notebooks, though it can be noticed with smaller ones as well. Here's some super light editing of a notebook in notebook-on-next, using React devtools with "Highlight Updates" checked:

![highlight-updates](https://user-images.githubusercontent.com/836375/34843855-8fdb8498-f6c4-11e7-8a32-6f30753015b7.gif)

Notice how changing one cell is causing a re-render of _all_ the cells. That should not be the case with our components. I _think_ this is because the `cellMap` is changing (as our top level prop), so then all the children have to re-render (from the `<NotebookApp/>` component).

I'm a total novice on performance optimizations here. There are components that we could actually write a custom `shouldComponentUpdate` for (or even switch some to `PureComponent`). _This_ however looks like an architectural issue possibly (?). Perhaps it's just that the cells themselves should have their own `shouldComponentUpdate` (or the components inside).

lgeiger:
As an very simple starting point we might want to enable the [`jsx-no-bind` ESLINT rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md):
> A bind call or arrow function in a JSX prop will create a brand new function on every single render. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary. It may also cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.

lgeiger:
Here's a nice little tool that logs potential unnecessary renders: https://github.com/maicki/why-did-you-update

rgbkrk:
At first I was like "that's silly, we don't do any binding anywhere that's passed through". Then I realized that all our cells are passing in arrow functions that call to dispatch with their ID... 🤔

rgbkrk:
It turned out to be much more than the bound functions. Because the `<NotebookApp />` had `cellMap` as a prop, literally any change to any cell was causing the whole tree to render because the root was. I'm looking forward to finishing this up in #2402, it's working _really_ well now!

rgbkrk:
I think we can declare this one as iteratively improving, though it would be super neat if we measured performance on CI someday.

PR #2402, opened 2018-01-21
Notebook App rendering performance improvements

rgbkrk:
Mitigates issues #2394 and #2362 (and likely others). This PR does some performance optimizations of the app itself by making a provider component for a cell.

master | PR
----|---
master | pr

I've got some cleanup to do here. Generally I think this could be a lot better. One idea I've had is to move the current global level keybindings (for running / focusing cells) to a separate keybinding component or even just deferring them to our codemirror component.

### DIFF
--- a/applications/notebook-on-next/pages/edit.js
+++ b/applications/notebook-on-next/pages/edit.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from "react";
 import fetch from "isomorphic-fetch";
-import { emptyNotebook, fromJS } from "@nteract/commutable";
+import { emptyNotebook, fromJS, toJS } from "@nteract/commutable";
 import { NotebookApp } from "@nteract/core/providers";
 import { Provider } from "react-redux";
 import { List as ImmutableList, Map as ImmutableMap } from "immutable";
@@ -12,6 +12,7 @@ const store = configureStore();
 
 async function fetchFromGist(gistId): ?Object {
   const path = `https://api.github.com/gists/${gistId}`;
+
   return fetch(path)
     .then(async response => {
       if (response.status !== 200) return null;
@@ -26,7 +27,7 @@ async function fetchFromGist(gistId): ?Object {
         }
       }
     })
-    .catch(err => null);
+    .catch(err => toJS(emptyNotebook));
 }
 
 const Error = () => (
--- a/packages/core/__tests__/epics/execute-spec.js
+++ b/packages/core/__tests__/epics/execute-spec.js
@@ -4,6 +4,7 @@ import {
   SEND_EXECUTE_REQUEST,
   ABORT_EXECUTION,
   ERROR_EXECUTING,
+  EXECUTE_CELL,
   CLEAR_OUTPUTS,
   UPDATE_CELL_STATUS,
   UPDATE_DISPLAY,
@@ -28,10 +29,9 @@ import { toArray, share, catchError, bufferCount } from "rxjs/operators";
 
 describe("executeCell", () => {
   test("returns an executeCell action", () => {
-    expect(executeCell("0-0-0-0", "import random; random.random()")).toEqual({
-      type: SEND_EXECUTE_REQUEST,
-      id: "0-0-0-0",
-      message: expect.any(Object)
+    expect(executeCell("0-0-0-0")).toEqual({
+      type: EXECUTE_CELL,
+      id: "0-0-0-0"
     });
   });
 });
@@ -64,7 +64,22 @@ describe("createExecuteCellStream", () => {
             status: "not connected"
           },
           notificationSystem: { addNotification: jest.fn() }
-        }
+        },
+        document: Immutable.fromJS({
+          notebook: {
+            cellMap: {
+              first: {
+                source: "woo",
+                cell_type: "code"
+              },
+              second: {
+                source: "eh",
+                cell_type: "code"
+              }
+            },
+            cellOrder: ["first", "second"]
+          }
+        })
       }
     };
     const action$ = ActionsObservable.of({ type: SEND_EXECUTE_REQUEST });
@@ -149,7 +164,7 @@ describe("executeCellEpic", () => {
   test("Errors on a bad action", done => {
     // Make one hot action
     const badAction$ = ActionsObservable.of({
-      type: SEND_EXECUTE_REQUEST
+      type: EXECUTE_CELL
     }).pipe(share());
     const responseActions = executeCellEpic(badAction$, store).pipe(
       catchError(error => {
@@ -181,14 +196,42 @@ describe("executeCellEpic", () => {
       err => done.fail(err)
     );
   });
-  test("Informs about disconnected kernels, allows reconnection", done => {
-    const action$ = ActionsObservable.of(executeCell("id", "source")).pipe(
-      share()
-    );
+  test.only("Informs about disconnected kernels, allows reconnection", done => {
+    const store = {
+      getState() {
+        return this.state;
+      },
+      state: {
+        app: {
+          kernel: {
+            status: "not connected"
+          },
+          channels: false,
+          notificationSystem: { addNotification: jest.fn() }
+        },
+        document: Immutable.fromJS({
+          notebook: {
+            cellMap: {
+              first: {
+                source: "woo",
+                cell_type: "code"
+              },
+              second: {
+                source: "eh",
+                cell_type: "code"
+              }
+            },
+            cellOrder: ["first", "second"]
+          }
+        })
+      }
+    };
+
+    const action$ = ActionsObservable.of(executeCell("first")).pipe(share());
     const responseActions = executeCellEpic(action$, store);
     responseActions.subscribe(
       x => {
-        expect(x.payload.toString()).toEqual("Error: kernel not connected");
+        expect(x.payload.toString()).toEqual("Kernel not connected!");
         done();
       },
       err => done.fail(err)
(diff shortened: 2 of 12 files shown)


applications/notebook-on-next/pages/edit.js
```diff
@@ -12,6 +12,9 @@ const store = configureStore();
 
 async function fetchFromGist(gistId): ?Object {
   const path = `https://api.github.com/gists/${gistId}`;
+  return toJS(emptyNotebook);
```
rgbkrk: Reminder to myself that I have to return this file back to how I found it. 😉

packages/core/src/actionTypes.js
```diff
@@ -124,6 +124,11 @@ export type SendExecuteMessageAction = {
   message: ExecuteRequest
 };
 
+export const EXECUTE_CELL = "EXECUTE_CELL";
+export type ExecuteCellAction = {
+  id: CellID
+};
```
rgbkrk: This brings _back_ `EXECUTE_CELL`, which is requesting that we execute a cell with a given ID. This helps us not have to have the cells and their contents bound to actions that will cause cell execution -- they should only have to have the cell ID.

packages/core/src/epics/execute.js
```diff
@@ -177,11 +179,19 @@ export function executeCellEpic(action$: ActionsObservable<*>, store: any) {
...
       cellActionStream.pipe(
-        // When a new SEND_EXECUTE_REQUEST comes in with the current ID, we create a
+        // When a new EXECUTE_CELL comes in with the current ID, we create a
         // a new stream and unsubscribe from the old one.
-        switchMap(({ message, id }) =>
-          createExecuteCellStream(action$, store, message, id)
-        )
+        switchMap(({ id }) => {
+          const state = store.getState();
+          const source = state.getIn(["document", "cellMap", id, "source"], "");
```
rgbkrk: I should probably add a comment in here that we create create the execute request based on the source of the ID from the `EXECUTE_CELL` action.

packages/core/src/providers/notebook-app.js
```diff
@@ -100,14 +326,8 @@ const mapStateToProps = (state: Object, ownProps: Props): Props => {
...
     lastSaved: state.app.get("lastSaved"),
     cellOrder: state.document.getIn(["notebook", "cellOrder"], ImmutableList()),
-    cellMap: state.document.getIn(["notebook", "cellMap"], ImmutableMap()),
-    transient: state.document.get("transient"),
-    cellPagers: state.document.get("cellPagers", ImmutableMap()),
-    cellFocused: state.document.get("cellFocused"),
-    editorFocused: state.document.get("editorFocused"),
     stickyCells: state.document.get("stickyCells"),
     kernelStatus: state.app.getIn(["kernel", "status"], "not connected"),
-    models: state.comms.get("models"),
```
rgbkrk, line 110: The more I was able to take out of our top level component here, the better it got.

packages/core/src/providers/notebook-app.js
```diff
@@ -178,27 +397,28 @@ export class NotebookApp extends React.Component<Props> {
...
+    return;
 
-    if (!cell) {
-      return;
-    }
+    /*
+    const id = this.props.cellFocused;
 
     if (e.shiftKey && !this.props.stickyCells.has(id)) {
+      // Couldn't focusNextCell just do it, no need to have this here
```
rgbkrk: Clarification on this comment, as well as a todo for me:

There should be a way to dispatch something like `focusTheCellAfterTheCurrentFocusedCell`

packages/core/src/providers/notebook-app.js
```diff
@@ -178,27 +397,28 @@ export class NotebookApp extends React.Component<Props> {
...
       this.context.store.dispatch(focusNextCell(this.props.cellFocused, true));
       this.context.store.dispatch(focusNextCellEditor(id));
     }
 
-    if (cell.get("cell_type") === "code") {
-      this.context.store.dispatch(executeCell(id, cell.get("source")));
-    }
+    // Could become "executeFocusedCell"
+    this.context.store.dispatch(executeCell(id));
```
rgbkrk: Just the same, we should be able to run `executeFocusedCell`.

packages/core/src/providers/notebook-app.js
```diff
@@ -178,27 +397,28 @@ export class NotebookApp extends React.Component<Props> {
...
-    const cellMap = this.props.cellMap;
-    const id = this.props.cellFocused;
-    const cell = cellMap.get(id);
+    return;
 
-    if (!cell) {
-      return;
-    }
+    /*
+    const id = this.props.cellFocused;
```
rgbkrk: Can't merge until the functionality here is brought back. I tore it out to reduce how much information the top level notebook was getting so I could get the nice update highlighting. 😄

packages/core/src/providers/notebook-app.js
rgbkrk: Thinking about this one in particular, I think the transforms could/should opt in to the models via `contextTypes`. Only model-debug is using this currently, so I think that's a fine assumption while we carve this out.

rgbkrk:
Alright! This should now pass and is ready for review!

packages/core/__tests__/epics/execute-spec.js
```diff
@@ -181,14 +196,42 @@ describe("executeCellEpic", () => {
       err => done.fail(err)
     );
   });
-  test("Informs about disconnected kernels, allows reconnection", done => {
-    const action$ = ActionsObservable.of(executeCell("id", "source")).pipe(
-      share()
-    );
+  test.only("Informs about disconnected kernels, allows reconnection", done => {
```
theengineear: ⚡️ 💀

packages/core/src/components/toolbar.js
```diff
@@ -21,7 +21,7 @@ import {
   TriangleRightOcticon
 } from "@nteract/octicons";
 
-declare type ToolbarProps = {|
+export type ToolbarProps = {|
```
theengineear, line 24: Does `declare` make the type available globally or something?

packages/core/src/epics/execute.js
```diff
@@ -166,7 +180,18 @@ export function createExecuteCellStream(
...
 export function executeCellEpic(action$: ActionsObservable<*>, store: any) {
   return action$.pipe(
-    ofType(SEND_EXECUTE_REQUEST),
+    ofType(EXECUTE_CELL, EXECUTE_FOCUSED_CELL),
+    mergeMap(action => {
+      if (action.type === EXECUTE_FOCUSED_CELL) {
+        const state = store.getState();
+        const id = state.document.get("cellFocused");
+        if (!id) {
+          throw new Error("attempted to execute without an id");
```
theengineear, line 189: If this just throws in here, why do we allow `id?: string` in the type defs?

packages/core/src/providers/notebook-app.js
```diff
@@ -43,30 +43,254 @@ import {
...
+
+    const cellFocused = this.props.cellFocused;
+    const editorFocused = this.props.editorFocused;
+
+    let element = null;
+
+    switch (cellType) {
+      case "code":
+        element = (
+          <React.Fragment>
```
theengineear, line 160: 👌

packages/core/src/providers/toolbar.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import React, { Component } from "react";
+import * as React from "react";
```
theengineear, line 2: 🐄 isn't the canonical import `import React from "react"`?

packages/core/src/providers/toolbar.js
```diff
@@ -60,7 +58,7 @@ class Toolbar extends Component<Props> {
 
   executeCell(): void {
     const { dispatch } = this.props;
-    dispatch(executeCell(this.props.id, this.props.cell.get("source")));
```
theengineear, line 63: 🎉

theengineear (changes requested):
Just un-skip that test. Otherwise, looking great!

packages/core/src/providers/toolbar.js
rgbkrk: From the [flow docs on React](https://flow.org/en/docs/react/components/)

> Note: We import React as a namespace here with import * as React from 'react' instead of as a default with import React from 'react'. When importing React as an ES module you may use either style, but importing as a namespace gives you access to React’s utility types.

Most of the code base is this way as well and occurred on some combo of React 16 and a more recent flow release.

packages/core/src/epics/execute.js
rgbkrk: The `EXECUTE_FOCUSED_CELL` action has no id, and relies on what's in the store for the current focused cell.  If for some reason a cell is not focused and they hit shift-enter, I figured this should be thrown (and caught in the epic as `catchError` (not in the diff here, it's in the epic much lower).

packages/core/src/components/toolbar.js
rgbkrk: You know, I'm not sure. I just switched this to export so I could explicitly use it in the provider component.

packages/core/__tests__/epics/execute-spec.js
rgbkrk: Whoops!

rgbkrk:
Fixed up the tests, now to watch CI run. 👁 👁

theengineear (approved):

packages/core/src/providers/toolbar.js
theengineear: ah, awesome. thanks! that's really good to know. i'm positive that saved me like at least 10 minutes down the road somewhere 😛

PR #2405, opened 2018-01-22
fix: post-pr updates

theengineear:
The PR got merged to keep things moving. Just doing some clean up based
on a review.

See https://github.com/nteract/nteract/pull/2399 for context.


rgbkrk (approved):

rgbkrk:
I've viewed `kernelspecs` as the entire list of all the individual `kernelspec`. Technically, the `spec` is the actual kernel spec. However, the Jupyter Notebook API has an "envelope" around this that has the name and resources.

PR #2410, opened 2018-01-24
fix: fix flow typing (missing // @flow)

theengineear:
Missed some `// @flow` annotations :(

We also figured out how to actually handle the `opaque` types in here. There's a bit of trickiness to get them to work properly. _Within_ the file, the types can be used as `string`, but outside the files (i.e., when you import them), they are no longer `string`, but actually a new type that you've defined.

### DIFF
--- a/packages/types/src/core/refs.js
+++ b/packages/types/src/core/refs.js
@@ -1,6 +1,10 @@
 // @flow
 import { v4 as uuid } from "uuid";
 
+// Note that within this file, these can all be used as `string`.
+// However, _outside_ this file, they are no longer `string`, but actually the
+// opaque types that we've set.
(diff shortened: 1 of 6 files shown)


packages/types/src/core/refs.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import { v4 as uuid } from "uuid";
+
+// Note that within this file, these can all be used as `string`.
+// However, _outside_ this file, they are no longer `string`, but actually the
+// opaque types that we've set.
+// See https://flow.org/en/docs/types/opaque-types/#toc-within-the-defining-file
```
rgbkrk, line 7: Awesome description and link back.

rgbkrk (approved):

Issue #2371, opened 2018-01-13
YError when starting nteract 0.5.5 on Ubuntu 17.10


bgodard:
When starting nteract 0.5.5 on Ubuntu 17.10, a popup window appear in the top right corner of the notebook with the message "ERROR" and an error message is displayed in the console (see below).

The notebook seems otherwise usable.

This happens both with AppImage and Debian package.
This does not happen with nteract version 0.4.3.

> $ nteract 
> YError: Invalid first argument. Expected boolean or string but received function.
>     at i (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:239215)
>     at c.optional.forEach.e (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:240072)
>     at Array.forEach (native)
>     at e.exports (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:239972)
>     at Object.S.version (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:916317)
>     at Object.<anonymous> (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:255879)
>     at t (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:256)
>     at /opt/nteract/resources/app.asar/lib/webpacked-main.js:1:606
>     at Object.<anonymous> (/opt/nteract/resources/app.asar/lib/webpacked-main.js:1:617)
>     at Object.<anonymous> (/opt/nteract/resources/app.asar/lib/webpacked-main.js:2:3)
> [10:46:29.260] [info] args { _: [],
>   version: false,
>   v: false,
>   help: false,
>   h: false,
>   kernel: 'python3',
>   k: 'python3',
>   '$0': '/opt/nteract/nteract',
>   verbose: undefined }
> [10:46:29.967] [warn] APPIMAGE env is not defined, current application is not an AppImage
> [10:46:30.230] [info] launching an empty notebook by default
> 

![nteract-error](https://user-images.githubusercontent.com/8047509/34906351-c40ed9fe-f863-11e7-9cbd-bcea87aaa0f1.png)

atbug:
duplicate of #2346

jkornblum:
duplicate #2369

Issue #2369, opened 2018-01-12
Installation and launch issues on Ubuntu 17.10


jkornblum:
I followed the link on nteract.io for Linux download.
I executed the AppImage and opted to save icon/info to application tray.
I noticed the following issues,
- I received an exception when I executed the app image
- There is no icon when executing AppImage (maybe this is normal?)
- The nteract icon in application tray does not launch program says cannot find file in `/opt/nteract`

```
YError: Invalid first argument. Expected boolean or string but received function.
    at i (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:239215)
    at c.optional.forEach.e (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:240072)
    at Array.forEach (native)
    at e.exports (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:239972)
    at Object.S.version (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:916317)
    at Object.<anonymous> (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:255879)
    at t (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:256)
    at /tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:606
    at Object.<anonymous> (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:1:617)
    at Object.<anonymous> (/tmp/.mount_nterac6Em50n/app/resources/app.asar/lib/webpacked-main.js:2:3)
```

![nteract-install-error-ubuntu17-10](https://user-images.githubusercontent.com/5368903/34893841-968548a2-f7a4-11e7-95bc-e739cc77fe08.png)
![nteract-failed-launch-ubuntu17-10](https://user-images.githubusercontent.com/5368903/34893842-98735ae6-f7a4-11e7-8ffb-c26350ff7c56.png)

PR #2412, opened 2018-01-24
fix(index): fix YERROR in notebook index.js

jkornblum:
The yargs.version() method was being sent a function when it needed to be sent result of that function which is the version.

This addresses part of #2369  
This addresses #2371

### DIFF
--- a/applications/desktop/src/main/index.js
+++ b/applications/desktop/src/main/index.js
@@ -43,7 +43,7 @@ const path = require("path");
 
 const yargs = require("yargs/yargs");
 const argv = yargs()
-  .version(() => require("./../../package.json").version)
+  .version((() => require("./../../package.json").version)())
   .usage("Usage: nteract <notebooks> [options]")
   .example("nteract notebook1.ipynb notebook2.ipynb", "Open notebooks")
   .example("nteract --kernel javascript", "Launch a kernel")


applications/desktop/src/main/index.js
```diff
@@ -43,7 +43,7 @@ const path = require("path");
 
 const yargs = require("yargs/yargs");
 const argv = yargs()
-  .version(() => require("./../../package.json").version)
+  .version((() => require("./../../package.json").version)())
```
rgbkrk, line 46: This could just be:

```javascript
require("./../../package.json").version
```

at this point now

rgbkrk:
Thanks for tackling this!

applications/desktop/src/main/index.js
jkornblum: That's the approach I would have gone for.

I googled it thinking the function was there for a reason and found some posts about using 
`require("./../../package.json").version` could expose the entire package.json in bundled code so I thought maybe the function was there for that. I have no idea how webpack/bundling works under the hood so I just left it 😄 .

Cheers

rgbkrk: Ok, cool, I'll just merge this in. No worries if the package.json ends up in our bundle, it's teeny compared to everything else.

PR #2414, opened 2018-01-24
hook up app reducer + epics to notebook-on-next

rgbkrk:
This hooks up execution / kernel launch logic to notebook-on-next.

We're a very short amount of work away from a full binder backed app on the web here.

### DIFF
--- a/applications/notebook-on-next/pages/edit.js
+++ b/applications/notebook-on-next/pages/edit.js
@@ -48,11 +48,25 @@ export default class Edit extends React.Component<*> {
   static async getInitialProps(context: Object) {
     const query = context.query;
     const isServer = context.isServer;
-    const serverNotebook = await fetchFromGist(query.gistid);
-    if (!serverNotebook) return {};
+
+    let serverNotebook;
+    if (query.gistid) {
+      serverNotebook = await fetchFromGist(query.gistid);
+    }
+
+    if (!serverNotebook) {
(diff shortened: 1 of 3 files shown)


applications/notebook-on-next/pages/edit.js
```diff
@@ -48,11 +48,25 @@ export default class Edit extends React.Component<*> {
...
+
+    if (!serverNotebook) {
+      // TODO: make this a notebook with one cell
+      serverNotebook = toJS(emptyNotebook);
+    }
+
     store.dispatch({
+      // TODO: This action _will_ set off a kernel kick off with our current
+      //       setup so as we start activating servers and kernels we'll want
+      //       to change how we do "SET_NOTEBOOK"
```
betatim, line 65: We contact mybinder.org from the server because `SET_NOTEBOOK` is executed there? And then again from the client where we also do `SET_NOTEBOOK`?

betatim:
This mostly brings in stuff that is already in use elsewhere (I think). So 👍

betatim (approved):

applications/notebook-on-next/pages/edit.js
rgbkrk: Not yet! My TODO here is more of a ":warning: once this is hooked up to run the kernels, it _could_ end up firing off actions that will cause binder to launch a server"

PR #2416, opened 2018-01-24
Ensure we are not trying to edit anonymous gists.

jdetle:
This prevents an error when uploading as a user and then uploading anonymously.

### DIFF
--- a/applications/desktop/__tests__/renderer/epics/github-publish-spec.js
+++ b/applications/desktop/__tests__/renderer/epics/github-publish-spec.js
@@ -110,17 +110,23 @@ describe("publishNotebookObservable", () => {
       notebook,
       "./test.ipynb",
       notificationSystem,
-      false,
+      true,
       store
     );
     const edit = jest.spyOn(github.gists, "edit");
+    let types = [];
     publishNotebookObs.subscribe(
       x => {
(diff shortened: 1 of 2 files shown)


rgbkrk:
Got some tests to fix up here, it looks like they need adapting to the _now_ proper behavior. 😄

rgbkrk (approved):

applications/desktop/__tests__/renderer/epics/github-publish-spec.js
```diff
@@ -110,17 +110,23 @@ describe("publishNotebookObservable", () => {
...
-      false,
+      true,
       store
     );
     const edit = jest.spyOn(github.gists, "edit");
+    let types = [];
     publishNotebookObs.subscribe(
       x => {
-        expect(x.type).toBe("OVERWRITE_METADATA_FIELD");
+        types.push(x.type);
```
rgbkrk, line 120: This could push the full object and do checks that ensure we get roughly the right actions with:

```
expect(actions).toEqual([
  {
    type: "OVERWRITE_METADATA_FIELD",
    name: "whatever",
    dunno: expect.any(Object)
  }
])
```

jdetle: happy to do that.

PR #2418, opened 2018-01-24
extract contents fetching from jupyter-extension into core

rgbkrk:
Attempted to match much of the style from the kernelspecs setup. This doesn't change much other than set up similar wording of actions, constants, etc.


rgbkrk:
Woohoo! Now the kernel is kicked off when the notebook is loaded. 😄

packages/core/src/actions.js
```diff
@@ -84,11 +84,37 @@ import type {
   // TODO: Needs an action creator
   DoneSavingConfigAction,
   // TODO: Needs an action creator
-  SetNotebookCheckpointAction
+  SetNotebookCheckpointAction,
```
mpacer, line 87: Why did you add this here if you aren’t also adding that action?

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -87,3 +92,36 @@ export const acquireKernelInfoEpic = (action$: ActionsObservable<*>) =>
     ofType(LAUNCH_KERNEL_SUCCESSFUL),
     switchMap(action => acquireKernelInfo(action.kernel.channels))
   );
+
+export const extractNewKernel = (
+  filename?: string,
+  notebook: ImmutableNotebook
+) => {
+  // TODO: There's some incongruence between desktop and web app here, regarding path vs. filename
+  const cwd = (filename && path.dirname(filename)) || "/";
```
mpacer, line 101: Why do we call it filename if it’s actually a path? Technically aren’t all filenames implicit (relative) paths while not all paths are filenames?  Wouldn’t filepath be the more appropriate name? This might make more sense in a later PR but is definitely worth considering.

mpacer: Also should this default to the root directory or is `/` here only being treated as a relative directory since it’s to be treated as a relative path and not an absolute path?

packages/core/src/epics/contents.js
```diff
@@ -0,0 +1,94 @@
...
+      const serverConfig = {
+        endpoint: host.serverUrl,
+        token: host.token,
+        crossDomain: false
+      };
+
+      return contents
+        .get(serverConfig, action.payload.path, action.payload.params)
+        .pipe(
+          tap(xhr => {
```
mpacer, line 49: Not a problem, just I’m curious; does throwing an error inside `tap` propagate the error up and halt execution of the subsequent code inside the `pipe`? If so is the the common way to throw errors when using lettable operators?

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -87,3 +92,36 @@ export const acquireKernelInfoEpic = (action$: ActionsObservable<*>) =>
...
+) =>
+  action$.pipe(
+    ofType(SET_NOTEBOOK),
+    map((action: SetNotebookAction) => {
+      const { cwd, kernelSpecName } = extractNewKernel(
+        action.filename,
+        action.notebook
+      );
+
+      return launchKernelByName(kernelSpecName, cwd);
```
mpacer, line 125: W00t!

packages/core/src/actions.js
mpacer: Oh nvm, comma.

packages/core/src/actions.js
```diff
@@ -84,11 +84,37 @@ import type {
...
+  FetchContent,
+  FetchContentFulfilled,
+  FetchContentFailed
```
mpacer, line 90: Is it possible to add a comma here for cleaner future diffs? Or is this a prettier opinion?

mpacer:
Merging because the only thing that I don’t agree with is stylistic. Really nice & clear code!

packages/core/src/epics/contents.js
rgbkrk: No further operators will run in the pipe, only a `catchError` operator will catch the error thrown here.

packages/core/src/actions.js
rgbkrk: That's a prettier thing.

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: I kept with the original nomenclature of this action as it was written, which was `filename`. We've got some incongruence between desktop and webapp here since the desktop app uses paths which many times need to be `path.resolve`d and the default should be their current working directory rather than `/`. In the case of the web app, `/` is the root of the notebook server.

PR #2419, opened 2018-01-25
hook up the kernel launch in jupyter extension

rgbkrk:
I had this working and was so excited I didn't commit it in #2418, instead playing with the live notebook. Sadly I'm in the middle of shipping so I'll need to ship again with this hooked up...


theengineear (approved):
🙃

PR #2420, opened 2018-01-25
Add menu to nteract-on-jupyter

theengineear:
🎉 first pass at some minimal menu functionality

I went with code-completeness > feature completeness. I.e., there's much more to flesh out here, but I tried my best to get this code right on the first pass so that we can easily extend.

TODO:

- [x] create `notebook-menu` presentation component (i made it a dir with`index.js` so it matches other components there)
- [x] connect presentation component to redux
- [x] snapshot test
- [x] test that we can mount and then assert that the appropriate actions are being called

### DIFF
--- a/packages/core/src/components/notebook-menu/index.js
+++ b/packages/core/src/components/notebook-menu/index.js
@@ -1,0 +1,212 @@
+// @flow
+import * as React from "react";
+import Menu, { SubMenu, Divider, MenuItem } from "rc-menu";
+import { localCss } from "./styles";
+import { connect } from "react-redux";
+import * as Immutable from "immutable";
+import * as actions from "../../actions";
+import { ACTIONS, MENUS } from "./constants";
+
+// To allow actions that can take dynamic arguments (like selecting a kernel
+// based on the host's kernelspecs), we have some simple utility functions to
+// stringify/parse actions/arguments.
+const createActionKey = (action, ...args) => [action, ...args].join(":");
+const parseActionKey = key => key.split(":");
+
+type Props = {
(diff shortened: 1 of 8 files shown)


theengineear:
screen shot 2018-01-25 at 11 00 40 am

packages/core/src/components/notebook-menu/index.js
```diff
@@ -0,0 +1,212 @@
...
+});
+
+const NotebookMenu = connect(mapStateToProps, mapDispatchToProps)(
+  PureNotebookMenu
+);
+
+// We export this for testing purposes.
+export { PureNotebookMenu };
+
+export default NotebookMenu;
```
rgbkrk, line 212: This is beautifully organized.

packages/core/src/components/notebook-menu/index.js
```diff
@@ -0,0 +1,212 @@
...
+      </div>
+    );
+  }
+}
+
+// TODO: this forces the menu to re-render on cell focus. The better option is
+// to alter how the actions work to just target the currently focused cell.
+// That said, we *may* not have a great way of getting around this as we'll need
+// information about the current document to decide which menu items are
+// available...
```
rgbkrk, line 185: Yeah that's a tricky one. We probably need to work with the focused cell and if we need to create more actions (like the "run all").

rgbkrk (approved):

rgbkrk:
I'm confused as to why that's failing on Appveyor, I'm going to go ahead and merge it, clear the cache on appveyor and figure out the build setup.

PR #2421, opened 2018-01-25
Add contents api spec to plan.md.

theengineear:
Opening this up as a pr into the flat-db.md branch so we can iterate easily before merging.

### DIFF
--- a/packages/types/core/plan.md
+++ b/packages/types/core/plan.md
@@ -113,6 +113,14 @@ type state = {
       }
     }
   },
+  contents: {
+    byRef: {
+      [ref: Ref]: {
+        loading: boolean,
+        error: ?Object
+      }
+    }
+  }
 
   // These are the actual data that we get back from
   //   * API Calls
@@ -190,6 +198,7 @@ type state = {
           token: string,
           serverUrl: string,
           crossDomain: boolean,
+          rootContentRef: Ref,
           messages: Array<string> // binder only
         }
       }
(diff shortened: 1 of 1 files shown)


packages/types/core/plan.md
```diff
@@ -190,6 +198,7 @@ type state = {
           token: string,
           serverUrl: string,
           crossDomain: boolean,
+          rootContentRef: Ref,
```
theengineear, line 201: @rgbkrk this is all we should theoretically need to drill into content, right? There should always be some root to our content tree and that logically should be stored in the `host`, right?

packages/types/core/plan.md
```diff
@@ -212,6 +221,16 @@ type state = {
         }
       }
     },
+    contents: {
+      byRef: {
+        [ref: Ref]: {
+          type: "directory" | "file", // TODO: is this right?
+          path: string,
+          name: string,
+          content: Array<Ref> | string // array of refs if directory?
```
theengineear: Is there a better standard we should follow? I feel like I commonly think of the `content` of a directory as a list of `children` and the `content` of a file as text? Might be worth thinking a little more on...

theengineear:
🙃 seems silly, but will you double check the `contents` vs `content` usage. I think it should mirror `hosts`, `host`, `selectedHostRef`. So, `contents`, `content`, `rootContentRef`.

packages/types/core/plan.md
```diff
@@ -113,6 +113,14 @@ type state = {
...
     }
   },
+  contents: {
+    byRef: {
+      [ref: Ref]: {
+        loading: boolean,
+        error: ?Object
+      }
+    }
+  }
```
rgbkrk, line 123: Assuming this is all nested under "communication", 👍

packages/types/core/plan.md
rgbkrk: Yeah, that makes sense.

packages/types/core/plan.md
rgbkrk: Here's a good reference, from deep within our monorepo -- full on flow types for the content API response: https://github.com/nteract/nteract/blob/fe09bf9ae5f77473dab2e2eec9a290ec2d80ca22/applications/commuter/src/content-providers/base/index.js

rgbkrk: The content API response is like a stat call + raw content inside of the `content` key:

```js
export type FileContent = {
  type: "file",
  mimetype: null | string,
  content: null | string,

  name: string,
  path: string,

  created: Date,
  last_modified: Date,
  writable: boolean,
  format: null | "text" | "base64"
};
```

theengineear:
@rgbkrk , ok, updated want to merge it?

PR #2426, opened 2018-01-26
fix: use defaultKeys/openKeys correctly in rc-menu

theengineear:
These *cannot* be `null`. I thought this was sort-of odd behavior at
first for the `rc-menu` comp, but it makes sense since `defaultProps`
will only target `undefined`. Allowing `null` is actually a bit of a
pain to implement in `rc-menu`.

I always get confused about the `undefined` vs `nil-y` types in Flow:

`type foo?: string // string or undefined (i.e., optional)`
`type foo: ?string // string or undefined or null`


theengineear:
@rgbkrk funny when you write your tests last, confirm that they work, and then accidentally checked a stale version in your browser before pushing it up 😅.

Small hot-fix, I'll just merge if I don't hear from you in ~1hr.

rgbkrk (approved):

PR #2428, opened 2018-01-26
Refactor notebook menu

theengineear:
Noticed an opportunity to refactor and make testing this component a little simpler and prevent bloat in what _could basically_ just be a presentation component.

### DIFF
--- a/packages/core/src/components/notebook-menu/extra-handlers.js
+++ b/packages/core/src/components/notebook-menu/extra-handlers.js
@@ -1,0 +1,46 @@
+// @flow
+import * as Immutable from "immutable";
+
+// Adding a level of indirection here to make it trivial to mock these out in
+// tests to ensure they're getting called as we expect. It also prevents the
+// NotebookMenu component from bloating too much.
+
+type FocusedCellId = ?string;
+type CellIds = Immutable.List<string>;
+type CellMap = Immutable.Map<string, *>;
+type CellIdFn = ?(string) => void;
+
+export const executeCells = (
+  executeCell: CellIdFn,
+  cellMap: CellMap,
+  cellIds: CellIds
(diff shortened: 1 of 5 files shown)


packages/core/src/components/notebook-menu/extra-handlers.js
```diff
@@ -0,0 +1,46 @@
...
+  cellMap: CellMap,
+  cellIds: CellIds,
+  focusedCellId: FocusedCellId
+) => {
+  const skipToIndex = focusedCellId ? cellIds.indexOf(focusedCellId) : -1;
+  if (~skipToIndex) {
+    executeCells(executeCell, cellMap, cellIds.skip(skipToIndex));
+  } else {
+    executeCells(executeCell, cellMap, cellIds);
+  }
```
rgbkrk, line 45: This all makes me wonder if we should have an action for "EXECUTE_ALL_CELLS_BELOW_FOCUSED_CELL". Same with Execute all. Force the epic to do the coordination with the state and cellMap.

rgbkrk (approved):
Totally cool with me, though I think we could get out of having the menu itself needing to know what the focused cell is and instead rely on an epic to handle the behavior (and know the state).

packages/core/src/components/notebook-menu/extra-handlers.js
theengineear: yup, totally, i just didn't want to get that deep yet

rgbkrk:
Something I've been thinking about, regarding how we handle `execute_all` -- we could "lock the cell UI" when waiting for execute all to finish (or error), which could either be just graying it out or literally preventing edits while it's in process. Something to think about for coordinating with the epics some day in the future.

PR #2431, opened 2018-01-27
feat: add download notebook option to file menu

theengineear:
I ended up using a 3rd party file downloader. I did also find this while looking around: https://github.com/jimmywarting/StreamSaver.js

If notebooks are quite large (which i imagine they could be), it'd be cool to flesh this out and alternatively stream the notebook to disk if we hit a limit on the blob size. Thoughts?

Chrome's blobs cap out at 500MiB according to https://stackoverflow.com/questions/28307789/is-there-any-limitation-on-javascript-max-blob-size. Any concern there?

### DIFF
--- a/packages/core/src/components/notebook-menu/extra-handlers.js
+++ b/packages/core/src/components/notebook-menu/extra-handlers.js
@@ -1,14 +1,20 @@
 // @flow
 import * as Immutable from "immutable";
+import FileSaver from "file-saver";
+import type { ImmutableNotebook, Notebook } from "@nteract/commutable";
+import * as commutable from "@nteract/commutable";
 
 // Adding a level of indirection here to make it trivial to mock these out in
 // tests to ensure they're getting called as we expect. It also prevents the
 // NotebookMenu component from bloating too much.
 
+// TODO: these exist in commutable/types. We should use export/use those.
 type FocusedCellId = ?string;
 type CellIds = Immutable.List<string>;
 type CellMap = Immutable.Map<string, *>;
 type CellIdFn = ?(string) => void;
+
+const DEFAULT_NOTEBOOK_FILENAME = "notebook.ipynb";
 
 export const executeCells = (
   executeCell: CellIdFn,
@@ -42,5 +48,18 @@ export const executeAllCellsBelow = (
     executeCells(executeCell, cellMap, cellIds.skip(skipToIndex));
   } else {
     executeCells(executeCell, cellMap, cellIds);
+  }
+};
+
+export const downloadNotebook = (
+  immutableNotebook: ?ImmutableNotebook,
+  path: ?string
+) => {
+  if (immutableNotebook) {
+    const notebook = commutable.toJS(immutableNotebook);
+    const filename = (path || DEFAULT_NOTEBOOK_FILENAME).split("/").pop();
+    const data = commutable.stringifyNotebook(notebook);
+    const blob = new Blob([data], { type: "text/json" });
+    FileSaver.saveAs(blob, filename);
   }
 };
--- a/packages/core/src/components/notebook-menu/index.js
+++ b/packages/core/src/components/notebook-menu/index.js
@@ -23,6 +23,8 @@ type Props = {
   executeCell: ?(cellId: ?string) => void,
   cutCell: ?(cellId: ?string) => void,
   copyCell: ?(cellId: ?string) => void,
+  filename: ?string,
+  notebook: Immutable.Map<string, *>,
   pasteCell: ?() => void,
   createCodeCell: ?(cellId: ?string) => void,
   createMarkdownCell: ?(cellId: ?string) => void,
@@ -38,6 +40,7 @@ class PureNotebookMenu extends React.Component<Props> {
     executeCell: null,
     cutCell: null,
     copyCell: null,
+    notebook: null,
     pasteCell: null,
     createCodeCell: null,
     createMarkdownCell: null,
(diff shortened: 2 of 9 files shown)


theengineear:
Also, @rgbkrk (+ all nteract folks) thanks for having a repo with nice surface area to write tests in. It's been a breeze so far.

rgbkrk:
> If notebooks are quite large (which i imagine they could be), it'd be cool to flesh this out and alternatively stream the notebook to disk if we hit a limit on the blob size. 
> Chrome's blobs cap out at 500MiB according to https://stackoverflow.com/questions/28307789/is-there-any-limitation-on-javascript-max-blob-size. Any concern there?

If they get that big, the notebook is _way too big_. Most notebooks are < 5 MB, some large plots put them ~25 MB, and I've seen some astrophysics notebooks that are ~100 MB. There is a `/files/` endpoint that you can stream the files from that this usually uses, so I'd use the size as a heuristic for choosing whether to serve it in-memory or via the server.

> Also, @rgbkrk (+ all nteract folks) thanks for having a repo with nice surface area to write tests in. It's been a breeze so far.

Woo!

packages/core/src/components/notebook-menu/extra-handlers.js
```diff
@@ -44,3 +50,18 @@ export const executeAllCellsBelow = (
     executeCells(executeCell, cellMap, cellIds);
   }
 };
+
+export const downloadNotebook = (notebook: Notebook, path: ?string) => {
+  if (notebook) {
+    const filename = (path || "notebook.ipynb").split("/").pop();
```
rgbkrk: I _think_ we have functions for some of this in desktop that could end up in core.

packages/core/src/components/notebook-menu/index.js
```diff
@@ -23,6 +23,8 @@ type Props = {
   executeCell: ?(cellId: ?string) => void,
   cutCell: ?(cellId: ?string) => void,
   copyCell: ?(cellId: ?string) => void,
+  filename: ?string,
+  notebook: Immutable.Map<string, *>,
```
rgbkrk, line 27: The type for this likely exists in commutable or in types (though I think the latter may be one of those that is currently wrong and not used...)

packages/core/src/components/notebook-menu/extra-handlers.js
```diff
@@ -44,3 +50,18 @@ export const executeAllCellsBelow = (
...
+    const {
+      cellMap = {},
+      cellOrder = [],
+      ...rest
+    }: NotebookJS = (notebook.toJS(): Object);
```
rgbkrk: There's a function for this in `@nteract/commutable` that you _have_ to use because our in-memory structure is _not_ the same as what the on-disk format has to be.

rgbkrk:
```js
import { toJS, stringifyNotebook } from "@nteract/commutable";

stringifyNotebook(toJS(notebook))
```

The way it gets done in the desktop app (since it also sets the version when saved): https://github.com/nteract/nteract/blob/d40b6de052bf6c034b1bd07cc9c1ade5268c4c61/applications/desktop/src/notebook/epics/saving.js#L39-L45

rgbkrk (changes requested):
Definitely need to change up / fix the notebook serializing, then this is good to go.

packages/core/src/components/notebook-menu/extra-handlers.js
theengineear: ah, excellent. yah, i didn't want to over-think this cause i figured that might be the case. i'll grab that. i was just mapping to `cells` in here manually.

theengineear:
Thanks! I'll finish this up tomorrow.

packages/core/src/components/notebook-menu/extra-handlers.js
theengineear: This one ends up being a little different from the one in desktop. I'm going to leave a TODO in here and we can continue to discuss. The difference is that when you download a file from the browser, you almost never want the full `path`, whereas desktop would absolutely need the full `path` on save.

packages/core/src/components/notebook-menu/index.js
theengineear: Seemed to work OK for me and I'd rather not add _another_ implementation of that type 😁

theengineear:
OK! I added a larger dummy notebook (it's called `bigDummy`) so that I could better test some of the execution flows. I'm glad you forced me to explore `commutable` a little more, this PR ended up become wayyyyy more solid PR because of that.

BUT, I now need to fix a weird issue with how `uuid` is mocked in the `notebook-preview` tests...

rgbkrk:
We used to have a moduleNameMapper in our top level jest config that locked the uuid that was used, I wonder if that has to be brought back.

rgbkrk:
This is what it used to be: https://github.com/nteract/nteract/pull/2285/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L70

theengineear:
OK @rgbkrk ready for review 🎉 (thanks for your patience!)

packages/core/src/components/notebook-menu/extra-handlers.js
```diff
@@ -44,3 +50,16 @@ export const executeAllCellsBelow = (
...
+
+export const downloadNotebook = (
+  immutableNotebook: ?ImmutableNotebook,
+  path: ?string
+) => {
+  if (immutableNotebook) {
+    const notebook = commutable.toJS(immutableNotebook);
+    const filename = (path || DEFAULT_NOTEBOOK_FILENAME).split("/").pop();
+    const data = commutable.stringifyNotebook(notebook);
+    const blob = new Blob([data], { type: "text/json" });
```
rgbkrk, line 62: The mime type here should be application/json i believe

rgbkrk (approved):

rgbkrk:
One little comment you can address in a follow up PR. Thanks for kicking the File menu off!

packages/core/src/components/notebook-menu/extra-handlers.js
theengineear: Yah, bad copy 🍝 !

PR #2433, opened 2018-01-27
Add pick theme menu option

theengineear:
Just lets you switch between the themes we provide.

screen shot 2018-01-27 at 3 45 46 pm

### DIFF
--- a/packages/core/src/components/notebook-menu/index.js
+++ b/packages/core/src/components/notebook-menu/index.js
@@ -30,7 +30,8 @@ type Props = {
   createCodeCell: ?(cellId: ?string) => void,
   createMarkdownCell: ?(cellId: ?string) => void,
   setCellTypeCode: ?(cellId: ?string) => void,
-  setCellTypeMarkdown: ?(cellId: ?string) => void
+  setCellTypeMarkdown: ?(cellId: ?string) => void,
+  setTheme: ?(theme: ?string) => void
 };
 
 class PureNotebookMenu extends React.Component<Props> {
@@ -47,7 +48,8 @@ class PureNotebookMenu extends React.Component<Props> {
     createCodeCell: null,
     createMarkdownCell: null,
     setCellTypeCode: null,
-    setCellTypeMarkdown: null
+    setCellTypeMarkdown: null,
+    setTheme: null
   };
   handleClick = ({ key }: { key: string }) => {
     const {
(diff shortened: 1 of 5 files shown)


packages/core/src/components/notebook-menu/index.js
```diff
@@ -7,13 +7,19 @@ import * as Immutable from "immutable";
...
 import * as extraHandlers from "./extra-handlers";
+import * as themes from "../../themes";
 
 // To allow actions that can take dynamic arguments (like selecting a kernel
 // based on the host's kernelspecs), we have some simple utility functions to
 // stringify/parse actions/arguments.
 const createActionKey = (action, ...args) => [action, ...args].join(":");
 const parseActionKey = key => key.split(":");
 
+// TODO: this sort of stinks to have to filter the themes api like this :/
```
theengineear: @rgbkrk thoughts? I was going to add a top-level `themeStrings` key or some such, but then I found code in here that uses `themes` (from `import * as themes...`) as a map of keys to themes. Which feels like a weird way to use a module object? `default` is the obvious oddity there when you do an ES6 import (not there if you `require`).

rgbkrk: We only plan on having two themes, light and dark, and this can be cleaned up so feel free to pick what you think is best here.

packages/core/src/components/notebook-menu/index.js
```diff
@@ -7,13 +7,19 @@ import * as Immutable from "immutable";
...
+const THEME_KEYS = Object.keys(themes)
+  .filter(theme => theme !== "default")
+  .sort();
```
rgbkrk: Ah yeah now that I read it, yes this would be better if they were each named.

packages/core/src/components/notebook-menu/styles.js
```diff
@@ -81,7 +81,7 @@ export const localCss = css`
     position: absolute;
   }
   .rc-menu-submenu > .rc-menu {
-    background-color: #fff;
+    background-color: var(--cell-bg-hover);
```
rgbkrk, line 84: Always make sure to use what the light theme uses for the second arg / default so these components work without explicit theme.

rgbkrk (approved):

packages/core/src/components/notebook-menu/index.js
theengineear: Ah, cool. I'll just be explicit about light/dark then 👍

theengineear:
@rgbkrk I'll make an update to this tomorrow. Let's hold off on merging.

theengineear:
OK @rgbkrk I think that `var( .. )` call is alright, it just wasn't set previously and should have been. The color from the rc-menu port was just close enough that I hadn't noticed. That's available in the light theme though, I believe.

theengineear:
Merge?

PR #2435, opened 2018-01-28
[jext] epic for saving notebooks in the web app

### LINKED ISSUE #2429: [jext] Save Notebook
* [x] Proper rx-jupyter epic for saving
* [x] Menu entry

rgbkrk:
Closes #2429. This PR accomplished all the following:

* [x] create a File -> Save menu item and menu action
* [x] Refactor `SAVE` and `SAVE_AS` actions to not require passing a filename and the raw notebook structure
  * [x] Action creators
  * [x] Desktop Epic
* [x] Refactor `DONE_SAVING` action to not pass back a notebook file
* [x] Store version information in the `app` state, for both desktop and web app
* [x] Create an epic that uses rx-jupyter and matches the style for how we do loading
* [x] Move save related reducers and tests from desktop to core

### DIFF
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -25,25 +25,22 @@ export function saveEpic(
 ) {
   return action$.pipe(
     ofType(SAVE),
-    map(action => {
+    mergeMap(action => {
       const state = store.getState();
-      const notebook = state.document.get("notebook");
-      const filename = state.document.get("filename");
-
-      return {
-        filename,
-        notebook: stringifyNotebook(
-          toJS(
-            notebook.setIn(
+      const notebook = stringifyNotebook(
+        toJS(
+          state.document
+            .get("notebook")
+            .setIn(
               ["metadata", "nteract", "version"],
               state.app.get("version", "0.0.0-beta")
             )
-          )
         )
-      };
-    }),
-    mergeMap(action =>
-      writeFileObservable(action.filename, action.notebook).pipe(
+      );
+
+      const filename = state.document.get("filename");
+
+      return writeFileObservable(filename, notebook).pipe(
         map(() => {
           if (process.platform !== "darwin") {
             const state = store.getState();
@@ -63,8 +60,8 @@ export function saveEpic(
             error: true
           })
         )
-      )
-    )
+      );
+    })
   );
 }
 
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -38,8 +38,7 @@ export function saveEpic(
           toJS(
             notebook.setIn(
               ["metadata", "nteract", "version"],
-              // TODO: set this in the state tree once on startup
-              remote.app.getVersion()
+              state.app.get("version", "0.0.0-beta")
             )
           )
         )
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -11,8 +11,6 @@ import { changeFilename, save, doneSaving } from "@nteract/core/actions";
 
 import { toJS, stringifyNotebook } from "@nteract/commutable";
 
-import { remote } from "electron";
-
 import { of } from "rxjs/observable/of";
 import { tap, mergeMap, catchError, map } from "rxjs/operators";
 
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -33,29 +33,20 @@ export function saveEpic(
       const filename = state.document.get("filename");
 
       return {
-        notebook,
-        filename
-      };
-    }),
-    mergeMap(action =>
-      writeFileObservable(
-        action.filename,
-        stringifyNotebook(
+        filename,
+        notebook: stringifyNotebook(
           toJS(
-            action.notebook.setIn(
+            notebook.setIn(
               ["metadata", "nteract", "version"],
+              // TODO: set this in the state tree once on startup
               remote.app.getVersion()
             )
           )
         )
-      ).pipe(
-        catchError((error: Error) =>
-          of({
-            type: "ERROR",
-            payload: error,
-            error: true
-          })
-        ),
+      };
+    }),
+    mergeMap(action =>
+      writeFileObservable(action.filename, action.notebook).pipe(
         map(() => {
           if (process.platform !== "darwin") {
             const state = store.getState();
@@ -66,8 +57,15 @@ export function saveEpic(
               level: "success"
             });
           }
-          return doneSaving(action.notebook);
-        })
+          return doneSaving();
+        }),
+        catchError((error: Error) =>
+          of({
+            type: "ERROR",
+            payload: error,
+            error: true
+          })
+        )
       )
     )
   );
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -27,11 +27,15 @@ export function saveEpic(
 ) {
   return action$.pipe(
     ofType(SAVE),
-    tap(action => {
-      // If there isn't a filename, save-as it instead
-      if (!action.filename) {
-        throw new Error("save needs a filename");
-      }
+    map(action => {
+      const state = store.getState();
+      const notebook = state.document.get("notebook");
+      const filename = state.document.get("filename");
+
+      return {
+        notebook,
+        filename
+      };
     }),
     mergeMap(action =>
       writeFileObservable(
@@ -66,9 +70,6 @@ export function saveEpic(
         })
       )
     )
-    // .startWith({ type: START_SAVING })
-    // since SAVE effectively acts as the same as START_SAVING
-    // you could just look for that in your reducers instead of START_SAVING
   );
 }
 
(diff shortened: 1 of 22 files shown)


rgbkrk:
Woo, initial cleanup / refactor done. Now for the web based epic!

rgbkrk:
Oh hey I just fixed our whole "Are you sure you want to exit before saving" issue. The state was being changed on save and was set as the `savedNotebook`, which would _never_ be equal to the current state.  🎉  It's working properly now! Hurrah! 🎊

/cc @lgeiger

rgbkrk:
This is worth a look over @theengineear! I'm finding the not-closing-after-click behavior of the menu rather odd. Is that configurable?

rgbkrk:
Super ready for review. I've grouped the commits in a way that you could review them individually, however it's fine to review in whole.

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -27,31 +25,26 @@ export function saveEpic(
...
+
+      return {
+        filename,
+        notebook: stringifyNotebook(
           toJS(
-            action.notebook.setIn(
+            notebook.setIn(
               ["metadata", "nteract", "version"],
-              remote.app.getVersion()
+              // TODO: set this in the state tree once on startup
```
rgbkrk: Now actually done so I'll delete this TODO.

rgbkrk:
Cleaned up a bunch of inline TODOs that I addressed in other commits as part of this PR.

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -27,11 +27,15 @@ export function saveEpic(
...
     ofType(SAVE),
-    tap(action => {
-      // If there isn't a filename, save-as it instead
-      if (!action.filename) {
-        throw new Error("save needs a filename");
-      }
+    map(action => {
+      const state = store.getState();
+      const notebook = state.document.get("notebook");
+      const filename = state.document.get("filename");
```
theengineear: Nice. Yah, this is way cleaner 👍

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -27,11 +27,15 @@ export function saveEpic(
...
+
+      return {
```
theengineear: This is a little weird to return this object that's not really an action. I see that it's basically just getting handled below. Not a huge deal, but I think it would make sense to just handle this section below in the `mergeMap`?

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -77,12 +78,22 @@ export function saveEpic(
...
-    ])
+    mergeMap(action => {
+      const state = store.getState();
+      const notebook = state.document.get("notebook");
+
+      return [
+        // order matters here, since we need the filename set in the state
+        // before we save the document
+        changeFilename(action.filename),
+        save()
```
theengineear, line 87: I wonder if we should create a `saveAs` action so that this isn't so order-dependent? Also not a big deal, 💯 for leaving a comment.

applications/desktop/src/notebook/menu.js
```diff
@@ -34,9 +34,7 @@ import {
 import { defaultPathFallback, cwdKernelFallback } from "./path";
 
 export function dispatchSaveAs(store, evt, filename) {
-  const state = store.getState();
-  const notebook = state.document.get("notebook");
-  store.dispatch(saveAs(filename, notebook));
+  store.dispatch(saveAs(filename));
```
theengineear, line 37: Does this not permanently change the filename in state or something? I left a comment above, but I didn't realize this action already existed.

packages/core/src/components/notebook-menu/index.js
```diff
@@ -154,6 +162,9 @@ class PureNotebookMenu extends React.Component<Props> {
           selectable={false}
         >
           <SubMenu key={MENUS.FILE} title="File">
+            <MenuItem key={createActionKey(MENU_ITEM_ACTIONS.SAVE_NOTEBOOK)}>
+              Save
+            </MenuItem>
```
theengineear, line 167: Sweet! Mind doing a copy-paste job in the menu-spec? (reviewing commit-by-commit so apologies if you did and i didn't see it yet). You should be able to basically copy one of the little blobs there and change it to use this new constant.

packages/core/src/epics/contents.js
```diff
@@ -20,7 +29,7 @@ import { contents } from "rx-jupyter";
 
 import { toJS, stringifyNotebook } from "@nteract/commutable";
```
theengineear: Just noticed this, but we have two imports from `@nteract/commutable` in here that should get merged.

packages/core/src/epics/contents.js
```diff
@@ -67,6 +76,55 @@ export function fetchContentEpic(
...
+        )
+      };
+    }),
+    mergeMap(({ filename, notebook }) => {
+      // TODO: Use the selector from the kernelspecs work
+      const host = store.getState().app.host;
+      const serverConfig = {
+        endpoint: host.serverUrl,
+        token: host.token,
+        crossDomain: false
```
theengineear: Maybe add a comment on the app.host.crossDomain vs setting this crossDomain to `false`?

packages/core/src/epics/contents.js
```diff
@@ -67,6 +76,55 @@ export function fetchContentEpic(
...
+      };
+
+      const model = {
+        content: notebook,
+        type: "notebook"
+      };
+
+      return contents.save(serverConfig, filename, model).pipe(
+        mapTo(doneSaving()),
```
theengineear, line 110: `mapTo`. haven't used that one yet. handy!

theengineear: Is there anything potentially weird about sending *literally* the same action a bunch of times? I guess not... feels sorta odd to me. But only because I guess I haven't seen the pattern before.

theengineear (approved):
Awesome work @rgbkrk! The commits you broke this down into made it a breeze to review! I have a couple non-blocking comments in there that are mostly stylistic. 💥 !

theengineear:
I'll let you take a look through the comments before merging. Let me know if you just want to handle them separately; that's cool too!

applications/desktop/src/notebook/epics/saving.js
rgbkrk: I think that'd be fine. I was mostly following the old order. Let me make this all in one here.

applications/desktop/src/notebook/epics/saving.js
rgbkrk: There _is_ a saveAs action, which is how this comes to be in the first place. I was a little worried about changing the current state of how the desktop app worked in the process of doing this. Then again, `save` used to take the `filename` and `notebook` so I suppose this is a bit different now...

applications/desktop/src/notebook/menu.js
rgbkrk: It doesn't change the filename until `changeFilename` is dispatched, though it seems like the way this works -- we could possibly get rid of `changeFilename` and make a reducer for `saveAs`?

packages/core/src/components/notebook-menu/index.js
rgbkrk: Ah, I had not done that. Good catch.

packages/core/src/epics/contents.js
rgbkrk: I think the weirdness about this is that we're not tracking a `ref` to denote that we finished saving a particular file. Right now we're in a state of our app having only one file at a time.

theengineear:
Also... Why did you change the title from `Here I come to save the notebook`!? That was great!

rgbkrk:
> Also... Why did you change the title from Here I come to save the notebook!? That was great!

¯\\\_(ツ)\_/¯

I at least kept the branch name.

mighty mouse

packages/core/src/components/notebook-menu/index.js
rgbkrk: Done.

packages/core/__tests__/components/notebook-menu-spec.js
```diff
@@ -175,6 +176,13 @@ describe("PureNotebookMenu ", () => {
       setThemeDarkItem.simulate("click");
       expect(props.setTheme).toHaveBeenCalledTimes(1);
       expect(props.setTheme).toHaveBeenCalledWith("dark");
+
+      const saveNotebookItem = wrapper
```
theengineear, line 180: ❤️

applications/desktop/src/notebook/epics/saving.js
rgbkrk: Cleaning this one up now, I want to make sure to test things out just to make sure.

applications/desktop/src/notebook/epics/saving.js
theengineear: ah. right, right. okay that makes sense. 👍

applications/desktop/src/notebook/menu.js
theengineear: Let's leave it as is for now 👍 . I think we could just listen to a new SAVE_AS action type in the same reducer that CHANGE_FILENAME is listening for. Even still allow a `filename` arg in the `save` action potentially. I think it's worth punting for now to have a proper discussion about.

packages/core/src/epics/contents.js
rgbkrk: One other potentially weird bit to this is that if _anyone_ mutates the action, every other receiver of this call gets the changes. Not that we use any fields here, but that would be very gross.

rgbkrk:
Ok, this is all set, at least for the bits we weren't going to punt on.

packages/core/src/epics/contents.js
theengineear: Ha, yah, that's interesting. Well, we could prevent all that by just `map(() => doneSaving())` 😸

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -25,25 +25,22 @@ export function saveEpic(
 ) {
   return action$.pipe(
     ofType(SAVE),
-    map(action => {
+    mergeMap(action => {
```
theengineear: 🙇

packages/core/src/epics/contents.js
rgbkrk: That or `mapTo(Object.freeze(doneSaving()))`. 😄 

Totally cool to switch this to a `map` operator, that's a welcome quick fix.

Issue #2271, opened 2017-12-30
[refactor] Clean up the editor package


rgbkrk:
As the codemirror editor react component got merged together, some of the old cruft was kept for the sake of timeliness and stability. _However_, this left some warts that manifest themselves as either bugs or extra complexity that we can do away with:

* [x] Additional props that we were allowing for configuration, which should just be set as codemirror options `props.options`.
  * [x] `cursorBlinkRate`
  * [x] `language` -- should switch to `mode`, being tackled by @mpacer 
  * [x] `preserveScrollPosition` -- I'm not sure if/when we'd use this one (?)
* [x] Switch tooltips to using `ReactDOM.createPortal` -- see https://github.com/nteract/nteract/blob/f41b58503325e844b1424a8ab29a4a71afadc102/packages/editor/src/index.js#L287
* [ ] `componentWillRecieveProps` has the proper (?) merging of new options into codemirror, whereas our constructor (and `componentDidMount`) have this weird mishmash of defaultOptions (by necessity for binding sake) 
* [x] The codemirror creates several subscriptions that it does not close, namely `keyupEvents`, `tip`s, and `completions`. I _think_ we should put these in the component's state and make sure to dispose of them on `componentDidUnmount`

Feel free to add more. 😋

rgbkrk:
Adding on a little more details about the subscriptions -- we need to unsubscribe when the component is unmounted (or maybe also `takeUntil`).

Within our epics, that cleanup happens naturally. With standalone Rx, you need to track the "subscription"

```js
> var Rx = require('rxjs/Rx');
undefined
> var obs = Rx.Observable.interval(2000).map(i => i*100)
undefined
> var subscription = obs.subscribe(x => console.log(x))
undefined
0
100
200
> subscription.unsubscribe()
```

rgbkrk:
This is probably in a good enough state to just close this since @captainsafia fixed this up at the 2017 winter sprint!

PR #2443, opened 2018-01-29
Use React.createPortal for tooltip

captainsafia:
Trying to figure out how to get it to remove the tip element properly. And if this is the right approach.

Closes part of #2271.

### DIFF
--- a/packages/editor/src/index.js
+++ b/packages/editor/src/index.js
@@ -75,6 +75,7 @@ class CodeMirrorEditor extends React.Component<
     this.hint = this.completions.bind(this);
     this.tips = this.tips.bind(this);
     this.hint.async = true;
+    this.state = { tipElement: null };
 
     this.defaultOptions = Object.assign(
       {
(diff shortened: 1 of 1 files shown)


rgbkrk:
Nice! I have yet to use `React.createPortal` so I look forward to learning from this PR.

captainsafia:
This passes the tests and does what it is supposed to.

I still need to remove all of the unneeded code from the `tips` function.

I'll do that tonight.

captainsafia:
This is ready for review!

packages/editor/src/index.js
```diff
@@ -321,8 +314,11 @@ class CodeMirrorEditor extends React.Component<
           node
         );
 
+        this.setState({ tipElement });
```
rgbkrk, line 317: Wow this is a really neat pattern, I like this.

packages/editor/src/index.js
```diff
@@ -321,8 +314,11 @@ class CodeMirrorEditor extends React.Component<
...
+
         editor.addWidget({ line: editor.getCursor().line, ch: 0 }, node, true);
 
+        const body = document.body;
```
rgbkrk, line 321: Pro-level flow.

packages/editor/src/index.js
```diff
@@ -285,19 +285,12 @@ class CodeMirrorEditor extends React.Component<
           return;
         }
 
-        const node = document.createElement("div");
-        node.id = "tooltip-that-should-be-done-with-react-portals-now";
+        const node = document.getElementsByClassName("tip-holder")[0];
```
rgbkrk, line 288: Would it make sense to use a `ref` here instead?

rgbkrk (approved):

packages/editor/src/index.js
captainsafia: Hm. Yeah. That makes most sense. TBH, I hadn't considered it because all the tutorials I read used the class name as the identifier. 😆

PR #2445, opened 2018-01-29
extract getServerConfig to 🆕 core selectors

rgbkrk:
Probably about time we create general selectors! This `getServerConfig` selector comes from kernelspecs and is now applied to all the `rx-jupyter`-using epics.

I've also gone ahead and introduced `reselect` to memoize this one. This is my first time _actually_ using `reselect` rather than just knowing about it. 😄

### DIFF
--- a/packages/core/src/selectors/index.js
+++ b/packages/core/src/selectors/index.js
@@ -3,10 +3,15 @@ import type { AppState } from "@nteract/types/core/records";
 
 import { createSelector } from "reselect";
 
-export const getHost = (state: AppState) => state.app.host;
+const serverUrl = (state: AppState) => state.app.host.serverUrl;
+const crossDomain = (state: AppState) => state.app.host.crossDomain;
+const token = (state: AppState) => state.app.host.token;
 
-export const getServerConfig = createSelector(getHost, host => ({
-  endpoint: host.serverUrl,
-  crossDomain: host.crossDomain,
-  token: host.token
-}));
+export const getServerConfig = createSelector(
+  [serverUrl, crossDomain, token],
+  (serverUrl, crossDomain, token) => ({
+    endpoint: serverUrl,
+    crossDomain,
+    token
+  })
+);
--- a/packages/core/src/selectors/index.js
+++ b/packages/core/src/selectors/index.js
@@ -1,9 +1,12 @@
 // @flow
 import type { AppState } from "@nteract/types/core/records";
 
-// TODO: This assumes a jupyter host for now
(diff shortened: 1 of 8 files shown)


packages/core/src/selectors/index.js
```diff
@@ -1,9 +1,12 @@
...
-export const getServerConfig = (state: AppState) => ({
-  endpoint: state.app.host.serverUrl,
-  crossDomain: state.app.host.crossDomain,
-  token: state.app.host.token
-});
+import { createSelector } from "reselect";
+
+export const getHost = (state: AppState) => state.app.host;
+
+export const getServerConfig = createSelector(getHost, host => ({
```
theengineear: Since `serverUrl`, `crossDomain`, and `token` are primitive types, you could also:

```
export const getServerConfig = createSelector(
  [
    state.app.host.serverUrl,
    state.app.host.crossDomain,
    state.app.host.token
  ],
  (endpoint, crossDomain, token) => ({ endpoint, crossDomain, token })
)
```

Right?

theengineear: I'm sure it's not a big enough performance increase to make a difference, but maybe we should get in the habit?

theengineear:
Noice! One comment, but looks great! Thanks for finally getting us on `reselect` 😁

packages/core/src/selectors/index.js
rgbkrk: I'm lost on this -- where does `state` come from? Is that intended to be `createSelector((state) => [...]`

packages/core/src/selectors/index.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import type { AppState } from "@nteract/types/core/records";
+
+import { createSelector } from "reselect";
+
+const serverUrl = (state: AppState) => state.app.host.serverUrl;
+const crossDomain = (state: AppState) => state.app.host.crossDomain;
+const token = (state: AppState) => state.app.host.token;
```
theengineear, line 8: 💯

packages/core/src/selectors/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const getServerConfig = createSelector(
+  [serverUrl, crossDomain, token],
+  (serverUrl, crossDomain, token) => ({
+    endpoint: serverUrl,
```
theengineear, line 13: Good idea to not switch the naming in the args 👍

theengineear (approved):

Issue #2446, opened 2018-01-29
The Model Modal OR the Portal vs Redux


theengineear:
This issue is meant to discuss how we want to do modals. Should we:

1. Use Redux & `<ModalController />`
2. Use React 16 Portals

Dan Abramov has a lovely answer to implementing modals as the accepted answer here: https://stackoverflow.com/questions/35623656/how-can-i-display-a-modal-dialog-in-redux-that-performs-asynchronous-actions/35641680

However, `react-portal` and now it's implemented with `React.createPortal` is also viable (if you scroll down in the SO post, you'll see answers along those lines as well.

In all my React projects, I've used the former approach where you dump a `ModalController` at the top level of your app and dispatch redux actions which set state and the modals are rendered based on app state. One difficulty (though I think it's fairly simple to overcome) is making it extensible in a mono-repo structure.

I'm still feeling in favor of using the Redux approach and here's a sketch of what that would roughly look like:

```js
// core/actionTypes.js
const OPEN_MODAL = "CORE/OPEN_MODAL";
const CLOSE_MODAL = "CORE/CLOSE_MODAL";

// core/actions.js
const openModal = ({ modalType, modalProps }) => ({
  type: actionTypes.OPEN_MODAL,
  payload: { modalType, modalProps }
});
const closeModal = () => ({ type: actionTypes.closeModal });

// core/reducers.js
const modalType = (state = "", action) => {
  switch(action.type) {
    case actionTypes.OPEN_MODAL:
      return action.payload.modalType;
    case actionTypes.CLOSE_MODAL:
      return "";
    default:
      return state;
  }
}
const modalProps = (state = {}, action) => {
  switch(action.type) {
    case actionTypes.OPEN_MODAL:
      return action.payload.modalProps;
    case actionTypes.CLOSE_MODAL:
      return {};
    default:
      return state;
  }
};
const modal = combineReducers({ modalType, modalProps });

// core/components/modal-controller/index.js
class ModalController extends React.Component {
  getModal = () => {
    const { modalType } = this.props;
    switch(modalType) {
      case "core/about-modal":
        return AboutModal;
      default:
        return null;
    }
  };
  render() {
    const { modalProps } = this.props;
    const Modal = this.getModal();
    if (!Modal) {
      return null;
    }
    return (
      <ModalOverlay>
        <Modal {...modalProps} />
      </ModalOverlay>
    );
  }
}

// not-core/components/modal-controller/index.js
[...]

PR #2447, opened 2018-01-29
feat: add help/about to web menu.

theengineear:
First pass at the about menu. I didn't want to get too far into it so it's pretty minimal / uninformative... I could use some help understanding the rest of the information that needs to make its way into this menu.

screen shot 2018-01-29 at 8 39 20 pm

### DIFF
--- a/packages/core/src/components/modal-controller/about-modal.js
+++ b/packages/core/src/components/modal-controller/about-modal.js
@@ -1,0 +1,93 @@
+// @flow
+
+// We need to allow to allow escape from the modal from an inner element.
+/* eslint jsx-a11y/no-noninteractive-tabindex: 0 */
+
+// TODO: we can remove this if we knew an appropriate role for the modal shadow.
+// We want to allow the user to click on the modal-backdrop to escape.
+/* eslint jsx-a11y/no-static-element-interactions: 0 */
+
+import React from "react";
+import { modalCss } from "./styles";
+import { connect } from "react-redux";
+import * as actions from "../../actions";
+
+type Props = {
+  appVersion?: string,
+  hostType?: string,
+  closeModal: ?() => void
+};
+
+// We need to do this so that you can immediately `Escape` out of the dialog.
+// Otherwise, we'd need to (a) put an event listener on the document or (b)
+// require that the user clicks the content first before attempting to escape.
+const focusOnRender = el => el && el.focus();
+
+class PureAboutModal extends React.Component<Props> {
+  handleKeyDown = (event: KeyboardEvent) => {
+    const { closeModal } = this.props;
+    const { key, metaKey, altKey, ctrlKey, repeat, shiftKey } = event;
+    if (
+      key === "Escape" &&
+      !(metaKey || altKey || ctrlKey || repeat || shiftKey) &&
+      closeModal
+    ) {
+      closeModal();
+    }
+  };
+  handleOverlayClick = (event: Event) => {
+    const { closeModal } = this.props;
+    if (closeModal && event.target && event.target === event.currentTarget) {
+      closeModal();
+    }
+  };
+  render() {
+    const { appVersion, hostType } = this.props;
+    return (
+      <div>
+        <div
+          className="modal--overlay"
+          onClick={this.handleOverlayClick}
+          onKeyDown={this.handleKeyDown}
+          tabIndex={-1}
+        >
+          <dialog tabIndex={0} className="modal--content" ref={focusOnRender}>
+            <div className="modal--content--header">
+              <h2>About nteract on Jupyter</h2>
+              <p>You are using an nteract-rendered notebook on Jupyter.</p>
+            </div>
+            <div className="modal--content--body">
+              <p>
+                <span className="modal--content--body--field">Version:</span>
+                <span className="modal--content--body--value">
+                  {appVersion}
+                </span>
+              </p>
+              <p>
+                <span className="modal--content--body--field">Host Type:</span>
+                <span className="modal--content--body--value">{hostType}</span>
+              </p>
+            </div>
+          </dialog>
+        </div>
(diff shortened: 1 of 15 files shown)


theengineear:
Related https://github.com/nteract/nteract/issues/2446

This component now has to store state, which seems unnecessary. Had I gone with the longer approach of creating an actionType/action/reducer/ControllerComponent setup, we could really have this in isolation. I.e., the action would be the only thing coupling the menu to the modal... which could potentially be cleaner. I think this will work for now though.

Note, reused bits of our set of css variables where I could in the modal css.

packages/core/src/components/notebook-menu/about-modal.js
```diff
@@ -0,0 +1,61 @@
...
+      <div>
+        <div
+          className="modal--overlay"
+          onClick={this.handleOverlayClick}
+          ref={this.overlayRef}
+        >
+          <div className="modal--content">
+            <div className="modal--content--header">
+              <h2>About Nteract on Jupyter</h2>
+              <p>You are using an Nteract-rendered notebook on Jupyter.</p>
```
rgbkrk: We lowercase "nteract".

packages/core/src/components/notebook-menu/styles.js
```diff
@@ -1,6 +1,72 @@
...
+      left: 100px;
+      right: 100px;
+    }
+  }
+  @media only screen and (max-width: 600px) {
+    .modal--content {
+      left: 40px;
+      right: 40px;
+    }
+  }
```
rgbkrk: Oh did it actually let you use `@media` queries? I thought for sure these fell under the "no nested" category. 👍 🎉

packages/core/src/components/notebook-menu/about-modal.js
theengineear: 😬 shame.

packages/core/src/components/notebook-menu/styles.js
theengineear: nah, this is totally normal CSS, the bug is when you have a newline above a media query for some reason. it seems to flake on that parse. but yah, seems to work just fine 🎉

rgbkrk:
> Had I gone with the longer approach of creating an actionType/action/reducer/ControllerComponent setup, we could really have this in isolation.

That's totally fine if you want to do that.

packages/core/src/components/notebook-menu/about-modal.js
```diff
@@ -0,0 +1,61 @@
...
+  hostType?: string,
+  isOpen: boolean,
+  close: () => void
+};
+class AboutModal extends React.Component<Props> {
+  overlay: ?HTMLDivElement;
+
+  handleOverlayClick = (event: Event) => {
+    const { close } = this.props;
+    if (this.overlay && event.target === this.overlay) {
```
rgbkrk: I _think_ if you use `currentTarget` and separate the two divs here so they're not nested you can go without the `ref`.

theengineear: ah, yup, totally right 👍

theengineear:
@rgbkrk  ⏪ ⏹ ▶️ OK, re-implemented with a `ModalController`.

packages/core/src/components/modal-controller/about-modal.js
```diff
@@ -0,0 +1,65 @@
+// @flow
+
```
rgbkrk: For eslint, you're going to have to set:

```
/* eslint jsx-a11y/no-static-element-interactions: 0 */
/* eslint jsx-a11y/click-events-have-key-events: 0 */
/* eslint jsx-a11y/no-noninteractive-element-interactions: 0 */
```

Since there's a click event on a div here.

rgbkrk, line 2: For eslint, you're going to have to set:

```
/* eslint jsx-a11y/no-static-element-interactions: 0 */
/* eslint jsx-a11y/click-events-have-key-events: 0 */
/* eslint jsx-a11y/no-noninteractive-element-interactions: 0 */
```

Since there's a click event on a div here.

rgbkrk (changes requested):

theengineear:
@rgbkrk, your comments made me realize I skipped some corners 😓. Now we:

1. allow escape-key-exit from the modal
2. have a real `dialog` (yay new html tags)

theengineear:
I like those `eslint` rules! Super handy!

rgbkrk:
Oh hey that's even better since the linter prompted better a11y conformance. :smile:

packages/core/src/components/modal-controller/about-modal.js
```diff
@@ -0,0 +1,93 @@
...
+// We need to allow to allow escape from the modal from an inner element.
+/* eslint jsx-a11y/no-noninteractive-tabindex: 0 */
+
+// TODO: we can remove this if we knew an appropriate role for the modal shadow.
+// We want to allow the user to click on the modal-backdrop to escape.
+/* eslint jsx-a11y/no-static-element-interactions: 0 */
```
rgbkrk, line 8: Superb TODOs for the eslint ignores. 😄

packages/core/src/components/modal-controller/about-modal.js
```diff
@@ -0,0 +1,93 @@
...
+    const { appVersion, hostType } = this.props;
+    return (
+      <div>
+        <div
+          className="modal--overlay"
+          onClick={this.handleOverlayClick}
+          onKeyDown={this.handleKeyDown}
+          tabIndex={-1}
+        >
+          <dialog tabIndex={0} className="modal--content" ref={focusOnRender}>
```
rgbkrk, line 54: 👍

rgbkrk (approved):

PR #2449, opened 2018-01-30
refactor: create keyboard shortcuts provider

theengineear:
Not sure if I ended up liking the direction this took... Thoughts?

Notes:

1. I tried using react-hotkeys. Couldn't get that working as expected (perhaps something with what react-hotkeys calls the *focused tree*)?
2. I tried just using Mousetrap, and there's something very subtly different about how we handle focusing the next cell and then focusing the next cell editor that wasn't right, but only after hitting a markdown cell? I think it's how the preventDefault is use inside the markdown cell... anyways. That was fun :)
3. Stuck to plain old `addEventListener`, but added a step that just parses out a command string such that you just need to register a command for each possible string. I really like the testing prospects for this, so I fleshed out a little test suite.

My major grievance with this is that folks now need to remember to wrap the whole app in this new component...

I'm going to hook up the save action real quick in a separate commit as a bit more of a POC.


rgbkrk:
You'll need to rebase since I merged the other PRs (?).

Some comments that I put in chat (that should have been here in the PR):

> I think you'd have to tear our the key handling from the markdown preview component. That needs to be done anyways so that keybindings are top level. I think you'd be able to use mousetrap then.

theengineear:
Closing this down. We can revisit later.

PR #2455, opened 2018-01-31
put regression test notebook in commuter samples

rgbkrk:
Solely to make it easy to test a commuter deployment.


cabhishek (approved):
LGTM

Issue #2430, opened 2018-01-27
[jext] approximate menu parity with desktop app


rgbkrk:
* [x] File
  * New Notebook (by kernel name) (deferred to #2438)
  *  Open (deferred to #2731)
  * [x] Save
  *  Duplicate (deferred to #2725)
  *  ~~Rename~~ This is now part of the title bar
  * [x] Download (.ipynb) 
  *  Close and Halt (deferred to #2724)
* [x] Edit
  * [x] Cut Cell
  * [x] Copy Cell
  * [x] Paste Cell
  * [x] Merge Cell Below
  * Move Cell Up (not essential for M1)
  * Move Cell Down (not essential for M1)
  * Edit Notebook Metadata (not essential for M1)
  * Find and Replace (not essential for M1)
* [x] View
  * Toggle Header (not essential for M1)
  * Toggle Toolbar (not essential for M1)
  * Presentation Mode (not essential for M1)
  * Report Mode
  * [x] Theme (light/dark)
* [x] Cell
  * [x] New Markdown Cell
  * [x] New Code Cell
  * [x] Run All
  * [x] Run All Below
  * [x] Cell Type
  * [x] Unhide All Outputs
  * [x] Clear All Outputs
* [ ] Runtime
  * [x] Interrupt #2448 
  * [x] Kill
  * [x] Restart
  * [ ] Reconnect
    * do the connect again with the same kernel id
  * [x] Switch to ▶ (Language)
  * [x] Restart and Clear Output
  * Restart and Run All (deferred to #2727)
* [x] Help
  * Keyboard Shortcuts (not essential for M1)
  * Custom Resources... (generated help links, tailored by operator)
  * Kernel Resources... (generated help links, come from kernel info) (not essential for M1)
  * [x] About #2447

![menu](https://user-images.githubusercontent.com/836375/35467281-5fd3d59e-02c1-11e8-95fe-7c3bb42cad3e.png)

Symbol legend:

* Green star === trivial to implement
* Red star === quite a bit work to implement
* Blue circle === "global" actions (not dependent on focused cell)
* Underline === First to be implemented in a section of the menu

rgbkrk:
To make the above a little easier to triage what's done and where we're at I'm going to type the above up as a TODO list.

rgbkrk:
After #2448 is merged, I'm going to move the menu order to match the above:

```
File  |  Edit  |  View  |  Cell  |  Runtime  |  Help
```

[...]

PR #2460, opened 2018-02-01
switch order to match more closely with jupyter

rgbkrk:
Switch around the menu order to match what was outlined in  https://github.com/nteract/nteract/issues/2430

```
File  |  Edit  |  View  |  Cell  |  Runtime  |  Help
```

🍽


rgbkrk:
I'm going to err on the side of 👌  here and keep hacking away.

PR #2462, opened 2018-02-01
clean out more desktop reducers

rgbkrk:
* move notification system into core app reducer
* delete unused configLastSaved property


rgbkrk:
I'm going to err on the side of 👌  here and keep hacking away.

Issue #2461, opened 2018-02-01
Move desktop kernel's cleanup into epics (and out of reducers)


rgbkrk:
While working on the nbextension I noticed probably the murkiest thing in the desktop app -- the restart, interrupt, and kill kernel logic all live inside reducers even though they're inherently asynchronous. The fix is pretty simple -- it's a bit more ritual around using epics for these, leaving room for us to do proper checking of the cleanup.

PR #2463, opened 2018-02-01
move interrupt zeromq kernels logic to epic

rgbkrk:
Handles part of #2461.

This does some cleanup of desktop's app reducer to bring the epic structure inline with the web app based version. After this I'll handle the launch kernel cleanup + shutdown logic.

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.js
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.js
@@ -2,6 +2,7 @@
 
 import { Observable } from "rxjs/Observable";
 import { of } from "rxjs/observable/of";
+import { fromEvent } from "rxjs/observable/fromEvent";
 import { merge } from "rxjs/observable/merge";
 
 import {
@@ -12,7 +13,8 @@ import {
   catchError,
   first,
   pluck,
-  switchMap
+  switchMap,
+  concatMap
 } from "rxjs/operators";
 
 import { launchSpec } from "spawnteract";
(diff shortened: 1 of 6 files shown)


applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -164,3 +169,52 @@ export const launchKernelEpic = (action$: ActionsObservable<*>) =>
       )
     )
   );
+
+export const interruptKernelEpic = (action$: *, store: *) =>
+  action$.pipe(
+    ofType(INTERRUPT_KERNEL),
+    filter(() => {
```
theengineear, line 176: Hmm. Definitely just for discussion. BUT, we could also do a bit of indirection instead of filtering here. Like instead of ignoring these, we could have yet another epic that just handles all of these actions and maps them to more specific actions:

```
handle INTERRUPT_KERNEL:
  if ({zeromq kernel}) {
    return INTERRUPT_KERNEL_ZEROMQ
  } else if ({ jupyter kernel }) {
    return INTERRUPT_KERNEL_JUPYTER
  }
```

What I'm worried about is us getting into a pattern of swallowing actions. I think we could alternatively just be explicit about routing.

I guess I'm thinking about trying to debug the INTERRUPT_KERNEL action (i know, i'm using nomenclature loosely). I'd probably `grep` for that string to see where it's handled. If it shows up in a bunch of epics, but is only handled by *some* of them, that feels like a non-linear debugging process.

Anyways 🐇 🕳 ... but I think we might want to rethink this pattern.

applications/desktop/src/notebook/reducers/app.js
```diff
@@ -41,16 +37,6 @@ function exit(state: AppRecord) {
   return cleanupKernel(state);
 }
 
-function interruptKernel(state: AppRecord) {
-  // TODO: This should be an epic instead
```
theengineear, line 45: 🎉

theengineear:
Awesome! I'm going to merge as is, but let's discuss observable redux patterns a bit?

PR #2464, opened 2018-02-02
kill kernel as epic

### LINKED ISSUE #2461: Move desktop kernel's cleanup into epics (and out of reducers)

rgbkrk:
Closes #2461 

* [x] Create an epic that causes proper cleanup of kernels
* [x] delete kill kernel reducer
* [x] make new kernel launching kill the old kernel
* [x] Figure out restarts
* [x] ~~handle app `unload` properly by creating a stateful epic that tracks shutdown~~ it's roughly the same as it was before, we need a bit more work on this


rgbkrk:
This PR was insane and all too necessary.

rgbkrk:
I'm going to tackle `unload` tomorrow.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -157,17 +178,34 @@ export const launchKernelEpic = (action$: ActionsObservable<*>) =>
...
-      merge(
+    // We must kill the previous kernel now
+    // Then launch the next one
+    mergeMap(action => {
+      const state = store.getState();
+      const kernel = state.app.kernel;
+
+      const lifecycle$ = kernel
+        ? merge(
+            of({ type: "KILLING_OLD_KERNEL" }),
```
rgbkrk: We'll want an action creator for this -- note that it's distinct from the KILL KERNEL action in that we're allowing for killing the old kernel while we create a new one. We can't fire off `KILL_KERNEL` because we could run into a timing issue where we end up killing off the kernel we just launched. Up until refs, we can't go down that path.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -157,17 +178,34 @@ export const launchKernelEpic = (action$: ActionsObservable<*>) =>
...
+            killKernelObservable(kernel)
+          )
+        : of({ type: "FRESH_START" });
```
rgbkrk: I _thought_ I could use `empty()` here -- that returns an `undefined` item though (?)

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -157,17 +178,34 @@ export const launchKernelEpic = (action$: ActionsObservable<*>) =>
...
+      return merge(
         of({
           type: ERROR_KERNEL_LAUNCH_FAILED,
           payload: error,
           error: true
         }),
         source
-      )
-    )
+      );
```
rgbkrk: We _do_ have an action creator for the above, I should adapt that.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +256,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+    ofMessageType("shutdown_reply"),
+    first(),
+    // If we got a reply, great! :)
+    mergeMap(msg =>
+      from([
+        {
+          // TODO: Do we want these as intermediate actions with follow on epics
+          //       or should we just merge with the stream below?
+          type: "SHUTDOWN_REPLY_SUCCEEDED",
+          payload: msg.content
```
rgbkrk: Since I was unsure about how we'd want these setup, I've just made plain actions. What do you think about this stream @theengineear? (we can make action creators after we scope out what this lifecycle looks like)

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +256,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+      ])
+    ),
+    // If we don't get a response within 2s, assume failure :(
+    timeout(1000 * 2),
+    catchError(err =>
+      of({
+        type: "SHUTDOWN_REPLY_EXCEEDED_TIMEOUT",
+        payload: err,
+        error: true
+      })
```
rgbkrk: Another action to think about.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +256,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+
+      // Kill the process fully
+      kernel.spawn.kill("SIGKILL");
+
+      // Delete the connection file
+      const del$ = unlinkObservable(kernel.connectionFile).pipe(
+        mapTo({ type: "DELETE_CONNECTION_FILE_SUCCESSFUL" }),
+        catchError(err =>
+          of({
+            type: "DELETE_CONNECTION_FILE_FAILED",
```
rgbkrk: These actions are more for transparency while debugging. At this point, we _know_ the kernel has been formally cleaned up if it reached here (or so we hope).

rgbkrk:
Reference of another app that did a workaround for app closing: https://github.com/xodio/xod/pull/838

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: I'm trying to understand why `empty` would not work here. I thought it would be something related to empty immediately terminating, but I don't think that matters (or would be different from `of` in this case) since you're `merge`-ing at the end here.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +251,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+    // If we don't get a response within 2s, assume failure :(
+    timeout(1000 * 2),
+    catchError(err =>
+      of({
+        type: "SHUTDOWN_REPLY_EXCEEDED_TIMEOUT",
+        payload: err,
+        error: true
+      })
+    ),
+    mergeMap(action => {
```
theengineear: This sorta feels like a distinct thing to me. Why not just have another, small epic that listens to either of the two actions that are spit out from the above? I feel like that would be easier to reason about.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +251,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+          payload: msg.content
+        },
+        setExecutionState("shutting down")
+      ])
+    ),
+    // If we don't get a response within 2s, assume failure :(
+    timeout(1000 * 2),
+    catchError(err =>
+      of({
+        type: "SHUTDOWN_REPLY_EXCEEDED_TIMEOUT",
```
theengineear: This might be misleading, it's not necessarily the timeout that caused the error, right?

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: I commented on this below.

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: I think just `SHUTDOWN_REPLY_FAILED` is more true?

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +251,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+        payload: err,
+        error: true
+      })
+    ),
+    mergeMap(action => {
+      // End all communication on the channels
```
theengineear: awesome commenting 💯 ⭐️

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +251,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
       );
     })
   );
+
+function killKernelObservable(kernel) {
```
theengineear: Should this be named something like `createKillKernelObservable`?

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -218,3 +251,112 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+        kernel.type === "zeromq" &&
+        kernel.spawn &&
+        kernel.channels
+      );
+    }),
+    concatMap(action => {
+      const app = store.getState().app;
+      const kernel = app.kernel;
+
+      return killKernelObservable(kernel);
```
theengineear: TBH this code was a little hard to grok for me. It's nice that it is all packed away into a separate function here, but I'm so used to the notion of `actions in, actions out` that having a function that is `kernel in, actions out` feels weird. It might be my lack of familiarity though. I think we should probably keep it if it's working in the interest of time, but we might want to discuss further.

packages/core/src/middlewares.js
```diff
@@ -13,9 +13,15 @@ export const errorMiddleware = (
   store: any,
   console: Console = global.console
 ) => (next: any) => (action: Action) => {
-  if (!action.type.includes("ERROR")) {
+  if (!action || !action.type) {
```
theengineear: huh, i'm surprised that nothing else complains before this. isn't it expected that an action `MUST` have a type? I'm worried about being overly-defensive here. What's the rationale for this?

theengineear:
I haven't run this yet. At some point this weekend I'll pull this down and run things to take a deeper dive on this!

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: Yeah that makes sense. I could call it `killKernel` and we'd know the return type is an observable based on usage here?

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: 🤔  There might be a better way to group these so that we catch timeout errors separate from any other errors.

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: When I was using `empty`, I was getting an `undefined` action passed to redux. I wasn't able to trace where it was coming from and it cleaned up after switching out the `empty`. 
¯\\\_(ツ)\_/¯

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: I guess as long as we track the types of ways it fails well. In my limited testing with these, I've been able to get the python kernel to not respond to the shutdown request some of the time. At least, I only care about this for the sake of debugging kernels (and informing users where possible). We'll have the error, either way, I just wouldn't want to have to parse the error when we already have the information.

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: Heh, I made this a separate function then realized I needed the `kernel` in scope... so I cheated and moved this back in.

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: This is really useful feedback!

The reasons that I made a separate function for this include:

* Needing to handle killing of kernels in a different way when
  * there's a new kernel being launched
  * the user requests to _just_ kill the current kernel (this is likely when an interrupt just _isn't_ cutting it)
* The fact that we don't have refs to keep track of kernels yet 😅

applications/desktop/src/notebook/global-events.js
```diff
@@ -5,28 +5,35 @@ import type { AppState } from "@nteract/types/core/records";
...
+    //       e.returnValue to provide information about unload to the main process
+    //       See https://github.com/electron/electron/blob/32a1395bcfd6db9cb3eb35e8072caca085be38c1/docs/api/browser-window.md#event-close
+    // $FlowFixMe: this is an Electron-only API for unload response
     e.returnValue = true;
+    e.preventDefault();
+    // TODO: we also need to clean the kernel if the user decides to exit without saving
+  } else {
+    // Notebook in memory matches that of what we last saved, can exit
+    // At this point, we issue kernel cleanup and set the state to "clean"
+    store.dispatch({ type: "EXIT" });
```
rgbkrk: This segment is simultaneously killing me and delaying the ease of finishing this PR. 😰

packages/core/src/middlewares.js
rgbkrk: This centers around that issue I described above with `empty()` where I was getting an `undefined`. I safe guarded this just so things didn't bork out while I was debugging.

packages/core/src/reducers/app.js
```diff
@@ -42,8 +42,7 @@ export function launchKernelSuccessful(
       break;
     case "zeromq":
       kernel = makeLocalKernelRecord(action.kernel);
-    default:
-      kernel = null;
+      break;
```
rgbkrk: 😅  This was a bug we didn't experience yet because we were using a separate reducer in desktop.

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: ^^ yah, that's fine as well

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: ah, gotcha, that's fine 👌

packages/core/src/middlewares.js
theengineear: Ah, ok, is it going to stay in the PR or get yanked before we merge?

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: Yeah once we have `ref`s we can break this apart.

packages/core/src/middlewares.js
rgbkrk: I can yank it out. It added some additional runtime safety, no big deal though.

applications/desktop/__tests__/renderer/epics/kernel-launch-spec.js
```diff
@@ -29,34 +29,58 @@ describe("launchKernelEpic", () => {
...
+    expect(responses).toEqual([
+      {
+        type: actionTypes.SET_KERNEL_INFO,
+        kernelInfo: { spec: "hokey", name: "woohoo" }
+      },
+      {
+        type: actionTypes.LAUNCH_KERNEL_SUCCESSFUL,
+        kernel: {
+          type: "zeromq",
+          channels: expect.anything(),
```
theengineear, line 72: oh wow. How did i not know about this? `expect.anything()` 💥

rgbkrk: Another cool one is `expect.any` -- I probably could have made this `expect.any(Subject)`, though I wasn't sure.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -157,20 +176,24 @@ export const launchKernelEpic = (action$: ActionsObservable<*>) =>
...
-          type: ERROR_KERNEL_LAUNCH_FAILED,
-          payload: error,
-          error: true
-        }),
-        source
-      )
-    )
+    // We must kill the previous kernel now
+    // Then launch the next one
+    mergeMap(action => {
```
theengineear: In general, I actually think I have a bunch more questions, but thinking about whether or not this should really be a `mergeMap` is what kicked off some concerns for me. Like, what happens if you launch the same kernel twice, is it the case that `kernel` will be set correctly on the second one? What if it's still in the process of being killed? What if it hasn't been connected to yet... then can we even kill it?

I'm wondering if we should have more guard rails here?

rgbkrk: Total cop out answer: The users are not fast enough with the desktop ui to launch a kernel while already launching a kernel \*

\* This is totally a lie, if the first kernel is slower to start and the second kernel is faster to start, you'd definitely have this problem

I only think we can get around this by having `ref`s. You're probably right though that this should be a `concatMap` instead.

theengineear (approved):

PR #2472, opened 2018-02-03
flow type the main process for desktop

rgbkrk:
This _very minimally_ flow types the main process files.

Note: I skipped the CLI, as I wasn't sure if that had to not be transpiled.


rgbkrk:
Also: I've only done limited testing to make sure that the changes I made to appease flow didn't break things. I'll come back to this PR to double check.

rgbkrk:
Lookin' good, erring on the side of getting this out so I can continue refactors of the main process with some sense of sanity.

PR #2485, opened 2018-02-06
fix: missed var() calls due to rebase timing.

theengineear:
The help/about PR got merged while the styles provider PR was being
worked on. Therefore, these custom CSS properties didn't get renamed.


rgbkrk:
Rock in'!

PR #2491, opened 2018-02-07
pin to an exact electron version

rgbkrk:
The `1.8.x` releases just came out for electron and they're currently not working with our `install-app-deps` step, so something has to be fixed up _somewhere_. For the sake of `master` allowing launch, I'm just pinning this for now.

/cc @lgeiger

### DIFF
--- a/applications/desktop/package.json
+++ b/applications/desktop/package.json
@@ -128,7 +128,7 @@
     "commonmark-react-renderer": "^4.3.3",
     "cross-env": "^5.1.3",
     "date-fns": "^1.29.0",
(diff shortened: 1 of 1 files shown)


lgeiger:
zeromq.js hast prebuilts for electron 1.8.0 that shouldn't be the problem.

rgbkrk:
Weird. Do we need underlying upgrades to zeromq in our code base / jmp then? I was getting module version mismatches.

rgbkrk:
So happy to say we good now. 😄

PR #2493, opened 2018-02-07
Update how contents api should be kept in memory.

theengineear:
The major change is that we had previously only really considered the
notebook-type, whereas a file-browsing app needs to let notebooks,
directories, and text files _all_ be first-class citizens.

### DIFF
--- a/packages/types/core/plan.md
+++ b/packages/types/core/plan.md
@@ -55,17 +55,15 @@ opaque type Id = string;
 opaque type Ref = string;
 
 type state = {
-  // The top level of core state can be considered a notebook, but we formalize
-  // that by nesting it under `notebook`..
-  notebook: {
+  // The top level of core state can be considered a entry (which can be a 
+  // "notebook", "directory", or plain "file").
+  app: {
       // On desktop we'll have the one built-in local host that connects to
       // zeromq directly. On jupyterhub backed apps, you'll be able to switch to
       // different hosts.
       selectedHostRef: Ref,
       hostRefs: Array<Ref>,
-      lastSaved: Date
-      
-      // TODO: language information? Does this belong at the notebook level?
+      selectedContentRef: Ref
   },
   
   preferences: {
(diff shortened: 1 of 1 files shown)


packages/types/core/plan.md
```diff
@@ -55,17 +55,15 @@ opaque type Id = string;
 opaque type Ref = string;
 
 type state = {
-  // The top level of core state can be considered a notebook, but we formalize
-  // that by nesting it under `notebook`..
-  notebook: {
+  // The top level of core state can be considered a entry (which can be a 
+  // "notebook", "directory", or plain "file").
+  app: {
```
theengineear, line 60: `app`? I didn't really know what to call this, it's really the top-most part of the `core` state. The alternative would be to assume that *all* of this is wrapped in `core` and so we don't need an `app` wrapper:

```
state = {
  core: {
    selectedContentRef: Ref,
    ...
    preferences: { .. },
    communication: { .. },
    ...
  }
}
```

packages/types/core/plan.md
```diff
@@ -229,10 +227,17 @@ type state = {
...
-          format: null | "json" | "text" | "base64" // "json" for dir / nb
-          content: ?Array<Ref> | string // Array<Ref> is for directory-type
+          format: null | "json" | "text" | "base64", // "json" for dir / nb
+          // The model is a little confusing. Think of it as the in-memory, app
+          // version of the content string that you get back from the contents
+          // api. So, for a plain file, which we don't necessarily know how to
+          // handle, the model will just be a string still. However, for a
+          // notebook, we basically flesh out all the references to cells in
+          // here.
+          model: ?Object, // null | DirectoryModel | NotebookModel | FileModel
```
theengineear, line 240: I didn't want to call it `content` since I feel like that's confusing the strings returned from the contents api with our in-app representation... thoughts? Also, it's nullable since we might not have the information if we asked for only `type` and `filename` from the api.

rgbkrk: I think `model` is fine

packages/types/core/plan.md
rgbkrk: I think it's cool at the top of core.

PR #2494, opened 2018-02-07
Small css improvements for Jupyter extension.

theengineear:
At some point I need to figure out how to get those little gaps next to the absolutely-positioned sub-menu items to disappear. Not going to mess with it right now. This just:

1. removes the `body` margin so that the menu extends across the whole page
2. removes box-shadow from the menu (I don't think it's super helpful)
3. removes radius from menu items (i think the blockier style looks better)

screen shot 2018-02-07 at 12 01 24 am


rgbkrk:
🙇  Thank you so much!

rgbkrk:
> removes the body margin so that the menu extends across the whole page

Hurrah! 🎉 

> removes box-shadow from the menu (I don't think it's super helpful)
> removes radius from menu items (i think the blockier style looks better)

Totally agree, good choices.

PR #2497, opened 2018-02-07
[WIP] move core records from types to core

### LINKED ISSUE #2408: Notebook types are wrong in @nteract/types/core/records
I just noticed that flow wasn't actually resolving the types for the records because of two things:

* `@nteract/types/core/records` resolves to the lib version
* The flow definitions are not copied for the records

It's not a simple switch though -- as soon as I "fixed" that, I found an overwhelming amount of flow errors to get through. Most of the issue is that `Immutable.Map`s are not used for the notebook when they should be (up until we switch to `Immutable.Record` in `commutable`. Honestly though, the records should be relying on commutable's types if at all possible.

rgbkrk:
Will close #2408.

Very much a work in progress.


rgbkrk:
Ruh roh, I think I've unraveled some major typing issues. I'm going to make a new approach that doesn't move core records for a moment.

rgbkrk:
Closing in favor of #2516 and other refinements.

PR #2499, opened 2018-02-07
Return block math if inline.js finds double $

tlyon3:
The remark `inline.js` was finding `$$` without a new line and returning an inline math. Both `$$...$$` and `$$\n...\n$$` should return the same thing. This fixes the issue so that they both return a block math div.

* Added a check to see if it found `$$` and returned the block math instead of inline.
* Updated tests to match the changes to `inline.js`

### DIFF
--- a/packages/markdown/__tests__/remark-math-spec.js
+++ b/packages/markdown/__tests__/remark-math-spec.js
@@ -7,7 +7,9 @@ const u = require("unist-builder");
 
 function remark() {
   return unified()
-    .use(parse, { position: false })
+    .use(parse, {
+      position: false
+    })
     .use(stringify);
 }
 
@@ -211,7 +213,7 @@ describe("remark-math", () => {
   it("must not set inlineMathDouble class", () => {
     const processor = remark().use(math);
 
-    const targetText = ["$$\\alpha$$"].join("\n");
+    const targetText = "$$\\alpha$$";
 
     const ast = processor.parse(targetText);
 
@@ -225,7 +227,7 @@ describe("remark-math", () => {
                 hChildren: [u("text", "\\alpha")],
                 hName: "div",
                 hProperties: {
-                  className: "math inlineMathDouble"
+                  className: "math"
                 }
               }
             },
@@ -241,13 +243,14 @@ describe("remark-math", () => {
       inlineMathDouble: true
     });
 
-    const targetText = ["$$\\alpha$$"].join("\n");
+    const targetText = "hello $$\\alpha$$ world";
 
     const ast = processor.parse(targetText);
 
     expect(ast).toEqual(
       u("root", [
         u("paragraph", [
+          u("text", "hello "),
           u(
             "math",
             {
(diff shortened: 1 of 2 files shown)


mpacer:
shouldn't this change the results of some of the jest tests? If not, we should cover this test case in them.

tlyon3:
@mpacer yeah it does. I updated the test cases to make sure that $$ always returns a block math, but you might want to double check them. You can see the changes I made in the commit

mpacer:
Should the below actually produce a math object? I’m not sure… I think that would be invalid LaTeX not sure about mathjax…

```
  $$$
    \alpha
  $$$
```

mpacer:
Why do https://github.com/tlyon3/nteract/blob/1ad8dfb268e880067b49d9eca54067d6e8e4fffd/packages/markdown/__tests__/remark-math-spec.js#L214 and https://github.com/tlyon3/nteract/blob/1ad8dfb268e880067b49d9eca54067d6e8e4fffd/packages/markdown/__tests__/remark-math-spec.js#L146 differ?


Could you include additional text  surrounding the math in the inline mathblock cases to indicate that they are being treated appropriately? 

I really don’t like remarkmaths distinction between inlinemathdouble vs math block. In LaTeX they’re the same, & in mathjax they’re the same. I think I’d prefer us to keep to exactly one way to interpret math blocks rather than following down that path.

mpacer:
Could you include more complicated mathematical expressions that conflict with markdown (e.g., including underscores for subscripts vs italics). I’ve found this to be a likely point of parsing errors in the classic notebook, so it’s worth including.

mpacer:
Could you add some tabular examples that have dollar signs inside the contents? That’s also a known place for parsing conflicts.

packages/markdown/__tests__/remark-math-spec.js
```diff
@@ -148,7 +150,7 @@ describe("remark-math", () => {
     const ast = processor.parse(targetText);
 
     expect(ast).toMatchObject(
-      u("root", [u("paragraph", [u("inlineMath", "\\alpha")])])
+      u("root", [u("paragraph", [u("math", "\\alpha")])])
```
mpacer, line 153: Is this equivalent to https://github.com/nteract/nteract/pull/2499/files#diff-44b50e8eee5aee1583efa30b48d59d94R221  because it uses `toMatchObject` vs. `toEqual`? I think I need to learn more about jest tests and their primitives, because that seems… strange since now they're literally the same string.

mpacer: From the [jest docs](https://facebook.github.io/jest/docs/en/expect.html#tomatchobjectobject): 

> Use `.toMatchObject` to check that a JavaScript object matches a subset of the properties of an object. It will match received objects with properties that are not in the expected object.

> Use .toEqual when you want to check that two objects have the same value. This matcher recursively checks the equality of all fields, rather than checking for object identity—this is also known as "deep equal". For example, toEqual and toBe behave differently in this test suite, so all the tests pass:

So I think this test is redundant with the test below, which is stricter.

packages/markdown/__tests__/remark-math-spec.js
```diff
@@ -241,28 +243,98 @@ describe("remark-math", () => {
...
-            "inlineMath",
+            "math",
             {
               data: {
                 hChildren: [u("text", "\\alpha")],
-                hName: "span",
+                hName: "div",
                 hProperties: {
-                  className: "inlineMath inlineMathDouble"
+                  className: "math inlineMathDouble"
```
mpacer: Why does this need to set the inlineMathDouble class? Do we really want to have different meanings for math blocks depending on whether they're surrounded by text?

mpacer: Note 

https://github.com/Rokt33r/remark-math#inlinemathdouble-experimental-use-with-caution

They use the `inlineMathDouble` class to keep the same paragraph but to display it as though it were a different paragraph. Notice the spacing for the two cases is different:

image

In that case (from their docs), it doesn't look too bad because the text doesn't continue, but if you were to extend the text out it would be too close to the math. 

Also, such a convention is a deviation from classic LaTeX and (I think) MathJax, and unless people feel strongly about it I'd prefer it to treat both `$$…$$` blocks in the same way. From this perspective, displayed math shouldn't be in the same paragraph.

Otherwise we're right back to these actually producing different results (though they'd be more similar now than they were before):

```
$$…$$
```

```
$$…
$$
```

```
$$
…$$
```

```
$$
…
$$
```

tlyon3: The "check set inline double" test was one of the original tests that the remark guys made.

I did read another thing in their docs as to why they treat `$$..$$` and `$$\n$$` differently but it didn't seem like many people were happy with it. I also feel that they shouldn't differ, because they don't normally.

mpacer:
This looks great! 

We could remove the commented out stuff, but I'm going to merge this as is and we can do that later in a cleanup PR.

Thank you for all the hard work!

rgbkrk:
Awesome! Thanks for pulling this through @tlyon3 and @mpacer for reviewing!

PR #2500, opened 2018-02-07
More selectors

theengineear:
Jumping on the selectors train! I want to start moving state around, but I won't be confident about doing that until we centralize access with these selectors.

Couple questions:

1. naming: holy `getCurrent*`, batman! It's a little wordy! Maybe `current` should be the default thing we assume to be getting when we're using selectors.
2. `notificationStystem` stuff: not touching it, it's like an object with methods on it? I'm just going to assume we're not moving that :0
3. putting selectors to work: since we're memoizing, we might as well make selectors that do all the work for us! One example here is the `getCurrentCodeCellIdsBelow` (i know, a mouthful...). you're on board for more of the same, right?
4. In general, we should try and keep selectors as specific as possible. The more general they are, the less we can change state! I.e., selectors should be a black box. We should never return hunks of state that we think we'd want to refactor the internals of. Ya?

### DIFF
--- a/applications/desktop/__tests__/renderer/epics/github-publish-spec.js
+++ b/applications/desktop/__tests__/renderer/epics/github-publish-spec.js
@@ -1,6 +1,6 @@
 import { ActionsObservable } from "redux-observable";
 
-import { dummyStore, dummyCommutable } from "@nteract/core/dummy";
+import { dummy, dummyStore, dummyCommutable } from "@nteract/core/dummy";
 
 import { toArray } from "rxjs/operators";
 
@@ -54,7 +54,9 @@ describe("publishNotebookObservable", () => {
     const notificationSystem = createNotificationSystem();
     const publishNotebookObs = publishNotebookObservable(
       GitHub(),
-      dummyCommutable,
+      dummy,
+      "fake-github-username",
+      "fake-gist-id",
       "./test.ipynb",
       notificationSystem,
       false,
@@ -68,7 +70,9 @@ describe("publishNotebookObservable", () => {
     const notificationSystem = createNotificationSystem();
     const publishNotebookObs = publishNotebookObservable(
       GitHub(),
-      dummyCommutable,
+      dummy,
+      "fake-github-username",
+      "fake-gist-id",
       "./test.ipynb",
       notificationSystem,
       false,
@@ -88,7 +92,9 @@ describe("publishNotebookObservable", () => {
     const notificationSystem = createNotificationSystem();
     const publishNotebookObs = publishNotebookObservable(
       github,
-      dummyCommutable,
+      dummy,
+      "fake-github-username",
+      "fake-gist-id",
       "./test.ipynb",
       notificationSystem,
       false,
@@ -103,11 +109,12 @@ describe("publishNotebookObservable", () => {
   test("edits gist that is already made", done => {
     const github = GitHub();
     const store = dummyStore();
-    const notebook = dummyCommutable.setIn(["metadata", "gist_id"], "ID123");
     const notificationSystem = createNotificationSystem();
     const publishNotebookObs = publishNotebookObservable(
       github,
-      notebook,
+      dummy,
+      "fake-github-username",
+      "fake-gist-id",
       "./test.ipynb",
       notificationSystem,
       true,
@@ -140,7 +147,9 @@ describe("createGistCallback", () => {
     const notificationSystem = createNotificationSystem();
     const publishNotebookObs = publishNotebookObservable(
       github,
-      dummyCommutable,
+      dummy,
+      "fake-github-username",
+      "fake-gist-id",
       "./test.ipynb",
       notificationSystem,
       false,
--- a/applications/desktop/src/notebook/epics/github-publish.js
+++ b/applications/desktop/src/notebook/epics/github-publish.js
@@ -8,7 +8,7 @@ import {
 
 import { overwriteMetadata, deleteMetadata } from "@nteract/core/actions";
 
-import { toJS, stringifyNotebook } from "@nteract/commutable";
+import * as selectors from "@nteract/core/selectors";
 
 const path = require("path");
 
@@ -90,14 +90,14 @@ export function createGistCallback(
  */
 export function publishNotebookObservable(
   github: any,
-  notebook: any,
+  notebookString: string,
+  githubUsername?: string,
+  gistId?: string,
   filepath: string,
   notificationSystem: any,
   publishAsUser: boolean
 ) {
   return Observable.create(observer => {
-    const notebookString = stringifyNotebook(toJS(notebook));
-
     const filename = filepath ? path.parse(filepath).base : "Untitled.ipynb";
     const files = {};
     files[filename] = { content: notebookString };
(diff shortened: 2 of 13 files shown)


rgbkrk:
> naming: holy `getCurrent*`, batman! It's a little wordy! Maybe `current` should be the default thing we assume to be getting when we're using selectors.

`current*` is way better than `getCurrent*`, especially since every selector is a `get*`.

applications/desktop/src/notebook/epics/github-publish.js
```diff
@@ -8,7 +8,7 @@ import {
 
 import { overwriteMetadata, deleteMetadata } from "@nteract/core/actions";
 
-import { toJS, stringifyNotebook } from "@nteract/commutable";
+import * as selectors from "@nterat/core/selectors";
```
rgbkrk: `from "@nteract/core/selectors"`

applications/desktop/src/notebook/epics/github-publish.js
```diff
@@ -110,10 +110,7 @@ export function publishNotebookObservable(
           message: `Authenticated as ${res.data.login}`,
           level: "info"
         });
-        if (
-          notebook.getIn(["metadata", "github_username"]) !==
-          (res.data.login || undefined)
-        ) {
+        if (githubUsername !== (res.data.login || undefined)) {
```
rgbkrk, line 113: solid cleanups in here

applications/desktop/src/notebook/epics/github-publish.js
theengineear: 🤣

theengineear:
>current* is way better than getCurrent*, especially since every selector is a get*.

Cool. Accepted 😎.

applications/desktop/__tests__/renderer/epics/github-publish-spec.js
```diff
@@ -140,7 +147,9 @@ describe("createGistCallback", () => {
     const notificationSystem = createNotificationSystem();
     const publishNotebookObs = publishNotebookObservable(
       github,
-      dummyCommutable,
+      dummy,
+      "fake-github-username",
+      "fake-gist-id",
       "./test.ipynb",
       notificationSystem,
       false,
```
rgbkrk, line 155: This should all make @jdetle happy! 🎉

packages/core/src/selectors/index.js
```diff
@@ -1,13 +1,17 @@
...
 import { createSelector } from "reselect";
+import { toggleCellInputVisibility } from "../../actions";
 
+const identity = thing => thing;
 const serverUrl = (state: AppState) => state.app.host.serverUrl;
 const crossDomain = (state: AppState) => state.app.host.crossDomain;
 const token = (state: AppState) => state.app.host.token;
 
-export const getServerConfig = createSelector(
+export const serverConfig = createSelector(
```
rgbkrk, line 12: 👍

packages/core/src/selectors/index.js
```diff
@@ -1,13 +1,17 @@
 // @flow
 import type { AppState } from "@nteract/types/core/records";
+import { toJS, stringifyNotebook } from "@nteract/commutable";
+import * as Immutable from "immutable";
 
 import { createSelector } from "reselect";
+import { toggleCellInputVisibility } from "../../actions";
 
+const identity = thing => thing;
```
rgbkrk: heh

packages/core/src/selectors/index.js
```diff
@@ -1,13 +1,17 @@
 // @flow
 import type { AppState } from "@nteract/types/core/records";
+import { toJS, stringifyNotebook } from "@nteract/commutable";
+import * as Immutable from "immutable";
 
 import { createSelector } from "reselect";
+import { toggleCellInputVisibility } from "../../actions";
```
rgbkrk: How'd an action sneak in here?

packages/core/src/selectors/index.js
```diff
@@ -16,52 +20,131 @@ export const getServerConfig = createSelector(
...
 export const isCurrentKernelJupyterWebsocket = createSelector(
-  [getCurrentHostType, getCurrentKernelType],
+  [currentHostType, currentKernelType],
   (hostType, kernelType) => {
     return hostType === "jupyter" && kernelType === "websocket";
   }
 );
+
+// TODO: if we're not looking at a notebook in the UI, there may not _be_ a
+// notebook object to get. Do we return null? Throw an error?
```
rgbkrk, line 79: I think `null` is fine

packages/core/src/selectors/index.js
```diff
@@ -16,52 +20,131 @@ export const getServerConfig = createSelector(
...
+
+export const currentSavedNotebook = createSelector(
+  (state: AppState) => state.document.get("savedNotebook"),
+  identity
+);
+
+export const currentNotebookGithubUsername = createSelector(
+  [currentNotebook],
+  notebook => notebook.getIn(["metadata", "github_username"])
+);
```
rgbkrk, line 93: Since we're having to put the selectors here for the github functionality, we might as well move the reducer from desktop's app reducer (and maybe the epic...). Not for this PR, but it'll be nice for desktop to have no custom reducers now.

packages/core/src/selectors/index.js
```diff
@@ -16,52 +20,131 @@ export const getServerConfig = createSelector(
...
+
+export const currentCodeCellIdsBelow = createSelector(
+  [currentFocusedCellId, currentCellMap, currentCellOrder],
+  (focusedCellId, cellMap, cellOrder) => {
+    const index = cellOrder.indexOf(focusedCellId);
+    return cellOrder
+      .skip(index)
+      .filter(id => cellMap.getIn([id, "cell_type"]) === "code");
+  }
+);
```
rgbkrk, line 137: At some point we'll want to re-evaluate how we do run-all and run-all-below, since they belong in an epic that tracks the execution rolling on through.

packages/core/src/selectors/index.js
rgbkrk: Looks like this didn't mean to be here

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -70,16 +67,10 @@ export function saveEpic(
...
-  action$: ActionsObservable<*>,
-  store: Store<any, any>
-) {
+export function saveAsEpic(action$: ActionsObservable<*>) {
   return action$.pipe(
     ofType(SAVE_AS),
     mergeMap(action => {
-      const state = store.getState();
-      const notebook = state.document.get("notebook");
-
```
rgbkrk, line 82: Ooooohhhh cool, we didn't need the notebook here anymore.

rgbkrk (changes requested):

rgbkrk:
Separately, we should extract `getCodeMirrorMode` out as a selector too: https://github.com/nteract/nteract/blob/master/packages/core/src/providers/notebook-app.js#L295

rgbkrk:
We'll probably want to switch a lot of our `mapStateToProps` to using these as well.

packages/core/src/selectors/index.js
theengineear: ah, didn't even realize that these were being set in desktop.

applications/desktop/src/notebook/global-events.js
```diff
@@ -23,8 +22,8 @@ export function unload(store: Store<AppState, Action>) {
 export function beforeUnload(store: Store<AppState, Action>, e: any) {
   const state = store.getState();
   const saved = is(
-    state.document.get("notebook"),
-    state.document.get("savedNotebook")
+    selectors.currentNotebook(state),
+    selectors.currentSavedNotebook(state)
   );
```
rgbkrk, line 27: This could be an `isSaved` selector.

packages/core/src/selectors/index.js
rgbkrk: That or `emptyNotebook` (from `commutable`)

rgbkrk (approved):

PR #2501, opened 2018-02-07
Extract and confirm types from @nteract/types to flow declares

rgbkrk:
Corrects messaging, channels, and redux types. They're now project global types to be used anywhere and everywhere.


rgbkrk:
Since I'm in the midst of other fixes, I think I'll also fix the other types while I'm at it in this PR.

Edit: not the records though

flow-typed/channels.js
```diff
@@ -24,7 +25,9 @@ export type ExecuteMessageContent = {
   stop_on_error: boolean
 };
 
-export type ExecuteRequest = JupyterMessage<
+declare type ExecuteRequest = JupyterMessage<
   "execute_request",
   ExecuteMessageContent
 >;
+
+declare type Channels = rxjs$Subject<JupyterMessage<*, *>>;
```
theengineear, line 33: 👍

flow-typed/npm/rxjs_v5.0.x.js
```diff
@@ -638,6 +638,90 @@ declare class rxjs$Observable<+T> {
...
+    op4: rxjs$OperatorFunction<C, D>,
+    op5: rxjs$OperatorFunction<D, E>,
+    op6: rxjs$OperatorFunction<E, F>,
+    op7: rxjs$OperatorFunction<F, G>,
+    op8: rxjs$OperatorFunction<G, H>,
+    op9: rxjs$OperatorFunction<H, I>,
+    op10: rxjs$OperatorFunction<I, J>,
+    op11: rxjs$OperatorFunction<J, K>,
+    op12: rxjs$OperatorFunctionLast<K, L>
+  ): L;
```
theengineear, line 723: 🙃

packages/core/src/epics/execute.js
```diff
@@ -146,7 +142,7 @@ export function createExecuteCellStream(
     channels &&
     !(kernel.status === "starting" || kernel.status === "not connected");
 
-  if (!kernelConnected) {
+  if (!kernelConnected || !channels) {
```
theengineear, line 145: 💯 flow caught this?

flow-typed/base.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+
+declare type PrimitiveImmutable = string | number | boolean | null;
+declare type JSONType = PrimitiveImmutable | JSONObject | JSONArray; // eslint-disable-line no-use-before-define
+declare type JSONObject = { [key: string]: JSONType };
+declare type JSONArray = Array<JSONType>;
```
theengineear, line 6: I almost wonder if we should namespace these types... is there any scenario where we run into possible collisions?

packages/types/src/commutable.js
```diff
@@ -1,19 +1,11 @@
...
-const Immutable = require("immutable");
-const ImmutableMap = Immutable.Map;
-const ImmutableList = Immutable.List;
-
-type PrimitiveImmutable = string | number | boolean | null;
-
-export type JSONType = PrimitiveImmutable | JSONObject | JSONArray; // eslint-disable-line no-use-before-define
-export type JSONObject = { [key: string]: JSONType };
-export type JSONArray = Array<JSONType>;
+import type { Map, List } from "immutable";
```
theengineear, line 3: Should we normalize this to `import * as Immutable from "immutable"`?

packages/types/src/commutable.js
```diff
@@ -1,19 +1,11 @@
 // @flow
 
-const Immutable = require("immutable");
-const ImmutableMap = Immutable.Map;
-const ImmutableList = Immutable.List;
```
theengineear, line 5: 🎉 😁

theengineear (approved):

packages/core/src/epics/execute.js
rgbkrk: Yup!

flow-typed/base.js
theengineear: ^^ nvm, these won't ever bleed out of our repo, so it's not an issue.

rgbkrk: 🎉

packages/types/src/commutable.js
rgbkrk: We can't do that with (direct) flow types (there aren't utility types like the React ones here).

PR #2505, opened 2018-02-08
Unsubscribe fromEvents on unmount in editor

captainsafia:
Closes part of #2271.

Does this right @rgbkrk?


rgbkrk:
Yeah I think that's perfect, I don't really know any alternatives.

rgbkrk (approved):

rgbkrk:
That is, other than figuring out some way that we'd do it through our state tree similar to modals. I think this is wayyyy less complicated though. I'll assume this works well and just merge it on in.

Issue #2506, opened 2018-02-08
[core] Sticky cells pass focus to next on shift-enter


theengineear:
I noticed this when I needed to refactor a test in the `notebook-spec.js` file.

## Expected behavior

When you `shift+enter` in a sticky cell, it should execute in place and not focus the next cell.

## Actual behavior

Not ^^.

theengineear:
@rgbkrk I wanted to link an issue in a PR I'm about to open. Have you noticed this one yet?

rgbkrk:
I've definitely noticed this one. I think it would be far easier for us to tear out the sticky cells and come back to them as a separate feature someday -- docked (and mirrored) output or something similar.

theengineear:
Yah, I think ripping it out would be a good idea. Any idea how much backlash there would be?

rgbkrk:
From https://github.com/nteract/nteract/pull/2507#discussion_r167042929:

Yes! We ran [an informal poll on twitter](https://twitter.com/nteractio/status/918636485263405056) in October of 2017.

----

_Do you use "pin cell" in nteract desktop?_

|  # | % | Choice | 
|---|---|------|
| 5 | 26.3% | Never, it's so not for me  |
| 5 | 26.3% | Once for funsies |
| **7** | 36.8% | **Hardly ever** |
| 2 | 10.5% | All the time, it's my jam |

**19 respondents**

----

I'm willing to say, based on those results, that it would be ok to remove it in favor of maintainability. We can always keep the component + styling in core, just rip it out of the standalone notebook app.

rgbkrk:
Sticky cells are now out so this is no longer relevant.

PR #2507, opened 2018-02-08
Selectors everywhere

theengineear:
In response to some comments @rgbkrk left in #2500 I used selectors in `mapStateToProps`. I also couldn't help myself and added a commit in here that updates some usage of `mapDispatchToProps`. I _think_ the latter change is an overall improvement. However, I don't know how cleverly/simply the `mapDispatchToProps` arg is handled in `connect`. It could potentially be a perf issue to re-create all the functions in `mapDispatchToProps`... just wanted to draw that out.

Also, I _think_ I found a test that was yielding false positives. You'll see it skipped in the diff. Here's an issue I opened about it:

https://github.com/nteract/nteract/issues/2506

Just to reiterate, the goal of these refactors are to ensure that our code doesn't depend on the structure of our state 🌲. We want to reorganize the state and add useful things like app `ref`s to our state (see the `plan.md` file for details on that).

### DIFF
--- a/packages/core/__tests__/providers/notebook-spec.js
+++ b/packages/core/__tests__/providers/notebook-spec.js
@@ -7,10 +7,7 @@ import { shallow, mount } from "enzyme";
 import renderer from "react-test-renderer";
 
 import { displayOrder, transforms } from "@nteract/transforms";
-import {
-  NotebookApp,
-  getCodeMirrorMode
-} from "../../src/providers/notebook-app";
+import { NotebookApp } from "../../src/providers/notebook-app";
 
 import { dummyStore, dummyCommutable } from "../../src/dummy";
 
@@ -47,20 +44,6 @@ describe("NotebookApp", () => {
     expect(component).not.toBeNull();
   });
 
-  describe("getCodeMirrorMode", () => {
-    test("determines the right mode from the notebook metadata", () => {
-      const mode = getCodeMirrorMode(dummyCommutable.get("metadata"));
-      expect(mode).toEqual(Immutable.fromJS({ name: "ipython", version: 3 }));
-
-      const lang2 = getCodeMirrorMode(
-        dummyCommutable
-          .setIn(["metadata", "language_info", "codemirror_mode", "name"], "r")
-          .get("metadata")
-      );
-      expect(lang2).toEqual(Immutable.fromJS({ name: "r", version: 3 }));
-    });
-  });
-
   describe("keyDown", () => {
     test("detects a cell execution keypress", () => {
       const focusedCell = dummyCommutable.getIn(["cellOrder", 1]);
@@ -68,7 +51,7 @@ describe("NotebookApp", () => {
       const context = { store: dummyStore() };
 
       context.store.dispatch = jest.fn();
-
+      const executeFocusedCell = jest.fn();
       const component = shallow(
         <NotebookApp
           cellOrder={dummyCommutable.get("cellOrder")}
@@ -78,6 +61,7 @@ describe("NotebookApp", () => {
           cellStatuses={dummyCellStatuses}
           stickyCells={new Immutable.Set()}
           cellFocused={focusedCell}
+          executeFocusedCell={executeFocusedCell}
         />,
         { context }
       );
@@ -90,9 +74,7 @@ describe("NotebookApp", () => {
 
       inst.keyDown(evt);
 
-      expect(context.store.dispatch).toHaveBeenCalledWith({
-        type: "EXECUTE_FOCUSED_CELL"
-      });
+      expect(executeFocusedCell).toHaveBeenCalled();
     });
     test("detects a focus to next cell keypress", () => {
       const focusedCell = dummyCommutable.getIn(["cellOrder", 1]);
@@ -100,7 +82,9 @@ describe("NotebookApp", () => {
       const context = { store: dummyStore() };
 
       context.store.dispatch = jest.fn();
-
+      const executeFocusedCell = jest.fn();
+      const focusNextCell = jest.fn();
+      const focusNextCellEditor = jest.fn();
       const component = shallow(
         <NotebookApp
           cellOrder={dummyCommutable.get("cellOrder")}
@@ -110,6 +94,9 @@ describe("NotebookApp", () => {
           cellStatuses={dummyCellStatuses}
           stickyCells={new Immutable.Set()}
           cellFocused={focusedCell}
+          executeFocusedCell={executeFocusedCell}
+          focusNextCell={focusNextCell}
+          focusNextCellEditor={focusNextCellEditor}
         />,
         { context }
       );
@@ -122,17 +109,21 @@ describe("NotebookApp", () => {
 
       inst.keyDown(evt);
 
-      expect(context.store.dispatch).toHaveBeenCalledWith({
-        type: "FOCUS_NEXT_CELL_EDITOR"
-      });
+      expect(executeFocusedCell).toHaveBeenCalled();
+      expect(focusNextCell).toHaveBeenCalled();
+      expect(focusNextCellEditor).toHaveBeenCalled();
     });
-    test("handles a focus to next cell keypress on a sticky cell", () => {
+
+    // TODO: This test was silently broken. It was loudly found during a refact.
+    test.skip("handles a focus to next cell keypress on a sticky cell", () => {
       const focusedCell = dummyCommutable.getIn(["cellOrder", 1]);
 
       const context = { store: dummyStore() };
 
       context.store.dispatch = jest.fn();
-
+      const executeFocusedCell = jest.fn();
+      const focusNextCell = jest.fn();
+      const focusNextCellEditor = jest.fn();
       const component = shallow(
         <NotebookApp
           cellOrder={dummyCommutable.get("cellOrder")}
@@ -142,6 +133,9 @@ describe("NotebookApp", () => {
           cellStatuses={dummyCellStatuses}
           stickyCells={new Immutable.Set([focusedCell])}
           cellFocused={focusedCell}
+          executeFocusedCell={executeFocusedCell}
+          focusNextCell={focusNextCell}
+          focusNextCellEditor={focusNextCellEditor}
         />,
         { context }
       );
@@ -154,11 +148,9 @@ describe("NotebookApp", () => {
 
       inst.keyDown(evt);
 
-      expect(context.store.dispatch).not.toHaveBeenCalledWith({
-        type: "FOCUS_NEXT_CELL",
-        id: focusedCell,
-        createCellIfUndefined: true
-      });
+      expect(executeFocusedCell).toHaveBeenCalled();
+      expect(focusNextCell).not.toHaveBeenCalled();
+      expect(focusNextCellEditor).not.toHaveBeenCalled();
     });
   });
 });
--- a/packages/core/__tests__/selectors-spec.js
+++ b/packages/core/__tests__/selectors-spec.js
@@ -1,0 +1,27 @@
+import Immutable from "immutable";
+import { dummyCommutable } from "../src/dummy";
+import * as selectors from "../src/selectors";
+
+describe("codeMirrorMode", () => {
+  test("determines the right mode from the notebook metadata", () => {
+    // TODO: better way to get dummy state?
+    const state1 = {
+      document: Immutable.Map({
+        notebook: dummyCommutable
+      })
+    };
+    const mode = selectors.codeMirrorMode(state1);
+    expect(mode).toEqual(Immutable.fromJS({ name: "ipython", version: 3 }));
+
+    const state2 = {
+      document: Immutable.Map({
+        notebook: dummyCommutable.setIn(
+          ["metadata", "language_info", "codemirror_mode", "name"],
+          "r"
+        )
+      })
+    };
+    const lang2 = selectors.codeMirrorMode(state2);
+    expect(lang2).toEqual(Immutable.fromJS({ name: "r", version: 3 }));
+  });
+});
--- a/packages/core/src/components/notebook-menu/index.js
+++ b/packages/core/src/components/notebook-menu/index.js
@@ -5,6 +5,7 @@ import { localCss } from "./styles";
 import { connect } from "react-redux";
 import * as Immutable from "immutable";
 import * as actions from "../../actions";
+import * as selectors from "../../selectors";
 import { MENU_ITEM_ACTIONS, MENUS } from "./constants";
 import * as extraHandlers from "./extra-handlers";
 import { MODAL_TYPES } from "../modal-controller";
@@ -286,11 +287,11 @@ class PureNotebookMenu extends React.Component<Props> {
 // information about the current document to decide which menu items are
 // available...
 const mapStateToProps = state => ({
-  cellFocused: state.document.cellFocused,
-  cellMap: state.document.getIn(["notebook", "cellMap"]),
-  cellOrder: state.document.getIn(["notebook", "cellOrder"]),
-  filename: state.document.get("filename"),
-  notebook: state.document.get("notebook")
+  cellFocused: selectors.currentFocusedCellId(state),
+  cellMap: selectors.currentCellMap(state),
+  cellOrder: selectors.currentCellOrder(state),
+  filename: selectors.currentFilename(state),
+  notebook: selectors.currentNotebook(state)
 });
 
 const mapDispatchToProps = dispatch => ({
--- a/packages/core/src/epics/contents.js
+++ b/packages/core/src/epics/contents.js
@@ -1,5 +1,5 @@
 /* @flow */
-
+import { empty } from "rxjs/observable/empty";
 import { of } from "rxjs/observable/of";
 import {
   tap,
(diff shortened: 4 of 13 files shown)


packages/core/__tests__/providers/notebook-spec.js
```diff
@@ -122,17 +109,21 @@ describe("NotebookApp", () => {
...
-        type: "FOCUS_NEXT_CELL_EDITOR"
-      });
+      expect(executeFocusedCell).toHaveBeenCalled();
+      expect(focusNextCell).toHaveBeenCalled();
+      expect(focusNextCellEditor).toHaveBeenCalled();
     });
-    test("handles a focus to next cell keypress on a sticky cell", () => {
+
+    // TODO: This test was silently broken. It was loudly found during a refact.
+    test.skip("handles a focus to next cell keypress on a sticky cell", () => {
```
rgbkrk, line 118: Yeah, that's fine for now. We probably want to tear out the sticky cells at this point.

packages/core/__tests__/selectors-spec.js
```diff
@@ -0,0 +1,27 @@
+import Immutable from "immutable";
+import { dummyCommutable } from "../src/dummy";
+import * as selectors from "../src/selectors";
+
+describe("codeMirrorMode", () => {
+  test("determines the right mode from the notebook metadata", () => {
+    // TODO: better way to get dummy state?
```
rgbkrk, line 7: On some level, I think it's ok to only use the state we're actually using for the selectors.

packages/core/src/components/notebook-menu/index.js
```diff
@@ -286,11 +287,11 @@ class PureNotebookMenu extends React.Component<Props> {
 // information about the current document to decide which menu items are
 // available...
 const mapStateToProps = state => ({
-  cellFocused: state.document.cellFocused,
-  cellMap: state.document.getIn(["notebook", "cellMap"]),
-  cellOrder: state.document.getIn(["notebook", "cellOrder"]),
-  filename: state.document.get("filename"),
-  notebook: state.document.get("notebook")
+  cellFocused: selectors.currentFocusedCellId(state),
+  cellMap: selectors.currentCellMap(state),
```
rgbkrk, line 291: Refresh my memory, why does the notebook-menu need the whole `cellMap` and `cellOrder`?

packages/core/src/epics/contents.js
```diff
@@ -80,6 +80,12 @@ export function saveContentEpic(
       const state = store.getState();
       const currentNotebook = selectors.currentNotebook(state);
 
+      // TODO: this will likely make more sense when this becomes less
+      // notebook-centric.
+      if (!currentNotebook) {
```
rgbkrk, line 86: We could / probably should emit an error rather than the empty.

packages/core/src/epics/contents.js
```diff
@@ -80,6 +80,12 @@ export function saveContentEpic(
...
+        return of(empty());
```
rgbkrk: Aren't you able to return the `empty()` observable as is?

packages/core/src/epics/execute.js
```diff
@@ -202,6 +202,14 @@ export function executeCellEpic(action$: ActionsObservable<*>, store: any) {
...
         switchMap(({ id }) => {
           const cellMap = selectors.currentCellMap(store.getState());
+
+          // If for some reason this is run *outside* the context of viewing
+          // a notebook (stray async task or something), it's possible that the
+          // cellMap will be null;
+          if (!cellMap) {
+            return empty();
+          }
+
```
rgbkrk, line 212: Since we're only using cell type and cell source here, I think the selector used above could be:

```js
const source = cellSourceById(store.getState(), { id });
const type = cellTypeById(store.getState(), { id });
```

packages/core/src/providers/editor.js
```diff
@@ -21,14 +22,14 @@ function mapStateToProps(
   state: Object,
   ownProps: CodeMirrorEditorProps
 ): Object {
+  const kernel = selectors.currentKernel(state);
+  const { cursorBlinkRate } = selectors.userPreferences(state);
```
rgbkrk, line 26: Good old silly one configurable bit for codemirror here. 😝

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
 // TODO: if we're not looking at a notebook in the UI, there may not _be_ a
 // notebook object to get. Do we return null? Throw an error?
 export const currentNotebook = createSelector(
   (state: AppState) => state.document.get("notebook"),
-  identity
+  notebook => {
+    if (notebook) {
+      return notebook;
+    }
+    return null;
```
rgbkrk: This could be simplifed back to identity by making the call above be:

```
state.document.get("notebook", null)
```

packages/core/__tests__/selectors-spec.js
theengineear: Yah, the one thing I got hung up on was that it's common for use to use `fromJS` in tests, but that causes selectors to fail if they expect a `Record` and use `.` (dot) accessors. I changed all the selectors to use `get` and `getIn`, which is a pretty easy compromise, but It is a little odd. I guess records all have defaults though, so there's no reason you couldn't instantiate one in a test with *only* the bits you need filled in?

packages/core/__tests__/providers/notebook-spec.js
theengineear: Not gonna tear them out in this PR, but I'm for doing it if we think that folks will not be terribly bothered.

Speaking of knowing if folks are terribly bothered, do we have any plans on trying to get anonymized usage data from folks? It'd be cool if we could see how many people are actually sticking cells. That's a whole can of worms, but I'm curious if it's been discussed?

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
+];
+const CODE_MIRROR_MODE_DEFAULT = "text";
+export const codeMirrorMode = createSelector(
+  [currentNotebookMetadata],
+  metadata => {
+    if (metadata) {
+      return (
+        CODE_MIRROR_MODE_PATHS.reduce(
+          (mode, path) => mode || metadata.getIn(path),
+          null
```
rgbkrk: Took me a sec to realize that instead of accumulation, it's replacement of a falsy (likely `null`) value.

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
+  (state: AppState) => state.document.get("stickyCells"),
+  identity
+);
+
+export const cellPagers = createSelector(
+  (state: AppState) => state.document.get("cellPagers"),
+  identity
+);
+
+export const currentLastSaved = createSelector(
```
rgbkrk, line 111: Since I _think_ this gets used for timing, we should probably call this `currentLastSavedTimestamp` or something.

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
+  (state: AppState) => state.app.get("lastSaved"),
+  identity
+);
+
+export const currentNotebookMetadata = createSelector(
+  [currentNotebook],
+  notebook => {
```
rgbkrk: Personally, I'd aim at just doing:

```
state.document.getIn(["notebook", "metadata"], Immutable.Map())
```

so we can rely on this being a map unconditionally. It helps with a lot of the code below that uses it and I think is a fair tradeoff.

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
 );
 
 export const currentNotebookGistId = createSelector(
   [currentNotebook],
-  notebook => notebook.getIn(["metadata", "gist_id"])
+  notebook => {
+    if (notebook) {
+      return notebook.getIn(["metadata", "gist_id"]);
+    }
+    return "";
```
rgbkrk: There's disparity here in that if `["metadata", "gist_id"]` is not in the notebook, this will return undefined. If the notebook wasn't defined it returns `""`. I'd aim to return null in this one as:

```js
notebook.getIn(["metadata", "gist_id"], null)
```

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
 
-export const currentNotebookJS = createSelector([currentNotebook], notebook =>
-  toJS(notebook)
-);
+export const currentNotebookJS = createSelector([currentNotebook], notebook => {
+  if (notebook) {
+    return toJS(notebook);
+  }
+  return null;
+});
```
rgbkrk, line 152: I see what you mean by if we should throw here... If it's not set we should fail and know where it occurred. Do people normally allow for throwing in exceptions?

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
 
 export const currentNotebookString = createSelector(
   [currentNotebookJS],
-  notebookJS => stringifyNotebook(notebookJS)
+  notebookJS => {
+    if (notebookJS) {
+      return stringifyNotebook(notebookJS);
+    }
+    return "";
```
rgbkrk, line 160: ⚠️ I'm a little afraid of us accidentally saving at some `$PATH` an empty notebook. I think I'm a fan of throwing here (or in `stringifyNotebook`). Perhaps the right choice though is to handle it in the epic -- if it returns `null` for instance, we should take a different path. I guess maybe that's the right choice though -- we do the handling in the epic where we have an ability to inform the UI and the user.

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
+  (state: AppState) => state.document.get("editorFocused"),
+  identity
 );
 
-export const currentCellOrder = createSelector([currentNotebook], notebook =>
-  notebook.get("cellOrder")
+export const transientCellMap = createSelector(
+  (state: AppState) =>
+    state.document.getIn(["transient", "cellMap"], Immutable.Map()),
+  identity
```
rgbkrk, line 177: This seems like one we'll want to do as a `byRef` setup :soon:, since no component ever needs _all_ the cells.

packages/core/src/selectors/index.js
```diff
@@ -75,74 +82,194 @@ export const isCurrentKernelJupyterWebsocket = createSelector(
...
 );
 
+export const currentCellMap = createSelector([currentNotebook], notebook => {
+  if (notebook) {
+    return notebook.get("cellMap");
+  }
+  return null;
+});
```
rgbkrk, line 185: I look forward to this being replaced by a selector which is `cellByRef`

rgbkrk (changes requested):
This is so fantastic! I made a lot of comments that I think are more for future reference, so I don't intend to block this. I'll go back and note which comments we should address first (and as part of this PR).

packages/core/src/epics/contents.js
rgbkrk: Saving I definitely want to make sure we're propagating enough information back up, both for trust and transparency for users, and debuggability for us.

packages/core/src/epics/execute.js
rgbkrk: Updating the above comment to match the use of our cell ids (which of course are in fact refs).

packages/core/__tests__/providers/notebook-spec.js
rgbkrk: Yes! We ran [an informal poll on twitter](https://twitter.com/nteractio/status/918636485263405056) in October of 2017.

----

_Do you use "pin cell" in nteract desktop?_

|  # | % | Choice | 
|---|---|------|
| 5 | 26.3% | Never, it's so not for me  |
| 5 | 26.3% | Once for funsies |
| **7** | 36.8% | **Hardly ever** |
| 2 | 10.5% | All the time, it's my jam |

**19 respondents**

----

I'm willing to say, based on those results, that it would be ok to remove it in favor of maintainability. We can always keep the component + styling in core, just rip it out of the standalone notebook app.

packages/core/__tests__/selectors-spec.js
rgbkrk: Ahhhhhh yeah. I suppose the records make it easier, which means using `makeAppRecord` and `makeDocumentRecord` should make building a dummy simpler than having to change dummy on the fly (or rely on its values).

packages/core/src/components/notebook-menu/index.js
theengineear: It needs those because it needs to do `runAllCells` and such:

https://github.com/nteract/nteract/blob/master/packages/core/src/components/notebook-menu/extra-handlers.js#L19

BUT, if (_when_) we have an epic that does this for us... we can rip these out.

packages/core/src/epics/contents.js
theengineear: 😬 ah, yah, you're right, this is returning an observable of an observable, so `mergeMap` will only shell off the outer one, leaving `empty()` instead of just er... nothing. thanks!

packages/core/src/components/notebook-menu/index.js
rgbkrk: ooooooook

I'll get that into an epic since that would help both menu implementations

packages/core/src/selectors/index.js
theengineear: Good point!

packages/core/src/selectors/index.js
theengineear: K, i was just following the naming conventions in the state. I think it might be clearer to change them both at the same time later.

packages/core/src/selectors/index.js
theengineear: But it probably needs to be a `Record`?

packages/core/src/selectors/index.js
theengineear: Yah, i can clean this up.

packages/core/src/selectors/index.js
rgbkrk: Okee doke

packages/core/src/selectors/index.js
rgbkrk: Metadata is an unstructured `Immutable.Map` because we don't know what keys they're using.

packages/core/src/selectors/index.js
theengineear: I'm not sure! I think `flow` _should_ protect us in this case when we try to do _anything_ with the selector, so i guess I'd rather return *something* and then have `flow` remind us that it could be `null`? I mostly say this because the error would likely originate somewhere in `reselect` code, which feels like it might be odd to debug?

packages/core/src/selectors/index.js
theengineear: Cool. I think the moral here is to return `null` in these cases. I don't think I want to throw in the selectors...

packages/core/src/epics/contents.js
theengineear: also i double-checked this (for sanity) and yes, just `empty()` should be returned.

packages/core/src/epics/execute.js
theengineear: Gonna do this after, but 💯

packages/core/src/selectors/index.js
rgbkrk: I think returning `null` and handling it elsewhere is perfect, unless there's some sensible default (like with codemirror mode, etc.)

packages/core/src/selectors/index.js
theengineear: 🤔should *all* of these `.get` operations return a default of `null` then?

packages/core/src/selectors/index.js
theengineear: ah, right, i was thinking for `currentNotebook` instead of metadata. yes, i agree 👍

packages/core/src/epics/contents.js
```diff
@@ -80,6 +81,12 @@ export function saveContentEpic(
       const state = store.getState();
       const currentNotebook = selectors.currentNotebook(state);
 
+      // TODO: this will likely make more sense when this becomes less
+      // notebook-centric.
+      if (!currentNotebook) {
+        return of(saveFailed());
```
rgbkrk: I do wish we provided more context here, so we know that it's because of a lack of `notebook`.

packages/core/src/selectors/index.js
```diff
@@ -87,62 +98,142 @@ export const currentSavedNotebook = createSelector(
...
+);
+
+const CODE_MIRROR_MODE_DEFAULT = "text";
+export const codeMirrorMode = createSelector(
+  [currentNotebookMetadata],
+  metadata =>
+    metadata.getIn(["language_info", "codemirror_mode"]) ||
+    metadata.getIn(["kernel_info", "language"]) ||
+    metadata.getIn(["kernelspec", "language"]) ||
+    CODE_MIRROR_MODE_DEFAULT
```
rgbkrk, line 129: Fantastic!

rgbkrk (approved):

PR #2508, opened 2018-02-08
Move from Object.entries to lodash.toPairs

### LINKED ISSUE #2498: Switch Object.entries usage to lodash equivalent
As reported by @tlyon3, with his old version of node (v6, though we suggest people use v8+), `Object.entries` is not a function. On all the browsers including desktop safari `Object.entries` is well supported. However, Safari mobile (the only option on iOS, even if using Chrome) does not support it. We should switch to the lodash equivalent.

/cc @theengineear

captainsafia:
Closes #2498.

### DIFF
--- a/packages/styles/buildStyles.js
+++ b/packages/styles/buildStyles.js
@@ -15,10 +15,12 @@ const makeName = name => {
   return `--${config.prefix}-${dashName}`;
 };
 
+const lodash = require("lodash");
+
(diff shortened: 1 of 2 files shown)


rgbkrk:
Thank you!

rgbkrk (approved):
I had no idea `toEntries` was an alias of `toPairs` until now!

PR #2512, opened 2018-02-09
Extract restart and clear output actions

rgbkrk:
This turns what was _way too much_ work from desktop's menu into appropriate epics and reducers for our core state. It does leave some of the old cruft, including use of the notification system when slightly inaccurate. Once the kernel refs are in, we can clean the restarts up even further.

Clear outputs is incredibly sane now that we don't have to do it for each and every cell as a separate action. It's looking more and more like `transient` should be a record, though I'll just wait for the core architecture plan to get implemented before I shuffle that around.

### DIFF
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -66,7 +66,7 @@ export function dispatchRestartKernel(store: *) {
     ? path.dirname(path.resolve(filename))
     : cwdKernelFallback();
 
-  store.dispatch(actions.killKernel);
+  store.dispatch(actions.killKernel());
   // TODO: Use the kernelspec directly, requires us having the kernelspecs available
   //       in the store.
   // TODO: `kernel &&` may be redundant if Record default is `null` for this.
@@ -201,7 +201,7 @@ export function dispatchUnhideAll(store: *) {
 }
 
 export function dispatchKillKernel(store: *) {
-  store.dispatch(actions.killKernel);
+  store.dispatch(actions.killKernel());
 }
 
 export function dispatchInterruptKernel(store: *) {
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -54,45 +54,6 @@ export function triggerWindowRefresh(store: *, filename: string) {
 
 export function dispatchRestartKernel(store: *) {
   store.dispatch(actions.restartKernel());
-
-  // TODO: Make this an action to dispatch that an epic consumes, which will stop the
-  //       current kernel and launch a new kernel of the same type
-  const state = store.getState();
-  const notificationSystem = state.app.notificationSystem;
-  const filename = selectors.currentFilename(state);
-  const kernel = selectors.currentKernel(state);
-
-  const cwd = filename
-    ? path.dirname(path.resolve(filename))
-    : cwdKernelFallback();
-
-  store.dispatch(actions.killKernel());
-  // TODO: Use the kernelspec directly, requires us having the kernelspecs available
-  //       in the store.
-  // TODO: `kernel &&` may be redundant if Record default is `null` for this.
-  const kernelName =
-    kernel && kernel.kernelSpecName ? kernel.kernelSpecName : null;
-
-  if (!kernelName) {
-    notificationSystem.addNotification({
-      title: "Failure to Restart",
-      message: `Unable to restart kernel, please select a new kernel.`,
-      dismissible: true,
-      position: "tr",
-      level: "error"
-    });
-
-    return;
-  }
-  store.dispatch(actions.launchKernelByName(kernelName, cwd));
-
-  notificationSystem.addNotification({
-    title: "Kernel Restarted",
-    message: `Kernel ${kernelName} has been restarted.`,
-    dismissible: true,
-    position: "tr",
-    level: "success"
-  });
 }
 
 export function triggerKernelRefresh(store: *): Promise<*> {
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -56,7 +56,7 @@ export function dispatchRestartKernel(store: *) {
   store.dispatch(actions.restartKernel());
 }
 
-export function triggerKernelRefresh(store: *): Promise<*> {
+export function triggerKernelRefresh(store: *, filename: string): Promise<*> {
   return new Promise(resolve => {
     dialog.showMessageBox(
       {
@@ -72,7 +72,13 @@ export function triggerKernelRefresh(store: *): Promise<*> {
       },
       index => {
         if (index === 0) {
-          dispatchRestartKernel(store);
+          const kernel = selectors.currentKernel();
+          const cwd = filename
+            ? path.dirname(path.resolve(filename))
+            : cwdKernelFallback();
+          store.dispatch(
+            actions.launchKernelByName(kernel.kernelSpecName, cwd)
+          );
         }
         resolve();
       }
@@ -84,7 +90,7 @@ export function triggerSaveAs(store: *) {
   showSaveAsDialog().then(filename => {
     if (filename) {
       triggerWindowRefresh(store, filename);
-      triggerKernelRefresh(store);
+      triggerKernelRefresh(store, filename);
     }
   });
 }
@@ -180,8 +186,7 @@ export function dispatchInterruptKernel(store: *) {
 }
 
 export function dispatchRestartClearAll(store: *) {
-  dispatchRestartKernel(store);
-  dispatchClearAll(store);
+  store.dispatch(actions.restartKernel({ clearOutputs: true }));
 }
 
 export function dispatchZoomIn() {
@@ -305,10 +310,9 @@ export function triggerSaveAsPDF(store: *) {
   showSaveAsDialog()
     .then(filename => {
       if (filename) {
-        return Promise.all([
-          triggerWindowRefresh(store, filename),
-          triggerKernelRefresh(store)
-        ]).then(() => storeToPDF(store));
+        return Promise.all([triggerWindowRefresh(store, filename)]).then(() =>
+          storeToPDF(store)
+        );
       }
     })
     .catch(e =>
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -72,7 +72,7 @@ export function triggerKernelRefresh(store: *, filename: string): Promise<*> {
       },
       index => {
         if (index === 0) {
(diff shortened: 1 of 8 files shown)


rgbkrk:
I'll be knocking each and every one of these anti-patterns in menu out and making them reusable across all apps. 🎉

rgbkrk:
I seem to have found a case for a selector that will have a specific implementation in desktop and the webapp version -- `cwd` (current working directory). I'm going to see how I can factor out the current way it works in a stable way. I'm starting with making the kernel keep track of its cwd. The reason I need this is to make a everything available that is needed for a kernel restart without having to write two separate epics.

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -112,13 +106,28 @@ export const launchKernelWhenNotebookSetEpic = (
...
+  action$.pipe(
+    ofType(actionTypes.RESTART_KERNEL),
+    mergeMap(x => {
+      const state = store.getState();
+      const kernel = selectors.currentKernel(state);
+      // TODO: Incorporate notification system bits
+
+      return merge(
+        actions.killKernel,
+        actions.launchKernelByName(kernel.kernelSpecName, kernel.cwd)
```
rgbkrk: This is much more cleaned up relative to what we've done in the menu in the past!

applications/desktop/src/notebook/menu.js
```diff
@@ -8,34 +8,13 @@ import * as fs from "fs";
...
-  loadConfig,
-  newNotebook,
-  pasteCell,
-  save,
-  saveAs,
-  setCursorBlink,
-  setGithubToken,
-  setTheme,
-  toggleOutputExpansion
-} from "@nteract/core/actions";
```
theengineear, line 32: 💯

packages/core/src/actions.js
```diff
@@ -723,3 +726,23 @@ export function shutdownReplyTimedOut(error: Error): ShutdownReplyTimedOut {
...
+
+export function restartKernelFailed(error: Error): RestartKernel {
+  return {
+    type: actionTypes.RESTART_KERNEL_FAILED,
+    payload: error,
+    error: true
+  };
+}
+
+export function restartKernelSuccessful(): RestartKernelSuccessful {
```
theengineear, line 759: Just to make sure I'm on the same page, when we have refs for kernels, these will all take a `KernelRef` as the payload, right?

applications/desktop/src/notebook/menu.js
```diff
@@ -70,51 +49,14 @@ export function triggerWindowRefresh(store: *, filename: string) {
...
-  }
-  store.dispatch(launchKernelByName(kernelName, cwd));
-
-  notificationSystem.addNotification({
-    title: "Kernel Restarted",
-    message: `Kernel ${kernelName} has been restarted.`,
-    dismissible: true,
-    position: "tr",
-    level: "success"
-  });
```
theengineear, line 114: Amazin'!

applications/desktop/src/notebook/menu.js
```diff
@@ -130,7 +72,13 @@ export function triggerKernelRefresh(store: *): Promise<*> {
       },
       index => {
         if (index === 0) {
-          dispatchRestartKernel(store);
+          const kernel = selectors.currentKernel(store.getState());
+          const cwd = filename
+            ? path.dirname(path.resolve(filename))
+            : cwdKernelFallback();
+          store.dispatch(
+            actions.launchKernelByName(kernel.kernelSpecName, cwd)
```
rgbkrk, line 80: What I wish we currently had in the `kernel` is either the `kernelSpec` itself or a `kernelSpecRef` so that we could just launch with a known kernelspec (at least for desktop, where we have to do a lookup).

applications/desktop/src/notebook/menu.js
```diff
@@ -8,34 +8,13 @@ import * as fs from "fs";
...
-  newNotebook,
-  pasteCell,
-  save,
-  saveAs,
-  setCursorBlink,
-  setGithubToken,
-  setTheme,
-  toggleOutputExpansion
-} from "@nteract/core/actions";
+import * as actions from "@nteract/core/actions";
```
rgbkrk, line 11: I became just a little bit more sane after switching to a `actions.whatever()` all over here. Much easier for me to scan through and refactor (though the types aren't really loaded because of that weird pseudo namespace).

applications/desktop/src/notebook/menu.js
```diff
@@ -359,10 +309,9 @@ export function triggerSaveAsPDF(store: *) {
   showSaveAsDialog()
     .then(filename => {
       if (filename) {
-        return Promise.all([
-          triggerWindowRefresh(store, filename),
-          triggerKernelRefresh(store)
```
theengineear, line 364: Refresh (😉) my understanding of why `triggerKernelRefresh` is going away here?

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -112,13 +108,53 @@ export const launchKernelWhenNotebookSetEpic = (
...
+        notificationSystem.addNotification({
+          title: "Failure to Restart",
+          message: `Unable to restart kernel, please select a new kernel.`,
+          dismissible: true,
+          position: "tr",
+          level: "error"
+        });
+
+        // TODO: Wow do we need to send notifcations through our store for
+        // consistency
```
theengineear, line 141: 😆 ya.

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -112,13 +108,53 @@ export const launchKernelWhenNotebookSetEpic = (
...
+      //       This only mirrors the old behavior of restart kernel (for now)
+      notificationSystem.addNotification({
+        title: "Kernel Restarted",
+        message: `Kernel ${kernel.kernelSpecName} has been restarted.`,
+        dismissible: true,
+        position: "tr",
+        level: "success"
+      });
+
+      return from([
```
theengineear: Ha, so _this_ is what I was saying about `of`. If you just have values (which actions are) you can `of(action1, action2)`. Just to be clear, I think `from` can be misleading in that it handles a larger set of input arguments. I'm not really that opinionated on it, but just wanted to point it out.

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -112,13 +108,53 @@ export const launchKernelWhenNotebookSetEpic = (
...
+        actions.killKernel({ restarting: true }),
+        actions.launchKernelByName(kernel.kernelSpecName, kernel.cwd)
+      ]);
```
theengineear: Nice and clean 😎

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -112,13 +108,53 @@ export const launchKernelWhenNotebookSetEpic = (
...
 
-      return launchKernelByName(kernelSpecName, cwd);
+      return actions.launchKernelByName(kernelSpecName, cwd);
+    })
+  );
+
+export const restartKernelEpic = (action$: ActionsObservable<*>, store: *) =>
+  action$.pipe(
+    ofType(actionTypes.RESTART_KERNEL),
+    mergeMap(action => {
```
theengineear: I feel like this keeps coming up, but, `concatMap`? Or, `switchMap`? I'm trying to think through what should happen for multiple kernel restarts.

theengineear: I.e., what's your rationale for `mergeMap` here?

packages/core/src/reducers/document.js
```diff
@@ -195,6 +197,44 @@ function clearOutputs(state: DocumentRecord, action: ClearOutputsAction) {
...
+    action.type === actionTypes.RESTART_KERNEL &&
+    !action.payload.clearOutputs
+  ) {
+    return state;
+  }
+
+  // For every cell, clear the outputs and execution counts
+  const cellMap = state
+    .getIn(["notebook", "cellMap"], new Immutable.Map())
+    // NOTE: My kingdom for a mergeMap
```
theengineear, line 216: 🎭 ! Dramatic comment award ⭐️

packages/core/src/reducers/document.js
```diff
@@ -723,6 +765,9 @@ function handleDocument(
       return focusCell(state, action);
     case actionTypes.CLEAR_OUTPUTS:
       return clearOutputs(state, action);
+    case actionTypes.CLEAR_ALL_OUTPUTS:
+    case actionTypes.RESTART_KERNEL:
+      return clearAllOutputs(state, action);
```
theengineear, line 770: Am I misunderstanding, or, are we just clearing all outputs on kernel restart either way here?

theengineear:
^^ some comments for you! This all looks great!

packages/core/src/actions.js
rgbkrk: Yes!

applications/desktop/src/notebook/menu.js
rgbkrk: 😁

applications/desktop/src/notebook/menu.js
rgbkrk: Sorry, I should have left a comment. I didn't understand why saving a PDF would force the kernel to reload here.

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: Gosh golly, lemme switch it.

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: It'll be even better with the refs here, since we'll be able to say "kill this one, start this one, and make this one the current one".

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: `concatMap` is probably best, though it may be a bit silly since the observable returned in both cases are completely synchronous. I'll switch to `concatMap` for consistency.

packages/core/src/reducers/document.js
rgbkrk: 🤣

packages/core/src/reducers/document.js
rgbkrk: In `clearAllOutputs` there's a check to see if the restart kernel action has `payload.clearOutputs` set to `true`.

applications/desktop/src/notebook/menu.js
theengineear: np, just wanted to make sure that was intentional

theengineear (approved):

PR #2516, opened 2018-02-09
[WIP] Add core state types

theengineear:
[WIP, DON'T MERGE]

Fleshing out our state types in `/core`.

- [x] add types for our desired core state
- [ ] setup reducers to track current actions in new state.


packages/core/src/reducers/root/communication/contents.js
```diff
@@ -0,0 +1 @@
+// @flow
```
theengineear, line 1: I auto-add these when I create files now :)

packages/core/src/reducers/root/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+export const root = combineReducers(
+  {
+    // TODO: selectedHostRef
+    // TODO: selectedContentRef
+    communication,
+    entities
+  },
+  makeStateRecord
+);
```
theengineear, line 15: OK, totally boiler-plate-y, BUT, the idea is that you just set it and forget it when you configure your store. The big win here is that `selectors` become the Official™️ way to look into the state tree. That means that we ca *finally* say the `core` state acts like a black box to all the other packages/applications 🎉.

I want to keep pausing at nice stopping points to make sure we're still on the same page. This still seem reasonable?

packages/core/src/reducers/root/communication/kernelspecs.js
```diff
@@ -0,0 +1,42 @@
...
+
+const byRef = createImmutableMapReducer(
+  action => action.kernelspecsRef,
+  combineReducers({ loading, error }, makeKernelspecsCommunicationRecord)
+);
+
+export const kernelspecs = combineReducers(
+  { byRef },
+  makeAllKernelspecsCommunicationRecord
+);
```
rgbkrk, line 42: This is all looking super good.

packages/core/src/reducers/root/communication/kernelspecs.js
```diff
@@ -0,0 +1,42 @@
...
+    case actionTypes.FETCH_KERNELSPECS_FAILED:
+      return action.payload.error;
+    default:
+      return state;
+  }
+};
+
+const byRef = createImmutableMapReducer(
+  action => action.kernelspecsRef,
+  combineReducers({ loading, error }, makeKernelspecsCommunicationRecord)
```
rgbkrk, line 36: What's the `byRef` doing with the `loading` states?

packages/core/src/reducers/root/index.js
rgbkrk: Woo, yeah!

packages/core/src/types/state/communication/contents.js
```diff
@@ -0,0 +1,26 @@
...
+
+export type ContentsCommunicationRecordProps = {
+  byRef: Map<ContentRef, RecordOf<ContentCommunicationRecordProps>>
+};
+
+export const makeContentsCommunicationRecord: RecordFactory<
+  ContentsCommunicationRecordProps
+> = Record({
+  byRef: Map()
+});
```
rgbkrk, line 26: Looking solid, glad you `type`d the `Map` above.

packages/core/src/types/state/communication/hosts.js
```diff
@@ -0,0 +1,26 @@
...
+
+export type HostCommunicationRecordProps = {
+  isSaving: boolean,
+  error: ?Object
+};
+
+export const makeHostCommunicationRecord: RecordFactory<
+  HostCommunicationRecordProps
+> = Record({
+  isSaving: false,
```
rgbkrk: Wait, why does `host` have an `isSaving`?

packages/core/src/types/state/communication/hosts.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import type { HostRef } from "../refs";
+import type { RecordFactory, RecordOf } from "immutable";
+import { Map, Record } from "immutable";
+
+export type HostCommunicationRecordProps = {
+  isSaving: boolean,
```
rgbkrk: I feel like the core plan has some bits about what we wanted to capture for communication about hosts.

packages/core/src/types/state/communication/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const makeCommunicationRecord: RecordFactory<
+  CommunicationRecordProps
+> = Record({
+  contents: makeContentsCommunicationRecord(),
+  hosts: makeHostsCommunicationRecord(),
+  kernels: makeKernelsCommunicationRecord(),
+  kernelspecs: makeAllKernelspecsCommunicationRecord(),
+  preferences: makePreferencesCommunicationRecord()
+});
```
rgbkrk, line 31: Oh myyyyy 😄

packages/core/src/types/state/communication/kernels.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import type { KernelRef } from "../refs";
+import type { RecordFactory, RecordOf } from "immutable";
+import { Map, Record } from "immutable";
+
+export type KernelCommunicationRecordProps = {
+  isSaving: boolean,
```
rgbkrk: I'm guessing this was all just copy-pasta'ed from how contents was working -- what's `isSaving` here for? Are there other properties we care to communicate from here?

packages/core/src/types/state/communication/kernelspecs.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import type { KernelspecsRef } from "../refs";
+import type { RecordFactory, RecordOf } from "immutable";
+import { Map, Record } from "immutable";
+
+export type KernelspecsCommunicationRecordProps = {
+  isSaving: boolean,
```
rgbkrk: ohhhh, now I'm understanding -- perhaps this is for `isLoading` -- that's probably what we want here

packages/core/src/types/state/entities/contents.js
```diff
@@ -0,0 +1,90 @@
+// @flow
+import type { ContentRef } from "../refs";
+import type { ImmutableNotebook } from "@nteract/commutable";
+import type { RecordFactory, RecordOf } from "immutable";
+import { List, Map, Record } from "immutable";
+
+// TODO: NotebookModel will likely diverge from this when we flatten
+// inputs/outputs. For now, this is nicely defined and is workable.
+type NotebookModel = ImmutableNotebook;
```
rgbkrk, line 9: 👍

packages/core/src/types/state/entities/contents.js
```diff
@@ -0,0 +1,90 @@
...
+  latSaved: ?Date,
+  writable: boolean
+};
+
+export type FileContentRecordProps = BaseContentRecordProps & {
+  type: "file",
+  mimetype: string,
+  format: "text" | "base64",
+  model: string
+};
```
rgbkrk, line 26: This is intended to be our in-memory version and likely different from the Jupyter API response, is that why that isn't used here?

packages/core/src/types/state/entities/contents.js
```diff
@@ -0,0 +1,90 @@
...
+export type DirectoryContentRecordProps = BaseContentRecordProps & {
+  type: "directory",
+  format: null | "json",
+  model: List<ContentRef>
+};
+
+export type NotebookContentRecordProps = BaseContentRecordProps & {
+  type: "notebook",
+  format: null | "json",
+  model: *
```
rgbkrk, line 54: 👍

packages/core/src/types/state/entities/contents.js
```diff
@@ -0,0 +1,90 @@
...
+// TODO: not sure about path/name... should these be nullable?
+// The default record is meant to basically be ridiculous... i.e., it'd never
+// get used... This type setup could probably use some work.
+export type ContentRecordProps = {
+  type: "directory" | "notebook" | "file",
+  mimetype: ?string,
+  path: ?string,
+  name: ?string,
+  created: ?Date,
+  latSaved: ?Date,
```
rgbkrk: Earlier today I saved my lats by stretching.

packages/core/src/types/state/entities/contents.js
```diff
@@ -0,0 +1,90 @@
...
+
+export type NotebookContentRecordProps = BaseContentRecordProps & {
+  type: "notebook",
+  format: null | "json",
+  model: *
+};
+
+// TODO: not sure about path/name... should these be nullable?
+// The default record is meant to basically be ridiculous... i.e., it'd never
+// get used... This type setup could probably use some work.
```
rgbkrk: In the lifecycle of the application we never want the path to be null. Probably inevitable for startup though that we don't have this set so we could default to `"."` and default the type to "directory" I suppose.

rgbkrk: We _do_ have an option for enforcing they get set on creation -- the publicly exported `makeContentRecord` can be a function of our _own_ design that uses the real RecordFactory underneath. We can enforce that these properties are set when that function is called.

packages/core/src/types/state/entities/hosts.js
```diff
@@ -0,0 +1,68 @@
...
+import type { HostRef, KernelspecsRef } from "../refs";
+import type { RecordFactory, RecordOf } from "immutable";
+import { List, Map, Record } from "immutable";
+
+export type BaseHostProps = {
+  id: ?HostId,
+  ref: ?HostRef,
+  kernelspecsRef: ?KernelspecsRef, // reference to a collection of kernelspecs
+  defaultKernelName: ?string,
+  kernelIds: List<KernelId>
```
rgbkrk: The `BaseHost` shouldn't have `kernelIds`, those only apply to the Jupyter server. Then again, I think maybe the uuid on the server neatly matches the connection file name written to disk.

At any rate I think `kernelIds` only belongs in the `JupyterHost`

packages/core/src/types/state/entities/hosts.js
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const makeJupyterHostRecord: RecordFactory<
+  JupyterHostRecordProps
+> = Record({
+  type: "jupyter",
+  ref: null,
+  id: null,
+  kernelspecsRef: null,
+  defaultKernelName: null,
```
rgbkrk: It's ok to default this to `python`

packages/core/src/types/state/entities/kernels.js
```diff
@@ -0,0 +1,74 @@
...
+import type { HostRef, KernelRef } from "../refs";
+import type { KernelId } from "../ids";
+import type { RecordFactory, RecordOf } from "immutable";
+import { ChildProcess } from "child_process";
+import { Map, Record } from "immutable";
+
+export type BaseKernelProps = {
+  ref: ?KernelRef,
+  name: ?string,
+  kernelSpecName: ?string,
```
rgbkrk, line 11: Note: we'll want to make sure we can look up a kernelspec by name in the kernelspecs structure. Names are unique per set of _kernelspecs_. We'll be able to get the collection of kernelspecs using the `hostRef` since that will give us the `kernelSpecsRef` (sorry about all the crazy plurality naming and potential confusion).

packages/core/src/types/state/entities/kernels.js
```diff
@@ -0,0 +1,74 @@
...
+  status: null
+});
+
+export type RemoteKernelRecord = RecordOf<RemoteKernelProps>;
+
+export type KernelRecord = LocalKernelRecord | RemoteKernelRecord;
+
+export type KernelsRecordProps = {
+  byRef: Map<KernelRef, KernelRecord>
+};
```
rgbkrk, line 70: 💯

packages/core/src/types/state/entities/kernelspecs.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const makeKernelspec: RecordFactory<KernelspecProps> = Record({
+  name: null,
+  resources: Map(),
+  language: null,
+  argv: List(),
+  env: Map(),
+  interrupt_mode: null
+});
```
rgbkrk: Most excellent.

packages/core/src/types/state/entities/kernelspecs.js
```diff
@@ -0,0 +1,32 @@
...
+
+export type KernelspecsRecordProps = {
+  byRef: Map<KernelspecsRef, RecordOf<KernelspecProps>>
```
rgbkrk: Ok, this is where things will probably start to get a bit confusing.

`KernelSpecsRef` gives you access to a collection of `KernelSpecs` each of which is identified by name.

Example lookup:

```js
function kernelSpec(kernelSpecsRef, kernelSpecName) {
  const kernelSpecs = kernelspecs.byRef[kernelSpecsRef];
  return kernelSpecs.byName[kernelSpecName];
}
```

Does that clarify it?

packages/core/src/types/state/ids.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+// The *Id type is for external resources like the trailing id in:
+//   /api/kernels/9092-7679-9978-8822
+// ... where 9092-7679-9978-8822 is the KernelId.
+//
+// We call it "castTo*" to make it clear that you're not getting anything unique
+// out of the function.
```
rgbkrk, line 8: Interesting.

packages/core/src/types/state/index.js
```diff
@@ -0,0 +1,22 @@
...
+import type { ContentRef, HostRef } from "./refs";
+import type { EntitiesRecordProps } from "./entities";
+import type { RecordFactory, RecordOf } from "immutable";
+import { Record } from "immutable";
+import { makeCommunicationRecord } from "./communication";
+import { makeEntitiesRecord } from "./entities";
+
+export type StateRecordProps = {
+  selectedHostRef: ?HostRef,
+  selectedContentRef: ?ContentRef,
```
rgbkrk, line 12: 👍

packages/core/src/types/state/ids.js
rgbkrk: Ok, yeah this makes sense since we have to take a string in from externally and declare them as "IDs".

packages/core/src/reducers/root/communication/kernelspecs.js
theengineear: The state should look something like:

```
{
  ...
  communication: {
    kernelspecs: {
      byRef: {
        [REF]: {
          loading: true,
          error: null
        }
      }
    }
  }
}
```

Where this assumes that you might have multiple `kernelspecs`. If we _know_ there is only one possible kernelspec, then this isn't necessary. I thought that was the case though right?

packages/core/src/types/state/communication/hosts.js
theengineear: 💀 oh man, totally bad copy pasta. Thanks for catching this!

packages/core/src/types/state/entities/contents.js
theengineear: 😆

packages/core/src/reducers/root/communication/kernelspecs.js
rgbkrk: Yeah we can have multiple kernelspecs, my brain was stuck in jupyter extension / one kernelspecs mode when in fact we'd want to support multiple since we support multiple hosts.

packages/core/src/types/state/entities/kernelspecs.js
theengineear: ah yah, that it confusing... OK, that makes sense though!

packages/core/src/types/state/entities/hosts.js
rgbkrk: I suppose, looking over this again, that we could always have `kernelRefs` for both. Maybe that's duplicative though. I guess what I'm saying is that we'll want to be able to list all the running kernels on a jupyter host and then all the ones active for the document are in `kernelRefs`?

packages/core/src/types/state/communication/hosts.js
theengineear: I don't think there's additional information for the `communication` plan (this should be `loading` not `isSaving`, btw)

packages/core/src/types/state/communication/kernels.js
theengineear: For something like `contents`, it might actually be nice to have `isSaving` and `isLoading`, which could of course be combined into `status: "saving" | "loading"` or something.

packages/core/src/types/state/entities/contents.js
theengineear: Yup, exactly.

packages/core/src/types/state/entities/contents.js
theengineear: 🤔 not a bad idea!

theengineear: I'm going to pass on that for now and just use the defaults you suggested though. I'll leave a comment about it.

packages/core/src/types/state/entities/hosts.js
theengineear: I think we decided that kernels know about their hosts, but hosts only have the `kernelIds` (so we're not duplicating state). So we'd have this list of running kernels via `kernelIds`.

packages/types/core/plan.md
```diff
@@ -164,15 +164,16 @@ type core = {
...
+            resources: Object,
+            hostRef: Ref,
+          byName: {
+            [name: string]: {
+              argv: Array<string>,
+              displayName: string,
+              env: Object,
+              language: string
+            }
+          },
```
theengineear: @rgbkrk am I getting warmer? 🔥?

packages/core/src/types/state/communication/kernels.js
rgbkrk: I think maybe both `isSaving` and `isLoading` is fine, since we could have both in process technically (even though we shouldn't).

packages/core/src/types/state/entities/contents.js
rgbkrk: That'll work.

packages/core/src/types/state/entities/hosts.js
rgbkrk: Yeah, the main reason we'd want to know all the kernel IDs is so that on the dashboard we are able to list which kernels and notebooks are currently running.

packages/core/src/reducers/root/index.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import { combineReducers } from "redux-immutable";
+import { makeStateRecord } from "../../types/state";
+import { communication } from "./communication";
+import { entities } from "./entities";
+
+export const root = combineReducers(
+  {
+    // TODO: selectedHostRef
+    // TODO: selectedContentRef
```
rgbkrk, line 10: ~~Are we just going to have `selectedHostRef` etc. be defaults for now in the state? It's probably ok to have a simple reducer action combo for `SELECT_HOST` where we'd pass a `ref`.~~ Already answered below with it being part of the state, though we could still do a reducer.

packages/core/src/types/state/entities/index.js
```diff
@@ -0,0 +1,29 @@
...
+import { makeKernelspecsRecord } from "./kernelspecs";
+import { makePreferencesRecord } from "./preferences";
+
+export type EntitiesRecordProps = {
+  contents: RecordOf<ContentsRecordProps>,
+  hosts: RecordOf<HostsRecordProps>,
+  kernels: RecordOf<KernelsRecordProps>,
+  kernelspecs: RecordOf<KernelspecsRecordProps>,
+  preferences: RecordOf<PreferencesRecordProps>
+};
```
rgbkrk, line 21: A+ to see all this.

packages/core/src/types/state/entities/kernelspecs.js
```diff
@@ -0,0 +1,42 @@
...
+  language: null
+});
+
+export type KernelspecsByNameRecordProps = {
+  hostRef: ?HostRef,
+  resources: Map<string, *>,
+  byName: Map<string, RecordOf<KernelspecProps>>
+};
+
+export const makeKernelspecsByNameRecord: RecordFactory<
```
rgbkrk: Total nit, but I'd probably call this `makeKernelspecByNameRecord`, noting lack of plurality, just so we don't confuse it with the `kernelspecs` related ones.

packages/core/src/types/state/index.js
rgbkrk: Ah ok, I see this answers what I asked above.

packages/types/core/plan.md
rgbkrk: Yeah, that makes sense. It's nested under the `byRef` version since you have to have the ref and the name.

`resources` belongs inside an individual kernelspec though.

`GET` on `/api/kernelspecs` returns

```js
{
  default: "python3",
  kernelspecs: {
    python3: {
      name: "python3",
      spec: {
        argv: ["python", "-m", "ipykernel_launcher", "-f", "{connection_file}"],
        env: {},
        display_name: "Python 3",
        language: "python",
        interrupt_mode: "signal",
        metadata: {}
      },
      resources: {
        "logo-64x64":
          "/user/nteract-vdom-q5kh4u9v/kernelspecs/python3/logo-64x64.png",
        "logo-32x32":
          "/user/nteract-vdom-q5kh4u9v/kernelspecs/python3/logo-32x32.png"
      }
    }
  }
}
```

theengineear: ah, excellent, that's super helpful!

theengineear:
Still plugging away at this. It sure does touch a lot of code 😬.

One thing that is sorta odd for us are failure actions... Take `LAUNCH_KERNEL_FAILED`. We want to type this as an `ErrorAction` and use `payload: Error, error: true`, yada-yada. HOWEVER, now that we have _many_ kernels that can fail, we actually want the payload to be something like `payload: { kernelRef: string, error: Error }`, hrm...

Anyways, I'd like to standardize on always providing `payload` as an envelope. I don't think I'll get there in this PR, but I wanted to throw it out there for thoughts.

Slowly, but surely...

rgbkrk:
There's too ways about it -- either the `payload` is an `Error` with extra properties on it or the `payload` has both the error and all other data we want to keep. Doesn't matter to me, just as long as if it's the `payload` option that we go ahead and standardize it and update the error middleware (which I think is only on desktop, though it should be in core).

rgbkrk:
> Still plugging away at this. It sure does touch a lot of code 

If you find smaller PRs or things that could be done separately, let me know. I'm happy to help. 😄

rgbkrk:
This became smaller PRs so I'm going to close it.

PR #2517, opened 2018-02-10
create unhide all redux setup

rgbkrk:
This sets up a menu item that unhides both input and outputs for all cells in the document.

### DIFF
--- a/packages/core/src/reducers/document.js
+++ b/packages/core/src/reducers/document.js
@@ -10,6 +10,7 @@ import * as actionTypes from "../actionTypes";
 // TODO: With the new document plan, I think it starts to make sense to decouple
 //       the document view actions and the underlying document format
 import type {
+  UnhideAll,
   RestartKernel,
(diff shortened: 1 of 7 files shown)


packages/core/src/reducers/document.js
```diff
@@ -576,6 +577,21 @@ function toggleCellOutputVisibility(
...
 }
 
+function unhideAll(state: DocumentRecord, action: UnhideAll) {
+  return state.updateIn(["notebook", "cellMap"], cellMap =>
+    cellMap.map(cell => {
+      if (cell.get("cell_type") === "code") {
+        return cell.mergeIn(["metadata"], {
+          // TODO: Verify that we convert to one namespace for hidden input/output
+          outputHidden: !action.payload.outputs,
+          inputHidden: !action.payload.input
```
rgbkrk: This makes me wonder if I should change the nomenclature and pattern to `action.payload.outputHidden`.

PR #2522, opened 2018-02-12
Migrate the rest of Desktop's app reducer into core

rgbkrk:
This finishes off the cleanup of the reducers that were in Desktop, everything is in core now. 🎉


theengineear:
Great!

Issue #2458, opened 2018-01-31
Cannot clear cell output of R documentation


jkornblum:
I'm running nteract 0.6.2 on Ubuntu 17.10
I cannot clear the output, toggle output visibility, or toggle output expansion when I view R documentation.
The output is also expanded automatically

This is my `R.version` output
```R
platform       x86_64-conda_cos6-linux-gnu 
arch           x86_64                      
os             linux-gnu                   
system         x86_64, linux-gnu           
status                                     
major          3                           
minor          4.2                         
year           2017                        
month          09                          
day            28                          
svn rev        73368                       
language       R                           
version.string R version 3.4.2 (2017-09-28)
nickname       Short Summer     
```

I can clear the first cell output but not perform any actions on the second.

![nteract-r-error](https://user-images.githubusercontent.com/5368903/35652982-a2a1cdc0-06ab-11e8-93ae-2682af6ecbd7.png)

rgbkrk:
OMG the Pager! I personally have just been running the cell again to clear that out.

Luckily that's not stored in the notebook. Would you want to just collapse it or full on delete it?

jkornblum:
I'm confused by what you're asking.
I'm able to delete it but I was just trying to use the "clear cell output", I suppose I can just rerun the cell 😄

rgbkrk:
I edited my comment so hopefully that shows up now, I delayed a bit in fixing what I wrote.

Basically, the R docs are coming through something separate from outputs even though we put them in the output area. We could actually convert them to an output, store them in the notebook, and if you clear or collapse them they'd be handled like normal outputs.

rgbkrk:
This will end up in the next desktop release (likely this weekend).

PR #2524, opened 2018-02-13
[NOMERGE] Pager as output

### LINKED ISSUE #2521: Docstring doesn't clear on Clear Cell Output
Using v.0.7.1 on Windows 10.  

Attempted to clear a docstring output after doing something like within a cell

```
pandas.DataFrame?
```
Tried clearing the output from the tools in the cell and the toolbar with no luck.

rgbkrk:
Closes #2521, #2458

This puts the pager data in the output directly. Note the style change, mostly because of constraints of Outputs.

screen shot 2018-02-13 at 10 26 45 am

There's an undesirable affect of this though, based on how IPython works -- the pager data almost always appears at the bottom.

screen shot 2018-02-13 at 10 28 14 am

I'm actually thinking I should just make sure that pager data is cleared when outputs are cleared instead of this PR. I want the help doc to be right at the top for users to see right away (since they asked for docs).


rgbkrk:
Closing in favor of #2525

PR #2525, opened 2018-02-13
Clear pager on output clearing

### LINKED ISSUE #2458: Cannot clear cell output of R documentation

### LINKED ISSUE #2521: Docstring doesn't clear on Clear Cell Output

rgbkrk:
Closes #2521 
Closes #2458

### DIFF
--- a/packages/core/__tests__/document-spec.js
+++ b/packages/core/__tests__/document-spec.js
@@ -9,7 +9,8 @@ import {
   emptyCodeCell,
   emptyMarkdownCell,
   fromJS,
-  appendCellToNotebook
+  appendCellToNotebook,
+  emptyNotebook
 } from "@nteract/commutable";
 
 import { makeDocumentRecord } from "@nteract/types/core/records";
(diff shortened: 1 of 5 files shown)


theengineear:
Just for context, because it wasn't clear to me when I first looked this over. The reason we want to do this is that the `cell_pager` which is the help-text that appears in the output area when you do something like `list?`. However, it's not output, so it previously didn't clear when you cleared all outputs. This is pretty confusing for users who shouldn't need to understand the difference between transient pager information and real cell output.

theengineear:
Looks great @rgbkrk !

rgbkrk:
Yeah it's pretty confusing. Users shouldn't have to understand and this data has never been stored in the document previously. We (nteract) interwove the help doc and the output to make them colocated with the cell they were requested from. Jupyter made it a little popup drawer that came up from the bottom (and there could only be one).

Issue #2483, opened 2018-02-06
[jext] Kernelspecs should generate Menu


rgbkrk:
We need generated menu items for:

* Switching kernels to other languages
* Making new notebooks with new languages

This can all rely on the kernelspecs call

rgbkrk:
Done in #2715

PR #2528, opened 2018-02-14
set up restart and restart+clear menu items

rgbkrk:
Sets up restart and the restart+clear menu items for the web app.

### DIFF
--- a/packages/core/src/components/notebook-menu/index.js
+++ b/packages/core/src/components/notebook-menu/index.js
@@ -38,6 +38,8 @@ type Props = {
   setCellTypeMarkdown: ?(cellId: ?string) => void,
   setTheme: ?(theme: ?string) => void,
   openAboutModal: ?() => void,
+  restartKernel: ?() => void,
+  restartKernelAndClearOutputs: ?() => void,
   interruptKernel: ?() => void
 };
 
@@ -61,6 +63,7 @@ class PureNotebookMenu extends React.Component<Props> {
     setCellTypeMarkdown: null,
     setTheme: null,
     openAboutModal: null,
+    restartKernel: null,
     interruptKernel: null
   };
   handleClick = ({ key }: { key: string }) => {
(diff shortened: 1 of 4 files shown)


packages/core/src/components/notebook-menu/index.js
```diff
@@ -285,6 +300,18 @@ class PureNotebookMenu extends React.Component<Props> {
...
+              Restart
+            </MenuItem>
+            <MenuItem
+              key={createActionKey(MENU_ITEM_ACTIONS.RESTART_AND_CLEAR_OUTPUTS)}
+            >
+              Restart and Clear All Cells
+            </MenuItem>
+            {/*
+              <Divider />
+              <SwitchToKernelSpecsMenu />
```
rgbkrk, line 313: Placeholder for #2483

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -126,7 +126,22 @@ export const restartKernelEpic = (action$: ActionsObservable<*>, store: *) =>
     concatMap(action => {
       const state = store.getState();
       const kernel = selectors.currentKernel(state);
-      const notificationSystem = state.app.notificationSystem;
+      const notificationSystem = state.app.notificationSystem
+        ? state.app.notificationSystem
+        : {
```
rgbkrk: The web app isn't invoking / creating the notification system, so I set up a hokey one here. I could make this part of the default state tree so we at least have it defined (though we still need to move to our own managed notification system).

rgbkrk: Ok, I've done that just now. Just a heads up here @theengineear.

theengineear: heh, nice. good enough for now!

packages/types/src/core/records.js
```diff
@@ -167,7 +167,20 @@ export const makeAppRecord: RecordFactory<AppRecordProps> = Record({
   kernel: null,
   host: null,
   githubToken: null,
-  notificationSystem: null,
+  notificationSystem: {
+    addNotification: msg => {
```
theengineear, line 171: heh, nice. good enough for now!

PR #2529, opened 2018-02-14
Halt Kernel in Menu

rgbkrk:
As part of this, I created parametrized error actions for the boilerplatey `*failed` actions. It works well with flow too:

```js
let x: ?KillKernelFailed;

// Flow will bomb out here, as it expects the literal "KILL_KERNEL_FAILED"
x = {
  type: KILL_KERNEL,
  payload: new Error("test"),
  error: true
};

// ok with flow, like we want 👌🏻
x = {
  type: KILL_KERNEL_FAILED,
  payload: new Error("test"),
  error: true
};
```

* [x] create killKernel(success,fail)
* [x] create web based epic for killing kernel
* [x] put halt in the menu

### DIFF
--- a/packages/core/src/actionTypes.js
+++ b/packages/core/src/actionTypes.js
@@ -27,6 +27,12 @@ import type { ExecuteRequest } from "@nteract/messaging";
 
 import type { Output, StreamOutput } from "@nteract/commutable/src/v4";
 
+export type ErrorAction<T: string> = {
+  type: T,
+  payload: Error,
+  error: true
+};
+
 export const OPEN_MODAL = "CORE/OPEN_MPODAL";
 export type OpenModal = {
   type: "CORE/OPEN_MPODAL",
@@ -394,11 +400,7 @@ export const INTERRUPT_KERNEL_SUCCESSFUL = "INTERRUPT_KERNEL_SUCCESSFUL";
 export type InterruptKernelSuccessful = { type: "INTERRUPT_KERNEL_SUCCESSFUL" };
 
 export const INTERRUPT_KERNEL_FAILED = "INTERRUPT_KERNEL_FAILED";
-export type InterruptKernelFailed = {
-  type: "INTERRUPT_KERNEL_FAILED",
-  payload: Error,
-  error: true
-};
+export type InterruptKernelFailed = ErrorAction<"INTERRUPT_KERNEL_FAILED">;
 
 export const KILL_KERNEL = "KILL_KERNEL";
 export type KillKernelAction = {
@@ -406,6 +408,14 @@ export type KillKernelAction = {
   payload: {
     restarting: boolean
   }
+};
+
+export const KILL_KERNEL_FAILED = "KILL_KERNEL_FAILED";
+export type KillKernelFailed = ErrorAction<"KILL_KERNEL_FAILED">;
+
+export const KILL_KERNEL_SUCCESSFUL = "KILL_KERNEL_SUCCESSFUL";
+export type KillKernelSuccessful = {
+  type: "KILL_KERNEL_SUCCESSFUL"
 };
 
 export const SET_GITHUB_TOKEN = "SET_GITHUB_TOKEN";
@@ -423,11 +433,7 @@ export type RestartKernel = {
 };
 
 export const RESTART_KERNEL_FAILED = "RESTART_KERNEL_FAILED";
-export type RestartKernelFailed = {
-  type: "RESTART_KERNEL_FAILED",
-  payload: Error,
-  error: true
-};
+export type RestartKernelFailed = ErrorAction<"RESTART_KERNEL_FAILED">;
 
 export const RESTART_KERNEL_SUCCESSFUL = "RESTART_KERNEL_SUCCESSFUL";
 export type RestartKernelSuccessful = {
@@ -442,11 +448,7 @@ export type LaunchKernelAction = {
 };
 
 export const LAUNCH_KERNEL_FAILED = "LAUNCH_KERNEL_FAILED";
-export type LaunchKernelFailed = {
-  type: "LAUNCH_KERNEL_FAILED",
-  payload: Error,
-  error: true
-};
+export type LaunchKernelFailed = ErrorAction<"LAUNCH_KERNEL_FAILED">;
 
 export const LAUNCH_KERNEL_SUCCESSFUL = "LAUNCH_KERNEL_SUCCESSFUL";
 export type NewKernelAction = {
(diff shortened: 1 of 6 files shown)


packages/core/src/actionTypes.js
```diff
@@ -27,6 +27,12 @@ import type { ExecuteRequest } from "@nteract/messaging";
 
 import type { Output, StreamOutput } from "@nteract/commutable/src/v4";
 
+export type ErrorAction<T: string> = {
+  type: T,
+  payload: Error,
+  error: true
+};
+
```
theengineear, line 35: ✨

packages/core/src/actionTypes.js
```diff
@@ -394,11 +400,7 @@ export const INTERRUPT_KERNEL_SUCCESSFUL = "INTERRUPT_KERNEL_SUCCESSFUL";
 export type InterruptKernelSuccessful = { type: "INTERRUPT_KERNEL_SUCCESSFUL" };
 
 export const INTERRUPT_KERNEL_FAILED = "INTERRUPT_KERNEL_FAILED";
-export type InterruptKernelFailed = {
-  type: "INTERRUPT_KERNEL_FAILED",
-  payload: Error,
-  error: true
-};
+export type InterruptKernelFailed = ErrorAction<"INTERRUPT_KERNEL_FAILED">;
```
theengineear, line 403: 💯

packages/core/src/epics/websocket-kernel.js
```diff
@@ -14,33 +14,26 @@ import { of } from "rxjs/observable/of";
...
+import * as actions from "../actions";
 import * as selectors from "../selectors";
-
-import {
-  LAUNCH_KERNEL,
-  LAUNCH_KERNEL_BY_NAME,
-  LAUNCH_KERNEL_SUCCESSFUL,
-  INTERRUPT_KERNEL
-} from "../actionTypes";
+import * as actionTypes from "../actionTypes";
```
theengineear, line 24: 👌

packages/core/src/epics/websocket-kernel.js
```diff
@@ -14,33 +14,26 @@ import { of } from "rxjs/observable/of";
...
 
 import { executeRequest, kernelInfoRequest } from "@nteract/messaging";
 
+function onlyJupyter(store: *) {
+  return filter(() => selectors.isCurrentHostJupyter(store.getState()));
+}
```
theengineear: IMO, this feels a little too clever. Just because you're passing the whole store into the function, if it was more than a one-liner, maybe, but I think the just seeing `filter(() => selectors.isCurrentHostJupyter(store.getState()))` is pretty clear. Anyways, just my opinion :)

packages/core/src/epics/websocket-kernel.js
```diff
@@ -82,8 +75,30 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+    })
+  );
+
+export const killKernelEpic = (action$: *, store: *) =>
+  action$.pipe(
+    ofType(actionTypes.KILL_KERNEL),
+    // This epic can only interrupt kernels on jupyter websockets
+    onlyJupyter(store),
+    // If the user fires off _more_ kills, we shouldn't interrupt the in-flight
+    // kill, instead doing it after the last one happens
```
theengineear: 🥇

packages/core/src/epics/websocket-kernel.js
```diff
@@ -82,8 +75,30 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+      const state = store.getState();
+      const serverConfig = selectors.serverConfig(state);
+      const kernel = selectors.currentKernel(state);
+      const id = kernel.id;
+
+      return kernels
+        .kill(serverConfig, id)
+        .pipe(
+          map(() => actions.killKernelSuccessful()),
+          catchError(err => actions.killKernelFailed(err))
```
theengineear: 🤔 this one always strikes me as funny. But... the thing that we're trying to return is the ajax response, which is why we have to flatten it. If that fails, we hit the `catchError`, which needs to match the nested structure of the failed ajax, right? I.e., don't we need like a `catchError(err => of(actions.killKernelFailed(err)))` here?

packages/core/src/epics/websocket-kernel.js
```diff
@@ -82,8 +75,30 @@ export const interruptKernelEpic = (action$: *, store: *) =>
       return kernels
         .interrupt(serverConfig, id)
         .pipe(
-          map(() => interruptKernelSuccessful()),
-          catchError(err => interruptKernelFailed(err))
+          map(() => actions.interruptKernelSuccessful()),
+          catchError(err => actions.interruptKernelFailed(err))
```
theengineear: Hmm. Now that I'm thinking about it, should this be wrapped as well?

packages/core/src/epics/websocket-kernel.js
theengineear: Ah, yah, this spec explains what's going on: https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/catchError.ts#L14

When one of the observables fails, you just switch to the one provided in the catchError instead. So yah, we definitely need this to be an observable.

theengineear (changes requested):
Looks great! I think some of the `catchError` statements need to be updated though.

packages/core/src/epics/websocket-kernel.js
rgbkrk: ok, I can move that back in

packages/core/src/epics/websocket-kernel.js
rgbkrk: ooooops

rgbkrk: Fixed it for both interrupt and kill just now!

rgbkrk:
It's too bad that flow doesn't recognize that the function passed to catchError must return an observable. I'm going to update the flow definition.

rgbkrk:
Hmmmm... Looks like `catchError` is `any` currently.

rgbkrk:
Welp, it looks like _all_ operators are currently untyped. 😭  Of course, I've already reported this (and made some PRs): https://github.com/flowtype/flow-typed/issues/1688

/cc @jayphelps

packages/core/src/epics/websocket-kernel.js
```diff
@@ -60,17 +49,17 @@ export const launchWebSocketKernelEpic = (action$: *, store: *) =>
...
     })
   );
 
 export const interruptKernelEpic = (action$: *, store: *) =>
   action$.pipe(
-    ofType(INTERRUPT_KERNEL),
+    ofType(actionTypes.INTERRUPT_KERNEL),
     // This epic can only interrupt kernels on jupyter websockets
-    filter(() => selectors.isCurrentKernelJupyterWebsocket(store.getState())),
+    filter(() => selectors.isCurrentHostJupyter(store.getState())),
```
theengineear, line 62: Thanks!

packages/core/src/epics/websocket-kernel.js
```diff
@@ -82,8 +71,30 @@ export const interruptKernelEpic = (action$: *, store: *) =>
...
+      const state = store.getState();
+      const serverConfig = selectors.serverConfig(state);
+      const kernel = selectors.currentKernel(state);
+      const id = kernel.id;
+
+      return kernels
+        .kill(serverConfig, id)
+        .pipe(
+          map(() => actions.killKernelSuccessful()),
+          catchError(err => of(actions.killKernelFailed(err)))
```
theengineear, line 97: 🙇

theengineear (approved):

PR #2535, opened 2018-02-15
create readdirObservable, statObservable, more fully type

rgbkrk:
Two more functions for `fs-observable` and some additional type annotations since `bindNodeCallback` is not typed...

In support of #2532 (mostly to fully flesh out the "Contents API" for desktop).

PR #2538, opened 2018-02-15
Even better tests for fs-observable

rgbkrk:
✨ 

Follow up to #2535 to test `readdirObservable` as requested by @MSeal.

### DIFF
--- a/applications/desktop/__mocks__/fs.js
+++ b/applications/desktop/__mocks__/fs.js
@@ -1,6 +1,0 @@
-module.exports = {
-  unlinkSync: jest.fn(),
-  unlink: jest.fn(),
-  existsSync: jest.fn(),
-  writeFile: (name, data, callback) => callback(null)
-};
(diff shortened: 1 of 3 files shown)


rgbkrk:
Extra note as to why I moved the `fs` mock from desktop:

```
jest-haste-map: duplicate manual mock found:
  Module name: fs
  Duplicate Mock path: /Users/kylek/code/src/github.com/nteract/nteract-ext/packages/fs-observable/__mocks__/fs.js
This warning is caused by two manual mock files with the same file name.
Jest will use the mock file found in:
/Users/kylek/code/src/github.com/nteract/nteract-ext/packages/fs-observable/__mocks__/fs.js
 Please delete one of the following two files:
 /Users/kylek/code/src/github.com/nteract/nteract-ext/applications/desktop/__mocks__/fs.js
/Users/kylek/code/src/github.com/nteract/nteract-ext/packages/fs-observable/__mocks__/fs.js
```

Because we use a global jest configuration, all mocks go together. We could/should put the mocks at the top level of the monorepo then. For now, this is alright since we (currently) only test against fs-observable usages.

PR #2532, opened 2018-02-15
[WIP] Switch load to fetch content

rgbkrk:
Iterating towards making the desktop app loading epic use the same `fetchContent*` actions that `@nteract/core` uses now (for web apps). Putting this into place to make #2516 easier to digest.

Not finished yet, littering this with `FIXME` while I get it worked out. Expect a bunch of commits. 😅

PR #2540, opened 2018-02-15
create contents API responses on Desktop

rgbkrk:
In order to get `@nteract/core` and Desktop inline, I've made the `LOAD` action create a contents API response. It gets dropped at the moment (you can `tap` it to see the payload).

Also a precursor to #2532.

### DIFF
--- a/applications/desktop/src/notebook/epics/loading.js
+++ b/applications/desktop/src/notebook/epics/loading.js
@@ -1,13 +1,15 @@
 /* @flow */
 
 import * as path from "path";
+import * as fs from "fs";
 
 import { of } from "rxjs/observable/of";
+import { forkJoin } from "rxjs/observable/forkJoin";
 import { map, tap, mergeMap, switchMap, catchError } from "rxjs/operators";
 
 import { ActionsObservable, ofType } from "redux-observable";
 
-import { readFileObservable } from "fs-observable";
+import { readFileObservable, statObservable } from "fs-observable";
 
 import { monocellNotebook, fromJS, parseNotebook } from "@nteract/commutable";
 import type { Notebook, ImmutableNotebook } from "@nteract/commutable";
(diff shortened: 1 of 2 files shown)


applications/desktop/src/notebook/epics/loading.js
```diff
@@ -52,13 +54,70 @@ export const convertRawNotebook = (filename: string, data: string) => ({
...
+}
+
 /**
  * Loads a notebook and launches its kernel.
  *
  * @param  {ActionObservable}  A LOAD action with the notebook filename
  */
-export const loadEpic = (actions: ActionsObservable<*>) =>
-  actions.pipe(
+export const loadEpic = (action$: ActionsObservable<*>) =>
```
theengineear, line 118: ;)

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -67,9 +126,24 @@ export const loadEpic = (actions: ActionsObservable<*>) =>
...
+          content,
+          stat
+        })
+      ).pipe(
+        map(({ content, stat }) =>
+          createContentsResponse(filepath, stat, content)
+        ),
+        // TODO: Switch this all to a response to fetchContents instead
+        //       and handle accordingly
+        map(({ content }) => convertRawNotebook(action.filename, content)),
```
theengineear: Could you use the `filename` from the `createContentsResponse` at this point? I.e., instead of `action.filename`?

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -67,9 +126,18 @@ export const loadEpic = (actions: ActionsObservable<*>) =>
...
+
+      return forkJoin(
+        readFileObservable(filepath),
+        statObservable(filepath),
+        // Project onto the Contents API response
+        (content, stat) => createContentsResponse(filepath, stat, content)
+      ).pipe(
+        // TODO: Switch this all to a response to fetchContents instead
+        //       and handle accordingly
+        map(({ content }) => convertRawNotebook(action.filename, content)),
```
theengineear: Gah, comment got swallowed. Could you use the `filename` from the `createContentsResponse` here?

theengineear:
Nice! One question in there for ya! Then I'll merge it for ya!

applications/desktop/src/notebook/epics/loading.js
rgbkrk: Yeah I can do that, good idea.

rgbkrk:
All set!

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -67,9 +126,18 @@ export const loadEpic = (actions: ActionsObservable<*>) =>
...
+
+      return forkJoin(
+        readFileObservable(filepath),
+        statObservable(filepath),
+        // Project onto the Contents API response
+        (content, stat) => createContentsResponse(filepath, stat, content)
+      ).pipe(
+        // TODO: Switch this all to a response to fetchContents instead
+        //       and handle accordingly
+        map(({ content, filePath }) => convertRawNotebook(filePath, content)),
```
theengineear: I think this is just `path` actually. Right? I.e., not `filePath`

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -52,13 +54,70 @@ export const convertRawNotebook = (filename: string, data: string) => ({
...
+
+    // TODO: Mimetype detection
+    return {
+      type: "file",
+      mimetype: null,
+      format: "text",
+      content,
+      writable: true,
+      name,
+      path: filePath,
```
theengineear, line 104: Just pointing out what I'm talking about below ^^

applications/desktop/src/notebook/epics/loading.js
rgbkrk: Yeah, `path` should be used here. `path` (the node module) is also in scope so I was working around it. The current commit should now be clean (and I've exercised it quite a bit locally).

applications/desktop/src/notebook/epics/loading.js
rgbkrk: This was the one I worked around with since node's `path` module needs to be used in this function, the parameter passed was `filepath` and I was reconstructing it (pointlessly) as `filePath`.

PR #2541, opened 2018-02-15
That's so fetch!

rgbkrk:
Desktop is now just like the webapp, in that it creates Contents API responses! No more `load` action or `LOAD` actionType. 😄 

This also fixes up `fetchContentFailed` returning the right payload now.

All of this is in support of making #2516 easier to digest.

### DIFF
--- a/applications/desktop/src/notebook/epics/loading.js
+++ b/applications/desktop/src/notebook/epics/loading.js
@@ -109,7 +109,7 @@ function createContentsResponse(
  *
  * @param  {ActionObservable}  A LOAD action with the notebook filename
  */
-export const loadEpic = (action$: ActionsObservable<*>) =>
+export const fetchContentEpic = (action$: ActionsObservable<*>) =>
   action$.pipe(
     ofType(actionTypes.FETCH_CONTENT),
     tap(action => {
--- a/applications/desktop/src/notebook/epics/loading.js
+++ b/applications/desktop/src/notebook/epics/loading.js
@@ -5,30 +5,37 @@ import * as fs from "fs";
 
 import { of } from "rxjs/observable/of";
 import { forkJoin } from "rxjs/observable/forkJoin";
-import { map, tap, mergeMap, switchMap, catchError } from "rxjs/operators";
+import {
+  map,
+  tap,
+  mergeMap,
+  switchMap,
+  catchError,
+  timeout
+} from "rxjs/operators";
 
 import { ActionsObservable, ofType } from "redux-observable";
 
 import { readFileObservable, statObservable } from "fs-observable";
 
+import * as Immutable from "immutable";
 import { monocellNotebook, fromJS, parseNotebook } from "@nteract/commutable";
 import type { Notebook, ImmutableNotebook } from "@nteract/commutable";
 
 import * as actionTypes from "@nteract/core/actionTypes";
 import * as actions from "@nteract/core/actions";
 
+import type { SetNotebookAction } from "@nteract/core/src/actionTypes";
+
 /**
- * Creates a new kernel based on the language info in the notebook.
- *
- * @param  {String}  filename  The filename of the notebook being loaded
- * @param  {Immutable<Map>}  notebook  The notebook to extract langauge info from
- *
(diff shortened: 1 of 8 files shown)


applications/desktop/src/notebook/epics/loading.js
```diff
@@ -117,41 +110,53 @@ function createContentsResponse(
...
       );
     })
   );
 
+export const launchKernelWhenNotebookSetEpic = (
+  action$: ActionsObservable<*>
+) =>
+  action$.pipe(
+    ofType(actionTypes.SET_NOTEBOOK),
+    map((action: SetNotebookAction) => {
```
theengineear, line 151: I gotta remember to set the type here! Nice.

packages/core/src/epics/contents.js
```diff
@@ -114,6 +114,7 @@ export function saveContentEpic(
 }
 
 // When content gets loaded, if it's a notebook, set it up as the notebook
+// TODO: should this be renamed as "fetchContentFulfilledEpic"
```
theengineear: Potentially. Is there a case when we have multiple epics listening for the same actions? If not, it's sorta nice when the naming lines up.

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -117,41 +110,53 @@ function createContentsResponse(
  */
 export const loadEpic = (action$: ActionsObservable<*>) =>
   action$.pipe(
-    ofType(actionTypes.LOAD),
+    ofType(actionTypes.FETCH_CONTENT),
```
theengineear: Cool, so the `load` action is gone now, and we just hook up this epic in place of the core `fetchContent` epic. I wonder if we should actually call this the `fetchContentEpic` now so that it's clear that we're using the desktop-specific `fetchContentEpic`, not the `core` `fetchContentEpic`?

rgbkrk: I like that, I'll switch the name up here.

packages/core/src/epics/contents.js
rgbkrk: There won't be more than one epic listening for this (I think).

rgbkrk:
Woo, I fixed the tests! On to addressing feedback.

applications/desktop/src/notebook/epics/loading.js
rgbkrk: Pretty handy since we don't have the rxjs operator types via flow. 😄

packages/core/src/actions.js
```diff
@@ -133,9 +133,9 @@ export const fetchContentFulfilled = (payload: {
 
 export const fetchContentFailed = (payload: {
   path: string,
-  error: Object
-}): FetchContentFulfilled => ({
-  type: actionTypes.FETCH_CONTENT_FULFILLED,
+  error: Error
+}): FetchContentFailed => ({
+  type: actionTypes.FETCH_CONTENT_FAILED,
```
theengineear, line 138: 👌 ha, i totally had this in a stashed change locally. you beat me :)

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -160,11 +166,19 @@ export const loadEpic = (action$: ActionsObservable<*>) =>
...
   action$.pipe(
     ofType(actionTypes.NEW_NOTEBOOK),
-    switchMap(action =>
-      of(
-        // TODO: Switch filename to an optional argument
-        actions.setNotebook(null, monocellNotebook),
-        actions.launchKernel(action.kernelSpec, action.cwd)
-      )
-    )
+    switchMap(action => {
```
theengineear: This is just synchronously returning an action, right? Couldn't we just `map` here?

packages/core/src/epics/contents.js
theengineear: Then, "yes". 😄 but you don't need todo that in this pr.

theengineear (approved):
:tada:

packages/core/src/epics/contents.js
rgbkrk: 😄 I'll take it out now.

PR #2545, opened 2018-02-16
[jext] create a title bar with adaptive theming, title, and logo

### LINKED ISSUE #2459: [jext] Create nteract header and file title bar
It should (could) be a shared thing in `@nteract/core`, as we have something similar on both commuter and play.

rgbkrk:
Closes #2459.

theme-switching

* [x] Bring in the nteract logo as an SVG React Component
* [x] Allow it to adapt to changing theme and filename
* [x] Test it
* [x] Hook it up to redux to adapt to:
  * [x] theme
  * [x] filename

I'm going to hold myself back from trying to tweak CSS too much here, open to some fresh style contributions.


rgbkrk:
theme-switching

rgbkrk:
Last up is tests and then I'm ready for review.

packages/core/src/components/title-bar/index.js
```diff
@@ -0,0 +1,39 @@
...
+  <React.Fragment>
+    <header>
+      <Logo height={20} />
+      <p>{props.title}</p>
+    </header>
+    <style jsx>{`
+      header {
+        display: flex;
+        justify-content: flex-start;
+        background-color: var(--theme-menu-bg, rgb(250, 250, 250));
```
theengineear, line 26: 👌

packages/core/src/components/title-bar/index.js
```diff
@@ -0,0 +1,39 @@
...
+        padding: 10px 16px;
```
theengineear, line 27: We have some `spacing` css variables at this point. Don't feel like you need to update this now, but it might help with consistency in the future.

packages/core/src/components/title-bar/index.js
```diff
@@ -32,8 +36,25 @@ const TitleBar = (props: TitleBarProps) => (
...
+    .split("/")
+    .pop()
+    .split(".ipynb")
+    .shift()
+});
+
+const mapDispatchToProps = dispatch => ({
+  titleChanged: () => {
+    // TODO: Once the content refs PR is finished use the ref to change
+    // the filename, noting that the URL path should also change
```
theengineear: 👌

packages/core/src/selectors/index.js
```diff
@@ -237,3 +237,8 @@ export const modalType = createSelector(
   (state: AppState) => state.modals.modalType,
   identity
 );
+
+export const currentTheme: (state: *) => string = createSelector(
```
theengineear, line 241: ✨

theengineear (commented):
Nice! This all looks great!

theengineear (approved):

packages/core/src/components/title-bar/index.js
rgbkrk: Ah, cool. What should I be using here?

theengineear: https://github.com/nteract/nteract/blob/master/packages/styles/src/config.json#L145

I think `var(--nt-spacing-m)` and then `var(--nt-spacing-l)` might work? There are always going to be places where you just need to push pixels a little bit, but I think we should strive for as much consistency here as possible

PR #2550, opened 2018-02-19
[WIP] create a scales segment of our styles

rgbkrk:
This incorporate's two of [vega's color scales](https://github.com/vega/vega/wiki/Scales), category10 and category20.

I don't know _how_ we'd want to do categorical palettes with our styles setup, so I'm just starting this to kick it off and get input from @theengineear.

I'm imagining the way we'd use this is with:

```css
.completion-type-keyword:before {
  content: "k";
}
.completion-type-keyword {
  background-color: var(--theme-scale-idx-1);
}

.completion-type-class:before {
  content: "C";
}
.completion-type-class {
  background-color: var(--theme-scale-idx-2);
}
```

Whereas the theme would set / use a particular scale. For the light theme, they could use:

```css
--theme-scale-idx-1: var(--nt-scale-vega-category10-1)
```

Thoughts?

### DIFF
--- a/packages/styles/src/config.json
+++ b/packages/styles/src/config.json
@@ -97,6 +97,42 @@
           "lightest": "hsl(33, 75%, 97%)"
         }
       },
+      "scale": {
+        "category10": [
+          "#1f77b4",
+          "#ff7f0e",
+          "#2ca02c",
+          "#d62728",
+          "#9467bd",
+          "#8c564b",
+          "#e377c2",
+          "#7f7f7f",
+          "#bcbd22",
+          "#17becf"
+        ],
+        "category20": [
+          "#1f77b4",
+          "#aec7e8",
+          "#ff7f0e",
+          "#ffbb78",
+          "#2ca02c",
+          "#98df8a",
+          "#d62728",
+          "#ff9896",
+          "#9467bd",
+          "#c5b0d5",
+          "#8c564b",
+          "#c49c94",
+          "#e377c2",
+          "#f7b6d2",
+          "#7f7f7f",
+          "#c7c7c7",
+          "#bcbd22",
+          "#dbdb8d",
+          "#17becf",
+          "#9edae5"
+        ]
+      },
       "borderRadius": {
         "none": 0,
         "s": "1px",


theengineear:
As far as `--theme-scale-idx-[index]`, i think we should likely give ourselves wiggle room to have a multiple scales, right? I.e., we might want to call it `--theme-scale-category10-1`. Right?

theengineear:
Also, these will change with light/dark theme ya? We'll have to be pretty careful about these if we're going to be changing colors in a chart based on theming... That might get confusing fast.

packages/styles/src/config.json
```diff
@@ -97,6 +97,42 @@
           "lightest": "hsl(33, 75%, 97%)"
         }
       },
+      "scale": {
+        "category10": [
+          "#1f77b4",
```
theengineear, line 102: If we set up the scales this way, can we follow the convention that we just `.join('-')` the keys together? I'm fine with it being an array since the keys here are just indices. I just want _some_ common rule that helps me look at this config and guess that I'd get `#1f77b4` from `--nt-scale-category10-0`. I.e., you were slipping an `idx` in there, which isn't specified anywhere in the config, so you'd just have to _know_ that.

theengineear: That said, this format in the config file should work. :+1:

rgbkrk:
> As far as `--theme-scale-idx-[index]`, i think we should likely give ourselves wiggle room to have a multiple scales, right? I.e., we might want to call it `--theme-scale-category10-1`. Right?

To give ourselves room for a 10 color palette and a 20 color palette?

Maybe what we do, on the theme side, is make this specific to the intent. `--theme-completion-idx-1` or if the completion types are truly known `--theme-completion-function-color`. Then it's all just based on the naming instead of indices (even though they're index based underneath).

theengineear:
One more comment... about comments. @rgbkrk and I chatted a little about wanting commenting for this config file. I think we'll likely change the input file to YAML so that it can support comments. What we *don't* want to do is add JS since it makes it far too easy to get too magical in what should be a very simple config file.

rgbkrk:
> Also, these will change with light/dark theme ya? We'll have to be pretty careful about these if we're going to be changing colors in a chart based on theming... That might get confusing fast.

Hopefully not confusing in the sense that it would stay consistent for the given theme. People shouldn't toggle between and expect them to stay the same (I think).

theengineear:
>People shouldn't toggle between and expect them to stay the same (I think).

Maybe? I'm actually not sure on this. But I'm okay with making the assumption for now and seeing how it goes.

theengineear:
>Maybe what we do, on the theme side, is make this specific to the intent. `--theme-completion-idx-1`

Yah, naming the themes is what I'm getting at for sure. If you want, you can do `--theme-category10-[index]` and then alias it to `--theme-completion-[index]` as well. But either way, let's sneak another piece of context in there as you're suggesting :+1:.

rgbkrk:
> Yah, naming the themes is what I'm getting at for sure. If you want, you can do `--theme-category10-[index]` and then alias it to `--theme-completion-[index]` as well. But either way, let's sneak another piece of context in there as you're suggesting :+1:.

I'm assuming here too, category10 just means "a category with a 10 color range" rather than vega's literal "category10" (which is the one we're using here).

rgbkrk:
Tabling this for later.

PR #2551, opened 2018-02-19
darker than the night's sky

rgbkrk:
Figured I might as well sneak in the `darker` and `darkest` variations on the already dark `midnight`, since the dark theme is implicitly using it.

🕶 🌃

### DIFF
--- a/packages/styles/src/config.json
+++ b/packages/styles/src/config.json
@@ -71,6 +71,9 @@
         },
         "midnight": {
           "": "hsl(0, 0%, 15%)",
+          "dark": "hsl(0, 0%, 10%)",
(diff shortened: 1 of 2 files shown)


packages/styles/src/config.json
```diff
@@ -71,6 +71,9 @@
         },
         "midnight": {
           "": "hsl(0, 0%, 15%)",
+          "dark": "hsl(0, 0%, 10%)",
+          "darker": "hsl(0, 0%, 5%)",
+          "darkest": "hsl(0, 0%, 0%)",
```
theengineear, line 76: so dark.

rgbkrk: If only I could have used negative numbers!

PR #2552, opened 2018-02-19
Alias more for jext

rgbkrk:
Figured I'd make this setup even nicer for the jupyter extension. I might go the way of rx and make a specific aliases module to put this in one place.


theengineear:
First off, I'm totally going to merge this. However, I think `aliases` can be a little confusing for folks new to a codebase (and for IDE's!). We're saving ~4 characters (per usage) and adding a fair amount of complexity. Dunno if it's totally worth it. OK, now I'll merge and we can discuss later.

theengineear:
Or, is there some other benefit I'm missing?

rgbkrk:
The benefit of these aliases is that you don't have to rebuild those packages to see updates in the jupyter extension UI. The benefit of a library approach to it is that we'd maintain the webpack aliases in one location.

theengineear:
>The benefit of these aliases is that you don't have to rebuild those packages to see updates in the jupyter extension UI.

Ah, ok, that makes sense then 💯

PR #2553, opened 2018-02-19
So nice, we aliased it twice!

theengineear:
This duplication got aliased twice this morning. Just removing the dups.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
@@ -18,8 +18,6 @@ const aliases = {
   "@nteract/editor": "@nteract/editor/src",
   "@nteract/commutable": "@nteract/commutable/src",
   "@nteract/dropdown-menu": "@nteract/dropdown-menu/src",
-  "@nteract/transforms": "@nteract/transforms/src",
-  "@nteract/transform-vdom": "@nteract/transform-vdom/src",
   "@nteract/transform-model-debug": "@nteract/transform-model-debug/src",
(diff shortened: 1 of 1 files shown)


applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
```diff
@@ -8,21 +8,23 @@ const isProd = nodeEnv === "production";
...
-  "@nteract/mathjax": "@nteract/mathjax/src",
-  "@nteract/core": "@nteract/core/src",
-  "@nteract/messaging": "@nteract/messaging/src",
-  "@nteract/editor": "@nteract/editor/src",
-  "@nteract/commutable": "@nteract/commutable/src",
-  "@nteract/dropdown-menu": "@nteract/dropdown-menu/src",
-  "@nteract/transform-model-debug": "@nteract/transform-model-debug/src",
-  "rx-jupyter": "rx-jupyter/src",
-  "rx-binder": "rx-binder/src",
-  ...rxAliases
```
theengineear: @rgbkrk this line was failing for me when I ran `npm run app:jupyter-extension`. Do we (1) have something transpiling this file and (2) if we do, any thoughts on why it wouldn't be working on my machine?

theengineear:
AH! I'm on Node 8.2.1, I think you might just be ahead of me. Looks like it's supported in Node 8.6.0+. Let me try that real quick.

theengineear:
^^ OK, updated my node and away we go. Duplicate declaration still stands 😛

rgbkrk:
Whoops! Yeah I upgraded recently.

Luckily I've got an automated setup for this that I'll post as a PR soon.

PR #2554, opened 2018-02-20
Webpack configurator

rgbkrk:
This makes our webpack setups consistent amongst each other, making reloading of next.js apps in tandem with our packages happen.

Note: hot reloading doesn't work properly for any apps using a `<MathJax.Context>` as well as those with `shouldComponentUpdate` blockers. I'm hopeful that for the mathjax context bits, we'll have hot reloading working well once React 16.3 comes out with the well supported context API.

PR #2558, opened 2018-02-20
unify the babel configuration

rgbkrk:
Just the babel unification bits out of #2554 

One consistent babel configuration across all our packages. Less `className` pollution from repeated styled-jsx setups.


rgbkrk:
Verified this is good to go across the apps, in dev and prod.

mpacer:
having global .babelrc files seems reasonable, and deduplicating jsx classes seems to surely be an improvement.

mpacer:
merging

PR #2559, opened 2018-02-20
pin packages within desktop to monorepo versions

rgbkrk:
Happened to notice these were out of wack. This happened because of a failed release step last time where desktop was having trouble building.


mpacer:
I don't know the differences for these version numbers, but it passes the tests. So, merging.

rgbkrk:
Yeah, you'd have to trust that I matched up the numbers from the other packages inside the repo (which is how I figured out these were out of sync).

PR #2562, opened 2018-02-21
Stop using createImmutableMapReducer.

theengineear:
We decided that it was more confusing than helpful to include this
construct. It was originally included in Play where we had a highly-
nested structure. However, in core, we're going for a much flatter
structure where this doesn't really provide much else.


applications/jupyter-extension/nteract_on_jupyter/app/store.js
```diff
@@ -49,6 +50,9 @@ const rootReducer = combineReducers({
   comms,
   config,
   modals,
+  communication: combineReducers({
+    kernelspecsByRef: communicationKernelspecsByRef
+  }),
```
theengineear, line 55: Eh, we're not using this yet, but might as well actually hook it up 😛

theengineear:
@rgbkrk as we talked about :)

applications/jupyter-extension/nteract_on_jupyter/app/store.js
rgbkrk: Works for me!

rgbkrk: It's fine iteration!

packages/core/src/reducers/kernelspecs.js
```diff
@@ -4,59 +4,51 @@ import {
...
+        makeCommunicationKernelspecs({ loading: false, error: null })
+      );
     case actionTypes.FETCH_KERNELSPECS_FAILED:
-      return action.payload.error;
+      return state.set(
+        action.payload.kernelspecsRef,
+        makeCommunicationKernelspecs({
+          loading: false,
+          error: action.payload.error
+        })
```
rgbkrk, line 32: These are perfect and I can completely understand them in one quick read.

rgbkrk (approved):

packages/core/src/reducers/kernelspecs.js
```diff
@@ -4,59 +4,51 @@ import {
...
+) => {
   switch (action.type) {
     case actionTypes.FETCH_KERNELSPECS_FULFILLED:
-      return Immutable.Map(action.payload.kernelspecs).map(
-        (kernelspec: KernelspecProps) => makeKernelspec(kernelspec)
+      return state.set(
+        action.payload.kernelspecsRef,
+        Immutable.Map(action.payload.kernelspecs).map(
+          (kernelspec: KernelspecProps) => makeKernelspec(kernelspec)
+        )
```
rgbkrk, line 49: 👌

rgbkrk: I'd probably leave comments in here that say

```
// This sets the kernelspecs as kernelspecsRef to contain
// a map of name -> KernelspecRecord
```

PR #2563, opened 2018-02-21
move records into core

### LINKED ISSUE #2408: Notebook types are wrong in @nteract/types/core/records

rgbkrk:
Closes #2408, moving core records into `@nteract/core` and out of `@nteract/types`.

Necessary follow up PRs:

* Correct type inside of `selectors`.
* Address the keypath `$FlowFixMe`s

There will be some duplication in the middle of this, followed by a magical rebase that makes it look like I did it all in one ✨


rgbkrk:
Rebased this against master, still more to go.

rgbkrk:
This will definitely be rebased...

rgbkrk:
Stopping this one without doing selectors. I'll do a separate PR for that ball of wax.

applications/desktop/__tests__/renderer/native-window-spec.js
```diff
@@ -4,7 +4,7 @@ import { remote } from "electron";
 import { from } from "rxjs/observable/from";
 
 import * as nativeWindow from "../../src/notebook/native-window";
-import { makeAppRecord, makeDocumentRecord } from "@nteract/types/core/records";
+import { makeAppRecord, makeDocumentRecord } from "@nteract/core/records";
```
rgbkrk, line 7: As you'll see in most of this PR, it was a lot of turning `@nteract/types/core/records` into `@nteract/core/records`

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -39,7 +39,9 @@ import * as jmp from "jmp";
 
 import type { NewKernelAction } from "@nteract/core/actionTypes";
 
-import type { KernelInfo, LocalKernelProps } from "@nteract/types/core/records";
+import type { KernelInfo, LocalKernelProps } from "@nteract/core/src/records";
```
rgbkrk, line 42: Since the top level `packages/core/records` doesn't re-export the types, we pull these out of `src`.

applications/desktop/src/notebook/index.js
```diff
@@ -26,8 +26,8 @@ import {
   makeAppRecord,
   makeDesktopHostRecord,
   makeDocumentRecord,
-  CommsRecord
-} from "@nteract/types/core/records";
+  makeCommsRecord
```
rgbkrk, line 29: Wise change I probably could have done in another PR. It was confusing me as to if this was a type or something else, when it was another record creator.

applications/desktop/src/notebook/menu.js
```diff
@@ -269,6 +269,7 @@ export function exportPDF(
   notificationSystem: *
 ): void {
   const state = store.getState();
+  // $FlowFixMe: This should be using a selector first and foremost
```
rgbkrk, line 272: I'll fix this in a follow up PR when I do the selector bits.

packages/core/src/actionTypes.js
```diff
@@ -1,9 +1,7 @@
 /* @flow */
 
-import type {
-  KernelspecsRef,
-  KernelspecProps
-} from "@nteract/types/core/records";
+import type { KernelspecsRef, KernelspecProps } from "./records";
```
rgbkrk: I'll merge this into a later import in this file.

packages/core/src/records/index.js
```diff
@@ -142,7 +150,7 @@ export type NotebookMetadata = {
   // orig_nbformat?: number,
 };
 
-export type Notebook = {
+export type NotebookRecordProps = {
```
rgbkrk, line 153: Note: This type isn't used at all here.

packages/core/src/records/index.js
```diff
@@ -168,7 +176,7 @@ export const makeAppRecord: RecordFactory<AppRecordProps> = Record({
   host: null,
   githubToken: null,
   notificationSystem: {
-    addNotification: msg => {
+    addNotification: (msg: { level?: "error" | "warning" }) => {
```
rgbkrk, line 179: Did this just to increase coverage.

packages/core/src/records/index.js
```diff
@@ -192,7 +200,10 @@ export const makeAppRecord: RecordFactory<AppRecordProps> = Record({
 export type AppRecord = RecordOf<AppRecordProps>;
 
 type DocumentRecordProps = {
-  notebook: ?Notebook,
+  // TODO: This _needs_ to become a Record
+  notebook: ?Map<string, any>,
+  savedNotebook: ?Map<string, any>,
```
rgbkrk, line 205: We were missing this property on the document record.

applications/desktop/__tests__/renderer/native-window-spec.js
theengineear: I've added `@nteract/core/types`, do you think we should keep `/records` in that? We're starting to duplicate work as you mentioned, we should make sure we have a plan to resolve the duplication in the future. Thoughts?

PR #2567, opened 2018-02-21
Add hosts to core state

theengineear:
Hosted!

screen shot 2018-02-21 at 10 44 21 am

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.js
@@ -54,7 +54,7 @@ function createApp(jupyterConfigData: JupyterConfigData) {
 
       // TODO: we should likely be passing in a hostRef to fetchContent too.
       store.dispatch(fetchContent({ path: jupyterConfigData.contentsPath }));
-      store.dispatch(fetchKernelspecs({ kernelspecsRef }));
+      store.dispatch(fetchKernelspecs({ hostRef, kernelspecsRef }));
     }
 
     render(): React$Element<any> {
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.js
@@ -12,9 +12,12 @@ import type { JupyterConfigData } from "./store";
 
 import { NotebookApp, Styles } from "@nteract/core/providers";
 
-import { fetchKernelspecs, fetchContent } from "@nteract/core/actions";
+import { addHost, fetchKernelspecs, fetchContent } from "@nteract/core/actions";
 
-import { createKernelspecsRef } from "@nteract/core/src/types/state/refs";
+import {
+  createHostRef,
+  createKernelspecsRef
+} from "@nteract/core/src/types/state/refs";
 
 import {
   ModalController,
@@ -30,7 +33,26 @@ function createApp(jupyterConfigData: JupyterConfigData) {
     notificationSystem: NotificationSystem;
 
     componentDidMount(): void {
+      const hostRef = createHostRef();
       const kernelspecsRef = createKernelspecsRef();
+      store.dispatch(
+        addHost({
+          hostRef,
+
+          // TODO: are we missing some host info here?
+          host: {
+            // id: string,
+            type: "jupyter",
+            // kernelIds: Array<Id>,
+            token: jupyterConfigData.token,
+            serverUrl: location.origin + jupyterConfigData.baseUrl
+            // crossDomain: boolean,
+            // rootContentRef: Ref,
+          }
+        })
+      );
+
+      // TODO: we should likely be passing in a hostRef to fetchContent too.
       store.dispatch(fetchContent({ path: jupyterConfigData.contentsPath }));
       store.dispatch(fetchKernelspecs({ kernelspecsRef }));
     }
--- a/packages/core/src/actionTypes.js
+++ b/packages/core/src/actionTypes.js
@@ -5,6 +5,8 @@ import type {
   KernelspecProps
 } from "@nteract/types/core/records";
 import type { ChildProcess } from "child_process"; // eslint-disable-line no-unused-vars
+import type { HostRef } from "./types/state/refs";
+import type { HostRecord } from "./types/state/entities/hosts";
 
 import type {
   ImmutableNotebook,
@@ -43,6 +45,12 @@ export type OpenModal = {
 export const CLOSE_MODAL = "CORE/CLOSE_MODAL";
 export type CloseModal = {
   type: "CORE/CLOSE_MODAL"
+};
+
+export const ADD_HOST = "CORE/ADD_HOST";
+export type AddHost = {
+  type: "CORE/ADD_HOST",
+  payload: { hostRef: HostRef, host: HostRecord }
 };
 
 export const FETCH_CONTENT = "CORE/FETCH_CONTENT";
--- a/packages/core/src/actionTypes.js
+++ b/packages/core/src/actionTypes.js
@@ -84,7 +84,8 @@ export const FETCH_KERNELSPECS = "CORE/FETCH_KERNELSPECS";
 export type FetchKernelspecs = {
   type: "CORE/FETCH_KERNELSPECS",
   payload: {
-    kernelspecsRef: KernelspecsRef
+    kernelspecsRef: KernelspecsRef,
+    hostRef: HostRef
   }
 };
 
@@ -93,6 +94,7 @@ export type FetchKernelspecsFulfilled = {
   type: "CORE/FETCH_KERNELSPECS_FULFILLED",
   payload: {
     kernelspecsRef: KernelspecsRef,
+    hostRef: HostRef,
     defaultKernelName: string,
     kernelspecs: { [string]: KernelspecProps }
   }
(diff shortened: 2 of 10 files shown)


theengineear:
~~oop, missed one of my rebase cues :p `temp: hosts types...`~~ ok ready!

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -30,9 +37,23 @@ function createApp(jupyterConfigData: JupyterConfigData) {
     notificationSystem: NotificationSystem;
 
     componentDidMount(): void {
+      const hostRef = createHostRef();
```
rgbkrk, line 36: Woohoo!

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -30,9 +37,23 @@ function createApp(jupyterConfigData: JupyterConfigData) {
...
+            // TODO: Use URL join, even though we know these are right
+            serverUrl: location.origin + jupyterConfigData.baseUrl
+          }
+        })
+      );
+
+      // TODO: we should likely be passing in a hostRef to fetchContent too.
       store.dispatch(fetchContent({ path: jupyterConfigData.contentsPath }));
-      store.dispatch(fetchKernelspecs({ kernelspecsRef }));
+      store.dispatch(fetchKernelspecs({ hostRef, kernelspecsRef }));
```
rgbkrk, line 57: Nice. 😄

packages/core/src/actionTypes.js
```diff
@@ -45,6 +43,36 @@ export type CloseModal = {
...
+  }
+};
+
+// TODO: fix host type
+export const FETCH_HOST_FULFILLED = "CORE/FETCH_HOST_FULFILLED";
+export type FetchHostFulfilled = {
+  type: "CORE/FETCH_HOST_FULFILLED",
+  payload: {
+    hostRef: HostRef,
+    host: *
```
rgbkrk: This would take `DesktopHostRecordProps | JupyterHostRecordProps`

packages/core/src/actionTypes.js
```diff
@@ -45,6 +43,36 @@ export type CloseModal = {
...
 };
 
+// TODO: fix config type
+export const FETCH_HOST = "CORE/FETCH_HOST";
+export type FetchHost = {
+  type: "CORE/FETCH_HOST",
+  payload: {
+    hostRef: HostRef,
+    hostType: "local" | "jupyter" | "binder",
+    hostConfig: *
```
rgbkrk: Not sure what this type would be since we don't have fetching specced at all yet. I would personally put a big `TODO: Determine what fetching a host even means -- this type is not used yet`

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -30,9 +37,23 @@ function createApp(jupyterConfigData: JupyterConfigData) {
...
+        fetchHost({
+          hostRef,
+          hostType: "jupyter",
+          hostConfig: {
+            token: jupyterConfigData.token,
+            // TODO: Use URL join, even though we know these are right
+            serverUrl: location.origin + jupyterConfigData.baseUrl
+          }
+        })
+      );
```
rgbkrk: Ok now that I've looked below, this is more like `setHost` rather than `fetchHost`. We're not making an ajax request whatsoever here. You could use the fulfilled version at this point.

packages/core/src/reducers/core/entities/hosts.js
```diff
@@ -0,0 +1,55 @@
...
+        }
+        default:
+          throw new Error(
+            `Unrecognized host type "${action.payload.host.type}".`
+          );
+      }
+    default:
+      return state;
+  }
+};
```
rgbkrk, line 42: Awesome, very easy to read here. ^^

packages/core/src/actionTypes.js
theengineear: Ah, yah, can do!

packages/core/src/actionTypes.js
theengineear: ^^ sg!

packages/core/src/selectors/index.js
```diff
@@ -246,6 +246,30 @@ export const currentTheme: (state: *) => string = createSelector(
...
+    state.core.getIn(["entities", "hosts", "byRef", hostRef]),
+  identity
+);
+
+export const currentHostCore = createSelector(
+  (state: AppState) => state.core.get("selectedHostRef"),
+  (state: AppState) => state.core.getIn(["entities", "hosts", "byRef"]),
+  (selectedHostRef, hostsByRef) => hostsByRef.get(selectedHostRef)
+);
+
```
rgbkrk: These are so awesome now.

packages/core/src/types/state/entities/hosts.js
```diff
@@ -0,0 +1,82 @@
...
+
+export type HostsRecordProps = {
+  byRef: Map<HostRef, HostRecord>,
+  refs: List<HostRef>
+};
+
+export const makeHostsRecord: RecordFactory<HostsRecordProps> = Record({
+  byRef: Map(),
+  refs: List()
+});
```
rgbkrk, line 86: We are definitely overlapping now with #2563. 😄

applications/jupyter-extension/nteract_on_jupyter/app/index.js
theengineear: Yah, that's what I was talking about last night. Is there a case where we *will* need an async `fetchHost` in the future?

packages/core/src/types/state/entities/hosts.js
theengineear: hehe, totally, i think that's ok though. i'd rather have some overlap and cruft for a little bit as we converge rather than have a huge pr that never gets merged 😛

rgbkrk: so true

applications/jupyter-extension/nteract_on_jupyter/app/index.js
rgbkrk: Not now. Let's create that in the future.

theengineear: Going with `addHost` since I think that's clearer. _Set_ sorta feels like _select_ to me.

theengineear:
OK @rgbkrk, final review?

applications/jupyter-extension/nteract_on_jupyter/app/index.js
rgbkrk: That makes sense. Do we also allow a way to select while adding this host?

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -30,9 +33,28 @@ function createApp(jupyterConfigData: JupyterConfigData) {
     notificationSystem: NotificationSystem;
 
     componentDidMount(): void {
+      const hostRef = createHostRef();
       const kernelspecsRef = createKernelspecsRef();
+      store.dispatch(
+        addHost({
+          hostRef,
+
+          // TODO: are we missing some host info here?
```
theengineear, line 42: Thoughts on this? I feel like we're not really fulfilling what we spec'd out in the plan.md doc.

packages/core/src/actions.js
```diff
@@ -115,6 +115,14 @@ export const closeModal = () => ({
   type: actionTypes.CLOSE_MODAL
 });
 
+export const addHost = (payload: {
+  hostRef: HostRef,
+  host: HostRecordProps
```
theengineear, line 120: OK?

rgbkrk: A+

applications/jupyter-extension/nteract_on_jupyter/app/index.js
rgbkrk: Probably want to have a `selected` property so that we're both adding and selecting this host.

rgbkrk: The host has its own ref too right?

theengineear: >Probably want to have a selected property so that we're both adding and selecting this host.

I _think_ we discussed having selected content, not a selected host? Right? Like our model puts the document at the center of everything, and then uses that to set all the other context?

theengineear: >The host has its own ref too right?

Nah, not unless it needs it, right? If we get to a place where the host needs it's own ref and it's not already being passed by a parent component, I'm down to add it, but I'm not sure it's totally necessary atm.

rgbkrk:
I'm going to go ahead and merge this so I can charge ahead with rebasing.

applications/jupyter-extension/nteract_on_jupyter/app/index.js
rgbkrk: The type here should be `HostRecordProps` for the payload (right?), so that would require a ref.

rgbkrk: > I think we discussed having selected content, not a selected host? Right? Like our model puts the document at the center of everything, and then uses that to set all the other context?

Oh right. Yeah.

PR #2570, opened 2018-02-21
Move /core/src/types/state to /core/src/state

theengineear:
@rgbkrk and I discussed moving `state` up a level since `types` is sorta a lie, the folder has more than types, e..g, `refs` has `createHostRef` and such.

The `core/src/records` is also going to do some moving around in a subsequent PR.


theengineear:
Eh, not so bad :)

packages/core/src/state/index.js
```diff
@@ -1,10 +1,10 @@
 // @flow
-import type { CommunicationRecordProps } from "./communication";
-import type { EntitiesRecordProps } from "./entities";
+import type { CommunicationRecordProps } from "./communication/index";
+import type { EntitiesRecordProps } from "./entities/index";
```
rgbkrk: Why do these have to reference `index` directly?

theengineear: hehe, cuz I let my IDE do the refact 😆 changing. thanks!

rgbkrk: Ok, I'll hold off on merging here. 😄

theengineear:
@rgbkrk thanks for the catch!

PR #2571, opened 2018-02-21
Switch menu.js over to using selectors fully

rgbkrk:
Call this a precursor to the flow type fixing of selectors.

### DIFF
--- a/packages/core/src/selectors/index.js
+++ b/packages/core/src/selectors/index.js
@@ -1,8 +1,18 @@
 // @flow
 
 // import type { AppState } from "../records";
-// TODO: Super wrong
-type AppState = Object;
+// FIXME FIXME FIXME SUPER WRONG FIXME FIXME FIXME
+type AppState = {
+  // The new way
+  core: Object,
+  modals: Object,
+
+  // The old way
+  app: Object,
+  comms: *,
+  document: Object,
+  config: Object
+};
 
 import { toJS, stringifyNotebook } from "@nteract/commutable";
 import * as Immutable from "immutable";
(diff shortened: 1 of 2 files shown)


packages/core/src/selectors/index.js
```diff
@@ -1,8 +1,18 @@
 // @flow
 
 // import type { AppState } from "../records";
-// TODO: Super wrong
-type AppState = Object;
+// FIXME FIXME FIXME SUPER WRONG FIXME FIXME FIXME
```
theengineear, line 4: 😆

packages/core/src/selectors/index.js
```diff
@@ -184,7 +194,7 @@ export const transientCellMap = createSelector(
 
 export const currentCellMap = createSelector([currentNotebook], notebook => {
   if (notebook) {
-    return notebook.get("cellMap");
+    return notebook.get("cellMap", Immutable.Map());
```
theengineear, line 197: 👍

packages/core/src/selectors/index.js
```diff
@@ -233,6 +243,19 @@ export const currentHiddenCellIds = createSelector(
   }
 );
 
+export const currentIDsOfHiddenOutputs = createSelector(
```
theengineear: `currentIdsOfHiddenOutputs`?

packages/core/src/selectors/index.js
```diff
@@ -233,6 +243,19 @@ export const currentHiddenCellIds = createSelector(
...
+  [currentCellMap, currentCellOrder],
+  (cellMap, cellOrder): Immutable.List<any> => {
+    if (!cellOrder || !cellMap) {
+      return Immutable.List();
+    }
+
+    return cellOrder.filter(
+      cellID => cellMap.getIn([cellID, "metadata", "outputHidden"]) === false
```
theengineear: I'm not sure on this one, but I _feel_ like this should be `=== true` based on the naming of these things.

PR #2572, opened 2018-02-22
Clean up some remaining selector usage

rgbkrk:
This switches out the remaining usages of `store.getState()` to use proper `selectors`.

I also didn't change tests that use `store.getState()`, opting instead to let it pick the path directly.

Note: I didn't touch the one use of the `githubToken`.

### DIFF
--- a/applications/desktop/src/main/menu.js
+++ b/applications/desktop/src/main/menu.js
@@ -214,6 +214,8 @@ export const named = {
 };
 
 export function loadFullMenu(store: * = global.store) {
+  // NOTE for those looking for selectors -- this state is not the same as the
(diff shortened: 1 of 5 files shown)


applications/desktop/src/main/menu.js
```diff
@@ -214,6 +214,8 @@ export const named = {
 };
 
 export function loadFullMenu(store: * = global.store) {
+  // NOTE for those looking for selectors -- this state is not the same as the
+  //      "core" state -- it's a main process side model in the electron app
```
theengineear, line 218: 👌 thanks for this!

theengineear (approved):
Excellent!

PR #2573, opened 2018-02-22
Fix flow types

rgbkrk:
This addresses some of the shortcomings of our current flow setup with `@nteract/core`. Ended up noticing a type mismatch as well, which I fixed by making `mapStateToProps` take a little less for options.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -4,15 +4,15 @@ types
 
 [options]
 suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
-module.name_mapper='^@nteract/core$' -> '<PROJECT_ROOT>/packages/core/src/'
+module.name_mapper='^@nteract/core$' -> '<PROJECT_ROOT>/packages/core/src'
 module.name_mapper='^@nteract/core/(?!src)(.+)$' -> '<PROJECT_ROOT>/packages/core/src/$1'
 module.name_mapper='^@nteract/core/src/(.+)$' -> '<PROJECT_ROOT>/packages/core/src/$1'
-module.name_mapper='^@nteract/(?!core)(.+)$' -> '<PROJECT_ROOT>/packages/$1/src/'
-module.name_mapper='^ansi-to-react$' -> '<PROJECT_ROOT>/packages/ansi-to-react/src/'
-module.name_mapper='^enchannel-zmq-backend$' -> '<PROJECT_ROOT>/packages/enchannel-zmq-backend/src/'
-module.name_mapper='^fs-observable$' -> '<PROJECT_ROOT>/packages/fs-observable/src/'
-module.name_mapper='^rx-binder$' -> '<PROJECT_ROOT>/packages/rx-binder/src/'
-module.name_mapper='^rx-jupyter$' -> '<PROJECT_ROOT>/packages/rx-jupyter/src/'
+module.name_mapper='^@nteract/(?!core)(.+)$' -> '<PROJECT_ROOT>/packages/$1/src'
+module.name_mapper='^ansi-to-react$' -> '<PROJECT_ROOT>/packages/ansi-to-react/src'
+module.name_mapper='^enchannel-zmq-backend$' -> '<PROJECT_ROOT>/packages/enchannel-zmq-backend/src'
+module.name_mapper='^fs-observable$' -> '<PROJECT_ROOT>/packages/fs-observable/src'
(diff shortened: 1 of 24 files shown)


.flowconfig
```diff
@@ -4,7 +4,9 @@ types
 
 [options]
 suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
-module.name_mapper='^@nteract/core/(.+)$' -> '<PROJECT_ROOT>/packages/core/src/$1'
+module.name_mapper='^@nteract/core$' -> '<PROJECT_ROOT>/packages/core/src/'
```
theengineear: Any reason for the trailing slash?

rgbkrk: Not really, lemme switch it to make it consistent

.flowconfig
```diff
@@ -4,7 +4,9 @@ types
...
+module.name_mapper='^@nteract/core/(?!src)(.+)$' -> '<PROJECT_ROOT>/packages/core/src/$1'
+module.name_mapper='^@nteract/core/src/(.+)$' -> '<PROJECT_ROOT>/packages/core/src/$1'
```
theengineear: We're going to eventually *not* allow `@nteract/core/*` imports, right? Like we're going to change this to be a defacto import from `@nteract/core`, ya?

.flowconfig
rgbkrk: oh, the rest of these down below have a trailing slash so I think I probably kept it consistent there.

.flowconfig
rgbkrk: Yeah. Figured I might as well push this to stay consistent for now at least. Once we've switched them, let's take the rule out.

applications/desktop/__tests__/renderer/epics/saving-spec.js
```diff
@@ -2,7 +2,7 @@ jest.mock("fs");
 import { ActionsObservable } from "redux-observable";
 import { dummyStore, dummyCommutable } from "@nteract/core/dummy";
 
-import { save, saveAs } from "@nteract/core/actions";
+import { actions } from "@nteract/core";
```
theengineear, line 5: 💯 this is much nicer, imo.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -52,33 +52,6 @@ import {
...
-  INTERRUPT_KERNEL,
-  LAUNCH_KERNEL_SUCCESSFUL,
-  LAUNCH_KERNEL,
-  LAUNCH_KERNEL_BY_NAME,
-  SET_LANGUAGE_INFO,
-  ERROR_KERNEL_LAUNCH_FAILED,
-  KERNEL_RAW_STDOUT,
-  KERNEL_RAW_STDERR
-} from "@nteract/core/actionTypes";
-
```
theengineear, line 81: 👋

applications/desktop/src/notebook/index.js
```diff
@@ -12,9 +12,8 @@ import NotificationSystem from "react-notification-system";
 
 import configureStore from "./store";
 
-import { NotebookApp, Styles } from "@nteract/core/providers";
-
-import { setNotificationSystem } from "@nteract/core/actions";
+import { actions, providers } from "@nteract/core";
+const { NotebookApp, Styles } = providers;
```
theengineear: Nice. So, the _only_ time I've seen this get hairy is when you have a sneaky cyclic import. I.e., it's possible to structure your code such that that `providers` is actually not yet defined here if there's some cyclic stuff going on, which will set NotebookApp and Styles to undefined for the life of this module. I don't _think_ we're going to run into an issue here, but it's something that I've seen with this pattern before and you just get a runtime error, which can be confusing.

theengineear: Otherwise, this is totally fine for now with the hope that we'll actually provide these as a top-level export later.

rgbkrk: Oh I could switch this to the `export * from 'providers'` setup in `@nteract/core` and just go from there.

applications/showcase/pages/index.js
```diff
@@ -196,7 +196,7 @@ export const Example = () => (
                   counter={cell.executionCount}
                 />
               )}
-              <Editor>
+              <Source>
```
theengineear, line 199: woo!

theengineear (approved):

PR #2578, opened 2018-02-22
migrate components to pure & connected

rgbkrk:
Moving everything from `providers` into `components`, bit by bit.

* [x] cell creator
* [x] toolbar
* [x] editor
* [x] notebook app

I ran out of laptop battery at the coffee shop so I'm going to skip `<Styles />` for now. Amusingly, it's the only actual provider. 😅


applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -10,7 +10,13 @@ import configureStore from "./store";
...
 import type { JupyterConfigData } from "./store";
 
-import { NotebookApp, Styles } from "@nteract/core/providers";
+import {
+  NotebookApp,
+  Styles,
+  ModalController,
+  NotebookMenu,
+  TitleBar
+} from "@nteract/core";
```
theengineear, line 19: 💯

packages/core/src/components/notebook-app.js
```diff
@@ -13,29 +13,26 @@ import {
...
 
 import { Display, RichestMime } from "@nteract/display-area";
 
-import CodeMirror from "./editor";
+import Editor from "./editor";
 
 import { displayOrder, transforms } from "@nteract/transforms";
 
-import { Input, Prompt, Source, Pagers, Outputs, Cell } from "../components";
+import { Input, Prompt, Source, Pagers, Outputs, Cell } from "./presentational";
```
theengineear, line 29: Cool. The `/components/presentational` works for me.

theengineear: The idea is that presentational components don't need connecting or anything right? They're like dumb components?

packages/core/src/providers/index.js
```diff
@@ -1,8 +1,4 @@
 // @flow
-import CellCreator from "./cell-creator";
-import Editor from "./editor";
-import NotebookApp from "./notebook-app";
 import Styles from "./styles";
-import Toolbar from "./toolbar";
 
-export { CellCreator, Editor, NotebookApp, Styles, Toolbar };
+export { Styles };
```
theengineear, line 4: 🐮 just `export { Styles } from "./styles";`

theengineear (approved):
Cool! Looks great! Sorry it took so long to look at. I'm going to merge this in and you can respond to comments at your leisure.

packages/core/src/components/notebook-app.js
rgbkrk: Yeah they're basically just styled divs that take certain kinds of children. Open to naming on these. I just moved them out of index.js so I wouldn't get a circular dependency.

packages/core/src/providers/index.js
rgbkrk: I'll make the next PR that just moves this into the components.

PR #2581, opened 2018-02-23
Refact more imports

theengineear:
Some more cleaning up! I haven't run this yet as a sanity check. I can do that in the am.

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.js
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.js
@@ -37,7 +37,7 @@ import { ipcRenderer as ipc } from "electron";
 import { createMainChannel } from "enchannel-zmq-backend";
 import * as jmp from "jmp";
 
-import type { NewKernelAction } from "@nteract/core/actionTypes";
+import type { NewKernelAction } from "@nteract/core/src/actionTypes";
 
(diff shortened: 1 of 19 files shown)


applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -296,19 +296,22 @@ export function watchSpawn(action$: *, store: *) {
   return action$.pipe(
     ofType(actionTypes.LAUNCH_KERNEL_SUCCESSFUL),
     switchMap((action: NewKernelAction) => {
+      // $FlowFixMe
       const spawn = action.kernel.spawn;
```
rgbkrk, line 300: These make sense since our kernel is nullable. It's only too bad that the rx types don't pass on through because then we could do a `filter` that ensured there was a kernel _and_ that it was a zeromq / local type.

I'll stick in some workarounds here at least.

rgbkrk:
Running this locally:

* [x] Desktop
* [x] Jupyter extension

rgbkrk:
Looking great! Thanks!

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: Ha, yah, I tried to filter it above to no avail :(.

PR #2583, opened 2018-02-23
switch to using already extracted notebookName

rgbkrk:
Noticed an unused suppression, ended up fixing this to use a previously selected value anyways.

### DIFF
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -348,11 +348,7 @@ export function storeToPDF(store: *) {
       }
     });
   } else {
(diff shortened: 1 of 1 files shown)


applications/desktop/src/notebook/menu.js
```diff
@@ -348,11 +348,7 @@ export function storeToPDF(store: *) {
       }
     });
   } else {
-    filename = path.join(
-      // $FlowFixMe: This should be using a selector first and foremost
```
theengineear, line 352: Ah, nice! 💥

theengineear (approved):

PR #2584, opened 2018-02-23
correct all action types

rgbkrk:
Got rid of the 35 flow errors that resulted from correcting `actionTypes` resolution.

### DIFF
--- a/packages/core/src/actionTypes.js
+++ b/packages/core/src/actionTypes.js
@@ -240,37 +240,37 @@ export type ExecuteFocusedCellAction = {
 };
 
 export const FOCUS_CELL = "FOCUS_CELL";
-export type FocusCellAction = { type: "FOCUS_CELL", id: CellID };
+export type FocusCellAction = { type: "FOCUS_CELL", id: ?CellID };
 
 export const FOCUS_NEXT_CELL = "FOCUS_NEXT_CELL";
 export type FocusNextCellAction = {
   type: "FOCUS_NEXT_CELL",
-  id: CellID,
+  id: ?CellID,
   createCellIfUndefined: boolean
 };
 
 export const FOCUS_PREVIOUS_CELL = "FOCUS_PREVIOUS_CELL";
 export type FocusPreviousCellAction = {
   type: "FOCUS_PREVIOUS_CELL",
-  id: CellID
+  id: ?CellID
 };
 
 export const FOCUS_CELL_EDITOR = "FOCUS_CELL_EDITOR";
 export type FocusCellEditorAction = {
   type: "FOCUS_CELL_EDITOR",
-  id: CellID | null
+  id: ?CellID
 };
 
 export const FOCUS_NEXT_CELL_EDITOR = "FOCUS_NEXT_CELL_EDITOR";
 export type FocusNextCellEditorAction = {
   type: "FOCUS_NEXT_CELL_EDITOR",
-  id: CellID
+  id: ?CellID
 };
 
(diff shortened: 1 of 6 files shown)


packages/core/src/actionTypes.js
```diff
@@ -326,10 +326,10 @@ export type CommMessageAction = {
 };
 
 export const SET_CONFIG_AT_KEY = "SET_CONFIG_AT_KEY";
-export type SetConfigAction = {
+export type SetConfigAction<T> = {
```
theengineear, line 329: So, this means that we are able to type `value` to whatever we want? I'll leave another comment where it's used.

packages/core/src/reducers/config.js
```diff
@@ -4,11 +4,11 @@ import { Map } from "immutable";
 
 import type { SetConfigAction, MergeConfigAction } from "../actionTypes";
 
-type ConfigAction = SetConfigAction | MergeConfigAction;
+type ConfigAction = SetConfigAction<*> | MergeConfigAction;
 
 type ConfigState = Map<any, any>;
 
-export function setConfigAtKey(state: ConfigState, action: SetConfigAction) {
+export function setConfigAtKey(state: ConfigState, action: SetConfigAction<*>) {
```
theengineear, line 11: I guess since this is set to `*`, i don't quite see what this is helping to assert? Not a disagreeing, just trying to understand.

packages/core/src/actionTypes.js
rgbkrk: Yeah, most of the time (maybe all the time) the generic type is string here.

packages/core/src/actions.js
```diff
@@ -769,7 +761,7 @@ export function deleteConnectionFileFailed(
   };
 }
 
-export function deleteConnectionFileSuccessful(): deleteConnectionFileSuccessful {
```
theengineear, line 772: huh, I'm surprised this wasn't an error before? What was `flow` thinking here to _not_ throw?

packages/core/src/reducers/config.js
rgbkrk: Every call to this function is in this file and their type is based on the config value they're setting and the action type they expect to return. Look for the calls to this as well as the other uses of `SetConfigAction<`

packages/core/src/actions.js
rgbkrk: LMAO, yeah this threw me completely off.

packages/core/src/components/notebook-app.js
```diff
@@ -298,11 +298,11 @@ type NotebookStateProps = {
 };
 
 type NotebookDispatchProps = {
-  moveCell: (sourceId: string, destinationId: string, above: boolean) => void,
-  selectCell: (id: string) => void,
-  executeFocusedCell: () => void,
-  focusNextCell: () => void,
-  focusNextCellEditor: () => void
+  moveCell: (sourceId: string, destinationId: string, above: boolean) => any,
```
theengineear: Should these be `*` since they _do_ have some type? By calling them `any`, flow is forced to give up trying to understand this type, right?

rgbkrk: Yeah, `*` would be better. I'll change these real quick.

theengineear (approved):
Looks good to me! Your flow skills far surpass mine, so if it's passing flow, +100. I left a couple comments, but I'll likely merge so I can do some work after it.

rgbkrk:
Thanks for the quick review! I changed the dispatch props as requested, I don't have time (right now that is) to update comments for the other ones. Feel free to add new commits to this one or new PRs, I'll be gone for ~6 hours.

PR #2587, opened 2018-02-24
Fixes #2117 - Add Title Elements to all Octicon SVGs

### LINKED ISSUE #2117: Provide a `<title>` to our toolbar buttons
For screen readers, all our SVG images that are used in toolbars (and other UI elements) should use [an SVG `<title>` element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) to describe what the button does to allow screen readers to know what the functionality is.

Example (for the "right triangle" svg):

```js
<svg xmlns="http://www.w3.org/2000/svg" width="6" height="16" viewBox="0 0 6 16">
  <title>Run Cell</title>
  <path fillRule="evenodd" d="M0 14l6-6-6-6z" />
</svg>
```

hydrosquall:
# Changelog

- Adds `<title>` tags to the `octicons` package, using guidelines from the `classNames` and `title` properties around where each of these icons are used elsewhere in the codebase.
- Update Jest snapshots to expect title tags in these icons

Other things I noticed

- The `DownArrow` icon (used to merge cells) turns out to be an upwards-pointing arrow when viewed on my computer. Possible bug.
- Couldn't actually find usage of the `CloudDownload` icon in the codebase, it's possibly dead code. However, it's labeled just in case.
- Couldn't figure out where to find the `Beaker` icon could actually be called within the app, it appears to pop up as part of a warning message. Labeled it  just in case, but could remove this if necessary.

As this is my first PR to this project, please feel free to let me know if there are any other documentation/styling/testing things that need to be taken care of. 

@rgbkrk

### DIFF
--- a/packages/core/__tests__/components/__snapshots__/pinned-cell-spec.js.snap
+++ b/packages/core/__tests__/components/__snapshots__/pinned-cell-spec.js.snap
@@ -21,6 +21,9 @@ exports[`PinnedPlaceHolderCell has a snapshot test at least 1`] = `
         width={12}
         xmlns="http://www.w3.org/2000/svg"
       >
+        <title>
+          Cell Placeholder for Pinned Cell
+        </title>
         <path
           d="M11 10h1v3c0 .55-.45 1-1 1H1c-.55 0-1-.45-1-1V3c0-.55.45-1 1-1h3v1H1v10h10v-3zM6 2l2.25 2.25L5 7.5 6.5 9l3.25-3.25L12 8V2H6z"
           fillRule="evenodd"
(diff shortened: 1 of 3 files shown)


rgbkrk:
Thank you! Extra thank you for the detailed description.

The cloud download icon gets used by the commuter front end package (or at least is supposed to) for downloading notebooks. The beaker is not used in an interactive way, though it does get used in the data resource transform package. Between the time I wrote that issue and now quite a bit changed. We might want to just pass title as a prop and set the intent in the apps in the future.

rgbkrk (approved):

rgbkrk:
Welcome to nteract, happy to have you!

PR #2588, opened 2018-02-25
Refact records

theengineear:
Woo! After some of the recent refacts, this change became much easier. I nested `records` at `state/old`.

I think we still need to think about naming and what we want to expose as a top-level export. However, we're getting a lot closer to the organization we're ultimately looking for in `/core`.


theengineear:
@rgbkrk A little overdue, but here you go 😅.

applications/desktop/__tests__/renderer/native-window-spec.js
```diff
@@ -4,7 +4,7 @@ import { remote } from "electron";
 import { from } from "rxjs/observable/from";
 
 import * as nativeWindow from "../../src/notebook/native-window";
-import { makeAppRecord, makeDocumentRecord } from "@nteract/core/records";
+import { state as stateModule } from "@nteract/core";
```
rgbkrk, line 7: It would be really tempting to call the state module "core".

rgbkrk (approved):
Looking super great here!

applications/desktop/__tests__/renderer/native-window-spec.js
theengineear: Heh, yah that's true. We should figure out the right names for all this stuff soon. I like that all this is under `/state` though, it makes it really easy to manage in the repo.

theengineear:
Thanks for the quick review!

applications/desktop/__tests__/renderer/native-window-spec.js
rgbkrk: Yeah my only worry is the super common use of the variable named state.

theengineear: Yup, definitely worthwhile to not block that variable name 😆.

Issue #2513, opened 2018-02-09
Alphabetize runtime menu entries


jkornblum:
Maybe there is a sort order to this? Not sure.
It doesn't seem to be ordered by kernel name or display name for me.
I'd prefer them to ordered by display name, what are your thoughts?

![runtime-menu](https://user-images.githubusercontent.com/5368903/36007946-776bfa58-0d0a-11e8-9302-d618e58e8ee3.png)

rgbkrk:
Sorting by display name would be cool.

jkornblum:
Closed with #2591

PR #2591, opened 2018-02-25
sort kernel by displayname on menu

jkornblum:
Will sort the display order of kernelspecs on the menu by the kernel display name.

*Note: I couldn't switch kernels on my branch when writing this, however I couldn't switch kernels on the master branch either so I don't think it has to do with any of these code changes.

addresses #2513

### DIFF
--- a/applications/desktop/src/main/menu.js
+++ b/applications/desktop/src/main/menu.js
@@ -8,6 +8,7 @@ import {
   BrowserWindow
 } from "electron";
 import * as path from "path";
+import { sortBy } from "lodash";
 import { launch, launchNewNotebook } from "./launch";
 import { installShellCommand } from "./cli";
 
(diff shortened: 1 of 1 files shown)


applications/desktop/src/main/menu.js
```diff
@@ -219,14 +220,16 @@ export function loadFullMenu(store: * = global.store) {
...
-      click: createSender("menu:new-kernel", kernelSpecs[kernelSpecName])
+      label: kernel.spec.display_name,
+      click: createSender("menu:new-kernel", kernel)
     };
   }
 
-  const kernelMenuItems = Object.keys(kernelSpecs).map(generateSubMenu);
+  const kernelMenuItems = sortBy(kernelSpecs, "spec.display_name").map(
+    generateSubMenu
+  );
```
rgbkrk, line 232: 👍

rgbkrk:
I'll check the `master` branch issue out, thanks for sorting these @jkornblum

PR #2593, opened 2018-02-26
create webpack configurator library

rgbkrk:
Tinier version of #2554, just starts off the package and exports some of the basics for use (aliases, alias merging, and the webpack configure function).


packages/webpack-configurator/index.js
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    if (
+      rule.loader === "babel-loader" ||
+      (rule.use && rule.use.loader === "babel-loader")
+    ) {
+      hasBabelLoader = true;
+    }
+
+    if (String(rule.exclude) === String(/node_modules/)) {
```
theengineear, line 103: Are there any other iterations of `/node_modules/` we'd want to guard against? `/node_modules`?  This is probably a naive comment, but it feels like this might be overly restrictive.

theengineear (approved):

packages/webpack-configurator/index.js
rgbkrk: Could be. I just matched the regex string patterns used by our own config and next.js'.

rgbkrk:
Thanks! I'll also get this updated with the webpack 4 upgrade. 😄

PR #2595, opened 2018-02-27
make flow and react imports consistent

rgbkrk:
Even more touchups from #2554, complete with a fix for an error handler. 😊


theengineear (approved):
Awesome, :tada: for consistency :)

PR #2601, opened 2018-02-27
Prepare state

theengineear:
This prepares the state to add kernel refs all over the place. Just some low-hanging fruit that can be pulled into a preliminary PR.


rgbkrk:
All these old prefixes will make me super ready for the next stage.

PR #2606, opened 2018-02-28
Flow fixme athon

rgbkrk:
Going on a bit of a rampage of fixes for flow types here to assist with the kernel refs setup.


theengineear (approved):
Woo 🎉

PR #2611, opened 2018-02-28
little bits of todo cleanup

rgbkrk:
Since I know this affects things with our flow setup, I'm removing these defaults for kill and restart.

### DIFF
--- a/packages/core/src/actions.js
+++ b/packages/core/src/actions.js
@@ -492,10 +492,10 @@ export function deleteMetadata(field: string): DeleteMetadataFieldAction {
   };
 }
 
-// TODO: we should probably remove the default here.
-export function killKernel(
-  payload: { restarting: boolean, ref?: KernelRef } = { restarting: false }
-): KillKernelAction {
+export function killKernel(payload: {
+  restarting: boolean,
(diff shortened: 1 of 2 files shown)


packages/core/src/actions.js
```diff
@@ -492,10 +492,10 @@ export function deleteMetadata(field: string): DeleteMetadataFieldAction {
   };
 }
 
-// TODO: we should probably remove the default here.
-export function killKernel(
-  payload: { restarting: boolean, ref?: KernelRef } = { restarting: false }
```
theengineear, line 497: 🙇 thanks!

packages/core/src/actions.js
```diff
@@ -847,10 +847,10 @@ export function shutdownReplyTimedOut(payload: {
   };
 }
 
-// TODO: probably should remove default here.
-export function restartKernel(
-  payload: { clearOutputs: boolean, ref?: KernelRef } = { clearOutputs: false }
```
theengineear, line 852: 🥇

theengineear (approved):
woo! yah, ditching defaults is definitely helpful here!

PR #2612, opened 2018-02-28
map names of modules for flow properly, fix resulting errors

rgbkrk:
😭 

Our flow config for the module name mapper was wrong this entire time. Regex groups needed to be inside escaped parens and the matcher is `\1`. Good news is that our monorepo is resolving properly now. Bad news is the pile of flow cleanup here. I ended up leaving some `$FlowFixMe` in while also addressing a large portion of them.


theengineear (approved):

theengineear:
^^ Man, the difference in syntax in the flowconfig is brutal 🐼.

PR #2613, opened 2018-02-28
Jest mapper

rgbkrk:
Set up all the proper jest name mappers to resolve from `src`


rgbkrk:
I'm going to assume this is 👌 , and I'm merging it in to keep going.

PR #2614, opened 2018-02-28
Use kernel refs

theengineear:
💥 💣 🎉 👏 Jupyter Extension is using kernels by ref now.

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.js
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.js
@@ -148,7 +148,8 @@ export const launchKernelByNameEpic = (
             actions.launchKernel({
               kernelSpec: specs[action.payload.kernelSpecName],
               cwd: action.payload.cwd,
-              ref: action.payload.ref
+              ref: action.payload.ref,
+              selectNextKernel: action.payload.selectNextKernel
             })
           )
         )
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.js
@@ -27,8 +27,14 @@ function createApp(jupyterConfigData: JupyterConfigData) {
   class App extends React.Component<*> {
     notificationSystem: NotificationSystem;
 
+    componentWillMount(): void {
+      // TODO: See reducers/core/index `useCore` reducer.
+      store.dispatch(actions.useCore());
+    }
+
     componentDidMount(): void {
       const hostRef = state.createHostRef();
+      const kernelRef = state.createKernelRef();
       const kernelspecsRef = state.createKernelspecsRef();
       store.dispatch(
         actions.addHost({
@@ -47,11 +53,11 @@ function createApp(jupyterConfigData: JupyterConfigData) {
       );
 
       // TODO: we should likely be passing in a hostRef to fetchContent too.
-      // TODO: provide a KernelRef
       store.dispatch(
         actions.fetchContent({
           path: jupyterConfigData.contentsPath,
-          params: {}
+          params: {},
+          kernelRef
         })
       );
       store.dispatch(actions.fetchKernelspecs({ hostRef, kernelspecsRef }));
--- a/packages/core/__tests__/epics/execute-spec.js
+++ b/packages/core/__tests__/epics/execute-spec.js
@@ -1,7 +1,7 @@
 // @flow
 import { ActionsObservable } from "redux-observable";
 
-import { actionTypes, actions } from "@nteract/core";
+import { actionTypes, actions, state as stateModule } from "@nteract/core";
 
 import { createExecuteRequest } from "@nteract/messaging";
 
@@ -49,11 +49,21 @@ describe("createExecuteCellStream", () => {
         return this.state;
       },
       state: {
+        core: stateModule.makeStateRecord({
+          useCore: true,
+          kernelRef: "fake",
+          entities: stateModule.makeEntitiesRecord({
+            kernels: stateModule.makeKernelsRecord({
+              byRef: Immutable.Map({
+                fake: stateModule.makeRemoteKernelRecord({
+                  channels,
+                  status: "not connected"
+                })
+              })
+            })
+          })
+        }),
         app: {
-          kernel: {
-            channels,
-            status: "not connected"
-          },
           notificationSystem: { addNotification: jest.fn() }
         },
         document: Immutable.fromJS({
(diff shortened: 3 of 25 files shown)


applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -148,7 +148,8 @@ export const launchKernelByNameEpic = (
             actions.launchKernel({
               kernelSpec: specs[action.payload.kernelSpecName],
               cwd: action.payload.cwd,
-              ref: action.payload.ref
+              ref: action.payload.ref,
+              selectNextKernel: action.payload.selectNextKernel
```
rgbkrk, line 152: Excellent, I figured there'd be some passthrough here.

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -27,8 +27,14 @@ function createApp(jupyterConfigData: JupyterConfigData) {
   class App extends React.Component<*> {
     notificationSystem: NotificationSystem;
 
+    componentWillMount(): void {
+      // TODO: See reducers/core/index `useCore` reducer.
+      store.dispatch(actions.useCore());
```
rgbkrk, line 32: 💭

packages/core/__tests__/epics/execute-spec.js
```diff
@@ -49,11 +49,21 @@ describe("createExecuteCellStream", () => {
         return this.state;
       },
       state: {
+        core: stateModule.makeStateRecord({
+          useCore: true,
```
rgbkrk, line 53: I look forward to when `useCore` is gone. 😉

packages/core/__tests__/epics/execute-spec.js
```diff
@@ -198,11 +218,21 @@ describe("executeCellEpic", () => {
...
+            kernels: stateModule.makeKernelsRecord({
+              byRef: Immutable.Map({
+                fake: stateModule.makeRemoteKernelRecord({
+                  channels: null,
+                  status: "not connected"
+                })
+              })
+            })
+          })
+        }),
```
rgbkrk, line 234: Very happy you cleaned these up to use the proper records

packages/core/src/actionTypes.js
```diff
@@ -511,7 +515,19 @@ export type LaunchKernelAction = {
...
+};
+
+export const LAUNCH_KERNEL_BY_NAME = "LAUNCH_KERNEL_BY_NAME";
+export type LaunchKernelByNameAction = {
+  type: "LAUNCH_KERNEL_BY_NAME",
+  payload: {
+    kernelSpecName: string,
+    cwd: string,
+    ref?: KernelRef,
+    selectNextKernel: boolean
```
rgbkrk, line 530: 👍

packages/core/src/components/notebook-menu/index.js
```diff
@@ -20,6 +21,7 @@ type Props = {
   defaultOpenKeys?: Array<string>,
   openKeys?: Array<string>,
   cellFocused: ?string,
+  currentKernelRef: ?KernelRef,
```
rgbkrk, line 24: 👍

packages/core/src/components/notebook-menu/index.js
```diff
@@ -48,6 +50,7 @@ class PureNotebookMenu extends React.Component<Props> {
   static defaultProps = {
     cellFocused: null,
     saveNotebook: null,
+    currentKernelRef: null,
```
rgbkrk, line 53: 👍

packages/core/src/components/notebook-menu/index.js
```diff
@@ -185,25 +189,22 @@ class PureNotebookMenu extends React.Component<Props> {
...
+          restartKernel({ ref: currentKernelRef });
         }
         break;
       case MENU_ITEM_ACTIONS.KILL_KERNEL:
         if (killKernel) {
-          // TODO: provide KernelRef
-          killKernel({});
+          killKernel({ ref: currentKernelRef });
         }
+        break;
```
rgbkrk, line 204: 😊

packages/core/src/dummy/index.js
```diff
@@ -86,6 +90,8 @@ export function dummyStore(config: *) {
   const dummyNotebook = buildDummyNotebook(config);
 
   return createStore(rootReducer, {
+    core: makeStateRecord(),
+    modals: makeModalsRecord(),
```
rgbkrk, line 94: Hooray for a fixed dummy store!

packages/core/src/reducers/core/communication/kernels.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+import {
+  makeKernelCommunicationRecord,
+  makeKernelsCommunicationRecord
+} from "../../../state/communication/kernels";
+import * as actionTypes from "../../../actionTypes";
+import { combineReducers } from "redux-immutable";
+import * as Immutable from "immutable";
+
+// TODO: we should spec out a way to watch the killKernel lifecycle.
```
rgbkrk, line 10: Maybe we need to set the status in the kernel object when we see a kill kernel for a given ref.

packages/core/src/reducers/core/entities/kernels.js
```diff
@@ -0,0 +1,58 @@
...
+  makeRemoteKernelRecord,
+  makeKernelsRecord
+} from "../../../state/entities/kernels";
+import * as actionTypes from "../../../actionTypes";
+import { combineReducers } from "redux-immutable";
+import * as Immutable from "immutable";
+
+// TODO: we need to clean up references to old kernels at some point. Listening
+// for KILL_KERNEL_SUCCESSFUL seems like a good candidate, but I think you can
+// also end up with a dead kernel if that fails and you hit KILL_KERNEL_FAILED.
```
rgbkrk, line 13: True true.

packages/core/src/state/entities/kernels.js
```diff
@@ -0,0 +1,76 @@
+// @flow
+import * as Immutable from "immutable";
+import type { ChildProcess } from "child_process";
+import type { HostRef, KernelRef } from "../refs";
+import type { KernelId } from "../ids";
+
+export type BaseKernelProps = {
```
rgbkrk, line 7: It scares me a bit that these are copied over from `old` when they're exactly the same. We should make one rely on the other or just get things aligned.

rgbkrk (approved):
I think the only thing that scares me in this PR is that types are copied from old in `packages/core/src/state/entities/kernels.js` when they should use the same type.

theengineear:
>I think the only thing that scares me in this PR is that types are copied from old in packages/core/src/state/entities/kernels.js when they should use the same type.

Yah, this was honestly just because I was not really trusting flow at the time and didn't want to make more waves than necessary. We should be able to sync them though in the follow up.

packages/core/src/reducers/core/communication/kernels.js
theengineear: ^^ That's a lovely idea ❤️

rgbkrk:
> Yah, this was honestly just because I was not really trusting flow at the time and didn't want to make more waves than necessary. We should be able to sync them though in the follow up.

Works for me, I'm happy to keep this all moving forward.

packages/core/src/state/entities/kernels.js
rgbkrk: On further digging, these came from the previous kernel refs PR (that we closed) and they're not consistent with the ones used in old currently.

PR #2617, opened 2018-03-01
Merge old and new types

theengineear:
This finishes merging the old/core stuff back together. I.e., I cleaned up my mess.


applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -41,7 +41,7 @@ import * as jmp from "jmp";
 
 import type { NewKernelAction } from "@nteract/core/src/actionTypes";
 
-import type { OldKernelInfo, LocalKernelProps } from "@nteract/core/src/state";
+import type { KernelInfo, LocalKernelProps } from "@nteract/core/src/state";
```
rgbkrk, line 44: 👍

applications/desktop/src/notebook/index.js
```diff
@@ -24,16 +24,15 @@ import { state } from "@nteract/core";
 
 const store = configureStore({
   app: state.makeAppRecord({
-    host: state.makeOldDesktopHostRecord(),
+    host: state.makeLocalHostRecord(),
```
rgbkrk, line 27: 😄

packages/core/src/reducers/index.js
```diff
@@ -3,7 +3,6 @@ import document from "./document";
 import comms from "./comms";
 import config from "./config";
 import app from "./app";
-import modals from "./modals";
 import core from "./core";
```
rgbkrk, line 7: Where are `modals` now?

rgbkrk: Ah in entities along with hosts! Hurrah!

rgbkrk (approved):
As the saying goes -- out with the old, in with the core.

🧘

packages/core/src/reducers/index.js
theengineear: Inside `core` so `core.modals`. ✨One less top-level reducer ✨

PR #2620, opened 2018-03-02
Prep for content refs

theengineear:
A couple more things I can factor out of my gargantuan PR 😁.

I'm getting flow errors locally on this... not sure why tbh!

### DIFF
--- a/packages/core/src/state/entities/contents.js
+++ b/packages/core/src/state/entities/contents.js
@@ -1,3 +1,4 @@
+// @flow
 import * as Immutable from "immutable";
 import { emptyNotebook } from "@nteract/commutable";
 
--- a/packages/core/src/state/entities/contents.js
+++ b/packages/core/src/state/entities/contents.js
@@ -1,0 +1,36 @@
+import * as Immutable from "immutable";
+import { emptyNotebook } from "@nteract/commutable";
+
+export type DocumentRecordProps = {
+  // TODO: This _needs_ to become a Record
+  notebook: Immutable.Map<string, any>,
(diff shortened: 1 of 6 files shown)


packages/core/src/state/entities/index.js
```diff
@@ -9,6 +9,7 @@ import { makeKernelsRecord } from "./kernels";
 import { makeKernelspecsRecord } from "./kernelspecs";
 import { makeModalsRecord } from "./modals";
 
+export * from "./contents";
```
theengineear, line 12: I don't get it. I'm exporting all-the-things from `contents` and all-the-things are getting re-exported from `state/index.js` in this file... flow fails, but tests pass...

theengineear:
^^ Not sure if you have light to shed on this flow error? Am I missing something dumb?

packages/core/src/state/entities/contents.js
```diff
@@ -0,0 +1,36 @@
+import * as Immutable from "immutable";
```
rgbkrk, line 2: This file is missing an `@flow` at the top.

rgbkrk: That was all that was missing here. Everything now passes!

rgbkrk:
Now for me to actually review. 😄

rgbkrk:
Looking good. This seems like a reasonable transition point.

packages/core/src/state/entities/contents.js
theengineear: face << palm. how many times...

theengineear:
Apparently I was tired last night 😛

PR #2616, opened 2018-03-01
enforce kernel ref usage

rgbkrk:
This furthers the kernel refs setup to enforce using kernel refs _everywhere_ they were optional before.

PR #2621, opened 2018-03-02
Extra cleanup

rgbkrk:
Some follow ups to comments on #2616.


rgbkrk:
I'm going to err on the side of A-OK here and merge this since it's really minor.

PR #2622, opened 2018-03-02
configure nteract showcase with our webpack config

rgbkrk:
Making it super easy to edit nteract components and work with them standalone in the showcase app. No more build step phooey, just run:

```
npm run app:showcase
```

Then load, for example, http://127.0.0.1:3000/dataresource and edit the `packages/transform-dataresource` component. :tada:

-----

Prepped this to make it so @emeeks can have a nice time hacking on the data resource transform.


rgbkrk:
live

PR #2623, opened 2018-03-02
Notebook Selection Menu

rgbkrk:
This creates a little "banner" that allows for clicking icons based on kernelspecs to launch new notebooks.

click-mah-notebook

* [x] Created `<NewNotebookNavigation />` component within core
* [x] Relied on `<button />`s to make accessibility a little bit simpler (and by extension some of the CSS and JS)
* [x] Created a new page on showcase for hacking on this component
* [x] Wrote a basic snapshot test

I'm not going to integrate this into the jupyter extension until our content refs are ready. This only starts setting up some components for use.

TODOs I left:

* Fallback to a simple logo when the kernel language isn't in our builtins
* Create a new way for kernels to specify an SVG for their logo
  * Most likely, we'd also want some way to note if a kernel is Spark backed or has other interesting features when there are > 2 kernels for that language. I'm probably going to start it as an informal spec based on `metadata` in the kernelspec and slap a spark logo on top. 😉


mpacer:
> Since showcase is now really nice for prototyping components, I've started it in there. I'll move it into core once ready.

I can put this into a new issue…, but I think it would help people begin working on the project if there were a tutorial showing a full walkthrough of how to use showcase for working on this kind of prototype.

Currently there is one design file in our `doc/` folder, I'm not even sure where that tutorial file would go…

rgbkrk:
A walkthrough on how to use showcase to prototype (it's next.js underneath with our custom webpack setup) or how I literally built this component?

mpacer:
Sort of… both? I was thinking that this would be a good example of a component that I wouldn't have had a good idea of how to properly prototype inside showcase. 

I had thought of showcase as a way to display the notebook components in a standard structure — it makes sense for it to be used more generally, I just hadn't thought about that in that way. 

In any case, even if it's not this component exactly a complete walkthrough of how you prototyped an entire actual component (i.e., _not_ a toy example but a component that was actually built for use in the nteract repository) would be super helpful.

rgbkrk:
> Sort of… both? I was thinking that this would be a good example of a component that I wouldn't have had a good idea of how to properly prototype inside showcase.

Yeah that's simple enough to kick off. I'll add on to showcase's README.

> I had thought of showcase as a way to display the notebook components in a standard structure — it makes sense for it to be used more generally, I just hadn't thought about that in that way.

The entire purpose of `showcase` is to display and develop on small UI components, similar to [storybook](https://storybook.js.org/).

> In any case, even if it's not this component exactly a complete walkthrough of how you prototyped an entire actual component (i.e., not a toy example but a component that was actually built for use in the nteract repository) would be super helpful.

At some point, I could imagine live streaming something like this. I don't know how I'd put together a decent tutorial on this because most of it is:

* [Mock a design](https://github.com/nteract/design/blob/master/nteract-web/hierarchical%20nteract%20Dashboard.png) (inspired by Google docs template selection)
* Make boxes with divs, color them various grays
* Search the web for each of these (in order, from my actual search history):
  * flex row (Oooh, I want to line these up)
  * flex row scroll (Oh, I just want only these notebooks to scroll when page is small)
  * flex space between items (I want some space between them)
  * only scroll div, not page (Why is the entire page scrolling while scrolling these little notebook icons)
  * no overscroll chrome (how do you get rid of that annoying bounce?)
  * margin width and scrolling (ok, I think it might be the margins)
  * vw css (I wonder if I should size according to view width)
  * padding overflow (ok, maybe it's padding)
  * Note: It was totally the padding making the "width" of the notebook selection area bigger than 100%
* Among each of those terms, I was iterating to make the design I wanted

As with all learning and exploration, there's a wide amount of knowledge of CSS that I both don't have and do have. There's also a lot of content out there on how to build particular site experiences. My favorite of the sites I found on the journey above was https://iamsteve.me/blog/entry/using-flexbox-for-horizontal-scrolling-navigation (and as always, both MDN and CSS Tricks https://css-tricks.com/fun-viewport-units/).

rgbkrk:
Turned this into a proper component into core, now to see if I can make each of these notebooks `<button />`s while retaining the style I was aiming for.

rgbkrk:
Now they're buttons and I'm pretty happy for the moment. 😄

rgbkrk:
Trying something a little fun on hover and focus:

color-switcher

rgbkrk:
I think I should switch this from a `<button />` to an `<a />` so that it's ctrl/cmd-clickable to allow the user to open in new tab and to eliminate weirdness around button focus for an element like this. 🤔

rgbkrk:
Whoops, just noticed I alternated the order of yellow and blue for the Python logo. Here it is fixed:

screen shot 2018-03-05 at 9 46 24 am

applications/showcase/pages/navigation.js
```diff
@@ -0,0 +1,98 @@
...
+            overflow: hidden;
+            height: 100%;
+          }
+
+          body {
+            overflow: auto;
+            height: 100%;
+            width: 100%;
+            margin: 0;
+            font-family: var(--nt-font-family-normal);
```
theengineear, line 90: 👌

packages/core/src/components/new-notebook-navigation/index.js
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * Create a listing of running notebooks
+ * Showcase the directory navigator
+ */
+
+import * as React from "react";
+
+import type { KernelspecProps } from "../../state/entities/kernelspecs";
+
+import { List, Map, Record } from "immutable";
```
theengineear, line 17: 🐄 `import * ...` style is our standard Immutable import, right?

packages/core/src/components/new-notebook-navigation/index.js
```diff
@@ -0,0 +1,161 @@
...
+
+import * as logos from "./logos";
+
+export type AvailableNotebook = {
+  kernelspec: KernelspecProps
```
theengineear, line 22: 🥇 nice!

theengineear (approved):
Awesome! Lookin' good! We might want to flesh out how this stuff will change under a `dark` theme, but we can fuss with that down the road. Those SVGs are slick!

packages/core/src/components/new-notebook-navigation/index.js
rgbkrk: Yes! Oops! I'll make another PR.

rgbkrk:
> We might want to flesh out how this stuff will change under a dark theme, but we can fuss with that down the road.

Yeah, that would be good to do. I aimed for matching my original mockup for better or worse, relying on the palette but not the theme due to not having everything available. Rather than get stuck in analysis paralysis, I went with getting this up as a prototype (albeit with tests and examples 😄 ). Thanks for merging it as it is!

PR #2626, opened 2018-03-03
Allow content ref in actions

theengineear:
This pr adds the *option* to add `contentRef` to a bunch of action payloads.


theengineear:
@rgbkrk slowly, but surely. I'm literally going action by action so that I can just keep pushing things up here. Not done yet, but going steadily forth.

rgbkrk:
We could always merge these part way ones, especially since you're doing a lot of moving things into `payload` objects.

rgbkrk:
Cool, let's move this forward.

PR #2628, opened 2018-03-05
Kernel communication

rgbkrk:
* improve messaging about kernel state
  * when it's the underlying process that has exited, denote it in the status
  * set the status when `KILL_KERNEL_(SUCCESSFUL|FAILED)`
* 🐛 the zeromq launch kernel epic was creating a brand new `kernelRef` instead of the one already specified


theengineear (approved):
Looks great!

PR #2642, opened 2018-03-07
Finish allowing content refs

theengineear:
OK. This step is done. All the actions that need content refs:

1. Allow them according to flow types
2. Use `action.payload` to nest the action information.

... I'm also sneaking the handling of contentRefs in epics here

- [x] make all content-y actions take optional `contentRef?: ContentRef`
- [x] pass contentRefs through all epics


rgbkrk (approved):
Great!

PR #2646, opened 2018-03-07
log stdout/err streams from raw kernel process

rgbkrk:
A little collapsible logger inside the console dev tools:

screen shot 2018-03-07 at 9 55 20 am

/cc @Jaykul for review

I also had to make the kernelRefs have a [subtyping constraint](https://flow.org/en/docs/types/opaque-types/#toc-subtyping-constraints) to allow them to be coerced to strings (for logging).

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.js
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.js
@@ -11,6 +11,8 @@ import { empty } from "rxjs/observable/empty";
 import { fromEvent } from "rxjs/observable/fromEvent";
 import { merge } from "rxjs/observable/merge";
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
Probably a more practical example, using pyspark since it launches a JVM process underneath:

screen shot 2018-03-07 at 10 00 18 am

I look forward to when we have a little kernel logs drawer instead. 😄

rgbkrk:
To make it easier to discern the kernels and to not have all these kind of useless collapsers (since you can't bring it all together from separate logging "sessions"), I switched to logging with colors and alternating colors if you restart or launch a new kernel:

screen shot 2018-03-07 at 11 13 39 am

PR #2654, opened 2018-03-08
ran npm t -- -u to update snapshots

rgbkrk:
I ran a fresh build with updated deps (flow is likely the culprit since babel + flow work together on the tests).


rgbkrk:
I'm really curious if things pass after this, it was a bit odd to have `master` locally failing. 🤔

rgbkrk:
Not sure what's up with my local copy here. 
¯\\\_(ツ)\_/¯

Closing.

PR #2658, opened 2018-03-09
Remove Sticky Cells

theengineear:
Sticky cells are a little buggy and I don't think our users are really using them (according to a poll, right?).

Also, they were jammed in _outside_ the document model. I'm not opposed to the idea in general, but I think we should perhaps punt this issue down the road until we can have a full discussion on the appropriate way to include them in our state model.


packages/core/src/components/pinned-cell.js
```diff
@@ -1,91 +0,0 @@
...
-          ref={ref => {
-            this.stickyCellContainer = ref;
-          }}
-        >
-          {this.props.children}
-        </div>
-        <style jsx>{`
-          .sticky-cell-container {
-            background: var(--theme-app-bg);
-            border-bottom: dashed var(--theme-app-border) 1px;
```
rgbkrk, line 64: I'll miss this most of all, as I tried to make some pretty sweet CSS for this feature I wasn't enjoying supporting.

rgbkrk (approved):

rgbkrk:
pinned-cell

rgbkrk:
Note for users who stumble upon this -- we'd love to support this kind of feature with some rethinking on how it's implemented. More than once it's been a stumbling block to maintaining the overall app. 🙇  Thank you.

PR #2659, opened 2018-03-09
Close Menu On Click

theengineear:
A bit of a workaround to get this to happen: https://github.com/react-component/menu/issues/89#issuecomment-321672963

Note that I think `activeKeys` is now `openKeys`?

### DIFF
--- a/packages/core/__tests__/components/notebook-menu-spec.js
+++ b/packages/core/__tests__/components/notebook-menu-spec.js
@@ -33,6 +33,9 @@ describe("PureNotebookMenu ", () => {
     });
     test("calls appropriate handlers on click", () => {
       const props = {
+        // keep menu open after clicks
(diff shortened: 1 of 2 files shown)


rgbkrk (approved):
Seems like an appropriate tradeoff. A neat property here is that I think the `openKeys` can be used to highlight clicking a menu entry to prompt a user to do something.

rgbkrk:
Very happy to have menu closing after clicking entries.

What happens if you click on one of the sub menus -- like change cell type. Does it close those when they should remain open?

theengineear:
>What happens if you click on one of the sub menus -- like change cell type. Does it close those?

Nope, just on item click, those are "submenus". So that click will not close the menu, which I think is the right behavior.

rgbkrk:
Ok sweet.

PR #2665, opened 2018-03-11
Highlighter cleanup

rgbkrk:
The current style of the highlighter was not matching up well with documents seen in the wild with commuter. I opted to switch to the `vs` theme here, relying on prism:

screen shot 2018-03-11 at 1 01 59 pm


rgbkrk:
Master:

screen shot 2018-03-11 at 1 05 45 pm

This branch:

screen shot 2018-03-11 at 1 06 13 pm

rgbkrk:
I've also got an alternate way of doing this, but what I might do is just switch commuter to using codemirror so it matches exactly. I'll leave the `core` module for highlighting to be without codemirror as requested by @tlyon3 originally.

theengineear:
Thanks for including the images 👌. Looks much better.

PR #2669, opened 2018-03-12
simplify initial boot

rgbkrk:
Create the initial state all in one rather than spreading it out.


rgbkrk:
Great, all the apps are failing because of vega's use of `util`. 😭

rgbkrk:
I'm going to aim at merging this and fixing the build for `master` in another PR.

PR #2671, opened 2018-03-12
Initialize state

rgbkrk:
Initializes more of the state on startup instead of relying on dispatching it first.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.js
@@ -111,46 +111,49 @@ function main(rootEl: Element, dataEl: Node | null) {
     return;
   }
 
+  const jupyterHostRecord = state.makeJupyterHostRecord({
+    id: null,
+    type: "jupyter",
+    defaultKernelName: "python",
+    token: config.token,
+    serverUrl: location.origin + config.baseUrl
+  });
+
+  const hostRef = state.createHostRef();
+  const contentRef = state.createContentRef();
+
   const initialState = {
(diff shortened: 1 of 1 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -111,46 +111,52 @@ function main(rootEl: Element, dataEl: Node | null) {
...
+      // TODO: Move into core as a "current" host
+      host: jupyterHostRecord
     }),
     comms: state.makeCommsRecord(),
     config: Immutable.Map({
       theme: "light"
     }),
-    core: state.makeStateRecord()
+    core: state.makeStateRecord({
+      // TODO: non-prioritized -- set the currentHostRef here
```
theengineear, line 136: OK for now. Thanks for dropping a note.

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -124,13 +124,16 @@ function main(rootEl: Element, dataEl: Node | null) {
 
   const initialState = {
     app: state.makeAppRecord({
-      version: `nteract-on-jupyter@${config.appVersion}`
+      version: `nteract-on-jupyter@${config.appVersion}`,
+      // TODO: Move into core as a "current" host
```
theengineear: Seems like this wouldn't be too hard. But no need to do it now.

theengineear (approved):
As an intermediate step. I don't see any issues with this approach. Until we need to handle two refs, this seems A-OK.

PR #2672, opened 2018-03-12
create a dummy content record

rgbkrk:
This is for use when you know the path of content (and are likely loading it)
but not the real content yet.

### DIFF
--- a/packages/core/src/state/entities/contents.js
+++ b/packages/core/src/state/entities/contents.js
@@ -46,6 +46,20 @@ export type NotebookContentRecordProps = {
   writable: boolean
 };
 
+export type DummyContentRecordProps = {
+  type: "dummy",
+  path: string
+};
+
+export const makeDummyContentRecord: Immutable.RecordFactory<
+  DummyContentRecordProps
+> = Immutable.Record({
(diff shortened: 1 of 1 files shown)


packages/core/src/state/entities/contents.js
```diff
@@ -64,8 +78,9 @@ export const makeContentRecord = makeNotebookContentRecord;
 export type NotebookContentRecord = Immutable.RecordOf<
   NotebookContentRecordProps
 >;
+
 // TODO: this will be a merger of notebook, directory, and file eventually.
-export type ContentRecord = NotebookContentRecord; // || DirectoryContentRecord
+export type ContentRecord = NotebookContentRecord | DummyContentRecord; // || DirectoryContentRecord
```
theengineear, line 83: heh. i'm surprised you didn't run into a crazy amount of flow errors here.

theengineear (approved):

packages/core/src/state/entities/contents.js
rgbkrk: 😄 

It helps that selectors isn't widely checked. Wait until my next PR. 😜

PR #2673, opened 2018-03-12
Filepathing

rgbkrk:
I'm switching us over to a single `filepath` property on our contents. Right now we have a few computed properties that need to be cleaned up.

In the process of making sure this was working I typed some of `selectors` to make it easier to ensure validity.

Totes ready.

### DIFF
--- a/applications/desktop/src/notebook/epics/loading.js
+++ b/applications/desktop/src/notebook/epics/loading.js
@@ -34,13 +34,13 @@ import type {
  * Determines the right kernel to launch based on a notebook
  */
 export const extractNewKernel = (
-  filename: ?string,
+  filepath: ?string,
   notebook: ImmutableNotebook
 ) => {
   // TODO: There's some incongruence between desktop and web app here, regarding path vs. filename
   //       Instead, this function is slightly repeated between here and @nteract/core
   const cwd =
-    (filename && path.dirname(path.resolve(filename))) || process.cwd();
+    (filepath && path.dirname(path.resolve(filepath))) || process.cwd();
   const kernelSpecName = notebook.getIn(
     ["metadata", "kernelspec", "name"],
     notebook.getIn(["metadata", "language_info", "name"], "python3")
@@ -116,14 +116,14 @@ export const fetchContentEpic = (action$: ActionsObservable<*>) =>
   action$.pipe(
     ofType(actionTypes.FETCH_CONTENT),
     tap((action: FetchContent) => {
-      // If there isn't a filename, save-as it instead
-      if (!action.payload.path) {
+      // If there isn't a filepath, save-as it instead
+      if (!action.payload.filepath) {
         throw new Error("fetch content needs a path");
       }
     }),
     // Switch map since we want the last load request to be the lead
     switchMap(action => {
-      const filepath = action.payload.path;
+      const filepath = action.payload.filepath;
 
       return forkJoin(
         readFileObservable(filepath),
@@ -135,7 +135,7 @@ export const fetchContentEpic = (action$: ActionsObservable<*>) =>
         timeout(60 * 1000),
         map(model =>
           actions.fetchContentFulfilled({
-            path: model.path,
+            filepath: model.path,
             model,
             kernelRef: action.payload.kernelRef,
             contentRef: action.payload.contentRef
@@ -144,7 +144,7 @@ export const fetchContentEpic = (action$: ActionsObservable<*>) =>
         catchError((err: Error) =>
           of(
             actions.fetchContentFailed({
-              path: filepath,
+              filepath,
               error: err,
               kernelRef: action.payload.kernelRef,
               contentRef: action.payload.contentRef
@@ -162,7 +162,7 @@ export const launchKernelWhenNotebookSetEpic = (
     ofType(actionTypes.SET_NOTEBOOK),
     map((action: SetNotebook) => {
       const { cwd, kernelSpecName } = extractNewKernel(
-        action.payload.filename,
+        action.payload.filepath,
         action.payload.notebook
       );
 
@@ -198,13 +198,12 @@ export const newNotebookEpic = (action$: ActionsObservable<*>) =>
       }
 
       return actions.setNotebook({
-        filename: null,
+        filepath: null,
         notebook,
         kernelRef,
         contentRef: action.payload.contentRef,
         created: null,
-        lastSaved: null,
-        name: null
+        lastSaved: null
       });
     })
   );
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -37,7 +37,7 @@ export function saveEpic(
         );
       }
 
-      const filename = selectors.currentFilename(state);
+      const filename = selectors.currentFilepath(state);
       // TODO: this default version should probably not be here.
       const appVersion = selectors.appVersion(state) || "0.0.0-beta";
       const notebook = stringifyNotebook(
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -86,7 +86,7 @@ export function saveAsEpic(action$: ActionsObservable<*>) {
         // order matters here, since we need the filename set in the state
         // before we save the document
         actions.changeFilename({
-          filename: action.payload.filename,
+          filepath: action.payload.filepath,
           contentRef: action.payload.contentRef
         }),
         actions.save({
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -37,7 +37,7 @@ export function saveEpic(
         );
       }
 
-      const filename = selectors.currentFilepath(state);
+      const filepath = selectors.currentFilepath(state);
       // TODO: this default version should probably not be here.
       const appVersion = selectors.appVersion(state) || "0.0.0-beta";
       const notebook = stringifyNotebook(
@@ -45,7 +45,7 @@ export function saveEpic(
           currentNotebook.setIn(["metadata", "nteract", "version"], appVersion)
         )
       );
-      return writeFileObservable(filename, notebook).pipe(
+      return writeFileObservable(filepath, notebook).pipe(
         map(() => {
           if (process.platform !== "darwin") {
             const state = store.getState();
--- a/applications/desktop/src/notebook/epics/saving.js
+++ b/applications/desktop/src/notebook/epics/saving.js
@@ -28,6 +28,15 @@ export function saveEpic(
     mergeMap((action: Save) => {
       const state = store.getState();
       const currentNotebook = selectors.currentNotebook(state);
+      if (!currentNotebook) {
+        return of(
+          actions.saveFailed({
+            error: new Error("no notebook loaded to save"),
+            contentRef: action.payload.contentRef
+          })
+        );
+      }
+
       const filename = selectors.currentFilename(state);
       // TODO: this default version should probably not be here.
       const appVersion = selectors.appVersion(state) || "0.0.0-beta";
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -192,14 +192,6 @@ export function dispatchPublishUserGist(
   store.dispatch({ type: "PUBLISH_USER_GIST" });
 }
 
-/**
- * Redux dispatch function to run the focused cell and all cells below it.
- * It obtains the focused cell cell id and all code cell cell ids below.
- * It dispatches the {@link executeCell} action on all of those retrieved cells.
- *
- * @exports
- * @param {Object} store - The Redux store
- */
 export function dispatchRunAllBelow(store: *) {
   const state = store.getState();
   const contentRef = selectors.currentContentRef(state);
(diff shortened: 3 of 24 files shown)


applications/desktop/src/notebook/menu.js
```diff
@@ -51,33 +51,33 @@ export function showSaveAsDialog(): Promise<string> {
...
       // This happens when the user cancels the dialog
-      if (filename === undefined) {
-        resolve(filename);
+      if (filepath === undefined) {
+        resolve(filepath);
         return;
       }
       // Assume it was a good path otherwise
-      resolve(filename);
+      resolve(filepath);
```
theengineear, line 68: Can't we have a `null` here still? Is that going to be a problem?

applications/desktop/src/notebook/menu.js
```diff
@@ -403,17 +403,17 @@ export function exportPDF(
         )
       );
 
-      fs.writeFile(`${filename}.pdf`, data, error_fs => {
+      fs.writeFile(`${filepath}.pdf`, data, error_fs => {
```
theengineear: Will this end up like `blah.ipynb.pdf`?

applications/desktop/src/notebook/menu.js
```diff
@@ -439,9 +439,9 @@ export function triggerSaveAsPDF(store: *) {
 export function storeToPDF(store: *) {
   const state = store.getState();
   const notebookName = selectors.currentFilename(state);
-  let filename = path.basename(notebookName, ".ipynb");
+  let filepath = path.basename(notebookName, ".ipynb");
```
theengineear: 🚨 wait a tick. 

```
path.basename('/foo/bar/baz/asdf/quux.html', '.html');
// Returns: 'quux'
```

^^ from the docs

I think we don't actually want to call this `filepath` here. It's not. maybe just `basename` and then compoese `filepath` later when it actually _is_ the `filepath`?

packages/core/src/actionTypes.js
```diff
@@ -57,7 +57,7 @@ export const FETCH_CONTENT = "CORE/FETCH_CONTENT";
 export type FetchContent = {
   type: "CORE/FETCH_CONTENT",
   payload: {
-    path: string,
+    filepath: string,
```
theengineear, line 60: 👌

packages/core/src/actionTypes.js
```diff
@@ -160,7 +160,7 @@ export const CREATE_CELL_AFTER = "CREATE_CELL_AFTER";
 export type CreateCellAfter = {
   type: "CREATE_CELL_AFTER",
   payload: {
-    id: CellID,
+    id?: CellID,
```
theengineear, line 163: Did this come up as part of changing the selector types?

packages/core/src/components/notebook-app.js
```diff
@@ -88,14 +88,20 @@ const mapStateToCellProps = (state, { id }) => {
   const outputExpanded =
     cellType === "code" && cell.getIn(["metadata", "outputExpanded"]);
 
+  const cellPagers = selectors.cellPagers(state);
+  const pager =
+    cellPagers && cellPagers.get
```
theengineear: huh, what the heck is a truthy `cellPagers` that `.get` can be missing?

packages/core/src/selectors/index.js
```diff
@@ -278,11 +283,10 @@ export const currentIdsOfHiddenOutputs = createSelector(
   }
 );
 
-// TODO: we can get the filename from the top-level instead of the model?
 export const currentFilename: (state: *) => string = createSelector(
   (state: AppState) => currentContent(state),
   currentContent => {
-    return currentContent.path;
+    return currentContent ? currentContent.filepath : "";
```
theengineear: Should this be renamed `currentFilepath`?

theengineear (commented):
Looking good! This will be awesome to have cleaned up for our sanity!

applications/desktop/src/notebook/menu.js
rgbkrk: The electron API returns `undefined` when the user didn't select a file, and we handle that with invocations of this function.

applications/desktop/src/notebook/menu.js
rgbkrk: ~~Yep (not that it should, just that it's how it's written at the moment both prior to this PR and now).~~ It ends up as `/some/long/path.pdf`

applications/desktop/src/notebook/menu.js
rgbkrk: Ohhhhh.

rgbkrk: Great catch.

rgbkrk: This will be much clearer now. 🙇 Thank you.

packages/core/src/components/notebook-app.js
rgbkrk: I... uh... may have just been working around flow here for a sec. I'll come to `FIXME` this one (and also fix it).

packages/core/src/selectors/index.js
rgbkrk: Excellent choice sir. 🍷

packages/core/src/actionTypes.js
rgbkrk: Now that we have dedicated content refs and know that if we're working on a notebook document that the focused cell is unique to that notebook document. Since there are so many occurrences where the callee to an action creator was getting the focused cell then dispatching an action, I'd really like to simplify this to state the intent as -- if the id isn't set, assume it's the focused cell.

rgbkrk:
This is ready to review! I'm going to do a self-review now.

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -198,13 +198,12 @@ export const newNotebookEpic = (action$: ActionsObservable<*>) =>
...
 
       return actions.setNotebook({
-        filename: null,
+        filepath: null,
         notebook,
         kernelRef,
         contentRef: action.payload.contentRef,
         created: null,
-        lastSaved: null,
-        name: null
```
rgbkrk, line 207: Note how we're rid of the "extra" `name` here.

applications/desktop/src/notebook/epics/saving.js
```diff
@@ -28,15 +28,24 @@ export function saveEpic(
     mergeMap((action: Save) => {
       const state = store.getState();
       const currentNotebook = selectors.currentNotebook(state);
```
rgbkrk, line 30: IIRC, we have a content ref at this point and could use that to `getNotebookByRef` or something... A future enhancement perhaps?

applications/desktop/src/notebook/menu.js
```diff
@@ -192,14 +192,6 @@ export function dispatchPublishUserGist(
...
 }
 
-/**
- * Redux dispatch function to run the focused cell and all cells below it.
- * It obtains the focused cell cell id and all code cell cell ids below.
- * It dispatches the {@link executeCell} action on all of those retrieved cells.
- *
- * @exports
- * @param {Object} store - The Redux store
- */
```
rgbkrk, line 202: This comment was odd as it literally described the implementation and was a bit out of date.

applications/desktop/src/notebook/menu.js
theengineear: I guess I meant, don't we set `filepath` to `null` in our Record and we _only_ handle `undefined` above, not `undefined || null`.

applications/desktop/src/notebook/menu.js
```diff
@@ -369,17 +355,19 @@ export function dispatchNewNotebook(
  * It will expand all cell outputs before printing and restore cells it expanded when complete.
  *
  * @param {object} store - The Redux store
- * @param {string} filename - filename of PDF to be saved.
+ * @param {string} basepath - basepath of the PDF to be saved.
```
theengineear, line 358: 🥇

applications/desktop/src/notebook/menu.js
```diff
@@ -438,10 +426,10 @@ export function triggerSaveAsPDF(store: *) {
 
 export function storeToPDF(store: *) {
   const state = store.getState();
-  const notebookName = selectors.currentFilename(state);
-  let filename = path.basename(notebookName, ".ipynb");
+  const notebookName = selectors.currentFilepath(state);
```
theengineear, line 429: super nit, but, this variable should probably be `filepath` to be consistent?

applications/desktop/src/notebook/native-window.js
```diff
@@ -70,7 +70,7 @@ export function createTitleFeed(state$: *) {
 
   const fullpath$ = state$.pipe(
     filter(state => selectors.currentContentRef(state)),
-    map(state => selectors.currentFilename(state) || "Untitled")
+    map(state => selectors.currentFilepath(state) || "Untitled")
```
theengineear, line 73: Couple notes:

1. `fullpath$` --> `filepath$`?
2. "Untitled" as the default seems not to qualify as a `filepath`, does it?
3. `selectors.currentFilepath(state)` should likely handle returning this default.

Non-blocking on these, just wondering if you had thoughts on it.

packages/core/__tests__/components/notebook-menu-spec.js
```diff
@@ -62,7 +62,7 @@ describe("PureNotebookMenu ", () => {
         cellMap: Immutable.Map(),
         cellOrder: Immutable.List(),
         notebook: Immutable.Map(),
-        filename: "fake.ipynb",
+        filepath: "fake.ipynb",
```
theengineear, line 65: So, maybe this is actually just my misunderstanding. But something like `fake.ipynb` means that this exists at `./fake.ipynb`? and `.` is the `cwd` where the notebook server was started?

packages/core/__tests__/document-spec.js
```diff
@@ -622,49 +624,100 @@ describe("setLanguageInfo", () => {
...
+    const secondId = uuid();
+    const thirdId = uuid();
 
-    const id = originalState.getIn(["notebook", "cellOrder"]).first();
-    const cell = originalState.getIn(["notebook", "cellMap", id]);
-    const state = reducers(originalState, actions.copyCell({ id }));
-    expect(state.getIn(["copied", "cell"])).toBe(cell);
-    expect(state.getIn(["copied", "id"])).toBe(id);
+    const originalState = makeDocumentRecord({
+      notebook: Immutable.fromJS({
```
theengineear, line 632: Need to record-ify notebooks, eh 😛

packages/core/src/components/notebook-menu/index.js
```diff
@@ -408,8 +402,7 @@ class PureNotebookMenu extends React.Component<Props, State> {
 // information about the current document to decide which menu items are
 // available...
 const mapStateToProps = state => ({
-  cellFocused: selectors.currentFocusedCellId(state),
```
theengineear, line 411: 👍

packages/core/src/selectors/index.js
```diff
@@ -278,11 +281,10 @@ export const currentIdsOfHiddenOutputs = createSelector(
   }
 );
 
-// TODO: we can get the filename from the top-level instead of the model?
-export const currentFilename: (state: *) => string = createSelector(
+export const currentFilepath: (state: *) => string = createSelector(
   (state: AppState) => currentContent(state),
   currentContent => {
-    return currentContent.path;
+    return currentContent ? currentContent.filepath : "";
```
theengineear, line 287: Can use use `""` instead of `Untitled` where we've defaulted to the latter in other places?

theengineear: Like, some parts of the app have used this selector, but defaulted to `Untitled`. I'm wondering if we should (a) return `Untitled` here, or (b) use `""` in place of `Untitled`?

theengineear (approved):
Awesome. I left some comments, but I'd rather merge this now and you can take a look and decide if things need updating in a subsequent PR. Nice work!

applications/desktop/src/notebook/menu.js
rgbkrk: The `filepath` here comes from electron's `dialog.showSaveDialog` and is a new `filepath`. We could change this to handle `null`, `undefined`, and `""`, returning `undefined` explicitly in that case.

applications/desktop/src/notebook/native-window.js
rgbkrk: Ah, hmmmm. I'll 

For (1), yeah that sounds good.

For (2), it's what we show as the "display name" of the file, so maybe this should be changed to `displayName` or something? It's the name when we have a filepath and Untitled when we don't.

For (3), maybe for a new selector that derives from filepath. (2) explains that reasoning, since this is basically used for what's displayed to the user rather than for file system operations.

packages/core/__tests__/components/notebook-menu-spec.js
rgbkrk: Yup.

packages/core/__tests__/document-spec.js
rgbkrk: Sooooo much so.

packages/core/src/components/notebook-menu/index.js
rgbkrk: 🎉

packages/core/src/selectors/index.js
rgbkrk: Yeah we should definitely clean this up to know what our consistent view on the filepath is. We probably should be returning `null` in the unset case. It made it easier to reason about if it was a string that I could manipulate in other selectors and uses, while still being falsy as needed (since an empty string). Perhaps that's poor form.

rgbkrk:
> Awesome. I left some comments, but I'd rather merge this now and you can take a look and decide if things need updating in a subsequent PR. Nice work!

Thank you!

Issue #2655, opened 2018-03-08
Failed to minify `@nteract/styles/lib/styles.js` with create-react-app


tlyon3:
When building a react application using create-react-app, running `npm run build` returns a “Failed to minify” error resulting from `@nteract/styles/lib/styles.js:3`.

tlyon3:
Here is the output of my log:

```
➜  client git:(master) ✗ npm run build

> client@0.3.0 build /Users/tlyon3/Work/quantecon/repos/Bookshelf/client
> react-scripts build

Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file: 

 	./node_modules/@nteract/styles/lib/styles.js:3 

Read more here: http://bit.ly/2tRViJ9

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! client@0.3.0 build: `react-scripts build`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the client@0.3.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/tlyon3/.npm/_logs/2018-03-09T04_48_02_441Z-debug.log
```

rgbkrk:
Ah, I see. [`create-react-app` won't allow template literals](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify). It's bombing on:

```
module.exports = `
:root {
```

rgbkrk:
We'll have to change our babel configuration to make sure template literals are transpiled.

rgbkrk:
In the _next_ release of `create-react-app` it looks like they'll transpile dependencies, based on https://github.com/facebook/create-react-app/pull/3776 being on their `next` branch since January (it's not on `master`, so not shipped).

rgbkrk:
Ok, so our `babel` config _would_ create the right output. The issue is that this file in particular is generated javascript:

```
// Generate themeable css. Overwrite the whole file for the first write.
fs.writeFileSync(
  CSS_PATH,
  "// This file is auto-generated. See buildStyles.js for details.\n\n"
);
fs.appendFileSync(CSS_PATH, "module.exports = `\n:root {\n");
lodash.toPairs(declarations).forEach(([cssProperty, cssValue]) => {
[...]

PR #2674, opened 2018-03-13
Fix transpile

tlyon3:
The way the `/packages/styles/buildStyles.js` currently works gives a "Failed to minify" error when using this package in a `create-react-app` application.

This changes the output from `/packages/styles/buildStyles.js` so that it outputs on a single line so that the file can be minified.

This closes issue https://github.com/nteract/nteract/issues/2655

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -11,7 +11,7 @@
     "prepublishOnly": "npm run build && npm run build:flow",
     "build": "npm run build:clean && npm run build:lib",
     "build:clean": "rimraf lib",
-    "build:lib": "node buildStyles.js && npm run build:copy-files && babel -d lib src --ignore '**/__tests__/**'",
+    "build:lib": "node buildStyles.js && npm run build:copy-files && babel -d lib src --ignore '**/__tests__/**' && babel -d lib build",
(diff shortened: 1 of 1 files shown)


packages/styles/buildStyles.js
```diff
@@ -90,11 +90,11 @@ fs.writeFileSync(
   CSS_PATH,
   "// This file is auto-generated. See buildStyles.js for details.\n\n"
 );
-fs.appendFileSync(CSS_PATH, "module.exports = `\n:root {\n");
-lodash.toPairs(declarations).forEach(([cssProperty, cssValue]) => {
-  fs.appendFileSync(CSS_PATH, `  ${cssProperty}: ${cssValue};\n`);
+fs.appendFileSync(CSS_PATH, 'module.exports = ":root { ');
+Object.entries(declarations).forEach(([cssProperty, cssValue]) => {
```
rgbkrk: You're going to want to keep this using `lodash.toPairs` -- the node version you were using before (maybe you're upgraded now) was what led us to switching to `lodash.toPairs`.

rgbkrk: Ref #2498

theengineear:
@tlyon3 great! Other than @rgbkrk's comment, it looks good to me! Thanks for the change.

rgbkrk:
I've switched this up to use `babel` to do the transpile for `lib` so we can look at the full output in `build`.

PR #2680, opened 2018-03-15
Directory Contents!

rgbkrk:
directory-navigation

Note: the above GIF I cut loading out to make it appear smoother. It's fully loading the entire app (including notebook stuff...) in development mode, so it loads really _slow_. 

* [x] Show directory listings in the jupyter extension
* [x] Navigate between directories, notebooks, and files
* [x] View files (in `<pre>` blocks, no editors yet)
* [x] Create records for files, directories, and our placeholder
* [x] Open notebooks in a new tab
* [x] Put mimetype field on `Contents` objects
* [x] Only render files we know we can render for sure
* [x] Replace inferred `*` types with explicit state
* [x] ~Comment on why we turn `config.contentsPath` set to `""` into `/`~ Relax rx-jupyter to allow contents path of `""`, as it reflects the actual contents API `""` is `"/"`
* [x] Rely on our themes in the directory components
* [ ] Unify the unknowables (`empty`, `dummy`, `unknown`, etc.)

Future PRs:

* Incorporate react router
* use `import("moduleName")` with webpack to only load components we need

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
@@ -4,6 +4,9 @@ import * as React from "react";
 import * as Immutable from "immutable";
 const urljoin = require("url-join");
 
+import TimeAgo from "@nteract/timeago";
+import { Book, FileText, FileDirectory } from "@nteract/octicons";
+
 import { selectors, actions, state as stateModule } from "@nteract/core";
 
 // Workaround flow limitation for getting these types
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
@@ -35,7 +35,9 @@ class DirectoryEntry extends React.PureComponent<DirectoryEntryProps, *> {
     );
 
     let type = this.props.entry.type;
-    type = type === "dummy" ? this.props.entry.assumedType : type;
+    if (this.props.entry.type === "dummy") {
+      type = this.props.entry.assumedType;
+    }
 
     let Icon = FileText;
 
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
@@ -1,0 +1,132 @@
+// @flow
+
+import * as React from "react";
+import * as Immutable from "immutable";
+const urljoin = require("url-join");
+
+import { selectors, actions, state as stateModule } from "@nteract/core";
+
+// Workaround flow limitation for getting these types
+type ContentRef = stateModule.ContentRef;
+type ContentRecord = stateModule.ContentRecord;
+type DirectoryContentRecord = stateModule.DirectoryContentRecord;
+
+import { connect } from "react-redux";
+
+type DirectoryEntryProps = {
+  basePath: string,
+  entry: ContentRecord
+};
+
+/**
+ * Note: the URLs we create for clicking have to take into account:
+ *    Server base URL   e.g. https://j.nteract.io/user/kylek/
+ *    App base URL      e.g. /nteract/edit/
+ *    Filepath          e.g. Untitled.ipynb
+ */
+
+class DirectoryEntry extends React.PureComponent<DirectoryEntryProps, *> {
+  render() {
+    const displayName = this.props.entry.filepath.split("/").pop();
+    const href = urljoin(
+      this.props.basePath,
+      "/nteract/edit/",
+      this.props.entry.filepath
+    );
+
+    let type = this.props.entry.type;
+    type = type === "dummy" ? this.props.entry.assumedType : type;
+
+    let Icon = FileText;
+
+    switch (type) {
+      case "notebook":
+        Icon = Book;
+        break;
+      case "directory":
+        Icon = FileDirectory;
+        break;
+      case "file":
+        Icon = FileText;
+        break;
+    }
+
+    return (
+      <React.Fragment>
+        <Icon />
+        <a
+          href={href}
+          // When it's a notebook, we open a new tab
+          target={type === "notebook" ? "_blank" : undefined}
+        >
+          {displayName}
+        </a>
+        <style jsx>{`
+          a {
+            padding-left: 10px;
+            text-decoration: none;
+            color: var(--theme-app-fg);
+            background-color: var(--theme-app-bg);
+          }
+        `}</style>
+      </React.Fragment>
+    );
+  }
+}
+
+const mapStateToEntryProps = (
+  state: Object,
+  ownProps: { contentRef: ContentRef }
+): DirectoryEntryProps => {
+  const host = selectors.currentHost(state);
+  if (host.type !== "jupyter") {
+    throw new Error("This component only works with jupyter servers");
+  }
+
+  return {
+    entry: selectors.contentByRef(state, ownProps),
+    basePath: host.basePath
+  };
+};
+
+const ConnectedEntry = connect(mapStateToEntryProps)(DirectoryEntry);
+
+type DirectoryProps = {
+  content: DirectoryContentRecord
+};
+
+export class Directory extends React.PureComponent<DirectoryProps, *> {
+  render() {
+    const atRoot = this.props.content.filepath === "/";
+
+    return (
+      <ul>
+        {atRoot ? null : (
+          // When we're not at the root of the tree, show `..`
+          <a href="..">..</a>
+        )}
+        {this.props.content.model.listing.map(contentRef => (
+          <li key={contentRef}>
+            <ConnectedEntry contentRef={contentRef} />
+          </li>
+        ))}
+        <style jsx>{`
+          ul {
+            list-style-type: none;
+          }
+        `}</style>
+      </ul>
+    );
+  }
+}
+
+const mapStateToDirectoryProps = (
+  state: Object,
+  ownProps: { contentRef: ContentRef }
+): DirectoryProps => {
+  return { content: selectors.contentByRef(state, ownProps) };
+};
+
+export const ConnectedDirectory = connect(mapStateToDirectoryProps)(Directory);
+
+export default ConnectedDirectory;
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
@@ -17,7 +17,7 @@ type DirectoryContentRecord = stateModule.DirectoryContentRecord;
 import { connect } from "react-redux";
 
 type DirectoryEntryProps = {
-  type: "dummy" | "notebook" | "directory" | "file",
+  type: "unknown" | "notebook" | "directory" | "file",
   href: string,
   displayName: string
 };
@@ -71,7 +71,7 @@ class DirectoryEntry extends React.PureComponent<DirectoryEntryProps, *> {
 }
 
 const mapStateToEntryProps = (
-  state: Object,
+  state: stateModule.AppState,
   ownProps: { contentRef: ContentRef }
 ): DirectoryEntryProps => {
   const host = selectors.currentHost(state);
@@ -80,6 +80,15 @@ const mapStateToEntryProps = (
   }
 
   const entry = selectors.contentByRef(state, ownProps);
+  if (!entry) {
+    // TODO: Determine what we do if we try to load content that isn't in the byRef structure
+    return {
+      type: "unknown",
+      href: "",
+      displayName: "Unknown Content"
+    };
+  }
+
   const basePath = host.basePath;
 
   const displayName = entry.filepath.split("/").pop() || "";
@@ -100,19 +109,26 @@ const mapStateToEntryProps = (
 const ConnectedEntry = connect(mapStateToEntryProps)(DirectoryEntry);
 
 type DirectoryProps = {
-  content: DirectoryContentRecord
+  content: DirectoryContentRecord,
+  basePath: string
 };
 
 export class Directory extends React.PureComponent<DirectoryProps, *> {
   render() {
     const atRoot = this.props.content.filepath === "/";
 
+    const dotdothref = urljoin(
+      this.props.basePath,
+      "/nteract/edit/",
+      urljoin(this.props.content.filepath, "..")
+    );
+
     return (
       <ul>
         {atRoot ? null : (
           // TODO: Create a contentRef for `..`, even though it's a placeholder
           // When we're not at the root of the tree, show `..`
-          <DirectoryEntry href="#yolo" displayName=".." type="directory" />
+          <DirectoryEntry href={dotdothref} displayName=".." type="directory" />
         )}
         {this.props.content.model.items.map(contentRef => (
           <li key={contentRef}>
@@ -130,10 +146,20 @@ export class Directory extends React.PureComponent<DirectoryProps, *> {
 }
 
 const mapStateToDirectoryProps = (
-  state: Object,
+  state: stateModule.AppState,
   ownProps: { contentRef: ContentRef }
 ): DirectoryProps => {
-  return { content: selectors.contentByRef(state, ownProps) };
+  const host = selectors.currentHost(state);
+  if (host.type !== "jupyter") {
+    throw new Error("This component only works with jupyter servers");
+  }
+
+  const basePath = host.basePath;
+
+  return {
+    content: selectors.contentByRef(state, ownProps),
+    basePath
+  };
 };
 
 export const ConnectedDirectory = connect(mapStateToDirectoryProps)(Directory);
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
@@ -17,8 +17,9 @@ type DirectoryContentRecord = stateModule.DirectoryContentRecord;
 import { connect } from "react-redux";
 
 type DirectoryEntryProps = {
-  basePath: string,
-  entry: ContentRecord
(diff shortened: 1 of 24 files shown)


rgbkrk:
Since there is a pretty darn opinionated routing thing here for the "directory app", I'm thinking I may just move this into the jupyter extension for now to make it easier to iterate without conflicting with core itself.

theengineear:
I'll fire this up locally after a review, but mind adding an image of the new feature in the top comment?

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
```diff
@@ -0,0 +1,125 @@
...
+import * as React from "react";
+import * as Immutable from "immutable";
+const urljoin = require("url-join");
+
+import TimeAgo from "@nteract/timeago";
+import { Book, FileText, FileDirectory } from "@nteract/octicons";
+
+import { selectors, actions, state as stateModule } from "@nteract/core";
+
+// Workaround flow limitation for getting these types
```
theengineear, line 12: What's the limitation here again?

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
```diff
@@ -0,0 +1,125 @@
...
+        break;
+      case "file":
+        Icon = FileText;
+        break;
+    }
+
+    return (
+      <React.Fragment>
+        <Icon />
+        <a href={href}>{displayName}</a>
```
theengineear: Should we be auto-opening this in a new tab? Or, is this the same behavior as the classic view?

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
```diff
@@ -0,0 +1,125 @@
...
+        <style jsx>{`
+          a {
+            padding-left: 10px;
+            text-decoration: none;
+            color: gray;
```
theengineear: Can we use a color from our theme?

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
```diff
@@ -0,0 +1,125 @@
...
+          }
+        `}</style>
+      </React.Fragment>
+    );
+  }
+}
+
+const ConnectedEntry = connect(
+  (
```
theengineear: Are you just writing this as an anon-function in the `connect` call to get around multiple `mapStateToProps` functions or something? I feel like we should probably just define it as its own function for clarity.

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
```diff
@@ -0,0 +1,125 @@
...
+};
+
+export class Directory extends React.PureComponent<DirectoryProps, *> {
+  render() {
+    const atRoot = this.props.content.filepath === "/";
+
+    return (
+      <ul>
+        {atRoot ? null : (
+          // When we're not at the root of the tree, show `..`
```
theengineear: 👌

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
```diff
@@ -0,0 +1,30 @@
...
+
+import { connect } from "react-redux";
+
+type FileProps = {
+  content: FileContentRecord
+};
+
+export class File extends React.PureComponent<FileProps, *> {
+  render() {
+    return <pre>{this.props.content.model.text}</pre>;
```
theengineear: Any issues around rendering images and such? Maybe we should leave a TODO in here?

applications/jupyter-extension/nteract_on_jupyter/app/contents/index.js
```diff
@@ -0,0 +1,69 @@
...
+} from "@nteract/core";
+
+import { default as Directory } from "./directory";
+import { default as File } from "./file";
+
+type ContentRef = stateModule.ContentRef;
+
+import { connect } from "react-redux";
+
+const mapStateToProps = (state: *, ownProps: *): * => {
```
theengineear, line 22: Can we type the state here or will `*` get us the same answer? I'm actually a little unclear if being specific ends up being redundant if `*` can suffice in a lot of cases?

applications/jupyter-extension/nteract_on_jupyter/app/contents/index.js
```diff
@@ -0,0 +1,69 @@
...
+  return {
+    contentType: selectors.currentContentType(state),
+    contentRef: selectors.currentContentRef(state)
+  };
+};
+
+class Contents extends React.PureComponent<*, *> {
```
theengineear: Define props / state types?

applications/jupyter-extension/nteract_on_jupyter/app/contents/index.js
```diff
@@ -0,0 +1,69 @@
...
+  render() {
+    switch (this.props.contentType) {
```
theengineear: ✨ squeaky clean ✨

applications/jupyter-extension/nteract_on_jupyter/app/contents/index.js
```diff
@@ -0,0 +1,69 @@
...
+          </React.Fragment>
+        );
+      case "file":
+        return (
+          <React.Fragment>
+            <TitleBar />
+            <File contentRef={this.props.contentRef} />
+          </React.Fragment>
+        );
+      case "dummy":
```
theengineear, line 47: I liked your idea on the `ContentsPlaceholder` idea here as a type instead of `dummy`. Not sure what the type string should be, `"placeholder"`?

applications/jupyter-extension/nteract_on_jupyter/app/index.js
```diff
@@ -95,6 +97,9 @@ function main(rootEl: Element, dataEl: Node | null) {
 
   try {
     config = JSON.parse(dataEl.textContent);
+    if (config.contentsPath === "") {
```
theengineear: Is this the default? Maybe a comment here on why we're assuming `"" === "/"`?

packages/core/src/components/notebook-app.js
```diff
@@ -74,9 +74,13 @@ const mapStateToCellProps = (state, { id }) => {
   const outputExpanded =
     cellType === "code" && cell.getIn(["metadata", "outputExpanded"]);
 
-  const pager = selectors
-    .currentModel(state)
-    .getIn(["cellPagers", id], Immutable.List());
+  const model = selectors.currentModel(state);
+
+  if (model.type !== "notebook") {
+    throw new Error("FAIL --- this aint a notebook and shouldn't have a cell");
```
theengineear: lol. Maybe a friendlier message 😀

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -910,6 +916,66 @@ const byRef = (state = Immutable.Map(), action) => {
         })
       );
     case actionTypes.FETCH_CONTENT_FULFILLED:
+      if (action.payload.model.type === "file") {
```
theengineear: maybe opt for a `switch`? We seem to use that pattern elsewhere over this `if`-ing.

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -910,6 +916,66 @@ const byRef = (state = Immutable.Map(), action) => {
...
+        );
+      }
+
+      if (action.payload.model.type === "directory") {
+        // TODO
+        // For each entry in the directory listing, create a new contentRef
+        // and a "filler" contents object
+        // Optional: run through all the current contents to see if they're
+        //           a file we already have (?)
+        const listingDraft: Array<Object> = action.payload.model.content;
```
theengineear: Are you creating a separate variable here for typing? I.e., why not map directly over `action.payload.model.content`?

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -910,6 +916,66 @@ const byRef = (state = Immutable.Map(), action) => {
...
+        // Create a map of <ContentRef, ContentRecord> that we merge into the
+        // content refs state
+        const dummyRecords = Immutable.Map(
+          listingDraft.map(entry => {
+            return [
+              createContentRef(),
+              makeDummyContentRecord({
+                // TODO: We can store the type of this content,
+                // it just doesn't have a model
+                // entry.type
```
theengineear: 💯 for your comments.

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -910,6 +916,66 @@ const byRef = (state = Immutable.Map(), action) => {
...
+                assumedType: entry.type,
```
theengineear: 👌

packages/core/src/selectors/index.js
```diff
@@ -166,8 +162,12 @@ export const currentLastSaved = createSelector(
...
 );
 
-export const currentNotebookMetadata = createSelector(currentModel, model =>
-  model.getIn(["notebook", "metadata"], Immutable.Map())
+export const currentNotebookMetadata = createSelector(
+  currentModel,
+  model =>
+    model.type === "notebook"
+      ? model.notebook.get("metadata", Immutable.Map())
+      : Immutable.Map()
```
theengineear, line 170: 🤔 do we _have_ a concept of non-notebook metadata? I'm fine with leaving this for safety. Just curious.

packages/core/src/selectors/index.js
```diff
@@ -212,16 +212,22 @@ export const currentNotebookString = createSelector(
   }
 );
 
-export const currentFocusedCellId = createSelector(currentModel, model =>
-  model.get("cellFocused")
+export const currentFocusedCellId = createSelector(
+  currentModel,
+  model => (model.type === "notebook" ? model.cellFocused : null)
```
theengineear, line 217: This is because we allow `null`, but not `undefined` in related action creators?

packages/core/src/selectors/index.js
```diff
@@ -212,16 +212,22 @@ export const currentNotebookString = createSelector(
...
 );
 
-export const transientCellMap = createSelector(currentModel, model =>
-  model.getIn(["transient", "cellMap"], Immutable.Map())
+export const transientCellMap = createSelector(
+  currentModel,
+  model =>
+    model.type === "notebook"
+      ? model.transient.get("cellMap", Immutable.Map())
+      : Immutable.Map()
```
theengineear, line 230: Ditto here, it's probably more accurate to return `null` here, but I get that this is likely easier on `Flow`.

packages/core/src/state/entities/contents/directory.js
```diff
@@ -0,0 +1,41 @@
...
+
+export type DirectoryModelRecordProps = {
+  type: "directory",
+  listing: Immutable.Set<ContentRef>
+};
+export const makeDirectoryModel: Immutable.RecordFactory<
+  DirectoryModelRecordProps
+> = Immutable.Record({
+  type: "directory",
+  listing: Immutable.Set()
```
theengineear: Why Set over List?

packages/core/src/state/entities/contents/dummy.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+import * as Immutable from "immutable";
+
+export type EmptyModelRecordProps = {
```
theengineear, line 5: We've now got `empty`, `placeholder`, `unknown` and `dummy` floating around. Let's consolidate our language around this to prevent future confusion.

packages/core/src/state/entities/hosts.js
```diff
@@ -3,6 +3,16 @@ import * as Immutable from "immutable";
...
 import type { HostRef } from "../refs";
 
+export type EmptyHost = {
+  type: "empty"
+};
+export type EmptyHostRecord = Immutable.RecordOf<EmptyHost>;
+export const makeEmptyHostRecord: Immutable.RecordFactory<
+  EmptyHost
+> = Immutable.Record({
+  type: "empty"
```
theengineear, line 13: Can you comment on when / why we'd have an `empty` host?

theengineear (commented):
Just took a pass through on this. It's looking awesome! Let me know if any of my comments need more explanation. Also, apologies if some of the comments are a bit terse, I kinda went through this quickly!

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
rgbkrk: Since I can't import them at a top level (or can I?), I pulled them off the `stateModule`.

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
rgbkrk: New tab for the notebooks certainly. I'll file this as a TODO above.

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
rgbkrk: Added to the TODO list

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
rgbkrk: Will do. 😄

rgbkrk: I was definitely trying to get a prototype up quick, so this is one of those quick-and-dirty bits I'll clean up. 😃

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
rgbkrk: Heh, I should probably incorporate some of the handling from commuter or include the `mimetype` field we don't currently have on our `Contents`.

I'll add a few more TODOs for that.

applications/jupyter-extension/nteract_on_jupyter/app/contents/index.js
rgbkrk: We can type the state here. I'll update the TODO. It's better for us in the long run to be specific (if not truly a type for inferring).

applications/jupyter-extension/nteract_on_jupyter/app/contents/index.js
rgbkrk: Seems sensible to me.

applications/jupyter-extension/nteract_on_jupyter/app/index.js
rgbkrk: If it ends up as `""`, it's because the user went to `/nteract/edit` instead of `/nteract/edit/`. We could also make sure that the server side does a redirect -- I figured it would be ok to handle this case here. I'll add a comment.

packages/core/src/components/notebook-app.js
rgbkrk: hahaha, I forgot I did this

packages/core/src/reducers/core/entities/contents.js
rgbkrk: Ah yeah, we could do that. I split it up while I was figuring out what to do here.

packages/core/src/selectors/index.js
rgbkrk: I think this will get better if we switch to selectors that operate only on `NotebookContents` or a `NotebookModel` instead of AppState, which has to dive in via `byRef` and might be undefined.

packages/core/src/selectors/index.js
rgbkrk: Flow is being extremely accurate, because of how we use this in other areas it made logic simpler.

packages/core/src/state/entities/contents/directory.js
rgbkrk: I suppose in my mind, these were just an existence of entries. Any sorted behavior would come from an additional property that declared an ordering of the original set. 🤔 Maybe this should be a `Map` and a `List`. One for the actual items, and another for the view we're giving users.

packages/core/src/state/entities/contents/dummy.js
rgbkrk: 👍

packages/core/src/state/entities/hosts.js
rgbkrk: Something that made working with our selectors, reducers, and flow is to have a type for our "unsets" instead of `null`.

packages/core/src/state/entities/contents/directory.js
rgbkrk: Looking at that again, seems like `List` is the way, and we'll just be sorting / displaying that one. Key is still the `ContentRef` for the React elements.

rgbkrk: Updated this, provided a comment about the way ahead for `groupedBy` and `sortedBy`.

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.js
rgbkrk: Done.

packages/core/src/components/notebook-app.js
rgbkrk: Done.

rgbkrk:
Next up for me is unifying the dummy, empty, and placeholder types.

My current thinking is that FileContents is really the placeholder type for truly unknown files, and it's the model attribute that holds a potential EmptyModel. This reflects the Contents API itself really well since when doing a contents listing, each entry's "content" key is null. While we could keep it nullable, it's muchhhhh easier to reason about with flow if we have a type on a model (currentModel).

applications/jupyter-extension/nteract_on_jupyter/app/index.js
theengineear: Awesome, thanks!

PR #2687, opened 2018-03-16
allow html in markdown

### LINKED ISSUE #2685: Enable HTML inside markdown on commuter
Papermill creates little colored `<span>`:

screen shot 2018-03-16 at 10 02 46 am

and they're not shown in papermill, commuter, etc. We should relax the markdown implementation to allow HTML inside (but likely disable scripting, using caja).

/cc @charsmith

rgbkrk:
Attempting to close #2685 

This doesn't work though because inline HTML doesn't work appropriately in `react-markdown`. See https://github.com/rexxars/react-markdown/issues/43#issuecomment-373865528

### DIFF
--- a/packages/markdown/src/index.js
+++ b/packages/markdown/src/index.js
@@ -6,6 +6,12 @@ import RemarkMathPlugin from "./remark-math";
 
 import PropTypes from "prop-types";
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Closing in favor of #2778.

PR #2691, opened 2018-03-21
Environment Rendering

tlyon3:
The markdown renderer wouldn't render latex environments unless it was wrapped inside $$ (in math mode). This functionality is _different_ to how jupyter runs.

I added a pre-processor to the markdown renderer that goes through any markdown source and wraps any environments that are outside a $$ block.


rgbkrk:
> This functionality is different to how jupyter runs.

As in the current behavior is different and this brings us closer to parity?

tlyon3:
Correct. This helps bring it closer to how Jupyter functions

rgbkrk:
I'm sure edge cases will come up in the parsing, we'll shake 'em out though. 😄 

We should definitely put in an error boundary on this in case there are errors.

tlyon3:
@rgbkrk I went ahead and put in a case if the parser can't find a matching end tag for the environment and it will return an error. The markdown component will then just fall back to the original, unmodified text and a console error will be logged.

packages/markdown/src/index.js
```diff
@@ -18,6 +19,9 @@ const MarkdownRender = (
   props: ReactMarkdown.ReactMarkdownProps,
   context: { MathJaxContext?: boolean }
 ) => {
+  const processedText = processEnv(props.source);
```
rgbkrk, line 22: My only hesitation on this PR is that it's going to recompute this every time it renders.

Perhaps we should just memoize the `processEnv` function. At least on desktop, it'll be expensive for anything that renders within the tree.
at least on desktop, it'll be expensive otherwise

mmcky:
+1 for getting this to work. We would really like to support math environments such as `\begin{equation}` for the upcoming QuantEcon bookshelf project (http://notes.quantecon.org/). A lot of notebook writers we have talked with use these environments without explicit ``$$`` frequently and it will improve the rendering of submitted notebooks and harmonise with the behaviour exhibited in `jupyter notebook` and `jupyter lab`

rgbkrk:
I've got some cycles this week to look at this while I'm also going to move our MathJax context API over to the new React 16.3 Context API.

lgeiger:
We've moved our MathJax setup to the new context API in #3384 and are now able to add new [plugins](https://github.com/nteract/nteract/tree/master/packages/markdown/src/remark-math).

@tlyon3 Do you want to revisit this PR?

rgbkrk:
Since this hasn't been touched in quite a few months and our mathjax + markdown setup has had a typescript overhaul, I'll assume that this would need to be a new PR. Thanks again @tlyon3!

PR #2693, opened 2018-03-22
[jext] cut plotly out of blocking assets

rgbkrk:
If the network is slow, loading plotly can block loading the app. Since we're not using the plotly transform in the jupyter extension yet, I'd like to take it off the page (for now).


theengineear (approved):
Cool with me (for now)

rgbkrk:
I think what I'd want to do for this is to do dynamic loading of the transforms. We'd boot up the app as fast as we can, run the `import()` calls (the promise backed version) through an epic, and update what the current transforms are. The outputs would then re-render with the better mimetype.

PR #2698, opened 2018-03-23
Font loader

rgbkrk:
Switch to using a font-loader instead of blocking page rendering. This will likely cause [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) on first load.


theengineear (approved):
Cool. Looks good!

PR #2699, opened 2018-03-23
Sessions!

rgbkrk:
Introduce the concept of sessions. Definitely a work in progress here.

* [x] Track session ID inside the Remote kernel object (?)
  * We put the session ID in the contents records in our plan document, whereas when we say to kill a kernel we pass the kernelRef onward
* [x] Creating a kernel should be done via the session API
* [x] Stopping a kernel should be done by `DELETE`ing a session
* [x] Ensure restart works properly

### DIFF
--- a/packages/core/src/epics/websocket-kernel.js
+++ b/packages/core/src/epics/websocket-kernel.js
@@ -20,7 +20,7 @@ import { from } from "rxjs/observable/from";
 import { merge } from "rxjs/observable/merge";
 import { empty } from "rxjs/observable/empty";
 
-import { kernels, shutdown } from "rx-jupyter";
+import { kernels, shutdown, sessions } from "rx-jupyter";
 import { v4 as uuid } from "uuid";
 
 import * as actions from "../actions";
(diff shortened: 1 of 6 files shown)


rgbkrk:
* [ ] Switching kernels must stop a session and create a new session

packages/core/src/actionTypes.js
```diff
@@ -789,6 +792,47 @@ export type ShutdownReplyTimedOut = {
...
+export type CreateSessionFulfilled = {
+  type: "CORE/CREATE_SESSION_FULFILLED",
+  payload: {
+    contentRef: ContentRef,
+    kernelRef: KernelRef,
+    sessionId: SessionId,
+    // Should we not do the `channels` connection when we create the session (?)
+    kernel: RemoteKernelProps
+  }
+};
```
rgbkrk: These actions are not used as I was able to simplify the kernel/session creation within the websocket kernel setup. I think I'll go ahead and delete these so they're not confusing by being with the other types before we really use them.

packages/core/src/epics/websocket-kernel.js
```diff
@@ -44,18 +44,41 @@ export const launchWebSocketKernelEpic = (action$: *, store: *) =>
...
+        kernel: {
+          id: null,
+          name: kernelSpecName
+        },
+        name: "",
+        path: content.filepath,
+        type: "notebook"
+      };
+
+      // TODO: Handle failure cases here
```
theengineear, line 72: Were you intending to TODONE these in this PR?

theengineear: Perhaps this happens in a later commit though. I'm going one-by-one.

packages/rx-jupyter/src/sessions.js
```diff
@@ -27,7 +26,10 @@ export function list(serverConfig: Object): Observable<*> {
  *
  * @return  {Object}  An Observable with the request/response
  */
-export function get(serverConfig: Object, sessionID: string): Observable<*> {
+export function get(
+  serverConfig: Object,
+  sessionID: string
```
theengineear, line 31: nit: `sessionId` (not that you really changed that here...)

packages/core/src/epics/websocket-kernel.js
```diff
@@ -71,13 +72,15 @@ export const launchWebSocketKernelEpic = (action$: *, store: *) =>
         mergeMap(data => {
           const session = data.response;
 
+          const sessionId = castToSessionId(session.id);
```
theengineear: Nice!

theengineear (approved):
Wow. Not as bad as I was expecting! Nice work! I'm assuming that handling some of the kernel actions directly via the kernels API is OK for now, then?

packages/rx-jupyter/src/sessions.js
rgbkrk: Totally fine to make a new-contributor-friendly issue to go through updating these, as this library was in place wayyy before we came to the casing agreement and it wouldn't be too hard to get these in line. 😄

rgbkrk:
> I'm assuming that handling some of the kernel actions directly via the kernels API is OK for now, then?

Yeah, interrupt, restart, and connection all are supposed to happen via the kernel API. Only create and stop need to happen via the sessions API (note -- this includes changing kernels, which we don't support yet).

PR #2701, opened 2018-03-25
(fix): sort file>new by kernel display_name

jkornblum:
Not sure how I missed this when I PR'd for runtime>kernel sort. The code is top of each other :laughing: 

Sorting file>new the same way runtime>kernels are sorted


rgbkrk:
This is great, thanks @jkornblum!

rgbkrk:
I forgot to hit that merge button! Whoops. Now it's in. 😄

PR #2707, opened 2018-03-28
Download epic

rgbkrk:
This moves the notebook file downloading into an epic, freeing the menu up from re-rendering on every single keystroke in the notebook...


rgbkrk:
Gotta fix up the tests then this is good to go.

rgbkrk:
Super good to go now!

packages/core/src/actions.js
```diff
@@ -138,7 +57,7 @@ export const fetchContent = (payload: {
   params: Object,
   kernelRef: KernelRef,
   contentRef: ContentRef
-}): FetchContent => ({
+}): actionTypes.FetchContent => ({
```
theengineear, line 60: Oh dang! Good idea!

packages/core/__tests__/components/notebook-menu-extra-handlers-spec.js
```diff
@@ -1,39 +0,0 @@
-// @flow
```
rgbkrk, line 1: Note -- these tests got moved into the contents tests (of which there were none previously).

packages/core/src/components/notebook-menu/index.js
```diff
@@ -407,23 +403,15 @@ const mapStateToProps = state => {
...
-    filepath: selectors.currentFilepath(state),
-    notebook: model.notebook,
     currentKernelRef: selectors.currentKernelRef(state),
     currentContentRef: contentRef
   };
 };
 
 const mapDispatchToProps = dispatch => ({
   saveNotebook: (payload: *) => dispatch(actions.save(payload)),
+  downloadNotebook: (payload: *) => dispatch(actions.downloadContent(payload)),
```
theengineear, line 414: 👍

packages/core/src/epics/contents.js
```diff
@@ -79,69 +79,97 @@ export function fetchContentEpic(
...
+        }
+
+        const serverConfig = selectors.serverConfig(host);
+
+        const model = {
+          content: notebook,
+          type: "notebook"
+        };
+
+        // if (action.type === actionTypes.SAVE)
```
theengineear: :zap:?

theengineear (approved):
Cool. I feel like it's a little odd that download / save share the epic as far as organization goes. We could potentially factor out the common code and create two epics... Not sure how you feel about it.

packages/core/src/actions.js
rgbkrk: Yeah I was going a little crazy with the import type then use pattern happening in here. Way simpler to get going and enforce a type on the action creators now. 😄

packages/core/src/epics/contents.js
rgbkrk: Perhaps I should have left this as a comment instead:

```
// NOTE: assumption is action.type === actionTypes.SAVE at this point
```

rgbkrk:
> Cool. I feel like it's a little odd that download / save share the epic as far as organization goes. We could potentially factor out the common code and create two epics... Not sure how you feel about it.

The amount of boilerplate we've been writing for this stuff has been driving me _slightly_ crazy. I'm going to push up a commit and see what you think.

PR #2709, opened 2018-03-28
Add currentKernelspecsRef to state.

theengineear:
Just sets the currentKernelspecsRef in our core state on FETCH_KERNELSPECS.

### DIFF
--- a/packages/core/src/reducers/core/index.js
+++ b/packages/core/src/reducers/core/index.js
@@ -30,10 +30,20 @@ const currentContentRef = (state = null, action) => {
   }
 };
 
+const currentKernelspecsRef = (state = null, action) => {
+  switch (action.type) {
+    case actionTypes.FETCH_KERNELSPECS:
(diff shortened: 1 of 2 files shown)


packages/core/src/reducers/core/index.js
```diff
@@ -30,10 +30,20 @@ const currentContentRef = (state = null, action) => {
   }
 };
 
+const currentKernelspecsRef = (state = null, action) => {
+  switch (action.type) {
+    case actionTypes.FETCH_KERNELSPECS:
+      return action.payload.kernelspecsRef;
```
rgbkrk, line 36: ohhhh, I was thinking we'd set it with `initialState` like we do for `currentContentRef`

theengineear: Ah, I see where you're coming from... yah, i suppose that makes sense. I can make a follow up PR to do that.

PR #2711, opened 2018-03-29
Fix kernelspecs epic

theengineear:
This fixes the casing from the api (snake) to the casing we want in memory (camel).

### DIFF
--- a/packages/core/src/epics/kernelspecs.js
+++ b/packages/core/src/epics/kernelspecs.js
@@ -37,7 +37,12 @@ export const fetchKernelspecsEpic = (
             kernelspecs[key] = {
               name: value.name,
               resources: value.resources,
-              ...value.spec
+              argv: value.spec.argv,
+              displayName: value.spec.display_name,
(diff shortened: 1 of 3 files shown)


packages/core/src/epics/kernelspecs.js
```diff
@@ -37,7 +37,12 @@ export const fetchKernelspecsEpic = (
             kernelspecs[key] = {
               name: value.name,
               resources: value.resources,
-              ...value.spec
+              argv: value.spec.argv,
+              displayName: value.spec.display_name,
+              env: value.spec.env,
+              interruptMode: value.spec.interrupt_mode,
+              language: value.spec.language,
+              metadata: value.spec.metadata
```
rgbkrk, line 45: 👍  Looking great

One thing we probably need to add on to the reducer is making this an `Immutable.Map()`

rgbkrk:
Tests need a fix. Amusingly, they now catch the proper behavior for us. 😄

PR #2715, opened 2018-03-29
Change kernels from menu

theengineear:
♻️🌽

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.js
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.js
@@ -146,7 +146,12 @@ export function launchKernelObservable(
           };
 
           observer.next(
-            actions.launchKernelSuccessful({ kernel, kernelRef, contentRef })
+            actions.launchKernelSuccessful({
+              kernel,
+              kernelRef,
+              contentRef,
+              selectNextKernel: true
+            })
           );
           // TODO: Request status right after
           observer.next(
--- a/packages/core/src/actions.js
+++ b/packages/core/src/actions.js
@@ -146,6 +146,17 @@ export function launchKernel(payload: {
   };
 }
 
+export function changeKernelByName(payload: {
+  kernelSpecName: any,
+  kernelRef: KernelRef,
+  contentRef: ContentRef
+}): actionTypes.ChangeKernelByName {
+  return {
+    type: actionTypes.CHANGE_KERNEL_BY_NAME,
+    payload
+  };
+}
+
 export function launchKernelByName(payload: {
   kernelSpecName: any,
   cwd: string,
--- a/packages/core/src/actions.js
+++ b/packages/core/src/actions.js
@@ -124,7 +124,8 @@ export function launchKernelFailed(payload: {
 export function launchKernelSuccessful(payload: {
   kernel: LocalKernelProps | RemoteKernelProps,
   kernelRef: KernelRef,
-  contentRef: ContentRef
+  contentRef: ContentRef,
+  selectNextKernel: boolean
 }): actionTypes.NewKernelAction {
   return {
     type: actionTypes.LAUNCH_KERNEL_SUCCESSFUL,
--- a/packages/core/src/actions.js
+++ b/packages/core/src/actions.js
@@ -148,7 +148,7 @@ export function launchKernel(payload: {
 
 export function changeKernelByName(payload: {
   kernelSpecName: any,
-  kernelRef: KernelRef,
+  oldKernelRef: ?KernelRef,
   contentRef: ContentRef
 }): actionTypes.ChangeKernelByName {
   return {
--- a/packages/core/src/components/notebook-app.js
+++ b/packages/core/src/components/notebook-app.js
@@ -6,7 +6,8 @@ import * as React from "react";
 import * as actions from "../actions";
 import * as selectors from "../selectors";
 const themes = require("../themes");
-import type { ContentRef } from "../state/refs";
+import type { AppState } from "../state";
+import type { ContentRef, KernelRef } from "../state/refs";
 
 import { Input, Prompt, Source, Pagers, Outputs, Cell } from "./presentational";
 
@@ -287,7 +288,8 @@ type PureNotebookProps = {
   // TODO: Once we're willing to do multi-contents views, we should require this
   //       to be passed in
   // TODO: Fill in more from the convo with Andrew here
-  contentRef?: ContentRef
+  contentRef?: ContentRef,
+  kernelRef?: KernelRef
 };
 
 type NotebookStateProps = {
@@ -299,7 +301,8 @@ type NotebookStateProps = {
   languageDisplayName: string,
   kernelStatus: string,
   codeMirrorMode: string | Immutable.Map<string, *>,
-  contentRef: ContentRef
+  contentRef: ContentRef,
+  kernelRef?: KernelRef
 };
 
 type NotebookDispatchProps = {
@@ -311,11 +314,13 @@ type NotebookDispatchProps = {
 };
 
 const mapStateToProps = (
-  state: Object,
+  state: AppState,
   ownProps: PureNotebookProps
 ): NotebookStateProps => {
   // TODO: Switch to ownProps
   const contentRef = ownProps.contentRef || selectors.currentContentRef(state);
+  const kernelRef = ownProps.kernelRef || selectors.currentKernelRef(state);
+
   if (!contentRef) {
     throw new Error("<Notebook /> has to have a contentRef");
   }
@@ -334,7 +339,7 @@ const mapStateToProps = (
       lastSaved: content.lastSaved,
       cellOrder: Immutable.List(),
       // TODO: deal with current kernel ref
-      kernelStatus: selectors.currentKernelStatus(state),
+      kernelStatus: "unknown",
       languageDisplayName: "unknown",
       transforms: ownProps.transforms || transforms,
       displayOrder: ownProps.displayOrder || displayOrder,
@@ -349,16 +354,32 @@ const mapStateToProps = (
     );
   }
 
+  let kernel = {
+    kernelSpecName: null,
+    status: null
+  };
+
+  if (kernelRef) {
+    kernel = selectors.kernel(state, { kernelRef }) || kernel;
+  }
+
+  // NOTE: We should use what the kernel reports before using what the notebook "reports"
+  // TODO: The kernel object should contain it's own information
+  const languageDisplayName =
+    kernel.kernelSpecName || selectors.notebook.displayName(model);
+  // TODO: Rely on the kernel's codeMirror version first and foremost, then fallback on notebook
+  const codeMirrorMode = selectors.notebook.codeMirrorMode(model);
+
   return {
     theme: selectors.userPreferences(state).theme,
     lastSaved: content.lastSaved,
     cellOrder: selectors.notebook.cellOrder(model),
     // TODO: deal with current kernel ref
-    kernelStatus: selectors.currentKernelStatus(state),
-    languageDisplayName: selectors.notebook.displayName(model),
+    kernelStatus: kernel.status || "not connected",
+    languageDisplayName,
     transforms: ownProps.transforms || transforms,
     displayOrder: ownProps.displayOrder || displayOrder,
-    codeMirrorMode: selectors.notebook.codeMirrorMode(model),
+    codeMirrorMode,
     contentRef
   };
 };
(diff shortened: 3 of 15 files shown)


theengineear:
@rgbkrk want to pull this down and try it out? This is definitely missing some polish, but as a first pass, it should be switching kernels for us.

rgbkrk:
> not really understanding why I don't need to manually kill the kernel? Is it possibly that that's just getting taken care of under the hood for us?

As you pointed out to me on slack, when updating the session with a new kernel it will kill the old kernel: https://github.com/jupyter/notebook/blob/master/notebook/services/sessions/handlers.py#L144

(noting this for the sake of having a record)

rgbkrk:
Ok, this made something explicit that was always hidden away -- the ui state for the kernel status should come from the kernel, not from what's in the notebook document. This gives them a consistent view of what they're really connected to, rather than what is in the document. I've made the "choice" of what the language to come from the kernel first and then fallback to the notebook.

A follow on PR needs to do several things:

* [ ] Track the codemirror mode, display name, etc. inside the kernel object
* [ ] Move the notebook fully off of `current*` selectors

rgbkrk:
I'm going to fix all the tests and then do an earnest review.

applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -146,7 +146,12 @@ export function launchKernelObservable(
           };
 
           observer.next(
-            actions.launchKernelSuccessful({ kernel, kernelRef, contentRef })
+            actions.launchKernelSuccessful({
+              kernel,
+              kernelRef,
+              contentRef,
```
rgbkrk, line 152: Why does launch kernel successful need the content ref?

packages/core/src/actions.js
```diff
@@ -145,6 +146,17 @@ export function launchKernel(payload: {
...
+export function changeKernelByName(payload: {
+  kernelSpecName: any,
+  kernelRef: KernelRef,
+  contentRef: ContentRef
+}): actionTypes.ChangeKernelByName {
+  return {
+    type: actionTypes.CHANGE_KERNEL_BY_NAME,
+    payload
+  };
+}
```
rgbkrk, line 158: On first glance (linear viewing based on the GitHub UI), I thought this was two events under the covers -- "create a new kernel with this kernel ref and set this contentRef to use this kernelRef"

After reading the epic though, it's "for this contentRef, create a new kernel using this name, and also here's the current kernel ref so it's easier to get access to the session".

We should probably allocate the new kernelRef at this point and optionally include the old kernel ref.

packages/core/src/components/notebook-app.js
```diff
@@ -311,11 +314,13 @@ type NotebookDispatchProps = {
 };
 
 const mapStateToProps = (
-  state: Object,
+  state: AppState,
```
rgbkrk, line 317: Yay for actual flow type checking. 😄

packages/core/src/components/notebook-app.js
```diff
@@ -349,16 +354,32 @@ const mapStateToProps = (
...
+  if (kernelRef) {
+    kernel = selectors.kernel(state, { kernelRef }) || kernel;
+  }
+
+  // NOTE: We should use what the kernel reports before using what the notebook "reports"
+  // TODO: The kernel object should contain it's own information
+  const languageDisplayName =
+    kernel.kernelSpecName || selectors.notebook.displayName(model);
+  // TODO: Rely on the kernel's codeMirror version first and foremost, then fallback on notebook
+  const codeMirrorMode = selectors.notebook.codeMirrorMode(model);
```
rgbkrk, line 371: We need to make this come through with the current reduction step that handles this for the notebook and set it in the kernel record.

packages/core/src/components/notebook-menu/constants.js
```diff
@@ -39,5 +40,6 @@ export const MENUS = {
   VIEW: "view",
   VIEW_THEMES: "view-themes",
   RUNTIME: "runtime",
+  RUNTIME_CHANGE_KERNEL: "runtime-change-kernel",
```
rgbkrk, line 43: What's the difference between `CHANGE_KERNEL` and `RUNTIME_CHANGE_KERNEL`?

packages/core/src/components/notebook-menu/index.js
```diff
@@ -229,7 +239,15 @@ class PureNotebookMenu extends React.Component<Props, State> {
           });
         }
         break;
-
+      case MENU_ITEM_ACTIONS.CHANGE_KERNEL:
+        if (changeKernelByName) {
+          changeKernelByName({
+            kernelRef: currentKernelRef,
+            contentRef: currentContentRef,
```
rgbkrk: Ok now that I see this I see my confusion with the action's types. We could just be setting the contentRef here since we're not creating a new kernel.

packages/core/src/components/notebook-menu/index.js
```diff
@@ -366,10 +388,34 @@ class PureNotebookMenu extends React.Component<Props, State> {
...
+              title="Change Kernel"
+              disabled={!currentKernelspecs}
+            >
+              {currentKernelspecs
+                ? currentKernelspecs.byName
+                    .keySeq()
+                    .map(name => [
+                      name,
+                      currentKernelspecs.byName.getIn([name, "displayName"])
+                    ])
```
rgbkrk, line 407: This could probably be created in the `mapStateToProps` call right? Since the Provider component handles `shouldComponentUpdate` for us, we could pass an object that has just what we need into this component.

packages/core/src/epics/websocket-kernel.js
```diff
@@ -94,22 +91,104 @@ export const launchWebSocketKernelEpic = (action$: *, store: *) =>
...
+      }
+
+      const content = selectors.content(state, { contentRef });
+      if (!content || content.type !== "notebook") {
+        return empty();
+      }
+      const { filepath, model: { notebook } } = content;
+      const { cwd } = extractNewKernel(filepath, notebook);
+
+      const kernelRef = createKernelRef();
```
rgbkrk, line 145: I feel like we should be:

* creating the new kernel ref outside of here (part of the action)
* _this_ action (change kernel by name) should be used on reduction to set the new kernel for the content

If necessary, we can track the oldKernelRef in the action as well

packages/core/src/epics/websocket-kernel.js
```diff
@@ -94,22 +91,104 @@ export const launchWebSocketKernelEpic = (action$: *, store: *) =>
...
+      return kernels.start(serverConfig, kernelSpecName, cwd).pipe(
+        mergeMap(({ response }) => {
+          const { id: kernelId } = response;
+          const sessionPayload = {
+            kernel: { id: kernelId, name: kernelSpecName }
+          };
+          return sessions.update(serverConfig, sessionId, sessionPayload).pipe(
```
rgbkrk: We should make a note in here about how the sessions API closes down the old kernel for us

packages/core/src/epics/websocket-kernel.js
```diff
@@ -94,22 +91,104 @@ export const launchWebSocketKernelEpic = (action$: *, store: *) =>
...
+      const state = store.getState();
+      const host = selectors.currentHost(state);
+      if (host.type !== "jupyter") {
+        // Dismiss any usage that isn't targeting a jupyter server
+        return empty();
+      }
+      const serverConfig = selectors.serverConfig(host);
+
+      const oldKernel = selectors.kernel(state, { kernelRef: oldKernelRef });
+      if (!oldKernel || oldKernel.type !== "websocket") {
```
rgbkrk: If the old kernel doesn't exist in this case, we need to create a new session with the requested kernel name (so the user can get to a working state).

rgbkrk (approved):

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -919,6 +919,13 @@ const byRef = (state = Immutable.Map(), action) => {
           // TODO: we can set kernelRef when the content record uses it.
         })
       );
+    case actionTypes.LAUNCH_KERNEL_SUCCESSFUL:
+      // TODO: is this reasonable? We launched the kernel on behalf of this
+      // content... so it makes sense to swap it, right?
+      return state.setIn(
+        [action.payload.contentRef, "model", "kernelRef"],
+        action.payload.kernelRef
+      );
```
theengineear, line 928: This is why launch kernel successful is taking a content ref. I didn't know *when* the content should update it's kernel ref though. should it do it immediately? or only after success here?

applications/desktop/src/notebook/epics/zeromq-kernels.js
theengineear: See https://github.com/nteract/nteract/pull/2715/files#r178346701

packages/core/src/actions.js
theengineear: Hmm. I was considering it like when you `killKernel`, you give it the kernel you want to kill. And change kernel gets the kernel you wan to `change`. Since the killing happens under-the-hood here in the sessions api, I think that fits a little better. Does that change your mind at all 😁?

packages/core/src/reducers/core/entities/contents.js
rgbkrk: Immediately. If the kernel is already being changed out, they shouldn't remain connected (visually) to the old kernel they were done with.

packages/core/src/components/notebook-app.js
theengineear: Is that something for this PR or can it hold off until a later date?

packages/core/src/components/notebook-menu/constants.js
theengineear: `RUNTIME_CHANGE_KERNEL` is a menu. Whereas `CHANGE_KERNEL` is a menu-action (meant to be suffixed with a kernelSpecName)

packages/core/src/components/notebook-menu/index.js
theengineear: Heh. Not sure I understand that comment. What other comment / thought is this referring to?

packages/core/src/components/notebook-menu/index.js
theengineear: Yep. Totally.

packages/core/src/actions.js
rgbkrk: Yeah that makes sense. I definitely think we could make this action be:

```
{
  oldKernelRef,
  nextKernelRef,
  contentRef,
  kernelName,
}
```

packages/core/src/epics/websocket-kernel.js
theengineear: Ya! I wanted to ask you if you had thoughts on this one. I didn't _really_ want to handle it here 😛. Can we do a follow up for this?

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: Oops, I meant to delete this comment after I found out. I take a lot of notes in the review, expecting to answer my own questions after. Sorry bout that.

Seems the answer, assuming we agree on my statement, is that we change the kernel in the `CHANGE_KERNEL_BY_NAME` rather than the successful launch. A successful launch merely updates the kernel record to provide the up to date info.

packages/core/src/epics/websocket-kernel.js
theengineear: Cool. You might be starting to convince me. I'll have a look and maybe update that!

packages/core/src/components/notebook-menu/constants.js
rgbkrk: Ohhhhhh. Thank you.

packages/core/src/components/notebook-menu/index.js
rgbkrk: The one earlier about what kind of kernelRef this was in the lifecycle -- was is the old or the new. It's the old kernel and the new kernel is launched here.

packages/core/src/epics/websocket-kernel.js
rgbkrk: Totally. I think for a bunch of these we can just add TODOs and go from there. This is super solid as is.

applications/desktop/src/notebook/epics/zeromq-kernels.js
rgbkrk: 🚩 Follow on PR for state reduction switch up.

packages/core/src/components/notebook-app.js
rgbkrk: Definitely something for another PR.

packages/core/src/components/notebook-menu/index.js
rgbkrk: Going to defer this for another PR, namely the PR that uses this same info for the New Notebook menu entries.

packages/core/src/reducers/core/entities/contents.js
rgbkrk: I'll handle this in a follow on PR. The tricky part is likely instantiating a kernel object without some of the info.

PR #2719, opened 2018-03-31
Kernel info

rgbkrk:
In an effort to surface the right information in the status bar in the kernel, I went after cleaning up our kernel info and making the notebook app take a `contentRef`.

## `setKernelInfo` action and `KernelInfo` record

We had an action called `setKernelInfo` that was only used on Desktop (as far as I can tell) that I've renamed to `setKernelspecInfo`. I've done this so we can have a `setKernelInfo` that actually matches the `kernel_info_request` and `kernel_info_reply` pairs.

There are now proper types for `KernelInfo`, and I've created a flattened version of the `kernel_info_reply` content for us to use.

## `<NotebookApp />`

In the process here, I've made the `<NotebookApp />` component require a `contentRef`. On Desktop it uses the `contentRef` set at the start (since we never swap it out on the same window).

## `<StatusBar />`

`<StatusBar />` is now a connected component that accepts props for `contentRef` and `kernelRef` so that it can look up exactly the information it needs.

## Generic import cleanups

In the process of doing this PR I ended up doing some cleanup of imports from core to simplify how I was working on this.


packages/core/src/state/entities/kernel-info.js
```diff
@@ -0,0 +1,71 @@
...
+  // (e.g. IPython.__version__ for the IPython kernel)
+  implementationVersion: string,
+
+  // A banner of information about the kernel,
+  // which may be displayed in console environments.
+  banner: string,
+
+  // Optional: A list of dictionaries, each with keys 'text' and 'url'.
+  // These can be displayed in the help menu for the UI.
+  help_links?: Array<{ text: string, url: string }>,
```
theengineear: Should this be Immutable here?

packages/core/src/state/entities/kernel-info.js
```diff
@@ -0,0 +1,71 @@
...
+
+  // Nbconvert exporter, if notebooks written with this kernel should
+  // be exported with something other than the general 'script'
+  // exporter.
+  nbconvert_exporter: string
+};
+
+export type KernelInfoRecord = Immutable.RecordOf<KernelInfo>;
+
+export function makeKernelInfoRecord(kernelInfo: KernelInfo): KernelInfoRecord {
```
theengineear: Nice, trying out the wrapper version of this?

rgbkrk: Yeah, works pretty well and forces these types to all be set and not have to do a big `null` hunt. The only strange thing is our creation of records. I wonder if I should make sure that two equivalent set of props passed into would evalute `Immutable.is(a,b)` to `true`.

applications/notebook-on-next/pages/edit.js
```diff
@@ -1,4 +1,5 @@
-// @flow
+// FIXME: To make it easier to iterate on supported apps, flow is disabled for
```
rgbkrk, line 1: I've fully disabled notebook-on-next until we have time/resources to work on it again.

packages/core/__tests__/epics/kernel-lifecycle-spec.js
```diff
@@ -26,21 +26,123 @@ describe("acquireKernelInfo", () => {
...
+            fileExtension: ".py",
+            pygmentsLexer: "ipython3",
+            codemirrorMode: { name: "ipython", version: 3 },
+            nbconvertExporter: "python"
+          }
+        }
+      }
+    ]);
+
+    done();
```
rgbkrk, line 145: Now with real tests (compared to what this did before)!

packages/core/src/components/notebook-app.js
```diff
@@ -354,33 +339,32 @@ const mapStateToProps = (
     );
   }
 
-  let kernel = {
-    kernelSpecName: null,
-    status: null
-  };
+  // TODO: Determine and fix things so we have one reliable place for the kernelRef
+  const kernelRef =
+    selectors.currentKernelRef(state) || ownProps.kernelRef || model.kernelRef;
```
rgbkrk, line 344: In particular, I noticed I couldn't reliably rely on `model.kernelRef` being set and didn't want to add on the reducers refactor here (yet).

packages/core/src/components/notebook-app.js
```diff
@@ -332,18 +321,14 @@ const mapStateToProps = (
...
       theme: selectors.userPreferences(state).theme,
-      lastSaved: content.lastSaved,
       cellOrder: Immutable.List(),
-      // TODO: deal with current kernel ref
-      kernelStatus: "unknown",
-      languageDisplayName: "unknown",
       transforms: ownProps.transforms || transforms,
       displayOrder: ownProps.displayOrder || displayOrder,
       codeMirrorMode: Immutable.Map({ name: "text/plain" }),
+      kernelRef: null,
```
rgbkrk, line 331: Is there a pattern for returning an alternate component if we can't create the suite of props that we need to in `mapStateToProps`?

packages/core/src/state/entities/kernel-info.js
rgbkrk: 😭 

```
> Immutable.is(Immutable.Record({a: 2})(), Immutable.Record({a: 2})())
false
```

packages/core/src/components/notebook-app.js
theengineear: Other than checking in the render method of _this_ component of the _parent_ component, I don't think there's a better pattern (if I'm understanding your question correctly).

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -878,8 +878,8 @@ function document(
       return updateCellStatus(state, action);
     case actionTypes.SET_LANGUAGE_INFO:
       return setLanguageInfo(state, action);
-    case actionTypes.SET_KERNEL_INFO:
-      return setKernelInfo(state, action);
+    case actionTypes.SET_KERNELSPEC_INFO:
```
theengineear, line 881: Nice, thanks for cleaning these all up as well.

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -86,7 +86,14 @@ export function acquireKernelInfo(
           langInfo: msg.content.language_info,
           kernelRef,
           contentRef
-        })
+        }),
+        {
+          type: "ACTUAL_KERNEL_INFO",
```
theengineear: 😁

applications/desktop/src/notebook/epics/github-publish.js
```diff
@@ -1,8 +1,12 @@
...
 import { shell } from "electron";
 
-import type { ContentRef } from "@nteract/core/src/state";
-import { selectors, actions, actionTypes } from "@nteract/core";
+import {
+  selectors,
+  actions,
+  actionTypes,
+  state as stateModule
+} from "@nteract/core";
```
theengineear, line 9: 💯 for standardizing these!

packages/core/src/components/notebook-app.js
rgbkrk: Seems like we'd have to have some wrapper component that delegates to a component based on whether we have "real" `content` and `model`. Something likely called `<NotebookApp />` as the export still, yet has its own `mapStateToProps` to basically dictate if we're showing a placeholder or the real notebook app. 🤷‍♂️ I'll get something better written up for this.

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: I should really learn better that I shouldn't have stray gunk in my commits so you don't see me fooling with data. 😄

packages/core/src/state/entities/kernel-info.js
rgbkrk: Tackled this one by the way.

packages/core/src/state/entities/kernel-info.js
rgbkrk: I did switch this to doing the proper thing with records by the way.

theengineear: Possibly fixed? https://github.com/facebook/immutable-js/commit/0c5ca8f344908e3c5b106bae7b008948e6a9f859

packages/core/__tests__/epics/websocket-kernel-spec.js
```diff
@@ -7,7 +7,7 @@ import { actions, state as stateModule, epics as coreEpics } from "../../src";
 import { toArray } from "rxjs/operators";
 
 describe("launchWebSocketKernelEpic", () => {
-  test("", async function() {
+  test("launches remote kernels", async function() {
```
theengineear, line 10: 🥇

packages/core/src/state/entities/kernel-info.js
```diff
@@ -0,0 +1,87 @@
...
+  implementationVersion: string,
+
+  // A banner of information about the kernel,
+  // which may be displayed in console environments.
+  banner: string,
+
+  // Optional: A list of dictionaries, each with keys 'text' and 'url'.
+  // These can be displayed in the help menu for the UI.
+  // SPEC: help_links
+  helpLinks?: Array<{ text: string, url: string }>,
```
theengineear: I think I left a comment on this before. But I didn't see the resolution. Should we have this as an Immutable `List`?

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -80,23 +81,35 @@ export function acquireKernelInfo(
     ofMessageType("kernel_info_reply"),
     first(),
     mergeMap(msg => {
+      const c = msg.content;
+      const l = c.language_info;
```
theengineear: Better variable names?

packages/core/src/state/entities/kernel-info.js
```diff
@@ -35,8 +47,11 @@ export type KernelInfo = {
 
   // Optional: A list of dictionaries, each with keys 'text' and 'url'.
   // These can be displayed in the help menu for the UI.
+  // NOTE: To make it easy to use the KernelInfo type as a plain object in
+  //       actions, we allow this type to be an array of plain objects or the
+  //       fully immutable variety
   // SPEC: help_links
-  helpLinks?: Array<{ text: string, url: string }>,
+  helpLinks?: Array<HelpLink> | Immutable.List<HelpLinkRecord>,
```
theengineear: Ah. OK, i see you handle it here. 🤔 Still trying to understand why we need the Array? If anything, maybe the correct thing to do is type the API response and then type the resulting state separately?

packages/core/src/reducers/core/entities/kernels.js
```diff
@@ -50,10 +50,22 @@ const byRef = (
...
+      }
+      switch (typeof codemirrorMode) {
+        case "string":
+          // already set as we want it
+          break;
+        case "object":
+          codemirrorMode = Immutable.Map(codemirrorMode);
+          break;
+        default:
+          // any other case results in falling back to language name
```
theengineear, line 66: 👌

applications/desktop/src/notebook/index.js
```diff
@@ -77,6 +77,8 @@ export default class App extends React.PureComponent<Object, Object> {
           <Styles>
             <MathJax.Context input="tex">
               <NotebookApp
+                // The desktop app always keeps the same contentRef in a browser window
```
theengineear, line 80: 👍

packages/core/src/components/notebook-app.js
```diff
@@ -285,10 +285,7 @@ type PureNotebookProps = {
   languageDisplayName?: string,
   kernelStatus?: string,
   codeMirrorMode?: string | Immutable.Map<string, *>,
-  // TODO: Once we're willing to do multi-contents views, we should require this
-  //       to be passed in
-  // TODO: Fill in more from the convo with Andrew here
-  contentRef?: ContentRef,
```
theengineear, line 291: 🎉

packages/core/src/components/notebook-app.js
theengineear: Yah. I think we *want* it to be `model.kernelRef`, but I'm guessing there's a race condition for getting that to be set? I.e., we might need to get here before the content gets returned or something?

packages/core/src/components/status-bar.js
```diff
@@ -1,14 +1,21 @@
 /* @flow */
 import React from "react";
+
+import { connect } from "react-redux";
+
 import distanceInWordsToNow from "date-fns/distance_in_words_to_now";
 
+import * as selectors from "../selectors";
+import type { AppState } from "../state";
+import type { ContentRef, KernelRef } from "../state/refs";
```
theengineear: You should be able to import all this at once from the top level state.

packages/core/src/components/status-bar.js
```diff
@@ -62,3 +69,27 @@ export default class StatusBar extends React.Component<Props> {
...
+const mapStateToProps = (
+  state: AppState,
+  ownProps: { contentRef: ContentRef, kernelRef: ?KernelRef }
+): Props => {
+  const { contentRef, kernelRef } = ownProps;
+  const content = selectors.content(state, { contentRef });
+  const kernel = kernelRef ? selectors.kernel(state, { kernelRef }) : null;
+
+  const lastSaved =
+    content && content.lastSaved ? content.lastSaved : undefined;
```
theengineear: Minor, but it feels weird to set it to undefined here. Can we use `null`?

packages/core/src/components/status-bar.js
```diff
@@ -62,3 +69,27 @@ export default class StatusBar extends React.Component<Props> {
...
+  const kernelStatus =
+    kernel && kernel.status ? kernel.status : "not connected";
+  const kernelSpecDisplayName =
+    kernel && kernel.kernelSpecName ? kernel.kernelSpecName : "kernel";
```
theengineear: 😁 i like the generic default 👍 .

packages/core/src/components/status-bar.js
```diff
@@ -80,10 +82,18 @@ const mapStateToProps = (
...
-  const kernelStatus =
-    kernel && kernel.status ? kernel.status : "not connected";
-  const kernelSpecDisplayName =
-    kernel && kernel.kernelSpecName ? kernel.kernelSpecName : "kernel";
+
+  const kernelStatus = kernel && kernel.status ? kernel.status : NOT_CONNECTED;
+
+  let kernelSpecDisplayName = " ";
+  if (kernelStatus === NOT_CONNECTED) {
+    kernelSpecDisplayName = "no kernel";
```
theengineear: even better!

packages/core/src/state/entities/kernel-info.js
```diff
@@ -94,9 +94,27 @@ export type KernelInfo = {
 
 export type KernelInfoRecord = Immutable.RecordOf<KernelInfo>;
 
+// NOTE: This is not exported as the "makeXRecord" for KernelInfo so that we
+//       don't have nullables and can rely on the proper types. The developer is
+//       required to pass a fully instantiated KernelInfo into
+//       makeKernelInfoRecord below.
```
theengineear: 👍 cool. i like it.

packages/core/src/state/entities/kernel-info.js
rgbkrk: This is at the level of needing payload vs. record typed appropriately. I suppose if I made these generics instead that would help.

packages/core/src/components/notebook-app.js
rgbkrk: 💥

packages/core/src/components/status-bar.js
rgbkrk: so true

PR #2722, opened 2018-04-01
Parameter cells

### LINKED ISSUE #2557: Show cells that are set as parameter cells from papermill
Starter mockup:

image

rgbkrk:
Closes #2557.

screen shot 2018-03-31 at 8 08 14 pm

Users can toggle parameter cells for papermill usage and it shows a banner at the top of the cell.

Not quite where I wanted the styling, though I also wanted to just set up the state bits for this. Thoughts?

### DIFF
--- a/packages/core/src/actionTypes.js
+++ b/packages/core/src/actionTypes.js
@@ -133,7 +133,8 @@ export type ToggleTagInCell = {
   type: "CORE/TOGGLE_TAG_IN_CELL",
   payload: {
     id: CellID,
-    tag: string
+    tag: string,
+    contentRef: ContentRef
   }
 };
 
--- a/packages/core/src/actionTypes.js
+++ b/packages/core/src/actionTypes.js
@@ -126,6 +126,17 @@ export type ChangeFilenameAction = {
   }
 };
 
+export const TOGGLE_TAG_IN_CELL = "CORE/TOGGLE_TAG_IN_CELL";
+export type ToggleTagInCell = {
+  // expectation is that if a tag doesn't exist, it will set it
+  // if the tag is already in the collection of tags it will delete it
+  type: "CORE/TOGGLE_TAG_IN_CELL",
+  payload: {
+    id: CellID,
+    tag: string
+  }
+};
+
 export const SET_IN_CELL = "SET_IN_CELL";
 export type SetInCell<T> = {
   type: "SET_IN_CELL",
--- a/packages/core/src/components/notebook-app.js
+++ b/packages/core/src/components/notebook-app.js
@@ -29,6 +29,7 @@ import { displayOrder, transforms } from "@nteract/transforms";
 
 type AnyCellProps = {
   id: string,
+  tags: Immutable.Set<string>,
   contentRef: ContentRef,
   cellType: "markdown" | "code" | "raw",
   theme: string,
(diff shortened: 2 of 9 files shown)


rgbkrk:
I think I probably shouldn't have changed the font-size here. I brought it down so it wasn't so "loud". 🤷‍♂️

@mseal, @mpacer, @willingc - thoughts?

packages/core/src/actionTypes.js
```diff
@@ -125,6 +125,17 @@ export type ChangeFilenameAction = {
...
 };
 
+export const TOGGLE_TAG_IN_CELL = "CORE/TOGGLE_TAG_IN_CELL";
+export type ToggleTagInCell = {
+  // expectation is that if a tag doesn't exist, it will set it
+  // if the tag is already in the collection of tags it will delete it
+  type: "CORE/TOGGLE_TAG_IN_CELL",
+  payload: {
+    id: CellID,
+    tag: string
```
theengineear: Are we allowing arbitrary tags? I.e., anything goes?

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -23,47 +23,6 @@ import { createContentRef } from "../../../state/refs";
...
-  FocusPreviousCell,
-  SetKernelInfo,
-  SetLanguageInfo,
-  UpdateCellStatus,
-  ToggleCellInputVisibility,
-  ToggleCellOutputVisibility,
-  SetInCell,
-  SendExecuteRequest
-} from "../../../actionTypes";
-
```
theengineear, line 66: 🙇

packages/core/src/reducers/core/entities/contents.js
```diff
@@ -160,6 +160,25 @@ function clearOutputs(state: NotebookModel, action: actionTypes.ClearOutputs) {
...
+  state: NotebookModel,
+  action: actionTypes.ToggleTagInCell
+): NotebookModel {
+  const { id, tag } = action.payload;
+
+  return state.updateIn(
+    ["notebook", "cellMap", id, "metadata", "tags"],
+    // When tags haven't been initialized, we set them up as a new set
+    Immutable.Set(),
+    tags => {
```
theengineear: Could shorten this one. `tags => tags.has(tag) ? tags.remove(tag) : tags.add(tag);`

packages/core/src/components/notebook-app.js
```diff
@@ -246,6 +269,15 @@ class AnyCell extends React.PureComponent<AnyCellProps, *> {
     return (
       <HijackScroll focused={cellFocused} onClick={selectCell}>
         <Cell isSelected={cellFocused}>
+          {/* The following banners come from when papermill's acknowledged
+              cell.metadata.tags are set
+          */}
+          {tags.has("parameters") ? (
+            <CellBanner>Papermill - Parametrized</CellBanner>
```
theengineear, line 276: Part of me is questioning wether it really makes sense to associate some tag `parameters` here with `Papermill`. The cell just has a tag, and we're currently using that tag in `Papermill`. However, I don't know if this code should really know about `Papermill`? Thoughts?

packages/core/src/actionTypes.js
rgbkrk: In the Jupyter [notebook format](http://nbformat.readthedocs.io/en/latest/format_description.html#cell-metadata) (within the last year), `tags` were introduced as an `Array<string>` on cell metadata.

I'm taking the stance that we'll only take unique tags so that in memory we can represent them as an `Immutable.Set`.

packages/core/src/components/notebook-app.js
rgbkrk: This is [what papermill expects for cells tagged for parameters](https://github.com/nteract/papermill/blob/90f5519804ea23d22d688f0032e244630a5e6633/papermill/execute.py#L240).

rgbkrk: Oh I see. Yeah this doesn't have to say "Papermill" to the user. I just put that in the mock originallly and then put it in here.

Issue #2324, opened 2018-01-04
Turn dataresource transform more into zeppelin style


rgbkrk:
Zeppelin has a pretty good interface for toggling between table and chart types:

<img width="1424" alt="screen shot 2018-01-04 at 2 16 35 pm" src="https://user-images.githubusercontent.com/836375/34586776-4c6362fe-f15a-11e7-9d96-8b19d0f8819c.png">

<img width="1424" alt="screen shot 2018-01-04 at 2 16 21 pm" src="https://user-images.githubusercontent.com/836375/34586720-1319e0ae-f15a-11e7-9106-6d4a96628c2d.png">

As well as a neat way to pick your "Keys" (Axes), "Groups", and "Values" on a chart

<img width="1435" alt="screen shot 2018-01-04 at 2 17 01 pm" src="https://user-images.githubusercontent.com/836375/34586721-13318e48-f15a-11e7-93ac-60574d8a7a66.png">

It's not the end all-be-all, it's a good starter though on top of what we have. We could persist the settings for the chart to metadata on the output type as well.

Some other thoughts, beyond what zeppelin is currently doing that we should think about:

* How do you filter?
* Look again at [voyager](http://vega.github.io/voyager/)
* Lightweight pivot tables as a visualization

rgbkrk:
/cc @thomasballinger @theengineear 

Also, I just noticed zeppelin has their own little suite of viz "plugins", I'm tempted to check out more of their approach and how well users are taking advantage of it.

https://zeppelin.apache.org/docs/0.7.0/development/writingzeppelinvisualization.html#how-it-works

rgbkrk:
Silly me, I forgot to paste the link when posting.

theengineear:
Yup. I think we can definitely improve the user experience here. I'm happy with making updates to this as long as we make sure not to overstep. For example, if we allow too much handy filtering in the UI for the visualizations, we may end up conditioning users to just `select *` all the time from tables.

One alternative is to try to nudge users into narrowing the number of columns they request in the first place. This is a super-common problem in viz. Another manifestation is creating tooling to handle down-sampling huge sets of data into a visualization. Arguably, the sampling should be done _prior_ to visualization in many cases.

Again, I think adding more functionality is definitely appropriate here, but let's make sure we consider how any features we add will condition users to do analysis.

emeeks:
[...]

Issue #2321, opened 2018-01-04
[dataresource] Create or acquire better icons for scatter and table views


rgbkrk:
From https://github.com/nteract/nteract/pull/2273, we need actual icons for scatter and table. Right now we're using a little heartbeat and database icon.

rgbkrk:
Done in #2732

PR #2732, opened 2018-04-02
Improve the chart rendering for the tabular data payloads

emeeks:

Here's a rough first shot at getting the same functionality seen in the Plotly integration but from [Semiotic](https://github.com/emeeks/semiotic). It's rough, all the code is in one file like the Plotly code was, and the UI elements are garbage, and also there's more functionality in the charts, so any feedback from any perspective (code, dataviz, structure, design) is welcome.


rgbkrk:
This is the first time we've needed any amount of options/selection UI so it's totally fine this is (self-described as) rough. The main focus is the visualization from rough user-provided data and this is looking pretty darn great from an interactive standpoint!

What I think would be useful, as a project, is if we provided some UI components for options/selection that work well inside the app. Happy to help carve them out in here or as a new PR/new package. I know @theengineear is interested too.

I'm going to be on the road tomorrow so I think what I'll do is make sure I have this cloned / installed so I can try it out.

rgbkrk:
Related issues for this component:

* Most similar issue, bringing us closer to Zeppelin's data viewer: https://github.com/nteract/nteract/issues/2324
* Improving the icons https://github.com/nteract/nteract/issues/2321

rgbkrk:
This is working pretty well inside the showcase!

screen shot 2018-04-03 at 10 32 33 am

On Desktop (and likely the jupyter extension), I'm getting errors:

screen shot 2018-04-03 at 10 33 38 am

rgbkrk:
That might be because we have Immutable.JS structures going in and we're not converting them (?). I'll tinker with it to figure it out.

emeeks:
Without being able to dig into it, it looks like the component isn't getting a `height` prop because it's the SVG itself that isn't drawing with height. There could be other problems but I'd look at that first.

packages/transform-dataresource/src/index.js
```diff
@@ -40,37 +69,389 @@ const DataResourceTransformGrid = ({
...
+    label: "Stacked Area Chart (Percent)"
+  },
+  {
+    type: "bumparea",
+    label: "Ranked Area Chart"
+  }
+];
 
-const plotViewTypes = Object.keys(viewTypeToPlotlyTraceProps);
+const metricDimSelector = (values, selectionFunction, title) => (
```
rgbkrk: Is this a semiotic pattern where these function signatures are not a props bag but it returns a React.Element?

rgbkrk:
Ok I've snuck in a default height property for semiotic to work with. Still rendering issues for some of the chart types, I'll keep at it. I'd really like to start breaking components up into separate files here.

rgbkrk:
It's amusing how much we ported over from the `infer` code from frictionless tableschema.

rgbkrk:
This is a bit of an aside to this component, likely essential for maintenance though.

We've got some decisions to make about the table schema inference. I recall we forked it internally into here to reduce dependencies originally. It's not currently flow typed (and doesn't match our current style well). The original library we built from is now using an asynchronous interface.

Some thoughts / direction:

* We tear all the inference code out, require that a `schema` is passed in
* We re-write the inference code ourselves
* We move over to the current version of https://github.com/frictionlessdata/tableschema-js, which means we probably need to do it at the top level component, relying on setting `state` internally to the component and potentially memoizing

At the very least, I'd like to group the inference code in a separate module

rgbkrk:
@emeeks is this getting close enough that we can merge the first pass here and then iterate with new PRs? I believe from our in-person work that you also upgraded the semiotic version.

rgbkrk:
Let's keep this moving on forward, thanks @emeeks!

PR #2733, opened 2018-04-03
Fix run spawn:debug

BenRussert:
Noticed this wasn't working when run from the monorepo root the way `npm run spawn` does.


rgbkrk:
Nice!

I've been running

```
DEBUG=true npm run spawn
```

This is super handy to have a shortened command at the root, thank you!

PR #2742, opened 2018-04-04
[WIP] Notebook records

rgbkrk:
Similar to the core records refactor, I wanted to take a second look at the notebook structures. This is definitely not ready for primetime as use of it would require refactoring records, selectors, and components. The aim is a structure like this:

```yaml
entities:
  outputs:
    byRef:
      [ref]: DisplayDataRecord
  cells:
    byRef:
      [ref]: CodeCell
  models:
    byRef:
      [ref]: Notebook
```

TODO:

* [ ] Create function to read from disk notebooks easily (like current commutable)
  * [ ] v3
  * [ ] v4
* [ ] Incorporate more of the notebook model as is used in `@nteract/core` as part of the main notebook model
* [ ] Ensure that only what should be written to disk/JSON happens with our toJS


packages/commutable/src/records/cells.js
```diff
@@ -0,0 +1,70 @@
...
+import type { OutputRef, CellRef } from "./refs";
+
+export type CellType = "markdown" | "code" | "raw";
+
+export type CodeCellProps = {
+  cellType: "code",
+  source: string,
+  executionCount: number | null,
+  outputs: Immutable.List<OutputRef>,
+  // TODO: Some of the metadata should be tightly specced
```
holdenk, line 16: So maybe this is a bit unrelated, but looking at https://vatlab.github.io/blog/post/sos-notebook/ & other commercial solutions in the same space -- would it make sense to have a metadata field for which lexer/language to use (e.g. allow overriding of traditionally notebook level fields)?

rgbkrk: SoS looks pretty cool! Integrating with it is certainly possible, though it might be a little bit too much for me (personally) to take on any time soon though.

Side note: Users can use magics to reach other languages `%%R`, `%%scala`, if they have them installed / available. Users can use the single line variant `%R` or the whole cell variant `%%R`.  There are interfaces to pass data between these, there just isn't a UI control to toggle languages like is shown in SoS. It goes something like this:

```
In [14]: Z = np.array([1,4,5,10]) # Python land

# Send Z over to an R session

In [15]: %R -i Z mean(Z)
Out[15]: array([ 5.])

# Bring an R result over to Python

In [16]: %R -o W W=Z*mean(Z)
Out[16]: array([  5.,  20.,  25.,  50.])

# 🎉

In [17]: W
Out[17]: array([  5.,  20.,  25.,  50.])
```

Granted I don't think that's the best way (or the only way) to pass data between processes. It's certainly possible though.

rgbkrk:
Tabling this one for now, I need to do this in a less drastic way so we can iterate on it.

PR #2747, opened 2018-04-10
switch from SET_NOTEBOOK to FETCH_CONTENT_*

rgbkrk:
This fully switches us over to the fetch contents pathway, no more "hidden" SET_NOTEBOOK action. I'm really happy to be rid of the Immutable.js object inside of actions (this was the only one I think).


rgbkrk:
Oh fun, actually I found out that `MERGE_CONFIG` was using a `Map` (yet not the Immutable kind). That may need a touchup.

PR #2753, opened 2018-04-12
allow configuring a remote for js assets and GA

rgbkrk:
This allows for simple configuration of:

* Setting a Google Analytics tracking code
* Pulling the JavaScript for the nteract web app from a remote server

For testing I deployed https://yup-jyusaxdvps.now.sh/ which just has

```
.
└── nteract
    └── static
        └── lib
            ├── app.js
            ├── app.js.map
            ├── vendor.js
            └── vendor.js.map
```

/cc @yuvipanda 

Admittedly, I was only going to do asset url here then realized I'd have an easier time figuring out configuration with only google analytics at first.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/config.py
+++ b/applications/jupyter-extension/nteract_on_jupyter/config.py
@@ -1,6 +1,7 @@
 from traitlets import HasTraits, Unicode, Bool
+from traitlets.config import Configurable
 
-class Config(HasTraits):
+class NteractConfig(Configurable):
     """The nteract application configuration object
     """
     name = Unicode('nteract',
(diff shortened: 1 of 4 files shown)


yuvipanda:
The traitlet code looks good to me. However, I always feel extremely uncomfortable when Google Analytics code insertion is a core part of an open source project, rather than an extension. Here it also seems to be enabled by default, which makes me doubly uncomfortable. However, I'm only peripherally involved so my discomfort should not guide your decisions :)

rgbkrk:
> Here it also seems to be enabled by default, which makes me doubly uncomfortable.

Oops! This should be an opt-in, where the script would only get written when the google analytics code is defined.

> However, I always feel extremely uncomfortable when Google Analytics code insertion is a core part of an open source project

I wanted to make this easier for others to use the work I'll end up doing around hooking up actions to google analytics events. I could just provide it as redux middleware though and then they'd have to get the `index.html` instantiated if they want to use it. It's basically build your own nteract_on_jupyter at that point though.

yuvipanda:
LGTM and GA is not on by default!

PR #2759, opened 2018-04-13
Use codemirror to display file contents

yuvipanda:
First serious React / Redux patch ever!

- Picks appropriate mimetype for codemirror
- Updates redux store state whenever codemirror contents changes
- Automatically saves contents with every keystroke

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
@@ -40,6 +40,11 @@ export class TextFile extends React.PureComponent<TextFileProps, null> {
         contentRef: this.props.contentRef
       })
     );
+    this.props.dispatch(
+      actions.save({
+        contentRef: this.props.contentRef
+      })
+    );
   }
   render() {
     return (
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
@@ -15,12 +15,17 @@ type FileContentRecord = stateModule.FileContentRecord;
 import { connect } from "react-redux";
 
 type FileProps = {
+  content: FileContentRecord,
+  contentRef: ContentRef
+};
+
+type TextFileProps = {
   dispatch: Dispatch<*>,
   content: FileContentRecord,
   contentRef: ContentRef
 };
 
-export class TextFile extends React.PureComponent<FileProps, null> {
+export class TextFile extends React.PureComponent<TextFileProps, null> {
   static handles(mimetype: string) {
     return (
       mimetype.startsWith("text/") ||
@@ -41,6 +46,10 @@ export class TextFile extends React.PureComponent<FileProps, null> {
       <CodeMirrorEditor
         cellFocused
         editorFocused
+        channels
+        kernelStatus={"not connected"}
+        tip
+        completion
         theme="light"
         // TODO: This is the notebook implementation leaking into the editor
         //       component. It shouldn't be here, I won't refactor it as part
@@ -64,7 +73,7 @@ export class TextFile extends React.PureComponent<FileProps, null> {
         }}
         value={this.props.content.model.text}
         onChange={this.handleChange.bind(this)}
-        contentRef={this.props.contentRefk}
+        contentRef={this.props.contentRef}
       />
     );
   }
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
@@ -70,9 +70,6 @@ export class TextFile extends React.PureComponent<TextFileProps, null> {
         //       completion should be enabled
         options={{
           lineNumbers: true,
-          extraKeys: {
-            "Ctrl-Space": "autocomplete"
-          },
           cursorBlinkRate: 0,
           mode: this.props.content.mimetype
         }}
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
@@ -15,7 +15,9 @@ type FileContentRecord = stateModule.FileContentRecord;
 import { connect } from "react-redux";
 
 type FileProps = {
-  content: FileContentRecord
+  dispatch: Dispatch<*>,
+  content: FileContentRecord,
+  contentRef: ContentRef
 };
 
 export class TextFile extends React.PureComponent<FileProps, null> {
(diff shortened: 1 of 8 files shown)


yuvipanda:
I've fixed all the flow errors I could, the last one eludes me.

yuvipanda:
Cell Updates are performed in the codeMirrorEditor component, while file content updates are performed elsewhere. I *think* the right thing to do is to move the current `onChange` and `onFocusChange` events in `CodeMirrorEditor` out of there to wherever notebook code is? But that doesn't need to be in this PR...

yuvipanda:
I think a better autosave would be something like 'after X keystrokes', and provide some indicators of saving being performed. Going to avoid buttons for now.

rgbkrk:
Woo, I can check this out now. Finally done with other triage.

rgbkrk:
> First serious React / Redux patch ever!

🎉 Thanks for picking this all up!

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
```diff
@@ -14,26 +15,71 @@ type FileContentRecord = stateModule.FileContentRecord;
...
+        focusBelow={() => {}}
+        // END TODO for notebook leakage
+        // TODO: kernelStatus should be allowed to be null or undefined,
+        //       resulting in thought of as either idle or not connected by
+        //       default. This is primarily used for determining if code
+        //       completion should be enabled
+        options={{
+          lineNumbers: true,
+          extraKeys: {
+            "Ctrl-Space": "autocomplete"
```
rgbkrk: Probably could delete this since we don't have a kernel here (yet).

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
```diff
@@ -14,26 +15,71 @@ type FileContentRecord = stateModule.FileContentRecord;
...
+          },
+          cursorBlinkRate: 0,
+          mode: this.props.content.mimetype
```
rgbkrk: Oh, will codemirror just recognize this based on the mimetype?

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
```diff
@@ -14,26 +15,71 @@ type FileContentRecord = stateModule.FileContentRecord;
...
+      actions.updateFileText({
+        text: source,
+        contentRef: this.props.contentRef
+      })
+    );
+    this.props.dispatch(
+      actions.save({
+        contentRef: this.props.contentRef
+      })
+    );
```
rgbkrk: This works for now. I'd rather save be separate, especially since this is going to trigger a save on _every single keystroke_.

packages/core/src/actionTypes.js
```diff
@@ -626,6 +626,15 @@ export type NewNotebook = {
...
 };
 
+export const UPDATE_FILE_TEXT = "UPDATE_FILE_TEXT";
+export type UpdateFileText = {
+  type: "UPDATE_FILE_TEXT",
+  payload: {
+    text: string,
+    contentRef: ContentRef
+  }
+};
```
rgbkrk, line 636: You figured this stuff out quick!

packages/core/src/actions.js
```diff
@@ -807,6 +807,19 @@ export function newNotebook(payload: {
...
+  text: string
+}): actionTypes.UpdateFileText {
+  return {
+    type: actionTypes.UPDATE_FILE_TEXT,
+    payload: {
+      contentRef: payload.contentRef,
+      text: payload.text
+    }
+  };
+}
```
rgbkrk, line 821: Easy peasy. 😄 You did exactly what I expected to be done for this.

packages/core/src/epics/contents.js
```diff
@@ -126,21 +123,28 @@ export function saveContentEpic(
...
         // TODO: this default version should probably not be here.
         const appVersion = selectors.appVersion(state) || "0.0.0-beta";
 
-        // contents API takes notebook as raw JSON whereas downloading takes
-        // a string
-        const notebook = toJS(
-          notebookModel.setIn(["metadata", "nteract", "version"], appVersion)
-        );
+        let serializedData;
+        if (content.type == "notebook") {
```
rgbkrk: Make sure to use `===` as standard practice when comparing.

packages/core/src/epics/contents.js
```diff
@@ -126,21 +123,28 @@ export function saveContentEpic(
...
+
         // TODO: this default version should probably not be here.
         const appVersion = selectors.appVersion(state) || "0.0.0-beta";
 
-        // contents API takes notebook as raw JSON whereas downloading takes
-        // a string
-        const notebook = toJS(
-          notebookModel.setIn(["metadata", "nteract", "version"], appVersion)
-        );
+        let serializedData;
```
rgbkrk: Just creating the serialized data and then passing on to contents is plenty smart here. 👍

packages/core/src/epics/contents.js
```diff
@@ -126,21 +123,28 @@ export function saveContentEpic(
...
+          // a string
+          serializedData = toJS(
+            content.model.notebook.setIn(
+              ["metadata", "nteract", "version"],
+              appVersion
+            )
+          );
+        } else if (content.type == "file") {
+          serializedData = content.model.text;
+        }
```
rgbkrk: Super!

packages/core/src/epics/contents.js
```diff
@@ -151,8 +155,9 @@ export function saveContentEpic(
             const serverConfig = selectors.serverConfig(host);
 
             const model = {
-              content: notebook,
-              type: "notebook"
+              content: serializedData,
+              type: content.type,
+              format: "text"
```
rgbkrk: This should only be "text" for the `content.type === "file"`. Notebooks are sent over as a JSON payload.

packages/core/src/reducers/core/entities/contents/file.js
```diff
@@ -0,0 +1,30 @@
...
+
+export function file(state: FileModelRecord, action: FileAction) {
+  switch (action.type) {
+    case actionTypes.UPDATE_FILE_TEXT:
+      return updateFileText(state, action);
+    default:
+      (action: empty);
+      return state;
+  }
+}
```
rgbkrk, line 30: For a first React/Redux PR, complete with having to work with epics, this is so well done.

have mercy

packages/core/src/reducers/core/entities/contents/index.js
```diff
@@ -195,6 +196,11 @@ const byRef = (state = Immutable.Map(), action) => {
       const model = state.getIn(path);
       return state.setIn(path, notebook(model, action));
     }
+    case actionTypes.UPDATE_FILE_TEXT: {
+      const path = [action.payload.contentRef, "model"];
+      const model = state.getIn(path);
+      return state.setIn(path, file(model, action));
+    }
```
rgbkrk: 🎉

rgbkrk (approved):

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
yuvipanda: I want it to be delayed - something like '5s after the last keystroke, unless there is another keystroke in the interim'. I know how to do this in normal JS, but here I am *guessing* I'll need to make an Epic?

I guess it'll be simpler to just make a button for this PR :)

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
yuvipanda: Yes, it will!

packages/core/src/actionTypes.js
yuvipanda: :D Thank you! The Redux docs are pretty good.

packages/core/src/actions.js
yuvipanda: :tada:

yuvipanda:
I think the TODO left for me is to move saving to a button.

applications/jupyter-extension/nteract_on_jupyter/app/contents/file.js
rgbkrk: There are ways to debounce it locally here. The better option is definitely in an epic where we can debounce or throttle the same save action for the same document.

I like the button approach for now. I'll target autosaving for this release as well.

rgbkrk:
Extra commentary now that I've played with this PR -- we should make the height of the codemirror element match that of the page like we do on play. Otherwise small files look a little silly:

screen shot 2018-04-13 at 2 30 57 pm

We also want it a bit bigger so that empty files are able to be edited. 😅

rgbkrk:
Since this passes and is already super good, we can make new PRs for improving this further. Thanks again @yuvipanda!

PR #2762, opened 2018-04-13
craft better messaging for stale and locked issues

rgbkrk:
Since our default stale message (that I copied from the app) is [not all that friendly and doesn't state our intent well](https://twitter.com/rgbkrk/status/984852813128712192), I reworded it.

/cc @jaredly @betatim

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -10,8 +10,22 @@ exemptLabels:
 staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
-  This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  This issue hasn't had any activity on it in the last 90 days. Unfortunately we don't
+  get around to dealing with every issue that is opened. Instead of leaving issues
+  open we're seeking to be transparent by closing issues that aren't being prioritized.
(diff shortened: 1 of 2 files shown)


.github/stale.yml
```diff
@@ -10,8 +10,22 @@ exemptLabels:
 staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
-  This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  This issue hasn't had any activity on it in the last 90 days. While we would
+  love to solve everything, we don't have the bandwidth. 😩 Instead of leaving issues
```
betatim: Maybe say "... the last 90 days. Unfortunately we don't get around to dealing with every issue that is opened. Instead of leaving issues open we're...".

A pet peeve of mine is the "Sorry for the late email reply...", to which I think "don't be sorry, just reply faster next time if you indeed put this off for no good reason or if you had a good reason then don't be sorry."

jaredly:
Love it!

betatim:
LGTM if the robots are happy too.

(I for one welcome our new locking-priority-setting-testing overlords.) ;)

rgbkrk:
CI bots have approved our social bots. ⚡️

PR #2763, opened 2018-04-13
auto save files and notebooks for the user

rgbkrk:
Saves files and notebooks every 7 seconds.

NOTE: This is only provided for the jupyter extension at the moment, not the desktop app.

TODO and TODONE:

* [x] Create an autosaving epic
* [x] Don't save over files that have been saved elsewhere
* [x] Ensure that saved files get updated last_modified
* [x] Remove current save on file change behavior in the editor

### DIFF
--- a/packages/core/src/epics/contents.js
+++ b/packages/core/src/epics/contents.js
@@ -247,20 +247,25 @@ export function saveContentEpic(
                 const diskDate = new Date(model.last_modified);
                 const inMemoryDate = content.lastSaved
                   ? new Date(content.lastSaved)
-                  : // FIXME: I'm unsure if we don't have a date if we should default
+                  : // FIXME: I'm unsure if we don't have a date if we should default to the disk date
                     diskDate;
 
-                if (Math.abs(diskDate - inMemoryDate) > 300) {
-                  // TODO: Determine our course of action or at the very least introduce
-                  throw new Error("open in another tab maybe");
+                if (Math.abs(diskDate - inMemoryDate) > 600) {
+                  return of(
+                    actions.saveFailed({
+                      error: new Error("open in another tab possibly..."),
+                      contentRef: action.payload.contentRef
+                    })
+                  );
                 }
 
                 return contents.save(serverConfig, filepath, saveModel).pipe(
-                  mapTo(
-                    actions.saveFulfilled({
-                      contentRef: action.payload.contentRef
-                    })
-                  ),
+                  map(xhr => {
+                    return actions.saveFulfilled({
+                      contentRef: action.payload.contentRef,
+                      model: xhr.response
+                    });
+                  }),
                   catchError((error: Error) =>
                     of(
                       actions.saveFailed({
--- a/packages/core/src/epics/contents.js
+++ b/packages/core/src/epics/contents.js
@@ -100,7 +100,7 @@ export function autoSaveCurrentContentEpic(
   store: Store<stateModule.AppState, *>
 ) {
   // Save every seven seconds, regardless of contentType
-  return interval(7000).pipe(
+  return interval(3000).pipe(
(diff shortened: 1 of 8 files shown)


packages/core/src/epics/contents.js
```diff
@@ -92,6 +95,35 @@ export function downloadString(
   FileSaver.saveAs(blob, filename);
 }
 
+export function autoSaveCurrentContentEpic(
+  action$: ActionsObservable<Action>,
+  store: Store<stateModule.AppState, *>
+) {
+  // Save every seven seconds, regardless of contentType
+  return interval(7000).pipe(
```
yuvipanda: Is there any way to make this configurable?

yuvipanda:
I see SAVE events every 7s. I also see UPDATE_FILE_TEXT everytime I type. However, the contents don't actually seem to get saved.

packages/core/src/epics/contents.js
rgbkrk: Yes\*

\* I'd really like to auto-adjust this according to how long it last took to save the current document. I'm a big proponent of not making things too configurable on the user side and instead wanting to rely on finding the right balance of UX (and in this case ops). If it's configurable for the sake of the server operator though, I can be swayed.

rgbkrk:
I think after I introduced the last modified checks the saving stopped happening. I need to fully flesh this out and test it. Marking as WIP.

rgbkrk:
Rebasing and coming back to working on this feature.

packages/core/src/epics/contents.js
```diff
@@ -194,19 +231,46 @@ export function saveContentEpic(
...
-                  actions.saveFailed({
-                    error,
-                    contentRef: action.payload.contentRef
-                  })
-                )
-              )
+            // Check to see if the file was modified since the last time we saved
+            // TODO: Determine how we handle what to do
+            // Don't bother doing this if the file is new(?)
+            contents.get(serverConfig, filepath, { content: 0 }).pipe(
```
rgbkrk: Forgot to `return` here, and the `default` case was returning `empty()` so kindly for us. 🙃

rgbkrk:
Finished off making autosave based on the _actual_ last modified timestamp. 🎉 

Next piece is tracking when we keep auto save on or off. Originally @yuvipanda and I thought about beginning autosaving once an initial autosave action was been kicked off, just like in the redux-observable 1.0.0 example:

```js
const autoSaveEpic = (action$, state$) =>
  action$.pipe(
    ofType(AUTO_SAVE_ENABLE),
    // The use of an exhaustMap here is basically a flipped switch to say
    // "we are in autosaving mode, don't re-enable it again until requested"
    exhaustMap(() =>
      state$.pipe(
        // Note that our setup is a little more complex because we have multiple documents (potentially)
        // and are autosaving per contentRef
        pluck('content'),
        distinctUntilChanged(),
        throttleTime(500, { leading: false, trailing: true }),
        concatMap(content =>
          saveContent(content).pipe(
            map(() => saveContentFulfilled()),
            catchError(e => of(saveContentRejected(e)))
          )
        ),
        takeUntil(action$.pipe(
          ofType(AUTO_SAVE_DISABLE)
        ))
      )
    )
  );
```

Instead we'd like to store it in the state tree for each content element, and update the state for that field.

rgbkrk:
Something deeply needed here is a last saved (and saving) indicator which I think belongs in the upper right of the title bar. I'll make a new PR for that.

rgbkrk:
Since this is pretty workable at the moment, I'm going to open new issues for things that I want to refine for it.

* [ ] Propagate errors when unable to autosave due to another tab being opened. The way to override it is by literal saving.
* [ ] Visual indicator for saving and last saved

PR #2775, opened 2018-04-17
extract presentational components to new lib

rgbkrk:
Problem being solved -- simple notebook rendering pages shouldn't have to opt in to all the dependencies that the full notebook app has. To kick this off, some of the components from `@nteract/core` have been moved into a new package called `@nteract/presentational-components`.

* [X] Re-package `presentational` from inside of `core`
* [X] Re-package the styles provider from inside of core into presentational
* [X] Re-package themes from inside of core into presentational
* [X] Re-export the original exports that core was providing from this new package
* [x] Migrate notebook preview to relying directly on the presentational components
* [x] Migrate showcase to use less core, more presentational
* [x] Migrate commuter to use presentational

cc @tlyon3


packages/notebook-preview/package.json
```diff
@@ -20,11 +20,10 @@
   },
   "dependencies": {
     "@nteract/commutable": "^4.0.1",
-    "@nteract/core": "^5.0.0",
     "@nteract/display-area": "^4.1.0",
-    "@nteract/editor": "^7.0.3",
```
rgbkrk, line 25: @tlyon3 turns out there was likely a simpler fix of just removing this unused dep on editor here. At any rate, should be much trimmer now.

PR #2783, opened 2018-04-20
retry websocket connection on Event Errors

rgbkrk:
Earlier today I paired with @jayphelps to make our websocket subject do automatic reconnection due to certain events.

The `Events` that will lead to websocket reconnection now include

* CloseEvents that aren't from us `.complete()`ing the subject
* Network Errors

However this doesn't include errors like deserialization issues internal to the
RxJS WebsocketSubject.

There's a little bit of cleanup left in this. I'm also going to provide some more commentary in this PR about using the heartbeat effectively here.

### DIFF
--- a/packages/rx-jupyter/src/kernels.js
+++ b/packages/rx-jupyter/src/kernels.js
@@ -9,6 +9,8 @@ import { Subscriber } from "rxjs/Subscriber";
 
 import { createAJAXSettings } from "./base";
 
+import { retryWhen, tap, delay, share } from "rxjs/operators";
+
 const urljoin = require("url-join");
 const URLSearchParams = require("url-search-params");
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Jay and I also talked about an approach that would automatically check for the heartbeat on a kernel connection and attempt to reconnect if the heartbeat isn't happening. The implementation we ended up at was something like this:

```js
const originalWebsocket = webSocket(
  formWebSocketURL(serverConfig, kernelID, sessionID)
);

const wsSubject = merge(
  originalWebsocket,
  timer(5000).pipe(
    exhaustMap(() => {
      const waitTime = 10000;

      originalWebsocket.next(heartBeatMessage);

      return throwError(`timed out after ${waitTime} ms`).pipe(
        timeout(waitTime),
        // Effectively, if we get _any_ message we assume that the connection is live
        takeUntil(originalWebsocket)
      );
    })
    // Somehow this needs to cause a reconnection
  )
);
```

rgbkrk:
This is working pretty well, even on `play`. There are some pieces we need to refine in terms of what users see (they should know we're reconnecting and that they're disconnected). I'm not sure how we should expose those from this subject.

rgbkrk:
This also seems to miss the case of when the websocket gets stuck in a pending state.

screen shot 2018-04-20 at 3 32 36 pm

rgbkrk:
I'm going to merge this as something that's a bit better, even though it doesn't cover all the cases we'll need. We'll keep on improving from here!

PR #2784, opened 2018-04-21
switch from github to @octokit/rest

rgbkrk:
In place of #2771, switch to the supported `@octokit/rest` library that has the same interface.


rgbkrk:
This doesn't quite work so I'm going to go ahead with just refactoring the github publishing to use Rx directly and play a little better with how our state and event flow works.

rgbkrk:
Ohhhh interesting you can _no longer_ post anonymous gists. That simplifies the code here and helps me understand what was going on...

rgbkrk:
Closing in favor of #2785

PR #2788, opened 2018-04-23
upgrade vega embed and tear out d3

rgbkrk:
Since we're no longer using d3 directly in our codebase, I've taken the liberty of ripping out the direct dependency while also upgrading vega-embed (which does use it under the covers).

We were facing an issue with some vega charts not rendering right. I finally noticed we were on a release candidate of the newer vega embed. This fixes up the examples that weren't working well. I'll plan to get a release out for this in the next few days.

Fixes #2694

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,6 @@
     "codemirror": "5.37.0",
     "cross-env": "^5.1.3",
     "css-loader": "^0.28.10",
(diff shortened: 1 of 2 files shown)


rgbkrk:
cc @nikitakit

PR #2779, opened 2018-04-18
Add a nteract/monaco-editor component

yuvipanda:
And hack at things until I have it rendering

PR #2798, opened 2018-04-27
override the webpack public path to use built dist

rgbkrk:
Originally I was just going after using `import()` to dynamically load transforms then realized I needed to fix up things in the webpack build. I ended up finding out that there was some webpack and jupyter extension futzing to do... That @yuvipanda might have already done in #2779 🤦‍♂️ 

Posting my work for now, I also noticed I could delete more from our webpack config and make the app load faster. 🎉


applications/jupyter-extension/nteract_on_jupyter/index.html
```diff
@@ -97,8 +98,7 @@
 <div id="root">
 </div>
 
-<script src="{{ asset_url }}nteract/static/lib/vendor.js" type="text/javascript" charset="utf-8"></script>
```
rgbkrk, line 100: I found out that this bundle wasn't actually being used. This was being set as an entrypoint, literally as if `vendor.js` was a separate app.

PR #2802, opened 2018-04-27
break out notebook app component from core

rgbkrk:
Trying to kill off bottlenecks for loading by breaking up core better.


rgbkrk:
Woo, this is working really well. Can't wait to switch webpack to loading this component with `import()`.

PR #2823, opened 2018-05-02
Dynamic Transforms

rgbkrk:
This uses `import()` to load all the special output transforms asynchronously.  On startup, any output matching these media types won't be rendered until they're loaded. 

If the output has a mimetype that is yet to be loaded then it will be rendered as null. If `text/html` or `application/javascript` is in an output bundle alongside one of the higher priority display types, we don't render the js or html, avoiding side effects on the page.

For example, if an output had `application/vnd.plotly.v1+json` and `text/html` or `application/javascript` then we will render nothing until the Plotly transform is loaded.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.js
@@ -9,6 +9,44 @@ import {
   transforms as defaultTransforms
 } from "@nteract/transforms";
 
+const displayOrder = [
+  "application/vnd.vega.v3+json",
+  "application/vnd.vega.v2+json",
+  "application/vnd.vegalite.v2+json",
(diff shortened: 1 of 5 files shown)


rgbkrk:
Just an idea of what this helps keep out of our main bundle:

```
4.0K	plotly-412cc5768c73c5d737ae.bundle.js
 52K	tabular-dataresource-642ec9c72b58e09398fe.bundle.js
3.4M	typescript.worker.js
4.0K	vega-transform-34d3a88282d23797f8f3.bundle.js
2.1M	vendors~plotly-ff6afe7ca61e4d55299f.bundle.js
1.1M	vendors~tabular-dataresource-24a83b53c1b5fd181c1e.bundle.js
1.5M	vendors~vega-transform-b77a2bd4a48d8057f22b.bundle.js
```

PR #2825, opened 2018-05-02
delete unused MonacoEnvironment

rgbkrk:
While trying to make Monaco's webworkers resolve when on subpaths I found out Monaco wasn't actually wasn't picking this up so I've cleaned it up.


yuvipanda:
Yeah, I found it wasn't working either :) This got committed as an accident of git commit -p iwith the pre-commit hooks

PR #2826, opened 2018-05-02
[jext] allow for hot module reloading

rgbkrk:
This turns `npm run app:jupyter-extension` into a hot module reloading setup. 😄 I didn't realize having the `asset_url` setup would make this super simple to do.


rgbkrk:
One major note though -- modifying things in `@nteract/core` tends to cause a full reload instead of an in-place reload. 😭 

Modifying the jupyter extension itself is pretty good though. 😄

rgbkrk:
Since our webpack build is a bit weighty though, it does take a while to go from blank page to full app right after running the command. Sometimes you have to reload the page.

rgbkrk:
Oh interesting -- for any compound components that check `child.type` we'll have to use `react-hot-loader`s `areComponentsEqual`.

PR #2828, opened 2018-05-04
include a sample notebook that has the happy data

rgbkrk:
@emeeks showed me this great dataset for playing with our table data viz component and since it's really useful to help figure out issues in the component too, I'd like to include it in the repo. 🎉 📊 

screen shot 2018-05-03 at 5 28 51 pm


mpacer:
Great! I'll merge once circleci is done :).

PR #2829, opened 2018-05-04
Connected components

rgbkrk:
Breaking the last suite of components out of `@nteract/core`. Going to do some restructuring on exports to make the stateModule easier to work with.


rgbkrk:
This is now working. I think I'm going to go ahead and refactor the `stateModule` export usages though while I'm at it.

PR #2830, opened 2018-05-04
apply hot before connect

rgbkrk:
This correct some of our 🌶 hotness so that hot reloading works.


rgbkrk:
⚠️ I did notice that hot reloading of styled-jsx does not work (it _will_ fail) so that's going to take some digging into what @zeit does for next.js.

screen shot 2018-05-04 at 1 15 08 pm

PR #2832, opened 2018-05-04
write npm commands to install & enable jupyter ext

rgbkrk:
cc @mpacer 

I was thinking of how we could expose these as commands that can be run from the root of the repo so I put the `pip` and `jupyter` commands in as new `npm` scripts. Let me know what you think.

### DIFF
--- a/applications/jupyter-extension/package.json
+++ b/applications/jupyter-extension/package.json
@@ -3,6 +3,9 @@
   "version": "0.4.5",
   "description": "nteract on jupyter, as an extension",
   "scripts": {
+    "ext": "npm run ext:install && npm run ext:enable",
+    "ext:install": "pip install -e .",
+    "ext:enable": "jupyter serverextension enable nteract_on_jupyter",
     "build:python": "python setup.py sdist",
     "upload:pypi": "twine upload dist/*",
     "prebuild:python": "rimraf dist",


mpacer:
I'm hesitant around this… 

Python packaging & virtual environments are complicated enough and extremely confusing to people. 

I don't know enough about how virtualenv works to know whether that would be an ok thing to do… with conda I don't know how it would treat `pip` . If we were going to do this it might be safer to do a `python -m pip` rather than plain `pip` the executable.

Also, this doesn't give them any insight as to how to uninstall it (which we should probably add to the README now that I think about it). Importantly, it wouldn't give you any idea of how to remove the configuration file (though of course… right now we don't mention this in the README either so that could be solved). 

Also, I think it'd be nicer if people could learn about this app exclusively as "nteract_on_jupyter" and not "jupyter-extension", but that's going to require a more general renaming approach than can be fixed right now (which I think we talked about).

All that said…  if we want people to install from the root of the repo they could do 
`pip install -e  applications/jupyter-extension` … but they'd still need to run the serverextension enabling command. Is the key win here being able to run this from the root of the repo or only having one command to run for both?

We could put a config file somewhere for them as part of the install (at `<your env path>/etc/jupyter/jupyter_notebook_config.d/nteract_on_jupyter.json` even though `setuptools` won't handle that itself with the `data_files` argument… but I don't know if there are any hooks in `pip uninstall` to then go back and remove it later (which is what a non-editable `pip uninstall nteract_on_jupyter` will do via the setuptools data_files logic).

rgbkrk:
You're so right. This gets wayyyy complicated. I'll go ahead and close this.

Issue #2844, opened 2018-05-07
Test with a screen reader on a Windows Machine


yuvipanda:
We should test with a screen reader and identify issues that crop up.

yuvipanda:
I'm doing this now.

1. Setup a Windows Test VM from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/. I'm using Windows 10.
2. Install NVDA screen reader - https://www.nvaccess.org/
3. Run nteract on your local machine, but listening on 0.0.0.0
4. Access nteract from inside the Windows VM, and see how NVDA sees it!

yuvipanda:
There's a bug in the Windows 10 Edge VM that prevents sound from working. See https://forums.virtualbox.org/viewtopic.php?t=81135 for a fix.

yuvipanda:
NVDA doesn't seem as great, so I'm using JAWS instead now. Much more usable!

yuvipanda:
With JAWS things seem mostly ok! Only problem is that the accessibility help popup window doesn't actually get read by JAWS, which is awkward.

yuvipanda:
This works, I'm going to close this now. We should do this again when we have 'run' functionality.

PR #2846, opened 2018-05-07
Dynamically generate nteract page titles

yuvipanda:
Helps with accessibility by providing context for the current page.

From #2844


rgbkrk:
Fantastic!

PR #2849, opened 2018-05-08
Use <header>, not <nav> for header element

yuvipanda:
`<nav>` is an implicit ARIA Landmark (https://www.marcozehe.de/2009/10/31/easy-aria-tip-4-landmarks/),
with special behavior from screen readers. However, things inside
a nav should all be navigational elements only, since they are
given fairly high priority by screen reader & spoken as 'navigation
links'. Currently, almost none of the links in the header are
navigation (except for the 'Home' link).

`<header>` sidesteps this problem, and is also the more appropriate
tag to use here. If / When we have actual navigational elements
inside the header, those can live inside a `<nav>`

Discovered via #2844


rgbkrk:
Great, thank you. I wondered what the semantic difference was.

PR #2862, opened 2018-05-10
Fixes grid switching bug.

### LINKED ISSUE #2860: [Data Explorer] Table view doesn't work after choosing a chart view
On the [binder preview](https://t.co/xVK5uV0i6T), when doing the initial render you see the table view but after selecting any of the chart views clicking on the table view again does nothing and I see the below error in the JS Concole:

image

jdetle:
@dhirschfeld, could you check this out?

This closes #2860.

### DIFF
--- a/packages/transform-dataresource/src/charts.js
+++ b/packages/transform-dataresource/src/charts.js
@@ -9,7 +9,7 @@ import {
   ResponsiveXYFrame,
   ResponsiveNetworkFrame
 } from "semiotic";
-
+import { VirtualizedGrid } from "./virtualized-grid";
 const parentPath = (d, pathArray) => {
   if (d.parent) {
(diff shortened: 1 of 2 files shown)


dhirschfeld:
I guess I have to build from source to do so? Can prob get around to that this weekend...

jdetle:
There's no rush on my end! I hope you have a great weekend.

emeeks:
@jdetle I see what you were doing to make up for my mistake but I think it ultimately was too ambitious. I corrected the issue by changing the `setGrid` to simply fire `setState` rather than get into `updateChart` (which calculates all the dataviz stuff) and it seems to work fine.

jdetle:
Okay, I trust that you know whats best for this code. Thanks!

rgbkrk:
Thank you both so much for fixing this!

PR #2863, opened 2018-05-10
Fix labels and update select behavior

emeeks:
* Updates `select` inputs to reflect state
* Update labels to make them better


packages/transform-dataresource/src/index.js
```diff
@@ -484,8 +507,7 @@ class DataResourceTransform extends React.Component<Props, State> {
         `}</style>
       </div>
     );
-    console.log("updatedState", updatedState);
```
jdetle, line 487: Love me some console.log removals!

packages/transform-dataresource/src/index.js
```diff
@@ -102,18 +102,29 @@ const availableLineTypes = [
...
+  selectionFunction,
+  title,
+  required,
+  selectedValue
+) => {
+  return (
+    <div style={{ display: "inline-block", margin: "0 10px" }}>
+      <h2>{title}</h2>
+      <select
+        value={selectedValue}
```
rgbkrk, line 116: 😄

rgbkrk:
Probably want to remove the changes to the notebook file here.

Thanks for making these fixes!

emeeks:
I changed the filtering a bit on the DF to make it show two regions so that it would be a more interesting data visualization experience. But it's still smaller than the whole DF.

rgbkrk:
Ah, cool. I'll go ahead and merge this in then.

PR #2866, opened 2018-05-11
Improve icons, add hexbin view

emeeks:
* Adds a hexbin option as a view, which groups points into hexes and styles them based on density and provides membership on hover.
* Adds some margin to avoid left-side cropping of tooltips
* Updates icons to better icons

### DIFF
--- a/packages/transform-dataresource/src/icons.js
+++ b/packages/transform-dataresource/src/icons.js
@@ -10,7 +10,7 @@ import {
 
 export const BoxplotOcticon = (props: any) => (
   <SVGWrapper width={16} height={16} viewBox="0 0 16 16" outerProps={props}>
-    <title>Create Text Cell</title>
+    <title>Summary Diagram</title>
     <path
       fill="lightgray"
       stroke="black"
@@ -21,7 +21,7 @@ export const BoxplotOcticon = (props: any) => (
 
 export const TreeOcticon = (props: any) => (
   <SVGWrapper width={16} height={16} viewBox="0 0 16 16" outerProps={props}>
-    <title>Create Text Cell</title>
+    <title>Dendrogram</title>
     <path
       fill="lightgray"
       stroke="black"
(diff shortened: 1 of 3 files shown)


packages/transform-dataresource/src/icons.js
```diff
@@ -43,11 +43,44 @@ export const NetworkOcticon = (props: any) => (
 
 export const ScatterplotOcticon = (props: any) => (
   <SVGWrapper width={16} height={16} viewBox="0 0 16 16" outerProps={props}>
-    <title>Create Text Cell</title>
+    <title>Scatterplot</title>
```
rgbkrk, line 46: A+ much better titles, I clearly didn't go over these with a fine tooth comb earlier.

rgbkrk (approved):

packages/transform-dataresource/src/icons.js
```diff
@@ -43,11 +43,44 @@ export const NetworkOcticon = (props: any) => (
...
+export const BarChartOcticon = (props: any) => (
+  <SVGWrapper width={16} height={16} viewBox="0 0 16 16" outerProps={props}>
+    <title>Bar Chart</title>
+    <path
+      fill="lightgray"
+      stroke="black"
+      d="M 12.279027,8.4511213 16.181209,8.4511213 16.181209,15.992862 12.279027,15.992862 Z M 6.5947919,4.0324659 10.524457,4.0324659 10.524457,15.92641 6.5947919,15.92641 Z M 0.52657735,0.52511126 4.8907923,0.52511126 4.8907923,15.88305 0.52657735,15.88305 Z"
+    />
+  </SVGWrapper>
+);
```
rgbkrk, line 86: Totally forgot to note -- the reason that some of our other icons have "Octicon" in the name is because we used [Octicons from GitHub](https://octicons.github.com/). These could be renamed in a future PR.

emeeks: That's a good point they could all be called Icon couldn't they.

PR #2869, opened 2018-05-11
Flow charts

emeeks:
* Add types to charts.js


rgbkrk:
Great, thanks for adding these!

If you ever want more bugs shaken out here, swap `d: Object` out to `d: *`. That will make it infer more types. Better yet though would be if you create types for these `d: MyThing`.

emeeks:
Will do. I need to think through how it relates to Semiotic and my own assumptions about what's being passed in and out. Which is great.

PR #2870, opened 2018-05-11
Update Icons

emeeks:
* Change name to -Icon instead of -Octicon.
* Adjust paths a bit so they're not clipped


packages/transform-dataresource/src/icons.js
```diff
@@ -8,18 +8,18 @@ import {
   ResponsiveNetworkFrame
```
rgbkrk, line 8: If you add the `//@flow` at the top of this file it'll really ensure the flow types in here. 😄

PR #2871, opened 2018-05-11
add bdist_wheel to the python build step

mpacer:
Improves the python step to build both a wheel and a sdist. Twine will still pick up both of them from `dist/*`.


rgbkrk (approved):
Thank you!

PR #2873, opened 2018-05-12
Autoenable dev

mpacer:
This improves developer experience while working on the app. 

Previously if you editably installed and were going to run the app in dev mode, you needed to separately enable the extension. Now this checks if you are in `dev_mode` (also settable via traitlet) and will automatically load the server extension for you. 

It also improves the documentation around dev mode making it easier to know what to do if you (for example) weren't reading the README but were running `jupyter nteract --help`.


rgbkrk (approved):
Sweet!

Issue #2848, opened 2018-05-08
refactor execution to be less notebook centric


rgbkrk:
In order to support `play` style running within the jupyter extension (that @yuvipanda wants to do), we need to switch up how we handle kernel messaging and reduction of messaging into state. Right now it's very notebook focused when it doesn't really need to be. My first pass here is likely to make the epics be less notebook centric, then we'll see what I can do with reducers.

rgbkrk:
## Plan

When a kernel gets connected, it sends all messages in batched collections as a new action:

```
type BatchedJupyterMessages = {
  type: "BATCHED_JUPTYER_MESSAGE_ACTION"
  payload: {
    contentRef: ContentRef,
    messages: Array<JupyterMessage>
  }
}
```

These then get routed to the contents model that is the intended receiver. Since this is now devoid of a `CellId`, we have to do a different kind of tracking. Within the notebook for instance, the last execute message id would need to be tracked for each cell (this is similar to how the classic notebook does the tracking). We then drop any messages that aren't "assigned" anywhere.

I'm guessing that the generic action that needs to be sent is:

```
type SendExecuteRequest = {
  payload: {
    contentRef: ContentRef,
    message: JupyterMessage
  }
}
```

What I'm a bit unsure of is how we'd track the CellID for the notebook. There seem to be a few options -- we could store some info in a metadata field in that `SendExecuteRequest` action or we can create this new action from the ol' execute cell action. 💭

rgbkrk:
Now that I've been able to think of this overnight, I'm going to see about introducing it as a new action and epic.

PR #2879, opened 2018-05-15
[WIP] buffer messages for reducer consumption

rgbkrk:
Starting on the approach for #2848.

I'm first grouping all response messages. Next I'll be taking a fresh look at the reducers. Then I'll hook it up to the text-based editor.

cc @yuvipanda


rgbkrk:
This should get re-visited in the future.

PR #2886, opened 2018-05-18
Better Legends

emeeks:
* Changes inline SVG legends into custom HTML legends and sorts the items to ensure that only a reasonable number of items are colored and then colors everything else as grey.

* Also adjusts the thresholds for hexbin


rgbkrk (approved):
Awesome!

packages/transform-dataresource/src/HTMLLegend.js
```diff
@@ -0,0 +1,35 @@
+/* @flow */
+import * as React from "react";
+
```
rgbkrk, line 3: Style wise, I'd prefer to name this component here, even if it ends up the default.

```js
const HTMLLegend = ...

export default HTMLLegend
```

One thing I tend to do is move props to their own type with something like:

```js
type HTMLLegendProps = {
  values: Array<string>,
  colorhash: Object
}
```

putting it right before the function itself. With only two props it's not a big deal, as soon as it gets bigger I move them out.

emeeks: That makes sense. Will do in future.

PR #2887, opened 2018-05-18
React and flow upgrade

rgbkrk:
This enforces us to be upgraded to React 16.3. I'm going to be switching our MathJax setup to the new context API as well as our theme provider after this PR.


rgbkrk:
There we go. Things were failing because the monaco webpack plugin was expecting a higher version of monaco that was not part of our semver range.

PR #2889, opened 2018-05-20
chore: adding open to NotebookApp menu for #2731

### LINKED ISSUE #2731: [jext] [menu] File -> Open from <NotebookApp />
Follow on to #2430.

From the notebook menu, clicking open will redirect to the directory the current notebook is in (or maybe `/tree` is fine...)

alexandercbooth:
Closes #2731. First time taking Jupyter extension for a spin; very cool!

### DIFF
--- a/packages/connected-components/src/notebook-menu/constants.js
+++ b/packages/connected-components/src/notebook-menu/constants.js
@@ -4,6 +4,7 @@
 // These actions map to a case in a switch handler. They are meant to cause a
 // unique action from the menu.
 export const MENU_ITEM_ACTIONS = {
+  OPEN_HOME: "open-home",
   SAVE_NOTEBOOK: "save-notebook",
   DOWNLOAD_NOTEBOOK: "download-notebook",
   EXECUTE_ALL_CELLS: "execute-all-cells",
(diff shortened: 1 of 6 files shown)


rgbkrk:
Awesome! Thanks for taking the jupyter extension for a spin!

This whole setup can be synchronous and part of the connected menu I think.

* `window.open` is synchronous (and there's not an asynchronous variant)
* `window.open`'s side effect doesn't change anything on our page

You are right in catching if `window.open` failed or not since it can return `null`. 🤔 I suppose we need a flow for popup handling detection.

However! There's another way to do this without having to deal with any JS issues. Create an `<a>` link that matches the style in the menu, likely like this:

```
<a href="/nteract/edit" target="_blank">Open Directory View</a>
```

Since this comes from user interaction, it doesn't count towards the browser's built in pop up throttling.

As an aside, the nteract logo in the upper left also has the behavior of navigating up a directory and showing the directory listing (though it's clearly a kinda hidden feature).

alexandercbooth:
Oops, was fiddling around with flow errors before I saw this 😅

alexandercbooth:
That is a good idea, I will give it a shot tomorrow 👍. I had noticed the logo behavior, but didn't think to mimic it -- I like it!

alexandercbooth:
@rgbkrk I think that idea is better and keeps things more simple,  I could close this in favor of #2890 instead 👍

alexandercbooth:
Closing in favor of #2890 😄

PR #2890, opened 2018-05-20
chore: adding open directory item to NotebookApp

### LINKED ISSUE #2731: [jext] [menu] File -> Open from <NotebookApp />

alexandercbooth:
Closes #2731. I thought this would be cleaner to start fresh from #2889 😄.

### DIFF
--- a/packages/connected-components/src/notebook-menu/index.js
+++ b/packages/connected-components/src/notebook-menu/index.js
@@ -306,6 +306,15 @@ class PureNotebookMenu extends React.Component<Props, State> {
             >
               Download (.ipynb)
             </MenuItem>
+            <MenuItem>
+              <a
+                href="/nteract/edit"
+                style={{ textDecoration: "none" }}
+                target="_blank"
+              >
+                Open Directory View
+              </a>
+            </MenuItem>
           </SubMenu>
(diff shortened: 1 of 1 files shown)


packages/connected-components/src/notebook-menu/index.js
```diff
@@ -306,6 +306,15 @@ class PureNotebookMenu extends React.Component<Props, State> {
...
             </MenuItem>
+            <MenuItem>
+              <a
+                href="/nteract/edit"
+                style={{ textDecoration: "none" }}
+                target="_blank"
+              >
+                Open Directory View
+              </a>
+            </MenuItem>
```
rgbkrk, line 317: A+, much better. 😄

rgbkrk:
Now that I've taken this for a spin locally, I think we should make a few style changes:

* Put this menu item at the top of the list
* Label it "Open..." (mostly to match the jupyter server)
* Override the `color` to match that of the other items (and not inherit the `a` style). It _might_ work to write `color: currentColor`.

alexandercbooth:
Good thinking, that looks better with those changes. I opened #2895 👍

PR #2898, opened 2018-05-22
Data explorer cleanup

emeeks:
* Made the flow types into a separate object and `HTMLLegend` declared and then exported
* Fixed how annotations were displaying in scatterplot
* Update Semiotic version to correct some graphical nonsense


rgbkrk:
Alright!

PR #2908, opened 2018-05-25
WIP Routing for jupyter extension

rgbkrk:
This is a placeholder PR so that I can take note of what @theengineear and I ended up figuring out we could do. I was feeling a bit blocked because of our contentRefs setup and the fact that location is managed a bit externally. Basically, when location changes we don't know what contentRef to use. Solution: make `<Link />`s in the directory listing include a query param for the contentRef, which we can strip off.


rgbkrk:
Nice, that todo bot is already on a roll.

rgbkrk:
I think with the query param stuff this will be off to the races very soon.

One issue I'm expecting to run into though is what happens when you hit the back button navigating through the directory listing. Those won't have the temporary contentRefs in the URL.

rgbkrk:
Closing, superseded by #3838.

PR #2915, opened 2018-05-28
pin vega things

rgbkrk:
master is broken for desktop, time to pin so it's fixed up for now


rgbkrk:
Wow I've also broken master in new ways with the new babel setup. I'm going to merge this since it fixes vega. I'll have to keep working at fixing the babel + jest setup.

PR #2919, opened 2018-05-28
Backoutta babel 7

rgbkrk:
Too early for us to move to babel 7 + next.js 6. :( Mostly because of jest.


rgbkrk:
Yay back to normalcy!

PR #2923, opened 2018-05-28
add notebook as dependency for nteract_on_jupyter

ivanov:


rgbkrk:
The babel issue on master is fixed now so I'm going to go ahead and merge this since the python bits here look good.

PR #2928, opened 2018-05-30
feat: adding close and halt to notebook menu app

### LINKED ISSUE #2724: [jext] Close and Halt menu entry
Follow on from #2430:

> `DELETE` the `session`, wait for success, then close the window

alexandercbooth:
Closes #2724.

### DIFF
--- a/packages/core/src/epics/kernel-lifecycle.js
+++ b/packages/core/src/epics/kernel-lifecycle.js
@@ -250,3 +250,18 @@ export const restartKernelEpic = (action$: ActionsObservable<*>, store: *) =>
       );
     })
   );
+
+export const closeAndHaltEpic = (action$: ActionsObservable<*>) =>
+  action$.pipe(
+    ofType(actionTypes.CLOSE_AND_HALT),
+    concatMap((action: actionTypes.CloseAndHalt) => {
+      const kernelRef = action.payload.kernelRef;
+      return of(
+        actions.killKernel({
+          restarting: false,
+          kernelRef: kernelRef
+        }),
+        window.close()
+      );
+    })
+  );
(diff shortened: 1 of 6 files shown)


packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -250,3 +250,18 @@ export const restartKernelEpic = (action$: ActionsObservable<*>, store: *) =>
...
+  action$.pipe(
+    ofType(actionTypes.CLOSE_AND_HALT),
+    concatMap((action: actionTypes.CloseAndHalt) => {
+      const kernelRef = action.payload.kernelRef;
+      return of(
+        actions.killKernel({
+          restarting: false,
+          kernelRef: kernelRef
+        }),
+        window.close()
```
rgbkrk, line 264: What might be a little scary here is that kernel might not be killed before the window is closed. This epic probably belongs in either the websocket kernel or as part of the jupyter extension itself. We need to wait for the server to respond after killing the kernel before we can close the window.

rgbkrk: Adding on, the kill kernel action will end up triggering the `killKernelEpic` from `packages/core/src/epics/websocket-kernel.js` (see
 https://github.com/nteract/nteract/blob/master/packages/core/src/epics/websocket-kernel.js#L297-L310)

I'd probably make the callback to the `concatMap` here use the `sessions` API directly

```js
const kernel = selectors.kernel(state, { kernelRef: action.payload.kernelRef })

return sessions.destroy(serverConfig, kernel.sessionId).pipe(
  map(() => {
    window.close();
    return actions.killKernelSuccessful({
      kernelRef: action.payload.kernelRef
    });
  }),
  catchError(err => {
    // Not sure if we want to error here or just close anyways
    window.close();
    return of(
      actions.killKernelFailed({
        error: err,
        kernelRef: action.payload.kernelRef
      })
    );
  })
);

```

alexandercbooth:
Since most browsers block pop-ups, a new notebook is created in the window that is already opened (tree view), so closing cannot happen from a new [script](https://www.w3.org/TR/html51/browsers.html#dom-window-close). At the moment, this PR will only close a notebook that is being reopened. Unless we change how a notebook is opened, this might be better off closed for now and revisited in the future 🤔.

rgbkrk:
Ok, we can close this for now.

🤔we could make the page show "You can close this window now" and have a button to close it, once we know background resources have been cleaned up.

PR #2934, opened 2018-05-30
Document input, outputs, pagers components

willingc:


packages/presentational-components/src/index.js
```diff
@@ -2,211 +2,16 @@
 import * as React from "react";
 
 import Highlighter from "./syntax-highlighter";
+import { Inputs } from "./components/inputs.js";
```
rgbkrk: This will need to be:

```js
import { Input } from "./components/input.js";
```

packages/presentational-components/src/index.js
```diff
@@ -2,211 +2,16 @@
...
-            /** Note omission of the li:only-child styling **/
-          `}</style>
-        </div>
-      );
-    }
-
-    return null;
-  }
-}
+export { themes, Inputs, Outputs, Pagers, Prompt, PromptBuffer, Source };
```
rgbkrk: `Input` here too

rgbkrk:
I've rebased this to get it working with everything on `master` while Carol has been meeting with folks.

PR #2939, opened 2018-05-31
include import code at the top of each example

rgbkrk:
Hey look @willingc, we can render static `jsx`!

screen shot 2018-05-31 at 6 29 54 am

I put `import { Thing } from "@nteract/presentational-components` at the top of each of our documented components.


willingc (approved):

willingc:
:heart:

PR #2947, opened 2018-06-04
Add NotebookRender component

ghost:
This component allows to render notebooks server-side using remark for
markdown rendering and katex for math display.

It's greatly inspired from the NotebookPreview component however NotebookPreview relies on document variable and can't be rendered server-side.

This is a work in progress, and needs to be improved with some help.
- [x] pass flow test
- [x] pass jest test
- [x] add a page in showcase.

**Preview**
image


packages/notebook-render/README.md
```diff
@@ -0,0 +1,26 @@
+# Notebook Preview
+
+This package provides a small React component that allows to render notebooks server-side with `ReactDOMServer.renderToStaticMarkup()`.
+
+Markdown is parsed with `marked`, and maths are rendered with `katex` so you need to include the katex.css stylesheet in your page.
```
rgbkrk, line 5: Excellent! Thank you for kicking off this new package so we have another version people can work with.

ghost:
@rgbkrk @lgeiger I'm trying to implement a simple test for this package based on the test you wrote for `NotebookPreview`.
However I have a problem with the cell IDs that change on every test run. it seems that the uuid mock should fix this but it isn't the case. Can you give me some advices to fix this.

packages/notebook-render/README.md
ghost: You're welcome, in fact you already done the hard job with all the packages provided with nteract.

I'd like to use this component inside the `gatsby-transformer-plugin`, but it could be more generally used as an alternative to the `notebook.js` lib used to convert notebooks to HTML.

packages/notebook-render/__tests__/index.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+import { dummyCommutable, dummyJSON } from "../../core/src/dummy";
+
+// In order to get reproducable snapshots we need to mock the uuid package
+jest.mock("uuid", () => {
+  let uuid = 1;
+  return {
+    v4: jest.fn(() => uuid++)
+  };
+});
```
rgbkrk, line 18: This is definitely the same as NotebookPreview, so I'm pretty surprised by this not working.

lgeiger: Guessing from the CI logs, it's working fine. I guess you need to update the snapshot.

ghost: I already updated the snapshot, and every time I run the test cell IDs are changed randomly. I also checked that the uuid mock was never called. However I'm surprised logs on the CI gives IDs 1,2,3…

rgbkrk: I'm running this locally now to check it out.

rgbkrk: I ran:

```
npm t -- --updateSnapshot
```

Followed by

```
npm t
```

and all is passing now with the expected incrementing IDs (1,2,3,...).

If I had to guess, locally on your machine things aren't linked right somehow @benjaminabel. You may have to run `git clean -xfd && npm i`.

If you like though I can submit a commit that fixes things up with new snapshots.

ghost: Thanks for your help @rgbkrk, `git clean -xfd && npm i`, worked. I'll now add the page in showcase.

rgbkrk:
Now that this passes CI, do you think this is still WIP?

I can do an earnest review of the PR if you're ready for it too. 😄

ghost:
@rgbkrk I'm working in a page on showcase to test the katex rendering, and then I'll remove the WIP for a review.

rgbkrk:
Ok fantastic! I'm super looking forward to using this with gatsby, I've been hoping to tinker with using it for our own docs. 😄

ghost:
@rgbkrk  I added the page to showcase and a simple server-side test, so I removed the WIP.

ghost:
@rgbkrk I've prepared a preview of `Notebookrender` and `Notebookrender` side by side at https://gatsby-contrib.github.io/gatsby-transformer-ipynb/Three-Body%20Problem/.

rgbkrk:
Fantastic! Happy to merge this in and ship soon so we can iterate on everything.

By the way, if you'd like we can certainly host the gatsby transformer in nteract as well.

rgbkrk:
Thank you again for submitting this! I've also extended an invite to join the nteract org both on github and on npm. Welcome to the nteract party! 🎉

ghost:
@rgbkrk thanks for the invitations, I'm happy to join the nteract org. I'll submit the gatsby-transformer soon. Maybe we could  move the showcase from next to gatsby.

rgbkrk:
> Maybe we could move the showcase from next to gatsby.

That's totally cool. We were mostly using showcase to hack on components without the full blown app having to run, switching to gatsby is cool since it has a live reloading setup.

----

Bit of an aside --

We should probably tinker with gatsby in general for documentation, since we've been using [React styleguidist](https://react-styleguidist.js.org/) for [light component documentation](https://components.nteract.io). The react style-guidist setup uses react-docgen underneath to grab all the type info + comments, so I imagine we could build some cohesive site(s) on top of that data. Likely `docs.nteract.io` for user facing documentation, `devdocs.nteract.io` for nteract libraries and components.

PR #2956, opened 2018-06-06
chore: decrease card size

### LINKED ISSUE #2951: [jext] Make new notebook cards smaller
From a user:

> Make the new notebook cards smaller, or decrease the image size for the icon, they feel like they are using a big portion of the page.

alexandercbooth:
Closes #2951. Not sure if this is too small/or not small enough 🤔


rgbkrk:
You'll need to update the jest snapshots by doing this:

```
npm t -- --updateSnapshot
```

As for size, I've noticed that everything looks a bit better at 90%. That's a totally made up Kyle heuristic though. I think we should go with this and try it out for a bit. 😄

alexandercbooth:
Ahh I have been wondering how to update those! Haha, made up Kyle heuristic sounds good to me 😄. Sounds good, though!

PR #2958, opened 2018-06-07
New records

rgbkrk:
# New Records for nteract

Philosophy: create very flat records for core data types used in nteract, to be used across applications.

> Wait, what about commutable?

Commutable has been _awesome_ and should be borrowed from to build this up. However, commutable relied heavily on `Immutable.Map`'s. These lose all the benefits of flow typing, defaults, and direct property access that records have.

> Could we add these directly to commutable?

I'd love to. However, I didn't want to make changes that required refactoring all the apps that use them, as well as any external usage (Hydrogen included) that would have to adapt to new interfaces.

> Where should I start in helping with this effort?

Look over how `src/outputs/stream.js` and `src/outputs/index.js` sets up the base types. Implement the other output types in a similar fashion.

- [x] stream
- [x] display_data
- [x] execute_result
- [x] error

## References

- [nbformat v4](https://raw.githubusercontent.com/jupyter/nbformat/master/nbformat/v4/nbformat.v4.schema.json)
- [Jupyter Message Format](http://jupyter-client.readthedocs.io/en/stable/messaging.html#general-message-format)
- [commutable's implementation](../commutable/src/v4.js) -- `packages/commutable/src/v4.js`

## Testing

Run

```
npm t -- records --watch
```

to run the tests locally while watching to all changes in the records setup.

## Types!

If your editor isn't already helping you with this, use

```
npm run flow
```

## Documentation

To help figure out where we _might_ end up with documentation for nteract libraries (as opposed to components, which use react-docgen / react-styleguidist ), use [documentation.js](https://github.com/documentationjs/documentation) to run docs for this library.

```
npm install -g documentation
documentation serve --watch packages/records/src
```

I have a feeling we'll end up having to do some custom things with our monorepo to make this more palatable.

### DIFF
--- a/packages/records/__tests__/index-spec.js
+++ b/packages/records/__tests__/index-spec.js
@@ -38,3 +38,50 @@ describe("stream output", () => {
     );
   });
 });
+
+describe("display_data output", () => {
+  test("can be converted from nbformat", () => {
+    expect(
+      nteractRecords.outputFromNbformat({
+        output_type: "display_data",
+        data: {
+          infinityStones: [
+            "mind\n",
+            "time\n",
+            "space\n",
+            "reality\n",
+            "power\n",
+            "soul"
+          ]
+        },
+        metadata: { "application/json": { expanded: true } }
+      })
+    ).toEqual(
+      nteractRecords.makeDisplayDataOutputRecord({
+        outputType: "display_data",
+        data: { infinityStones: "mind\ntime\nspace\nreality\npower\nsoul" },
+        metadata: { "application/json": { expanded: true } }
+      })
+    );
+  });
+
+  test("can be converted from jupyter messages", () => {
+    expect(
+      nteractRecords.displayDataRecordFromMessage({
+        header: {
+          msg_type: "display_data"
+        },
+        content: {
+          data: { anotherDay: "anotherDoug" },
+          metadata: { "application/json": { expanded: true } }
+        }
+      })
+    ).toEqual(
+      nteractRecords.makeDisplayDataOutputRecord({
+        outputType: "display_data",
+        data: { anotherDay: "anotherDoug" },
+        metadata: { "application/json": { expanded: true } }
+      })
+    );
+  });
+});
--- a/packages/records/__tests__/index-spec.js
+++ b/packages/records/__tests__/index-spec.js
@@ -1,0 +1,40 @@
+// @flow
+import * as nteractRecords from "@nteract/records";
+
+describe("stream output", () => {
+  test("can be converted from nbformat", () => {
+    expect(
+      nteractRecords.outputFromNbformat({
+        output_type: "stream",
+        name: "stdout",
+        text: ["sup\n", "yall"]
+      })
+    ).toEqual(
+      nteractRecords.makeStreamOutputRecord({
+        outputType: "stream",
+        name: "stdout",
+        text: "sup\nyall"
+      })
+    );
+  });
+
+  test("can be converted from jupyter messages", () => {
+    expect(
+      nteractRecords.streamRecordFromMessage({
+        header: {
+          msg_type: "stream"
+        },
+        content: {
+          name: "stdout",
+          text: "it is love we must hold on to\nnever easy but we try"
+        }
+      })
+    ).toEqual(
+      nteractRecords.makeStreamOutputRecord({
+        outputType: "stream",
+        name: "stdout",
+        text: "it is love we must hold on to\nnever easy but we try"
+      })
+    );
+  });
+});
--- a/packages/records/__tests__/index-spec.js
+++ b/packages/records/__tests__/index-spec.js
@@ -151,7 +151,7 @@ describe("error output", () => {
         output_type: "error",
         ename: "Thor",
         evalue: "Pirate Angel",
-        traceback: Immutable.List(["sweet", "rabbit"])
+        traceback: ["sweet", "rabbit"]
       })
     );
   });
@@ -165,7 +165,7 @@ describe("error output", () => {
         content: {
           ename: "cats",
           evalue: "good",
-          traceback: Immutable.List(["squirrel"])
+          traceback: ["squirrel"]
         }
       })
     ).toEqual(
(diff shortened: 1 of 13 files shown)


packages/records/__tests__/index-spec.js
```diff
@@ -38,3 +38,50 @@ describe("stream output", () => {
...
+      nteractRecords.outputFromNbformat({
+        output_type: "display_data",
+        data: {
+          infinityStones: [
+            "mind\n",
+            "time\n",
+            "space\n",
+            "reality\n",
+            "power\n",
+            "soul"
```
rgbkrk, line 55: 😆 

thanos crushhhhh

packages/records/src/outputs/display-data.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ */
+
+export type DisplayDataType = "display_data";
+
+export const DISPLAYDATA = "display_data";
+
+// In-memory version
+type DisplayDataOutput = {
+  outputType: DisplayDataType,
```
rgbkrk, line 24: For the in-memory types I'd prefer to keep `camelCase`, keeping with JS conventions. For the messages and nbformat, we'll go with what the spec has (`snake_case`, unless otherwise indicated).

packages/records/__tests__/index-spec.js
alexandercbooth: Haha, I might be obsessed... 😄

packages/records/src/outputs/display-data.js
alexandercbooth: Sorry, which ones do you mean? Do you mean `type DisplayDataOutput` should be `type displayDataOutput`?

packages/records/src/outputs/error.js
```diff
@@ -0,0 +1,92 @@
...
+export type ErrorType = "error" | "pyerr";
+
+export const ERROR = "error";
+
+// In-memory version
+type ErrorOutput = {
+  outputType: ErrorType,
+  ename: string,
+  evalue: string,
+  traceback: Immutable.List<string>
```
rgbkrk: I think the traceback in our data structure here should be a single string, joined.

packages/records/src/outputs/index.js
```diff
@@ -29,6 +33,8 @@ export function outputFromNbformat(output: NbformatOutput): OutputRecord {
       return displayData.displayDataRecordFromNbformat(output);
     case executeResult.EXECUTERESULT:
       return executeResult.executeResultRecordFromNbformat(output);
+    case error.ERROR:
+      return error.errorRecordFromNbformat(output);
     default:
       // $FlowAllowFeatureDetection: At runtime, allow fallback
       return unrecognized.unrecognizedRecordFromNbformat(output);
```
rgbkrk: Now that we've done all this work, we should also make a `JupyterMessage -> Output` function.

packages/records/src/outputs/display-data.js
rgbkrk: Whoops! No I thought I read this as `output_type`. Nevermind!

alexandercbooth: Oh gotcha, no problem!

packages/records/src/outputs/error.js
alexandercbooth: Do you mean like `common.MultilineString` and `common.demultiline` like with the stream `text`?

rgbkrk: Ah right, this one is weird because the message spec returns them like this.

```
{
   'status' : 'error',
   'ename' : str,   # Exception name, as a string
   'evalue' : str,  # Exception value, as a string
   'traceback' : list(str), # traceback frames as strings
}
```

🤔 

Maybe nothing to do here then, we'll keep them as `Immutable.List`. Part of the reason I was hoping to step away from having fields that are immutable is that we can support components that take non-immutable and immutable Record varieties when they're records. When they're `Immutable.List`s we're tied into it.

packages/records/src/outputs/index.js
alexandercbooth: Do you mean something like this?
```js
import type { JupyterMessage } from "@nteract/messaging";
.
.
.
/**
 * Turn any output that was in JupyterMessage into a record
 */
export function outputFromMessage(msg: JupyterMessage<any,any>): OutputRecord {
  switch (msg.header.msg_type) {
    case stream.StreamType:
      return stream.streamRecordFromMessage(msg);
    case displayData.DisplayDataType:
      return displayData.displayDataRecordFromMessage(msg);
    case executeResult.ExecuteResultType:
      return executeResult.executeResultRecordFromMessage(msg);
    case error.ErrorType:
      return error.errorRecordFromMessage(msg);
    default:
      // $FlowAllowFeatureDetection: At runtime, allow fallback
      return unrecognized.unrecognizedRecordFromMessage(msg);
  }
}
```
If this is the right direction, I am not sure what the correct arguments for `JupyterMessage<,>` should be , though, as flow doesn't like these one bit..🤔 
Also, what would be the right default? Would we make a `unrecognized.unrecognizedRecordFromMessage` function?

packages/records/src/outputs/error.js
alexandercbooth: Right, I am not sure of a way out here without `Immutable.List`, but that makes sense and it would be good if we could find a way.

packages/records/src/outputs/index.js
rgbkrk: Yeah that's pretty good. Since there's not really an unrecognized output type of message really, I wonder if we should just return `null`. Perhaps we could just use the unrecognized type, using the `header.msg_type` and `content`.

packages/records/src/outputs/error.js
rgbkrk: I suppose we could make it `Immutable.List<string> | Array<string>` or even just `Iterable<string>` (this is the [general form for iterables in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols)).

alexandercbooth: 👍

rgbkrk:
I think I'm going to call this pretty darn good, and we can keep iterating on it before release. I'll go ahead and bring it in and connect it up to some examples. 

Thanks @alexandercbooth!

alexandercbooth:
Awesome, looks great! My pleasure to help 😄

PR #2961, opened 2018-06-07
Host cache

rgbkrk:
image

Some of you may know about some components I've been [putting together that use React 16.3's new context API to "acquire" hosts from binder](https://github.com/rgbkrk/react-context-backed-jupyter). In this PR I've extracted _just_ the part that allocates and stores binder host info in [LocalStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).

```js
import { LocalHostStorage } from "@mybinder/host-cache";

const lhs = new LocalHostStorage();

import { LocalHostStorage } from "@mybinder/host-cache";

const lhs = new LocalHostStorage();

// Allocate returns a Promise<HostConfig>, ensuring the host is up too
const host = await lhs.allocate({ repo: "nteract/vdom" });

console.log("Endpoint: ", host.endpoint);
console.log("Token: ", host.token);

// If you call allocate with the same params and the host is still active, you'll get the same host
const host2 = await lhs.allocate({ repo: "nteract/vdom" });

console.log(
  "Same?",
  host.endpoint === host2.endpoint && host.token === host2.token
);

// You're not limited to mybinder.org. If you have your own binderhub deployed, use it!

const ownHost = await lhs.allocate({
  repo: "nteract/vdom",
  ref: "master",
  binderURL: "https://example.com"
})
```


rgbkrk:
Admittedly, there are some rough edges in this code base as I was prototyping. Probably good to clean some of those up.

rgbkrk:
cc @gnestor @yuvipanda @willingc @alexandercbooth 

Questions:

* Should we include the React Component(s) that use this host cache and return a `host`?
  - I could also include it in a `@nteract/experimental` package or something. ¯\\\_(ツ)\_/¯
* How do people feel about the current interface?

packages/host-cache/README.md
```diff
@@ -0,0 +1,37 @@
+# Host Cache
+
+> Come one, come all. Step right up and get a host!
+
+This package allows you to get and store a host configuration from binder, "caching" it via [LocalStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).
+
+```js
+import { LocalHostStorage } from "@mybinder/host-cache";
```
rgbkrk: I need to make sure to export this name again since I changed it to only export the component

packages/host-cache/src/components/host.js
```diff
@@ -0,0 +1,68 @@
...
+type HostProps = {
+  children?: React.Node,
+  repo: string,
+  ref?: string,
+  binderURL?: string
+};
+
+import type { ServerConfig } from "../host-storage";
+
+class HostAcquirer extends React.Component<HostProps, ServerConfig> {
```
rgbkrk: I seriously have no idea what to call this component that is, in my mind, a redux like Provider (when it uses a named provider)

packages/host-cache/src/host-storage.js
```diff
@@ -0,0 +1,200 @@
+// @flow
+
+import { formBinderURL, binder } from "rx-binder";
+import { kernels, apiVersion } from "rx-jupyter";
+
+import * as rxJupyter from "rx-jupyter";
+import * as operators from "rxjs/operators";
```
alexandercbooth: I apologize if I am mistaken, but do we need these two * imports?

packages/host-cache/README.md
alexandercbooth: Hm, I am not quite sure what you mean here. Are you saying you have this ->
```js
import { LocalHostStorage } from "@mybinder/host-cache";

const lhs = new LocalHostStorage();
```
in here twice because you only export the component in host-storage?

packages/host-cache/src/host-storage.js
```diff
@@ -0,0 +1,200 @@
...
+
+  constructor() {
+    this.localForage = new LocalForage();
+    window.addEventListener("storage", this.handleStorageEvent);
+  }
+
+  // Developer has to call this to cleanup :(
+  close() {
+    window.removeEventListener("storage", this.handleStorageEvent);
+  }
```
alexandercbooth, line 83: When the developer calls this to cleanup, does that mean to clean the localStorate up? If so, would we need to also call `localStorage.removeItem()` or `localStorage.clear()`?

alexandercbooth (approved):
> Should we include the React Component(s) that use this host cache and return a host?

I think it is helpful to include the starter component here.  

>How do people feel about the current interface?

The interface seems clean and straightforward to use 👍.

I posted a few comments inline, but from what I can follow, this looks good! 😄

packages/host-cache/src/host-storage.js
rgbkrk: We want to leave it in localstorage because it's what provides the cache. We could provide a utility to clean the cache keys that are ours though. 🤔

packages/host-cache/README.md
rgbkrk: I should also fix that!

I was saying that `src/index.js` doesn't export `LocalHostStorage`, so I need to export it again.

packages/host-cache/src/host-storage.js
rgbkrk: So right, cleaned up. Thanks!

rgbkrk:
Ok, I say we roll with this for now since @alexandercbooth had a look. Time to ship our first `@mybinder` component and library. 😄 

Let's try using these!

gnestor:
Sorry for chiming in late here, but I've been using these for the past week or so and I'm **so happy** 👍👍

> Should we include the React Component(s) that use this host cache and return a host?

I think it's good to include this in the `Host` component because it reduces work on binder's end, right? 

My use case: giving code blocks in a markdown editor access to a Jupyter kernel so that they can run code, display output, etc.

For my use case, I ended up refactoring this so that they aren't components anymore, just a single function that can get a `kernel` for a repo/kernelName:

```js
// ./binder.js

import { kernels } from 'rx-jupyter';
import { LocalHostStorage } from './storage.js';

function getHost(
	options: BinderOptions = {
		repo: 'nteract/vdom',
		ref: 'master',
		binderURL: 'https://mybinder.org'
	}
): Promise<BinderState> {
	const localHostStorage = new LocalHostStorage();
	return localHostStorage.allocate(options);
}

function getKernel(options: KernelOptions): Observable<KernelState> {
	const { host, kernelName } = options;
	return kernels
		.start(host, kernelName, '/')
		.toPromise()
		.then(xhr => {
			return {
				...xhr.response,
				channels: kernels.connect(host, xhr.response.id),
				last_activity: new Date(xhr.response.last_activity)
			};
		});
}

export async function getBinder(repo: string = 'nteract/vdom', kernelName: string = 'python'): Promise<?Binder> {
	const host = await getHost({
		repo,
		ref: 'master',
		binderURL: 'https://mybinder.org'
	});
	const kernel = await getKernel({ kernelName, host });
	return { host, kernel };
}
```

Then, in the doc (above the code blocks) I create binder instance(s):

```jsx
import { getBinder } from './binder';

const binder = getBinder();
```

The binder instance is passed to a `Session` that provides an interface to a kernel connection:

```jsx
import * as messaging from '@nteract/messaging';

export default class Session extends React.Component<Props, State> {
  state = {
    messages: []
  };

  componentDidMount() => {
    this.subscription = this.props.kernel.channels.subscribe(handleMessage});
  };

  handleMessage = (msg: JupyterMessage) => {
    this.setState({ messages: [...this.state.messages, msg] });
  }

  executeRequest = (source: string): Promise<void> => {
    this.props.kernel.channels.next(messaging.executeRequest(source));
  };

  render() {
    const { messages } = this.state;
    return (
      <Cell messages={messages} executeRequest={this.executeRequest} />
    )
  }
}
```

Finally, `messages` and `executeRequest` are passed to a `CodeBlock`. With that, `CodeBlock` can consume message to derive output and kernel state and send messages to affect state.

The HOC interface could work similarly:

```js
withSession(CodeBlock);
```

The only thing that I don't like about HOCs is that they hide where props come from, whereas the context API makes it very visible (arguably too visible at times). After playing with this for a couple weeks, I think that an ideal interface would be:

```jsx
<SessionProvider repo="nteract/vdom" kernelName="python">
  <Doc />
<SessionProvider>
```

Then, anywhere inside the doc (as there may be several Python code blocks that should share the same session/kernel):

```jsx
<SessionConsumer>
  {session => <CodeBlock session={session} />}
<SessionConsumer>
```

In conclusion: For my use case, `Session` -> `CodeBlock` is a better abstraction than `Host` -> `Kernel` -> `CodeBlock` because all `CodeBlock` needs is to be able to run code and display output.

rgbkrk:
I'll go ahead and implement the Kernel provider you outlined above. If we do one called `<Session />` then I think that it would likely need to be mapped to `/api/sessions`, which gets you a kernel while also establishing a link to some content.

I've been a bit more iffy on what the interface should be for anything that passes back the messages and the way to execute. _Mostly_ because if we send all the messages down, the entire tree will re-render on every single message received. 

The interface that gets to just messages or just outputs seems like a good area for us to explore with new libraries in the monorepo.

rgbkrk:
Oook, this branch is worth trying out now that I stuck some of the component we hacked on inside the styleguide.

Clone this branch, `npm install` it and run:

```
npx styleguidist server
```

Then open http://127.0.0.1:6060/#!/Kernel

gnestor (approved):

gnestor:
I tried this branch and it works! It's great to see one of these binder component land in the style guide. This looks good. We can always add higher-level components later that maybe only pass derived state like `output` and `kernelStatus` and callbacks like `execute`.

rgbkrk:
Alright I'm going to get this PR passing and then we can merge it on in. 😄

rgbkrk:
Away we go!

PR #2963, opened 2018-06-11
feat: adding loading/saving/error icon to jext

### LINKED ISSUE #2847: [jext] Show contents status
For each `contentRef`, there's an entry in `state.core.communication.contents.byRef`. We can use this information to get if a piece of content is saving, loading, and if it's in error state.

I started playing around with this, using some `iron-icon`s to display each of these states, next to the `<LastSaved />` component.

alexandercbooth:
Closes #2847.

I wasn't sure which iron-icons to use, so at the moment I went with -- 
- saving: icons:save
- loading: icons:autorenew
- error: icons:error

Also, I am not sure if these icons should be in @nteract/octicons since they are not octicons, but I didn't know where else to put them at the moment 🤔

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
@@ -6,6 +6,8 @@ import * as Immutable from "immutable";
 import { selectors, actions } from "@nteract/core";
 import type { ContentRef, FileContentRecord, AppState } from "@nteract/core";
 
+import { LoadingIcon, SavingingIcon, ErrorIcon } from "@nteract/octicons";
+
 import moment from "moment";
 
 import { ThemedLogo } from "../../components/themed-logo";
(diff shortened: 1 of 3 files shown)


rgbkrk:
Whoa nice! I've been thinking of bringing in the iron icons as a new React library inside our repo for now. This is looking good.

I think we probably want to use a [React transition group](https://github.com/reactjs/react-transition-group) to make sure to animate when saving is rather fast, so we don't have a flicker. We can do that after this PR though.

rgbkrk:
Thank you @alexandercbooth!

alexandercbooth:
>I think we probably want to use a React transition group to make sure to animate when saving is rather fast, so we don't have a flicker. 

Oo I like that idea a lot -- the flicker was definitely too fast to get much from.

And that makes sense to create a new library for the iron-icons. I will create a couple of issues to track these ideas 👍.

Thanks!

PR #2969, opened 2018-06-12
desktop: Fix font loading for bundled app

lgeiger:
`require.resolve` was failing in the bundled app: c45b8b0

Fixes #2797

/cc @rgbkrk

PR #2970, opened 2018-06-12
desktop: Don't split webpack build in production mode

lgeiger:
This reduces the time between `first paint` and `DOMLoaded` from 1.69s to 1.52s and slightly reduces the overall bundle size.

Found while debugging #2969

### DIFF
--- a/applications/desktop/webpack.common.js
+++ b/applications/desktop/webpack.common.js
@@ -34,20 +34,7 @@ const mainConfig = {
 const rendererConfig = {
   mode: "development",
   entry: {
-    app: "./src/notebook/index.js",
-    vendor: [
-      "react",
-      "react-dnd",
-      "react-dnd-html5-backend",
-      "react-dom",
-      "react-redux",
-      "redux",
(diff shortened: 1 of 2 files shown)


jdetle:
Shouldn't there be a vendor.[hash].js removed from an index.html somewhere?

lgeiger:
> Shouldn't there be a vendor.[hash].js removed from an index.html somewhere?

Ahh you're right. Guess this needs some more cleanup.

Closing for now

jdetle:
Alright, thanks for the PR. Hope you're doing well!

rgbkrk:
You know, I think we probably can remove `../lib/vendor.js` from the index.html as well. I noticed on the jupyter extension that app ended up with the dependency as well. What we _really_ need to do is make sure chunks are able to be created and loaded on demand via webpack.

jdetle:
Is that the same as dynamic imports?

rgbkrk:
You need chunks to load properly if you're going to use dynamic imports, they're not the same thing though.

PR #2974, opened 2018-06-12
more bits for the changelog

rgbkrk:
Added sections in the changelog for vega + open recent, updated the gist publishing with emojis. 🍣 🌱 

cc @piercefreeman -- I added a tiny blurb about "Open Recent" to the changelog. I'm also going to send over a PR because I'm not seeing updates to the menu to show recent. Not sure if that's a dev-copy only thing or if there's some other issue.


alexandercbooth:
Thanks!

PR #2975, opened 2018-06-12
switch recentdocuments casing

rgbkrk:
In my hope to see recent documents working on my machine I made the casing match that of
https://github.com/electron/electron/blob/18362eb948ef91e002e286a835663fd6a3977048/lib/browser/api/menu-item-roles.js

I'm still only seeing recently opened documents in the dock though, so maybe we're running into https://github.com/electron/electron/pull/11166#issuecomment-394115690 ?

~~Perhaps there is an even later version of electron we have to upgrade to?~~ Apparently 2.0.2 is the latest.

cc @piercefreeman

### DIFF
--- a/applications/desktop/src/main/menu.js
+++ b/applications/desktop/src/main/menu.js
@@ -392,19 +392,17 @@ export function loadFullMenu(store: * = global.store) {
         role: "close"
       }
     );
-  }
-  else if (process.platform === "darwin") {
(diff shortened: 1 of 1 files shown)


rgbkrk:
I'll go ahead and assume lowercase is the right thing considering what I see in Atom's source. We definitely need to work on this feature some more. Thanks @piercefreeman, I'm looking forward to using this. 😄

PR #2978, opened 2018-06-13
check subprocess when using jupyter nteract --dev

rgbkrk:
This checks to see if the webpack-dev-server has exited early during the initial startup when using `jupyter nteract --dev`. Additionally, it waits just long enough (at least for the cases I'm on) so that you don't have to refresh your `localhost:8888/nteract/edit` page, webpack will reload the page once the bundle is ready.

Also took the liberty of switching the ports so it doesn't overlap with RethinkDB's admin page nor other default webpack servers.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/nteractapp.py
+++ b/applications/jupyter-extension/nteract_on_jupyter/nteractapp.py
@@ -4,21 +4,25 @@ from subprocess import Popen
 from notebook.notebookapp import NotebookApp, flags
 from traitlets import Unicode, Bool
 
+import time
+
 
 from . import EXT_NAME, PACKAGE_DIR
 from .config import NteractConfig
 from .extension import load_jupyter_server_extension
 from .utils import cmd_in_new_dir
 
-webpack_hot = {"address": 'http://localhost:8080/',
-               "command":["lerna", "run", "hot",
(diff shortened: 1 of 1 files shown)


applications/jupyter-extension/nteract_on_jupyter/nteractapp.py
```diff
@@ -41,14 +46,13 @@ class NteractApp(NotebookApp):
...
-    help="""Whether to start the app in dev mode. Expects resources to be loaded
+                    help="""Whether to start the app in dev mode. Expects resources to be loaded
     from webpack's hot reloading server at {address}. Run
     `{command}`
     To serve your assets.
     This is only useful if NteractApp is installed editably e.g., using `pip install -e .`.
     """.format(address=webpack_hot["address"],
-               command=" ".join(webpack_hot["command"])))
-
+                        command=" ".join(webpack_hot["command"])))
```
mpacer: could we unindent this line to what it was before?

rgbkrk: Yeah, I can switch it back. `yapf` was automatically doing this (that or Atom). I'd love for us to adopt [black](https://github.com/ambv/black), as recommended by @willingc so that we have one consistent style.

PR #2979, opened 2018-06-13
Add gatsby-transformer-ipynb package

ghost:
This package relies on `@nteract/notebook-render` component to generate
html from a notebook. It is added here as discussed on #2947

### DIFF
--- a/packages/gatsby-transformer-ipynb/.gitignore
+++ b/packages/gatsby-transformer-ipynb/.gitignore
@@ -1,0 +1,13 @@
+# Project dependencies
+.cache
+node_modules
+yarn-error.log
+
+# Builds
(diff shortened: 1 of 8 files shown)


ghost:
@rgbkrk I've moved the gatsby-plugin here, as you proposed it on #2947. I also updated the example site at https://github.com/gatsby-contrib/gatsby-transformer-ipynb/tree/move-to-nteract-monorepo.
I'd like to publish it to npm, but I don't know if it is possible to publish packages in the monorepo separately.

rgbkrk:
Great, I'll get this in and go ahead and do the publish. The monorepo does have to publish together, however it will only publish packages with changes. I've been planning on releasing today anyways. 😄

PR #2992, opened 2018-06-14
[WIP] upgrade monaco

rgbkrk:
~~Upgrades monaco so that web workers get loaded even when the notebook server is running on a subpath, like with JupyterHub!~~ 🎉  @yuvipanda @minrk @willingc 

~~This makes for a very smooth experience when editing a file.~~ 🥇 

Just kidding. 

screen shot 2018-06-14 at 2 22 36 pm

I'll look into this.


rgbkrk:
Closing for now since this didn't work out as intended and isn't worth a bother at the moment.

PR #3011, opened 2018-06-20
feat:  add directory-listing package

### LINKED ISSUE #2985: [jext] Directory Listing Styling
Some feedback I've gotten from folks about the directory listing, that I agree with:

* Make the font of the directory listing smaller than it currently is (pretty easy)
* People like the directory listing UI that is in commuter
  * I originally intended to use the same component, then realized I couldn't do a straight `mv` because of differences in routing (`Link` from next.js).

My general recommendation is to create a directory listing package as part of the monorepo, creating them as compound components like this:

```jsx
<DirectoryListing>
  <Entry>
    <Entry.Icon type="notebook" />
    <Entry.Name>
      <a href="..." /> 
    </Entry.Name>
    <Entry.Time>
      <TimeAgo date={row.last_modified} />
    </Entry.Time>
  </Entry>
</DirectoryListing>
```

alexandercbooth:
Will fix #2985.

This will add a new package to create a unified directory listing that can be utilized across nteract applications. 

Also makes the font smaller.

### Current Status
At the moment, this consists of a compound `<Entry>` component. This `<Entry>` component is then being utilized by the respective directory listings in Juptyer Extension and 🚗commuter to render the tree view. The next step is to create a `<DirectoryListing>` component directly that will be used instead of these individual directory listing components. 

### TODO
- [x] create `<DirectoryListing>` 
- [x] add tests 🙃

Note: the styling is currently crazy 😀 

Below is a snapshot of the tree views from both jext and commuter using the new package.
image
image


rgbkrk:
I can help tweak that styling to get it working out. What would you like me to start helping with first?

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+      `}</style>
+    </td>
+  );
+  render() {
+    return React.Children.map(this.props.children, childElement =>
+      React.cloneElement(childElement)
+    );
+  }
+}
```
rgbkrk: 👍

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,121 @@
...
+          text-align: right;
+          color: #6a737d;
+          padding-right: 10px;
+        }
+      `}</style>
+    </td>
+  );
+  render() {
+    return React.Children.map(this.props.children, childElement =>
+      React.cloneElement(childElement)
```
rgbkrk: You can just return `this.props.children` here since no props are being added on with `cloneElement`.

alexandercbooth: Oops, didn't realize that wasn't actually doing anything 😅

alexandercbooth:
Thanks @rgbkrk, that would be great! A couple of questions -- 

I am wondering about the design of `<DirectoryListing>`. Since commuter's directory-listing has only the tree view, it is easy enough to just incorporate that directly into this package (which I have done locally). If we do it that way, then for Jext, we could use the connected component we have already that would do the `NewNotebookNavigation` stuff plus get the actual contents of the directory and then be handed down as props to `<DirectoryListing contents={this.props.contents} basepath={this.props.host.basePath} />`. Is this the direction we want this to go in?

Could you also help me get started on testing  😬?

Styling help would also be much appreciated. With the compound component pattern I had to move it around and now it's just all over the place 🙃. Do we want to include a theme that can do jext and commuter styles (like exists in commuter)?

rgbkrk:
`NewNotebookNavigation` should be its own component, not part of the directory listing component.

The "directory listing 'app'" uses the navigator while also using the directory listing components.

The basepath stuff I need to fix with a proper routing setup. It's something we'll pass into the respective `<Link />`s or `<a />` we pass as children to an entry.

> Could you also help me get started on testing 😬?

Since these are visual components, I'd start with using [jest snapshots](https://facebook.github.io/jest/docs/en/snapshot-testing.html). That way you can write declarative tests and just make sure the "rendered" versions look like the right output.

rgbkrk:
As for styling, I'll come back to help you with this soon. 😄

alexandercbooth:
>NewNotebookNavigation should be its own component, not part of the directory listing component.

Sorry, I don't think I explained what I meant correctly -- I wasn't suggesting including the  `NewNotebookNavigation` as part of this package. What I meant with 
>we could use the connected component we have already that would do the NewNotebookNavigation stuff

was to use the existing directory listing app and include the new`<DirectoryListing contents={this.props.contents} />` from `@nteract/directory-listing` there.

>Since these are visual components, I'd start with using jest snapshots. That way you can write declarative tests and just make sure the "rendered" versions look like the right output.

Awesome, I will do this 😄

rgbkrk:
> <DirectoryListing contents={this.props.contents} />

As much as possible I'd like to see the components get used only in the "compound" sense from this one, which likely means writing `<DirectoryListing />` for each separately since they both use different routers (basepath handling, `Link`s, etc.).

alexandercbooth:
What about passing the routers and basepath as props instead? I tooled around a bit with that last night and it worked OK. It allows us to have just the one `<DirectoryListing>`, but I am not sure if this is a desirable or expected pattern 🤔. It might over complicate things unnecessarily, though 😬. I'll commit/push and let me know what you think.

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,119 @@
...
+
+import TimeAgo from "@nteract/timeago";
+import { Book, FileText, FileDirectory } from "@nteract/octicons";
+
+// type EntryProps = {
+//   fileType: "unknown" | "notebook" | "directory" | "file" | "dummy",
+//   link: any,
+//   last_modified: string
+// };
+
```
rgbkrk: This should be using some `props.children` here, so that the passing of:

```
<Entry>
  <Entry.Icon />
</Entry>
```

works

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,119 @@
...
+import * as React from "react";
+
+import TimeAgo from "@nteract/timeago";
+import { Book, FileText, FileDirectory } from "@nteract/octicons";
+
+// type EntryProps = {
+//   fileType: "unknown" | "notebook" | "directory" | "file" | "dummy",
+//   link: any,
+//   last_modified: string
+// };
```
rgbkrk: The EntryProps will be:

```
children: React.Node
```

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,119 @@
...
+        .timeago {
+          text-align: right;
+          color: #6a737d;
+          padding-right: 10px;
+        }
+      `}</style>
+    </td>
+  );
+  render() {
+    return this.props.children;
```
rgbkrk: I'm doing some thinking on the interface and it seems like we should make `<Entry>`'s render be:

```jsx
<tr>
  {children}
</tr>
```

so that it really is the row. The great thing about us having the `children` interface is that you can swap order of the icons, names, etc. by swapping where they appear in between `<Entry></Entry>`

applications/commuter/components/contents/directory-listing.js
```diff
@@ -93,33 +94,25 @@ const DirectoryListing = (props: DirectoryListingProps) => {
...
+                    query: { viewPath: row.path }
+                  }}
+                  as={base + "/" + row.path}
+                >
+                  <a>{row.name}</a>
+                </Link>
+              );
 
               return (
                 <tr key={index}>
```
rgbkrk: With the change suggested below (making `<Entry>` contain a `<tr>` will eliminate the need to use the `<tr>` here (and leaking the implementation details).

This code will then be:

```jsx
<Entry key={index}>
  <Entry.Icon fileType={row.type} />
  <Entry.Name link={link} />
  <Entry.LastSaved last_modified={row.last_modified} />
</Entry>
```

applications/commuter/components/contents/directory-listing.js
```diff
@@ -93,33 +94,25 @@ const DirectoryListing = (props: DirectoryListingProps) => {
               ) {
```
rgbkrk: The `<table>` and `<tbody>` portion of this should end up as some `<Listing>` component, whose render probably looks like this:

```jsx
class Listing extends React.Component {
  render() {
    return (
      <React.Fragment>
        <table>
          <tbody>{this.props.children}</tbody>
        </table>
        <style jsx>{`
          .table {
            ... all them styles ...
         }
        `}</style>
      </React.Fragment>
    );
  }
}

```

rgbkrk: Then it would get used like:

```jsx
<Listing>
  {contents.map((entry, index) => (
    <Entry key={index}>
      <Entry.Icon fileType={row.type} />
      <Entry.Name link={link} />
      <Entry.LastSaved last_modified={row.last_modified} />
    </Entry>
  ))}
</Listing>
```

packages/directory-listing/src/entry.js
alexandercbooth: 👍

packages/directory-listing/src/entry.js
alexandercbooth: 👍 I like how easy that makes moving them around too

applications/commuter/components/contents/directory-listing.js
alexandercbooth: Nice, this definitely makes things look cleaner! Do you mean we should use this `<Listing>` component in commuter's existing `directory-listing` (as opposed to the new one in this package)? Would we then use the `<Entry>...</Entry>` interface directly in Jext's existing `directory.js` (instead of using the new `<DirectoryListing>` from this package)?

rgbkrk: Yep! This is the view-only suite of directory components that can be used in these slightly overlapping apps.

alexandercbooth: OK gotcha, thanks! I'll give this a shot late tonight/tomorrow 🌙 😄

applications/commuter/components/contents/directory-listing.js
```diff
@@ -77,135 +76,32 @@ const DirectoryListing = (props: DirectoryListingProps) => {
...
+          );
+          return (
+            <Entry key={index}>
+              <Entry.Icon fileType={entry.type} />
+              <Entry.Name link={link} />
+              <Entry.LastSaved last_modified={entry.last_modified} />
+            </Entry>
+          );
+        })}
+      </Listing>
```
rgbkrk: This is definitely looking beautiful now. 👌🏻 👨‍🍳 💋

alexandercbooth:
Hm, not actually sure why this failed 🤔. I think this is starting to be in pretty good shape now, but styling isn't quite right yet.

applications/commuter/components/contents/directory-listing.js
alexandercbooth: Thank you 😁 🎉

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,85 @@
...
+  static Icon = ({
+    fileType
+  }: {
+    fileType: "unknown" | "notebook" | "directory" | "file" | "dummy"
+  }) =>
+    fileType === "notebook" ? (
+      <td className="icon">
+        <Book />
+        <style jsx>{`
+          .icon {
```
rgbkrk: The styles are all the same here so I'd make this component be:

```jsx
let icon = <FileText />;
switch (fileType) {
  case "notebook":
    icon = <Book />;
    break;
  case "directory":
    icon = <FileDirectory />;
    break;
  case "file":
    icon = <FileText />;
    break;
  default:
    icon = <FileText />;
}

return (
  <td className="icon">
    {icon}
    <style jsx>{`
      .icon {
        padding-right: 2px;
        padding-left: 10px;
        width: 17px;
        vertical-align: middle;
        text-align: center;
        opacity: 0.95;
      }
    `}</style>
  </td>
);
```

applications/commuter/components/contents/directory-listing.js
alexandercbooth: I really appreciate all the help on this. It can be difficult for me to see the larger picture sometimes, and I think this really looks/feels much cleaner than what I was initially thinking 😄

packages/directory-listing/src/entry.js
```diff
@@ -0,0 +1,85 @@
...
+        .timeago {
+          text-align: right;
+          color: #6a737d;
+          padding-right: 10px;
+        }
+      `}</style>
+    </td>
+  );
+  render() {
+    return <tr>{this.props.children}</tr>;
```
rgbkrk: I'm beginning to wonder if we can make our interface a bit more generic by doing this:

```jsx
return (
  <tr className="directory-entry">
    {React.Children.map(this.props.children, child => (
      <td className="directory-entry-field">{child}</td>
    ))}
  </tr>
);
```

or even just detect if it already is a `td`:

```jsx
return (
  <tr className="directory-entry">
    {React.Children.map(this.props.children, child => {
      if (child.type === "td") {
        // Already a <td>, pass it on through
        return child;
        // Could always supplement this child with an additional className too for CSS reachability
        //
        // return React.cloneElement(child, {
        //   className:
        //     typeof child.props.className === string &&
        //     child.props.className !== ""
        //       ? child.props.className + " directory-entry-field"
        //       : "directory-entry-field"
        // });
      } else {
        return <td className="directory-entry-field">{child}</td>;
      }
    })}
  </tr>
);
```

This will make it easier to drop in new fields without having to change the directory package.

```jsx
<Listing>
  <Entry>
    <Icon />
    <pre>💥</pre>
    <LastSaved />
  </Entry>
</Listing>
```

rgbkrk:
Now that we're super close on interface, I'd say it's time to add this to the styleguidist so it can be documented and experimented with as an interface.

cc @willingc 😄

alexandercbooth:
99.999999% there, but I cannot seem to get the scoping right on the styles for `<Listing>`. I can only make it work by adding global, but I know that's not right 🙃. What's the way out here?

rgbkrk:
Oooook, let's give it a shot. I'm cloning this locally now. I'll also help get it into the styleguide.

It's totally fine by the way if you use `:global` as it's still scoped to just that component and its children.

rgbkrk:
How's this?

screen shot 2018-06-27 at 8 47 50 am

Needs more style changes I think. This just relocates the styles. The diff:


```diff
diff --git a/packages/directory-listing/src/entry.js b/packages/directory-listing/src/entry.js
index 03c7dc7c..1af15c21 100644
--- a/packages/directory-listing/src/entry.js
+++ b/packages/directory-listing/src/entry.js
@@ -54,6 +54,19 @@ export default class Entry extends React.Component<EntryProps, *> {
           vertical-align: middle;
           font-size: 0.9em;
         }
+
+        :global(a) {
+          text-decoration: none;
+        }
+
+        :global(a:hover) {
+          text-decoration: underline;
+          outline-width: 0;
+        }
+        :global(a:hover) {
+          outline-width: 0;
+          text-decoration: underline;
+        }
       `}</style>
     </td>
   );
@@ -89,6 +102,24 @@ export default class Entry extends React.Component<EntryProps, *> {
             return <td className="directory-entry-field">{child}</td>;
           }
         })}
+        <style jsx>{`
+          tr {
+            border-top: 1px solid #eaecef;
+          }
+
+          tr:hover {
+            background-color: #f6f8fa;
+            transition: background-color 0.1s ease-out;
+          }
+
+          tr:first-child {
+            border-top: none;
+          }
+
+          tr:last-child {
+            border-bottom: none;
+          }
+        `}</style>
       </tr>
     );
   }
diff --git a/packages/directory-listing/src/listing.js b/packages/directory-listing/src/listing.js
index 4f64b015..f24f12c5 100644
--- a/packages/directory-listing/src/listing.js
+++ b/packages/directory-listing/src/listing.js
@@ -13,43 +13,12 @@ export default class Listing extends React.Component<ListingProps> {
           </table>
         </div>
         <style jsx>{`
-          tr {
-            border-top: 1px solid #eaecef;
-          }
-
-          td a {
-            text-decoration: none;
-          }
-
-          td a:hover {
-            text-decoration: underline;
-            outline-width: 0;
-          }
-
-          tr:hover {
-            background-color: #f6f8fa;
-            transition: background-color 0.1s ease-out;
-          }
-
-          tr:first-child {
-            border-top: none;
-          }
-
-          tr:last-child {
-            border-bottom: none;
-          }
-
           table {
             width: 100%;
             border-collapse: collapse;
             border-radius: 2px;
             border-spacing: 0;
           }
-
-          td a:hover {
-            outline-width: 0;
-            text-decoration: underline;
-          }
         `}</style>
       </React.Fragment>
     );
```

rgbkrk:
I'll go ahead and tack this on as a commit.

alexandercbooth:
That is awesome, thanks Kyle!

alexandercbooth:
### Commuter
image
### Jext
image

rgbkrk:
Slick. I think we should use commuter's blues inside the directory listing too (I think ?)

alexandercbooth:
Ahh good thinking. I added padding to match commuter too -->

image

image

rgbkrk:
I played around with these components inside of the react style guide (I'll add a commit once you're ready) and noticed that we should change a few things:

`<Entry.Name>` should take `children` rather than as a `link` prop, especially since we really use it as a child element.

We should export the components as named exports `export class Entry` for each separate component. Also! We should structure the directory like this:

```
$ tree
directory-listing
└── src
    ├── components
    │   ├── entry.js
    │   ├── entry.md
    │   ├── listing.js
    │   └── listing.md
    └── index.js
```

alexandercbooth:
><Entry.Name> should take children rather than as a link prop, especially since we really use it as a child element.

That definitely makes sense, thanks! And that structure is cleaner/more professional 👍.

Are `entry.md` and `listing.md` for the react style guide?

alexandercbooth:
Rats, I forgot that TODO was still there and boty 🤖 is going to be yelling at me with each commit.

rgbkrk:
> Are entry.md and listing.md for the react style guide?

Yeah. That's where you document how to use the components, using [react styleguidist](https://github.com/styleguidist/react-styleguidist).

alexandercbooth:
Ahh, gotcha. I can give that a shot 👍

alexandercbooth:
After discussing offline with @rgbkrk, I decided to try refactoring this so that Name, Icon and LastSaved would become separate components outside of Entry. Although a little larger, I think I like this much better 🤗. The type annotations are more clear and the relationship between the components more explicit.

packages/directory-listing/src/components/icon.md
```diff
@@ -0,0 +1,8 @@
+```jsx static
+import { Icon } from "@nteract/directory-listing"
+```
+Display an icon
+```jsx
+const { Icon } = require("../");
```
rgbkrk, line 6: When the component is the one that matches the file you don't have to do the require, you can use it directly. 😄 

```jsx
<Icon fileType="notebook" />
```

applications/commuter/components/contents/directory-listing.js
```diff
@@ -77,135 +81,32 @@ const DirectoryListing = (props: DirectoryListingProps) => {
...
+              as={base + "/" + entry.path}
+            >
+              <a>{entry.name}</a>
+            </Link>
+          );
+          return (
+            <Entry key={index}>
+              <Icon fileType={entry.type} />
+              <Name>{link}</Name>
+              <LastSaved last_modified={entry.last_modified} />
```
rgbkrk, line 105: Super legit

rgbkrk (approved):

rgbkrk:
I think at this point it's cool to take the WIP tag off and we can merge this!

packages/directory-listing/src/components/icon.md
alexandercbooth: Oh awesome, I didn't realize that -- will remove that this weekend 👍

applications/commuter/components/contents/directory-listing.js
alexandercbooth: 😄

alexandercbooth:
Great, will do. Thanks so much for all the help on this, Kyle!
:shipit:

packages/directory-listing/src/components/icon.md
rgbkrk: Ok cool, that can be a new PR. 😄

rgbkrk:
Away we go!

rgbkrk:
Heh, that TODO issue now exists. It's probably a good thing though. 😅

PR #3013, opened 2018-06-21
Display refactor

rgbkrk:
Closes #2984, in a workable way for now. We can make this much smoother in the near term future though by:

* Switching commutable to using `@nteract/records`
* Relying on those same records in the output react components
* Only passing necessary data to outputs that need them (models are one set)

~~To make this work for now I had to disable some functionality (models being the sneaky one). I've got to figure out a good path for that and then I can make sure this is well situated to iterate with.~~ All set -- `models` get to be immutable up until consumed by a component, protected for perf via `shouldComponentUpdate`. 😍 

NOTE! This kind of deprecates using `<DisplayArea outputs={...} expanded />` over using:

```js
<Outputs expanded>
  <Output />
</Outputs>
```

Yay for compound components!


rgbkrk:
Wow, I'm much happier with this and at a much quicker pace than I expected! Care to check it out @alexandercbooth?

jdetle:
I checked it out and it ran great on my notebooks. It looks like some snapshots need to be updated before a merge.

rgbkrk:
Thank you so much for checking this out John, it gives me much more confidence to have another pair of eyes on it!

rgbkrk:
Made sure Immutable was included as part of display area so the release goes well.

alexandercbooth:
Works great on my end, too. Awesome to have a fix for this 👍

PR #3027, opened 2018-06-26
Fix props access in transform-plotly

gnestor:
Fixes `TypeError: Cannot read property 'layout' of undefined` error when `layout` is not passed to PlotlyTransform.


alexandercbooth:
Thanks @gnestor!

rgbkrk:
Love when PRs go through while I'm not watching. 😊

PR #3057, opened 2018-07-08
chore: remove unnecessary imports in styleguidist

alexandercbooth:
Follow up from #3011

### DIFF
--- a/packages/directory-listing/src/components/entry.md
+++ b/packages/directory-listing/src/components/entry.md
@@ -7,7 +7,6 @@ This component is used to create individual entries in a directory. It is not me
 
 Display an Icon, Name, and time since last saved of an entry in a directory.
 ```jsx
-const { Entry, Icon, Name, LastSaved } = require("../");
 const link = (
                 <a href={"http://nteract.io"}>
                   {"Example-Notebook.ipynb"}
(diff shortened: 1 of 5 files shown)


packages/directory-listing/src/components/entry.md
```diff
@@ -7,7 +7,6 @@ This component is used to create individual entries in a directory. It is not me
 
 Display an Icon, Name, and time since last saved of an entry in a directory.
 ```jsx
-const { Entry, Icon, Name, LastSaved } = require("../");
```
rgbkrk, line 10: I think `Icon`, `Name`, and `LastSaved` are still required (unless I'm mistaken), since they're not `Entry`.

rgbkrk:
Just merging this and seeing how it goes!

packages/directory-listing/src/components/entry.md
rgbkrk: I was totally mistaken. This is awesome! It works if they're in the same directory. ❤️

rgbkrk: We can totally make the presentational components simpler now too! 😄

alexandercbooth: Definitely 👍! I think I am loving _everything_ about react styleguidist 😄

PR #3060, opened 2018-07-09
provide a docs:promote command to alias the deploy

rgbkrk:
Now to build and promote the docs we run:

```
npm run docs:deploy
```

Then check the site it returns to make sure all went well. If it's working, promote it!

```
npm run docs:promote
```

:boom: It will now be live at components.nteract.io

------

~~NOTE: I'm trying the now integration to see what it's like to check out a now deployment on every PR.~~ 😈

PR #3061, opened 2018-07-09
[WIP] Promote docs

rgbkrk:
Just like #3060, except now ▲ a deployment will happen.

🙃

### DIFF
--- a/now.json
+++ b/now.json
@@ -1,0 +1,7 @@
+{
+  "name": "nteract-components",
+  "alias": "components.nteract.io",
+  "static": {
+    "public": "styleguide"
+  }
+}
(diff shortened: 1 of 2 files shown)


rgbkrk:
> Saving deployment image (218.0M)

LOL, this is a terrible idea.

rgbkrk:
If we just moved the electron app out to another repo, we'd sure have a differently shaped monorepo. 😑

rgbkrk:
What I was hoping for with this would be that we could continuously deploy sample versions of our UI, backed by binder for kernel compute, that we can tinker with. May have to build something a bit custom for this to work. 🤔

PR #3062, opened 2018-07-09
npm run start now launches a built copy of docs

rgbkrk:
Like #3061 with more gusto. Let's see how this is.

Still going to be a long install since it's doing the full monorepo.


rgbkrk:
Oh yeah, the build is still huge. I'm thinking the only reasonable option we might have is to take the built apps, put them in circle ci artifacts, and then publish to `now`.

rauchg:
@rgbkrk please let me know if we need to tweak something on our end. How long do your builds usually take?

rgbkrk:
> How long do your builds usually take?

So long as we're not building the desktop app, it's under 4 minutes.

This repo is a big monorepo with ~3 next apps, a hefty desktop app with native deps, and the static styleguidist setup I was hoping to push out (which ends up being _tiny_). If anything, I feel bad about all the things I need to tweak on my end. I would _love_ if we could deploy several next.js apps from our monorepo.

Happy to collaborate to make our overall setup easy for collaborators and make full use of zeit!

PR #3070, opened 2018-07-11
feat: create tray launcher

alexandercbooth:
For users that consume mountains of notebooks, this might be helpful 😄 

Thoughts on menu items/structure/wording? Below are some screen shots. ⬇️ 

image

At the moment, I don't think the logo is a high enough resolution 😬

image


rgbkrk:
Super cool! I think I can trim down that logo so it's only the icon and give you a higher resolution version. This gives a good idea of it though!

applications/desktop/src/main/index.js
```diff
@@ -244,14 +251,19 @@ openFile$
...
 fullAppReady$.subscribe(() => {
   kernelSpecsPromise
     .then(kernelSpecs => {
       if (Object.keys(kernelSpecs).length !== 0) {
         store.dispatch(setKernelSpecs(kernelSpecs));
         const menu = loadFullMenu();
         Menu.setApplicationMenu(menu);
+        const trayImage = join(__dirname, "..", "static", "nteractlogo.png");
+        tray = new Tray(trayImage);
+        tray.setToolTip("nteract Quick Launch");
```
rgbkrk: I was wondering what other apps and native mac utilities report in their tooltip and only my VPN shows one. Based on that, I'm not sure if we should.

My thinking here is totally counter to what I usually say with web tools and toolbars. 🤷‍♂️

applications/desktop/src/main/menu.js
```diff
@@ -653,3 +653,128 @@ export function loadFullMenu(store: * = global.store) {
...
+        }
+      ]
+    });
+  }
+
+  template.push(fileWithNew);
+
+  const menu = Menu.buildFromTemplate(template);
+  return menu;
+}
```
rgbkrk, line 780: 👏

applications/desktop/src/main/index.js
alexandercbooth: That's a good point, none of the others on my machine show anything either. Will remove 👍

applications/desktop/src/main/menu.js
```diff
@@ -653,3 +653,128 @@ export function loadFullMenu(store: * = global.store) {
   const menu = Menu.buildFromTemplate(template);
   return menu;
 }
+
+export function loadTrayMenu(store: * = global.store) {
+  // NOTE for those looking for selectors -- this state is not the same as the
+  //      "core" state -- it's a main process side model in the electron app
```
BenRussert, line 659: This is a good tip, maybe we should remember it for the [dev docs](https://github.com/nteract/global-sprint/blob/master/projects/docs.md#devdocsnteractio)

applications/desktop/src/main/index.js
```diff
@@ -244,14 +251,18 @@ openFile$
...
-
+let tray = null;
 fullAppReady$.subscribe(() => {
   kernelSpecsPromise
     .then(kernelSpecs => {
       if (Object.keys(kernelSpecs).length !== 0) {
         store.dispatch(setKernelSpecs(kernelSpecs));
         const menu = loadFullMenu();
         Menu.setApplicationMenu(menu);
+        const trayImage = join(__dirname, "..", "static", "nteractlogo.png");
```
rgbkrk: It looks like we can create multiple resolutions of this same image: https://electronjs.org/docs/api/native-image#high-resolution-image

alexandercbooth:
## Update
- fixed logo and much sharper 👍 😎 
- included white logo for dark menu bar and windows

image


image

Thanks for the feedback and looking this over, @rgbkrk and @BenRussert!

rgbkrk:
Slick.

rgbkrk:
Now that I'm using this, can we make the launcher just be "New" and "Open" instead of "Launch New Notebook" then "New", "Open", etc.?

alexandercbooth:
Certainly!

PR #3076, opened 2018-07-13
chore: bump electron builder

alexandercbooth:
Closes #3072.

Not sure if this will work, but it seems like electron-builder tries to publish if CI is detected (I can run dist locally fine, but it fails if I force a `-p always`). We'll see ¯\\\_(ツ)_/¯


rgbkrk (approved):
Looking good to me.

alexandercbooth:
Nice, thanks!

PR #3081, opened 2018-07-14
Include missing deps

rgbkrk:
Commuter coudln't be deployed in production uses because of missing `react-hot-loader`.


alexandercbooth (approved):
Thanks!

PR #3085, opened 2018-07-15
Fix icons on linux

BenRussert:
- fix desktop icons not displaying on linux
- see https://github.com/electron-userland/electron-builder/issues/2577


BenRussert:
Desktop shortcut icons are not showing on Ubuntu. It looks like the top image circled in the screenshot before and the correct icon after.

linux-icons


Btw: I don't know if this will break the dist for windows/mac, but I don't think it would.

alexandercbooth:
Thanks for tackling this, @BenRussert! I will test this out on mac tonight.

alexandercbooth (approved):
Looks good on mac, too. Thanks @BenRussert!

PR #3106, opened 2018-07-21
Handle zero dimension issues

emeeks:
* Don't crash views when there are no dimensions in the DF
* Treat boolean as a possible dimension (toString on the uniques)

### DIFF
--- a/packages/transform-dataresource/src/charts.js
+++ b/packages/transform-dataresource/src/charts.js
@@ -404,7 +404,12 @@ const semioticBarChart = (
 
   if (dim1 && dim1 !== "none") {
     const uniqueValues = sortedData.reduce(
-      (p, c) => (!p.find(d => d === c[dim1]) && [...p, c[dim1]]) || p,
+      (p, c) =>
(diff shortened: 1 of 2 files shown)


packages/transform-dataresource/src/charts.js
```diff
@@ -404,7 +404,12 @@ const semioticBarChart = (
 
   if (dim1 && dim1 !== "none") {
     const uniqueValues = sortedData.reduce(
-      (p, c) => (!p.find(d => d === c[dim1]) && [...p, c[dim1]]) || p,
+      (p, c) =>
```
rgbkrk, line 407: I realize this is notation from before -- what do the `p` and `c` signify?

emeeks: "Previous" an "Current" in the reducer, in the logic of the reducer function I'm just trying to reduce the data to unique values baed on the selected dimension for coloring (`dim1`) and using the spread syntax lets me do it in a single line, which looks cool but if it's not the best approach lmk.

PR #3110, opened 2018-07-21
enable react hot loader only for packages in need

### LINKED ISSUE #2998: Published modules require react-hot-loader
The new release of `@nteract/` packages requires `react-hot-loader` as a dependency due to the [babel config](https://github.com/nteract/nteract/blob/master/.babelrc#L4) used during publishing.

We should remove this babel plugin for the published packages.

Checkout https://github.com/nteract/hydrogen/issues/1335, https://github.com/nteract/hydrogen/pull/1336
Also related to #2977

rgbkrk:
Closes #2998

I'm also going to fix up the regenerator runtime and styled-jsx setups.

### DIFF
--- a/applications/commuter/.babelrc
+++ b/applications/commuter/.babelrc
@@ -11,7 +11,6 @@
     "next/babel"
   ],
   "plugins": [
-    "react-hot-loader/babel",
     "transform-class-properties",
     "transform-flow-strip-types",
     "transform-object-rest-spread"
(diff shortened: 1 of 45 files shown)


rgbkrk:
Another thing that can be cleaned up is the dependence on `@nteract/types` which we ended up migrating.

rgbkrk:
Ok, this is ready for a review. I included `babel-runtime` as a dependency in all the packages based on https://babeljs.io/docs/en/babel-plugin-transform-runtime.html.

cc @jdpigeon @lgeiger

rgbkrk:
I'm going to err on the side of getting this out and shipped since we're in a current funky state. Thanks all for reporting this and helping make external usage of the packages easier. 😇

PR #3121, opened 2018-07-27
No primary keys fix

### LINKED ISSUE #3005: Data explorer crashes with a notebook (that I think is ok)
I'm creating [this](https://gist.github.com/davidanthoff/f17f89736dc9016d209d66047ceefd45) notebook from within julia. It has the tabular data stored as a ``application/vnd.dataresource+json``, so I would have expected it to show up nicely in nteract 0.9.

Instead, I get the following error:

error

I guess either I'm output incorrect ``application/vnd.dataresource+json``, or there is a bug somewhere else? Any pointer would be greatly appreciated, I can't wait to fully enable this scenario!

emeeks:
Should resolve https://github.com/nteract/nteract/issues/3005

### DIFF
--- a/packages/transform-dataresource/package.json
+++ b/packages/transform-dataresource/package.json
@@ -40,7 +40,6 @@
     "react-virtualized": "9.20.1",
     "semiotic": "^1.12.0",
     "tv4": "^1.3.0",
-    "numeral": "^2.0.6",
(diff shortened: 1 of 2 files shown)


rgbkrk:
Cool, let's get this shipped soon. Thanks!

Issue #3095, opened 2018-07-19
Handle Time Series Data Correctly


emeeks:
Data Explorer currently shows a line chart that is not time series, it's just a stacked bar chart kind of thing. Instead there should be a way to recognize and handle time series data. I don't know what typical "time series data" looks like in a df and whether this requires recognizing dates as a separate data type from strings and numbers or otherwise.

PR #3125, opened 2018-07-27
Handle datetime, allow sorting

emeeks:
* recognize datetime type and put it in both metrics and dimensions
* parse datetime strings into js dates when data is first sent to data transform
* allow sorting by metrics in line chart
* format axes labels so when dates there's more space and they're rotated
* show x value (date or metric value) in tooltip

### DIFF
--- a/packages/transform-dataresource/src/charts.js
+++ b/packages/transform-dataresource/src/charts.js
@@ -2,7 +2,7 @@
 import * as React from "react";
 
 import { nest } from "d3-collection";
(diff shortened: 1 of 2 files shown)


emeeks:
Should resolve https://github.com/nteract/nteract/issues/3095

emeeks:
Almost forgot to post an image.

screen shot 2018-07-27 at 3 21 49 pm

rgbkrk:
Whoa 😮

PR #3132, opened 2018-07-28
Address source hidden toolbar issue.

jdetle:
This adds a source hidden attribute to the toolbar so that a minimized cell cant be lost in the document.


packages/notebook-app-component/src/toolbar.js
```diff
@@ -190,7 +191,7 @@ export class PureToolbar extends React.Component<PureToolbarProps> {
           }
 
           .cell-toolbar-mask {
-            display: none;
+            display: ${sourceHidden ? "block" : "none"};
```
rgbkrk, line 194: This works. We could instead return `null` in render when hidden though.

PR #3140, opened 2018-07-29
Start with the strictification.

jdetle:

### DIFF
--- a/packages/notebook-app-component/src/cell-creator.js
+++ b/packages/notebook-app-component/src/cell-creator.js
@@ -146,7 +146,7 @@ class CellCreator extends React.Component<ConnectedProps> {
       contentRef
     } = this.props;
 
-    if (id == null || id.length === 0) {
+    if (id == null || typeof id != "string") {
       createCellAppend({ cellType: type, contentRef });
       return;
     }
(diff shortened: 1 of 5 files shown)


packages/notebook-app-component/src/cell-creator.js
```diff
@@ -159,21 +146,21 @@ class CellCreator extends React.Component<ConnectedProps> {
...
       : createCellAfter({ cellType: type, id, source: "", contentRef });
-  }
+  };
 
   mergeCell(): void {
     const { mergeCellAfter, id, contentRef } = this.props;
 
     // We can't merge cells if we don't have a cell ID
-    if (id) {
+    if (id != null && id.length === 0) {
```
rgbkrk: Could we use `typeof id === 'string'` instead of checking length?

The reason I ask is because if I'm reading this my brain immediately wonders if it's checking to make sure that `id` is long enough (as if it's a container, like an array).

packages/notebook-app-component/src/draggable-cell.js
```diff
@@ -39,13 +48,13 @@ const cellDragPreviewImage = [
 
 type Props = {|
   connectDragPreview: (img: Image) => void,
-  connectDragSource: (el: ?React$Element<any>) => void,
-  connectDropTarget: (el: ?React$Element<any>) => void,
+  connectDragSource: (el: ?React$Element<*>) => void,
+  connectDropTarget: (el: ?React$Element<*>) => void,
```
rgbkrk: I wonder if these types are inside `react-dnd` and that we could use them here instead of having them written by us.

jdetle: Totally makes sense, fixed it

jdetle:
:shipit:

PR #3141, opened 2018-07-29
Remove immutable.js

BenRussert:


packages/records/src/outputs/error.js
```diff
@@ -47,23 +47,24 @@ type ErrorMessage = {
   }
 };
 
-export type ErrorOutputRecord = Immutable.RecordOf<ErrorOutput>;
+export type ErrorOutputRecord = Object;
```
rgbkrk, line 50: Was this meant to be `ErrorOutput`?

packages/records/src/outputs/stream.js
```diff
@@ -49,7 +48,7 @@ type StreamMessage = {
   }
 };
 
-export type StreamOutputRecord = Immutable.RecordOf<StreamOutput>;
+export type StreamOutputRecord = Object;
```
rgbkrk, line 51: We should just replace all usage of `StreamOutputRecord` with `StreamOutput` here.

packages/records/src/outputs/error.js
rgbkrk: Rather, like I said for stream, we should get rid of the `*Record` types and just use the data structure type since they should be one and the same as one JS object structure.

packages/records/src/outputs/stream.js
```diff
@@ -17,12 +17,11 @@ import * as common from "../common";
...
 
 export type StreamName = "stdout" | "stderr";
-
-export const STDOUT = "stdout";
-export const STDERR = "stderr";
-
 export type StreamType = "stream";
 
+export const STDOUT: StreamName = "stdout";
+export const STDERR: StreamName = "stderr";
```
rgbkrk, line 23: This is an improvement!

packages/records/src/outputs/display-data.js
```diff
@@ -23,14 +23,14 @@ export const DISPLAYDATA = "display_data";
 type DisplayDataOutput = {
   outputType: DisplayDataType,
   data: common.MimeBundle,
-  metadata: mixed
+  metadata: Object
```
jdetle, line 26:
```
{ ?collapsed: boolean, ?autoscroll: boolean | 'auto', ?deletable: boolean, ?format: string, ?name: string, ?tags: Array<string>, source_hidden: boolean, ?outputs_hidden: boolean, ?isolated: boolean }
```

PR #3172, opened 2018-08-01
remove dangling backticks

rgbkrk:


alexandercbooth (approved):
Good catch, thanks! 😄

PR #3175, opened 2018-08-01
npm -> yarn

eseiver:
updates dev install info to use yarn instead of npm


willingc:
Yay!!!!! Way to go @eseiver :cake: 🌊 :tada: :sparkles:

PR #3178, opened 2018-08-02
Clusterbar sometimes

emeeks:
Simple heuristics to choose cluster bar when cardinality is high.
Also update Semiotic so the tooltip works right

Beyond that, the only thing Zeppelin has is a pie chart, which we can implement at some point if there's public desire, and then a bunch of UI bits like dragging and dropping and other things, so this was focused on chart parity and the only thing missing from that perspective were grouped bar charts.

### DIFF
--- a/packages/transform-dataresource/src/charts/bar.js
+++ b/packages/transform-dataresource/src/charts/bar.js
@@ -92,8 +92,17 @@ export const semioticBarChart = (
     }
   }
 
+  //replace with incoming cardinality when df.describe metadata is implemented
+  const cardinality =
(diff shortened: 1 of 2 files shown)


packages/transform-dataresource/src/charts/bar.js
```diff
@@ -92,8 +92,17 @@ export const semioticBarChart = (
     }
   }
 
+  //replace with incoming cardinality when df.describe metadata is implemented
```
rgbkrk, line 95: We are on it!

rgbkrk (approved):

PR #3180, opened 2018-08-03
Hide scroll bar when cell is not in focus

emeeks:
Initial CSS to only allow scrolling in an overflowed cell when that cell is in focus. Adding some indication that text is being cut off seems like it will require some js comparing scrollHeight with clientHeight.

Data Explorer Grid uses `react-virtualized` container which uses js scrolling so also needs a custom solution to lock when grid is not in focus. We need to get the cell being in focus sent to the datatransform component to disable scrolling on it.


emeeks:
@rgbkrk Let me know what I need to do to regenerate the HTML so the tests don't fail on comparison.

rgbkrk:
```
yarn test --updateSnapshot
```

PR #3182, opened 2018-08-03
just the records

alexandercbooth:
After in person discussion with @rgbkrk and @mpacer at the summer sprints, we decided that the best way forward here is to get the changes to the records submitted separate from any implementation. 

While updating the records to immer, @BenRussert and I had started implementation of this package [here](#3151), [here](https://github.com/nteract/nteract/tree/output-components) and [here](https://github.com/alexandercbooth/nteract/tree/output-components).

Going forward:
- ~~test the conversion between these records `fromMsg`, `fromDisk`, `fromMem` (this way we can see if they are working without direct implementation)~~ I forgot, we did these as we went along
- implement reducers directly in the records package where it makes sense (e.g. appendOutput, stream munging)


rgbkrk:
Rebasing and helping move this forward now!

packages/records/src/common.js
```diff
@@ -6,6 +6,32 @@ import produce from "immer";
 export type MultilineString = string | Array<string>;
 
 export type OnDiskMimebundle = {
+  "text/plain"?: MultilineString,
```
rgbkrk, line 7: rockin'

PR #3183, opened 2018-08-03
move delete to the right of menu and run, put nonclickable bar inbetween

mpacer:
This fixes the issue where you can accidentally delete a notebook when trying to click either the run button or the menu, with delete sandwiched inbetween.

There isn't an iron icon that on it's own does this without modification (as far as I can tell) and it was easier to make it non-clickable to just include a span with " | ". I think it gets the quick fix out there.

image

### DIFF
--- a/packages/notebook-app-component/src/toolbar.js
+++ b/packages/notebook-app-component/src/toolbar.js
@@ -61,15 +61,6 @@ export class PureToolbar extends React.Component<PureToolbarProps> {
               </span>
             </button>
           )}
-          <button
-            onClick={removeCell}
-            title="delete cell"
-            className="deleteButton"
-          >
-            <span className="octicon">
-              <TrashOcticon />
-            </span>
-          </button>
           <DropdownMenu>
             <DropdownTrigger>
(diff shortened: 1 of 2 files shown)


packages/notebook-app-component/src/toolbar.js
```diff
@@ -149,6 +140,16 @@ export class PureToolbar extends React.Component<PureToolbarProps> {
               </DropdownContent>
             )}
           </DropdownMenu>
+          <span> | </span>
```
rgbkrk: Since right now this is darker than the other icons, we should get/make an SVG that is our divider here.

How's the dropdown menu look / work when we have this?

mpacer: the right part of the dropdown menu almost perfectly aligns with the cell's right boundary:

image

If you want i could add some more padding to see if I can get it more precisely correct.

mpacer:
Ok, updated to use span and boundary:

image

rgbkrk:
🎉  Thank you!

PR #3186, opened 2018-08-04
cleanup unused

rgbkrk:
Clean out unused packages and their footprint.


rgbkrk:
Woo! This cuts the cruft down heavily. 😄😃

PR #3198, opened 2018-08-07
Strictify nbextension

jdetle:

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/components/nav.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/components/nav.js
@@ -4,7 +4,7 @@ import * as React from "react";
 import { areComponentsEqual } from "react-hot-loader";
 
 type NavSectionProps = {
-  children: React.ChildrenArray<null | string | number | React.Element<*>>
+  children: React.Node
 };
 
 export class NavSection extends React.Component<NavSectionProps, null> {
--- a/applications/jupyter-extension/nteract_on_jupyter/app/components/nav.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/components/nav.js
@@ -1,10 +1,10 @@
-// @flow
+/* @flow strict */
 import * as React from "react";
 
 import { areComponentsEqual } from "react-hot-loader";
 
 type NavSectionProps = {
-  children: React.ChildrenArray<null | string | number | React.Element<any>>
+  children: React.ChildrenArray<null | string | number | React.Element<*>>
 };
 
 export class NavSection extends React.Component<NavSectionProps, null> {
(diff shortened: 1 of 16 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/components/nav.js
```diff
@@ -1,10 +1,10 @@
-// @flow
+/* @flow strict */
 import * as React from "react";
 
 import { areComponentsEqual } from "react-hot-loader";
 
 type NavSectionProps = {
-  children: React.ChildrenArray<null | string | number | React.Element<any>>
+  children: React.ChildrenArray<null | string | number | React.Element<*>>
```
rgbkrk: This _might_ be able to be simplified to `React.Node`.

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
```diff
@@ -47,7 +47,7 @@ type FileProps = {
   lastSavedStatement: string,
   saving: boolean,
   loading: boolean,
-  error: ?Object
+  error: ?{}
```
rgbkrk, line 50: What's the difference in these? Asking so I know how to apply this in the future. 😅

flow-typed/nb-format.js
```diff
@@ -13,3 +13,5 @@ declare type KernelSpec = {
 declare type KernelSpecs = {
   [key: string]: KernelSpec
 };
+
+declare type NotebookTypes = "unknown" | "directory" | "notebook" | "file";
```
rgbkrk, line 17: 👍

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
jdetle: Object is *any* object, this is an object that never gets accessed. It seems to be a shortcoming (or feature) of Flow that an Object that is never accessed, but only checked for existence, is essentially an empty object. I'm also slowly realizing that type doesn't do type inference unless you specifically tell it to, which is cool. I figure if we ever start using a `msg` key or something similar, then Flow will error, so it acts a bit as like our snapshot tests.

applications/jupyter-extension/nteract_on_jupyter/app/components/nav.js
jdetle: Good catch.

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
rgbkrk: Ahhhhh, yeah. In the long run we'll want to explicitly set the type for these (actual object structure) and we did a bit of cheating here with `Object` originally. 😄  I'll use `{}` now instead of `Object` from here on out when we haven't scoped it well.

Thanks @jdetle!

rgbkrk:
🎉

jdetle:
^ i live for those confetti emojis

PR #3206, opened 2018-08-07
Transform-Dataresource button styles

binx:
* Selected state for the chart icon button column
* Standardizing button display cross components
* Standardizing tooltip display


rgbkrk:
🎉  Slick!

PR #3208, opened 2018-08-08
create <TooltipContent /> as component

rgbkrk:
This extracts the `.tooltip-content` styles into a single wrapper component you can use like this:

```jsx
import TooltipContent from "./tooltip-content";

...

<TooltipContent>
  <h1>15</h1>
  <p>Country: Guatemala</p>
<TooltipContent>
```

cc @binx @emeeks

### DIFF
--- a/packages/transform-dataresource/src/ParallelCoordinatesController.js
+++ b/packages/transform-dataresource/src/ParallelCoordinatesController.js
@@ -7,6 +7,8 @@ import HTMLLegend from "./HTMLLegend";
 import { numeralFormatting } from "./utilities";
 import buttonGroupStyle from "./css/button-group";
 
+import TooltipContent from "./tooltip-content";
+
 type State = {
   filterMode: boolean,
   data: Array<Object>,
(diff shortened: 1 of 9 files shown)


emeeks (approved):

rgbkrk:
From chat between me and Elijah:

```
emeeks: So style-jsx is supposed to be an XML child of the parent it applies to
  but you still use CSS classes?

kylek: yes. you don't _have_ to use CSS classes there though. I could have done `div`
  and it would have been specific enough

emeeks: Nah because it would have hit every div in the tree, right?

kylek: because styled-jsx is going to create some gobbeldy gook name like `jsx-158723950`
  for that component's css

emeeks: I mean in the sub-tree.

kylek: It acts "shadow-dom" like, in that it will only target the elements that are
  declared right there. No children.

emeeks: Oh. What if you wanted CSS that affected children, then you use global?
  Or presumably some kind of `:children` proto-class?

kylek: when you apply `.tooltip-content :global(h3)` it affects all of the subtree `h3`

emeeks: Oh right I see that. so "global" within this subtree.

kylek: yeah it's a weird name. https://github.com/zeit/styled-jsx/tree/v2#how-it-works
```

PR #3210, opened 2018-08-08
target with :global only on nested selectors

rgbkrk:
largely this is so that the jsx-whateverID gets attached to the start like documented in https://github.com/zeit/styled-jsx/tree/v2#how-it-works

cc @alexandercbooth


rgbkrk:
This ended up affecting production deployment of commuter and was hard to track down (and there are other styled-jsx stylesheet issues I need to track down... 😞 )

PR #3211, opened 2018-08-08
More :global cleaning

rgbkrk:
Making sure we're targeting the right things here. I've mostly been testing by running desktop and making sure things look "right". I wish we had a better way to measure this stuff -- I suppose it means I need to do a better job reviewing these 😅 .

Behavior to know about with styled-jsx:

```jsx
<ul>
  {props.children}
</ul>
<style jsx>{`
  :global(li) {
    background-color: DeepPink;
  }
`}</style>
```

Will target _all_ `li` because there's no selector preceding it whereas:

```jsx
<ul>
  {props.children}
</ul>
<style jsx>{`
  ul :global(li) {
    background-color: DeepPink;
  }
`}</style>
```

will create `ul` with a className like `jsx-2352357` and the css becomes:

```css
ul.jsx-2352357 li {
    background-color: DeepPink;
}
```

[Further background on how styled-jsx works](https://github.com/zeit/styled-jsx/tree/v2#how-it-works)


rgbkrk:
Oops, I forgot a `ul` in my example up there. Edited the second block above to include `ul` in the style jsx section.

alexandercbooth (approved):
Great catch, @rgbkrk!

PR #3212, opened 2018-08-08
[WIP] set optimizeForSpeed to true to showcase issue

rgbkrk:
This PR is up merely to show an issue we're currently having with `styled-jsx`. If you run `npm run app:commuter` on this branch, open your devtools, and navigate between pages with the directory listing, you'll see the following warnings in the console:

screen shot 2018-08-08 at 12 14 20 pm

Navigate again and the whole page will break down with errors:

screen shot 2018-08-08 at 12 15 08 pm

In summary, it looks like Chrome is rejecting stylesheets that are being added (saying they're invalid) and then later styled-jsx appears to be trying to remove styles that are not there (due to never getting added properly).

This is styled-jsx 2.2.6 and affects both next 5 and next 6.

### DIFF
--- a/applications/commuter/.babelrc
+++ b/applications/commuter/.babelrc
@@ -8,7 +8,12 @@
         }
       }
     ],
-    "next/babel"
+    [
+      "next/babel",
+      {
+        "styled-jsx": { "optimizeForSpeed": true }
+      }
+    ]
   ],
   "plugins": [
     "react-hot-loader/babel",


rgbkrk:
Reported this in styled jsx as well: https://github.com/zeit/styled-jsx/issues/474

rgbkrk:
This was only a demonstration of a problem. We're now upgraded anyway and styled-jsx is going after the fix so I'm going to just close this PR.

PR #3217, opened 2018-08-09
trim naming boilerplate in records

rgbkrk:
This takes a new approach to setting up our base little types. To create a `displayData` record you:

```js
import { displayData } from "@nteract/records"

const dd = displayData()
```

converting from notebook format looks like:

```js
import { displayData } from "@nteract/records"

const dd = displayData.fromNbformat({
  output_type: "display_data",
  data: {
    "text/plain": [
      "mind\n",
      "time\n",
      "space\n",
      "reality\n",
      "power\n",
      "soul"
    ]
  },
  metadata: { "application/json": { expanded: true } }
});
```

similarly, there's `displayData.fromJupyterMessage`.

----

Posting this to see what folks think of the interface.

### DIFF
--- a/packages/records/__tests__/index-spec.js
+++ b/packages/records/__tests__/index-spec.js
@@ -1,5 +1,6 @@
 // @flow
 import * as nteractRecords from "@nteract/records";
+import { displayData } from "@nteract/records";
 
 describe("stream output", () => {
   test("can be converted from nbformat", () => {
@@ -42,7 +43,7 @@ describe("stream output", () => {
 describe("display_data output", () => {
   test("can be converted from nbformat", () => {
     expect(
-      nteractRecords.outputFromNbformat({
+      displayData.fromNbformat({
         output_type: "display_data",
         data: {
           "text/plain": [
@@ -57,7 +58,7 @@ describe("display_data output", () => {
         metadata: { "application/json": { expanded: true } }
       })
     ).toEqual(
-      nteractRecords.makeDisplayDataOutputRecord({
+      displayData({
         outputType: "display_data",
         data: {
           "text/plain": "mind\ntime\nspace\nreality\npower\nsoul"
@@ -71,7 +72,7 @@ describe("display_data output", () => {
 
   test("can be converted from jupyter messages", () => {
     expect(
-      nteractRecords.displayDataRecordFromMessage({
+      displayData.fromJupyterMessage({
         header: {
           msg_type: "display_data"
         },
@@ -81,7 +82,7 @@ describe("display_data output", () => {
         }
       })
     ).toEqual(
-      nteractRecords.makeDisplayDataOutputRecord({
+      displayData({
         outputType: "display_data",
         data: { "text/plain": "another\nDoug" },
         metadata: { "application/json": { expanded: true } }
--- a/packages/records/__tests__/index-spec.js
+++ b/packages/records/__tests__/index-spec.js
@@ -1,4 +1,4 @@
-// @flow
+// @flow strict
 import * as nteractRecords from "@nteract/records";
 
 import { streamOutput } from "@nteract/records/src/outputs/stream";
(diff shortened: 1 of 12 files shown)


alexandercbooth:
I like this interface! It feels clean 🚿 and looks much more explicit 👍

rgbkrk:
Ok awesome. I can go through making the same format for the other types too.

rgbkrk:
Here's what I would hope this comes out to look like:

```js
codeCell({
  source: "import pandas as pd; pd.read_csv(...)",
  outputs: [
    executeResult()
  ]
})
```

rgbkrk:
Ok this is ready for a review!

I ended up not deleting or mucking with the cell + notebook stuff here, so this is only about outputs.

packages/records/__tests__/index-spec.js
```diff
@@ -1,16 +1,21 @@
...
     expect(
-      nteractRecords.outputFromNbformat({
+      streamOutput.fromNbformat({
         output_type: "stream",
         name: "stdout",
         text: ["sup\n", "yall"]
       })
     ).toEqual(
-      nteractRecords.makeStreamOutputRecord({
+      streamOutput({
```
alexandercbooth, line 18: I am loving this new interface. It is so much clearer!

packages/records/__tests__/index-spec.js
```diff
@@ -30,7 +35,7 @@ describe("stream output", () => {
         }
       })
     ).toEqual(
-      nteractRecords.makeStreamOutputRecord({
+      streamOutput({
         outputType: "stream",
         name: "stdout",
         text: "it is love we must hold on to\nnever easy but we try"
```
alexandercbooth, line 41: >"it is love we must hold on to\nnever easy but we try"

Is this Beauty and the Beast :lion:?

packages/records/src/common.js
```diff
@@ -132,8 +132,7 @@ export function createImmutableMimeBundle(
       bundle[key] = demultiline(mimeBundle[key]);
     } else {
       // we now know it's an Object of some kind
-      // TODO: DeepFreeze the mimebundle object for @nteract/records
-      bundle[key] = Object.freeze(mimeBundle[key]);
+      bundle[key] = deepFreeze(mimeBundle[key]);
```
alexandercbooth, line 135: Ahh, this makes more sense to me now 👍

packages/records/src/freeze.js
```diff
@@ -0,0 +1,21 @@
+/* @flow strict */
+
+export function deepFreeze(object: *) {
```
alexandercbooth, line 3: ❄️

packages/records/src/outputs/stream.js
```diff
@@ -1,4 +1,4 @@
-// @flow
+// @flow strict
 
 import produce from "immer";
```
alexandercbooth: I think we can remove the `immer` import here and in `error.js`, since we are using `Object.freeze` instead. Is the reason for using `Object.freeze` that we don't need the stronger immutability of immer for `streamOutput` and `errorOutput` since we aren't modifying these outputs?

alexandercbooth (approved):
This is a great improvement, thanks @rgbkrk! 👍

packages/records/__tests__/index-spec.js
rgbkrk: Yeahhhhhh back when I wrote this it was playing in the background 😁

packages/records/src/outputs/stream.js
rgbkrk: In [development, `immer` does freezing for you](https://github.com/mweststrate/immer#performance). It might be better to just use `immer` like was being done before so we're not having to do it ourselves. 🤔

rgbkrk: Oh right, one of the reasons that I ended up not using `produce` was for `display_data` and `execute_result` where the mimebundle payloads were different. For flow to accept it, it had to know that we were matching the right type.

BenRussert (approved):
Merging with more ideas to come later as discussed on slack. (we have a new outputs-project slack channel, everyone is welcome to join)

PR #3221, opened 2018-08-11
Media Type & <RichMedia />

rgbkrk:
~~Super WIP here! Just posting so everyone knows which segment I'm touching.~~

Taking a stab at a new interface for the `<ExecuteResult />`, `<DisplayData />`, and `<RichMedia />`. This starts with _only_ `<RichMedia />` for the moment.

Demo: https://declarative-trial.nteract.io/#!/RichMedia

Here's a snippet of the new interface:

```jsx
// Declare a component for rendering plaintext
const Plain = props => <pre>{props.data}</pre>;
Plain.defaultProps = {
  mediaType: "text/plain"
};

// Declare a component for rendering HTML
const HTML = props => <div dangerouslySetInnerHTML={{ __html: props.data }} />;
HTML.defaultProps = {
  mediaType: "text/html"
};

<RichMedia
  data={{
    "text/plain": "plain was richer",
    "text/html": "<b>HTML is so much richer</b>"
  }}
>
  <HTML />
  <Plain />
</RichMedia>
```

Miscellaneous TODO as I come across them, so I don't forget:

* ~~`@nteract/presentational-components` now depends on `@nteract/records` solely for types -- need to figure out if this should be a hard dependency~~ included a simple type for now...

For later:

* Much of the styling of `<Outputs />` should only be affecting the html and markdown types. Right now it's this global grab bag of things.
* ~~We need to make it super easy to special case handling of the data explorer, to give it control over height~~ Hopefully this is good now with our declarative `<RichMedia>` component
* ~~How we do transforms should become simpler _somehow_~~ Possibly solved here
* ~~`onMetadataChange` should be passed in here~~ Hey I think we can at least do this with the Data Explorer with this setup.

### DIFF
--- a/packages/presentational-components/src/components/execute-result.js
+++ b/packages/presentational-components/src/components/execute-result.js
@@ -2,12 +2,44 @@
 
 import * as React from "react";
 
-import type { ExecuteResultOutput } from "@nteract/records";
+import type { MimeBundle } from "@nteract/records";
 
-type ExecuteResultProps = {} & ExecuteResultOutput;
+type ExecuteResultProps = {
+  /**
+   * The n-th result of execution
+   *
+   * Typically used to show what execution count the user is on. When working at
+   * the `IPython` or `jupyter console` for example, it's the number between the
+   * `[ ]` on the Out:
+   *
+   * ```
+   * In [1]: 2 + 2
+   * Out[1]: 4
+   *
+   * In [2]: "should be the 2"
+   * Out[2]: 'should be the 2'
+   * ```
+   *
+   */
+  executionCount: ?number,
+  /**
+   * Big ol' [payload of mimetype -> data](http://jupyter-client.readthedocs.io/en/stable/messaging.html#id6)
+   */
+  data: MimeBundle,
+  /**
+   * Grab bag of random things
+   */
+  metadata: {}
+};
 
 export class ExecuteResult extends React.Component<ExecuteResultProps, null> {
+  static defaultProps = {
+    executionCount: null,
+    data: {},
+    metadata: {}
+  };
+
   render() {
-    return <div />;
+    return <pre>{JSON.stringify(this.props.data, null, 2)}</pre>;
   }
 }
--- a/packages/presentational-components/src/components/execute-result.js
+++ b/packages/presentational-components/src/components/execute-result.js
@@ -29,7 +29,8 @@ type ExecuteResultProps = {
   /**
    * Grab bag of random things
    */
-  metadata: {}
+  metadata: {},
+  children: React.Node
 };
 
 export class ExecuteResult extends React.Component<ExecuteResultProps, null> {
@@ -40,6 +41,30 @@ export class ExecuteResult extends React.Component<ExecuteResultProps, null> {
   };
 
   render() {
-    return <pre>{JSON.stringify(this.props.data, null, 2)}</pre>;
+    let chosenOne = null;
+
+    const data = this.props.data;
+
+    React.Children.forEach(this.props.children, child => {
+      if (chosenOne) {
+        // Already have a selection
+        return;
+      }
+      if (child.props && child.props.mimetype && child.props.mimetype in data) {
+        chosenOne = child;
+        return;
+      }
+    });
+
+    if (chosenOne === null) {
+      return null;
+    }
+
+    const mimetype = chosenOne.props.mimetype;
+
+    return React.cloneElement(chosenOne, {
+      data: this.props.data[mimetype],
+      metadata: this.props.metadata[mimetype]
+    });
   }
 }
--- a/packages/presentational-components/src/components/execute-result.js
+++ b/packages/presentational-components/src/components/execute-result.js
@@ -1,0 +1,13 @@
+// @flow strict
+
+import * as React from "react";
+
+import type { ExecuteResultOutput } from "@nteract/records";
+
+type ExecuteResultProps = {} & ExecuteResultOutput;
+
+export class ExecuteResult extends React.Component<ExecuteResultProps, null> {
+  render() {
+    return <div />;
+  }
+}
--- a/packages/presentational-components/src/components/execute-result.js
+++ b/packages/presentational-components/src/components/execute-result.js
@@ -30,6 +30,9 @@ type ExecuteResultProps = {
    * Grab bag of random things
    */
   metadata: {},
+  /**
+   * React elements that accept mimebundle data
+   */
   children: React.Node
 };
 
@@ -41,6 +44,7 @@ export class ExecuteResult extends React.Component<ExecuteResultProps, null> {
   };
 
   render() {
+    // We must pick only one child to render
     let chosenOne = null;
 
     const data = this.props.data;
--- a/packages/presentational-components/src/components/execute-result.md
+++ b/packages/presentational-components/src/components/execute-result.md
@@ -33,7 +33,7 @@ HTML.defaultProps = {
   mimetype: "text/html"
 };
 
-<ExecuteResult data={{ "text/plain": "test data" }}>
+<ExecuteResult data={{ "text/plain": "SparkContext ⚡️" }}>
   <HTML />
   <Plain />
 </ExecuteResult>;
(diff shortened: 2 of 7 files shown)


rgbkrk:
> `@nteract/presentational-components` now depends on `@nteract/records` solely for types -- need to figure out if this should be a hard dependency

I now realize this makes documenting the flow types weird so I think I'll write them again here so that I can declare what the props mean in our components.

rgbkrk:
Ok, I'm loving this interface experiment. It rethinks how we've been doing `displayOrder` and `transforms` for mimebundle outputs.

Play with it now via https://declarative-trial.nteract.io/#richmedia

As a preview, you declare which components you'll render with as children, in order of decreasing priority:

```jsx
<RichMedia
  data={{
    "text/plain": "plain was richer",
    "text/html": "<b>HTML was richer</b>"
  }}
>
  <HTML />
  <Plain />
</RichMedia>;
```

Any props that get passed in to these children get passed on through when rendered so you can pass in custom things:

```jsx
<RichMedia
  data={{
    "text/plain": "plain was richer",
    "text/html": "<b>HTML was richer</b>"
  }}
>
  <HTML iframeMe="yesplease" />
  <Plain />
</RichMedia>;
```

🎉 

In terms of how our current interface works, we can adapt old code with something like this:

```jsx
<RichMedia
  data={{
    "text/plain": "plain was richer",
    "text/html": "<b>HTML was richer</b>"
  }}
>
  {displayOrder.map(
    mimetype =>
      transforms[mimetype]
        ? React.createElement(transforms[mimetype])
        : null
  )}
</RichMedia>;
```

Noting that we'd have to handle the `mimetype` prop _properly_.

packages/presentational-components/src/components/execute-result.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    const mimetype = chosenOne.props.mimetype;
+
+    return React.cloneElement(chosenOne, {
+      data: this.props.data[mimetype],
+      metadata: this.props.metadata[mimetype]
+    });
+  }
+}
```
alexandercbooth: I love this! It is so much easier to understand this way 👏.

packages/presentational-components/src/components/execute-result.js
```diff
@@ -0,0 +1,74 @@
...
+        return;
+      }
+      if (child.props && child.props.mimetype && child.props.mimetype in data) {
+        chosenOne = child;
+        return;
+      }
+    });
+
+    if (chosenOne === null) {
+      return null;
```
BenRussert: I wondered if `<Plain />` should just be available by default. Or maybe an indication that the kernel did not output any mime-types that the client can support?

Or is null the way to go?

packages/presentational-components/src/components/execute-result.md
```diff
@@ -0,0 +1,185 @@
...
+The execute result component, like `DisplayData`, takes in the mimebundle that the `execute_result` message has as part of its payload. The payload looks like this:
+
+```json
+{
+  "text/plain": "SparkContext ⚡️",
+  "text/html": "<b>SparkContext ⚡️</b>"
+}
+```
+
+The standard practice in Jupyter apps is to pick the "richest" of these. This Component lets
```
BenRussert: Some ideas not necessarily for this PR: 
I think we could reword to be slightly more beginner friendly, example:
> Jupyter frontends generally pick the "richest" output type they support, while ignoring any output types they do not.

If the audience is jupyter front end app developers, we could speak to them like:
> Your frontend should pick the "richest" output type it can support, while ignoring any other output types.

packages/records/src/outputs/execute-result.js
```diff
@@ -22,7 +22,7 @@ export type ExecutionCount = ?number;
 export const EXECUTE_RESULT = "execute_result";
 
 // In-memory version
-export type ExecuteResult = {
+export type ExecuteResultOutput = {
   outputType: ExecuteResultType,
```
BenRussert, line 26: Calling this `ExecuteResultOutput` is an improvement. I'm thinking we might be able to make this more intuitive if we change the name of this type somehow: `type ExecuteResultType = "execute_result";`

It seems confusing to me that the `ExecuteResultOutput` type *contains* a type called `ExecuteResultType`, what do you all think? Bringing it up here since it will be easier to change sooner than later after releasing. Comment applies to the similar types as well.

I was trying to look at the [flow types with redux](https://flow.org/en/docs/react/redux/) docs for inspiration, since actions are similar in a way with the string constant action types.

BenRussert (approved):
This is such a great idea and I am on board to move forward with this api. This is getting to be quite blog-post worthy! 

I'm already thinking about how I will integrate into hydrogen and hopefully produce some good examples on how to use this.

I think we will need some out of the box transform components somewhere. The custom transforms make sense to those of us who know the idea, but could be a challenge to get started for new devs. I may end up with better ideas after starting to plug this into hydrogen!

BenRussert:
Should we start a dev branch as this unfolds instead of waiting until we can merge into master?

packages/presentational-components/src/components/execute-result.js
rgbkrk: I think `null` is the way to go. People sometimes emit mimetypes that aren't for rendering and are just for forming records. Papermill does this.

packages/presentational-components/src/components/execute-result.md
rgbkrk: Oh excellent, I'll incorporate this today.

rgbkrk:
I just made some changes while traveling and will incorporate feedback later today. Thanks all for the feedback, this is great!

> Should we start a dev branch as this unfolds instead of waiting until we can merge into master?

I think we'll be able to merge this tonight without worrying about it interfering with the main app since it's a new interface. 😊

packages/records/src/outputs/execute-result.js
rgbkrk: Yeah, especially since this is an output type to match the jupyter type.

rgbkrk:
Ok I've updated a lot of the language in the component docs and switched just about everything over to `Media*` style naming. https://declarative-trial.nteract.io/#richmedia is updated and this PR is ready for the next round of reviews (annnnd maybe a merge ✅😄).

packages/presentational-components/src/components/rich-media.md
```diff
@@ -0,0 +1,226 @@
...
+  "text/plain": "SparkContext ⚡️",
+  "text/html": "<b>SparkContext ⚡️</b>",
+  "application/json": {
+    "spark": "awesome ⚡️",
+    "version": 2
+  }
+}
+```
+
+This object structure is called a **media bundle** (formerly known as a mimebundle), so dubbed because it's a bundle of [media types types](https://www.iana.org/assignments/media-types/media-types.xhtml) and associated data. Jupyter frontends pick the _richest_ media type amongst these for rendering for the user, by selecting via a **display order**.
```
alexandercbooth: Possible typo here?
>"media types types"

alexandercbooth (approved):
I think you've really hit on the right interface here 😄. I am also really happy with the new naming! Thanks @rgbkrk!

alexandercbooth:
Being able to review things with accompanying styleguidist entries rocks 😄

packages/presentational-components/src/components/rich-media.md
rgbkrk: Whoops! Fixing now.

rgbkrk:
Ok those were my last last last changes. I'm off to bed. 😅  🛏

packages/presentational-components/src/components/rich-media.md
```diff
@@ -1,3 +1,30 @@
...
+};
+
+const HTML = props => <div dangerouslySetInnerHTML={{ __html: props.data }} />;
+HTML.defaultProps = {
+  mediaType: "text/html"
+};
+
+<RichMedia
+  data={{
+    "text/plain": "plain is sooooo basic",
```
alexandercbooth, line 17: 😂

BenRussert (approved):
Love the new explanations for the docs!

PR #3224, opened 2018-08-14
[WIP] add appendOutput to records

alexandercbooth:
Continuing the outputs work, starting to bring in some reducers using the new outputs.
Lots of outputs 😄


BenRussert:
Nice work Alex! I am integrating with hydrogen at the moment, to see if tests pass.

rgbkrk:
I feel like we don't _have_ to call these reducers. This library can be used outside of a redux setup (though we'll make them all pure functions with immutable data types anyways).

Let's walk through using the API and see how it looks / works:

```js
import { appendOutput, outputFromMessage  } from "@nteract/records"

let outputs = [];

channels.subscribe(msg => {
  const output = outputFromMessage(msg);
  outputs = appendOutput(outputs, output)
}
```

From the immer docs, it seems like we should have the produce call happen as "high up" as possible, so maybe we provide mutable APIs 😬 😱 and use it like this:

```js
import { mutate, outputFromMessage  } from "@nteract/records"

let outputs = [];

// Same signature of appendOutput because of immer handling a function's length directly
appendOutput = produce(mutate.appendOutput)

channels.subscribe(msg => {
  const output = outputFromMessage(msg);
  outputs = appendOutput(outputs, output);
}
```

```js
import { mutate } from "@nteract/records";

// If we had a batch of outputs to deal with
produce(draft => {
  draft.outputs = batchOutputs.reduce(
    // For each output in the batch, append it to the prior collection of `outputs`
    mutate.appendOutput,
    // Use the prior outputs
    draft.outputs
  )
})
```

packages/records/src/reducers/outputs.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const streamOutput: StreamOutput = output;
+
+  function appendText(text: string): string {
+    if (typeof streamOutput.text === "string") {
+      return escapeCarriageReturnSafe(text + streamOutput.text);
+    }
+    return text;
+  }
```
rgbkrk: it might be best for us to bring this function outside the scope of this function and just make it take `text1` and `text2`.

packages/records/src/reducers/outputs.js
```diff
@@ -0,0 +1,77 @@
...
+          Object.assign(draftOutputs[draftLength], {
+            text: appendText(draftOutputs[draftLength].text)
+          });
+        }
+      });
+    }
+  }
+  return produce(outputs, draftOutputs => {
+    draftOutputs.push(streamOutput);
+  });
```
BenRussert: Somewhat similar to one of @rbgkrk's points, I think there is a way to wrap this whole function in one `produce` function, so you would be referring to and modifying `draftOutputs` the whole time.

If I'm not making sense I might have to code this out when I get home.

alexandercbooth:
>I feel like we don't have to call these reducers. This library can be used outside of a redux setup (though we'll make them all pure functions with immutable data types anyways).

I hadn't thought of that, but definitely true!

>Let's walk through using the API and see how it looks / works:

This is a really helpful idea and great way to think about things generally, thanks!

>so maybe we provide mutable APIs 😬 😱

Mutable APIs in our immutable records 😨. I think I see what you're saying, though. I will iterate on this tonight 👍

packages/records/__tests__/appendOutput-spec.js
```diff
@@ -0,0 +1,108 @@
...
+        traceback: []
+      }
+    ];
+    const newOutputs = appendOutput(outputs, {
+      outputType: "displayData",
+      data: {},
+      metadata: {}
+    });
+
+    expect(JSON.stringify(newOutputs)).toEqual(
```
BenRussert: I sent a PR to you last night removing all the boiler from tests that was to accommodate Immutable.js. figured it would be easier than explaining in the comments.

packages/records/src/reducers/outputs.js
alexandercbooth: Right, I think I get what you guys are saying. That will definitely help simplify this 👍. Will give it a shot tonight 😄

packages/records/__tests__/appendOutput-spec.js
alexandercbooth: You sent a PR? I don't I received it 😬. Was it through slack?

rgbkrk:
Here's the PR @alexandercbooth: https://github.com/alexandercbooth/nteract/pull/1

packages/records/__tests__/appendOutput-spec.js
rgbkrk: Here's the PR: https://github.com/alexandercbooth/nteract/pull/1

alexandercbooth:
Trying out a mutable API 👍. It can be used like ⬇️
```jsx
import { mutate, outputFromMessage } from "@nteract/records";

let outputs = [];
const appendOutput = produce(mutate.appendOutput);

channels.subscribe(msg => {
  const output = outputFromMessage(msg);
  outputs = appendOutput(outputs, output)
}

```

or, with the top level immer wrapped produce, ⬇️ 

```jsx
import { outputFromMessage } from "@nteract/records";
import appendOutput from "@nteract/records";

let outputs = [];

channels.subscribe(msg => {
  const output = outputFromMessage(msg);
  outputs = appendOutput(outputs, output)
}

```
I think maybe I like this better as it gives even more freedom for users.

packages/records/src/outputs/mutate-outputs.js
```diff
@@ -0,0 +1,74 @@
...
+): Array<OutputType> {
+  const last = outputs[outputs.length - 1];
+
+  if (
+    output.outputType !== "stream" ||
+    !last ||
+    (outputs.length > 0 && last["outputType"] !== "stream")
+  ) {
+    // If it's not a stream type, we just fold in the output
+    outputs.push(output);
```
BenRussert: Assuming the `outputs` object here is frozen/immutable, the outputs object wont be changed... right? It works with `produce`, but what if not used with `produce`?

packages/records/__tests__/appendOutput-spec.js
```diff
@@ -0,0 +1,156 @@
...
+    const newOutputs = appendOutput(outputs, {
+      outputType: "displayData",
+      data: {},
+      metadata: {}
+    });
+
+    expect(() => {
+      newOutputs[0] = { outputType: "stream", name: "stdout", text: "Boo" };
+    }).toThrow(
+      (TypeError: "Cannot assign to read only property '0' of object '[object Array]'")
```
BenRussert, line 118: Here is what I would call another "quirk" of `immer`. When you try to mutate outside of `produce`, it will either just not mutate (like a regular frozen object) or it might throw like here.

BenRussert: Interesting test here, I didnt know you could test that something throws!

packages/records/src/outputs/append-output.js
```diff
@@ -0,0 +1,79 @@
+// @flow strict
+import { escapeCarriageReturnSafe } from "escape-carriage";
+import produce from "immer";
+
+import type { OutputType, StreamOutput } from "../outputs";
+
+export function mutate() {}
```
BenRussert, line 7: I wasn't familiar with this pattern, is this preferable to writing as a class? It might transpile the same anyway, but I was just curious.

BenRussert (approved):
My "review" is basically just questions I had. Let's move along and try the records package in action!

Thanks for the work @alexandercbooth, I will see if I can show you all how `appendOutputs` works in hydrogen now!

PR #3225, opened 2018-08-14
catch when component's children error

rgbkrk:
This brings in (roughly) the `componentDidCatch` logic from the ol' `RichestMime` component, with one enhancement -- you can override the way the error renders. Check out `rich-media.md` for more on how to use it.

### DIFF
--- a/packages/presentational-components/src/components/rich-media.md
+++ b/packages/presentational-components/src/components/rich-media.md
@@ -248,3 +248,96 @@ class Output extends React.Component {
 
 <Output />;
 ```
+
+### Handling Errors from `<Media />` components
+
+The `<RichMedia />` component comes with a built-in `componentDidCatch` fallback
+
+```jsx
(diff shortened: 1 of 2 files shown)


packages/presentational-components/src/components/rich-media.md
```diff
@@ -248,3 +248,96 @@ class Output extends React.Component {
...
+  data={{
+    "text/plain": "<b>.____.</b>"
+  }}
+  renderError={IssueCreator}
+>
+  <Plain />
+</RichMedia>;
+```
+
+⚠️ `this.props.data` can contain sensitive data, so be mindful of what you enable for automatically populating an issue with ⚠️
```
alexandercbooth, line 343: Definitely good idea to note this 👍

alexandercbooth (approved):
Great to bring this in, thanks @rgbkrk!

Issue #3191, opened 2018-08-06
nteract stays blank on notebook open


r-zip:
When opening a notebook, nteract stays blank, never loading any of the content from the file (not even a code cell).

When opening the same notebook with `jupyter notebook`, there is no issue.

rgbkrk:
Howdy! Is this in the jupyter extension or the desktop app? If it's the jupyter extension does it show the menu?

r-zip:
This is the desktop app.

BenRussert:
Hi @rzpilgrim, what platform are you on and how did you install?

Is the problem only when you try to open an existing notebook? Or does this happen with a new notebook at startup?

BorisVarbanov:
Hello, just wanted to give some info, after I experienced the same issue today (desktop version) - it was due to the presence of an inline LaTeX equation embedded in a markdown cell (simply $\phi$). Removing this enabled the notebook to be shown and interacted with. This problem is more specifically referenced in a newly submitted issue #3226.

rgbkrk:
Ok I'm taking a look at this one now.

rgbkrk:
Closed by #3228 

We'll get a release out within the next 12 hours.

r-zip:
I had some inline equations in mine as well, but unfortunately, removing them did not solve my issue.

I am on macOS High Sierra 10.13.6, and I downloaded the installer from your website. It only occurs when I try to open an existing notebook. Jupyter notebook has no issue viewing/editing the file.

BenRussert:
Have you had a chance to try the new release @rzpilgrim?

r-zip:
Just tried it. It opens fine now.

BenRussert:
Excellent!

PR #3228, opened 2018-08-14
migrate to createRef API

### LINKED ISSUE #3223: Error: Invariant Violation: Minified React error #254
Any Qobj output from the `qutip` package causes the following error:

> Error: Invariant Violation: Minified React error #254; visit https://reactjs.org/docs/error-decoder.html?invariant=254&args[]=node for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

Tested with qutip version 4.3.1 and nteract 0.11.4

Full text of error (found via link above) is: 
>Element ref was specified as a string (node) but no owner was set. This could happen for one of the following reasons: 1. You may be adding a ref to a functional component 2. You may be adding a ref to a component that was not created inside a component's render method 3. You have multiple copies of React loaded See https://fb.me/react-refs-must-have-owner for more information.

The output from `qutip` has been problematic before (see #2509) not sure if this is related. If there is a better way to craft the output on the `qutip` end, I'm happy to work on a fix from that end. I don't yet know enough about nteract to submit a solution from this side of things.

### LINKED ISSUE #3226: Can't type formulas in the Markdown cell.
Hello! 

Displaying any formula or Greek notations in Markdown requires the formula expression to be  in __$__ *my-formula* __$__.  But the initial <kbd>shift</kbd> + <kbd>4</kbd> works fine and prints a __$__ . 

But any subsequent combination of  <kbd>shift</kbd> + <kbd>4</kbd> just makes the nteract window go white and renders it inactve  with only the menubar in display.

I have tested it out with multiple iterations and gotten the same results. 

General Info : 
1. nteract Windows app - 0.11.4
2. Windows 10 
3. Background applications open were : Chrome, Notepad++, MS Excel

rgbkrk:
Fixes #3226  #3191 

Possibly fixes #3223

😬  The old ref API was deprecated and I didn't even realize it was used here. This adapts us to the latest API for React + Refs. After this I will get another release out tonight.

### DIFF
--- a/packages/mathjax/src/node.js
+++ b/packages/mathjax/src/node.js
@@ -15,6 +15,7 @@ type Props = {
 
 class Node extends React.Component<Props, *> {
   script: ?HTMLScriptElement;
+  nodeRef: React.ElementRef<*>;
 
   static defaultProps = {
     inline: false,
@@ -23,6 +24,8 @@ class Node extends React.Component<Props, *> {
 
   constructor(props: Props) {
     super(props);
+
+    this.nodeRef = React.createRef();
 
     (this: any).typeset = this.typeset;
   }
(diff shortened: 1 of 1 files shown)


alexandercbooth (approved):
Thanks for the quick fix, @rgbkrk!

PR #3229, opened 2018-08-15
update the altair / vega notebook

rgbkrk:
Made the altair notebook include an interactive example and removed the bespoke vegalite function I wrote in this notebook long ago.


alexandercbooth (approved):
Awesome, thanks @rgbkrk!

rgbkrk:
We really need to re-vamp these example notebooks someday. 😅  It's pretty new contributor friendly if we outline the work I think.

alexandercbooth:
That's a great idea 🙂

PR #3236, opened 2018-08-18
create an outputs package

rgbkrk:
## Plan

Create a package(s) to replace `display-area` and `transforms`.

Component interface thinking:

```jsx
<Outputs>
  <PlainText /> {/* stdout */}
  <PlainText />
  <RichMedia>
    <DataExplorer />
    <Media.HTML />
    <Media.Plain />
  </RichMedia>
</Outputs>
```

Might end up making helper components that accept the same signature as `@nteract/records` does for output types.

_For now_, this just exports the `<RichMedia />` component within `@nteract/outputs` instead of `@nteract/presentational-components`.


alexandercbooth (approved):
Great idea, thanks! I think planning on having the transforms in here could be great as well and would make it even more clear what their purpose is (to new and old folks alike).

styleguide.config.js
```diff
@@ -15,7 +15,11 @@ module.exports = {
       components: "packages/presentational-components/src/components/*.js"
     },
     {
-      name: "Binder Components",
+      name: "Outputs",
+      components: "packages/outputs/src/components/*.js"
+    },
+    {
+      name: "Binder Hosts and Kernels",
```
alexandercbooth, line 22: This is a more clear name, too 🙂

PR #3245, opened 2018-08-21
Replace f-strings to fix compatibility with python < 3.6

adikus:

When trying out nteract in my python 3.4 env, I've noticed that `nteractapp.py` uses f-strings (Literal String Interpolation), which is a python 3.6 feature (and hence it crashed on start for me). I've replaced the two instances in the file with `.format` and managed to get it to work.
I've done a text search for `f'` and `f"` over the whole codebase and it seems like this is the only file where this is used.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/nteractapp.py
+++ b/applications/jupyter-extension/nteract_on_jupyter/nteractapp.py
@@ -14,7 +14,7 @@ from .utils import cmd_in_new_dir
 
 webpack_port = 8357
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Thank you @adikus!

I love f-strings. What I love even more is more people being able to use this extension. 😄  👍

rgbkrk:
Not sure why Circle CI ended early there, I think based on the run I can merge this one and I'll make sure `master` is good.

PR #3259, opened 2018-08-25
Xsrffixing

yangt:
passing xsrf token

### DIFF
--- a/packages/rx-jupyter/src/base.js
+++ b/packages/rx-jupyter/src/base.js
@@ -1,4 +1,5 @@
 // @flow
+import * as Cookies from "js-cookie";
 
 export function normalizeBaseURL(url: string): string {
   return url.replace(/\/+$/, "");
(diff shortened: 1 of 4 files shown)


packages/rx-jupyter/src/base.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import * as Cookies from "js-cookie";
```
rgbkrk, line 2: :100: This looks like the [best of the cookie](https://github.com/js-cookie/js-cookie/tree/latest#readme) libraries on npm.

* Works in all browsers
* Accepts any character
* Heavily tested
* No dependency
* Unobtrusive JSON support
* Supports AMD/CommonJS
* RFC 6265 compliant
* Useful Wiki
* Enable custom encoding/decoding
* ~900 bytes gzipped!

😄 

Thanks!

packages/rx-jupyter/src/base.js
```diff
@@ -11,7 +12,9 @@ export function createAJAXSettings(
 ): Object {
   const baseURL = normalizeBaseURL(serverConfig.endpoint || serverConfig.url);
   const url = `${baseURL}${uri}`;
+  const xsrfToken = Cookies.get("_xsrf");
```
rgbkrk, line 15: We may want to enable opting in to this because we use this library in environments where the cookies are not associated with the jupyter API (or even a way to pass it in). For now this is great and I'm happy it closes issues where people are trying to use the jupyter extension!

rgbkrk (approved):

PR #3264, opened 2018-08-27
Fix additional broken links in README.md

susannah146:


rgbkrk:
Thank you so much for all these fixes!

PR #3280, opened 2018-08-31
produce less errors and warnings from React DOM

rgbkrk:
Working on docs while doing the MathJax component and saw these areas for cleaning up. 😄


alexandercbooth (approved):
Good finds, thanks @rgbkrk!

alexandercbooth:
We don't have `now` deploying a preview of the styleguidist components on PRs here, is that something we might want to consider? 

I wonder if it can be configured to PR a preview only when files deployed by it are changed (like our components) 🤔

rgbkrk:
I tried to get our components docs auto-deploying with this repo, the monorepo proved to be too heavyweight.

alexandercbooth:
Oh OK gotcha 👍

PR #3282, opened 2018-09-02
remove unused <MathJax.Text />

rgbkrk:
We're not using `<MathJax.Text />`, so in the spirit of having less to maintain I'd like to take it out (we directly render with `<MathJax.Node />` instead).


rgbkrk:
After this PR (much later today), I'm going to modernize `<MathJax.Node />` followed by `<MathJax.Context />`.

alexandercbooth (approved):
Sounds good, thanks @rgbkrk!

PR #3283, opened 2018-09-02
[WIP] add appendOutput & typing from @nteract/records in reducer

alexandercbooth:
Continuing with our previous work, this begins to integrate some of the records work. In particular, this brings in the appendOutput function and typings for use within our reducers.

Also changes `appendOutput()` to a named export.

Still need to fix flow later, 😔. 

This W is very much IP and really just a starting point to continue the momentum 🙃.

### DIFF
--- a/packages/core/src/reducers/core/entities/contents/notebook.js
+++ b/packages/core/src/reducers/core/entities/contents/notebook.js
@@ -8,8 +8,6 @@ import * as actionTypes from "../../../../actionTypes";
 import * as Immutable from "immutable";
 
 import { appendOutput as appendOutputRecord } from "@nteract/records";
-import type { NbformatOutput } from "@nteract/records";
-type Immouts = Immutable.List<NbformatOutput>;
 
 import type {
   ImmutableCell,
@@ -210,7 +208,7 @@ function appendOutput(state: NotebookModel, action: actionTypes.AppendOutput) {
   ) {
     return state.updateIn(
       ["notebook", "cellMap", cellID, "outputs"],
-      (outputs: ImmutableOutputs): ImmutableOutputs =>
+      (outputs: ImmutableOutputs): mixed =>
         Immutable.fromJS(appendOutputRecord(outputs.toJS(), output))
     );
   }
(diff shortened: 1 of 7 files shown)


packages/core/src/reducers/core/entities/contents/notebook.js
```diff
@@ -206,8 +208,8 @@ function appendOutput(state: NotebookModel, action: actionTypes.AppendOutput) {
   ) {
     return state.updateIn(
       ["notebook", "cellMap", cellID, "outputs"],
-      (outputs: ImmutableOutputs): ImmutableOutputs =>
-        reduceOutputs(outputs, output)
+      (outputs: ImmutableOutputs): mixed =>
+        Immutable.fromJS(appendOutputRecord(outputs.toJS(), output))
```
alexandercbooth, line 212: Incorporating this means mixing our immer state (or I guess I should say our immutable state) with our Immutable state. At least immer objects are already JS, but I worry this `fromJS(...toJS())` seems like it could be costly 🤔.

packages/core/src/reducers/core/entities/contents/notebook.js
```diff
@@ -7,6 +7,8 @@ import * as actionTypes from "../../../../actionTypes";
 
 import * as Immutable from "immutable";
 
+import { appendOutput as appendOutputRecord } from "@nteract/records";
```
alexandercbooth, line 10: Had to import as a different name here as we already have a reducer (in which this function is used) named `appendOutput`.

packages/core/src/reducers/core/entities/contents/notebook.js
rgbkrk: The from and to are definitely costly. We definitely want to be switched over to `Array<Output>` as our objects and not use the `Immutable.List<ImmutableOutput>`

alexandercbooth: OK good, that definitely makes sense 👍

rgbkrk:
I'm going to assume we should go ahead with closing this one to revisit as a new PR. Thanks @alexandercbooth, look forward to seeing you again soon. Hope your thesis is going well!

PR #3297, opened 2018-09-07
match version of styled-jsx with that of next

rgbkrk:
Something I probably could have done in the babel PR that I didn't think of was to upgrade our styled-jsx to match that of next.js. I look forward to when these aren't on their canary versions.


rgbkrk:
This is looking good with the apps!

PR #3322, opened 2018-09-13
React table for dx grid

emeeks:
Use `react-table` instead of `react-virtualized-grid`
Fix some formatting issues (precision of axis labels)
Fix some graphical ugliness (viz area was too wide so horizontal scrolling happened)


emeeks:
This should resolve the flurry of DX grid issues as well as a couple older DX issues about buttons and decimal precision in axes.

emeeks:
The one thing I ignored feature-wise was in theming, so that will have to be handled later. Otherwise I think the react-table grid looks and handles better than the react-virtualized grid.

PR #3326, opened 2018-09-13
Disable ReactTableFixedColumns til prefixed uniqid

rgbkrk:
Stuck with using regular `react-table` until https://github.com/GuillaumeJasmin/react-table-hoc-fixed-columns/pull/12 is merged.

because sometimes uniqid will generate an invalid selector (leading digit). This ends up
being deterministic per machine because they're using the mac address of the system
as part of the hash.

### DIFF
--- a/packages/transform-dataresource/src/charts/grid.js
+++ b/packages/transform-dataresource/src/charts/grid.js
@@ -16,7 +16,7 @@ export const DataResourceTransformGrid = ({
 
   return (
     <div style={{ width: "calc(100vw - 150px)" }}>
-      <ReactTableFixedColumns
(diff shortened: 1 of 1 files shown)


BenRussert:
Don't want to wait until the PR is merged?

rgbkrk:
After the fix is merged and shipped, we can bring back the fixed columns. @emeeks and I had a hilarious time finding where the bug we were hitting was occurring.

BenRussert:
>  @emeeks and I had a hilarious time finding where the bug we were hitting was occurring.

Pretty interesting bug though, thanks for chasing down.

PR #3329, opened 2018-09-14
create Output Picker that defers to a child element

rgbkrk:
@ivanov and I experimented with a new API for rendering outputs
that lets you provide what you want to use for rendering `execute_result`,
`display_data`, `stream`, etc.

Here's the API outline:

```jsx
import { Output, DisplayData, ExecuteResult, StreamText } from "@nteract/outputs";

// Outputs from a cell!
const outputs = [
  {
    outputType: "stream",
    text: "Hello World\nHow are you?",
    name: "stdout"
  },
  {
    outputType: "display_data",
    data: {
      "text/plain": "O____O"
    },
    metadata: {}
  },
  {
    outputType: "stream",
    text: "Pretty good I would say",
    name: "stdout"
  }
];

// Render the outputs
<div>
  {outputs.map(output => (
    <Output output={output}>
      <StreamText />
      <DisplayData>
        <Plain />
        <HTML />
      </DisplayData>
      <ExecuteResult>
        <Plain />
        <HTML />
      </ExecuteResult>
    </Output>
  ))}
</div>;
```

The main motivation here is to make an API that allows doing configuration by composition, so that we can pass things like `onMetadataChange` to a media component without having to refactor several other components. With `@nteract/display-area`, this was a regular problem requiring a lot of "prop-drilling" to pass a prop through N components.

We'll want to follow this PR up with elements for each of the Jupyter output types. I wrote
hokey ones in the docs here to get this started.

All that being said, this may be ok to bring in as is since it's not integrated into the overall app!

----

To hack on this, you should:

* Grab this branch
* Run:

```
yarn
yarn docs
```

* Open http://127.0.0.1:6060/#!/Output

-----

Next steps I see:

* [ ] Write our "official" output components that would handle all the output types just like what we previously handled in `packages/display-area/src/output.js`
  * [ ] `<ExecuteResult>`
  * [ ] `<DisplayData>`
  * [ ] `<Traceback />` (or call it `<JupyterError />`
  * [ ] `<Stream />`
* [ ] Write a jest test for this component. Seems pretty simple! Not too many edge cases to knock out.
* [ ] Re-write (?) the "transforms" components with the new `<Media />` names

Some of these might be ok as separate steps.

-----

Hey look @BenRussert & @alexandercbooth, we're using the output record types here. 😄 

cc @emeeks

### DIFF
--- a/packages/outputs/src/components/output.js
+++ b/packages/outputs/src/components/output.js
@@ -1,0 +1,62 @@
+// @flow strict
+
+import * as React from "react";
+
+// We might only need this as a devDependency as it is only here for flow
+import type OutputType from "@nteract/records";
+
+type Props = {
+  /**
+   * React elements that accept Output
+   */
+  children: React.Node,
+  /**
+   * The raw output, as expected from @nteract/records
+   */
+  output: OutputType
+};
+
+type State = {};
+
+export class Output extends React.Component<Props, State> {
+  static defaultProps = {
+    output: null
+  };
+
+  render() {
+    // We must pick only one child to render
+    let chosenOne = null;
+
+    if (this.props.output == null) {
+      return null;
+    }
+
+    const outputType = this.props.output.outputType;
+
+    // Find the first child element that matches something in this.props.data
+    React.Children.forEach(this.props.children, child => {
+      if (chosenOne) {
+        // Already have a selection
+        return;
+      }
+      if (
+        child.props &&
+        child.props.outputType &&
+        child.props.outputType === outputType
+      ) {
+        chosenOne = child;
+        return;
+      }
+    });
+
+    // If we didn't find a match, render nothing
+    if (chosenOne === null) {
+      return null;
+    }
+
+    // Render the output component that handles this output type
+    return React.cloneElement(chosenOne, {
+      output: this.props.output
+    });
+  }
+}
--- a/packages/outputs/src/components/output.md
+++ b/packages/outputs/src/components/output.md
@@ -1,0 +1,99 @@
+The `<Output />` element is a glorified Switch statement for picking what to render an output with.
+
+```jsx
+// Until we create <Stream />
+const StreamText = props => (
+  <pre
+    style={{
+      backgroundColor: props.output.name === "stderr" ? "#f9e0db" : "#e0f9db"
+    }}
+  >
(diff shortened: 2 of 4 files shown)


rgbkrk:
What I'm unsure of is _what_ to call this component. `<OutputPicker />`? `<OutputSwitch />`?

rgbkrk:
Rebased to make sure this was up to date.

packages/outputs/src/components/output.js
```diff
@@ -0,0 +1,62 @@
...
+    if (chosenOne === null) {
+      return null;
+    }
+
+    // Render the output component that handles this output type
+    return React.cloneElement(chosenOne, {
+      output: this.props.output
+    });
+  }
+}
```
alexandercbooth, line 62: This is great, @rgbkrk & @ivanov! I'm really liking this compositional style in general 👍.

packages/outputs/src/components/output.md
```diff
@@ -0,0 +1,99 @@
...
+const outputs = [
+  {
+    outputType: "stream",
+    text: "Hello World\nHow are you?",
+    name: "stdout"
+  },
+  {
+    outputType: "display_data",
+    data: {
+      "text/plain": "O____O"
```
alexandercbooth, line 77: Love watching this little one float on by. Reminds me of webpages I made about a 1000 years ago 🙂

alexandercbooth:
I like `<OutputPicker />` as it feels the most clear, but `<OutputSwitch />` also makes sense syntactically. `<OutputSwitch />` also kind of makes me think of an on/off button for outputs. Both good, though 👍

rgbkrk:
The [`Switch` terminology comes from React Router](https://reacttraining.com/react-router/core/api/Switch) which possibly is a carry over from a [Network Switch](https://en.wikipedia.org/wiki/Network_switch), as well as the `switch` statement. We definitely don't want there to be confusion though. I both _like_ this pattern of composability and also find it kind of weird. It's like... a template we're stamping out?

emeeks:
I'm going to iterate on this and I think it's a good place to start, so I'm merging it in.

rgbkrk:
Ok I'll make a new issue for the things that need to be handled next. 😄

PR #3330, opened 2018-09-14
Better plotting dx

emeeks:
Canvas rendering for 1000+ point scatterplots (canvas instead of SVG as well as not hover interactivity, so again fewer DOM elements)
Smart numeric formatting for axes (extra decimal points, scientific notation for anything about 999 trillion or below 0.0001 (still doesn't know how to handle percent, so we should expose some controls for users)
Primary keys are fixed columns instead of just the first column.


rgbkrk:
Awesome, thanks.

PR #3331, opened 2018-09-15
Restore fixedColumnsHOC

emeeks:
Pretty straightforward, just turns fixed columns back on for the grid.


BenRussert:
I see the review request there, but I know this was the plan. Thanks!

PR #3343, opened 2018-09-19
Output picker

emeeks:
Create `<StreamText>`
Spread props for Output

### DIFF
--- a/packages/outputs/src/components/output.js
+++ b/packages/outputs/src/components/output.js
@@ -55,8 +55,6 @@ export class Output extends React.Component<Props, State> {
     }
 
     // Render the output component that handles this output type
-    return React.cloneElement(chosenOne, {
-      output: this.props.output
-    });
+    return React.cloneElement(chosenOne, this.props.output);
   }
 }
--- a/packages/outputs/src/components/stream-text.js
+++ b/packages/outputs/src/components/stream-text.js
@@ -5,19 +5,18 @@ import Ansi from "ansi-to-react";
 
 import type { StreamOutput } from "@nteract/records";
 
-type Props = {
-  output: StreamOutput,
-  outputType: "stream"
-};
+type Props = StreamOutput;
 
 const StreamText = (props: Props) => {
-  const { text, name } = props.output;
+  const { text, name } = props;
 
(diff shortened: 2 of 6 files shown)


packages/outputs/src/components/output.js
```diff
@@ -55,8 +55,6 @@ export class Output extends React.Component<Props, State> {
     }
 
     // Render the output component that handles this output type
-    return React.cloneElement(chosenOne, {
-      output: this.props.output
-    });
+    return React.cloneElement(chosenOne, this.props.output);
```
alexandercbooth, line 58: Good thinking, more concise 👍

packages/outputs/src/components/stream-text.js
```diff
@@ -0,0 +1,22 @@
...
+  return <Ansi className={`"nteract-display-area-${name}`}>{text}</Ansi>;
+};
+
+StreamText.defaultProps = {
+  outputType: "stream",
+  text: "",
+  name: "stdout"
+};
+
+export default StreamText;
```
alexandercbooth, line 22: We could also think about doing a named export here to match the other outputs components, but looks good 👍

alexandercbooth (approved):
Thanks for getting this rolling, @emeeks!

packages/outputs/src/components/output.js
rgbkrk: The main motivation was to make it so that we could use the new record output structures directly as the props:

```jsx
<Stream 
  text="hello world"
  name="stdout"
/>
```

(we totally met in person to work through some of this, figured I'd share the context 😅)

packages/outputs/src/components/stream-text.js
rgbkrk: Ooh, smart yeah we probably should/could.

packages/outputs/src/components/output.js
alexandercbooth: Ah gotcha, I think I see what you're saying 👍

rgbkrk (approved):

rgbkrk:
On to the next ones!

Issue #3342, opened 2018-09-19
<Output> Picker


rgbkrk:
Next steps for the `@nteract/outputs`:

* [x] Write our "official" output components that would handle all the output types just like what we previously handled in `packages/display-area/src/output.js`
  * [x] `<ExecuteResult>`
  * [x] `<DisplayData>`
  * [x] `<Traceback />` (or call it `<JupyterError />`
  * [x] `<Stream />`
* [x] Create `<Media />` components for all the transforms (or adapt them to the new style which only needs adding a `mediaType` key) 
* [ ] Extract styles out of `packages/presentational-components/src/components/outputs.js` that belong in the respective media or outputs, possibly packaged as:

```js
export userOutputStyle = css`
  .userOutput :global(table) {
    /* yay css */
  }
`;
```

then use them in `<Media.HTML />` and `@nteract/markdown`'s markdown.

* [x] Integrate the Output picker, RichMedia, and all these components into the notebook app component
* [ ] Deprecate `@nteract/display-area`

rgbkrk:
The big "why" on this effort is to make it so that a cell effectively looks like this in the React Tree:


```jsx
<Cell>
  <Input>
    <Prompt counter={1} />
    <Source language="python">{`import pandas as pd\npd.DataFrame([1,2,3])\n\n# Alternate between hovering the cursor over this cell and outside of the cell`}</Source>
  </Input>
  <Outputs>
    {outputs.map(output => (
      <Output>
        <DisplayData>
          <DataExplorer onMetadataChange={handleMetadataUpdate} />
          <Media.HTML isolated />
          <Media.Image mediaType="image/gif" />
          <Media.Image mediaType="image/jpeg" />
          <Media.Image mediaType="image/png" />
          <Media.Text />
        </DisplayData>
        <KernelOutputError />
        <StreamText />
      </Output>
    ))}
  </Outputs>
</Cell>
```

This lets frontends pass in what they want to use for output types and for handling media types, including any `onChange` handlers.

captainsafia:
Here's a breakdown of the transforms that need to be migrated to `Media` components that we can track.

**Components**
- [X] HTML
- [X] Image
- [x] JavaScript (@captainsafia)
- [X] JSON
- [x] LaTeX (@lgeiger)
- [X] Markdown
- [X] SVG
- [X] Plaintext

**Documentation**
- [x] HTML (@captainsafia)
- [X] Image
- [x] JavaScript (@captainsafia)
[...]

PR #3347, opened 2018-09-20
add error output component

alexandercbooth:
Towards #3342, this adds the `<JuptyerError />` component.

I went with JupyterError since it sounds more clear, but I'm OK with Traceback if folks would rather it be that.

### DIFF
--- a/packages/outputs/src/components/jupyter-error.js
+++ b/packages/outputs/src/components/jupyter-error.js
@@ -1,0 +1,29 @@
+// @flow strict
+import * as React from "react";
+import Ansi from "ansi-to-react";
+
+import type { ErrorOutput } from "@nteract/records";
+type Props = ErrorOutput;
+
+export const JupyterError = (props: Props) => {
+  const { ename, evalue, traceback } = props;
+
+  if (!Array.isArray(traceback) || !traceback.length) {
+    return (
+      <Ansi className="nteract-display-area-traceback">{`${ename}: ${evalue}`}</Ansi>
+    );
+  }
+
+  return (
+    <Ansi className="nteract-display-area-traceback">
+      {traceback.join("\n")}
+    </Ansi>
+  );
+};
+
+JupyterError.defaultProps = {
+  outputType: "error",
+  ename: "",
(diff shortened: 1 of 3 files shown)


packages/outputs/src/components/jupyter-error.js
```diff
@@ -0,0 +1,29 @@
...
+
+import type { ErrorOutput } from "@nteract/records";
+type Props = ErrorOutput;
+
+export const JupyterError = (props: Props) => {
+  const { ename, evalue, traceback } = props;
+
+  if (!Array.isArray(traceback) || !traceback.length) {
+    return (
+      <Ansi className="nteract-display-area-traceback">{`${ename}: ${evalue}`}</Ansi>
```
rgbkrk, line 13: I wonder if we want to take note of this class name being old (maybe deprecated). It's primarily Hydrogen that's using the class names to target some styling.

packages/outputs/src/components/jupyter-error.md
```diff
@@ -0,0 +1,9 @@
+The `<JuptyerError />` component will render a Jupyter error with or without a traceback.
+
+```jsx
+<JupyterError ename="NameError" evalue="Yikes!" traceback={["Yikes, never heard of that one!"]} />
```
rgbkrk, line 4: Ha!

packages/outputs/src/components/jupyter-error.md
```diff
@@ -0,0 +1,9 @@
...
+```
+
+```jsx
+<JupyterError ename="NameError" evalue="Yikes!" />
```
rgbkrk, line 8: I wonder if we should call this component `KernelOutputError` or `KernelError`.

rgbkrk (approved):

rgbkrk:
I'm going to go ahead and merge this in. Since it's still a new API we can change names and whatnot later.

packages/outputs/src/components/jupyter-error.md
alexandercbooth: I like that idea 👍

PR #3348, opened 2018-09-20
[WIP] test output component

alexandercbooth:
Towards #3342.

Getting the testing kicked off for the output component in the same way we had done with display-area output. 

Just a WIP until #3347 goes in, since it includes a component from that and I branched from there.

### DIFF
--- a/packages/outputs/__tests__/output-spec.js
+++ b/packages/outputs/__tests__/output-spec.js
@@ -1,0 +1,47 @@
+import * as React from "react";
+import { shallow } from "enzyme";
+
+import { Output, StreamText, JupyterError } from "../src";
+
+describe("Output", () => {
+  it("handles stream data", () => {
+    const output = { outputType: "stream", name: "stdout", text: "hey" };
+
+    const component = shallow(
+      <Output output={output}>
+        <StreamText />
+      </Output>
+    );
+    console.log(component);
+    expect(component.type()).toEqual(StreamText);
+  });
+
+  it("handles errors/tracebacks", () => {
+    const output = {
+      outputType: "error",
+      traceback: ["Yikes, Will is in the upsidedown again!"],
+      ename: "NameError",
+      evalue: "Yikes!"
+    };
+
+    const component = shallow(
+      <Output output={output}>
+        <JupyterError />
+      </Output>
+    );
+    expect(component.type()).toEqual(JupyterError);
+
+    const outputNoTraceback = {
+      outputType: "error",
+      ename: "NameError",
+      evalue: "Yikes!"
+    };
+
+    const component2 = shallow(
+      <Output output={output}>
+        <JupyterError />
+      </Output>
+    );
+    expect(component2.type()).toEqual(JupyterError);
+  });
+});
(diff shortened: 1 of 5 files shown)


rgbkrk:
#3347 is in now, so this can have the WIP tag removed. 😄

packages/outputs/src/index.js
```diff
@@ -1,2 +1,6 @@
-// @flow
-export { RichMedia } from "./components/rich-media.js";
+// @flow strict
+export { RichMedia } from "./components/rich-media";
+export { Output } from "./components/output";
+export { JupyterError } from "./components/jupyter-error";
+import StreamText from "./components/stream-text";
```
rgbkrk, line 5: Guess we need that named export for easier exporting here. 😄

BenRussert: Along the lines of other todos, I'm thinking we should change to `<OutputSwitch />` sooner rather than later. Or are we sticking to `<Output />`?

packages/outputs/__tests__/output-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(component.type()).toEqual(JupyterError);
+
+    const outputNoTraceback = {
+      outputType: "error",
+      ename: "NameError",
+      evalue: "Yikes!"
+    };
+
+    const component2 = shallow(
+      <Output output={output}>
```
BenRussert, line 41: Were you meaning to go:
```jsx
 <Output output={outputNoTraceback}>
```
Otherwise `outputNoTraceback` is just hanging out there :smiley:

packages/outputs/src/components/jupyter-error.js
```diff
@@ -0,0 +1,29 @@
...
+
+import type { ErrorOutput } from "@nteract/records";
+type Props = ErrorOutput;
+
+export const JupyterError = (props: Props) => {
+  const { ename, evalue, traceback } = props;
+
+  if (!Array.isArray(traceback) || !traceback.length) {
+    return (
+      <Ansi className="nteract-display-area-traceback">{`${ename}: ${evalue}`}</Ansi>
```
BenRussert, line 13: We use something like this  `className` in hydrogen today, but I think we could drop this class here in the stock error component. 

I have a long standing todo to create a custom error component for hydrogen, and this new api is going to make that happen I think! [See this issue](https://github.com/nteract/hydrogen/issues/894) if you want to find out how long I have had this on my mind :calendar:

BenRussert (approved):
I am a big fan of how this new api is working out and these tests gave me a chance to try it in action.Thanks @alexandercbooth for putting this together!

I iterated a bit on the tests as I played around, I think we should merge this and I'll show what I mean in a separate PR!

packages/outputs/src/index.js
rgbkrk: I think that's great, want to go ahead and make a PR?

packages/outputs/__tests__/output-spec.js
alexandercbooth: I was, good catch @BenRussert! Thank you 😄

alexandercbooth:
Thanks @BenRussert 😄

PR #3349, opened 2018-09-22
Remove boundary in cell toolbar

captainsafia:
Saw this in the roadmap and figured it would be an easy fix.

screen shot 2018-09-22 at 10 24 45 am


rgbkrk:
This went in to help create some space between delete and the other buttons. Perhaps we should just make this a spacer with no icon?

captainsafia:
Makes sense. On it.

PR #3350, opened 2018-09-22
cleanup output error test

alexandercbooth:
Following up on #3348 and #3347, this 
- renames `<JupyterError />` to `<KernelOutputError />` 
- uses the right output in the second test
- removes old className


rgbkrk:
Thanks for the follow up clean up!

alexandercbooth:
My pleasure 😄

PR #3351, opened 2018-09-22
Add DisplayData component to Outputs package

captainsafia:
Part of https://github.com/nteract/nteract/issues/3342.

### DIFF
--- a/packages/outputs/__tests__/output-spec.js
+++ b/packages/outputs/__tests__/output-spec.js
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { shallow } from "enzyme";
 
-import { Output, StreamText, KernelOutputError } from "../src";
+import { Output, StreamText, KernelOutputError, DisplayData } from "../src";
 
 describe("Output", () => {
   it("handles stream data", () => {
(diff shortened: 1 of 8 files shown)


packages/outputs/src/components/display-data.js
```diff
@@ -0,0 +1,23 @@
...
+import type { DisplayDataOutput } from "@nteract/records";
+type Props = DisplayDataOutput;
+
+export const DisplayData = (props: Props) => {
+  const { data, metadata, children } = props;
+
+  return (
+    <RichMedia data={data} metadata={metadata}>
+      {children}
+    </RichMedia>
```
rgbkrk, line 15: Perfect, this is exactly what I was expecting.

rgbkrk (approved):

packages/outputs/__tests__/output-spec.js
```diff
@@ -44,4 +44,25 @@ describe("Output", () => {
...
+
+    const component = shallow(
+      <Output output={output}>
+        <DisplayData>
+          <Plain />
+        </DisplayData>
+      </Output>
+    );
+
+    expect(component.type()).toEqual(DisplayData);
```
rgbkrk, line 66: It's pretty neat how this tests the _output_ of `Output`. 😄 

We may want to have a test that exercises `DisplayData` directly, since this shallow test is only a top level render.

PR #3353, opened 2018-09-23
Add DisplayData spec

captainsafia:


packages/outputs/__tests__/display-data-spec.js
```diff
@@ -0,0 +1,27 @@
...
+      mediaType: "text/plain"
+    };
+
+    const component = mount(
+      <DisplayData {...output}>
+        <Plain />
+      </DisplayData>
+    );
+
+    expect(component.type()).toEqual(DisplayData);
```
rgbkrk, line 24: I think it's cool for tests like these, especially since these are stateless functional components that we just use a jest snapshot.

PR #3355, opened 2018-09-23
add HTML media component

alexandercbooth:
Continuing with the output work, this kicks off bringing in the transforms by starting with HTML. At the moment, I have not yet included a `.md` for it alone, but used it in the `output.md` instead.

### DIFF
--- a/packages/outputs/__tests__/media-spec.js
+++ b/packages/outputs/__tests__/media-spec.js
@@ -1,0 +1,34 @@
+import React from "react";
+import { mount } from "enzyme";
+
+import * as Media from "../src/components/media";
+
+describe("HTML", () => {
+  it("renders direct HTML", () => {
+    const component = mount(<Media.HTML data={"<b>cats are best</b>"} />);
+    expect(component.html()).toEqual("<div><b>cats are best</b></div>");
+  });
+  it("correctly chooses to update with data changing", () => {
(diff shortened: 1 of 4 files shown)


packages/outputs/__tests__/media-spec.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+  it("updates the underlying HTML when data changes", () => {
+    const wrapper = mount(<Media.HTML data={"<b>cats are best</b>"} />);
+    expect(wrapper.html()).toEqual("<div><b>cats are best</b></div>");
+
+    wrapper.setProps({ data: "<b>squirrels are pretty great, too</b>" });
+
+    expect(wrapper.html()).toEqual(
+      "<div><b>squirrels are pretty great, too</b></div>"
+    );
```
rgbkrk, line 32: I'm guessing you pulled these tests from the old setup, it might be worth it to switch these to snapshot tests instead.

Then again... the `setProps` tests for the component updating is probably a good idea. Might be able to snapshot there too?

rgbkrk (approved):

packages/outputs/__tests__/media-spec.js
alexandercbooth: That's a good idea. I'll investigate 👍

rgbkrk: Cool, thanks. I might be wrong too, so it's worth looking at what the snapshots look like.

PR #3357, opened 2018-09-23
document reasoning for compiler config settings

rgbkrk:
I was trying to change something else and figured I should/could document why the compiler config was setup the way it was. Just a few extra comments.


alexandercbooth (approved):
Sounds good, thanks @rgbkrk!

PR #3372, opened 2018-09-28
Change errorMiddleWare to log/notify about more types of failure-related actions

### LINKED ISSUE #3369: Log EXECUTE_FAILED actions
I see many epics going to the trouble of generating EXECUTE_FAILED actions, but I don't see them being logged or otherwise handled anywhere, outside of testing that they're generated.

jameskoch:
Fixes #3369.

The following actions will now be logged to the console, and users will see notifications:
FETCH_CONTENT_FAILED
LAUNCH_KERNEL_FAILED
KILL_KERNEL_FAILED
INTERRUPT_KERNEL_FAILED
EXECUTE_FAILED
SAVE_FAILED
UPDATE_DISPLAY_FAILED
DELETE_CONNECTION_FILE_FAILED
SHUTDOWN_REPLY_TIMED_OUT
RESTART_KERNEL_FAILED
CORE/ERROR

### DIFF
--- a/packages/core/__tests__/middleware-spec.js
+++ b/packages/core/__tests__/middleware-spec.js
@@ -24,7 +24,7 @@ describe("The error middleware", () => {
       reducer: jest.fn()
     };
     const next = action => store.dispatch(action);
-    const action = { type: "ERROR", payload: "This is a payload", err: true };
+    const action = { type: "ERROR", payload: "This is a payload" };
 
     const fakeConsole = {
       error: msg => {}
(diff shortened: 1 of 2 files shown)


packages/core/__tests__/middleware-spec.js
```diff
@@ -75,4 +75,76 @@ describe("The error middleware", () => {
     });
     expect(store.reducer).toBeCalled();
   });
+  test("treats an action w/ 'error: true' as an error", () => {
```
captainsafia, line 78: Thank you for adding these tests!

packages/core/src/middlewares.js
```diff
@@ -16,13 +17,21 @@ export const errorMiddleware = (
...
+  if (!(action.type.includes("ERROR") || action.error)) {
     return next(action);
   }
   console.error(action);
   let errorText;
+
   if ("payload" in action) {
-    errorText = JSON.stringify(action.payload, null, 2);
+    if (
+      action.payload.hasOwnProperty("error") &&
```
captainsafia: Flow isn't a huge fan of this because the `payload` property is typed as `any` in `type Action` above. You probably want to check that the `payload` is an object first (`action.payload instanceof Object && ...`) to work around this.

captainsafia (commented):
Neat-o! Posted some comments in-line.

packages/core/src/middlewares.js
rgbkrk: This definitely shows off where we were mid-migration to a new structure for the actions and reducers setup that @theengineear and I went through. Probably worth us figuring out what our final direction is for our state structure and finish that work off as a milestone item.

jameskoch: Fixed. I rewrote my commit. Let me know if you all prefer a chain of commits for revisions like this. I've been away from github for a few years as well, so have fallen out of touch w/ the workflow.

rgbkrk:
Thank you so much for figuring this one out.

rgbkrk:

PR #3374, opened 2018-09-28
Adds support for kernel "Restart and Run All Cells"

### LINKED ISSUE #3367: [menu] Add Runtime, Restart and Run All
Coming from classic Notebook, this feature seems to be missing.

#2727 describes jupyter_extension variant of this

I'd like to take a swing at this myself.

jameskoch:
Fixes #3367.

Still a work in progress, but wanted to get some eyes on it, like @jdetle.

TODO:
- [x] Serious review of the redux-observable bits because this is my first time touching it. See kernel-lifecycle.js
- [x] Unit testing around the redux-observable
- [x] Manually test apps other than desktop

### DIFF
--- a/applications/desktop/src/notebook/menu.js
+++ b/applications/desktop/src/notebook/menu.js
@@ -8,7 +8,12 @@ import * as fs from "fs";
 
 import { throttle } from "lodash";
 
-import { actions, selectors, createKernelRef } from "@nteract/core";
+import {
+  actions,
+  actionTypes,
+  selectors,
+  createKernelRef
+} from "@nteract/core";
 import type { AppState, ContentRef, KernelRef } from "@nteract/core";
 
 import type { Store } from "redux";
@@ -87,19 +92,20 @@ export function triggerWindowRefresh(
 
 export function dispatchRestartKernel(
   ownProps: { contentRef: ContentRef },
-  store: Store<AppState, *>
+  store: Store<AppState, *>,
+  outputHandling: actionTypes.RestartKernelOutputHandling
 ) {
   const state = store.getState();
   const kernelRef = selectors.kernelRefByContentRef(state, ownProps);
 
   if (!kernelRef) {
-    console.error("kernelRef not set in state for menu");
+    store.dispatch(actions.coreError(new Error("kernel not set")));
     return;
   }
 
   store.dispatch(
     actions.restartKernel({
-      clearOutputs: false,
+      outputHandling: outputHandling,
       kernelRef,
       contentRef: ownProps.contentRef
     })
@@ -347,26 +353,6 @@ export function dispatchInterruptKernel(
 
     store.dispatch(actions.interruptKernel({ kernelRef }));
   }
-}
-
-export function dispatchRestartClearAll(
-  ownProps: { contentRef: ContentRef },
-  store: Store<AppState, *>
-) {
-  const state = store.getState();
-  const kernelRef = selectors.kernelRefByContentRef(state, ownProps);
-  if (!kernelRef) {
-    store.dispatch(actions.coreError(new Error("kernel not set")));
-    return;
-  }
-
-  store.dispatch(
-    actions.restartKernel({
-      clearOutputs: true,
-      kernelRef,
-      contentRef: ownProps.contentRef
-    })
-  );
 }
 
 export function dispatchZoomIn() {
@@ -647,10 +633,17 @@ export function initMenuHandlers(
     "menu:interrupt-kernel",
     dispatchInterruptKernel.bind(null, opts, store)
   );
-  ipc.on("menu:restart-kernel", dispatchRestartKernel.bind(null, opts, store));
+  ipc.on(
+    "menu:restart-kernel",
+    dispatchRestartKernel.bind(null, opts, store, "None")
+  );
   ipc.on(
     "menu:restart-and-clear-all",
-    dispatchRestartClearAll.bind(null, opts, store)
+    dispatchRestartKernel.bind(null, opts, store, "Clear All")
+  );
+  ipc.on(
+    "menu:restart-and-run-all",
+    dispatchRestartKernel.bind(null, opts, store, "Run All")
   );
   ipc.on("menu:theme", dispatchSetTheme.bind(null, opts, store));
   ipc.on("menu:set-blink-rate", dispatchSetCursorBlink.bind(null, opts, store));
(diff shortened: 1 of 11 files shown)


jameskoch:
I've tested nteract_on_jupyter and Desktop w/ this change and it works nicely.

I added some unit testing using the RxJS marbles technique. This involved altering restartKernelEpic to accept alternate means of generating new kernel refs. I tried another approach here, stubbing out a simple epic that would mimic receiving the kernel launch action and generating the eventual successful action, but I couldn't get it all to hang together within the simple TestScheduler harness.

@jdetle or @rgbkrk could one of you take a look, at your convenience?

packages/core/src/actionTypes.js
```diff
@@ -690,10 +690,11 @@ export type SetGithubTokenAction = {
 };
 
 export const RESTART_KERNEL = "RESTART_KERNEL";
+export type RestartKernelOutputHandling = "None" | "Clear All" | "Run All";
```
jdetle, line 693: Yay types!

packages/core/src/reducers/core/entities/contents/notebook.js
```diff
@@ -164,7 +164,7 @@ function clearAllOutputs(
   // handle it
   if (
     action.type === actionTypes.RESTART_KERNEL &&
-    !action.payload.clearOutputs
+    action.payload.outputHandling !== "Clear All"
```
jdetle, line 167: it looks like "Clear All" might be worth extracting out to a constant?

applications/desktop/src/notebook/menu.js
```diff
@@ -647,10 +633,17 @@ export function initMenuHandlers(
...
+    dispatchRestartKernel.bind(null, opts, store, "None")
+  );
   ipc.on(
     "menu:restart-and-clear-all",
-    dispatchRestartClearAll.bind(null, opts, store)
+    dispatchRestartKernel.bind(null, opts, store, "Clear All")
+  );
+  ipc.on(
+    "menu:restart-and-run-all",
+    dispatchRestartKernel.bind(null, opts, store, "Run All")
```
rgbkrk, line 646: Each of these seem good to encapsulate into constants. That or make individual actions for each of these that get handled by the epics `ofType(RestartAndRunAll, ClearAll, Restart)`.

packages/core/__tests__/epics/kernel-lifecycle-spec.js
```diff
@@ -171,3 +176,171 @@ describe("watchExecutionStateEpic", () => {
...
+            })
+          })
+        }),
+        app: stateModule.makeAppRecord({
+          notificationSystem: { addNotification: () => {} }
+        })
+      }
+    };
+
+    const testScheduler = new TestScheduler((actual, expected) =>
```
rgbkrk, line 209: I'm so happy you've dove in to using the `TestScheduler`. It's been needed for a while here.

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -198,7 +200,11 @@ export const launchKernelWhenNotebookSetEpic = (
     })
   );
 
-export const restartKernelEpic = (action$: ActionsObservable<*>, store: *) =>
+export const restartKernelEpic = (
+  action$: ActionsObservable<*>,
+  store: *,
+  kernelRefGenerator: () => KernelRef = createKernelRef
```
rgbkrk, line 206: I'm assuming this is here for overriding for tests?

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -223,30 +229,68 @@ export const restartKernelEpic = (action$: ActionsObservable<*>, store: *) =>
...
+          (action: actionTypes.NewKernelAction) =>
+            action.payload.kernelRef === newKernelRef
+        ),
+        take(1),
+        timeout(60000), // If kernel doesn't come up within this interval we will abort follow-on actions.
+        concatMap(_ => {
+          const restartSuccess = actions.restartKernelSuccessful({
+            kernelRef: newKernelRef,
+            contentRef: initiatingContentRef
+          });
```
rgbkrk, line 272: Excellent, this is looking like a great flow. I'll be curious to see how this works out in practice with flaky kernels or poor connections.

rgbkrk (approved):

packages/core/src/epics/kernel-lifecycle.js
jameskoch: Yeah.  I originally tried to address this by combining this epic w/ a dummyKernelStartingEpic that would receive the launchKernelByName(new_UUID) and respond w/ a launchSuccessful(new_UUID), but couldn't get the two epics to interact nicely, presumably because I'm working w/ pure Observables and not the full redux stack. My first time in this stack, so could be off base...

applications/desktop/src/notebook/menu.js
jameskoch: Something like constants feels better; will probably keep these bind calls, and the menu testing code, simpler and easier to grok than passing around higher-order action-building functions.

Trying something along the lines of constants and not loving it:
```
// actionTypes:
export const RestartKernelOutputHandling = {
  // Feels like some double-encoding going on here.
  // Could just assign non-sense values like 1,2,3 but that also feels silly.
  None: "None",
  ClearAll: "Clear All",
  RunAll: "Run All"
};

// And then to use these ...
// ... either export a type w/ a different name ...
export type RestartKernelOutputHandlingType = Keys$<RestartKernelOutputHandling>;

// ... or have downstream code just use Keys$ themselves. 
// Having to remember or consider using Keys$ vs. Values$ seems annoying.
```

So none of these are feeling great to me. Notably the `bind` calls above are type-checked accurately by Flow, however the "Clear All" literal in notebook.js that John flagged does go unchecked. So keeping the literals would be somewhere between wildly unsafe and fully checked.

I'm happy to change to whichever pattern you'd prefer.

rgbkrk: Since flow is happy with these, I'm alright. As long as it feels good to work on, I'd say this is forward progress.

jameskoch:
Alright to self-merge?

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: You are definitely picking it up at rapid pace.

rgbkrk:
Looks great, I'll merge it in now.

PR #3376, opened 2018-09-29
[WIP] React 16+ compliant @nteract/mathjax API

rgbkrk:
This switches us over to the new React 16 API for our mathjax components.

```jsx
var MathJax = require("@nteract/mathjax");

const tex = String.raw`f(x) = \int_{-\infty}^\infty
    \hat f(\xi)\,e^{2 \pi i \xi x}
    \,d\xi`;

<MathJax.Provider>
  <p>
    This is an inline math formula: <MathJax.Node inline>a = b</MathJax.Node>
    <span> and a block one:</span>
    <MathJax.Node>{tex}</MathJax.Node>
  </p>
</MathJax.Provider>;
```

What's left to do:

* [ ] Include tests
* [x] Update documentation
* [ ] Adapt all uses in the monorepo
* [ ] Clean up all TODOs left behind

### DIFF
--- a/packages/mathjax/src/node.js
+++ b/packages/mathjax/src/node.js
@@ -86,7 +86,7 @@ class MathJaxNode_ extends React.Component<Props & MathJaxContextValue, null> {
 
     if (!MathJax || !MathJax.Hub) {
       throw Error(
-        "Could not find MathJax while attempting typeset! It's likely the MathJax script hasn't been loaded or MathJax.Context is not in the hierarchy"
+        "Could not find MathJax while attempting typeset! It's likely the MathJax script hasn't been loaded or MathJax.Context is not in the hierarchy."
       );
     }
 
@@ -98,10 +98,9 @@ class MathJaxNode_ extends React.Component<Props & MathJaxContextValue, null> {
 
     if (forceUpdate || !this.script) {
       this.setScriptText(text);
-    }
-    // As an invariant of above, this shouldn't occur
-    if (!this.script) {
-      return;
+      if (!this.script) {
+        return;
+      }
     }
 
     MathJax.Hub.Queue(MathJax.Hub.Reprocess(this.script, this.props.onRender));
@@ -149,7 +148,7 @@ class MathJaxNode extends React.PureComponent<Props, null> {
     return (
       <MathJaxContext.Consumer>
         {({ MathJax, input, hasProviderAbove }: MathJaxContextValue) => {
-          // If no provider in the above tree, create our own
+          // If there is no <Provider /> in the above tree, create our own
           if (!hasProviderAbove) {
             return (
               <Provider>
--- a/packages/mathjax/src/node.js
+++ b/packages/mathjax/src/node.js
@@ -12,6 +12,8 @@ import MathJaxContext, {
   type MathJaxContextValue
 } from "./context";
 
+import Provider from "./provider";
+
 type Props = {
   inline: boolean,
   children: string,
@@ -21,11 +23,6 @@ type Props = {
 class MathJaxNode_ extends React.Component<Props & MathJaxContextValue, null> {
   script: ?HTMLScriptElement;
   nodeRef: React.ElementRef<*>;
-
-  static defaultProps = {
-    inline: false,
-    onRender: null
-  };
 
   constructor(props: Props & MathJaxContextValue) {
     super(props);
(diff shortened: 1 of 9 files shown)


packages/mathjax/src/node.js
```diff
@@ -68,7 +60,11 @@ class Node extends React.Component<Props, *> {
    * Clear the jax
    */
   clear() {
-    const MathJax = this.context.MathJax;
+    const MathJax = this.props.MathJax;
```
jdfreder, line 63: nit: Why don't you unpack this like you do on line 85?

packages/mathjax/src/node.js
```diff
@@ -86,9 +82,9 @@ class Node extends React.Component<Props, *> {
    * @param { Boolean } forceUpdate
    */
   typeset(forceUpdate: boolean = false) {
-    const { MathJax } = this.context;
+    const { MathJax } = this.props;
 
-    if (!MathJax) {
+    if (!MathJax || !MathJax.Hub) {
       throw Error(
         "Could not find MathJax while attempting typeset! It's likely the MathJax script hasn't been loaded or MathJax.Context is not in the hierarchy"
```
jdfreder: nit: Since this is user facing I think it makes sense to add a period, right? Do you mind doing that while you're touching this file?

packages/mathjax/src/node.js
```diff
@@ -103,6 +99,10 @@ class Node extends React.Component<Props, *> {
     if (forceUpdate || !this.script) {
       this.setScriptText(text);
     }
+    // As an invariant of above, this shouldn't occur
```
jdfreder: Do you want to actually throw an error since it shouldn't occur? Or log it somehow?

packages/mathjax/src/node.js
```diff
@@ -139,8 +139,44 @@ class Node extends React.Component<Props, *> {
...
+  static defaultProps = {
+    inline: false,
+    onRender: null
+  };
+
+  render() {
+    return (
+      <MathJaxContext.Consumer>
+        {({ MathJax, input, hasProviderAbove }: MathJaxContextValue) => {
+          // If no provider in the above tree, create our own
```
jdfreder: nit: "If there is"

packages/mathjax/src/node.js
```diff
@@ -139,8 +139,44 @@ class Node extends React.Component<Props, *> {
...
+          if (!hasProviderAbove) {
```
jdfreder, line 152: This seems a little weird to me for a couple of reasons:

1. This component is aware of the context which it is embedded.
2. Above is ambiguous, even within the context.

Maybe to resolve both, invert it and call it `addProvider` so that the parents are telling the component what to do?

Or if 1 is not an issue, maybe rename it to `withinAProvider` or `withinProvider`?

Please let me know what your thinking is here. I may be missing something fundamental (e.g. it's okay for children to know about their context).

packages/mathjax/src/node.js
```diff
@@ -139,8 +139,44 @@ class Node extends React.Component<Props, *> {
...
+            return (
+              <Provider>
+                <MathJaxNode {...this.props} />
```
jdfreder, line 155: Isn't this recursive, with hasProviderAbove always set to false? I'm having trouble seeing where you flip hasProviderAbove.

jdfreder (commented):
Hey Kyle, looks great! I have a few, mostly nitpick comments (not useful, sorry).

packages/mathjax/src/node.js
rgbkrk: I'll document this one better. Basically, [the consumer's context](https://reactjs.org/docs/context.html#reactcreatecontext) will be the default value passed in the `React.createContext` call. When there is no `<Provider />` somewhere in the tree (above it) that has this node as child, then `hasProviderAbove` defaults to null. When this is the case, we can create a provider and render the requested MathJax node.

rgbkrk: The reason I tried to implement this API was because the `latex` and `markdown` outputs were using this property.

packages/mathjax/src/node.js
rgbkrk: I'm basically saying `this.script` got set via `document.createElement` in the above code, though Flow is saying that `this.script` can be nullable (because of type), even though we've set it earlier on.

packages/mathjax/src/node.js
jdfreder: Oh I see, the provider sets hasProviderAbove!

packages/mathjax/src/node.js
jdfreder: I see now, provider sets this property. If I'm right, ignore my name change suggestion

rgbkrk: Yep!

PR #3378, opened 2018-09-29
update yarn lock with integrity hashes

rgbkrk:
On fresher versions of yarn, these integrity hashes are created. This keeps us up to date. 😄


jdfreder:
Looks like yarn is failing in CI

rgbkrk:
Yeah, looks like the `canvas` optional dependency is making failure happen when it really shouldn't.

PR #3384, opened 2018-09-30
React 16+ compliant @nteract/mathjax API

rgbkrk:
This switches us over to the new React 16 API for our mathjax components.

```jsx
var MathJax = require("@nteract/mathjax");

const tex = String.raw`f(x) = \int_{-\infty}^\infty
    \hat f(\xi)\,e^{2 \pi i \xi x}
    \,d\xi`;

<MathJax.Provider>
  <p>
    This is an inline math formula: <MathJax.Node inline>a = b</MathJax.Node>
    <span> and a block one:</span>
    <MathJax.Node>{tex}</MathJax.Node>
  </p>
</MathJax.Provider>;
```

What's left to do:

* [x] Include tests
* [x] Update documentation
* [x] Adapt all uses in the monorepo
* [x] Clean up all TODOs left behind

### DIFF
--- a/applications/desktop/src/notebook/index.js
+++ b/applications/desktop/src/notebook/index.js
@@ -1,4 +1,5 @@
 /* @flow strict */
+import * as path from "path";
 import * as React from "react";
 import ReactDOM from "react-dom";
 import { ipcRenderer as ipc, remote } from "electron";
@@ -70,6 +71,15 @@ const store = configureStore({
 // Register for debugging
 window.store = store;
 
+const mathJaxPath = path.join(
+  "..",
+  "node_modules",
+  "mathjax-electron",
+  "resources",
+  "MathJax",
+  "MathJax.js?config=electron"
+);
+
 initNativeHandlers(contentRef, store);
 initMenuHandlers(contentRef, store);
 initGlobalHandlers(contentRef, store);
@@ -88,8 +98,7 @@ export default class App extends React.PureComponent<{}, null> {
       <Provider store={store}>
         <React.Fragment>
           <Styles>
-            {/* TODO: Provide MathJax here with mathjax-electron */}
-            <MathJax.Provider input="tex">
+            <MathJax.Provider src={mathJaxPath} input="tex">
               <NotebookApp
                 // The desktop app always keeps the same contentRef in a browser window
                 contentRef={contentRef}
--- a/applications/desktop/src/notebook/index.js
+++ b/applications/desktop/src/notebook/index.js
@@ -88,14 +88,15 @@ export default class App extends React.PureComponent<{}, null> {
       <Provider store={store}>
         <React.Fragment>
           <Styles>
-            <MathJax.Context input="tex">
+            {/* TODO: Provide MathJax here with mathjax-electron */}
+            <MathJax.Provider input="tex">
               <NotebookApp
                 // The desktop app always keeps the same contentRef in a browser window
                 contentRef={contentRef}
                 transforms={transforms}
                 displayOrder={displayOrder}
               />
-            </MathJax.Context>
+            </MathJax.Provider>
 
             <NotificationSystem
               ref={notificationSystem => {
--- a/packages/mathjax/src/provider.js
+++ b/packages/mathjax/src/provider.js
@@ -63,8 +63,10 @@ class Provider extends React.Component<Props, State> {
     if (src == null) {
       return this.onLoad();
     }
-
-    loadScript(src, this.onLoad);
+    if (typeof MathJax === "undefined" || !MathJax || !MathJax.Hub) {
+      return loadScript(src, this.onLoad);
+    }
+    this.onLoad();
   }
 
   onLoad = () => {
--- a/packages/mathjax/src/provider.js
+++ b/packages/mathjax/src/provider.js
@@ -12,18 +12,11 @@ import MathJaxContext, {
 // MathJax expected to be a global and may be undefined
 declare var MathJax: ?MathJaxObject;
 
-declare type onLoad = () => void;
 type Props = {
   src: ?string,
   children: React.Node,
   didFinishTypeset: ?() => void,
-  // Provide a way to override how we load MathJax and callback to the onLoad
-  // For Hydrogen, for instance we can set
-  //
-  //  loader={(onLoad) => loadMathJax(document, onLoad)}
-  //
-  onLoad: ?onLoad,
-  loader: ?(cb: onLoad) => void,
+  onLoad: ?() => void,
   input: "ascii" | "tex",
   delay: number,
   options: ?*,
@@ -43,7 +36,6 @@ class Provider extends React.Component<Props, State> {
       "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-MML-AM_CHTML",
     input: "tex",
     didFinishTypeset: null,
-    loader: null,
     delay: 0,
     options: {},
     loading: null,
@@ -72,11 +64,7 @@ class Provider extends React.Component<Props, State> {
       return this.onLoad();
     }
 
-    if (!this.props.loader) {
-      loadScript(src, this.onLoad);
-    } else {
-      this.props.loader(this.onLoad);
-    }
+    loadScript(src, this.onLoad);
   }
 
   onLoad = () => {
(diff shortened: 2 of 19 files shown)


packages/mathjax/src/provider.js
```diff
@@ -0,0 +1,132 @@
...
+  src: ?string,
+  children: React.Node,
+  didFinishTypeset: ?() => void,
+  // Provide a way to override how we load MathJax and callback to the onLoad
+  // For Hydrogen, for instance we can set
+  //
+  //  loader={(onLoad) => loadMathJax(document, onLoad)}
+  //
+  onLoad: ?onLoad,
+  loader: ?(cb: onLoad) => void,
```
lgeiger: @rgbkrk Do you think we really need the `onLoad` and `loader` props? I think we can get away with only defining `src` for desktop and hydrogen.

rgbkrk: In that case, I'm happy for us to switch to just `src`. This will be a major version bump anyways.

lgeiger: For the desktop this worked fine. I guess it'll work for Hydrogen as well then.

applications/desktop/src/notebook/index.js
```diff
@@ -70,6 +71,15 @@ const store = configureStore({
...
 window.store = store;
 
+const mathJaxPath = path.join(
+  "..",
+  "node_modules",
+  "mathjax-electron",
+  "resources",
+  "MathJax",
+  "MathJax.js?config=electron"
+);
```
rgbkrk, line 81: @jameskoch hopefully getting you unblocked 🔜 , we're pretty close here

packages/mathjax/src/provider.js
```diff
@@ -51,12 +51,18 @@ class Provider extends React.Component<Props, State> {
...
       input: this.props.input,
       hasProviderAbove: true
     };
   }
 
+  static getDerivedStateFromProps(props: Props, state: State) {
+    if (state.input !== props.input) {
+      return { ...state, input: props.input };
+    }
+    return null;
```
lgeiger, line 63: @rgbkrk  Is this what you had in mind for this TODO?

lgeiger:
This is ready for review now.

I added a simple snapshot test, but we still need to figure out how to properly test the MathJax loading

applications/desktop/src/notebook/index.js
rgbkrk: The only thing I think we'll need to do is re-bloat the mathjax electron config to match up with `?config=TeX-MML-AM_CHTML`

packages/mathjax/src/provider.js
rgbkrk: Yeah this is great!

rgbkrk:
Weird... I thought I wrote some code to do automatic handling of making a new Context if one wasn't available.

rgbkrk:
I am somehow up to date with this branch locally and yet not seeing the changes.

lgeiger:
> Weird... I thought I wrote some code to do automatic handling of making a new Context if one wasn't available.

Isn't it [here](https://github.com/nteract/nteract/pull/3384/files#diff-9de225a5a6a9736c030f6572bac8e85bR151)?

applications/desktop/src/notebook/index.js
lgeiger: Switching mathjax-electron to `CHTML` would require a lot of tinkering (read: replacing all of `mathjax-electron`) since it's based on `SVG` rendering. Checkout https://docs.mathjax.org/en/latest/output.html for a comparison between the output formats. Is there a specific reason why you want to switch to `TeX-MML-AM_CHTML`?

What's exactly missing. I remember closely following the jupyter setup when originally putting together the package.

rgbkrk:
D'oh. I was looking at an old branch with similar work and a similar name:

```
(new-context-api-for-mathjax=) kylek@nfml-83G ~/code/src/github.com/nteract/nteract$
```

lgeiger:
> D'oh. I was looking at an old branch with similar work

This branch is on version `-for-real` 😜

rgbkrk:
`-draft3.final-docx.pdf`

applications/desktop/src/notebook/index.js
lgeiger: Is there a reason why we need `MML` and `AM` support? Isn't `TEX` enough?

This way we could use the [`TEX-AMS-CHTML`](http://docs.mathjax.org/en/latest/config-files.html#the-tex-ams-chtml-configuration-file) config, which may loads less javascript.

rgbkrk: This is an area where I'm less knowledgable about internals. @tlyon3 brought forward this setup which added in more LaTeX features (?) that people said resembled the current notebook better. We really need a definitive answer on this we can document here in our own docs.

@mpacer -- what's the ideal mathjax configuration to match well with both classic and jupyterlab?

mpacer: So this is the default configuration from notebook: 
https://github.com/jupyter/notebook/blob/b663bf1b14734d68518e8cdcb1d993d034ede0f9/notebook/base/handlers.py#L245-L247
```
    @property
    def mathjax_config(self):
        return self.settings.get('mathjax_config', 'TeX-AMS-MML_HTMLorMML-full,Safe')
```

lgeiger: Cool, then I think our `mathjax-electron` setup is pretty close here. It renders with SVG elements, but we can handle this in future PRs.

jameskoch:
I don't have much to say along the lines of meaningful code review, but thanks for working on this!

Only random idea I can bring to the 🎉is some sort of CI integration test that ensures desktop app is functional without an Internet connection. I've seen that done on the JVM (security APIs let you chop off internet when launching your tests), and also using containers (at a higher cost of orchestration, because you need to build in an env w/ a connection, but test it without). I couldn't find any prior art here in the first two pages of Google, so mostly thinking this is all a pipe dream.

applications/desktop/src/notebook/index.js
mpacer: Jupyterlab is more complicated since they also allow swapping out mathjax entirely with KaTeX https://github.com/jupyterlab/jupyter-renderers/tree/master/packages/katex-extension

lgeiger: I had a closer look at `mathjax-electron` and I think it's fine to keep it this way. Previous LaTeX issues we're mostly caused by our old Markdown parsing.

I think we should move the discussion to the mathjax-electron repo: https://github.com/nteract/mathjax-electron/issues/30#issuecomment-428723578

lgeiger:
One annoying bug that was introduced in #2350 and is still present in the PR is the wrong escaping of math displayed via `IPython.display.Math`. `IPython.display.Math` wraps it's output in `$$` which aren't escaped by the latex transform:
bildschirmfoto 2018-10-10 um 22 56 47

rgbkrk:
> some sort of CI integration test that ensures desktop app is functional without an Internet connection

I think that's liable for inclusion on the separate [desktop integration tests repo](https://github.com/nteract/desktop-integration-tests) that's been in the works. I've done iptables based limiting before and I think it's reasonable for us to keep it as part of the suite.

rgbkrk:
> wraps it's output in `$$` which aren't escaped by the latex transform

Ok yeah let's adapt the LaTeX transform to strip matching leading and ending dollars. I wish I understood what a `text/latex` really is as a media type since it's not really a full LaTeX document. Maybe we should be using the MathJax.Text nodes that I didn't add into this package. We... could add that in.

applications/desktop/src/notebook/index.js
rgbkrk: Thanks for the pointer to the right mathjax config @mpacer!

People can use our setup along with KaTeX too, this component is scoped only to MathJax. In fact that's exactly what @benjaminabel did for the [gatsby notebook renderer](https://github.com/nteract/nteract/blob/master/packages/notebook-render/src/index.js#L133).

rgbkrk: Also I swear I'm listening now @lgeiger, I'll head over to the mathjax electron issue. 😄

rgbkrk:
Alright, let's close this chapter of MathJax fun and I'll tackle the `<MathJax.Text />` use case for latex output in a separate PR. 😄

PR #3392, opened 2018-10-02
Cleanup duplicated dev dependencies

lgeiger:
Yarn workspaces allow us to either define dev dependencies inside each package or to include them globally. Currently we mix both approaches.

Since most of the dev dependencies are already part of the root `package.json` this PR removes all duplicated entries inside the subpackages.


applications/desktop/package.json
```diff
@@ -114,48 +114,6 @@
...
-    "rxjs": "^5.5.6",
-    "shell-env": "^2.0.0",
-    "spawn-rx": "^2.0.12",
-    "spawnteract": "^5.0.0",
-    "styled-jsx": "^3.0.3-canary.0",
-    "uuid": "^3.1.0",
-    "webfontloader": "^1.6.28",
-    "webpack": "^4.17.2",
-    "yargs": "^12.0.0"
+    "electron": "3.0.2"
```
lgeiger, line 117: This is just for `electron-builder` to know what version to use for rebuilding native modules. We could also use the `electronVersion` key in the builder config, but then renovate wouldn't help us updating.

rgbkrk: Is the dist ok after all this? I suppose it is since webpack handles building these in.

rgbkrk:
The only worrisome bit for me as our root grows is that we can't tell what's using what at a top level. 
¯\\\_(ツ)\_/¯

lgeiger:
> The only worrisome bit for me as our root grows is that we can't tell what's using what at a top level.

True. But to be honest, I also can't tell what's using what right now ¯\_(ツ)_/¯

applications/desktop/package.json
lgeiger: Yes, dist works fine

rgbkrk:
> True. But to be honest, I also can't tell what's using what right now ¯_(ツ)_/¯

😆

rgbkrk:
This needs a rebase, I'm in favor of this though. 😄

PR #3042, opened 2018-07-03
[WIP] chore: bump rxjs and redux-observable

alexandercbooth:
Closes #2793 and #3019.

So far, 
- completed the new imports for rxjs (though entirely possible with this many files that I missed some 😨 )
- started changing the epic pattern to accept state as the `redux-observable` bump requires. Started with core epics since I was using Jext as the test app. Kernel not currently connecting up, though 😞
- started changing the store configuration (I think just Jext so far) for `redux-observable`

## TODO

* [ ] Update redux-observable flow definition
* [ ] Write flow types for creation operators we use
  * [ ] `bindNodeCallback`
  * [ ] `fromEvent`
  * [ ] `merge`
* [ ] Remove all usage of `pluck` due to lack of type safety
* [ ] Reduce the longer `pipe()` chains OR make more definitions of `pipe` that can handle more arguments
* [ ] Send PRs to flow type for each flow definition update

PR #3395, opened 2018-10-03
Flow typed cleanup

lgeiger:
This PR cleans up our flow definitions, with the goal to simplify `rxjs` and `redux-observable` upgrades going on in #3042

This PR introduces the following changes
- I removed all stubs, since to my knowledge they don't provide a lot of benefit, other than cluttering the repo
- installed the latest definitions using `flow-typed install --skip`, skipping stubs for definitions that are not available
- added `redux-observable` typings converted from [typescript](https://github.com/redux-observable/redux-observable/blob/master/index.d.ts) using [`flowgen`](https://github.com/joarwilk/flowgen/) since the previously used typings aren't compatible
- fixed all remaining flow issues and added FixMe's in cases were flow definitions are still incomplete (mainly redux connect)


rgbkrk:
This is great, thank you!

rgbkrk:
We should probably just go ahead and delete the TODOs that come from upstream flow definitions.

lgeiger:
> We should probably just go ahead and delete the TODOs that come from upstream flow definitions.

I agree, but this may cause problems, when later upgrading them via the CLI, because flow-typed will recognize the change and abort

rgbkrk:
I see, so we'll just let [the todo bot](https://todo.jasonet.co/) make the issues, then we'll keep 'em open or close 'em?

lgeiger:
> Then we'll keep 'em open or close 'em?

I'd say close them 😇

PR #3399, opened 2018-10-03
Upgrade rxjs and redux-observable

lgeiger:
This is an cherry-picked version of #3042 on top of #3395

Just parking this here for now, since I probably won't finish this myself. Hopefully this PR should be easier to work with than #3042 due to cleaned up flow types.


rgbkrk:
Looks like we'll need to rebase again. I wonder what else we can pare down from the PR to make this more manageable.

jameskoch:
> Looks like we'll need to rebase again. I wonder what else we can pare down from the PR to make this more manageable.

I'm happy to take the two instances where I added TestScheduler, which I believe changed syntax in newer rxjs builds.

packages/messaging/__tests__/messaging-spec.js
```diff
@@ -99,7 +98,7 @@ describe("ofMessageType", () => {
         tap(val => {
           expect(val.header.msg_type === "a" || val.header.msg_type === "d");
         }),
-        pluck("header", "msg_type"),
+        map(entry => entry.header.msg_type),
```
lgeiger, line 101: Since flow is happy I assume we don't need to check existence of `entry.header`

applications/desktop/__tests__/renderer/epics/close-notebook-spec.js
```diff
@@ -77,12 +72,12 @@ describe("closeNotebookEpic", () => {
...
       };
 
-      const store = buildStore(true);
+      const state = buildState(true);
       const responses = await closeNotebookEpic(
         ActionsObservable.of(
           actions.closeNotebook({ contentRef: "contentRef1" })
         ),
-        store
+        { value: state }
```
lgeiger, line 80: We could build a real `StateObservable` here, but this does the job for now

lgeiger:
This PR is now in a pretty good state. Let's track what's left to do:
- [x] Make tests pass
- [x] Check applications
  - [x] desktop app
  - [x] play
    Fails during creation of store `TypeError: Class constructor Observable cannot be invoked without 'new'`. Maybe webpack problem @rgbkrk?
  - [x] jupyter extension
- [x] Fix `rxjs` and `redux-observable` flow types or type errors in the code ~~(@lgeiger won't do this one 😇)~~

rgbkrk:
I can figure out the play one. This is getting close!

rgbkrk:
Working through play now. Some scattered notes:

```
@nteract/play:  warning  in /Users/kylek/code/src/github.com/nteract/nteract/packages/core/src/epics/contents.js
@nteract/play: "export 'DownloadContent' (imported as 'actionTypes') was not found in '../actionTypes'
```

For some reason when I'm running `play` it's saying it can't find modules that should be installed along with it already, including `ansi-to-react`.

lgeiger:
I can get play to work when removing [`es2015`](https://github.com/nteract/nteract/blob/master/packages/webpack-configurator/index.js#L88) from the `webpack.mainFields` in our shared webpack config. Does this have a lot of negative side effects, or is this a reasonable workaround @rgbkrk?

rgbkrk:
Interesting! I think that's appropriate to remove for mainFields. We really only need it for the sake of setting the "nteractDesktop" mainField. The only potential bad effect is that we won't transpile packages that provide us nice es2015 source code.

rgbkrk:
@lgeiger and I talked about this PR and he's going to see about taking it all the way to completion. I'm going to extract some messaging bits from Lukas' zeromq experimental PR.

types/redux-observables.js
```diff
@@ -0,0 +1,103 @@
+declare type redux$Action = { type: $Subtype<string> }
```
rgbkrk: This file should probably be called `redux-observable.js`, though it doesn't affect anything since these become global types.

applications/jupyter-extension/Untitled.ipynb
```diff
@@ -0,0 +1,23 @@
+{
```
jameskoch: Accidental file addition here?

lgeiger: Whoops, thanks

jameskoch:
Here's one more commit to add to modernize the marble testing. I wasn't sure if I could push directly to your repo; didn't try.

https://github.com/lgeiger/nteract/pull/1/commits/cb00f77b97c55f96e395ee9cd4fcc606c76915a2

rgbkrk:
Here we go!

applications/jupyter-extension/Untitled.ipynb
captainsafia: Is this addition intentional?

applications/play/redux/epics.js
```diff
@@ -45,6 +45,7 @@ const activateServerEpic = action$ =>
               actionsArray.push(actions.killServer({ serverId: oldServerId }));
             }
           }
+          // $FlowFixMe
```
captainsafia, line 48: Can you give some context into why this wasn't easy to type in case someone decides to help with it in the future?

applications/play/redux/epics.js
```diff
@@ -253,6 +255,7 @@ const runSourceEpic = (action$, store) =>
     })
   );
 
+// $FlowFixMe
```
captainsafia, line 258: Any context into what made this hard to type?

packages/core/src/epics/contents.js
```diff
@@ -194,20 +194,21 @@ export function autoSaveCurrentContentEpic(
...
         action: actionTypes.Save | actionTypes.DownloadContent
-      ): ActionsObservable<Action> => {
-        const state = store.getState();
+      ): ActionsObservable<redux$Action> => {
+        const state = state$.value;
 
         const host = selectors.currentHost(state);
         if (host.type !== "jupyter") {
           // Dismiss any usage that isn't targeting a jupyter server
+          // $FlowFixMe
```
captainsafia, line 211: Context into what errors you got while trying to type this?

packages/core/src/epics/kernelspecs.js
```diff
@@ -1,26 +1,26 @@
 // @flow
 import * as actionTypes from "../actionTypes";
 import * as actions from "../actions";
+import { empty, of } from "rxjs";
```
captainsafia, line 4: I notice that `of` is imported in different ways throughout this PR. Is there a reason for this or can we standardize around a certain way?

packages/editor/src/jupyter/complete.js
```diff
@@ -105,11 +105,10 @@ export function codeCompleteObservable(
   editor: CMI,
   message: Object
 ) {
-  // $FlowFixMe: revisite after rxjs@6
   const completion$ = channels.pipe(
     childOf(message),
     ofMessageType("complete_reply"),
-    pluck("content"),
+    map(entry => entry.content),
```
captainsafia, line 111: I noticed the previous comment said this should be revisited after rxjs@6. Was this the change that needed to be made?

captainsafia:
Hahaha. I started reviewing this then left the tab open then when I came back it was already merged.

Most of my comments are just questions.

applications/jupyter-extension/Untitled.ipynb
rgbkrk: This one was outdated, @lgeiger removed it after @jameskoch's review.

rgbkrk:
Hey all good! I would have waited on merging had I known.

packages/core/src/epics/kernelspecs.js
lgeiger: Using `ActionObservables.of` was mostly done to appease flow, everywhere else I tried to use the default `of` imported from `rxjs`.

lgeiger:
The `FlowFixMe`'s inside this PR work around the shortcomings/bugs of our current flow definitions. I should've definitely clarified this in the comments.
Unfortunately I cannot really provide more insights than that. That said, I'll pair up with Jay next week to see if we can get those flow problems with `rxjs` and `redux-observables` ironed out.

packages/core/src/epics/kernelspecs.js
rgbkrk: I think once we're on typescript here, we can use the `rxjs` `of`.

PR #3403, opened 2018-10-06
Add rimraf as a devDependency to packages that use it.

jdfreder:
Problem:
Users cannot develop individual packages outside of the context of the monorepo without rimraf installed globally because the package.json files do not accurately reflect dependencies required in the scripts.

Solution:
Add rimraf to all package.json files that reference it and do not currently have it defined.

### DIFF
--- a/packages/ansi-to-react/package.json
+++ b/packages/ansi-to-react/package.json
@@ -30,5 +30,8 @@
   "peerDependencies": {
     "react": "^16.3.2",
     "react-dom": "^16.3.2"
+  },
+  "devDependencies": {
+    "rimraf": "^2.6.2"
   }
 }
--- a/packages/commutable/package.json
+++ b/packages/commutable/package.json
@@ -30,5 +30,8 @@
     "babel-runtime": "^6.26.0",
     "immutable": "^4.0.0-rc.9",
     "uuid": "^3.1.0"
+  },
+  "devDependencies": {
+    "rimraf": "^2.6.2"
   }
 }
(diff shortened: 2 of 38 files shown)


rgbkrk:
Generally, we only work on the packages within the monorepo itself. There's also the dependence on a centralized babel config and other build tools. You might be interested in the [`yarn workspace` command](https://yarnpkg.com/lang/en/docs/cli/workspace/) for managing packages within the workspace. What do you want to develop outside the context of the monorepo?

cc @lgeiger who has more of an understanding of why the devDeps are only in the root of the monorepo.

jdfreder:
I think we can close this actually - I was trying to `lerna exec npm i` in attempt to resolve a node_modules issue in one of my new packages ... but it turns out the issue is actually with my .tsconfig + yarn workspaces.

jdfreder:
... Although I still think it's a little weird the package.json files don't describe all of the dependencies used in that package. Maybe @lgeiger can comment before I close this?

lgeiger:
Yarn manages all of our development dependencies and hoists them in the top level `node_modules` folder in order to speed up installation and prevent duplicates.

If you want to execute package specific commands without using `lerna exec`, the easiest way is to `cd` into the desired package and run it via yarn: `yarn run my-build-command` (note that `npm run my-build-command` will not work due to the shared node modules).

In general I'm not against moving dev dependencies into their respective packages, but to have our mono repo in a consistent state we should do this for *all* deps, including our shared babel configs. In my opinion this would introduce quite a bit of tedious work and add a lot of unnecessary boilerplate.

jdfreder:
That makes sense to me, thank you.

PR #3411, opened 2018-10-09
Add menu items and shortcuts for cell actions

rsnk96:
Added menu options for 
* Delete cell (`CmdorCtrl Shift D`)
* Change cell type to md/code (`CmdorCtrl Shift M/Y`)
* Insert cell above (`CmdorCtrl Shift A`)
* Insert cell below (`CmdorCtrl Shift B`)

Choice of shortcuts: For similarity to jupyter-notebook. Also had to remove shortcut for inserting a text cell below as the same shortcut was to be used for changing cell type to Markdown

Updated `USER_GUIDE.md` with a table of cell shortcuts

### DIFF
--- a/applications/desktop/src/main/menu.js
+++ b/applications/desktop/src/main/menu.js
@@ -366,16 +366,24 @@ export function loadFullMenu(store: * = global.store) {
         type: "separator"
       },
       {
-        label: "New Code Cell",
+        label: "Insert Code Cell Above",
         enabled: BrowserWindow.getAllWindows().length > 0,
-        accelerator: "CmdOrCtrl+Shift+N",
-        click: createSender("menu:new-code-cell")
+        accelerator: "CmdOrCtrl+Shift+A",
+        click: createSender("menu:new-code-cell-above")
       },
       {
-        label: "New Text Cell",
+        label: "Insert Code Cell Below",
         enabled: BrowserWindow.getAllWindows().length > 0,
-        accelerator: "CmdOrCtrl+Shift+M",
-        click: createSender("menu:new-text-cell")
+        accelerator: "CmdOrCtrl+Shift+B",
+        click: createSender("menu:new-code-cell-below")
+      },
+      {
+        label: "Insert Text Cell Below",
+        enabled: BrowserWindow.getAllWindows().length > 0,
+        click: createSender("menu:new-text-cell-below")
+      },
+      {
+        type: "separator"
       },
       {
         label: "Copy Cell",
@@ -394,6 +402,12 @@ export function loadFullMenu(store: * = global.store) {
         enabled: BrowserWindow.getAllWindows().length > 0,
         accelerator: "CmdOrCtrl+Shift+V",
         click: createSender("menu:paste-cell")
+      },
+      {
+        label: "Delete Cell",
+        enabled: BrowserWindow.getAllWindows().length > 0,
+        accelerator: "CmdOrCtrl+Shift+D",
+        click: createSender("menu:delete-cell")
       }
     ]
   };
@@ -401,6 +415,21 @@ export function loadFullMenu(store: * = global.store) {
   const cell = {
     label: "Cell",
     submenu: [
+      {
+        label: "Change Cell Type to Code",
+        enabled: BrowserWindow.getAllWindows().length > 0,
+        accelerator: "CmdOrCtrl+Shift+Y",
+        click: createSender("menu:change-cell-to-code")
+      },
+      {
+        label: "Change Cell Type to Text",
+        enabled: BrowserWindow.getAllWindows().length > 0,
+        accelerator: "CmdOrCtrl+Shift+M",
+        click: createSender("menu:change-cell-to-text")
+      },
+      {
+        type: "separator"
+      },
       {
         label: "Run All",
         enabled: BrowserWindow.getAllWindows().length > 0,
--- a/packages/core/src/reducers/core/entities/contents/notebook.js
+++ b/packages/core/src/reducers/core/entities/contents/notebook.js
@@ -410,7 +410,7 @@ function removeCellFromState(
   state: NotebookModel,
   action: actionTypes.RemoveCell
 ) {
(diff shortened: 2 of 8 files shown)


packages/core/src/reducers/core/entities/contents/notebook.js
```diff
@@ -410,7 +410,7 @@ function removeCellFromState(
   state: NotebookModel,
   action: actionTypes.RemoveCell
 ) {
-  const { id } = action.payload;
+  const id = action.payload.id ? action.payload.id : state.cellFocused;
```
rgbkrk, line 413: Smart.

applications/desktop/src/main/menu.js
```diff
@@ -366,16 +366,24 @@ export function loadFullMenu(store: * = global.store) {
...
+        label: "Insert Code Cell Below",
         enabled: BrowserWindow.getAllWindows().length > 0,
-        accelerator: "CmdOrCtrl+Shift+M",
-        click: createSender("menu:new-text-cell")
+        accelerator: "CmdOrCtrl+Shift+B",
+        click: createSender("menu:new-code-cell-below")
+      },
+      {
+        label: "Insert Text Cell Below",
+        enabled: BrowserWindow.getAllWindows().length > 0,
```
rgbkrk, line 382: We should so extract this into a variable within this function in another PR. 😅

rgbkrk:
For CI, it looks like some tests need updates in light of this. If you're able to go after it great, otherwise I'm happy to lend a hand with some new commits.

yarn.lock
```diff
@@ -68,7 +68,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.1.2":
+"@babel/core@^7.0.0":
```
jameskoch: You probably didn't intend to commit this. I haven't worked w/ yarn enough to know if there are easy ways to avoid this. I've just been catching it when manually reviewing my commits.

applications/desktop/src/main/menu.js
```diff
@@ -394,13 +402,34 @@ export function loadFullMenu(store: * = global.store) {
...
+      },
+      {
+        label: "Change Cell Type to Text",
+        enabled: BrowserWindow.getAllWindows().length > 0,
+        accelerator: "CmdOrCtrl+Shift+M",
+        click: createSender("menu:change-cell-to-text")
+      },
+      {
+        type: "separator"
+      },
```
jameskoch, line 432: I imagine we'll want to add the new menu entries to the jupyter-extension as well, for consistency. Code [here](https://github.com/nteract/nteract/blob/master/packages/connected-components/src/notebook-menu/index.js#L122). I'm not seeing where, or if, that mode of operation includes many shortcut keys. @rgbkrk are most extension users you talk to just not leaning on shortcuts much?

rgbkrk: We definitely need shortcuts, people have been asking for them. I'd like to integrate [blueprint's hotkeys](https://blueprintjs.com/docs/#core/components/hotkeys) since we're likely adopting other components from there (beginning in #3412).

yarn.lock
rgbkrk: Hmmm, this should be something that stays in sync. I can check out what's going on here. @lgeiger may have a clue too.

jameskoch:
FWIW, I am working on an inventory of shortcuts defined by default in Lab/Notebook but not available in nteract. Should be able to share shortly.

applications/desktop/src/main/menu.js
jdfreder: I was just talking to M about this

rsnk96:
Hey. Was just wondering where the Merge Cells capability in #314 and #2432 has shifted to, unable to find it in the current codebase @rgbkrk

yarn.lock
rsnk96: Thanks for pointing it out @jameskoch . Changed it back to 7.1.2

applications/desktop/src/main/menu.js
rsnk96: I'm sorry, I didn't quite get what you meant :sweat_smile:

rsnk96:
> For CI, it looks like some tests need updates in light of this. If you're able to go after it great, otherwise I'm happy to lend a hand with some new commits.

Hey. Looks like I'll be needing your help on this one @rgbkrk  :sweat_smile: . Not able to figure out why Circle CI is failing

rgbkrk:
It... doesn't exist as a button anymore. It kept confusing folks. Perhaps we trimmed a bit too much and took out the ability to do it too. We may have to dig up some old commits or visit it anew.

rsnk96:
Yeah, in the absence of a Command Mode, it might get confusing in the cell toolbar if it's just "merge cells". I'll still try digging it up, seems like a nice feature to have if it's user-friendly. And the tests are finally passing @rgbkrk :) . How do you think we can integrate [blueprint's hotkeys](https://blueprintjs.com/docs/#core/components/hotkeys) now?

rgbkrk:
> How do you think we can integrate blueprint's hotkeys now?

For starters we'll want to get #3412 past the finish line so that we have blueprint integrated into the repo.

After that, I think we'll want to define / use `<HotKey>`s at a [global scope](https://blueprintjs.com/docs/#core/components/hotkeys.scope) in the notebook app component for a command or two and see if they collide with the electron native app commands that are registered in the main menu. Might as well try and start to see! Would you want to go after that @rsnk96?

applications/desktop/src/main/menu.js
rgbkrk: We're setting the property `BrowserWindow.getAllWindows().length > 0` in many of these menu items. It only needs to be calculated once, before the creation of this object.

rgbkrk:
Since this is passing and moving forward in the right ways, let's go ahead with merging and iterating. My feedback in this PR can be done in new PRs (refactor the usage of the browser window count calculation, trying out blueprint's hot keys, etc.).

Issue #1503, opened 2017-02-24
Creating a "header editor"


rgbkrk:
As a follow on to #1465 for metadata keys and https://github.com/nteract/commuter/pull/58 (for listing notebooks), I'd like to see a way to edit the "header section" of a notebook. One way that I'd like to try this out is a top level header that appears above all the cells with a slight amount of animation dropping it down:

![screen shot 2017-02-24 at 1 50 20 pm](https://cloud.githubusercontent.com/assets/836375/23322333/3b1be14c-fa98-11e6-9ca3-fc01aa951a44.png)

On a technical level these fields will end up in the notebook metadata:

```
{
  "title": "Named Display Updates",
  "authors": [{"name": "Paul Ivanov"}, {"name": "Kyle Kelley"}],
  "nteract": {
    "tags": ["ipython", "display updates", "progress-bars"],
    "description": "A Progressive look at updating outputs programmatically from IPython"
  }
}
```

<details>
<summary>
Expand this to see the HTML and CSS I mocked this with
</summary>

```html

<div id="hokey-header" style="
    font-family: &quot;Source Sans Pro&quot;;
    background-color: rgb(230, 230, 230);
    margin: 0px;
    padding: 50px;
    box-shadow: 0 1px 1px 0px #c5c5c5;
"><h1 style="
    /* padding: 100px; */
    /* background-color: lightgray; */
    margin-bottom: 30px;
">Learning About Display Updates</h1>
  <h2 style="
    font-weight: normal;
">By <i>Paul Ivanov, Kyle Kelley</i></h2>  
  <div style="
    /* background: #f1f8fa; */
    /* border-radius: 8px; */
    /* display: inline-block; */
"><p style="
    font-family: SourceSansPro-Regular;
    font-size: 18px;
    color: #4078c0;
    margin: 0 0.5em 0.5em 0;
    vertical-align: middle;
    text-align: center;
    padding: 0.2em 0.9em;
    display: inline-block;
    background-color:  #f2f8fa;
    border-radius: 3px;
">ipython</p><p style="
    font-family: SourceSansPro-Regular;
    font-size: 18px;
    color: #4078c0;
    margin: 0 0.5em 0.5em 0;
    vertical-align: middle;
    text-align: center;
    padding: 0.2em 0.9em;
    display: inline-block;
    background-color:  #f2f8fa;
    border-radius: 3px;
">progress-bars</p><p style="
    font-family: SourceSansPro-Regular;
[...]

PR #3412, opened 2018-10-09
Blueprint header editor

rgbkrk:
Introducing a new component in pursuit of #1503 to make an editor for author-relevant metadata of the notebook.

itt_muse_3

The component was crafted by @emeeks while I primarily worked on some build tooling to make blueprint.js easy to work with across the monorepo.

## Tooling

This introduces [blueprint.js](https://blueprintjs.com/) into our codebase as a component toolkit we can adopt throughout the project. Since we want styles to come with the components this includes a way for us to automate vendorizing styles using styled-jsx's webpack loader. It reads the css files from blueprint and generates a css string for embedding in a vendored folder that styled-jsx can use.  🎉 

~~Problem though... 😫 We need a font to be loaded. We load fonts in different ways on desktop, web, and everything else. However, it looks like [we don't actually need the icon font as it has been considered deprecated](https://blueprintjs.com/docs/#core/components/icon). Is my interpretation of the docs and the current release correct @giladgray?~~

From https://github.com/palantir/blueprint/issues/3012#issuecomment-428307629:

> the fonts are still supported (so the styles need to be there) but I don't think they actually need to be served at this point. have you tried simply not providing the font files? i expect all components to work, but you may find a rough edge or two. report those and we'll fix ASAP.

We are now off to the races for using blueprint.js! 💙 

TODO:

* [x] Snapshot tests
* [x] Flow type round up
* [x] ~~Figure out how to do font loading~~ Don't need it, palantir/blueprint#3012 demonstrates we're good

Useful references:

* [nbformat schema for notebook metadata](https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json#L67)


rgbkrk:
Made an issue for my question over on blueprint's repo: https://github.com/palantir/blueprint/issues/3012

rgbkrk:
Would love for you to review this one @alexandercbooth & @captainsafia.

We will also need some flow fixes, I'm pretty happy with this for forward progress. 😄

rgbkrk:
One thing I talked to Elijah about in person was possibly making our use of blueprint css wrapped in one package, possibly `@nteract/vendorized-blueprintjs`. Mostly to encapsulate the styles as is done here, yet not bound to `@nteract/presentational-components`.

In practical terms though, we started thinking about how we'd want to keep a consistent `<Tag />` that we use across the project(s) where our fellow contributors don't need to know about our applied styling on the blueprint `<Tag />` (as is done in this PR). Instead, they import it from one central spot that is our "wrapper" or at least the guide for how we use blueprint. That place... should probably be `@nteract/ion`

```jsx
import { Tag, TagSelection } from "@nteract/ion"

<Tag />
```

The `<Tag />` API should still match blueprint, while being surfaced in our main components docs and including our own styles.

packages/presentational-components/package.json
```diff
@@ -16,8 +16,11 @@
   },
   "dependencies": {
     "@babel/runtime-corejs2": "^7.0.0",
+    "@blueprintjs/core": "^3.7.0",
     "babel-runtime": "^6.26.0",
-    "react-syntax-highlighter": "^9.0.0"
+    "react-syntax-highlighter": "^9.0.0",
+    "styled-jsx": "^3.1.0",
+    "immer": "^1.5.0"
```
BenRussert: I was all excited to see what you would do with `immer`!

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+                  large={true}
+                  minimal={true}
+                  style={authorStyle}
+                  onRemove={
+                    editable
+                      ? () => {
+                          onChange({
+                            ...headerData,
+                            authors: headerData.authors.filter(p => p !== t)
+                          });
```
BenRussert: Enter the same name twice then try to delete one, both would be removed. Just an observation, not going to come up much.

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+              disabled={!editable}
+              onChange={newText => {
+                onChange({
+                  ...headerData,
+                  title: newText
+                });
+              }}
+            />
+          </H1>
+          {headerData.authors.length <= 0 ? null : (
```
BenRussert: I noticed the add author button disappears if you haven't entered any authors yet. You could fix this to work like the tags do if you shift the logic around a little:

```jsx
<p>
              {headerData.authors.length <= 0 ? null : "By "}
              {headerData.authors.map(t => (
                <Tag....
```

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+                  content={<span>Add another author</span>}
+                  position={Position.RIGHT}
+                  usePortal={false}
+                >
+                  <Button
+                    icon="add"
+                    onClick={() => this.setState({ editMode: "author" })}
+                    minimal={true}
+                  />
+                </Tooltip>
```
BenRussert: We could conditionally render this tooltip and add button so it doesnt show when edit is disabled.

BenRussert: Checking out the blueprint docs, you can just add: `disabled={!editable}` to both the tooltip and the button component props.

BenRussert (approved):
This is great, I am very glad we are using these blueprint components, they are amazing!

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+    },
+    onChange: () => {}
+  };
+
+  render() {
+    // Otherwise assume they have their own editor component
+    const { editable, headerData, onChange } = this.props;
+    return (
+      // NOTE: with styled-jsx we have to use React.Fragment, not <> for the time being
+      <React.Fragment>
```
captainsafia: Can we use `header` as the parent tag so this component uses semantic HTML?

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+                  <Button
+                    icon="add"
+                    onClick={() => this.setState({ editMode: "author" })}
+                    minimal={true}
+                  />
+                </Tooltip>
+              )}
+            </p>
+          )}
+          <p>
```
captainsafia: Is there a reason we are using `p` tags and not `div`s?

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+              {(this.state.editMode === "author" && (
+                <Tag style={{ ...authorStyle, color: "black" }}>
+                  <EditableText
+                    maxLength={40}
+                    placeholder="Enter Author Name..."
+                    selectAllOnFocus={true}
+                    onConfirm={e => {
+                      onChange({
+                        ...headerData,
+                        authors: [...headerData.authors, e]
```
captainsafia: Does this account for the fact that the nbformat spec for author is a list of objects?

captainsafia (approved):
Neat-o! Left some comments and questions inline.

packages/presentational-components/package.json
rgbkrk: Oh ha, we'll have to take that out. I added it in thinking we'd use it to manage some local state. 😬

packages/presentational-components/src/components/headerEditor.js
rgbkrk: Smart!

packages/presentational-components/src/components/headerEditor.js
rgbkrk: Good catch, this came from a prior iteration I think.

packages/presentational-components/src/components/headerEditor.js
rgbkrk: It was a little bit easier to work with this in a flat structure. What happened in the discussion that led to us adding it as an object? Was this to leave room for institution and other undecided bits (email, etc.)?

packages/presentational-components/src/components/headerEditor.js
```diff
@@ -0,0 +1,219 @@
...
+                      });
+                      this.setState({ editMode: "none" });
+                    }}
+                    onCancel={() => this.setState({ editMode: "none" })}
+                  />
+                </Tag>
```
rgbkrk: Now that we've done all this work for tag creation, I just found blueprint's [`<TagInput />`](https://blueprintjs.com/docs/#core/components/tag-input) component which also has some other nice props. 😄 

I also really like that their examples make sure to point out anti-patterns and what should be done differently in real use cases:

https://github.com/palantir/blueprint/blob/ac38072b019a21bd6d1fc302180010711610363c/packages/docs-app/src/examples/core-examples/tagInputExample.tsx#L72-L73

packages/presentational-components/src/components/headerEditor.js
alexandercbooth: This was the only issue that jumped out at me. `disabled={!editable}`, nice idea @BenRussert 👍-- the greyed out button looks more consistent this way.

packages/presentational-components/src/components/headerEditor.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+### Uneditable with editable={false}
+
+```js
+<HeaderEditor
+  editable={false}
+  headerData={{
+    authors: ["Elijah Meeks", "Kyle Kelley"],
+    description: "A cool notebook for kids and animals!",
```
alexandercbooth: 👍 💯for animals 😄

alexandercbooth (approved):
Nice work @emeeks and @rgbkrk 👍. I don't have much to add to what folks have already said 🙂. 

This is a great place to get going with these components. In nteract web in particular I have really wanted to be able to edit the name 😓.

packages/presentational-components/src/components/headerEditor.js
captainsafia: Yep. Looks like it was done for flexibility.

https://github.com/jupyter/nbformat/pull/46/files#r80823257

packages/presentational-components/src/components/headerEditor.js
emeeks: As I understood it, we aren't going to allow duplicate tags/authors. So really I should also do a unique check when updating the list. But if we do want to allow duplicates then we need to change this. :/

packages/presentational-components/src/components/headerEditor.js
emeeks: Good catch!

packages/presentational-components/src/components/headerEditor.js
emeeks: So weird I thought I was suppressing these when editable was set to false.

packages/presentational-components/src/components/headerEditor.js
emeeks: Accessiblicious!

packages/presentational-components/src/components/headerEditor.js
emeeks: Updated docs and component to expect an array of `{ name: string }`

packages/presentational-components/src/components/headerEditor.js
rgbkrk: > really I should also do a unique check when updating the list

👍 

> If we do want to allow duplicates then we need to change this

Nah I think these should be unique, even though this list of objects isn't unique as it doesn't have a key for each entry. Maybe it makes sense to have duplicates for Authors, just not with Tags.

rgbkrk:
Ok I've done the one cleanup that was left here. I'm going to add some snapshot tests now.

rgbkrk:
One extra thing I noticed -- due to case inconsistency between the major OSes the file should be called `header-editor.js`. Changing that now.

rgbkrk:
Based on trying to test this, I'm beginning to think we should make some smaller components like this:

```jsx
<AuthorTagInput authors={[]} onChange={} />
```

and wrap them together with a compound header editor component, which can coordinate the passing of data as well as stitching together `onChange` in the same way that's already done in this component, just across multiple smaller components:

```jsx
<HeaderEditor headerData={{}}>
  <Title />
  <Description />
  <AuthorTagInput />
  <TagInput />
</HeaderEditor>
```

Since this is in a good state though, I don't want to hold this up. It should make things easier to test though by having it be a little bit more compositional.

PR #3413, opened 2018-10-09
set jest to only use one worker on circleCI

rgbkrk:
Every now and then we're running out of memory on CircleCI. Jest by default runs in parallel when we really only have one vCPU (even if Linux itself is reporting more). This sets the maxWorkers to 1. Another thing we might consider in the future is [splitting tests along with CircleCI parallelism](https://circleci.com/docs/2.0/parallelism-faster-jobs/).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       # run tests!
       - run: yarn run test:lint
       - run: yarn run build:packages
-      - run: yarn run test
+      - run: yarn run test --maxWorkers 1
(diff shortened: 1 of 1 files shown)


rgbkrk:
This is _one_ of the things causing flaky builds on CircleCI. I would like to also see us invest in more Azure pipelines testing, especially to help with Windows tests. May want to introduce more exhaustive testing to https://github.com/nteract/desktop-integration-tests for that though.

rgbkrk:
Going ahead with merging to help other builds.

Issue #3134, opened 2018-07-29
Update READMEs across the repo to use yarn instead of npm


rgbkrk:
Hey all! 👋 

We're on `yarn` (again...). You'll want to personally upgrade your `yarn` version locally.

As for the bigger issue though -- we need to give people good guidance on using `yarn`. It would be great to have folks go after cleaning up the README, contributing docs, etc. to take this into account. Any takers?

ivanov:
"not on" you mean "now on"?

hydrosquall:
Pushing a PR up for this shortly :)

PR #3415, opened 2018-10-10
[Issue-3134] Replace references to NPM with Yarn in Documentation

hydrosquall:
## Overview
This is a PR to fulfill the request from #3134.

I went about this by doing a global search using Sublime for the following key phrases in all files with the `.md` suffix

- `npm install`
- `npm i` (npm install shorthand)
- `npm t` (npm test)
- `npm r` (npm run)

I believe this covers everything that would affect contributors. 

There were some different instructions used in the release/packaging documents, but I wasn't sure if we wanted to change those since the original issue only asked about changing `README.md` documents / it would require us to add Yarn to some of the CI pipelines. I left those out since I was already touching a large number of files.

These files in question which still include `npm install` references include `PACKAGING-NOTES.md` and the travis / circleci / appveyor CI yaml files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ When changes are made to the root nteract/nteract, they can then be pulled from 
 6. `git pull upstream master`
 7. `yarn clean`
 8. `yarn install`
+7. yarn install
 
 #### Windows Caveats
 
@@ -135,7 +136,7 @@ specific package, use this command, replacing `packageName` with the package you
 want to hack on:
 
 ```
-$(npm bin)/lerna run build --scope packageName
+$(yarn bin)/lerna run build --scope packageName
 ```
 
 ### Hacking on the Desktop application
@@ -147,7 +148,7 @@ yarn app:desktop
 ```
 
 As you make changes, you will have to close the entire app (CMD-q on macOS or
-CNTL-c at the terminal) and then run `npm run app:desktop` again to see the
+CNTL-c at the terminal) and then run `yarn run app:desktop` again to see the
 changes.
 
 #### Progressive Webpack build (automatic)
@@ -200,7 +201,7 @@ rebuild those using [the instructions for building specific packages](#building-
 > I want to debug redux actions and state changes.
 
 * Enable [redux-logger](https://github.com/evgenyrodionov/redux-logger) by
-  spawning the application with `npm run spawn:debug`.
+  spawning the application with `yarn run spawn:debug`.
 
 > I keep getting a pop-up asking: *Do you want the application "nteract Helper.app" to accept
   incoming network connections?* while developing or using a custom build of
--- a/README.md
+++ b/README.md
@@ -131,11 +131,17 @@ to do this.
 
 In some cases you'll want to modify an individual base package (i.e. commutable
 or transforms) and not rebuild all of the other packages. To target a build of a
-specific package, use this command, replacing `packageName` with the package you
+specific package, use this command, replacing `packageName` with the fully qualified name of the package you
 want to hack on:
 
 ```
-$(yarn bin)/lerna run build --scope packageName
+yarn build:only packageName
+```
+
+For example, to hack on the `transforms` package, use
+
+```
+yarn build:only @nteract/transforms
 ```
 
 ### Hacking on the Desktop application
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ To get started developing, [set up the nteract monorepo](#set-the-monorepo-up-in
 
 #### Set the monorepo up in dev mode
 
-Requires [Node.js and npm 3+](https://docs.npmjs.com/getting-started/installing-node#installing-npm-from-the-nodejs-site).
+Requires [Node.js](https://docs.npmjs.com/getting-started/installing-node) and [yarn](https://yarnpkg.com/lang/en/docs/install/).
 
 1. Fork this repo
 2. Clone your fork or this repo `git clone https://github.com/nteract/nteract`
--- a/README.md
+++ b/README.md
@@ -110,7 +110,6 @@ When changes are made to the root nteract/nteract, they can then be pulled from 
 6. `git pull upstream master`
 7. `yarn clean`
 8. `yarn install`
-7. yarn install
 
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -110,6 +110,7 @@ When changes are made to the root nteract/nteract, they can then be pulled from
 6. `git pull upstream master`
 7. `yarn clean`
 8. `yarn install`
+7. yarn install
```
captainsafia: This line is extraneous.

README.md
```diff
@@ -200,7 +201,7 @@ rebuild those using [the instructions for building specific packages](#building-
 > I want to debug redux actions and state changes.
 
 * Enable [redux-logger](https://github.com/evgenyrodionov/redux-logger) by
-  spawning the application with `npm run spawn:debug`.
+  spawning the application with `yarn run spawn:debug`.
```
captainsafia: We can nuke the `run` bit in this line and the one above.

packages/rx-jupyter/README.md
```diff
@@ -35,12 +35,12 @@ To install `rx-jupyter` for development, do the following:
 ```
 git clone https://github.com/nteract/nteract.git
 cd nteract
-npm i
+yarn install
```
captainsafia: We might wanna just do `yarn` here to take advantage of Yarn's terseness. Ditto in the line above.

captainsafia (commented):
Thanks for submitting this pull request! I left a few comments inline.

README.md
hydrosquall: Oops, a blip from resolving merge conflicts. I'll scrub this!

rgbkrk:
Thank you so much for making our contribution and maintenance docs more consistent!

README.md
```diff
@@ -135,7 +135,7 @@ specific package, use this command, replacing `packageName` with the package you
 want to hack on:
 
 ```
-$(npm bin)/lerna run build --scope packageName
+$(yarn bin)/lerna run build --scope packageName
```
rgbkrk: ~~Not that you need to go after fixing this, but I wonder if there's a good way for us to run this directly with yarn.~~ 

I just found out this can be `yarn build:only packageName`, which has to be the fully qualified package name like `yarn build:only @nteract/transforms`.

applications/commuter/README.md
```diff
@@ -29,7 +29,7 @@ Try **commuter** today and take your notebooks wherever you need them.
 ## Installation
 
 ```
-npm install @nteract/commuter -g
+yarn install @nteract/commuter -g
```
rgbkrk: This command, as a global command should either be:

* the `npm` instructions as it tells people how to install the package (rather than work in the monorepo)
* `yarn global add @nteract/commuter` (that's the command for [`yarn`'s global install](https://yarnpkg.com/lang/en/docs/cli/global/)

applications/commuter/README.md
```diff
@@ -86,9 +86,9 @@ COMMUTER_BUCKET=sweet-notebooks commuter
 
 ## Tests
 
-1. `npm t`
+1. `yarn test`
 
 ## Deployment
 
-1. Install commuter cli `npm install @nteract/commuter`
+1. Install commuter cli `yarn install @nteract/commuter`
```
rgbkrk, line 97: It's nice seeing what these docs have, here's another inconistency where it has to be installed globally. Again, we should use either `npm install -g @nteract/commuter` (recommended for user facing directions) or `yarn global add @nteract/commuter`.

rgbkrk:
I'm realizing across here that we have an annoying little problem. Each of our `package/*/README`s are the user facing side which likely need to list _both_ `npm install @nteract/packageName` as well as [`yarn add @nteract/packageName`](https://yarnpkg.com/en/docs/cli/add), and then it's up to the user to determine which they need. Within the monorepo development instructions, everything must be `yarn`.

rgbkrk:
To see what I mean, look at how [commuter is documented on npm](https://www.npmjs.com/package/@nteract/commuter).

packages/ansi-to-react/README.md
```diff
@@ -3,7 +3,7 @@
 Convert ANSI Escape Codes to pretty text output for React.
 
 ```
-npm install --save ansi-to-react
+yarn install --save ansi-to-react
```
rgbkrk: Like mentioned above, these sections are `npm install --save` and `yarn add`. Sucks that there are two options, that's the life for the libraries though. 😅

README.md
hydrosquall: Nice find! I'll expand this section to reflect this new information, thanks for figuring this out.

hydrosquall:
Thank you for the very detailed feedback @rgbkrk and @captainsafia! I don't actually use `yarn` myself, so it was very useful for me to learn the correct syntax for installing things :).

I've pushed up commits that hopefully address the issues raised above, and am happy to continue iterating as necessary.

applications/commuter/README.md
```diff
@@ -28,8 +28,12 @@ Try **commuter** today and take your notebooks wherever you need them.
 
 ## Installation
 
+You may use whichever package manager (`npm` or `yarn`) best suits your workflow. The `nteract` team internally uses `yarn`.
+
 ```
 npm install @nteract/commuter -g
+# OR
+yarn global add @nteract/commuter
```
rgbkrk, line 36: 👍

README.md
```diff
@@ -94,7 +94,7 @@ To get started developing, [set up the nteract monorepo](#set-the-monorepo-up-in
 
 #### Set the monorepo up in dev mode
 
-Requires [Node.js and npm 3+](https://docs.npmjs.com/getting-started/installing-node#installing-npm-from-the-nodejs-site).
+Requires [Node.js](https://docs.npmjs.com/getting-started/installing-node) and [yarn](https://yarnpkg.com/lang/en/docs/install/).
```
rgbkrk, line 97: 👍 good cleanup

rgbkrk (approved):
It looks like there's only one left to fix from an old comment I put in, let's go ahead with all this for now though. Thank you!

PR #3416, opened 2018-10-10
Use markdown parser to escape math delimiters in text/latex

lgeiger:
Fixes https://github.com/nteract/nteract/pull/3384#issuecomment-428729447

## This PR
bildschirmfoto 2018-10-11 um 00 02 44

## v0.11.9
bildschirmfoto 2018-10-11 um 00 02 47

## Jupyter Notebook
bildschirmfoto 2018-10-11 um 00 10 36

### DIFF
--- a/packages/transforms/__tests__/__snapshots__/latex-spec.js.snap
+++ b/packages/transforms/__tests__/__snapshots__/latex-spec.js.snap
@@ -1,10 +1,82 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`LaTeXDisplay processes basic LaTeX 1`] = `
-<MathJaxNode
-  inline={false}
-  onRender={null}
+<LaTeXDisplay
+  data="Pseudo LaTeX $x^2 + y = 3$"
 >
-  x^2 + y = 3
-</MathJaxNode>
+  <MarkdownRender
+    allowedTypes={Array []}
+    source="Pseudo LaTeX $x^2 + y = 3$"
+    unwrapDisallowed={true}
+  >
+    <ReactMarkdown
+      allowedTypes={Array []}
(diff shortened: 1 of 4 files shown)


rgbkrk:
That's kind of hilarious. 😄 

How does classic handle say:

```latex
\begin{equation} \label{eq1}
\begin{split}
A & = \frac{\pi r^2}{2} \\
 & = \frac{1}{2} \pi r^2
\end{split}
\end{equation}
```

or

```latex
\begin{equation} \label{eu_eqn}
e^{\pi i} + 1 = 0
\end{equation}
 
The beautiful equation \ref{eu_eqn} is known as the Euler equation
```

lgeiger:
The classic notebook does better in this case, because our markdown parser doesn't support parsing LaTeX environments like `equation` or `align`.

classic | this
--- | ---
bildschirmfoto 2018-10-11 um 00 18 59|bildschirmfoto 2018-10-11 um 00 19 11

rgbkrk:
Here's how others have done `<MathJax.Text />` nodes before -- https://github.com/wko27/react-mathjax/blob/master/src/Text.js#L19. Perhaps we could/should do the same and just have mathjax typeset a whole div?

lgeiger:
> Perhaps we could/should do the same and just have mathjax typeset a whole div?

We could definitely do that. But since latex environments should also be support in markdown cells, we would need to adapt our markdown parser anyway (at least thats how the classic notebook handles it).

rgbkrk:
Yeah I mean that `text/latex` gets typeset as a "whole div".

For markdown, yeah we'll have to adapt our markdown parser to allow `\begin{equation}`

lgeiger:
OK, let's don't do this lazy approach then

PR #3418, opened 2018-10-11
Add ExecuteResult component

captainsafia:
Decided to duplicate the code from DisplayData instead of setting up some sort of inheritance structure to allow use to have more fine-grained control of the doc strings on the types in the future.


rgbkrk:
This is great, we're getting close to finished up with the components side of this and can start on integration into the notebook app component soon!

PR #3419, opened 2018-10-11
Activate eslint-react plugin

lgeiger:
I noticed we have a lot of unused variables and imports. Configuring configuring the ESLINT React plugin allows us to add the `no-unused-vars` ESLINT rule without complaining on React imports for JSX.

I added the `no-unused-vars` rule as a warning so we can start reducing the amount of unused variables and imports.

### DIFF
--- a/packages/transform-dataresource/src/charts/xyplot.js
+++ b/packages/transform-dataresource/src/charts/xyplot.js
@@ -110,24 +110,26 @@ export const semioticScatterplot = (
         >
           ID, {metric1}, {metric2}
         </h3>
-        {d.binItems.map(d => (
-          <p
-            style={{
-              fontSize: "12px",
-              textTransform: "uppercase",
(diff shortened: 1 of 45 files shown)


rgbkrk:
Excellent, thank you! Great to have guard rails for the stuff like this.

packages/transform-dataresource/src/charts/xyplot.js
```diff
@@ -110,24 +110,26 @@ export const semioticScatterplot = (
...
+        {d.binItems.map(d => {
+          const id = dimensions
+            .map(
+              dim =>
+                (d[dim.name].toString && d[dim.name].toString()) || d[dim.name]
+            )
+            .join(",");
+          return (
+            <p
+              key={id}
```
BenRussert, line 122: Thanks for all these key prop fixes as well!

PR #3425, opened 2018-10-11
Cleanup unused vars

lgeiger:
This is a follow up on #3419.

While most of the fixes are straight forward, in a few places I'm not sure if there are bugs present. I left TODOs in those places.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -9,3 +9,5 @@ build/
 packages/**/node_modules
 applications/**/node_modules
 coverage
+node_modules/
+*/**/__snapshots__/
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,7 +20,7 @@
     "react/display-name": 0,
     "react/no-unescaped-entities": 0,
     "react/no-deprecated": 1,
-    "no-unused-vars": 1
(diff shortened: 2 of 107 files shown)


rgbkrk:
Each of those TODOs can be removed.

It looks like that `const config` (in the commuter view page) needs to be `config =`, since there's a `let config` above. I'm surprised that wasn't caught by flow or other linting.

lgeiger:
> It looks like that const config (in the commuter view page) needs to be config =

Thanks for taking a look 👍 

> Each of those TODOs can be removed.

Also the `writable` TODO? It looks like this should be used in the content response below:

https://github.com/nteract/nteract/blob/cf6606e261900194749ceba0fea2462d6e34812e/applications/desktop/src/notebook/epics/loading.js#L51-L67

rgbkrk:
I think we could just set `writable`, it just happens to be a bit silly since commuter doesn't have a way to write back with local mode.

lgeiger:
> I think we could just set writable, it just happens to be a bit silly since commuter doesn't have a way to write back with local mode.

👍

Issue #3417, opened 2018-10-10
Desktop status bar shows no feedback until kernel process is running


jameskoch:
I've got a slow-starting kernel for reasons I won't go into here.

The status bar in the bottom left doesn't actually change until the kernel process is up, which leaves users wondering if things are progressing.

The "launching" state set in [this reducer](https://github.com/nteract/nteract/blob/54c50e856930e9489e9b920365dacc3f10320af4/packages/core/src/reducers/core/entities/kernels.js#L47-L49) is never rendered.

A proximate cause is that that reducer's call to setIn just ends up creating an Immutable `Map` for our newly-generated kernelRef, rather than a full `LocalKernelRecord` (that gets set later). If I modify the status-bar to check `kernel.get("status")` for the Map case, in addition to the already-checked `kernel.status` for the record case, then I get _closer_ to the feedback behavior I desire. I am still looking around the code for a more redux-idiomatic way to resolve this (always generating the skeleton KernelRecord rather than creating a Map); hints are welcomed 😄 .

This surfaces another issue in that during launch I'm shown `Kernel Display Name | launching` and then once it's live I see `kernel-folder-name | idle`, yielding an odd bit of flicker. Actually, looks like there's already a [TODO in this space](https://github.com/nteract/nteract/blob/8ef408e7d2f2203cf6feb422f04da9f021837c3f/packages/notebook-app-component/src/status-bar.js#L87-L88).

jameskoch:
I'll share a PR for this shortly where I patch up the core kernel reducer only, to produce an Immutable.Record w/ a state field instead of a Map, during launch.

This still feels sub-optimal, but I struggled to find a more elegant solution due to the many different ways in which kernels are launched, with different info available at slightly different times for websocket vs. ZMQ kernels, and the strong local-vs-remote type difference currently modeled. I'll will look at it w/ fresh eyes in the morning.

captainsafia:
Sounds good! Looking forward to reading your PR.

rgbkrk:
We definitely want to move away from `Immutable.Map` in favor of `Immutable.Record` so that's great. Thank you for finding and fixing these inconsistencies. Our transition over from desktop to web (and remote kernels) has not been a full clean sweep. 😬  I'm extremely grateful you keep coming back and making it better.

jameskoch:
Closed by #3456

PR #3427, opened 2018-10-11
Report kernel "launching" properly.

jameskoch:
Currently, reducers in kernel.js react to LAUNCH_KERNEL/LAUNCH_KERNEL_BY_NAME actions by populating `kernels.byRef[newKernelRefUUID]` with a `Map({state: "launching"})`. Later, when more complete kernel info is available (LAUNCH_KERNEL_SUCCESSFUL), the `Map` is replaced with either a LocalKernelRecord or a RemoteKernelRecord. Reading Maps and Records isn't symmetrical and so this creates issues like #3417.

I find myself unfamiliar w/ all the different states kernels might be in, especially for the non-desktop applications. The big question I have is how early in the launch process can we identify the kernel as local vs. remote. If we can do it very early, then I would keep the Local and Remote types as they are, and just create the Records sooner. If we cannot determine local vs. remote very early, I’d recommend changing the Record types so that this early, uncertain, state can be represented cleanly in the store.

This PR assumes we can identify local vs remote early. I modified the LAUNCH_KERNEL actions to require Local vs Remote in their payload. If this feels directionally good, then I could use an assist w/ the TODO-annotated lines, figuring out if/how local-vs-remote can be decided in each case.

### DIFF
--- a/applications/desktop/src/notebook/epics/loading.js
+++ b/applications/desktop/src/notebook/epics/loading.js
@@ -174,6 +174,7 @@ export const launchKernelWhenNotebookSetEpic = (
 
       return actions.launchKernelByName({
         kernelSpecName,
+        kernelType: "zeromq",
         cwd,
         kernelRef: action.payload.kernelRef,
         selectNextKernel: true,
--- a/packages/core/src/epics/kernel-lifecycle.js
+++ b/packages/core/src/epics/kernel-lifecycle.js
@@ -153,11 +153,13 @@ export const extractNewKernel = (
 };
 
 /**
- * NOTE: This function is _exactly_ the same as the desktop loading.js version
- *       with one strong exception -- extractNewKernel
+ * NOTE: This function is _highly similar_ to the desktop loading.js version
+ *       with two exceptions -- extractNewKernel and kernelType
  *       Can they be combined without incurring a penalty on the web app?
- *       The native functions used are `path.dirname`, `path.resolve`, and `process.cwd()`
- *       We could always inject those dependencies separately...
+ *       extractNewKernel - The native functions used are `path.dirname`, `path.resolve`, and `process.cwd()`
+ *                          We could always inject those dependencies separately...
+ *       kernelType - See #3427, this relates to the currently-unrepresentable state of "launching a kernel
+ *                    but we aren't sure if it's local or remote yet".
  */
 export const launchKernelWhenNotebookSetEpic = (
   action$: ActionsObservable<*>,
@@ -189,6 +191,7 @@ export const launchKernelWhenNotebookSetEpic = (
       return of(
         actions.launchKernelByName({
           kernelSpecName,
+          kernelType: "websocket",
           cwd,
           kernelRef: action.payload.kernelRef,
           selectNextKernel: true,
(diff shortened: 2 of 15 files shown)


rgbkrk:
I think we can determine the type early. They should be known a priori.

rgbkrk:
It may mean that the action needs to include the type along with it when the caller dispatches the action though.

applications/desktop/src/notebook/epics/loading.js
```diff
@@ -186,6 +186,7 @@ export const launchKernelWhenNotebookSetEpic = (
 
       return actions.launchKernelByName({
         kernelSpecName,
+        kernelType: "zeromq", // TODO - Is this always the case?
```
jameskoch: Can someone confirm that the desktop app will only use ZMQ kernels? That's my read of the code, but want to make sure. If not, will need to figure out how to resolve name to type here in desktop.

packages/core/src/epics/kernel-lifecycle.js
```diff
@@ -189,6 +189,7 @@ export const launchKernelWhenNotebookSetEpic = (
       return of(
         actions.launchKernelByName({
           kernelSpecName,
+          kernelType: "zeromq", // TODO - Is this always the case?
```
jameskoch: This variant of `launchKernelWhenNotebookSetEpic` is NOT used in the desktop. Would we be okay w/ hard-coding it to use the remote/websocket kernel type for now? (Line above says "zeromq", ignore that, I'm proposing hard-coding it to websocket for now.) If we can't do this, will need an app-agnostic way to determine local-vs-remote here when opening a new notebook.

jameskoch (commented):
@rgbkrk I dropped some more-concrete questions in here as PR comments.

packages/core/src/epics/kernel-lifecycle.js
jameskoch: Or I can drop the kernelType from launchKernelByName, and force the two different epics we have for implementing that epic to do the dirty work.  Will play w/ that right now...

applications/desktop/src/notebook/epics/loading.js
rgbkrk: Right now the desktop app only uses ZMQ kernels. We did want to set up a way to do remote kernels, this never got driven all the way through though.

packages/core/src/epics/kernel-lifecycle.js
rgbkrk: I think hard coding this one to websocket is probably fine.

Ideally, it's about the host we're connected to for which we're local or remote. It's either a native zeromq host (I think we called this desktop host) or a jupyter server host (which we called remote, even if running on localhost). This area definitely needs a revisit.

jameskoch:
@rgbkrk - I updated the tests, comments and commit msg to reflect that this is just hardcoding kernel types in specific apps/components, and that this will some day cause pain and require us to correctly determine local vs remote before updating the state.

I would be fine if you wanted to reject this given the hard-coding, and that more a substantial overhaul to the state and epics is likely the correct long term solution. I am really just interested in resolving #3417, which I can instead do by having the status bar handle the case of "kernelRef state is a Map". Your call.

rgbkrk:
As long as we come back to this, I'm fine with it. You've already been quite adept at fixing some previous old refactors.

These definitely need to be records and writing them down as `Map`s was an oversight. It's also ok if we create an indeterminate type (something like `type: NotReadyKernel`) though I feel like it's known at launch time which kernel type it is, even if the kernel is not fully ready yet.

Given the hesitation, handling the case of the kernel state being a map seems like a decent area we can leave a TODO and come back to it.

jameskoch:
Opened #3456 for a quicker fix to this.

PR #3428, opened 2018-10-11
Fixes desktop reload feature, which I broke in f89e94c89.

jameskoch:
The closeNotebookEpic initiated in the renderer's onbeforeunload was being fired for reload as well, and was hard-coded to finish the workflow by asking the main process to close the window.

The simplest way I found to distinguish between a close and a reload was to wire a custom "reload" message, to be sent from the Reload menu/shortcut in the main process to the
renderer. This controls how the closeNotebookEpic will finish up, messaging back to the main process asking it to either close or reload the window.

I considered renaming the whole epic to couple it less tightly to close and acknowledge that it handles reload, but there are a lot of types/constants around and it doesn't seem worth the effort.

A failed attempt, recorded for posterity:
I initially tried to distinguish between close and reload cases purely within the renderer. By some readings of the docs the close event should fire only on closes, and the beforeunload even should fire in both close and reload, making it possible to distinguish between the two. Unfortunately it seems that close events wired from within the renderer do not actually work.

### DIFF
--- a/applications/desktop/src/notebook/global-events.js
+++ b/applications/desktop/src/notebook/global-events.js
@@ -2,6 +2,8 @@
 
 import type { Store } from "redux";
 
+import { ipcRenderer as ipc } from "electron";
+
 import { selectors } from "@nteract/core";
 
 import type { ContentRef } from "@nteract/core";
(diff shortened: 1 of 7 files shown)


applications/desktop/src/notebook/global-events.js
```diff
@@ -14,9 +16,10 @@ import {
   DESKTOP_NOTEBOOK_CLOSING_READY_TO_CLOSE
 } from "./state.js";
 
-export function beforeUnload(
+export function onBeforeUnloadOrReload(
```
captainsafia, line 19: Style nit: I think it’s OK to have this handler defined without the “on” prefix and just use “on” for the event action below.

captainsafia (approved):
Thanks for leaving the informative code comments and the background to the problem in the issue.

applications/desktop/src/notebook/global-events.js
jameskoch: Please keep the style nits coming; I've been out of the JS world for a few years and need the help.

PR #3431, opened 2018-10-11
Wait until MathJax is configured before passing it to children

lgeiger:
Currently MathJax may error or won't render when opening an existing notebook containing math. This is due to MathJax beeing fully ready when passed down by the provider.

With this PR The Provider waits for MathJax before passing it down. Currently MathJax is only updated inside the Markdown Cells. For normal output this needs some more React wizardry.

## This PR
pr2

## master
master

### DIFF
--- a/packages/mathjax/src/provider.js
+++ b/packages/mathjax/src/provider.js
@@ -84,8 +84,9 @@ class Provider extends React.Component<Props, State> {
     }
 
     const options = this.props.options;
-
-    MathJax.Hub.Config(options);
+    if (options != null && Object.keys(options).length > 0) {
+      MathJax.Hub.Config(options);
+    }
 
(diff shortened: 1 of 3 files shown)


packages/mathjax/src/provider.js
```diff
@@ -99,6 +100,10 @@ class Provider extends React.Component<Props, State> {
       if (this.props.didFinishTypeset) {
         this.props.didFinishTypeset();
```
lgeiger, line 101: This callback doesn't make much sense in our current setup, since mathjax-electron is configured to skip initial typeset after load.

rgbkrk:
What necessitated switching from PureComponent?

lgeiger:
> What necessitated switching from PureComponent?

MathJax didn't re-render otherwise, probably because the change in the provider state didn't propagate down.

PR #3438, opened 2018-10-12
made raw cells editable

kelleyblackmore:
This adds the ability to edit raw cells

- [x] I have read the [Contributor Guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md)

PR #3439, opened 2018-10-12
fixing linting errrors

kelleyblackmore:
Since there are linting errors on master preventing #3438 from being merged easily


rgbkrk:
Thanks for the linting fixes!

PR #3442, opened 2018-10-12
Fix inconsistent names for functions and actions/types

rsnk96:
* Renamed functions and actions/types so that they are consistent with the Jupyter conventions
* Earlier: Cells were either to be created "before" or "after" the focused cell. However, the functions I added in the menu were in continutation with Jupyter conventions, which use "above" and "below"
* All "before"s changed to "above"
* All "after"s changed to "below"

### DIFF
--- a/applications/desktop/__tests__/renderer/menu-spec.js
+++ b/applications/desktop/__tests__/renderer/menu-spec.js
@@ -6,7 +6,7 @@ import { actions, actionTypes } from "@nteract/core";
 import * as Immutable from "immutable";
 
 describe("dispatchCreateCellAbove", () => {
-  test("dispatches a CREATE_CELL_BEFORE with code action", () => {
+  test("dispatches a CREATE_CELL_ABOVE with code action", () => {
     const store = {
       dispatch: jest.fn()
     };
@@ -16,7 +16,7 @@ describe("dispatchCreateCellAbove", () => {
 
     menu.dispatchCreateCellAbove(props, store);
     expect(store.dispatch).toHaveBeenCalledWith(
-      actions.createCellBefore({
+      actions.createCellAbove({
         cellType: "code",
         contentRef: props.contentRef
       })
@@ -25,7 +25,7 @@ describe("dispatchCreateCellAbove", () => {
 });
 
 describe("dispatchCreateCellBelow", () => {
-  test("dispatches a CREATE_CELL_AFTER with code action", () => {
+  test("dispatches a CREATE_CELL_BELOW with code action", () => {
     const store = {
       dispatch: jest.fn()
     };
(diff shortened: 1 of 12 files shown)


rgbkrk:
I'm cool with the display name (menu name) saying something matching conventions. I'm not a big fan of changing all these internals as it's a breaking change for the packages, so we'd have to call this a major bump and state migration strategies. An alternative would be to duplicate the old name / export as well as acknowledge those action names, so that it's only a minor bump. cc @jdpigeon -- any issues with the interface changing here?

rsnk96:
* > An alternative would be to duplicate the old name / export as well as acknowledge those action names, so that it's only a minor bump. cc @jdpigeon -- any issues with the interface changing here?

  My understanding of what you're trying to suggest is aliasing the function calls. Please correct me if I'm wrong. I'm fine with aliasing, but wouldn't that be redundant and confusing? Also, there is even the `removeCell()` functions which are finally called `Delete Cell`.

* > I'm not a big fan of changing all these internals as it's a breaking change for the packages

  I tried to make sure this doesn't break any other functionality elsewhere and so updated all occurences of the functions/actions. Please let me know how else it could break the package, I thought because it passed all tests it wouldn't be breaking.

I also noticed that `createCellBefore()` (or `createCellAbove()` if using the nomenclature suggested in this PR) doesn't have `source`(source of the cell) as an argument, while `createCellAfter()` does. Any reason why?

rgbkrk:
Yep, this would be aliasing the function calls. We can put an explicit deprecation notice above them.

The reason this is a breaking change is that this package is published publicly and the functions are exported as part of the package.

rgbkrk:
To clarify, I like this as it's great to have common naming!

I'm primarily wanting to make sure there isn't a heavy dependence on the names as is, including the action names, types, and functions (since the actions get used in epics by their `type` name). Since we can't know for sure who is using it and we'd like them to be able to adapt to the new names, we provide both the old and new name so they can make a simple transition.

rsnk96:
I've added back the older function calls and actionTypes with a deprication warning (some as comments and some as `console.log()`s)

I've however updated core functionality of the notebook (at `packages/connected-components/src/notebook-menu/index.js`, `applications/desktop/src/notebook/menu.js` and `packages/notebook-app-component/src/cell-creator.js`) to use the updated function calls.

Is this alright? If it is, I'll go ahead with updating `removeCell()`s in the project too (with a `deleteCell()` maybe?)

captainsafia:
Thanks for all the work you put into this, @rsnk96. This not only matches Jupyter conventions, but conventions used across other products too (Microsoft Word, Google Sheets, Apple Pages). So I think this is a good UX move overall.

Thanks again!

captainsafia:
> Is this alright? If it is, I'll go ahead with updating removeCell()s in the project too (with a deleteCell() maybe?)

Renaming remove to delete is a good call as it also matches conventions used across other products.

Issue #3440, opened 2018-10-12
Add eslint-plugin-react-perf rules to our ESLINT config


lgeiger:
Objects created at render time will often cause a re-render when passed down as a prop, even if they have the same content. We need to reduce this usage especially for critical paths. See #3434

To enforce this, we should add the [`eslint-plugin-react-perf`](https://github.com/cvazac/eslint-plugin-react-perf) rules to our ESLINT config. Which prevent patterns like `<Item config={{foo: "bar"}} />`.

max-arias:
Hey @lgeiger , I'm interested in this one. Mind if I grab it?

lgeiger:
@max-arias That's awesome! Welcome on board 🎉

hydrosquall:
Whoa, didn't know about this rule! Thanks @lgeiger

PR #3443, opened 2018-10-12
Add eslint-plugin-react-perf to .eslint

max-arias:

Adds the `eslint-plugin-react-perf` to `.eslint`

When running `yarn lint`, you'll now see errors like these:
```
applications\play\components\Main.js
161:21  error  prop value should not be an Object created in render()  react-perf/jsx-no-new-object-as-prop
  177:19  error  prop value should not be an Array created in render()   react-perf/jsx-no-new-array-as-prop
  212:22  error  prop value should not be an Object created in render()  react-perf/jsx-no-new-object-as-prop
  233:21  error  prop value should not be an Array created in render()   react-perf/jsx-no-new-array-as-prop
```

Issue: https://github.com/nteract/nteract/issues/3440

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,7 +21,9 @@
     "react/display-name": 0,
     "react/no-unescaped-entities": 0,
     "react/no-deprecated": 1,
-    "no-unused-vars": 2
+    "no-unused-vars": 2,
+    "react-perf/jsx-no-new-object-as-prop": "warn",
+    "react-perf/jsx-no-new-array-as-prop": "warn"
   },
   "settings": {
     "react": {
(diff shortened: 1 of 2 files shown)


rgbkrk:
I wonder if there's a way to set these as warnings so that we can clear through them in separate PRs instead of having one big PR.

rgbkrk:
Ah that's right, we can [set each individual rule](https://eslint.org/docs/user-guide/configuring#extending-configuration-files) like this in the `rules` section:

```
rules: {
  "react-perf/jsx-no-new-object-as-prop": ["warn"] // my best guess
}
```

max-arias:
Good idea @rgbkrk , made `react-perf/jsx-no-new-object-as-prop` and `react-perf/jsx-no-new-array-as-prop` warnings

rgbkrk:
Perfect! Now we can do these bit by bit. 😄

rgbkrk:

PR #3444, opened 2018-10-13
Add Media.JSON component

BenRussert:


packages/outputs/__tests__/media-spec.js
```diff
@@ -15,3 +17,26 @@ describe("HTML", () => {
...
+  it("renders JSON data", () => {
+    const component = renderer
+      .create(<Media.JSON data={population} />)
+      .toJSON();
+    expect(component).toMatchSnapshot();
+  });
+
+  it("updates the theme if it changes", () => {
+    const component = renderer
+      .create(<Media.JSON data={population} />)
```
alexandercbooth: Since this is testing the update if the theme changes, would we want to include the theme change here with something like `<Media.JSON data={population} theme="dark" />`?

packages/outputs/src/components/media/json.js
```diff
@@ -0,0 +1,71 @@
...
+    if (
+      nextProps.theme !== this.props.theme ||
+      nextProps.data !== this.props.data
+    ) {
+      return true;
+    }
+    return false;
+  }
+
+  //   #3435
```
alexandercbooth, line 51: Nice 👍

alexandercbooth (approved):
This is great, thanks for adding this @BenRussert!

packages/outputs/__tests__/media-spec.js
BenRussert: Yep, that's what I meant to do! I'll fix that

BenRussert:
Ok, I fixed that test and squashed @alexandercbooth

alexandercbooth:
Awesome, thanks @BenRussert!

PR #3445, opened 2018-10-13
Integration points for jupyter-widgets in nteract.

jdfreder:
This commit contains the integration points required to integrate Jupyter
Widgets into nteract. It does not contain any widget logic, which will
be added in subsequent commits.

Design doc: https://docs.google.com/document/d/1gSr6SAbg_IqxFbdYulsZ0w248sYG4FamQb6o35VOnEg/edit#

#### Jupyter Notebook support of widgets
Widget state was stored at the notebook level, and mirror by views displayed in the cell output:

old

#### Proposed nteract support of widgets
Notebook level state will be moved into outputs, and for now duplicated. In the future we can be clever to eliminate duplicated state. The advantage of this is that we can greatly simplify the integration with nteract, and other application like the colab notebook.

new

The component looks something like this:

component

DO NOT SUBMIT until a sufficient amount of people have had a change 
to see this.


jdfreder:
@rgbkrk , please add additional reviewers as you see fit!

applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.js
```diff
@@ -107,6 +108,12 @@ export default class Notebook extends React.Component<Props, State> {
       }
     );
 
+    import(/* webpackChunkName: "jupyter-widgets" */ "@nteract/transform-widgets").then(
+      module => {
+        this.registerTransform(module.WidgetDisplay);
+      }
+    );
```
rgbkrk, line 116: Excellent, thanks for the dynamic loading.

applications/jupyter-extension/nteract_on_jupyter/package.json
```diff
@@ -37,6 +37,7 @@
     "@nteract/transform-model-debug": "^3.2.3",
     "@nteract/transform-plotly": "^3.2.3",
     "@nteract/transform-vega": "^3.2.4",
+    "@nteract/transform-widgets": "^0.0.1",
```
rgbkrk: Before I forget, we should probably call this nteract/jupyter-widgets and drop the transform name (it can still obey the interface though)

packages/transform-widgets/README.md
```diff
@@ -0,0 +1,33 @@
...
+...
+
+<PureWidgetDisplay 
+    data={{model_id: 'widget-model-guid'} 
+    currentKernel={remoteKernelPropsInstance}}/>
+```
+
+See `RemoteKernelProps` of 
+`/home/jon/github.com/nteract/nteract/packages/core/src/state/entities/kernels.js` 
+for a description of `remoteKernelPropsInstance`.
```
rgbkrk: Thank you for the README! Probably want to update this path so it's not referencing your local system.

packages/transform-widgets/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+import React from "react";
+// import { OuterShim } from "./outer-shim";
+import { actions, selectors } from "@nteract/core";
+import { connect } from "react-redux";
+
+
+type Props = {
+  data: { model_id: string },
+  currentKernel: LocalKernelProps | RemoteKernelProps
```
rgbkrk: I am expecting to fear what's next to come since this expects a kernel.

packages/transform-widgets/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+  static MIMETYPE = "application/vnd.jupyter.widget-view+json";
+
+  // TODO: Uncomment this and related code in a follow-up PR.
+  // The outer shim is responsable for managing the rendered cross domain
+  // iframe, communicating with it, and relaying information to and from the 
+  // kernel.
+  // shim: OuterShim;
+
+  // A reference to the div which we can inject the cross domain widget iframe.
+  container: { current: null | HTMLDivElement } = React.createRef();
```
rgbkrk: 👍

packages/transform-widgets/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+    return false;
+  }
+
+  /** @override */
+  render(): ?React$Element<any> {
+    this.createOrUpdateShim();
+
+    return (<pre>
+      Jupyter-Widgets are not yet supported in nteract.
+      <div ref={this.container} /></pre>);
```
rgbkrk: +1

packages/transform-widgets/src/index.js
rgbkrk: I think there's a better way to type this in accordance with React 16.

packages/transform-widgets/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+    // this.shim.setCommMsgsSubject(this.props.currentKernel.channels);
+    // this.shim.setModelId(this.props.data.model_id);
+  }
+}
+
+// Connect the PureWidgetDisplay component to the current kernel. The resultant
+// component only require display data to render (the widget model id).
+const mapStateToProps = (
+    state: AppState, { contentRef }: { contentRef: ContentRef }) => 
+        ({currentKernel: selectors.currentKernel(state)});
```
rgbkrk: Let me think on what API we should expose here at the top level of the component.

rgbkrk (commented):
The only thing that should likely be done here is to add the further docs relative to this module, moving your current docs into markdown files. :)

packages/transform-widgets/src/index.js
rgbkrk: Ok, I'm glad I made note of this. I would make this package expose an interface where it's required to get a kernel object instead of binding it to the whole `currentKernel` thing when a given page may someday have more than one kernel or content on a page. The default prop can be `null` and we'd pass the kernel object directly like this:

```jsx
<RichMedia data={data}>
  <JupyterWidget kernel={kernel} />
  <DataExplorer />
  <Media.JSON />
  <Media.HTML />
  <Media.Plain />
</RichMedia>
```

jdfreder:
Yeah that makes sense. I think I will go one step further, and require the
"channels" property directly, because that's all that will be used from the
kernel.

packages/transform-widgets/src/index.js
jdfreder: Yeah, really it just requires access to msg channels. I'll restrict the API to clarify.

... I juggled the idea of trying to record and replay comm messages on display, but that's not possible with the spec as-is because it's not symmetric. In other words, state changes caused by interactions with the front-end would not be captured and replayed when widgets are re-displayed (so the state would be butchered).

packages/transform-widgets/README.md
jdfreder: Whoops! Doh! Stupid copy pasta

packages/transform-widgets/src/index.js
jdfreder: Looking at flow's source (https://github.com/facebook/flow/blob/master/lib/react.js#L231), looks like `{current: null | React$ElementRef<ElementType>}` is more correct, although I'm not sure that's what you were thinking.

jdfreder: ... although, changing it to this causes flowtype to fail on the line below where I `<div ref=`. So I've reverted it to its current state.

packages/transform-widgets/src/index.js
jdfreder: Looks like if I'm not going to `connect()` from within the `Jupyter-Widget` component, the next appropriate place is notebook-app,js, from which I'll have to tunnel the propert down through the `<Output>` component. Does that sound right?

jdfreder: So, in the current state of nteract, this would look like the following:

- Connect at `notebook-app-component/src/notebook-app.js`
- Tunnel to `display-area/src/output.js`
- Tunnel to `display-area/src/richest-mime.js`
- Tunnel to `display-area/src/richest-mime.js` end at `<Tranform` which will point to the `JupyterWidgets` media transform.

applications/jupyter-extension/nteract_on_jupyter/package.json
jdfreder: Done.

jdfreder:
Comments addressed, PTAL. Also, there are some unresolved questions that I left in response.

packages/notebook-app-component/src/notebook-app.js
```diff
@@ -223,6 +234,7 @@ class AnyCell extends React.PureComponent<AnyCellProps, *> {
                   transforms={this.props.transforms}
                   theme={this.props.theme}
                   models={this.props.models}
+                  channels={this.props.channels}
```
rgbkrk, line 237: This won't actually pass the channels down (check out `Output` in the display area package for why). You'll want to pass channels directly and _only_ to the widget media type. Since this component doesn't actually need the channels yet, you might want to just forgo passing it through for the moment.

rgbkrk: Oh nevermind, you wove it through. We'll just have to refactor that later once we integrate the [compound version of the Output component](https://components.nteract.io/#output).

packages/jupyter-widgets/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+import React from "react";
+// import { OuterShim } from "./outer-shim";
+import { actions, selectors } from "@nteract/core";
+import { connect } from "react-redux";
+
+
+type Props = {
+  data: { model_id: string },
+  channels: rxjs$Subject<any>,
```
rgbkrk, line 18: To make this forwards compatible with the `Media` components, create a `propType` for `mediaType` that is the same as the constant used in the static `MIMETYPE` used below. I'm going to go ahead and merge this though, so a new PR is welcome.

rgbkrk (approved):

jdfreder:
Woah, I wasn't expecting you to merge this that quickly! I will follow this up with another PR, after COB. This was still missing the design doc -> markdown, too.

rgbkrk:
Gotta keep you coming back. Would love some tests too, even if you start out with [only snapshot tests](https://jestjs.io/docs/en/snapshot-testing).

jdfreder:
Yeah, I'll definitely add some tests. I didn't add any to this PR just because I wanted to establish the integration and API with no functional changes ... oh wait, looks like I snuck in a rendering of "widgets not supported" hehe.

rgbkrk:
That rendering is helpful though! Could even link to the relevant issue and start getting yourself some new contributors to the widgets bits.

PR #3388, opened 2018-10-01
[WIP] Try out zeromq-ng

lgeiger:
This is a proof of concept if it's feasible for us to start using [`zeromq-ng`](https://github.com/rolftimmermans/zeromq-ng). 

`zeromq-ng` is a rewrite of [`zeromq.js`](https://github.com/zeromq/zeromq.js) done by @rolftimmermans using [N-API](https://nodejs.org/api/n-api.html). I like to try this out here and see how it goes.

This means we can use it in Node 10 and Electron 3 and all future versions without rebuilding 🎉.

## Status of the PR

- [x] Get it to work in the desktop app
  I had to rip out `jmp` because ES6 classes of `zeromq-ng/compat` didn't play well with the ES5 style of subclassing in `jmp`. It turned out to be easier to start from scratch instead of converting `jmp` to ES6.
- [ ] Figure out what to do with `ijavascript` since it relies on `zeromq.js`
- [ ] Get multiplatform build working (or provide alternative)
- [ ] Add tests

PR #3447, opened 2018-10-14
[WIP] Raw messaging

rgbkrk:
This takes the wire protocol bits from @lgeiger's #3388 and moves them into here. Since they pretty much come directly from JMP I made sure to add co-author bits to the commits.

I think that `Message` no longer needs to be a class here, it can just be a type. This works for now though.

### DIFF
--- a/packages/enchannel-zmq-backend/src/messages.js
+++ b/packages/enchannel-zmq-backend/src/messages.js
@@ -1,0 +1,144 @@
+// @flow
+
+import { createHmac } from "crypto";
+
+/**
+ *
+ * This file implements the wire protocol for use with zeromq backed kernels
+ * https://jupyter-client.readthedocs.io/en/stable/messaging.html#the-wire-protocol
+ *
+ * Raw protocol example, as multiple frames:
+
+  [
+    'u-u-i-d',         // zmq identity(ies)
+    '<IDS|MSG>',       // delimiter
+    'baddad42',        // HMAC signature
+    '{header}',        // serialized header dict
+    '{parent_header}', // serialized parent header dict
+    '{metadata}',      // serialized metadata dict
+    '{content}',       // serialized content dict
+    '\xf0\x9f\x90\xb1' // extra raw data buffer(s)
+    ...
+  ]
+
+ */
+
+// The delimiter is expected after zmq identities and before the HMAC signature for a message
+const DELIMITER = "<IDS|MSG>";
+
+export class Message {
+  idents: Array<string>;
+  header: Object;
+  parent_header: Object;
+  metadata: Object;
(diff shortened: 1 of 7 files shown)


rgbkrk:
On some level, this is kind of silly unless we go ahead with bringing in `jmp.Socket`. I wonder if it would be worth it to adapt the current code with `jmp.Socket` to use the current version of zeromq. 🤔  Late night musings.

jdfreder:
looks good!

packages/enchannel-zmq-backend/src/messages.js
```diff
@@ -0,0 +1,144 @@
...
+    ...
+  ]
+
+ */
+
+// The delimiter is expected after zmq identities and before the HMAC signature for a message
+const DELIMITER = "<IDS|MSG>";
+
+export class Message {
+  idents: Array<*>; // string ?
```
captainsafia: Based on the 0MQ [docs on identities](http://zguide.zeromq.org/php:chapter3#toc15), I think it is OK to type this as `Array<string>`.

packages/enchannel-zmq-backend/src/messages.js
```diff
@@ -0,0 +1,144 @@
...
+  header: Object;
+  parent_header: Object;
+  metadata: Object;
+  content: Object;
+  buffers: Array<*>; // Array<string> ?
+  constructor(properties: ?Object) {
```
captainsafia: Is there a way to split out the properties into a type definition and use it to explicitly type this parameter?

captainsafia (approved):
Coolio! Left two comments inline.

packages/enchannel-zmq-backend/src/messages.js
rgbkrk: Yeah, we could make a more ideal interface here. Originally @lgeiger just extracted this from https://github.com/n-riesco/jmp and added types. I think we should make this line up with `@nteract/messaging` (or maybe even stick this logic in there?)

rgbkrk: The interface I think I'd want is to just have plain objects and functions, rather than this class structure.

rgbkrk:
Reviving this now and converting it to typescript.

packages/enchannel-zmq-backend/src/messages.js
rgbkrk: Made this use `@nteract/messaging`

rgbkrk:
Something for me to re-look at soon.

rgbkrk:
I'm going to close this and revisit it as part of `kernel-relay` efforts.

PR #3448, opened 2018-10-15
Add Media.SVG component

captainsafia:


packages/outputs/src/components/media/svg.js
```diff
@@ -0,0 +1,40 @@
+/* @flow */
+import React from "react";
+
+type Props = {
+  data: string
+};
+
+export class SVG extends React.Component<Props> {
+  el: ?HTMLElement;
+  static MIMETYPE = "image/svg+xml";
```
BenRussert: One thing we've been changing vs the transforms is using `mediaType` instead of `MIMETYPE` so that these components work with the `RichMedia` component.
```
static defaultProps = {
    mediaType: "image/svg+xml",
    data: ""
  };
```

captainsafia: Roger that! Done.

alexandercbooth (approved):
Looks great to me, thanks @captainsafia!

I think all that is left is to update the snapshot 👍

BenRussert:
Cool, thanks!

PR #3449, opened 2018-10-15
Add styleguidist setup for Media components

captainsafia:


rgbkrk:
At some point we'll want to figure out how to merge the narrative from `RichMedia` into the top level `Media` setup, then make the `RichMedia` component only be about being a "switch" for picking the richest media type.

PR #3453, opened 2018-10-16
Plain and Markdown media components

BenRussert:
* Add Plain and Markdown components
* Fix up exports

### DIFF
--- a/packages/outputs/src/components/media/plain.js
+++ b/packages/outputs/src/components/media/plain.js
@@ -1,0 +1,23 @@
+/* @flow strict */
+import * as React from "react";
+
+import { ansiToInlineStyle } from "ansi-to-react";
+
+type Props = {
+  data: string,
+  mediaType: string
+};
+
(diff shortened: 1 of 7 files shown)


BenRussert:
[Check this out in a real live hydrogen example!](https://github.com/nteract/hydrogen/compare/master...BenRussert:use-new-output-components)

packages/outputs/src/components/media/plain.js
```diff
@@ -0,0 +1,23 @@
...
+      <span style={bundle.style} key={key}>
+        {bundle.content}
+      </span>
+    ))}
+  </pre>
+);
+
+Plain.defaultProps = {
+  mediaType: "text/plain"
+};
```
BenRussert, line 23: If you guys want I will change this back to a class with `shouldComponentUpdate`, but I thought I'd try this.

rgbkrk:
I think this is great, thank you @BenRussert

Issue #3314, opened 2018-09-10
Completion timeout occasionally too aggressive for Almond kernel


jameskoch2:
In general, it would be great if timeouts like this were configurable w/o rebuilding the product.

Using the latest Almond (formerly jupyter-scala) kernel it's easy to spam the Console w/ timeout errors due to the 2s timeout hardcoded here - https://github.com/nteract/nteract/blob/54c50e856930e9489e9b920365dacc3f10320af4/packages/editor/src/jupyter/complete.js#L108-L115

I can collect a histogram of completion times if helpful, but just "feeling it out" at the keyboard the sense I have is that:
1. Most completion requests come back in ~200ms
2. At the start of an expression when the search space is least constrained (ex: "FOOOOOO_" where _ is the cursor) completions regularly take 750-1250ms.
3. Since nteract is attempting completions as I'm typing (👍), the kernel occasionally struggles to keep up (probably queuing them up, though I'm not positive) and the timeouts become common.

It's possible that the kernel could be more intelligent about avoiding nonsense work in this scenario.

jameskoch2:
@alexarchambault might be interested.

rgbkrk:
This is one area where we're not tracking "old" completion requests. In Rx parlance, we should be `switchMap`ping when a new completion request is going to be sent as well as throttling new ones.

alexarchambault:
On the almond side, I've been wanting to look more closely at the completion requests nteract sends to it, but didn't yet… Maybe cancelling (on the kernel side) the currently running completion when a new completion request arrives could help. If that's the case, support for that could be added to almond.

jameskoch:
I've been looking at this. I wanted to explore front-end solutions first, though it might be tricky to keep both fast and slow kernels happy w/ one solution.

The best UX I've been able to obtain involves a debounce on the front end. Any sense if that would be tolerable for other kernels @rgbkrk ? I've been using 200ms which makes a _world_ of difference; I can try ratcheting that down further.

Without some sort of hard-coded debounce, the next best front-end solution I can come up with is capping the number of concurrently active completion requests at one. The downside is that you end up needing to execute two completion requests where one debounced one would have been sufficient. Example:
1. complete("foo.") starts
1. complete("foo.b") encountered; discard since (1) is still running
1. complete("foo.ba") encountered; discard since (1) is still running
1. complete("foo.") finishes; now trigger complete("foo.ba") so we aren't left with non-sense results

[...]

PR #3455, opened 2018-10-16
Debounce code completion requests.

jameskoch:
Related to #3314.

Previously we'd fire off a code completion request to the kernel for almost every keystroke. Some kernels don't currently handle this well. I don't have a view across other kernels, but I know that Almond (Scala) currently queues up these requests and lacks the ability to cancel a request-in-process when a new one arrives.

We can mitigate this partially by debouncing. To be clear, this is not a comprehensive solution. I cannot imagine our aggressive completion attempts working seamlessly for users without a kernel that can prioritize newer requests. But debouncing can make the problem much less pronounced
by reducing the flood of requests that occur while one types quickly.

This PR hard-codes a debounceTime interval of 150ms, such that it should have relatively unnoticeable impact to other kernels. If this proves problematic, I can imagine making this configuration driven, or language driven, or kernel-metadata driven.

Lower-level details:
- Explicit invocation of completion via Ctrl+Space circumvents debouncing since it is unambiguously the right time to go get completions.
- I've made it so that typing a period initiates a completion. It seemed odd to me that many alphanumeric keystrokes would initiate one, but not this.
- Keystrokes and Ctrl+Space will now initiate completion regardless of kernelStatus. Previous code attempted to only initiate non-Ctrl+Space strokes when idle, but had a bug in it where it only checked against the status at mount() time. The feedback I'm getting from my users is that completion behaves inconsistently, so rather than have this subtle dependency that can lead to confusion I'd rather we shoot these requests over to the kernel and let it get to them when it can.
- There are some code gymnastics around dealing w/ CodeMirror completions() callback. I tried to find less obtuse ways to interact w/ CodeMirror but came up empty.
- I reorganized and renamed some tests that seems slightly off. Found two editor tests skip()'d, so rejiggered them to run again, and let me test the new functionality. Things still feel a little janky here, VERY narrow unit testing that method A calls methods B; I can't quite justify the split between the editor itself and ./jupyter/complete.js, which is the particular boundary being tested here. Open to feedback, but mostly think I've made things marginally better, and not worse.
- Given my lack of RxJS experience, slightly worried about memory leaks
  around Subject and manual subscription; so please look from that angle.

### DIFF
--- a/packages/editor/src/index.js
+++ b/packages/editor/src/index.js
@@ -3,10 +3,22 @@
 import * as React from "react";
 import ReactDOM from "react-dom";
 
+import { empty } from "rxjs/observable/empty";
 import { of } from "rxjs/observable/of";
 import { fromEvent } from "rxjs/observable/fromEvent";
+import { merge } from "rxjs/observable/merge";
 import type { Subscription } from "rxjs/Subscription";
-import { switchMap } from "rxjs/operators";
+import { Subject } from "rxjs/Subject";
+import {
+  catchError,
+  debounceTime,
+  map,
+  partition,
+  repeat,
+  switchMap,
+  takeUntil
+} from "rxjs/operators";
+
 
 import { RichestMime } from "@nteract/display-area";
 
@@ -54,6 +66,8 @@ type CodeMirrorEditorState = {
   tipElement: ?any
 };
 
+type CodeCompletionEvent = { editor: Object, callback: Function, debounce: boolean };
+
 class CodeMirrorEditor extends React.Component<
   CodeMirrorEditorProps,
   CodeMirrorEditorState
@@ -62,6 +76,9 @@ class CodeMirrorEditor extends React.Component<
   cm: CMI;
   defaultOptions: Object;
   keyupEventsSubscriber: Subscription;
+  completionSubject: Subject<CodeCompletionEvent>;
+  completionEventsSubscriber: Subscription;
+  debounceNextCompletionRequest: boolean;
 
   static defaultProps = {
     theme: "light",
@@ -82,6 +99,7 @@ class CodeMirrorEditor extends React.Component<
     (this: any).deleteTip = this.deleteTip.bind(this);
     // $FlowFixMe: weirdness in the codemirror API
     this.hint.async = true;
+    this.debounceNextCompletionRequest = true;
     this.state = { isFocused: true, tipElement: null };
 
     this.defaultOptions = Object.assign(
@@ -100,7 +118,10 @@ class CodeMirrorEditor extends React.Component<
           }
         },
         extraKeys: {
-          "Ctrl-Space": "autocomplete",
+          "Ctrl-Space": editor => {
+             this.debounceNextCompletionRequest = false;
+             return editor.execCommand("autocomplete");
+          },
           Tab: this.executeTab,
           "Shift-Tab": editor => editor.execCommand("indentLess"),
           Up: this.goLineUpOrEmit,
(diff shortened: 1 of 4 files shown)


jameskoch:
Hey @rgbkrk I've gotten this debouncing stuff to a good place, IMO. I updated my opening comment here with all the nitty gritty details.

rgbkrk:
Ok excellent, taking a look now.

packages/editor/src/index.js
```diff
@@ -173,27 +194,65 @@ class CodeMirrorEditor extends React.Component<
...
+      partition(ev => ev.debounce === true)
+    );
+
+    const mergedCompletionEvents = merge(
+      immediate,
+      debounce.pipe(
+        debounceTime(150),
+        takeUntil(immediate), // Upon receipt of an immediate event, cancel anything queued up from debounce.
+                              // This handles "type chars quickly, then quickly hit Ctrl+Space", ensuring that it
+                              // generates just one event rather than two.
```
rgbkrk, line 235: Great, this was a good idea. Thank you for the comments explaining why here.

packages/editor/src/jupyter/complete.js
```diff
@@ -112,8 +112,8 @@ export function codeCompleteObservable(
     pluck("content"),
     first(),
     map(expand_completions(editor)),
-    timeout(2000)
-  ); // 2s
+    timeout(15000) // Large timeout for slower languages; this is just here to make sure we eventually clean up resources
```
rgbkrk, line 115: The original amount was arbitrary from me, I'm glad you've bumped this up and stated the intent for cleanup.

rgbkrk (approved):

rgbkrk:
> There are some code gymnastics around dealing w/ CodeMirror completions() callback. I tried to find less obtuse ways to interact w/ CodeMirror but came up empty.

Yeah I've been stuck on that one in the past. It makes me wonder if we should be handling it outside of codemirror and only using their cursor decorations API.

rgbkrk:
> I reorganized and renamed some tests that seems slightly off. Found two editor tests skip()'d, so rejiggered them to run again, and let me test the new functionality. Things still feel a little janky here, VERY narrow unit testing that method A calls methods B; I can't quite justify the split between the editor itself and ./jupyter/complete.js, which is the particular boundary being tested here. Open to feedback, but mostly think I've made things marginally better, and not worse.

These are definitely improved tests.

rgbkrk:
> Given my lack of RxJS experience, slightly worried about memory leaks
around Subject and manual subscription; so please look from that angle.

Let me take another fine toothed comb on this one.

One method we might want to think about is managing all the completion entries outside of codemirror as well as all the rxjs subscriptions (managing them through proper epics and state in redux).

packages/editor/src/index.js
```diff
@@ -269,8 +329,11 @@ class CodeMirrorEditor extends React.Component<
 
   completions(editor: Object, callback: Function): void {
     const { completion, channels } = this.props;
+    const debounceThisCompletionRequest = this.debounceNextCompletionRequest;
+    this.debounceNextCompletionRequest = true;
     if (completion && channels) {
-      codeComplete(channels, editor).subscribe(callback);
+      const el = { editor: editor, callback: callback, debounce: debounceThisCompletionRequest };
+      this.completionSubject.next(el);
```
rgbkrk, line 336: This is a definite improvement since the subscription above technically could have been attached forever. Because we take only one event and make sure to timeout, that's not quite the case though. Still wise to unsubscribe in the unmount like you did above though.

PR #3456, opened 2018-10-16
Report kernel "launching" properly - Take two.

jameskoch:
Previously, reducers in kernel.js reacted to LAUNCH_KERNEL/LAUNCH_KERNEL_BY_NAME by populating kernels.byRef[newKernelRefUUID] with a Map({state: "launching"}). Later, when more complete kernel info is available (LAUNCH_KERNEL_SUCCESSFUL), the Map was replaced with either a LocalKernelRecord or a RemoteKernelRecord. Reading Maps and Records isn't symmetrical and so this creates issues like #3417.

The fix applied here is to create a special type to model the kernel at this early launch point.

See also #3427 which explores an alternate approach of requiring the specification of local vs remote kernel type before dispatching any launch action.

### DIFF
--- a/packages/core/src/state/entities/kernels.js
+++ b/packages/core/src/state/entities/kernels.js
@@ -8,6 +8,35 @@ import { Subject } from "rxjs/Subject";
 import type { KernelInfo } from "./kernel-info";
 export type { KernelInfo };
 
+// See #3427. This represents the kernel early in the launch process.
+// With a bit more work we could probably drop this and just use either
+// Local or RemoteKernelProps as our initial representation of the kernel,
+// deriving local-vs-remote from known sources of truth about kernels.
+export type KernelNotStartedProps = {
+  kernelSpecName: ?string,
+  status: ?string,
+  // The following properties are not known immediately at the start of
+  // launch; they are just included to keep Flow happy and minimize the
+  // impact of this likely-to-be-deleted type.
+  type: "unknown",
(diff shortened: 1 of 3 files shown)


packages/core/src/state/entities/kernels.js
```diff
@@ -8,6 +8,35 @@ import { Subject } from "rxjs/Subject";
...
+  kernelSpecName: ?string,
+  status: ?string,
+  // The following properties are not known immediately at the start of
+  // launch; they are just included to keep Flow happy and minimize the
+  // impact of this likely-to-be-deleted type.
+  type: "unknown",
+  cwd: ".",
+  channels: rxjs$Subject<*>,
+  info: ?KernelInfo
+};
```
rgbkrk, line 25: I'm definitely a fan of this not started type, it's explicit in intent.

packages/core/src/state/entities/kernels.js
```diff
@@ -8,6 +8,35 @@ import { Subject } from "rxjs/Subject";
 import type { KernelInfo } from "./kernel-info";
 export type { KernelInfo };
 
+// See #3427. This represents the kernel early in the launch process.
+// With a bit more work we could probably drop this and just use either
```
rgbkrk, line 12: Thanks for adding these notes.

rgbkrk (approved):

PR #3457, opened 2018-10-16
[WIP] nteract/ion wrapper edition

rgbkrk:
This 🆕 package wraps elements from blueprint.js with CSS provided via `styled-jsx`. By doing this, you can start using:

```jsx
import { Tag } from "@nteract/ion"
```

The approach I took was to wrap _each_ element with the stylesheet. This makes each component work out of the box, but may be a bit of an overkill (it's a lot of CSS text to be passing through). When transpiled, the higher order component looks like this:

```jsx
// Takes in a blueprint component, wraps it with styled-jsx
function provideStyle(Component) {
  return function (props) {
    return React.createElement(React.Fragment, null, React.createElement(Component, props), React.createElement(_style.default, {
      styleId: _blueprintCss.default.__hash,
      css: _blueprintCss.default
    }));
  };
}
```

Another approach would be that whenever someone wants to use a bunch of these elements, they add the styled jsx tag themselves.

Thoughts?

cc @captainsafia @alexandercbooth @emeeks @BenRussert


rgbkrk:
Wow, I had no idea `lerna` would autogenerate a README.

captainsafia:
> Thoughts?

I'm cool with this change. I like that we remove the need for the developer to include the styled-jsx wrapper themselves which adds some friction to the tools.

I'm queasy about including the entire CSS with each component myself but splunking around the [Blueprint issue board](https://github.com/palantir/blueprint/issues/248) shows that this won't be changed any time soon. Maybe we can provide a Webpack config recommendation to remove the duplicate CSS?

packages/ion/src/index.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import * as React from "react";
+
+import blueprintCSS from "./vendor/blueprint.css.js";
+
+import * as Blueprint from "@blueprintjs/core";
+
+// Takes in a blueprint component, wraps it with styled-jsx
+function provideStyle<Props>(
```
alexandercbooth: This seems like the simplest way to add the styles, but to echo @captainsafia's point, it would be nice if we didn't have to pass in the whole css for each.

packages/ion/__tests__/ion.test.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const ion = require('..');
```
alexandercbooth: What are we thinking for how we want testing to work here? Since each wrapped component is tested on Blueprint's end, do we assume that they are unit/integration tested there and we just do snapshots for our wrapped versions?

alexandercbooth (approved):
Thanks for kicking this off, @rgbkrk 👍

packages/ion/__tests__/ion.test.js
rgbkrk: The main thing we probably have to test is that blueprint css -> styled jsx code. I think we can readily assume blueprint is testing the component behavior (and that we'll contribute more tests to the blueprint code base when we find issues).

rgbkrk:
Ok based on this feedback and where we're at, I'm going to say that any apps making use of these components should import our styled jsx css. I'll make it nice to import. Not sure what this means for the style guide though. 🤔

rgbkrk:
> Maybe we can provide a Webpack config recommendation to remove the duplicate CSS?

I'd certainly enjoy something like that, though I don't know how to approach doing that.

rgbkrk:
Closing in favor of #3469.

PR #3465, opened 2018-10-19
Migrate ansi-to-react to TypeScript

jdetle:


rgbkrk:
Wow that was fast.

rgbkrk:
Do our webpack setups need to be adapted to load TS files directly?

jdetle:
Maybe! I suspect that there could be some issues with trying to load typescript files in jest? How can I test in a way that lets us know if we need to modify the webpack config?

rgbkrk:
Does the desktop and jupyter extension behave like normal?

On a secondary note, does the build get triggered when you change this file while running the watched builds?

babel.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = function(api) {
   const env = api.env(); // eslint-disable-line no-unused-vars
 
   const config = {
-    presets: ["@babel/preset-env", "@babel/preset-react", "@babel/preset-flow"],
+    presets: ["@babel/preset-env", "@babel/preset-react", "@babel/preset-typescript", "@babel/preset-flow"],
```
rgbkrk: babel 7 has made this too easy

rgbkrk:
More annoying questions from me since this is the beginning -- how do we know the types are being checked? What will warn? Should we enable something like tslint?

jdetle:
`Tsconfig.json` allows us to use `tsc` to typecheck. Type checking just worked out of the box for me in vscode, but I dont know if theres a necessity to as something like tslint for people who are using other IDEs.

rgbkrk:
Silly me. Yeah I'm thinking we should add `tsc` to our steps for circle ci.

jdetle:
Yeah, like some sort of `tsc packages/**` command?

rgbkrk:
If it alerts on failure, I'm in favor.

lgeiger:
How can we publish automatic typescript definitions for our packages?

rgbkrk:
It looks like we create a `types` field in the package.json of every package, following their [publishing guidelines](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

hydrosquall:
> Maybe! I suspect that there could be some issues with trying to load typescript files in jest? How can I test in a way that lets us know if we need to modify the webpack config?

@jdetle FWIW in the past when doing this, I've brought in https://github.com/kulshekhar/ts-jest

lgeiger:
>  I suspect that there could be some issues with trying to load typescript files in jest

This looks like a good read for our use case: https://medium.com/@francesco.agnoletto/how-to-set-up-typescript-with-babel-and-webpack-6fba1b6e72d5

rgbkrk:
> I suspect that there could be some issues with trying to load typescript files in jest?

I think that jest will handle the ts files just fine via our babel config.

rgbkrk:
That's perfect Lukas!

jdetle:
Brought in some changes from @jdfreder's work, This should pass circleci 🤞

packages/display-area/src/output.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import React from "react";
+// $FlowFixMe
 import Ansi from "ansi-to-react";
```
lgeiger, line 4: Can this be fixed by removing the `ansi-to-react` package from our [`.flowconfig`](https://github.com/nteract/nteract/blob/master/.flowconfig#L10)?

jdetle: I just tested in hopes that it would and unfortunately no, it wont.

lgeiger: Oh probably because it's still ignored then: https://github.com/nteract/nteract/blob/master/.flowconfig#L21

jdetle: We would have to create some sort of stub for ansi-to-react if we didn't want flow to error, right?

lgeiger: > We would have to create some sort of stub for ansi-to-react if we didn't want flow to error, right?

I don't think so, since we have a lot of packages in `node_modules` that don't have subs.

jdetle: Sweet, thanks @lgeiger

package.json
```diff
@@ -51,6 +51,7 @@
     "test:coverage": "npm run test -- --coverage",
     "test:lint": "npm run lint",
     "test:flow": "npm run flow",
+    "test:typescript": "yarn run tsc",
```
lgeiger: Shouldn't it be just `tsc`?  I don't see any `tsc` npm skript  here.

package.json
```diff
@@ -79,7 +80,20 @@
...
+      "jsx",
+      "ts",
+      "tsx"
+    ],
+    "testMatch": [
+      "**/__tests__/**/*.[tj]s?(x)",
+      "**/?(*.)+(spec|test).[tj]s?(x)"
+    ],
+    "transform": {
+      ".*": "babel-jest"
```
lgeiger, line 95: Do we really need this?  It should've been using `babel-jest` before anyway, right?

package.json
```diff
@@ -100,9 +114,11 @@
     "@babel/plugin-syntax-dynamic-import": "^7.0.0",
     "@babel/plugin-transform-flow-strip-types": "^7.0.0",
     "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/plugin-transform-typescript": "^7.1.0",
```
lgeiger: This should be already included in `@babel/preset-typescript`

package.json
```diff
@@ -100,9 +114,11 @@
...
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
+    "@babel/preset-typescript": "^7.1.0",
```
lgeiger: I think we should stick to `^7.0.0` like we do on the other babel packages due to `next.js`

packages/ansi-to-react/package.json
```diff
@@ -5,14 +5,14 @@
...
+    "prepublishOnly": "yarn run build && yarn run build:flow",
+    "build": "yarn run build:clean && yarn run build:lib",
     "build:flow": "flow-copy-source -v -i '**/tests/**' src lib",
     "build:clean": "rimraf lib",
-    "build:lib": "babel -d lib src --ignore '**/__tests__/**' --config-file ../../babel.config.js",
-    "build:lib:watch": "npm run build:lib -- --watch",
-    "build:watch": "npm run build:clean && npm run build:lib:watch"
+    "build:lib": "babel -d lib src --extensions '.ts,.tsx' --ignore '**/__tests__/**' --config-file ../../babel.config.js",
+    "build:lib:watch": "yarn run build:lib -- --watch",
+    "build:watch": "yarn run build:clean && yarn run build:lib:watch"
```
lgeiger: Why the switch to `yarn` inside `npm` skripts`?

packages/ansi-to-react/package.json
```diff
@@ -30,5 +30,9 @@
   "peerDependencies": {
     "react": "^16.3.2",
     "react-dom": "^16.3.2"
+  },
+  "devDependencies": {
+    "@types/react": "^16.4.18",
+    "typescript": "^3.1.3"
```
lgeiger: To be consistent we should move those to the `root` `package.json`

lgeiger (commented):
Thanks a lot for getting this started. I added a few comments and questions since this is the first time using typescript in this repo.

packages/ansi-to-react/package.json
jdetle: Yes. Thanks.

rgbkrk: I keep getting tripped up on this too. 
¯\\\_(ツ)\_/¯

rgbkrk:
I can't wait for this to get merged in, I'm eager to switch my in progress `@nteract/ion` PR over to typescript.

packages/ansi-to-react/package.json
jdetle: I thought that was a thing we we're trying to move over to doing.

.flowconfig
```diff
@@ -18,7 +17,6 @@ emoji=true
 [ignore]
 <PROJECT_ROOT>/node_modules/webpack-cli/.nyc_output/.*
 <PROJECT_ROOT>/packages/.*/node_modules
-<PROJECT_ROOT>/packages/.*/lib
```
rgbkrk, line 21: This line needs to stick around.

packages/ansi-to-react/src/index.ts
```diff
@@ -1,9 +1,15 @@
-/* @flow */
-
 import * as React from "react";
-import { ansiToJson } from "anser";
+import { ansiToJson, AnserJsonEntry } from "anser";
```
captainsafia, line 2: Does `anser` provide a type library that we can use instead of declaring our own types?

packages/ansi-to-react/src/index.ts
```diff
@@ -1,9 +1,15 @@
...
 import { escapeCarriageReturn } from "escape-carriage";
 
+declare type AnsiStyleType = {
+  backgroundColor?: string,
+  color?: string,
+}
+declare type ParsedAnsiBundleType = {
+  style: AnsiStyleType;
+  content: React.ReactNode | string;
```
captainsafia: I believe the `ReactNode` type already includes string as a subtype.

captainsafia (approved):
Neat-o! Left some comments inline.

packages/ansi-to-react/src/index.ts
lgeiger: Oh yeah. Looks like `anser` has typings: https://github.com/IonicaBizau/anser/blob/master/lib/index.d.ts

jdetle:
I'll get some changes up on wednesday, if anyone wants to take this over the finish line feel free.

package.json
jdfreder: I don't think we actually need this line. Running tsc will try to compile, nothing, and succeed.

jdetle: Tsc actually ends up compiling everything without any outDir, so it all gets compiled in place. I think we need to use lerna to invoke tsc from the actual package directory. Does that seem like it makes sense?

I'm probably using tsc here when I actually should be using tslint.

package.json
jdetle: Probably not, I'll check.

packages/ansi-to-react/src/index.ts
jdetle: TIL, thanks

jdfreder:
> I'll get some changes up on wednesday, if anyone wants to take this over the finish line feel free.

I can help try to push this across the finish line !

jdfreder:
:boom: this should be good to go

jdetle:
@jdfreder, as a verification step I tried to add a type and declare it and I got no .d.ts file in `/lib`. I'm afraid I'm not sure how to fix it at the moment.

jdfreder:
We might be missing some Babel config to get it to dump the d.ts. I
wouldn't let a lack of d.ts files hold up the PR, we can solve that in a
follow up PR.

jdetle:
Thats fine with me.

willingc:
Tested yarn build:desktop locally. Looks good. I would ship and iterate from here.

willingc (approved):

package.json
```diff
@@ -29,6 +29,8 @@
     "spawn:debug": "lerna run spawn:debug --scope nteract",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
+    "tslint": "tslint -c tslint.json '**/*{.ts,.tsx}'",
+    "tslint:fix": "tslint --fix -c tslint.json '**/*{.ts,.tsx}'",
```
lgeiger: Why do we need `tslint`? We already use ESLINT and prettier, why the need for a second linter?

I though usually `tsc` is used to do the typechecking: https://github.com/Microsoft/TypeScript-Babel-Starter/blob/8c72057e6b74e1d3969a17be4e169621e0b991a7/package.json#L6-L10

lgeiger:
> We might be missing some Babel config to get it to dump the d.ts. I
wouldn't let a lack of d.ts files hold up the PR, we can solve that in a
follow up PR.

This can be done using `tsc --emitDeclarationOnly`: https://github.com/Microsoft/TypeScript-Babel-Starter/blob/master/package.json#L6-L10 and some [small modifications to the `tsconfig`](https://github.com/Microsoft/TypeScript-Babel-Starter/blob/8c72057e6b74e1d3969a17be4e169621e0b991a7/tsconfig.json#L10)

package.json
jdetle: I believe having tslint gives us the ability to use prettier with typescript? Tsc also having an `--emitDeclarationOnly`, isn't the build step I was imagining. `yarn build` should build the javascript and the types simultaneously. Invoking `yarn build` as a type check in CI seems a bit weird to me.

lgeiger: > I believe having tslint gives us the ability to use prettier with typescript?

Prettier has native support for typescript since version 1.4: https://prettier.io/blog/2017/06/03/1.4.0.html

Don't get me wrong, I'm not suggesting to move away from babel in favor of the typescript compiler for building.

I had a quick look at the [official typescript babel starter example](https://github.com/Microsoft/TypeScript-Babel-Starter/blob/8c72057e6b74e1d3969a17be4e169621e0b991a7/package.json#L6-L10):
They're using `babel` to build, `tsc --noEmit` to typecheck, and `tsc --emitDeclarationOnly` to build the `.d.ts` type definitions.

It seams `tslint` needs quite a lot of [configuration](https://github.com/nteract/nteract/pull/3465/files#diff-ace19bd0c04529e685320269e3c05de9) compared to using `tsc`. What are the benefits of using `tslint` for typechecking rather than the default typescript compiler?

lgeiger: > Don't get me wrong, I'm not suggesting to move away from babel in favor of the typescript compiler for building.

Thinking about this again, it wouldn't be too much of a stretch to go full typescript for our packages later down the road, since our applications and docs all use babel through webpack anyway. Thus rapid prototyping without typechecking would still be possible, since babel just strips the types.

jdetle: Okay, this is all very helpful. Thanks @lgeiger. I think maybe I'm thinking about typescript the wrong way. I will remove tslint based on your feedback. From here we can rely on the build process failing in CI as a sign that our typings are wrong.

jdetle:
@lgeiger provided some quite useful feedback. Throwing back the wip tag for a day.

packages/ansi-to-react/package.json
```diff
@@ -3,14 +3,14 @@
   "version": "3.3.5",
   "description": "ANSI to React Elements",
   "main": "lib/index.js",
+  "types": "lib/index.d.ts",
   "nteractDesktop": "src/index.js",
```
lgeiger: This should be `"nteractDesktop": "src/index.ts",` now

packages/ansi-to-react/package.json
```diff
@@ -3,14 +3,14 @@
...
+  "types": "lib/index.d.ts",
+  "nteractDesktop": "src/index.ts",
   "scripts": {
     "prepare": "npm run build",
     "prepublishOnly": "npm run build && npm run build:flow",
     "build": "npm run build:clean && npm run build:lib",
-    "build:flow": "flow-copy-source -v -i '**/tests/**' src lib",
     "build:clean": "rimraf lib",
-    "build:lib": "babel -d lib src --ignore '**/__tests__/**' --config-file ../../babel.config.js",
+    "build:lib": "tsc",
```
lgeiger, line 13: This will use the typescript compiler instead of babel and typecheck everything.

PR #3466, opened 2018-10-19
Convert jupyter-widgets to typescript.

jdfreder:
and conver jupyter-widgets to typescript...
and ad some sanity tests.


tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "noImplicitAny": true,
+    "module": "esnext",
+    "target": "es2015",
+    "jsx": "react",
+    "baseUrl": "./packages",
+    "lib": ["es2017", "dom"],
+    "moduleResolution": "node",
+    "composite": true,
+    "paths": {
```
rgbkrk: Wait, so `paths` is package names?

jdetle:
@jdfreder it looks like the tests are picking up javascript flowtypes as if they were typescript?

applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
```diff
@@ -39,12 +39,15 @@ module.exports = {
       {
         test: /\.css$/,
         use: ["style-loader", "css-loader"]
+      },
+      {
+        test: /\.tsx?$/,
```
jdetle: Why tsx versus ts?

packages/jupyter-widgets/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.json",
+  "composite": true,
+  "compilerOptions": {
+    "outDir": "./dist/"
```
jdetle: We usually build packages with lib over dist I think? Or is this doing something else?

applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
rgbkrk: Hmmm yeah we want both here.

jdetle: I think we don't want to use the ts-loader at all. We should be using the babel loader at all times.

rgbkrk: ah that's probably true since we're using babel 7

packages/jupyter-widgets/tsconfig.json
rgbkrk: yeah

jdfreder:
Weird, this passed `yarn test` locally.

jdetle:
@jdfreder, what do you want to do you with this?

rgbkrk:
John usually works on this stuff on nights and weekends so you might not see a response until tonight or tomorrow. 😄

jdfreder:
Kyle's right, my day job is mostly unrelated so I'm left to work on this on my time off, sorry. I plan to rebase this off of the PR you and I collaborated on. I probably won't get to it tonight, but I'll let you know as soon as I do.

packages/jupyter-widgets/tsconfig.json
jdfreder: fixed - thanks

jdfreder:
I just rebased this and I doubt it will pass because I haven't tested it locally, but I haven't forgotten about it. Just leaving a comment so that you don't close the PR. :)

rgbkrk:
Cool! One thing you may want to check out is the new doc for the typescript migration: https://github.com/nteract/nteract/blob/master/doc/design/typescript.md

jdfreder:
Nice doc!

tsconfig.json
jdfreder: The asterisk is, yes

packages/jupyter-widgets/package.json
```diff
@@ -6,13 +6,13 @@
   "nteractDesktop": "src/index.js",
   "scripts": {
     "prepare": "npm run build",
-    "prepublishOnly": "npm run build && npm run build:flow",
+    "prepublishOnly": "npm run build",
     "build": "npm run build:clean && npm run build:lib",
     "build:clean": "rimraf lib",
-    "build:flow": "flow-copy-source -v -i '**/__tests__/**' src lib",
     "build:lib": "babel -d lib src --ignore '**/__tests__/**' --config-file ../../babel.config.js",
```
charlesjolley: this script (and a few others I think) are wrong. Correct scripts are in the [TypeScript Doc](https://github.com/nteract/nteract/blob/master/doc/design/typescript.md)

packages/jupyter-widgets/src/index.tsx
```diff
@@ -25,10 +26,10 @@ export class WidgetDisplay extends React.Component<Props, null> {
   // The outer shim is responsable for managing the rendered cross domain
   // iframe, communicating with it, and relaying information to and from the
   // kernel.
-  // shim: OuterShim;
+  // private shim: OuterShim;
 
   // A reference to the div which we can inject the cross domain widget iframe.
-  container: { current: null | HTMLDivElement } = React.createRef();
+  private container: { current: null | HTMLDivElement } = React.createRef();
```
charlesjolley: I would omit the explicit type here, since assigning the createRef value will correctly set the type. createRef has a generic to set the element by the way:  `container = React.createRef<HTMLDivElement>();`

packages/jupyter-widgets/src/index.tsx
```diff
@@ -64,7 +65,7 @@ export class WidgetDisplay extends React.Component<Props, null> {
   }
 
   /** @override */
-  render(): ?React$Element<any> {
+  render(): React.ReactElement<any> {
```
charlesjolley: personal opinion, I would remove the type here altogether. I think this is unnecessarily expanding more strict typing you will get from React. I might be wrong though as I haven't tried it myself.

tsconfig.base.json
```diff
@@ -7,6 +7,9 @@
     "esModuleInterop": true,
     "lib": ["esnext", "dom"],
     "jsx": "react",
-    "typeRoots": ["./node_modules/@types", "types"]
+    "typeRoots": ["./node_modules/@types", "types"],
+    "paths": {
```
charlesjolley: Why did you add the `paths` property back? I took it out because it was causing some conflicts with normal package resolution. Ideally we can find a way to avoid using paths since yarn workspaces already makes local packages available

charlesjolley (changes requested):
Glad to see work on this! I posted a few comments.  I would especially like to avoid adding the paths property back to the tsconfig.base if possible.

packages/jupyter-widgets/package.json
jdfreder: Thanks, yeah I saw the doc because Kyle shared it with me, it's very nice. I haven't had time to actually update this PR until now (I've just been rebasing it so that it doesn't fall too far behind). I'll ping back when it's ready for a review.

jdfreder: :heavy_check_mark:

packages/jupyter-widgets/src/index.tsx
jdfreder: :heavy_check_mark:

packages/jupyter-widgets/src/index.tsx
jdfreder: :heavy_check_mark: that makes sense

tsconfig.base.json
jdfreder: :heavy_check_mark:  - it was left over from earlier

jdfreder:
This is ready to be reviewed and/or merged - please take a look when you have a moment.

charlesjolley (approved):

charlesjolley:
LGTM!

PR #3467, opened 2018-10-19
[WIP] Use Azure pipelines to run all our tests

lgeiger:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,8 +4,23 @@ jobs:
       vmImage: "ubuntu 16.04"
     steps:
       - template: ci-steps.yml
+      - script: yarn test:lint
+        displayName: "Lint"
+      - script: yarn test:flow
+        displayName: "Flow type check"
+      - script: yarn build:packages
+        displayName: "Build packages"
+      - script: yarn docs:build
+        displayName: "Build Docs"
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -6,6 +6,8 @@ jobs:
       - template: ci-steps.yml
 
   - job: Windows
+    variables:
+      npm.config.cache: $(Agent.WorkFolder)/npm-cache
```
lgeiger, line 25: We need to change the `npm` cache folder since `prebuild-install` would fail with
[`EPERM: operation not permitted, rename 'C:\npm\cache\_prebuilds\506537-zeromq-v5.0.0-node-v64-win32-x64.tar.gz.1948-53728ae7d9d52.tmp' -> 'C:\npm\cache\_prebuilds\506537-zeromq-v5.0.0-node-v64-win32-x64.tar.gz'`](https://dev.azure.com/nteract/nteract/_build/results?buildId=452&view=logs)

@vijayma  Is there a better way to do this?

rgbkrk:
Want to have a go at this again? 😸

rgbkrk:
I'm going to assume if you revisit this that you'll make a new PR since we've diverged on master.

cc @captainsafia

PR #3468, opened 2018-10-19
Remove canvas dependency from vega

### LINKED ISSUE #3379: Remove that hidden vega optional dep
As has been pain for us over the last year or so, all the vega dependencies rely on `canvas` as an optional dependency. It's been annoying because _sometimes_ it seems to trip up the CircleCI build and generally confuses contributors to the project as it mixes error messages in what's typically a successful build. For those of us with systems that `canvas` builds just fine on, it makes the build longer.

Since we currently have to have vega2, vega3, vega-lite 1, and vega-lite 2, due to supporting each of the vega media types, I've been thinking that we could just make our own trimmed down forks of each of these (as well as `vega-embed`). I think newer versions of vega (v4) no longer have this issue, as they use canvas-prebuilt.

I'm opening this issue to start tracking some of the relevant issues to this and come up with a plan to handle this.

* https://github.com/vega/vega/issues/1073
* https://github.com/vega/vega-embed/issues/15

lgeiger:
The new [`@nteract/vega-embed-v2`](https://github.com/nteract/vega-embed-v2) package bundles vega using rollup, allowing us to remove the optional canvas dependency that's been spamming us with warnings on every install.

Closes #3379


rgbkrk:
Great plan, thank you!

rgbkrk:
I originally was thinking we'd figure out where our dependencies overlapped so that we could have a minimal build. This makes the most quick sense for this though. Thank you.

README.md
```diff
@@ -113,22 +113,6 @@ When changes are made to the root nteract/nteract, they can then be pulled from
...
-
-```
-npm install -g --production windows-build-tools
-```
-
-Additionally, you'll need to install GTK 2 for node-canvas. Follow [step 2 of
-this instruction
-set](https://github.com/Automattic/node-canvas/wiki/Installation%3A-Windows#2-installing-gtk-2)
-to do this.
-
```
rgbkrk, line 131: Wahoo!

applications/desktop/webpack.common.js
```diff
@@ -5,8 +5,6 @@ const configurator = require("@nteract/webpack-configurator");
 
 const nodeModules = {
   jmp: "commonjs jmp",
-  canvas: "commonjs canvas",
-  "canvas-prebuilt": "commonjs canvas-prebuilt",
```
rgbkrk, line 9: Finally.

lgeiger:
> I originally was thinking we'd figure out where our dependencies overlapped so that we could have a minimal build. This makes the most quick sense for this though. Thank you.

That was my first idea too. Though, doing this correctly would be quite hard since different versions of  Vega require different versions of the same package.

That said, `@nteract/vega-embed-v2` only [bundles `vega`, `vega-lite`, and `vega-embed`](https://github.com/nteract/vega-embed-v2/blob/master/rollup.config.js#L23) and shares all external dependencies:
```js
import datalib from 'datalib';
import d3 from 'd3';
import load from 'datalib/src/import/load';
import d3Cloud from 'd3-cloud';
import jsonStableStringify from 'json-stable-stringify';
import util from 'datalib/src/util';
```

lgeiger:
The final size of the bundle is 741 KB without minification. I think that's not too bad.

rgbkrk:
Woo, I'm pretty happy with that size. We can dynamically load it anyways.

rgbkrk:
Super mega thank you Lukas!

jdfreder:
Thank you!

Issue #3471, opened 2018-10-20
Add missing dependencies to our packages


lgeiger:
One of the inherit problems when developing packages in a monorepo is that it's easy to forget to include all necessary dependencies in the respective `package.json` files. This often leads to bugs when people are trying to use  our packages outside the context of the monorepo.

We should enforce proper declaration of dependencies by adding the [`import/no-extraneous-dependencies`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md) ESLINT rule to the repo and fix the issues it catches.
For this we probably need to ignore our `__test__` directories since, dev dependencies are currently declared in the root `package.json`.

Note: One convention we currently follow is to require `react`, `react-dom` and `immutable` as `peerDependencies` rather than `dependencies`.

jamessral:
I'd love to take a crack at this if that's alright

lgeiger:
> I'd love to take a crack at this if that's alright

That's awesome!

You might want to checkout #3472 since it already adds the plugin as a dependency. Thus you'd only need to add the rule to the config to get started.

lgeiger:
@jamessral #3472 is now merged into master, all set to go 🎉

PR #3472, opened 2018-10-20
Enforce order in our imports

lgeiger:
This adds [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) to our ESLINT setup and enforces correctly ordered imports. This helps to quickly find the right place to add additional dependencies. The nice thing about the plugin is that it integrates well with our prettier setups since this rule is auto-fixable. So almost no action required by our contributors.

This PR was intended to lower the entry barrier for #3471 since it already adds the the plugin to our ESLINT config and I figured I can do a bit of cleanup when I'm at it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,7 +7,14 @@
     "plugin:react-perf/recommended"
   ],
   "parser": "babel-eslint",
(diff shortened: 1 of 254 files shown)


rgbkrk:
This needs a rebase, likely due to removeCell PR I just merged.

rgbkrk (approved):

PR #3473, opened 2018-10-21
Jsral/add missing dependencies

jamessral:
Add missing dependencies to packages

* Attempts to address issue #3471 by adding `import/no-extraneous-dependencies` rule
  and adding appropriate packages to package.json file

### DIFF
--- a/applications/commuter/package.json
+++ b/applications/commuter/package.json
@@ -63,6 +63,7 @@
     "elasticsearch": "^15.0.0",
     "express": "^4.15.4",
     "fs-extra": "^7.0.0",
+    "immutable": "^3.8.2",
     "jquery": "^3.2.1",
     "jsonfile": "^5.0.0",
     "lodash": "^4.17.4",
@@ -76,10 +77,15 @@
     "react": "^16.3.2",
     "react-dom": "^16.3.2",
     "react-hot-loader": "^4.1.2",
+    "rxjs": "^6.3.3",
     "s3-proxy": "^1.1.0",
+    "styled-jsx": "^3.1.0",
     "util": "^0.11.0",
     "winston": "^3.0.0",
     "xhr2": "*",
     "xmlhttprequest": "^1.8.0"
+  },
+  "devDependencies": {
+    "webpack": "^4.21.0"
   }
 }
--- a/applications/desktop/package.json
+++ b/applications/desktop/package.json
@@ -108,12 +108,38 @@
   },
   "dependencies": {
     "@nteract/examples": "3.0.0",
+    "electron-context-menu": "^0.10.0",
+    "electron-log": "^2.2.17",
+    "electron-updater": "^3.1.2",
+    "enchannel-zmq-backend": "^6.3.3",
     "ijavascript": "lgeiger/ijavascript",
+    "immutable": "^3.8.2",
     "jmp": "^2.0.0",
+    "jupyter-paths": "^2.0.0",
+    "kernelspecs": "^2.0.0",
+    "lodash": "^4.17.11",
     "mathjax-electron": "^3.0.0",
-    "nteract-assets": "^4.0.0"
+    "nteract-assets": "^4.0.0",
+    "react": "^16.3.2",
+    "react-dom": "^16.3.2",
+    "react-notification-system": "^0.2.17",
+    "react-redux": "^5.0.7",
+    "redux": "^4.0.1",
+    "redux-electron-store": "^0.6.2",
+    "redux-logger": "^3.0.6",
+    "redux-observable": "^1.0.0",
+    "rxjs": "^6.3.3",
+    "shell-env": "^2.1.0",
+    "spawn-rx": "^3.0.0",
+    "spawnteract": "^5.1.0",
+    "webfontloader": "^1.6.28",
+    "yargs": "^12.0.2"
   },
   "devDependencies": {
-    "electron": "3.0.5"
+    "@nteract/webpack-configurator": "^1.2.4",
+    "electron": "3.0.5",
+    "lodash-webpack-plugin": "^0.11.5",
+    "webpack": "^4.21.0",
+    "webpack-merge": "^4.1.4"
   }
 }
--- a/packages/connected-components/package-lock.json
+++ b/packages/connected-components/package-lock.json
@@ -1,0 +1,597 @@
+{
+  "name": "@nteract/connected-components",
+  "version": "2.0.2",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/node": {
+      "version": "10.12.0",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.12.0.tgz",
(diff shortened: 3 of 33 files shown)


jamessral:
For some reason, I'm having difficulty with the `jupyter-extension` errors and the `records` errors. For the former, I've tried installing the packages (`npm i -S immutable` and `npm --save-dev lodash-webpack-plugin`), but I still seem to get the errors. For the latter, I'm not sure really what to do about that error since it seems to be referencing itself as a package.

applications/commuter/package.json
```diff
@@ -63,6 +63,7 @@
     "elasticsearch": "^15.0.0",
     "express": "^4.15.4",
     "fs-extra": "^7.0.0",
+    "immutable": "^3.8.2",
```
lgeiger, line 66: This looks like you just installed it with `npm` rather than copying it to from another `package.json`. We're requiring version `"immutable": "^4.0.0-rc.9"` instead of version 3. This needs to be changed in a few other places too.

applications/desktop/package.json
```diff
@@ -108,12 +108,38 @@
...
+    "redux": "^4.0.1",
+    "redux-electron-store": "^0.6.2",
+    "redux-logger": "^3.0.6",
+    "redux-observable": "^1.0.0",
+    "rxjs": "^6.3.3",
+    "shell-env": "^2.1.0",
+    "spawn-rx": "^3.0.0",
+    "spawnteract": "^5.1.0",
+    "webfontloader": "^1.6.28",
+    "yargs": "^12.0.2"
```
lgeiger, line 136: These should all be dev dependencies since we're using webpack to bundle the app. Otherwise they would be included in the electron bundle two times. We're only requiring the ones we explicitly exclude from the webpack bundle as dependencies.

packages/connected-components/package-lock.json
```diff
@@ -0,0 +1,597 @@
+{
```
lgeiger, line 1: We're using `yarn` to install our dependencies. Thus there should be no `package-lock.json` files in the repo.

packages/core/package.json
```diff
@@ -21,6 +21,7 @@
     "babel-runtime": "^6.26.0",
     "escape-carriage": "^1.2.0",
     "file-saver": "^1.3.3",
+    "lodash": "^4.17.11",
```
lgeiger, line 24: Nice

packages/markdown/package.json
```diff
@@ -28,6 +28,16 @@
...
     "@nteract/mathjax": "^2.1.4",
     "babel-runtime": "^6.26.0",
-    "react-markdown": "^4.0.0"
+    "react": "^16.3.2",
+    "react-markdown": "^4.0.0",
+    "remark-parse": "^5.0.0",
+    "remark-stringify": "^5.0.0",
+    "trim-trailing-lines": "^1.1.1",
+    "unified": "^7.0.0",
+    "unist-builder": "^1.0.3"
```
lgeiger, line 37: Oh wow so many missing dependencies 👀

packages/notebook-render/package.json
```diff
@@ -26,6 +26,7 @@
     "@nteract/transforms": "^4.4.4",
     "babel-runtime": "^6.26.0",
     "katex": "^0.10.0-alpha",
+    "react-dom": "^16.3.2",
```
lgeiger, line 29: This should be a `devDependency` since it's required in `__test__` only.

packages/outputs/package-lock.json
```diff
@@ -0,0 +1,766 @@
+{
```
lgeiger, line 1: We're using `yarn` to install our dependencies. Thus there should be no `package-lock.json` files in the repo.

packages/records/package.json
```diff
@@ -20,7 +20,7 @@
     "escape-carriage": "^1.2.0",
     "immer": "^1.5.0",
     "lodash": "^4.17.4",
-    "uuid": "^3.1.0"
+    "uuid": "^3.3.2"
```
lgeiger, line 23: We require `^3.1.0` every where else in the repo. It's better to keep the in sync.

packages/styled-blueprintjsx/package-lock.json
```diff
@@ -0,0 +1,164 @@
+{
```
lgeiger, line 1: We're using `yarn` to install our dependencies. Thus there should be no `package-lock.json` files in the repo.

packages/styleguide-components/package-lock.json
```diff
@@ -0,0 +1,137 @@
+{
```
lgeiger, line 1: We're using `yarn` to install our dependencies. Thus there should be no `package-lock.json` files in the repo.

lgeiger (changes requested):
Thanks a lot for tackling this issue 🎉 

As mentioned in the relevant issue, a convention we currently follow is to require `react`, `react-dom` and `immutable` as `peerDependencies` rather than `dependencies`. There are a few places where you added them as direct dependencies. You probably followed the prompt to install them via `npm` rather than copying from another `package.json`, I should've clarified this in the issue.

lgeiger:
> For some reason, I'm having difficulty with the jupyter-extension errors and the records errors. For the former, I've tried installing the packages (npm i -S immutable and npm --save-dev lodash-webpack-plugin), but I still seem to get the errors

Probably the easiest thing would be to just add the dependency to the `package.json` manually. I'm not surprised that installing via `npm` this doesn't work since we're using [`yarn`](https://yarnpkg.com) to manage our dependencies. Sorry, I should've mentioned this earlier.

> For the latter, I'm not sure really what to do about that error since it seems to be referencing itself as a package.

`@nteract/records` requires itself inside the tests. The best thing here would probably to switch it to relative imports as we do in the other tests.

rgbkrk:
Within our `yarn` workspace setup the easiest way to add deps is using [`yarn workspace @nteract/packageName add somePackage`](https://yarnpkg.com/en/docs/cli/workspace).

This functions the same as `yarn add` with workspaces in mind so you can include flags like `--peer` and `--dev`.

lgeiger:
This makes me think if we should start moving all the dev dependencies to their respective packages 🤔 

Installing everything to the root `packages.json` has caused some confusion in recent PRs and we would now have a ESLINT rule to enforce  the correct dev dependencies in our packages.

@rgbkrk What do you think?

rgbkrk:
There has certainly been ask for us to have `devDependency` on the packages so that people can hack on them when they're trying to work on a separate code base. I think the `yarn` workspace stuff is more efficient with this now, we're no longer having to worry about `lerna` install a package N times then reduce it to 1 at the root.

jamessral:
Do you have a preferred way for me to do it, then? I can do it either way

rgbkrk:
Seems like the verdict is -- include the right `devDependency` for each package. It would be ok if you started new PRs that did it package by package. Right now there are quite a few conflicts that may not be fun to work though. 

You'll want to use:

```
yarn workspace @nteract/packageName add somePackage --dev
```

rgbkrk:
Hey thanks for giving this a try, sorry these didn't make it through in due time.

jamessral:
No worries. Everyone gets busy :smile: Perhaps I'll have time to try again some time. I only closed these because they should probably be done again with more recent versions

PR #3474, opened 2018-10-21
Enforce react/jsx-no-bind rule

### LINKED ISSUE #3435: No .bind() or Arrow Functions in JSX Props
We should remove the use of `.bind()` and arrow functions in JSX props:
> A bind call or arrow function in a JSX prop will create a brand new function on every single render. This is bad for performance, as it will result in the garbage collector being invoked way more than is necessary. It may also cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.

This can be easily enforced by adding the [`react/jsx-no-bind` rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) to our ESLINT config.

Arrow functions inside props can be avoided by using [class properties](https://reactjs.org/docs/faq-functions.html#class-properties-stage-3-proposal).

Arrow functions for assigning `refs` can be replaced by using the new [`React.createRef` API](https://reactjs.org/docs/react-api.html#reactcreateref).

These are some small improvements we can make to our components, in order to improve performance in some cases. See #3434

showmeyourhits:

Fixes #3435


applications/commuter/components/contents/html.js
```diff
@@ -8,6 +8,10 @@ export default class HTMLView extends React.Component<*> {
     return false;
   }
 
+  setIframeRef = f => {
```
lgeiger: Any reason why not using the [`React.createRef` API](https://reactjs.org/docs/react-api.html#reactcreateref)?

packages/transforms/src/html.js
```diff
@@ -20,42 +20,46 @@ export function createFragment(html: string): Node {
 }
 
 export default class HTMLDisplay extends React.Component<Props> {
-  el: ?HTMLElement;
+  elRef: React.Ref<HTMLElement>;
```
lgeiger: I guess this should be `elRef: React.ElementRef<*>;` to appease flow.

packages/transforms/src/html.js
```diff
@@ -20,42 +20,46 @@ export function createFragment(html: string): Node {
...
   static MIMETYPE = "text/html";
 
+  constructor(props) {
+    super(props);
+
+    this.elRef = React.createRef();
+  }
+
```
lgeiger, line 26: I think this can benefit from some cleanup. Instead of using a constructor:
```js
export default class HTMLDisplay extends React.Component<Props> {
  elRef: elRef: React.ElementRef<*>;
  constructor(props: Props) {
    super(props);
    this.elRef = React.createRef();
  }
}
 ```
This can be simplified to:
```js
export default class HTMLDisplay extends React.Component<Props> {
  elRef: React.ElementRef<*> = React.createRef();
}
 ```

lgeiger (changes requested):
@showmeyourhits Thanks a lot for the great work 🎉 . This is quite a huge PR now.

We're using [flow](https://flow.org/) to typecheck our code. Unfortunately your changes introduced [some type errors](https://circleci.com/gh/nteract/nteract/3255?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), it would be wonderful if you could fix them. You can typecheck the code by running `yarn flow` which also integrates well with a lot of text editors or IDEs.

If you need help, please join our slack at http://slack.nteract.io. I'm also happy to pair up with you on video to tackle those errors.

lgeiger (approved):
@showmeyourhits Thanks a lot for this work. This is getting quite a huge PR now 👀

Looks like there's only one file left: https://circleci.com/gh/nteract/nteract/3408?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@emeeks This touches quite a bit of code from the data explorer. Could you take a quick look?

rgbkrk:
Closing as we're using tslint and adhering to the rules along with our typescript config.

PR #3475, opened 2018-10-21
Some easy fixes to reduce the creation of new elements during render

lgeiger:
Let's tackle a few easy fixes to reduce warnings introduced in #3440

### DIFF
--- a/applications/play/components/Main.js
+++ b/applications/play/components/Main.js
@@ -17,6 +17,8 @@ import { BinderConsole } from "./consoles";
 const NTERACT_LOGO_URL =
   "https://media.githubusercontent.com/media/nteract/logos/master/nteract_logo_cube_book/exports/images/svg/nteract_logo_wide_purple_inverted.svg";
 
+const emptyList = [];
+
 class Main extends React.Component<*, *> {
   constructor(props) {
     super(props);
@@ -175,7 +177,7 @@ class Main extends React.Component<*, *> {
             logs={
               currentServer && currentServer.messages
                 ? currentServer.messages
-                : []
+                : emptyList
             }
             repo={repoValue}
             gitref={gitrefValue}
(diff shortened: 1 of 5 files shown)


applications/play/components/Main.js
```diff
@@ -17,6 +17,8 @@ import { BinderConsole } from "./consoles";
 const NTERACT_LOGO_URL =
   "https://media.githubusercontent.com/media/nteract/logos/master/nteract_logo_cube_book/exports/images/svg/nteract_logo_wide_purple_inverted.svg";
 
+const emptyList = [];
```
rgbkrk, line 20: I'd probably `Object.freeze` this as well.

packages/notebook-app-component/src/notebook-app.js
```diff
@@ -60,6 +58,30 @@ type AnyCellProps = {
   focusBelowCell: () => void
 };
 
+const markdownEditorOptions = {
```
rgbkrk, line 61: Good plan. It's probably a good idea to freeze this too.

packages/presentational-components/src/components/header-editor.js
```diff
@@ -111,7 +117,7 @@ export class HeaderEditor extends React.Component<
               </Tag>
             ))}
             {(this.state.editMode === "author" && (
-              <Tag style={{ ...authorStyle, color: "black" }}>
+              <Tag style={authorStyleBlack}>
```
rgbkrk, line 120: The [`jsx-no-new-object-as-prop` rule shouldn't be triggering on `style`](https://github.com/cvazac/eslint-plugin-react-perf/blob/master/docs/rules/jsx-no-new-object-as-prop.md#rule-details).

lgeiger: Hmm, maybe it triggers because we're styling a component, instead of an DOM element 🤔

Issue #3434, opened 2018-10-12
Reduce unnecessary rerenders of editor component in notebook


lgeiger:
Currently all code editors are rerendered on every key press or focus change:
![perf](https://user-images.githubusercontent.com/13285808/46870764-d5c45a00-ce2f-11e8-94db-977740644680.gif)

I think there are a few things that may be the culprits:
- [mapDispatchToProps](https://github.com/nteract/nteract/blob/master/packages/notebook-app-component/src/editor.js#L41-L58) and [mapDispatchToCellProps](https://github.com/nteract/nteract/blob/master/packages/notebook-app-component/src/notebook-app.js#L115-L130) creates a new function on every prop change since it relies on the [ownProps](https://github.com/reduxjs/react-redux/blob/master/docs/api.md#arguments)
- [kernelStatus](https://github.com/nteract/nteract/blob/master/packages/editor/src/index.js#L44) often changes when requesting a code completion
- [codeMirrorOptions](https://github.com/nteract/nteract/blob/master/packages/notebook-app-component/src/notebook-app.js#L193) creates a new object when passing it to the editor props

We should definitely improve this component to improve the responsiveness in larger notebooks.

jameskoch:
(Nevermind, you already found this) I happen to be poking around code completion (#3314) and noted this relevant note:

https://github.com/nteract/nteract/blob/a9719367c974737be3879f76e5cc3d904595d36f/packages/editor/src/index.js#L41-L43

rgbkrk:
This greatly saddens me as I fixed this particular issue for nteract web earlier this year. 😭  Clearly I need to work on implementing tests for perf like this, as the editor is the main component for each cell.

jameskoch:
FWIW, it looks like kernelStatus changes aren't flowing in all that quickly, at least in desktop.

1. Start from new notebook, not-connected-to-a-kernel.
2. Type some characters in the default/first cell. Observe via breakpoint in `keyupEventsSubscriber` that auto-completion is not firing due to kernelStatus === "not connected".
3. Start a kernel via Runtime menu.
4. Observe that auto-completion still not happening, kernelStatus is unchanged.
5. Restart kernel.
6. Observe that auto-completion still not happening, kernelStatus is unchanged.

Proximate cause: [This line](https://github.com/nteract/nteract/blob/a9719367c974737be3879f76e5cc3d904595d36f/packages/editor/src/index.js#L192) is reading kernelStatus out of props only once, at mount time.

[...]

PR #3502, opened 2018-10-24
switch to a makeMapStateToProps function in file

rgbkrk:
This intends to replace `ownProps` usage with `makeMapStateToProps`.

I've started on #3434 via the approach outlined in https://github.com/nteract/nteract/issues/3434#issuecomment-432724332. I'm running into a few issues though. Flow is expecting an object instead of a function returned (do we have a global type override set somewhere...?):

screen shot 2018-10-24 at 9 48 26 am

The notebook doesn't show on load immediately (pure components blocking updates or something related to how I'm wielding `connect()` here?):

screen shot 2018-10-24 at 9 49 07 am

Due to hot reload though if I make a change it will force a render:

screen shot 2018-10-24 at 9 50 07 am

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
@@ -43,7 +43,7 @@ type FileProps = {
   error: ?{}
 };
 
-export class File extends React.PureComponent<FileProps, *> {
+export class File extends React.PureComponent<FileProps, null> {
   render() {
     // Determine the file handler
     let choice = null;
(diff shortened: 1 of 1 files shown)


rgbkrk:
Note: we're on the latest definition of the flow type for react-redux. It's missing the `connect(function, function)` form though.

According to https://github.com/flow-typed/flow-typed/issues/2109, this is fixed in https://github.com/flow-typed/flow-typed/pull/2105. It looks like it won't be mergeable as the maintainers don't want people to have to declare the type for their connect. 😭  That's something we can do though!

lgeiger:
Very cool. The same mechanism is available for `mapDispatchToProp` too.

rgbkrk:
Since I was digging around for a separate PR, I see now the most impactful PR might be to do this switch over the `mapStateToCellProps` instead (for Cells).

rgbkrk:
We can now start doing this all over the place now that we're on typescript which handles this case _perfectly_ with types. Glad to be out of the perpetual flow muck around this stuff. We can now optimize like crazy. Suppose I could make a little tutorial too.

PR #3504, opened 2018-10-24
move logging middleware into nteract/core

rgbkrk:
This moves the logging middleware that was only in desktop to be part of core so we can use it in nteract web/jext (and elsewhere).


packages/core/src/middlewares.js
```diff
@@ -49,3 +52,18 @@ export const errorMiddleware = (
   }
   return next(action);
 };
+
+export function clogger() {
```
jameskoch: I might have renamed while in here because it's not obvious to me what this does w/o reading the impl. Or, I just don't get plumbing jokes.

rgbkrk: It was me trying not to trip on the same `logger` variable name. I'll switch these around as I was thinking the same as I made this PR.

PR #3506, opened 2018-10-24
ignore the gatsby transformer built output

rgbkrk:
When the gatsby transformer is built, it doesn't write out to `lib` (not sure if this has to be this way, but alas). The generated files don't match our linting setup so I've set them up to be ignored by eslint.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,3 +11,6 @@ applications/**/node_modules
 coverage
 node_modules/
 */**/__snapshots__/
+
+# the gatsby transformer writes output to the root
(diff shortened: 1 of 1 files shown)


lgeiger (approved):
Does prettier also recognize this or do wee need to add it to `.prettierignore` too?

rgbkrk:
> Does prettier also recognize this or do wee need to add it to .prettierignore too?

Since it's not version controlled (`.gitignore` inside `gatsby-transformer-ipynb`), I didn't notice.

PR #3507, opened 2018-10-24
Throw a clear error when opening a notebook whose kernel cannot be located

### LINKED ISSUE #3500: nteract v0.12.1 halts and doesn't able to close when open notebook with not available kernel
nteract v0.12.1 halts and doesn't able to close window when open notebook with not available kernel. nteract v0.11.9 works well -- simply gives an error that he cannot.

* Windows 7 x64
* nteract v0.12.1

jameskoch:
Fixes #3500.

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.js
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.js
@@ -159,18 +159,25 @@ export const launchKernelByNameEpic = (
     }),
     mergeMap((action: actionTypes.LaunchKernelByNameAction) =>
       kernelSpecsObservable.pipe(
-        mergeMap(specs =>
-          // Defer to a launchKernel action to _actually_ launch
-          of(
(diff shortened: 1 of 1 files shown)


applications/desktop/src/notebook/epics/zeromq-kernels.js
```diff
@@ -159,18 +159,25 @@ export const launchKernelByNameEpic = (
...
-            })
-          )
-        )
+            });
+          } else {
+            return actions.launchKernelFailed({
+              error: new Error(`Kernel named ${action.payload.kernelSpecName} does not appear to be available.`),
+              kernelRef: action.payload.kernelRef,
+              contentRef: action.payload.contentRef
+            });
```
rgbkrk, line 178: Yeah this is much clearer, can't believe we didn't have this check before.

rgbkrk (approved):

PR #3508, opened 2018-10-24
Correctly use MathJax Queue to render math

lgeiger:
Calling `MathJax.Hub.Typeset` executes the command right away and doesn't make use of the MathJax Queue wich leads to MathJax errors or broken rendering when opening a new notebook containing math.

It would be great to make a bugfix release for the `@nteract/mathjax` package so we can start using it in Hydrogen.

This largely reverts #3431 and properly fixes the root cause of the problem.

### DIFF
--- a/packages/core/src/epics/contents.js
+++ b/packages/core/src/epics/contents.js
@@ -152,7 +152,7 @@ export function autoSaveCurrentContentEpic(
 
       // document.hidden appears well supported
       if (typeof document.hidden !== "undefined") {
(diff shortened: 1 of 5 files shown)


rgbkrk:
Ok great, thank you.

rgbkrk:
Happy to go ahead with just shipping packages now and a wider release tonight.

rgbkrk:
Packages now on the way, desktop will happen tonight.

PR #3510, opened 2018-10-24
Add markdown docs

stormpython:
Adding markdown docs.

### DIFF
--- a/packages/outputs/src/components/media/markdown.js
+++ b/packages/outputs/src/components/media/markdown.js
@@ -3,11 +3,11 @@ import * as React from "react";
 import MarkdownComponent from "@nteract/markdown";
 
 type Props = {
-  /*
+  /**
    * Markdown string.
    */
   data: string,
-  /*
+  /**
    * Mime type. Defaults to `text/markdown`.
    */
   mediaType: string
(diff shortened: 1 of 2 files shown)


rgbkrk:
It looks like there's a merge conflict, you'll need to rebase.

stormpython:
@rgbkrk I rebased against master and now all checks have passed. Should I add comments for the descriptions? I noticed that they were missing. I can add them quickly and push the new changes before merging.

packages/outputs/src/components/media/markdown.js
```diff
@@ -3,7 +3,13 @@ import * as React from "react";
 import MarkdownComponent from "@nteract/markdown";
 
 type Props = {
+  /**
+   * Markdown string.
+   */
   data: string,
+  /**
+   * Mime type. Defaults to `text/markdown`.
```
rgbkrk, line 11: This should be media type, we're switching off the mime type naming ([same as IANA](https://www.iana.org/assignments/media-types/media-types.xhtml)).

rgbkrk:
Looks like you got those comments in for the descriptions, those are definitely helpful since they render for users.

packages/outputs/src/components/media/markdown.js
stormpython: Ok. I will make changes to the descriptions and resubmit in another PR.

PR #3511, opened 2018-10-24
Adjust tooltip

emeeks:
* Adjust Height by removing fixed height for data explorer cells so height is based on total child height
* Adjust tooltip CSS so that tooltips don't go outside the cell
* Adjust bin legends so hexbins and heatmaps tell you the number of items rather than an obscure percent

### DIFF
--- a/packages/transform-dataresource/src/ParallelCoordinatesController.js
+++ b/packages/transform-dataresource/src/ParallelCoordinatesController.js
@@ -252,19 +252,21 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
               : null
           }
           pieceHoverAnnotation={!filterMode}
-          tooltipContent={d => (
-            <TooltipContent>
-              <h3>{primaryKey.map(key => d[key]).join(", ")}</h3>
-              {d[dim1] && (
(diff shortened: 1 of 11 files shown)


emeeks:
Also cleans up axes for the xy plots by hiding them for heatmap and suppressing baselines for contour and hexbin.

packages/transform-dataresource/src/ParallelCoordinatesController.js
```diff
@@ -252,19 +252,21 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
...
-                <h3 style={{ color: colorHash[d[dim1]] }}>
-                  {dim1}: {d[dim1]}
-                </h3>
-              )}
-              <p>
-                {d.metric}: {d.rawvalue}
-              </p>
-            </TooltipContent>
-          )}
+          tooltipContent={d => {
```
rgbkrk, line 255: A bit of feedback I've gotten from folks that I should pass on -- it would be fantastic if this variable was named. `datum` is probably fine, open to something better that will make this easier for other contributors to dive into. Just enough that people can figure out the context.

Issue #3497, opened 2018-10-23
Fix production build of components docs


lgeiger:
The production build of our component docs fails on some components with `TypeError: JSON.stringify is not a function`: https://3325-37496521-gh.circle-artifacts.com/0/styleguide/index.html

The dev build is fine though.

PR #3514, opened 2018-10-25
Correct errors for JSON in styleguidist

willingc:
Addresses #3497 

- Add @component comment [Ref](https://react-styleguidist.js.org/docs/thirdparties.html#styled-components)

### DIFF
--- a/packages/outputs/__tests__/__snapshots__/media-spec.js.snap
+++ b/packages/outputs/__tests__/__snapshots__/media-spec.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`<Json />; renders JSON data 1`] = `
+exports[`<Json /> renders JSON data 1`] = `
 <ul
   style={
     Object {
@@ -495,7 +495,7 @@ exports[`<Json />; renders JSON data 1`] = `
 </ul>
 `;
 
-exports[`<Json />; updates the theme if it changes 1`] = `
+exports[`<Json /> updates the theme if it changes 1`] = `
 <ul
   style={
     Object {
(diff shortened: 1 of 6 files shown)


lgeiger:
Well I think I know whats going on here:
It looks like the `JSON` component conflicts with the global `JSON` variable in the bundle.

I'm 99% confident that renaming our `JSON` component to `Json` will fix this problem.

rgbkrk:
Ah yeah the old way I got around this was calling the component `JSONDisplay` or something like that.

willingc:
Thanks @lgeiger and @rgbkrk. I've renamed to `Json` to avoid the global conflict and all seems well now.

rgbkrk:
Thank you so much for figuring this one out, I was stumped.

rgbkrk:
Guess we can deploy the components site after the next release. 😄

lgeiger:
> Guess we can deploy the components site after the next release.

Awesome! Can we automate this somehow, to deploy with now automatically when there's a new tag?

rgbkrk:
I tried that a bit ago and the electron deps and other gunk in the monorepo made it practically impossible.

lgeiger:
> I tried that a bit ago and the electron deps and other gunk in the monorepo made it practically impossible.

Can't we do this on circle CI since we already build the static assets there?

rgbkrk:
Yeah! We can do it with a separate Circle CI process -- that's actually why I did the artifact publishing. We just can't use zeit's integration to do it. If we wanted to use zeit directly, we'd have to make another repo for it. 😬

rgbkrk:
For now I've published the packages and am now publishing the site.

rgbkrk:
Looks good, I'll promote this: https://nteract-components-dwhmkfttcy.now.sh/#json

rgbkrk:
https://components.nteract.io/ is now live

willingc:
Thanks @lgeiger and @rgbkrk. I'm going to open an issue in the docs repo and/or project to keep visibility for auto deploy of docs on the radar (even if it's not elegant now).

cc/ @captainsafia

PR #3520, opened 2018-10-25
Adds Data Explorer docs

stormpython:
Initial attempt at adding docs for the Data Explorer.

@rgbkrk would you mind taking a look at this? @emeeks and I were trying to add this to the docs, but we were encountering the following error and couldn't figure out the cause.

screen shot 2018-10-25 at 12 35 48 pm

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -13,6 +13,10 @@ module.exports = {
   defaultExample: false,
   sections: [
     {
+      name: "Data Explorer",
+      components: "packages/transform-dataresouce/src/*.js"
+    },
+    {
       name: "Presentational Components",
       components: "packages/presentational-components/src/components/*.js"
     },


rgbkrk:
That error in the console is ok -- it's actually showing the `componentDidCatch` of the `<RichMedia />` component. Perhaps what we should do instead is showcase the error component and not demonstrate the componentDidCatch on our components page.

rgbkrk:
I'm going to wrap up my current PR for the error output and then I'll take this for a spin.

styleguide.config.js
```diff
@@ -10,6 +10,10 @@ module.exports = {
   title: "nteract components",
   defaultExample: false,
   sections: [
+    {
+      name: "Data Explorer",
+      components: "packages/transform-dataresouce/src/*.js"
```
rgbkrk, line 17: looks like a little typo

```suggestion
      components: "packages/transform-dataresource/src/*.js"
```

rgbkrk:
Go ahead and rebase this one when you get a chance, the errors from earlier are gone now.

styleguide.config.js
stormpython: 🤦‍♂️

stormpython:
Will do.

rgbkrk:
It looks like, after the typo is fixed, the data explorer will need to be broken up into a `components` directory that matches the style of the other packages (one component per file, at least if we want to document them).

rgbkrk:
From Slack:

@emeeks: Kyle and I were talking data explorer docs.

@rgbkrk: My suggestion was to make data explorer have a `src/components` folder. Stick each component in their own file. We can always make just the one `DataExplorer` export for now
`data-explorer.js` and `data-explorer.md` then the `index.js` can reference that component as the default (in the way it does now).

Ideally we'd go ahead and document the other components, so long as we think we want to support those.

@emeeks: I was wondering if there was more to break out, but I don't imagine that's the case. The color picking legend or the individual charts aren't really going to be useful to external projects.

@rgbkrk: Yeah, let's do less. If there's demand (and we want to support it) we can go for it

@emeeks: Sounds good.

rgbkrk:
Lets do this as a new PR and likely as a new package we can just call `data-explorer`. The `transform-dataresource` package can rely on it, and data explorer will be a suite of components that we start exposing in the components guide.

PR #3521, opened 2018-10-25
Create <Media.Image />

emeeks:
Creates an `Image` component with embedded style


packages/presentational-components/src/components/outputs.js
```diff
@@ -62,11 +62,6 @@ export class Outputs extends React.Component<OutputsProps> {
               word-wrap: break-word;
             }
 
-            .outputs :global(img) {
-              display: block;
-              max-width: 100%;
-            }
```
rgbkrk: We can't remove this yet, as we need the current app to work until we switch over to this component.

packages/outputs/src/components/media/image.md
```diff
@@ -0,0 +1,15 @@
...
+<Image data={matplotlibImage} mediaType="image/png" />
+```
+
+You can change the size of the image by changing the height and width in metadata.
+
+```
+const { matplotlibImage } = require("../../../sampleData/sampleImage");
+
+<Image data={matplotlibImage} mediaType="image/png" metadata={{ height: "200px", width: "800px"}} />
+```
```
rgbkrk, line 21: Within the `<RichMedia />` component let's document how we'd set up the 3 component handlers for the two other image types:

```jsx
<RichMedia>
  <Image mediaType="image/png" />
  <Image mediaType="image/jpeg" />
  <Image mediaType="image/gif" />
</RichMedia>
```

rgbkrk:
I'll go ahead and fix up the snapshots and add in some more documentation. We're getting closer on https://github.com/nteract/nteract/issues/3342. I'll update that issue with work left to do.

packages/outputs/src/components/media/image.md
rgbkrk: I think it's ok for us to handle this in a future PR to further en-_rich_ the overall Media portion of the docs.

captainsafia:
I think this is good to merge as is. I can take a stab at adding even more content to the docs later tonight.

Merging so we can move forward!

PR #3522, opened 2018-10-25
include styles for errors inside the component

rgbkrk:
In pursuit of https://github.com/nteract/nteract/issues/3342

I noticed that the error component on our components docs doesn't include the default styling. This sticks it in and includes a _real_ kernel error.

screen shot 2018-10-25 at 1 43 25 pm

We're probably going to need to do the same bundling of style for `Media.HTML`, `Media.Image`, and 
possibly others. Possible issue though -- Hydrogen may want to override these styles but use the base component.

One approach that @emeeks and I talked about is passing in a `styled-jsx` stylesheet to override the default one.

```
<KernelOutputError styledJsx={css``} />
```

Thoughts @lgeiger @BenRussert?

### DIFF
--- a/packages/outputs/src/components/kernel-output-error.js
+++ b/packages/outputs/src/components/kernel-output-error.js
@@ -46,9 +46,7 @@ export const KernelOutputError = (props: Props) => {
         <Ansi>{kernelOutputError.join("\n")}</Ansi>
         <style jsx>{`
           .kernel-output-error :global(code) {
-            font-family: "Source Code Pro";
             white-space: pre-wrap;
-            font-size: 14px;
           }
         `}</style>
       </div>
--- a/packages/outputs/src/components/kernel-output-error.js
+++ b/packages/outputs/src/components/kernel-output-error.js
@@ -26,11 +26,34 @@ type Props = {
 export const KernelOutputError = (props: Props) => {
   const { ename, evalue, traceback } = props;
 
-  if (!Array.isArray(traceback) || !traceback.length) {
-    return <Ansi>{`${ename}: ${evalue}`}</Ansi>;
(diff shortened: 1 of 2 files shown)


rgbkrk:
Actually I'll go ahead and remove the font styling and only include the `white-space: pre-wrap` style.

jdetle:
How do I review this?

rgbkrk:
> How do I review this?

* Does this match current behavior with our desktop app and other frontends?
* What would I need as a frontend developer to understand this component / output type (for the markdown file)?


For the hydrogen style concern:

* What level of included styles should components like these bake in?
* How easy is it to override this type of output if I want "my own thing"? (Answer: you can pass whatever outputType component you want to handle error output directly into the [output picker](https://github.com/nteract/nteract/blob/master/packages/outputs/src/components/output.js))

lgeiger:
I think things like `white-space` are really useful to have backed into the components.

For Hydrogen I'd love to have more styling embedded in the components. Our global css approach in Hydrogen is a bit sketchy at the moment. The only place where we're forced to rely on global css is for colors, since we should match the global theme in Atom. I don't think Atom has the possibility to access the theme options from anywhere other than the global css style sheet. @BenRussert Am I missing something important here?

rgbkrk:
I'd love for our layer between that we use common CSS variables and you set these from less. Basically like a LESS -> nteract CSS variable translation layer. That'll help keep things going in lock step with the data explorer and any other visual components.

rgbkrk:
Rebased since I merged #3528

captainsafia (approved):
Looks good. Thanks for adding the clarifying point about some kernels not sending back stacktraces. I started writing that but wasn't confident if it was true.

PR #3524, opened 2018-10-25
Modernize ansi-to-react and try out React.memo

lgeiger:
Currently blocked by https://github.com/airbnb/enzyme/issues/1875 and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29990


jdetle:
This is helping prevent large rerenderings?

rgbkrk:
Might want to try this as a new PR, deps should be up to date by now and it's reasonable to rely on `React.memo` now.

PR #3525, opened 2018-10-25
Typescript: Convert @nteract/display-area

lgeiger:


packages/display-area/package.json
```diff
@@ -24,10 +23,8 @@
     "access": "public"
   },
   "dependencies": {
-    "@babel/runtime-corejs2": "^7.0.0",
     "@nteract/transforms": "^4.4.7",
     "ansi-to-react": "^3.3.5",
-    "babel-runtime": "^6.26.0",
```
lgeiger, line 30: We don't need those anymore since we're using typescript to build the package.

packages/display-area/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "lib",
+    "noImplicitAny": false
```
lgeiger, line 6: Since `@nteract/transforms` isn't converted to typescript we need to allow implicit `any`.

lgeiger (commented):
That actually went very smooth 🎉

packages/display-area/tsconfig.json
jdetle: makes sense, thanks

lgeiger:
Actually setting up typescript in our monorepo isn't that easy: https://github.com/Microsoft/TypeScript/issues/25376

jdetle:
~~oh man... what an unfortunate turn to this friday~~	

Scratch that! You got it 🙌

packages/display-area/tsconfig.json
jdfreder: We may want to prioritize converting `@nteract/transforms`, because I don't think it's a good idea to turn off `noImplicitAny` for an extended period of time - it makes turning it back on difficult.

PR #3531, opened 2018-10-27
Convert monaco-editor to TypeScript

charlesjolley:
In addition, fixes a few bugs surfaced by TS typechecking. Namely:

  - monaco-editor import was broken because the current package has
    the same name and TypeScript is setup to resolve with the local
    packages first. Fixed by amending the tsconfig for this package
    to ignore other local packages. You should still be able to import
    other packages by referring them as @nteract/foo

  - componentDidUpdate() was calling the wrong method to change the
    language and theme, which means it wasn't working before at all.
    I actually added the correct calls, but maybe better to consider
    deleting these altogether since these weren't have an effect before
    and no one seemed to think it was a bug.

  - converted the ref assignment to use a saved method on the class.
    this is more efficient and easier to do in TypeScript with
    assignment functions.

  - removed unnecessary eslint-disable command

  - removed $FlowFixMe calls which seemed to be unnecessary


packages/monaco-editor/package.json
```diff
@@ -30,6 +30,7 @@
     "@nteract/display-area": "^4.4.8",
     "@nteract/messaging": "^4.2.2",
     "@nteract/transforms": "^4.4.7",
+    "@types/lodash": "^4.14.117",
```
captainsafia: I believe this type can be moved over to `devDependencies`.

packages/monaco-editor/src/index.tsx
```diff
@@ -1,47 +1,44 @@
...
+  onChange: (value: string) => void;
+  value: string;
+  editorFocused: boolean;
+}
 
 class MonacoEditor extends React.Component<MonacoEditorProps> {
-  monaco: ?monaco.IStandaloneCodeEditor;
-  monacoContainer: ?HTMLElement;
+  monaco?: editor.IStandaloneCodeEditor;
+  monacoContainer?: HTMLElement | null;
```
captainsafia: Hm. Any reason we have the `| null` here if the previous type was a possibly-undefined HTMLElement?

packages/monaco-editor/src/index.tsx
```diff
@@ -60,22 +57,24 @@ class MonacoEditor extends React.Component<MonacoEditorProps> {
...
-      // $FlowFixMe: We should be detecting monaco above
       this.monaco.setValue(this.props.value);
     }
 
-    // $FlowFixMe: We should be detecting monaco above
-    this.monaco.updateOptions({
-      language: this.props.mode,
-      theme: this.props.theme
-    });
+    const model = this.monaco.getModel();
```
captainsafia, line 66: Ooooh. Thanks for catching this. I think we should still keep the logic in the codebase. I suspect it was filed as a bug because people didn't recognize something was off right away.

captainsafia (approved):
Great! I left a few comments and questions in-line. Once they are addressed, this can be merged!

packages/monaco-editor/package.json
charlesjolley: Sure thing. While I'm at it, It appears that the only dependencies this package actually uses is the "monaco-editor" and "lodash" packages. want me to remove all of the others?

    "@nteract/display-area": "^4.4.8",
    "@nteract/messaging": "^4.2.2",
    "@nteract/transforms": "^4.4.7",
    "babel-runtime": "^6.26.0",
    "monaco-editor-webpack-plugin": "^1.5.1",
    "rxjs": "^6.3.3"

packages/monaco-editor/src/index.tsx
charlesjolley: TypeScript was complaining because the value can sometimes be null. Right now we have TS set to a pretty strict mode so it requires you to deal with undefined and null cases independently (or at least acknowledge in the type as I do here).  

We could loosen that constraint a bit on the compiler if you want. I'd be for it.

packages/monaco-editor/src/index.tsx
charlesjolley: +1

charlesjolley:
OK i just pushed a new commit that:

1- moves @types/lodash to dev dependencies
2- rebases from master
3- added an include line to the tsconfig file as required for new projects [this is has caught me on other packages]

packages/monaco-editor/src/index.tsx
lgeiger: Using the `React.createRef` API here may also help with properly typing this https://github.com/nteract/nteract/pull/3474/files#diff-4de70bc2c4e3ef16f6d02b74a92534dd

packages/monaco-editor/package.json
```diff
@@ -2,17 +2,17 @@
...
     "prepare": "npm run build",
     "prepublishOnly": "npm run build && npm run build:flow",
     "build": "npm run build:clean && npm run build:lib",
     "build:clean": "rimraf lib",
-    "build:flow": "flow-copy-source -v -i '**/__tests__/**' src lib",
-    "build:lib": "babel -d lib src --ignore '**/__tests__/**' --config-file ../../babel.config.js",
+    "build:lib": "tsc",
     "build:lib:watch": "npm run build:lib -- --watch",
-    "build:watch": "npm run build:clean && npm run build:lib:watch && npm run build:flow"
+    "build:watch": "npm run build:clean && npm run build:lib:watch"
```
lgeiger: I think you can remove all of these scripts since we build the typescript packages from the [root `package.json`](https://github.com/nteract/nteract/blob/master/package.json#L36)

packages/monaco-editor/package.json
```diff
@@ -25,6 +25,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "devDependencies": {
+    "@types/lodash": "^4.14.117"
```
lgeiger, line 20: Since we rely on `lodash` in several places you could move this to the top level `package.json`.

lgeiger (commented):
Thanks for tackling the conversion 🎉

packages/monaco-editor/package.json
charlesjolley: Will do. I figured that out eventually for the other pull requests.

packages/monaco-editor/package.json
charlesjolley: I thought the purpose of monorepo was that you managed each dependency separately for the packages and lerna/yarn take care of collapsing the dependencies to a top-level node_modules?

I'll follow whatever pattern you guys do, but it seems to me keeping shared dependencies doesn't get you any benefits but will make it harder for keep track of  which deps are still needed?

charlesjolley:
Pushed all changes based on feedback [except for moving the lodash dependency until I hear back]

charlesjolley:
Merged since i saw the PR was approved. I can submit a new PR to move the lodash dependency if requested.

packages/monaco-editor/package.json
rgbkrk: Now that we're on yarn workspaces I think it's much better for us to have `devDependencies` with each individual package (like you've done here). We were seeing a bit of a speedup in the past to have dependencies like these in the top level so that lerna wasn't going through installing it N times, then reducing it. Nowadays, it's much smarter about creating symlinks for the individual packages.

PR #3532, opened 2018-10-27
Convert Octicons package to TypeScript

charlesjolley:
This package is very simple but I wanted to try to solve the issue of converting a package to TS breaking flow types (and therefore test:flow check.) One solution that was used for ansi-to-react is to just disable typing around every callsite using $FlowFixMe.  Instead, in this pull request I am proposing that we include a .js.flow file alongside the .ts(x) files.

The downside is that this file will need to be maintained until all dependent packages are also converted to TypeScript. The upside however, is that it will centralize all the needed maintenance to one file and allow flow typing to still work everywhere else.


captainsafia (approved):
Neat-o. I think it's fine that we've got a small centralized dependency to maintain here. Hopefully, it won't stick around for too long. I'll make an issue to keep note that we need to track these and clear them out once they are no longer needed.

charlesjolley:
Woot! First commit!

rgbkrk:

PR #3533, opened 2018-10-28
Convert messaging to TypeScript

charlesjolley:
Trying a more complicated package this time. I also added an enum type for MessageType which allows for stricter type checking. (per a TODO that was in the code) and eliminated one of the ambient type definition (channel) since TypeScript & monorepos play better fewer ambient types.


packages/messaging/__tests__/messaging-spec.ts
```diff
@@ -85,17 +86,19 @@ describe("childOf", () => {
...
-      { header: { msg_type: "a" } },
-      { header: { msg_type: "d" } }
-    ])
+    from(([
+      { header: { msg_type: "stream" } },
+      { header: { msg_type: "error" } },
+      { header: { msg_type: "status" } },
+      { header: { msg_type: "stream" } },
+      { header: { msg_type: "status" } }
+    ] as any[]) as JupyterMessage[])
```
captainsafia, line 95: Is there a reason we as casting it as `any[]` then as `JupyterMessage[]` message here and elsewhere?

packages/messaging/src/index.ts
```diff
@@ -60,15 +58,15 @@ export const childOf = (parentMessage: JupyterMessage<*, *>) => (
  * @return {Observable}                 the resulting observable
  */
 export const ofMessageType = (
-  ...messageTypes: Array<string> | [Array<string>]
-) => {
+  ...messageTypes: (string | string[])[]
```
captainsafia: I believe it's better practice (and cleaner) to use the `Array<*>` syntax for complex array types.

captainsafia (commented):
Cool! Left some questions and comments inline.

packages/messaging/src/index.ts
charlesjolley: For TypeScript? tslint default rules strongly prefers the foo[] notation unless you're defining an interface. I can make it whatever you want though.

packages/messaging/__tests__/messaging-spec.ts
charlesjolley: The mocks don't fully satisfy the JupyterMessage interface. We could add all the extra dummy properties to fully satisfy the interface, but is not needed and it would make the code a lot harder to follow, IMO.  Another option would be to relax the JupyterMessage interface but I assume that except for these units tests, we generally want to be strict so that we get the full benefits of typing.  I opted for a third route, which is to just coerce the types in this special testing case and leave the strict types in place for regular code.

You have to coerce the types to 'any' before TS will let you coerce into a more specific type.

packages/messaging/src/index.ts
captainsafia: Mmmm. I guess my bias is showing. In other TypeScript codebases I've worked with, we use the [`array-type: array-simple`](https://palantir.github.io/tslint/rules/array-type/) constraint on tslint. I tend to prefer it because it makes things like this a little bit easier to read when you're skimming through code.

I'm fine leaving it as is and setting our tslint configuration at a later time once everything has been migrated.

charlesjolley: Staring at the original code more, I realized that what this function actually expects is vargs of either strings or 1d tuple (i.e. array with a single string element).  I just pushed a change that makes that will enforce that. TS will complain if you try to call with empty arrays or arrays with > 1 string.

charlesjolley: How about Array<string|[string]>?

charlesjolley: Is that what you had in mind?

charlesjolley:
@captainsafia I just pushed a change that adopts the "Array<string|[string]>" format and also a change to tsconfig to make TS play nice with commonjs modules such as uuid.  [And I rebased from master so it should be a clean merge]

charlesjolley:
This PR is included in #3536 anyway so I'm closing this one.

PR #3536, opened 2018-10-28
Convert enchannel-zmq-backend to TypeScript

charlesjolley:
This pull request depends on #3533 being merged first since it relies on types from messaging.

This pull request converts the enchannel-zmq-backend package to TypeScript. It also fixes a few bugs in the tests and integration test script that caused it not to even work before.


packages/enchannel-zmq-backend/src/index.ts
```diff
@@ -13,53 +13,60 @@ export const ZMQType = {
...
+  shell_port: number;
+  stdin_port: number;
+  control_port: number;
+  // Allows practically any string, they're really constrained though
+  signature_scheme: "hmac-sha256" | string;
+  hb_port: number;
+  ip: string;
+  key: string;
+  // Only known transports at the moment, we'll allow string in general though
+  transport: "tcp" | "ipc" | string;
```
jdetle: I dont know if this is how we want to do things. I would just say "tcp" | "ipc" and "hmac-sha256". We can use type checking to determine whether we need to add to this.

packages/enchannel-zmq-backend/src/jmp.d.ts
```diff
@@ -0,0 +1,48 @@
+// jmp does not publish type definitions. Provide a skeleton here just to
```
jdetle, line 1: I believe this should be moved into the root level `types/` folder.

packages/messaging/src/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  | "comm_open"
+  | "comm_msg"
+  | "comm_close"
+  | "execute_reply";
+
+export interface JupyterMessageHeader<MT extends MessageType = MessageType> {
+  msg_id: string;
+  username: string;
+  date: string; // ISO 8601 timestamp
+  msg_type: MT; // this could be an enum
```
jdetle: It looks like an enum :D

packages/enchannel-zmq-backend/src/index.ts
charlesjolley: so you'd like me to drop the string generic type for signature_scheme and transport? Easy to do but just a note that the flow typings were open ended, so that would be tightening the API.

jdetle: Yeah, our flow typings weren't perfect. I think taking the least-permissive route with enums whenever we can is the way to go.

packages/enchannel-zmq-backend/src/jmp.d.ts
charlesjolley: I'll try to make that work. At present TS doesn't seem to pick up types that I put in there

jdetle: I think you need to add it to the tsconfig.

packages/messaging/src/types.ts
charlesjolley: haha. that comment is left over from before. I'll take it out now that I added the enum

packages/enchannel-zmq-backend/src/jmp.d.ts
charlesjolley: I tried that but it ends up causing some other issues because of how tsconfig projects work. In general with the projects activated, you can't have 'ambient' shared folders across all your TS.

charlesjolley: I have an idea though on how to make this work

packages/enchannel-zmq-backend/src/index.ts
charlesjolley: Changed.

packages/messaging/src/types.ts
charlesjolley: Changes

charlesjolley:
@jdetle I have made all the updates you requested except for moving the jmp.d.ts types. Because of our use of TSC build projects, moving this to a global location (like types) in a scalable/maintainable way probably needs some discussion on its own as to how we want the process to work in the future.

Can you approve this PR as-is and I will submit a separate PR / proposal that will give us a scalable way to handle typings like this in the future?

packages/messaging/src/types.ts
rgbkrk: Ha, I'm the one that put in "could be an enum".

rgbkrk:
Rebase this one and let's bring it on in, keep this typescript train moving.

PR #3537, opened 2018-10-28
Convert commutable to TypeScript

charlesjolley:
-  Translated core files to TypeScript
 -  Added flow-type file to keep flow-based packages happy
 -  Deleted ambient types in base.js. Updated dependencies to either
    import the types from commutable or - in the case or record -
    copied the JSON types in there since the record seems to consider
    itself a replacement.

Note that in general it's a lot easier to transition to TypeScript if
we keep cross package imports to just the default.


packages/commutable/src/index.js.flow
```diff
@@ -0,0 +1,259 @@
+// @flow
+/* eslint-disable */
+
+/**
+ * These types are to be maintained by hand until all dependent packages have
+ * converted to typescript.
+ */
```
rgbkrk, line 7: Thanks for making these available to do the migration.

charlesjolley:
@rgbkrk am I approved to merge this PR? I just pushed a rebased branch; no other changes

charlesjolley:
Trying to get these knocked out before the weekend is over and I get busy IRL. :)

rgbkrk:
Done deal, thank you!

PR #3539, opened 2018-10-28
Update styleguidist docs for Media.JSON

captainsafia:

### DIFF
--- a/packages/outputs/src/components/media/json.md
+++ b/packages/outputs/src/components/media/json.md
@@ -1,5 +1,13 @@
-Renders JSON data.
+JSON, or JavaScript Object Notation, is a popular data format that represents its contents in a human-readable form. JSON consists of a series of attribute value pairs, like `name: "Tester McTester"`.
+
+You can render JSON data into an explorable tree-like structure using the `Media.Json` component. JSON data is identified using the `application/json` media type.
 
 ```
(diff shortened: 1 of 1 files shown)


packages/outputs/src/components/media/json.md
```diff
@@ -1,5 +1,13 @@
...
+
+```
+<Json data={{ name: "Tester McTester", age: 20, location: "The North Pole", job: { title: "Senior Engineer", employer: { name: "Acme Inc.", id: 1 } } }}/>
 ```
-<Json data={{ name: "Tester McTester", age: 20, location: "The North Pole" }}/>
+
+Any nested JSON data will be rendered in expanded form by default. If you would like to render nested data in collapsed form by default, you can pass the object `{ expanded: false }` to the `metadata` prop.
+
 ```
+<Json metadata={{expanded: false}} data={{ name: "Tester McTester", age: 20, location: "The North Pole", job: { title: "Senior Engineer", employer: { name: "Acme Inc.", id: 1 } } }}/>
```
rgbkrk, line 12: An aside from this PR -- I wonder if we should splat the metadata as individual props so that we can write:

```jsx
<Json
    expanded
    data={{}}
  />
```

I've been thinking about this the more we have to deal with these untyped metadata structures and wanting to identify settings quickly. 🤔

rgbkrk:
Thank you Safia!

PR #3540, opened 2018-10-28
Update styleguidist docs for Media.HTML

captainsafia:

Experimenting with explaining our implementation internals for the components. Let me know your thoughts on this!

### DIFF
--- a/packages/outputs/src/components/media/html.js
+++ b/packages/outputs/src/components/media/html.js
@@ -10,7 +10,7 @@ type Props = {
    * The media type associated with the HTML
    * string. This defaults to text/html.
    */
-  mediaType: string
+  mediaType: 'text/html'
 };
 
 // Note: createRange and Range must be polyfilled on older browsers with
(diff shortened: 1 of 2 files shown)


packages/outputs/src/components/media/html.js
```diff
@@ -2,7 +2,14 @@
...
 type Props = {
+  /**
+   * The HTML string that will be rendered.
+   */
   data: string,
+  /**
+   * The media type associated with the HTML
+   * string. This defaults to text/html.
+   */
   mediaType: string
```
jdetle: In the future I think it would be nice to make these mediaType props an enum of different mimetypes.

rgbkrk: For individual components they should likely just take the literal types they accept. For this one, just `text/html`.

```suggestion
  mediaType: 'text/html'
```

rgbkrk: Nice, I just tried the whole "Apply Suggestion" inline thing. Pretty cool.

PR #3542, opened 2018-10-28
Add Media.Javascript component

captainsafia:

Code is the same as the JavaScript transform. While testing, I noticed that the stacktrace can overflow past its parent div a line is too long so I added a wrap. That's the only difference.

### DIFF
--- a/packages/outputs/src/components/media/javascript.js
+++ b/packages/outputs/src/components/media/javascript.js
@@ -22,7 +22,7 @@ export function runCodeHere(el: ?HTMLElement, code: string): any {
     return eval(code); // eslint-disable-line no-eval
   } catch (err) {
     const pre = document.createElement("pre");
-    pre.style = "white-space: pre-wrap";
+    pre.style.whiteSpace = "pre-wrap";
     if (err.stack) {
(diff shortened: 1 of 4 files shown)


packages/outputs/src/components/media/javascript.js
```diff
@@ -0,0 +1,72 @@
+/* @flow */
+import React from "react";
+
+type Props = {
+  /**
+   * The media type associated with our component. This should be one of
+   * application/javascript, application/json, or text/*.
```
rgbkrk, line 7: For this one we can say it only takes a literal (even though it can be changed at runtime) and is fully just `application/javascript`.

packages/outputs/src/components/media/javascript.md
```diff
@@ -0,0 +1,26 @@
+The Media.JavaScript component allows you to execute JavaScript in the context of the current document.
+
```
rgbkrk, line 2: Probably want to stick a warning in about this component, similar to the `Media.Markdown` docs.

rgbkrk:
Going to merge and assume that we can handle the feedback afterward.

PR #3547, opened 2018-10-29
[WIP] Improve and document TypeScript workflow

charlesjolley:
This PR cleans up the typescript build process to eliminate a bunch of edge
cases where things would break, improve build performance, and allows us to
have project-wide typings at the top level.

Consider this PR a proposal since it has some impact on how people actually
work. The code is in a mergable state however.

## Recongiured tsconfigs

The shared config all projects import is renamed to `tsconfig.base.json` and
there is now a new top-level tsconfig that refences the tsconfig in packages
and can reference a tsconfig in applications, once we have at least one app
using tsc.

This way, it is now much easier to invoke tsc, and even select specific
packages to build. You just call `tsc --build` (or equivalent yarn script) from
any folder and it will do the right thing for the scope. You can further
limit scope by naming something, such as `tsc --build packages/ansi-to-react`.

I've also removed/added configs that were causing conflicts with the way tsc
handles imports and project builds

## Renamed / Better Integrated tsc scripts

It turns out that because of how we have our packages configured, all
of our tooling -- except for jest -- actually looks for the built files, not
the original ts source. Thus we need to make sure that essentially tsc builds
are always up-to-date. The good news is that with the tsc build system, that is
very easy and fast.

In support of this, I added/renamed the 'ts' tasks to prefix 'tsc:' and then
added them as precursers or concurrent tasks to run everywhere that you need
to make sure your files are always up to date.

The reason for prefixing tsc instead of postfixing as we did before is because
it's often important to run only these commands and at least when I've been
working I could often mistype and end up causing broken builds.

I'm proposing tsc as the prefix because this setup is really about leveraging
the tsc build mode, which is not actually related to typescript other than you
get at it through the same binary.

## Top-level Typings

With all of that work done, I was able to setup the top-level types directory
so that we can place ambient type declarations there and they will
automatically show up in every package. I moved jmp here and added a readme
explaining how and when to use that directory.

## Docs

To help explain the process, I added a [typescript.md](https://github.com/nteract/nteract/blob/42f7a6cdcf37b4cbcc25c9e40f293702d2c95177/doc/design/typescript.md) in docs/design that we can
use to explain to document how the typescript-based workflow works and also
how to convert packages to typescript. I borrowed some of the text from the
tracking issue.


willingc (commented):
Thanks @charlesjolley. Docs are beautiful. 🎉  I will leave the TypeScript/JS and any workflow changes for others to review.

@charlesjolley what do you recommend for documenting the source? TypeDoc http://typedoc.org/ ?

packages/messaging/src/types.ts
```diff
@@ -53,7 +53,7 @@ export interface ExecuteMessageContent {
   code: string;
   silent: boolean;
   store_history: boolean;
-  user_expressions: Object;
+  user_expressions: object;
```
jdetle, line 56: what is the significance here?

rgbkrk: From [the TypeScript handbook](https://www.typescriptlang.org/docs/handbook/basic-types.html)

> `object` is a type that represents the non-primitive type, i.e. any thing that is not `number`, `string`, `boolean`, `symbol`, `null`, or `undefined`.

I am going to have quite the time adapting from flow to TypeScript. 🙃

applications/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+// tsc built packages
+// Any folders listed here will be built when you run tsc:build:app
+{
+  "compilerOptions": {
+    "composite": true
+  },
+  "files": [],
+  "references": [
+    // { "path": "desktop" }
```
lgeiger: I don't think we need a `tsconfig` in our application folder. To build the applications we use `next.js`/`webpack` which are configured via the `webpack-configurator` package to use the source `.ts` and `.js` files.

package.json
```diff
@@ -20,28 +20,30 @@
...
-    "app:commuter": "lerna run dev --scope @nteract/commuter --stream",
-    "app:jupyter-extension": "lerna run dev --scope nteract-on-jupyter --stream",
+    "app:dev": "npm run tsc:build && npm run tsc:watch & lerna run dev --stream --scope",
+    "app:desktop": "npm run app:dev -- nteract",
+    "app:play": "npm run app:dev -- @nteract/play",
+    "app:commuter": "npm run app:dev -- @nteract/commuter",
+    "app:jupyter-extension": "npm run app:dev -- nteract-on-jupyter",
     "start": "npm run app:desktop",
-    "spawn": "lerna run spawn --scope nteract",
+    "spawn": "npm run tsc:build && lerna run spawn --scope nteract",
```
lgeiger, line 29: Our applications require the source files directly using webpack, we therefor don't need to execute the typescript build.

package.json
```diff
@@ -53,17 +55,18 @@
     "test:unit": "npm run test",
     "test:coverage": "npm run test -- --coverage",
     "test:lint": "npm run lint",
-    "test:flow": "npm run flow",
+    "test:flow": "npm run tsc:build && npm run flow",
     "flow": "flow",
     "diagnostics": "scripts/kernelspecs-diagnostics.js",
-    "verifyBeforePublish": "tsc -b packages && lerna run prepublishOnly && lerna run prepare",
+    "install": "npm run tsc:build",
```
lgeiger, line 61: Do we really need to build all packages after install?

packages/messaging/src/types.ts
charlesjolley: yep TS has gotten more particular about this one over the years. Basically anywhere you use Object in flow, change to 'object'

doc/design/typescript.md
```diff
@@ -0,0 +1,204 @@
...
+
+    yarn tsc:build packages/core  # builds just core and its dependencies
+
+Presently if you run a general command like `yarn build`, it will first invoke
+the top-level `tsc` version of the command and then runs the equivalent command
+– if it exists – in each package or application.
+
+### 2. tsc files are built sepatately
+
+The rest of our tools (webpack, IDE integrations) basically expect to see
```
lgeiger, line 32: For webpack this is not the case. Webpack [is configured](https://github.com/nteract/nteract/blob/master/packages/webpack-configurator/index.js) to build the packages in our monorepo from source which allows us to hack on the applications or docs without a watching the `tsc` build.

charlesjolley:
@willingc I've never setup a separate ts documentation process because vscode already has pretty exceptional support built in. (If you add jsdoc style comments, they appear when you hover over any typed value).  That said, TypeDoc seems to be the big boy in town. I'd keep an eye on https://github.com/Microsoft/tsdoc since that is effectively coming from the TypeScript team

lgeiger (changes requested):
This is awesome! @charlesjolley Thanks for leading the typescript efforts and documenting the setup.

I added some quick comments related to how we build our applications. I'll elaborate later today.

applications/tsconfig.json
charlesjolley: OK. I can remove it.

package.json
charlesjolley: that's not correct. because the package.json's all point to a main file at lib/index.js and types at lib/index.d.ts -- webpack will look for those files and ignores the typescript in src.

Previously this may have been partially obscured by having the paths property set in the base tsconfig.  However, using paths in this way conflicts with composite/build-mode [which requires you to specify all of your files and to keep them under the same root, which is src in each package].

package.json
charlesjolley: Because we are using packages in a monorepo, all of our tooling (webpack, vscode language support, babel) look for types and source in lib/* not src/*.  this means that to even open the project in an IDE without the having it throw errors everywhere, you need to have at least an initial build of your source present. i.e. the first step after installing will always be a clean build anyway.

I think it's worth hooking here so that developers don't immediately hit a speed bump when they first open the project.

doc/design/typescript.md
charlesjolley: See my note above. Our packages say to look for main and types in lib/* not src/* which means when you import via Webpack, it goes there.  Ditto when you're importing from another package via jest.

FWIW - I think this is probably on balance the best behavior because it means that consuming a package local to the monorepo follows the exact same code path as if you installed the package from npm.

rgbkrk: These are all the main fields we currently accept in our webpack config:

```
mainFields: ["nteractDesktop", "jsnext:main", "module", "main"],
```

`nteractDesktop` is more a legacy than anything else, as it's largely meant "our nteract babel configured `src` directory"

package.json
lgeiger: > that's not correct. because the package.json's all point to a main file at lib/index.js and types at lib/index.d.ts -- webpack will look for those files and ignores the typescript in src.

Maybe this behavior has changed in the last few days, I don't have time to verify this right now. But webpack should recognize the [`nteractDesktop` key](https://github.com/nteract/nteract/blob/master/packages/webpack-configurator/index.js#L99) and prefer it over the `main` field. The same *should* be true for `jest` since we use a [`moduleMapping`](https://github.com/nteract/nteract/blob/master/package.json#L79) to enable it to read the source files.

doc/design/typescript.md
charlesjolley: Thanks that's good to know. I missed that when I looked at the configuration. I'll go back and see if I can figure out why that wasn't working before.

charlesjolley: why do we even have the tsc-build process then? It feels like a recipe for ongoing issues to have parallel build infrastructure.

charlesjolley: OK. I did some investigating. There seems to be a confluence of things going here:

1- I'm not sure why but it appears that if you have a `lib` directory in a package, then webpack will prefer to use that code instead of what it finds in src. Once you get into any compiled lib directory, you're in commonjs land using require everywhere and so then it expects all other packages to be compiled also.

2- There isn't really a good way to do a build clean and even when I try to run the build:clean task in each project, it doesn't seem to consistently clean out the old lib directories in non-tsc-based packages.

Thus you end up in a situation where the webpack build is nondeterministic. If you've run a yarn build sometime recently, you can end up with old artifacts included in the build and unexpected errors.

Furthermore, looking at the webpack config it feels like there are a number of overlapping methods in play here to try to resolve various paths in different ways. Further making it hard to reason about how to includes will happen.

To further complicate matters, vscode / tsserver doesn't see any of this custom webpack trickery which means that you _must_ keep at least the built versions of your typescript packages up to date for intellisense to work across packages.  Which gets us back into the original problem of webpack getting confused if you have some packages built.

---

I have a proposal on how to fix all of this. Maybe I should write an RFC as an issue and we can discuss it there?

jdetle: What `tsc-build` process are you referring to specifically? I'm afraid I've lost context here.

charlesjolley: I mean we have this entire setup to use the tsc build mode for packages.  (i.e. yarn build:packages:ts)

charlesjolley:
Replaced by #3559

Issue #3549, opened 2018-10-29
Desktop app crashes on close when run as non-Admin user on Windows


jameskoch:
Desktop 0.12.1 and newer crash during shutdown. Everything seems to shutdown fine, but you get the Windows prompt that something has gone awry, and you see an Access Violation in the Application Event Log:

```
Exception code: 0xc0000005
Fault offset: 0x0000000001870b14
```

Only occurs as non-Admin user.

There's no shortage of [Electron bugs in this space](https://github.com/electron/electron/issues?q=Close+crash).

With Electron symbols loaded, the stack trace looks like this:
```
nteract.exe!aura::Window::GetRootWindow() Line 212   C++
nteract.exe!aura::Window::GetHost() Line 216    C++
nteract.exe!atom::NativeWindowViews::GetAcceleratedWidget() Line 1028          C++
nteract.exe!atom::NativeWindowViews::IsClosable() Line 686      C++
nteract.exe!atom::NativeWindowViews::Close() Line 314               C++
nteract.exe!mate::internal::Dispatcher<void __cdecl(atom::api::Menu * __ptr64)>::DispatchToCallback(const v8::FunctionCallbackInfo<v8::Value> & info) Line 239       C++
node.dll!000007fec5e0de84()     Unknown
node.dll!000007fec5e7f8fd()      Unknown
node.dll!000007fec5e4db89()     Unknown
node.dll!000007fec5e4da6b()     Unknown
0000000c95f84561()        Unknown
```

jameskoch:
Filed https://github.com/electron/electron/issues/15457.  Will look at work-arounds now.

rgbkrk:
One thing we could do is pin to the version of electron that vs-code is on, since we'd hope Microsoft is on top of that. We've tended towards staying up to date with Electron (and all our dependencies) potentially to our own demise.

jameskoch:
#3551 fixed a repro of an Electron issue I was able to create on a random AWS Windows 10 image.

Once that was fixed I moved to a different test environment and now shutdown is crashing w/ a stack that looks like https://github.com/electron/electron/issues/14411 - an issue when an http proxy is in the mix. I can work-around that by launching w/ `--no-proxy-server` to avoid ever initializing the proxy-related stack. I might instead quickly add an env var to disable app auto-updates, since I already saw one user request that in #3517, and I know it can generate some log spew in certain environments without Internet access.

[...]

PR #3551, opened 2018-10-29
Prevent crashing of desktop app when running on Windows as a non-admin

jameskoch:
Works-around #3549, which I would like to leave open so we eventually unwind this to ensure any initiation of an app-wide shutdown gracefully shuts down kernels. For example, if someone added a Quit option to the Windows system tray menu tomorrow, it would break given the hack in this PR.

### DIFF
--- a/applications/desktop/src/main/index.js
+++ b/applications/desktop/src/main/index.js
@@ -198,7 +198,14 @@ electronReady$
   .subscribe(createSplashSubscriber());
 
 app.on("before-quit", e => {
-  // When notebook windows are open we need to orchestrate shutdown manually.
+  // We use Electron's before-quit to give us a hook to into full app quit events,
+  // such as Command+Q on macOS.
+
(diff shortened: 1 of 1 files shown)


applications/desktop/src/main/index.js
```diff
@@ -198,7 +198,14 @@ electronReady$
...
 
 app.on("before-quit", e => {
-  // When notebook windows are open we need to orchestrate shutdown manually.
+  // We use Electron's before-quit to give us a hook to into full app quit events,
+  // such as Command+Q on macOS.
+
+  // This is broken on Windows due to a bug in Electron; see #3549.
+  // For most Windows workflows the user will be closing individual notebook windows directly,
+  // so we just avoid this code path for now.
+  if (process.platform === "win32") return;
```
rgbkrk, line 207: This is smart for now.

rgbkrk (approved):

Issue #3517, opened 2018-10-25
Feature request: Prevent nteract from trying to elevate UAC and update on exit


kiwi0fruit:
nteract tries to update every time I exit it. It's annoying. But I cannot update to version that halts and need to be shutdown in Ctrl+Alt+Del... But I find it a good idea - update on exit. But I guess it's better to ask for update first instead of trying to elevate UAC without notice. And also would be good to add somewhere deep in the menues the ability to supress the next update (such button **shoud not** be in the question for update I mentioned).

jameskoch:
I'd been seeing some errors on shutdown when running under a very low-privileged account. I bet this is it.

kiwi0fruit:
I guess not. I've got no errors concerning this feature request. My error was already fixed and hopefully the next release would be OK. So I wait for it and see annoying UAC prompts that would update nteract if I say yes.

PR #3555, opened 2018-10-29
Allow environment variable NTERACT_DESKTOP_DISABLE_AUTO_UPDATE to disable desktop auto updater.

jameskoch:
- Requested in #3517.
- Seems generally useful in minimizing log spew on machines not connected to the Internet.
- Gives an option for working around Electron 3.x crashing upon shutdown in certain Windows httpproxy environments - https://github.com/electron/electron/issues/14411


rgbkrk (approved):
I'm happy with this approach. We probably need to document this env var in a `PACKAGING.md` or another document.

PR #3559, opened 2018-10-30
Improved TS workflow [v2]

charlesjolley:
This is a second attempt at an improved TS workflow based on feedback from #3547 and further conversation with @jdetle . 

This version of the workflow largely eliminates the need to manually build typescript files, except for when you are building a package for release.

It works by requiring each typescript package to include an `index.ts` (or equivalent for any other public entry point) at the *root* of each package. This file typically will just re-export src.

With this file in place, TypeScript can now automatically discover and use dependent packages without needing any kind of compile step.

TS files are only compiled as part of a webpack build or if you run `yarn build:packages`.

As an additional benefit, this workflow is simpler now. You no longer need to manually maintain references between packages in the tsconfig files.

Note that there is draft (incorrect) documentation for the TS workflow here as well. It needs to be updated before this is merged.

Other Changes:

- designated the top-level `types` folder for type declaration. All
  flow types were consolidated to flow-typed.
- brings back individual build tasks for each project, TS projects just
  run tsc. This makes everything more consistent
- eliminated all tsc-specific tasks from the top-level package
- added `build:clean` tasks everywhere and added a top-level task of
  the same name. This way you can eliminate all build artifacts without
  having to fully reinstall all dependencies.

- stricter types for messaging
- add type stubs required for different projects to build


charlesjolley:
@rgbkrk Changing circleci to build in series seems to work. Want to merge?

doc/design/typescript.md
```diff
@@ -0,0 +1,146 @@
...
+can continue to use the package.
+
+> NOTE: if you search the source and determine that no other code depends on
+> your package, or all dependent packages are already converted to TypeScript,
+> then you can omit this file. An easy way to check for dependents is to use
+> `yarn why @nteract/your-package-name`.
+
+### Convert the files
+
+Exercise left to reader. Be sure to convert both the source files and tests.
```
jdetle, line 81: As a matter of principle I have to remove this :D

rgbkrk:
Thank you both so much, I'm glad to have this documented and for the velocity to keep up. 🚀

doc/design/typescript.md
charlesjolley: PR's encouraged! ;)

jdetle: Haha I just _hated_ reading this back in my math days :D thanks for the great work.

PR #3566, opened 2018-10-31
Include blueprint in deps

emeeks:
Pretty simple, just adds blueprint/select into the package.json of transform-dataresource


rgbkrk:
Thank you. Do we need blueprint core in here too before we ship?

PR #3567, opened 2018-10-31
Better names for Data Explorer Vars

emeeks:
Just going through this and using better names for all the vars so other human beings can read my code.

### DIFF
--- a/packages/transform-dataresource/src/ParallelCoordinatesController.js
+++ b/packages/transform-dataresource/src/ParallelCoordinatesController.js
@@ -172,8 +172,8 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
         { uniqueValues: [], valueHash: {} }
       );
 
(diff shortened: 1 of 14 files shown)


packages/transform-dataresource/src/ParallelCoordinatesController.js
```diff
@@ -24,16 +24,16 @@ type Props = {
 
 const axisSize = [40, 380];
 
-const connectorFunction = d => d.Country;
+const connectorFunction = columnDatapoint => columnDatapoint.Country;
```
rgbkrk, line 27: Wait, this isn't a hardcoded bit from the happiness dataset is it?

emeeks: Oh good catch how did that happen? I'll correct this in a separate PR.

BenRussert (approved):
I find this soooo much easier to read. Thank you for doing this @emeeks!

PR #3571, opened 2018-10-31
dont serialize objects for rxjs 6 WebSockets

rgbkrk:
RxJS 6 now serializes objects for us so we no longer need to `JSON.stringify` messages.

Major thank you to @mpacer for tracking this down on the python side to notice it was double serializing. 🙇

### DIFF
--- a/packages/rx-jupyter/src/kernels.ts
+++ b/packages/rx-jupyter/src/kernels.ts
@@ -169,7 +169,7 @@ export const connect = (
             }
           };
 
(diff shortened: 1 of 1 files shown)


mpacer:
Could this also be a problem?: 

https://github.com/nteract/nteract/blob/ce4d79184990645c257066ed90435fd1949d0820/packages/rx-jupyter/src/kernels.ts#L161-L162

rgbkrk:
> Could this also be a problem?

Amusingly I just tore out our old ability to send on raw strings as a result of rxjs 6 changes.

PR #3595, opened 2018-11-01
attempt to enhance message types for rx-jupyter

rgbkrk:
Bit of a follow up to https://github.com/nteract/nteract/pull/3571.

Since I'm just getting started on adapting to typescript, I'd love any feedback.

### DIFF
--- a/packages/rx-jupyter/src/kernels.ts
+++ b/packages/rx-jupyter/src/kernels.ts
@@ -6,6 +6,8 @@ import { createAJAXSettings, ServerConfig } from "./base";
 import urljoin from "url-join";
 import URLSearchParams from "url-search-params";
 
+import { JupyterMessage } from "@nteract/messaging";
+
 /**
  * Creates an AjaxObservable for listing running kernels.
  *
@@ -119,7 +121,7 @@ export const connect = (
   kernelID: string,
   sessionID?: string
 ) => {
-  const wsSubject = webSocket<string>(
+  const wsSubject = webSocket<JupyterMessage>(
     formWebSocketURL(serverConfig, kernelID, sessionID)
   );
 
@@ -156,7 +158,7 @@ export const connect = (
   // and ensuring it's serialized
   return Subject.create(
     Subscriber.create(
-      (message: any) => {
+      (message?: JupyterMessage) => {
         if (typeof message === "object") {
           const sessionizedMessage = {
             ...message,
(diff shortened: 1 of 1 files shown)


packages/rx-jupyter/src/kernels.ts
```diff
@@ -114,12 +114,17 @@ export const formWebSocketURL = (
   return url.replace(/^http(s)?/, "ws$1");
 };
 
+// TODO: Figure out how/where to bring over `@nteract/messaging` types for messages into `rx-jupyter`
```
captainsafia:
```suggestion
import { JupyterMessage } from '@nteract/messaging';
```

packages/rx-jupyter/src/kernels.ts
```diff
@@ -114,12 +114,17 @@ export const formWebSocketURL = (
...
+  header: {};
+};
+
 export const connect = (
   serverConfig: ServerConfig,
   kernelID: string,
   sessionID?: string
 ) => {
-  const wsSubject = webSocket<string>(
+  const wsSubject = webSocket<Message>(
```
captainsafia:
```suggestion
  const wsSubject = webSocket<JupyterMessage>(
```

packages/rx-jupyter/src/kernels.ts
```diff
@@ -156,7 +161,7 @@ export const connect = (
   // and ensuring it's serialized
   return Subject.create(
     Subscriber.create(
-      (message: any) => {
+      (message?: Message) => {
```
captainsafia:
```suggestion
      (message?: JupyterMessage) => {
```

captainsafia (commented):
Posted some suggestions on how I think it might look.

I think we might be able to use the generics in the `JupyterMessage` interface to constrain the particular message we expect to get from the `connect`.

packages/rx-jupyter/src/kernels.ts
rgbkrk: I love that we can import typescript types without an import type syntax.

rgbkrk:
I think I'll rebase since I left the default message when using the "suggested changes" feature.

packages/rx-jupyter/src/kernels.ts
```diff
@@ -114,12 +114,17 @@ export const formWebSocketURL = (
   return url.replace(/^http(s)?/, "ws$1");
 };
 
+import { JupyterMessage } from '@nteract/messaging';
+type Message = {
```
jdetle: I dont think this is getting used anywhere.

rgbkrk:
This works really well now, thank you @captainsafia.

PR #3597, opened 2018-11-01
Update Pagers styleguidist documentation

captainsafia:


packages/presentational-components/src/components/pagers.md
```diff
@@ -1,9 +1,15 @@
+The `Pagers` component is a container component that serves as the container for the pagers of a cell. Pagers are generally used to display information that is not an output. For example, you might use a pager to display the output of a cell magic. You'll generally have both a `Pagers` component and an `Outputs` component as part of your notebook cell so that you can display both pagers and outputs.
```
rgbkrk: Classically, `Pagers` hold the introspection output (`?` and `??`).

```python
import pandas as pd

# in outputs
display(pd)

# in pager
pd?
```

screen shot 2018-11-01 at 10 14 10 am

willingc: Some additional suggestions:

As a container component, `Pagers` manage a collection of the cell's pagers.  A pager is used to provide, but not modify, additional information. Common uses include introspection with `?` and `??` and help text for commands. For example, you might use a pager to display the output of a cell magic. 

A cell may have an `Outputs` component and a `Pagers` component. `Outputs` typically contain calculation results or actions on a cell, and `Pagers` hold information, like help text, found from introspecting a cell.

packages/presentational-components/src/components/pagers.md
```diff
@@ -1,9 +1,15 @@
...
+
 ```jsx static
 import { Pagers } from "@nteract/presentational-components"
 ```
 
-Display a pager area
+Similar to other components, the `Pagers`  component provides a `hidden` flag that determines whether or not the contents of the component should be rendered. You can try it out below by toggling the value of the `hidden` prop in the code from `false` to `true`.
```
willingc, line 12: Looks like an extra space between Pagers and component. 

...that determines if a Pager component should be rendered.

willingc (commented):
Looks good Safia. I made some suggestions. Not sure if they are helpful but tossing out as a different way to define a difficult to describe term.

captainsafia:
Thank you for the feedback everyone! I've made changes based on it.

rgbkrk:
Awesome, thanks Safia!

PR #3598, opened 2018-11-01
Update Prompt styleguidist documentation

captainsafia:


packages/presentational-components/src/components/prompt.md
```diff
@@ -1,26 +1,29 @@
+The `Prompt` component is typically a child of the `Input` component and serves two purposes.
+First, it serves as a visual indicator of input entry for end users. Secondly, it allows us to display relevant information about the state of a cell to the end user.
```
rgbkrk, line 2: 👍  This is a solid intro.

packages/presentational-components/src/components/prompt.md
```diff
@@ -1,26 +1,29 @@
...
+
 ```jsx static
 import { Prompt } from "@nteract/presentational-components"
 ```
 
-Display an empty prompt
+A `<Prompt/>` component with no props will render an empty set of brackets.
```
rgbkrk, line 8: "This tells our users that the cell hasn't been run yet"

packages/presentational-components/src/components/prompt.md
```diff
@@ -1,26 +1,29 @@
...
 
 ```
 <Prompt />
 ```
 
-Set the counter to 12
+You can use the `counter` prop to pass a number that will be rendered within the brackets. If you're working in the Jupyter ecosystem, this number will likely be the `execution_count` returned by the Jupyter kernel. This `execution_count` indicates the ordinal position in which the current cell was executed.
```
rgbkrk, line 14: I'm not sure about the `ordinal` position here, though I don't know how else to describe what is effectively the N-th execution. 🤔

rgbkrk (approved):
I left a little bit of feedback, I'm not sure if what I said is actionable right now though. I think we can refine it with future PRs as this text is better than the current lack of text.

PR #3599, opened 2018-11-01
Data Explorer clean up parallel coordinates

emeeks:
* Remove reference to Country in connector and replace with primary key based method
* Do the same for the grey color of points
* Go back to a fixed color regardless of brushing
* Suppress legend if < 18 instead of < 10 to match 18 max colors elsewhere
* Adjust point style and transparency to make filtered lines less prominent
* Make filtered values appear in grey on hover in Explore mode
* update valuePosition -> index in HTMLLegend

### DIFF
--- a/packages/transform-dataresource/src/ParallelCoordinatesController.js
+++ b/packages/transform-dataresource/src/ParallelCoordinatesController.js
@@ -24,8 +24,6 @@ type Props = {
 
 const axisSize = [40, 380];
 
-const connectorFunction = columnDatapoint => columnDatapoint.Country;
-
 function parallelizeData(data, metrics, schemaFields, primaryKey) {
   const minmax = {};
   const screenScales = {};
@@ -120,6 +118,9 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
 
     const hiddenHash = new Map();
 
+    const connectorFunction = columnDatapoint =>
+      primaryKey.map(key => columnDatapoint[key]).join(",");
+
     Object.keys(columnExtent).forEach(key => {
       if (columnExtent[key]) {
         const extent = columnExtent[key].sort((a, b) => a - b);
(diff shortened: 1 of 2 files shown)


packages/transform-dataresource/src/ParallelCoordinatesController.js
```diff
@@ -120,6 +118,9 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
 
     const hiddenHash = new Map();
 
+    const connectorFunction = columnDatapoint =>
+      primaryKey.map(key => columnDatapoint[key]).join(",");
```
rgbkrk, line 122: Excellent

packages/transform-dataresource/src/ParallelCoordinatesController.js
```diff
@@ -172,12 +173,20 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
...
+        []
+      );
+
+      uniqueDimsForColors.forEach((value: string, index: number) => {
         colorHash[value] = colors[index % colors.length];
       });
 
       additionalSettings.afterElements =
-        uniqueValues.length < 10 ? (
+        uniqueValues.length <= 18 ? (
```
rgbkrk, line 189: What's the magic number 18 mean?

packages/transform-dataresource/src/ParallelCoordinatesController.js
```diff
@@ -238,7 +250,7 @@ class ParallelCoordinatesController extends React.Component<Props, State> {
             opacity: hiddenHash.get(
               primaryKey.map(key => datapoint[key]).join(",")
             )
-              ? 0.5
+              ? 0.15
```
rgbkrk, line 253: Each of these thresholds... how were they tuned? How would you want us to evaluate if someone seeks to change this later?

packages/transform-dataresource/src/ParallelCoordinatesController.js
emeeks: I have 18 colors in the default color scheme. It's hard to get color differentiability for points and lines beyond 18 and by hard I mean really hard. Maybe in the future when we have more palette functionality, I'll make this dynamic.

packages/transform-dataresource/src/ParallelCoordinatesController.js
emeeks: There's no good research on parallel coordinates styling, so I made a judgement call that we'd want to really scale back the appearance of points when they're not in the filter bands. Once we see people using this mode more, we might end up with a more user-driven approach but until then this is a good place to start. There's not a uniform opacity value that would be applied across lines and points, so I don't see much to pull out into a constant, but we might allow users to style things dynamically later on if we implement a more exploratory data explorer. I wouldn't want to invest too much into that right now, because I want to see if people are actually going to use it.

rgbkrk: Some form of this summary might be good to describe just above this line, as a comment.

packages/transform-dataresource/src/ParallelCoordinatesController.js
rgbkrk: This would make a good comment just above the line.

PR #3608, opened 2018-11-03
Update ExecuteResult styleguidist documentation

captainsafia:


packages/outputs/src/components/execute-result.md
```diff
@@ -1,4 +1,6 @@
-The `<ExecuteResult />` component will render execution outputs (SVGs, HTML, text, etc) in the front end.
+The `ExecuteResult` component is a container component that is used to render output data from the kernel. Similar to the `DisplayData` component, the `ExecuteResult` component acts as a simple switch over its children and correctly associates the media type of the data with the media type that a component supports.
```
willingc: Perhaps "...used to render data returned by the kernel. The returned data may be in a number of output formats. Similar...the media type of the returned data..."

willingc (approved):
Looks good @captainsafia. Some minor suggestions for reader clarity.

packages/outputs/src/components/execute-result.md
```diff
@@ -1,4 +1,6 @@
...
+
+Why do so many component employ this pattern? It's in the interest of readability. Although each component executes the same logic under the hood, the separation of components by output types allows you to write meaningful JSX from these React components.
```
rgbkrk, line 3: ✨

One extra thing about ExecuteResult that we aren't using at the moment is the execution count.

This component is the result of the cell execution (as opposed to display data).

packages/outputs/src/components/execute-result.md
captainsafia: Thanks! I've updated the PR.

rgbkrk (approved):

PR #3609, opened 2018-11-03
Update Output styleguidist documentation

captainsafia:


packages/outputs/src/components/output.md
```diff
@@ -1,4 +1,6 @@
-The `<Output />` element is a glorified Switch statement for picking what to render an output with.
+The `Output` element takes a single `output` prop and several child components and renders the appropriate components to render each output type.
```
willingc, line 1: ...components needed for each output type.

rgbkrk (approved):
The main takeaway, at least for hydrogen is that they can completely override how to render a given output. We probably need an example that uses a custom component for Jupyter errors or something like that.

packages/outputs/src/components/output.md
```diff
@@ -14,6 +16,7 @@ const output = Object.freeze({
   <StreamText />
 </Output>;
 ```
+We can also render errors that are returned from the Jupyter kernel we're connected to.
```
willingc: Errors returned from the connected Jupyter kernel can be rendered.

willingc (approved):
Minor suggestions.

captainsafia:
Thanks for the feedback! I made changes.

PR #3610, opened 2018-11-03
Update StreamText styleguidist documentation

captainsafia:


packages/outputs/src/components/stream-text.md
```diff
@@ -1,3 +1,9 @@
+The Jupyter kernel will return outputs that contain data from standard out or standard error using the `stream` message type. The data contained in this messages consists of a string containing the contents of the output. To render these types of outputs from the kernel in your own Jupyter front-end, you can use the `StreamText` component.
+
+The `StreamText` component takes a `text` prop which contains the contents of the stream. The `name` prop, which can be one of `stdout` or `stderr` dictates whether the text stream originated from standard out or standard error respectively.
+
+The `name` component is used to set the class of the `StreamText` component to `nteract-display-area-${name}` so that you can add custom styles to stream outputs.
```
rgbkrk: Oh no, I totally don't want to make this class be an interface we have to support.

rgbkrk:
```suggestion
```

rgbkrk: Let's trim this out for now and then update our component to not rely on this class name.

PR #3611, opened 2018-11-03
Update Host styleguidist documentation

captainsafia:
The documentation is done but I am having some trouble with the changes I'm trying to to make the styleguidist config. I have an image in this tutorial and I've been trying to use the `styles` override to set a width of 100% so it doesn't overflow the parent container but it is not working. Not sure what I am doing wrong.

[This](https://react-styleguidist.js.org/docs/cookbook.html#how-to-change-styles-of-a-style-guide) is what I am using.

### DIFF
--- a/packages/host-cache/src/components/host.md
+++ b/packages/host-cache/src/components/host.md
@@ -31,4 +31,10 @@ The configuration details for the Binder instance that you can connect to are pe
 
 You can also see this information stored under the "@BinderKey@${props}` key in your browser's LocalStorage.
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
😅  If you want a cheap workaround, you could [render it with React and tell styleguidist not to make an editor for it](https://react-styleguidist.js.org/docs/cookbook.html#how-to-render-react-components-that-arent-part-of-the-style-guide):

```jsx noeditor
<img src="" style={} />
```

captainsafia:
Alright! Went ahead and used the React rendering trick to show the image with the correct width.

PR #3612, opened 2018-11-03
Update Kernel styleguidist documentation

captainsafia:
I simplified the example component so it's not as interactive but a little bit cleaner. I'll reserve a more interactive approach for using this component to the nteract play tutorial.


rgbkrk:
You're probably right about needing to simplify this one. I tend to showcase this little component to people having it run R code and it blows their mind. Somehow people seem a little boxed in for thinking and this starts to show how this all can be used together.

PR #3614, opened 2018-11-03
Migrate MathJax to TypeScript

captainsafia:
Closes https://github.com/nteract/nteract/issues/3573.

Had to sprinkle a few any's in there, especially in `load-script.ts`.

This builds and passes both flow and unit tests.


captainsafia:
Welp! @rgbkrk brought to my attention that the styleguidist configuration needs to be updated to accommodate this change. Fiddling with that now.

captainsafia:
> Welp! @rgbkrk brought to my attention that the styleguidist configuration needs to be updated to accommodate this change. Fiddling with that now.

This is now done. I had to change the way MathJax was exporting its classes.

packages/mathjax/index.ts
```diff
@@ -0,0 +1 @@
+export * from './src';
```
rgbkrk: What made this file be required? Isn't our usual `index` file as `src/index`?

packages/mathjax/src/load-script.tsx
```diff
@@ -1,16 +1,14 @@
...
+interface Options {
+  type: string;
+  charset: string;
+  async: any;
+  text?: string;
+  attrs?: { [key: string]: any };
 };
 
-type Callback = (error: ?Error, script: HTMLScriptElement) => void;
+type Callback = (script: HTMLScriptElement, error?: Error) => void;
```
rgbkrk, line 11: Is there a necessity for jumping out from the typical node.js callback style of `(err, ...data)` here? We don't export this file so it's ok that it's an inner API change, would love to know the reasons (typescript ease of typing?)

packages/mathjax/src/text.tsx
```diff
@@ -1,16 +1,15 @@
...
-  onRender: ?Function
+interface Props {
+  children: string;
+  onRender?: Function;
 };
 
-class MathJaxText_ extends React.Component<Props & MathJaxContextValue, null> {
-  nodeRef: React.ElementRef<*>;
+class MathJaxText_ extends React.Component<Props & MathJaxContextValue> {
+  nodeRef: React.RefObject<HTMLDivElement>;
```
rgbkrk, line 12: 👍

packages/mathjax/index.ts
captainsafia: Nope. Didn't need this at all. It's an artifact from my export tinkering. Removing it.

packages/mathjax/src/load-script.tsx
captainsafia: Yep. TypeScript doesn't like it when you have a possibly-undefined variable before a definitely defined one.

rgbkrk: Forgive my lack of experience, I'm curious if there's some way to [write it like this (example on ts playground)](https://www.typescriptlang.org/play/#src=interface%20Callback%3CResult%3E%20%7B%0D%0A%20%20(error%3A%20null%2C%20result%3A%20Result)%3A%20void%3B%0D%0A%20%20(error%3A%20Error%2C%20result%3A%20null)%3A%20void%3B%0D%0A%7D%0D%0A%0D%0Afunction%20doThatThing(x%3A%20boolean%2C%20cb%3A%20Callback%3Cnumber%3E)%20%7B%0D%0A%20%20%20%20if%20(x)%20%7B%0D%0A%20%20%20%20%20%20%20%20cb(null%2C%2042)%3B%0D%0A%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20cb(new%20Error(%22errors%20happen%22)%2C%20null)%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Aconst%20logIt%3A%20Callback%3Cnumber%3E%20%3D%20(err%2C%20num)%20%3D%3E%20%7B%0D%0A%20%20%20%20if%20(err)%20%7B%0D%0A%20%20%20%20%20%20%20%20console.error(err)%3B%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20console.log(num)%0D%0A%20%20%20%20%20%20%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0AdoThatThing(true%2C%20logIt)%3B%0D%0AdoThatThing(false%2C%20logIt)%3B):

```tsx
interface Callback<Result> {
  (error: null, result: Result): void;
  (error: Error, result: null): void;
}
```

then declare the callback directly like this:

```tsx
const logIt: Callback<number> = (err, num) => {
    if (err) {
        console.error(err);
    }
    else {
        console.log(num);
  }
}
```

It does mean we have to explicitly call with the last argument having a `null` when there is an error.

captainsafia: That definitely is a way to do it. The original error was about `error` being potentially undefined.  TypeScript [recommends](https://www.typescriptlang.org/docs/handbook/functions.html) that optional parameters follow required parameters so I think it's good to follow this pattern.

Is this the only place in the code where we were using Node-style callbacks or is there more?

rgbkrk: Oops, hit the wrong button there.

There are other spots in the code base using the node callback pattern though they're probably interfacing with node APIs directly. The way ahead is probably to adapt this to a promises backed interface.

rgbkrk:
Thank you for doing this and verifying the docs setup. This is definitely paving the way forward!

PR #3619, opened 2018-11-06
Add Markdown styleguidist documentation

captainsafia:


packages/markdown/examples.md
```diff
@@ -0,0 +1,23 @@
+The `MarkdownRender` component allows you to render Markdown text per the Commonmark specification. It's got one major power-up though, it allows you to render in-line and block math using LaTeX syntax too.
```
rgbkrk, line 1: The way I think of our `@nteract/markdown` package is that it's a manifestation of how _we_ render markdown since there are competing ways of extending on top of commonmark. This provides a way to have a consistent render across different apps.

packages/markdown/examples.md
```diff
@@ -0,0 +1,23 @@
...
+```
+const MarkdownRender = require("@nteract/markdown").default;
+
+<MarkdownRender source={`Just some $\delta_{\alpha}$ math.`}/>
+```
+
+Alternatively, you can pass the source as a child of the `MarkdownRender` component.
+
+```
+const MarkdownRender = require("@nteract/markdown").default;
```
rgbkrk, line 20: These make me think we should make them named exports instead.

There is a way to work with styleguidist where you don't have to do the `require`s though. Check out how presentational components does it.

captainsafia: I think this happens because we have the `components` property set in the styleguide config for Presentational Components. The docs for examples are set up as a single file markdown.

PR #3622, opened 2018-11-06
commutable: switch Notebook and Outputs from Maps to Records

rgbkrk:
My goal is to turn everything that should be a record into a record (cells, outputs, notebook). Many of these have been `Map`s since the early days. One big benefit is that we can have a lot less `null` checking (no more `const outputType = output.get('output_type'); if(outputType) {...}`, we can just do `output.outputType`).

Some things must remain maps, like the aptly titled cell map (`Map<string, ImmutableCell>`) and then others should remain maps because they're unspecced areas like metadata.

* [x] Notebook
* [x] Output
  * [x] v4 Output
  * [x] v3 Output

Since this PR has gotten big enough, I will happily do cells in a new PR. We'll then be on fully typed records which should make it easier for us to transition off of Immutable (if we end up wanting to for sure).

### DIFF
--- a/packages/commutable/src/index.ts
+++ b/packages/commutable/src/index.ts
@@ -54,9 +54,6 @@ export const fromJS = (
       `commutable was passed an Immutable.Record structure that is not a notebook`
     );
   }
-  if (notebook.nbformat === 4) {
-    console.log(notebook);
-  }
 
   if (notebook.nbformat === 4 && notebook.nbformat_minor >= 0) {
     var v4Notebook = notebook as v4.Notebook;
--- a/packages/commutable/src/index.ts
+++ b/packages/commutable/src/index.ts
@@ -1,4 +1,4 @@
-import { Map as ImmutableMap } from "immutable";
+import { Map as ImmutableMap, Record, RecordOf } from "immutable";
 import * as v4 from "./v4";
 import * as v3 from "./v3";
 import { ImmutableNotebook, JSONType } from "./types";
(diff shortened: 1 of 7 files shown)


rgbkrk:
Something that was really easy to do in flow was using Union types and then checking properties to find out which is which. I'm unsure how to do this in typescript, which is too bad because a lot of other code (that hasn't been converted yet) uses this property.

screen shot 2018-11-06 at 7 35 38 am

This is the v4 notebook type

```ts
export type Notebook = {
  cells: Array<Cell>;
  metadata: Object;
  nbformat: 4;
  nbformat_minor: number;
};
```

and this is the v3 Notebook type

```ts
export type Notebook = {
  worksheets: Worksheet[];
  metadata: object;
  nbformat: 3;
  nbformat_minor: number;
};
```

These are probably better called `OnDiskNotebook` 😄

rgbkrk:
Seems like the only way is to make it explicit with `as`:

```ts
  if (notebook.nbformat === 4 && notebook.nbformat_minor >= 0) {
    var v4Notebook = notebook as v4.Notebook;

    if (
      Array.isArray(v4Notebook.cells) &&
      typeof notebook.metadata === "object"
    ) {
      return v4.fromJS(v4Notebook);
    }
  } else if (notebook.nbformat === 3 && notebook.nbformat_minor >= 0) {
    return v3.fromJS(notebook as v3.Notebook);
  }
```

Surely there's a better way considering how we'll be using a switch for actions in redux land.

lgeiger:
The problem is probably the `ImmutableNotebook` type.

Typescript has great support for union types. Checkout [this playground example](https://www.typescriptlang.org/play/index.html#src=type%20Notebookv4%20%3D%20%7B%0D%0A%20%20cells%3A%20object%5B%5D%3B%0D%0A%20%20metadata%3A%20Object%3B%0D%0A%20%20nbformat%3A%204%3B%0D%0A%20%20nbformat_minor%3A%20number%3B%0D%0A%7D%3B%0D%0A%0D%0Atype%20Notebookv3%20%3D%20%7B%0D%0A%20%20worksheets%3A%20object%5B%5D%3B%0D%0A%20%20metadata%3A%20object%3B%0D%0A%20%20nbformat%3A%203%3B%0D%0A%20%20nbformat_minor%3A%20number%3B%0D%0A%7D%3B%0D%0A%0D%0Atype%20Notebook%20%3D%20Notebookv4%20%7C%20Notebookv3%0D%0A%0D%0Afunction%20test(notebook%3A%20Notebook)%20%7B%0D%0A%20%20if%20(notebook.nbformat%20%3D%3D%3D%204)%20%7B%0D%0A%20%20%20%20console.log(notebook.cells)%0D%0A%20%20%7D%20else%20%7B%0D%0A%20%20%20%20console.log(notebook.worksheets)%0D%0A%20%20%7D%0D%0A%7D).

rgbkrk:
Perhaps it's just something in my editor, but it does infer that it's a `Notebook` rather than an `ImmutableNotebook` there. What it hasn't been able to narrow to is `v3.Notebook` and `v4.Notebook`.

rgbkrk:
This is now ready for review. 😄

rgbkrk:
Putting WIP back on as I need to adapt v3 and all the types to use the same immutable output types so that no maps are sneaking on through.

rgbkrk:
It's likely I won't touch this PR again until tonight as I'm at GraphQL training.

packages/commutable/src/index.ts
```diff
@@ -25,45 +25,50 @@ export {
...
       return notebook;
     }
     throw new TypeError(
-      `commutable was passed an Immutable.Map structure that is not a notebook`
+      `commutable was passed an Immutable.Record structure that is not a notebook`
     );
   }
 
   if (notebook.nbformat === 4 && notebook.nbformat_minor >= 0) {
+    var v4Notebook = notebook as v4.Notebook;
```
rgbkrk, line 62: At least for me, it appeared that `notebook` wasn't being narrowed to a `v4.Notebook` for me here.

captainsafia: I believe the fix for this that you can do as a fast follow is to create an NbFormatEnum and use that as the type of the `nbformat` field and in this comparison here.

```js
enum NbFormatEnum {
    "v4" = 4
}

if (notebook.format === NbFormatEnum.v4) {}
```

Also, I'm not sure if the second clause after the `&&` might be throwing it off.

captainsafia (approved):

rgbkrk:
Thanks for the approval and the enum suggestion. I'll try that now.

PR #3626, opened 2018-11-07
Migrate NotebookRender to TypeScript

captainsafia:
Closes https://github.com/nteract/nteract/issues/3576.

Documenting this was a bit challenging because it isn't a siloed component so I felt weird about providing full-interactive examples in what should be an API reference. Not sure how to approach documenting these types of components yet.


captainsafia:
Looks like I need to install some type declarations and write a few up. I'll work on this tomorrow.

rgbkrk:
As we talked about in chat, I don't think we should document monolith components like this in our style guide. The style guide should live as a standalone section for us and other contributors to reference when building apps. This belongs in docs, though likely in a section devoted to blogging with notebooks, in this case using Gatsby.

packages/notebook-render/src/index.tsx
```diff
@@ -29,15 +28,15 @@ import {
...
 
-type Props = {
-  displayOrder: Array<string>,
-  notebook: any,
-  transforms: Object,
-  theme: "light" | "dark"
+interface Props  {
+  displayOrder: Array<string>;
+  notebook: any;
+  transforms: Object;
```
charlesjolley: Should be object (lowercase o)

packages/notebook-render/src/index.tsx
```diff
@@ -29,15 +28,15 @@ import {
   Cells
 } from "@nteract/presentational-components";
 
-type Props = {
-  displayOrder: Array<string>,
-  notebook: any,
-  transforms: Object,
-  theme: "light" | "dark"
+interface Props  {
+  displayOrder: Array<string>;
```
charlesjolley: Should be string [] per our earlier discussion? (Simple srrays use bracket. Complex types get array<>)

packages/notebook-render/src/index.tsx
```diff
@@ -29,15 +28,15 @@ import {
...
+  notebook: any;
+  transforms: Object;
+  theme: "light" | "dark";
 };
 
-type State = {
-  notebook: any
+interface State {
+  notebook: any;
```
charlesjolley: Can/should this be made more specific? You lose typing basically

charlesjolley (commented):
Suggested a few new type colors for the bikeshed

packages/notebook-render/src/index.tsx
rgbkrk: Now that commutable is completely typescripted with some nice records, we could probably use the definition there. This component probably takes the raw notebook format or commutable format as props and stores the derived state as a commutable notebook. Both those types are in commutable.

types/nteract__presentational-components/index.d.ts
```diff
@@ -0,0 +1,12 @@
...
+    export const Input: any;
+    export const Outputs: any;
+    export const Pagers: any;
+    export const Prompt: any;
+    export const PromptBuffer: any;
+    export const Source: any;
+    export const Cell: any;
+    export const Cells: any;
+    export const HeaderEditor: any;
+}
```
rgbkrk, line 12: There's a part of me that wants to convert these to typescript next.

captainsafia: Agreed! I mentioned this to @jdetle and I think he's working on it at the moment.

packages/notebook-render/src/index.tsx
captainsafia: Done. I used `Notebook | ImmutableNotebook`.

types/nteract__presentational-components/index.d.ts
jdetle: Yeah, on my todolist for today.

captainsafia:
OK! I think this is good to merge. Took a while for me to get the tests passing with all the new types.

rgbkrk:
Awesome, thank you

PR #3629, opened 2018-11-07
switch commutable cells from map to records

rgbkrk:
Follow on to #3622. Now for cells! We are now complete in `Record`ification of commutable.

This overhauls the structure of commutable while keeping (many of) the same exports. Records are grouped together by type `cells`, `outputs`, and `notebook`. Conversion utilities for going from formats are still called `v3` and `v4` (maybe it's worth calling these `convertFromV4` and `convertFromV3`). The file formerly known as `types`, which kept becoming this large dumping ground for things, now is called `primitives` that are shared across a few of the other files.


rgbkrk:
Ready for review!

captainsafia (approved):

PR #3636, opened 2018-11-10
[WIP] Initial metadata persistence for data explorer

emeeks:
* Honor incoming metadata
* Fire onMetadataChange when chart settings change


packages/transform-dataresource/src/index.js
```diff
@@ -133,15 +131,66 @@ const MetadataWarning = ({ metadata }) => {
 class DataResourceTransform extends React.Component<Props, State> {
   static MIMETYPE = mediaType;
 
+  //FOR TESTING PURPOSES ONLY THE METADATA HAS SAMPLE SETTINGS FOR A GRADUATED SYMBOL PLOT
+
   static defaultProps = {
-    metadata: {},
+    metadata: {
+      dx: {
```
rgbkrk, line 129: We'll definitely want to fully type our segment of metadata since we're going to have to track and adapt to previous versions of our own metadata (a notebook created today should render next year).

rgbkrk:
Does the component work (change) when:

* `onMetadataChange` is not set (locally managed state)?
* `metadata` changes by the props changing (does it trigger the `setState` that looks like it needs to be called)?

Could we have tests for both?

rgbkrk:
Something to consider -- should [`getDerivedStateFromProps` be implemented to react to props changes](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)? Note: [React folks recommend making it a fully controlled component](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html).

PR #3637, opened 2018-11-10
convert fs-observable to typescript!

rgbkrk:
This is really great because we can remove some strange code and let typescript do it's thing thanks to typescript knowing node's `fs` well and `rxjs` having already been written in TypeScript. There's now less code and it's _more_ accurately typed.


rgbkrk:
This one was strange enough for flow that I didn't create the `index.js.flow` file, instead opting to remove it from the flowconfig resolution.

rgbkrk:
Now with some decent flow types!

I did have to `$FlowFixMe` one thing that I feel confident we'll be able to tackle as part of the typescript migration.

packages/fs-observable/src/index.ts
```diff
@@ -0,0 +1,72 @@
+import * as fs from "fs";
```
rgbkrk, line 1: Could have sworn I did a `git mv index.js index.ts`. Perhaps my rebasing and the size of these changes caused `git` to say _nah_.

charlesjolley (approved):
LGTM!

packages/fs-observable/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { Observable, bindNodeCallback } from "rxjs";
+import { mergeMap } from "rxjs/operators";
+
+import mkdirp from "mkdirp";
+
+/**
+ * Deletes a file, if it exists
```
jdfreder, line 9: Hurray for 3rd person descriptive comments!

PR #3639, opened 2018-11-10
convert @mybinder/host-cache to typescript

rgbkrk:
Converted host-cache to typescript.

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -1,5 +1,10 @@
 // @format
+
 const path = require("path");
+
+const reactDocgenTypescript = require("react-docgen-typescript").withCustomConfig(
+  "./tsconfig.base.json"
+);
 
 const babelFlowConfig = require("./babel.flow.config");
 const babelTypescriptConfig = require("./babel.typescript.config");
@@ -7,6 +12,8 @@ var {
   exclude,
   mergeDefaultAliases
 } = require("./packages/webpack-configurator");
+
+const typescriptPropsParser = reactDocgenTypescript.parse;
 
 module.exports = {
   title: "nteract components",
@@ -28,7 +35,8 @@ module.exports = {
     },
     {
       name: "Binder Hosts and Kernels",
-      components: "packages/host-cache/src/components/*.ts"
+      components: "packages/host-cache/src/components/*.ts",
+      propsParser: typescriptPropsParser
     },
     {
       name: "Directory Listing",
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -28,7 +28,7 @@ module.exports = {
     },
     {
       name: "Binder Hosts and Kernels",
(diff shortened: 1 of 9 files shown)


styleguide.config.js
```diff
@@ -28,7 +35,8 @@ module.exports = {
     },
     {
       name: "Binder Hosts and Kernels",
-      components: "packages/host-cache/src/components/*.js"
+      components: "packages/host-cache/src/components/*.ts",
```
captainsafia:
```suggestion
      components: "packages/host-cache/src/components/*.tsx",
```

rgbkrk: _@rgbkrk smacks his forehead_

rgbkrk:
This is looking super good now, thank you for finishing it off @captainsafia!

rgbkrk:
It seems like props descriptions aren't coming up with the react docgen typescript stuff. Also, there's a syntax error with the `Host` component that affects rendering the component + props. `Kernel` renders fine though. Probably not a deal breaker for now.

captainsafia:
Hm. Yeah. I'm hesitant to spend too much time squaring away any of these build issues once we've fully migrated. If the problem persists then, we can figure it out.

Other than that, this is good to go!

rgbkrk:
I left this unsent yesterday -- 

The `propsParser` isn't getting called for this individual suite of components (seems we can't set it per suite). We'll have to put it at the top level, making a hybrid function that calls the typescript props parser for the tsx files and the regular parser for the others. I've got the hook setup so it's starting to be promising. Hitting the classic error from styleguidist though:

```
Warning: packages/host-cache/src/components/kernel.tsx matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.

It usually happens when using third-party libraries, see possible solutions here:
https://react-styleguidist.js.org/docs/thirdparties.html
```

That's largely due to the kernel component having multiple exports which was handled well by regular react-docgen.

> I'm hesitant to spend too much time squaring away any of these build issues once we've fully migrated. If the problem persists then, we can figure it out.

Ok, that sounds good. I'll make a follow up PR to see if I can get this handled and timebox myself so I don't try too much on it given that we're migrating anyway.

rgbkrk:
Code I tried to use to adapt things:

```js
const reactDocgenTypescript = require("react-docgen-typescript").withCustomConfig(
  "./tsconfig.base.json"
);
const reactDocgen = require("react-docgen");

function propsParser(filePath, source, resolver, handlers) {
  // If it ends in tsx, use the typescript props parser
  if (filePath.endsWith("tsx")) {
    return reactDocgenTypescript.parse(source, resolver, handlers);
  }

  // Otherwise use react-docgen's, as documented in https://react-styleguidist.js.org/docs/configuration#propsparser
  return reactDocgen.parse(source, resolver, handlers);
}
```

For now I'll leave this one alone.

PR #3640, opened 2018-11-11
Migrate transform-vdom to TypeScript

captainsafia:
Closes https://github.com/nteract/nteract/issues/3587.

I left some comments and questions inline.


packages/transform-vdom/src/object-to-react.ts
```diff
@@ -35,16 +33,22 @@
...
+
+export interface VDOMEl {
+  tagName: string; // Could be an enum honestly
+  children: VDOMNode;
+  attributes: Attributes;
+  key: number | string | null;
 };
 
-type ReactArray = Array<React$Element<any> | ReactArray | string>;
+interface ReactArray extends ConcatArray<React.ReactElement<any> | string | ReactArray> {};
```
captainsafia: Is this type supposed to be equivalent to the type defined in `React.ReactArray`?

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -95,7 +99,7 @@ export function objectToReactElement(obj: VDOMEl): React$Element<any> {
       if (args[1] === undefined) {
         args[1] = null;
       }
-      args = args.concat(arrayToReactChildren(children));
+      args = args.concat(arrayToReactChildren(children) as any);
```
captainsafia: There were some complaints here about the recursive definition in the `ReactArray` type.

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -35,16 +33,22 @@
...
+  tagName: string; // Could be an enum honestly
+  children: VDOMNode;
+  attributes: Attributes;
+  key: number | string | null;
 };
 
-type ReactArray = Array<React$Element<any> | ReactArray | string>;
+interface ReactArray extends ConcatArray<React.ReactElement<any> | string | ReactArray> {};
+
+type VDOMNode = VDOMEl | string | any;
```
captainsafia: I had trouble getting the circular reference to `VDOMNode` working here. TypeScript outlines a technique for using circular references in Array-like types which I used above but not for union types. I don't know much about the goal of this type to be able to define it without using circular types.

packages/transform-vdom/src/object-to-react.ts
rgbkrk: I think so.

rgbkrk:
In case it helps (perhaps for simplification), this has to support the VDOM spec:

* [Human Readable spec](https://github.com/nteract/vdom/blob/0.5/docs/spec.md)
* [JSON Schema](https://github.com/nteract/vdom/blob/master/vdom/schemas/vdom_schema_v1.json)

The `master` branch has some additional work that @gnestor is going to wrap up before our next release (around event binding).

packages/transform-vdom/src/object-to-react.ts
rgbkrk: For better or worse, VDOM is close to but not quite a total match to the React types. We support a subset React does, though we could allow it to expand a little bit.

captainsafia:
OK. I think I've found a good blend for typing by leveraging some of React's built-in types.

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -35,16 +33,17 @@
...
-type ReactArray = Array<React$Element<any> | ReactArray | string>;
+interface Attributes {
+  style: object;
+  key: number;
+}
 
-type VDOMNode = VDOMEl | string | Array<VDOMNode>;
+export interface VDOMEl {
+  tagName: string; // Could be an enum honestly
+  children: React.ReactNode | VDOMEl | Array<React.ReactNode | VDOMEl>;
```
captainsafia, line 43: `React.ReactNode` captures the `string` part of this (and related DOM elements). I got around the circular definition error by just copy pasting the `React.ReactNode | VDOMEl` as the value in the array *shrug*.

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -95,11 +94,11 @@ export function objectToReactElement(obj: VDOMEl): React$Element<any> {
       if (args[1] === undefined) {
         args[1] = null;
       }
-      args = args.concat(arrayToReactChildren(children));
+      args = args.concat(arrayToReactChildren(children as VDOMEl[]) as any);
```
captainsafia, line 97: I had to add some typecasting here to specify which specific type within the union type we were in based on the if-statement.

packages/transform-vdom/src/index.js.flow
```diff
@@ -0,0 +1,8 @@
+// @flow
+/* eslint-disable */
+
```
rgbkrk, line 3: I feel like the prop types can go here too, not sure how the hand made flow defs for react components should work.

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -35,16 +33,17 @@
...
-  tagName: string, // Could be an enum honestly
-  children: VDOMNode,
-  attributes: Object,
-  key: number | string | null
-};
-
-type ReactArray = Array<React$Element<any> | ReactArray | string>;
+interface Attributes {
+  style: object;
+  key: number;
```
rgbkrk, line 38: Is it ok that `Attributes` can have other properties at runtime?

rgbkrk:
@blink1073  -- we're migrating `@nteract/transform-vdom` over to typescript. Want to check this out when you get a chance?

packages/transform-vdom/src/object-to-react.ts
blink1073: Should be okay, this is the contract for the minimal required properties.

blink1073 (commented):
Looks reasonable to me, thanks @captainsafia and @rgbkrk!  TypeScript and React is still a somewhat shaky union, cf https://github.com/DefinitelyTyped/DefinitelyTyped/issues/28015

rgbkrk:
> TypeScript and React is still a somewhat shaky union

We can go without `prop-types` though right? We're just using the templated props.

blink1073:
Yup!  I was just giving you a heads up about `@types/react` being a moving target that doesn't follow semver...

rgbkrk:
😬

captainsafia:
@blink1073 Noted. Do you happen to know any workarounds for writing recursive references in union types in TypeScript? I managed to get a workaround for a recursive reference in an Array type but not a union type.

blink1073:
I've never come across a recursive reference of either type.  You jumped right into the deep end :smile:   @gnestor, have you come across them?

gnestor:
@captainsafia I have run into something like that before. 

Here's an example in phosphor: https://github.com/phosphorjs/phosphor/blob/master/packages/coreutils/src/json.ts#L10-L35

And here's the same exact example in a typescript issue: https://github.com/Microsoft/TypeScript/issues/3496#issuecomment-128553540

PR #3641, opened 2018-11-11
Migrate transform-geojson to TypeScript

captainsafia:
Closes https://github.com/nteract/nteract/issues/3584.

I left some comments and questions inline.


packages/transform-geojson/src/index.tsx
```diff
@@ -1,16 +1,21 @@
...
+interface Props {
+  data: GeoJsonObject;
+  metadata: Metadata;
+  mediaType: "application/geo+json";
+  theme: string;
 };
+
 type TileTheme = "dark" | "light";
-type TileLayer = [string, Object];
+type TileLayer = { urlTemplate: string, layerOptions: object };
```
captainsafia, line 18: I refactored the way we create TileLayer options to match up with the type definitions for the `tileLayer` function in Leaflet.

packages/transform-geojson/src/index.tsx
```diff
@@ -19,14 +24,14 @@ L.Icon.Default.imagePath = "../node_modules/leaflet/dist/images/";
 export function getLuma(el: HTMLElement): number {
   // https://en.wikipedia.org/wiki/Luma_(video)
   const style = window.getComputedStyle(el);
-  const [r, g, b] = style.backgroundColor
+  const [r, g, b] = style.backgroundColor!
```
captainsafia, line 27: The `!` here is a TypeScript way of saying that I know this type might be undefined but trust me, in this case, it is not.

packages/transform-geojson/src/index.tsx
```diff
@@ -44,11 +49,11 @@ export function getTheme(theme: string = "light", el: HTMLElement): TileTheme {
...
-  MIMETYPE: string;
-  map: Object;
-  el: ?HTMLElement;
-  geoJSONLayer: Object;
-  tileLayer: Object;
+  MIMETYPE!: string;
+  map!: L.Map;
+  el!: HTMLDivElement | null;
+  geoJSONLayer!: L.GeoJSON;
+  tileLayer!: L.TileLayer;
```
captainsafia, line 56: The `!` here are a way to allow adding types for static properties outside the context of a constructor.

I also pulled in some types from the Leaflet type definitions.

packages/transform-geojson/src/index.tsx
```diff
@@ -57,11 +62,12 @@ export class GeoJSONTransform extends React.Component<Props> {
...
   componentDidMount(): void {
-    this.map = L.map(this.el);
+    this.map = L.map(this.el!);
     this.map.scrollWheelZoom.disable();
-    this.tileLayer = L.tileLayer(...this.getTileLayer()).addTo(this.map);
+    const tileLayerOptions = this.getTileLayer();
+    this.tileLayer = L.tileLayer(tileLayerOptions!.urlTemplate, tileLayerOptions!.layerOptions).addTo(this.map);
     const geoJSON = this.props.data;
-    this.geoJSONLayer = L.geoJson(geoJSON).addTo(this.map);
+    this.geoJSONLayer = L.geoJSON(geoJSON).addTo(this.map);
```
captainsafia, line 70: According to the type definitions provided by Leaflet, there is no `geoJson` function defined in the library, just `geoJSON`. I think this is a result of us being on an older version of Leaflet. I can try installing the older type definitions. Any context into why our implementation and the type definitions might be off here?

rgbkrk: That would probably explain why this one hasn't been working. I think we could / should just upgrade this one and make sure it's working.

packages/transform-geojson/src/index.tsx
rgbkrk: Ohhhh this is so helpful. I wondered how to handle the "only initialized on mount" case.

packages/transform-geojson/src/index.tsx
captainsafia: I upgrade leaflet and tested against the geojson example notebook.

packages/transform-geojson/src/index.tsx
rgbkrk: Wow I don't even know what this original type was.

packages/transform-geojson/src/index.tsx
```diff
@@ -57,11 +62,12 @@ export class GeoJSONTransform extends React.Component<Props> {
   static MIMETYPE = MIMETYPE;
 
   componentDidMount(): void {
-    this.map = L.map(this.el);
+    this.map = L.map(this.el!);
     this.map.scrollWheelZoom.disable();
-    this.tileLayer = L.tileLayer(...this.getTileLayer()).addTo(this.map);
+    const tileLayerOptions = this.getTileLayer();
+    this.tileLayer = L.tileLayer(tileLayerOptions!.urlTemplate, tileLayerOptions!.layerOptions).addTo(this.map);
```
rgbkrk, line 68: It's ok if these are optionally null though for `tileLayer` right?

packages/transform-geojson/src/index.tsx
```diff
@@ -107,10 +114,10 @@ export class GeoJSONTransform extends React.Component<Props> {
         '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>',
       id: `mapbox.${theme}`
     };
-    return [urlTemplate, layerOptions];
+    return { urlTemplate, layerOptions };
   };
 
-  render(): ?React$Element<any> {
+  render(): React.ReactElement<any> {
```
rgbkrk, line 120: We can probably remove the return here and let TypeScript infer it instead. A lot of these got encoded due to an old flow limitation that no longer exists.

rgbkrk (approved):

Issue #1385, opened 2017-01-30
are there plans to support interactive widgets? like ipywidgets?


cornhundred:

rgbkrk:
Thanks for posting. No one is currently working on supporting the current version of the ipython widgets. It's open for someone to work on.

When the version for JupyterLab is more finalized, I'll begin working on integration if no one else gets to it.

cornhundred:
Thanks

paulnakroshis:
Widgets are useful; in testing out NTERACT, I just discovered that they do not work. 
Also bqplot doesn't work in NTERACT; see https://nbviewer.jupyter.org/gist/anonymous/2923cbda7f825eb81a5fc50dc47a33e9
for example (last cell in notebook)

rgbkrk:
If someone is able to support widgets, I'd be happy to have them.

cogmind:
I also support the inclusion of ipywidgets. A "workaround" is to export to HTML:

    `ipywidgets.embed.embed_minimal_html`

By the way, I think it's great that `IPython.display.Markdown` is supported.

spinicist:
I'm another one who would love to see widgets supported!

jdfreder:
I'm looking into this now...

captainsafia:
Commenting to remove stale tag. This is still something we need to do.

rgbkrk:
Added the `pinned` tag so it sticks around.

wadethestealth:
[`@jupyterlab/widgets`](https://github.com/jasongrout/jupyterlab/tree/master/src/widgets) that @rgbkrk mentioned during the meeting. Plus [`ipywidgets` docs](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Low%20Level.html) that explain how they work under the hood. Also link to nteract/hydrogen#27

jasongrout:
The jlab widget manager is at https://github.com/jupyter-widgets/ipywidgets/tree/master/packages/jupyterlab-manager

In addition to the docs, see the controls schema specification at https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema/jupyterwidgetmodels.latest.md and the message spec at https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema/messages.md#widget-messaging-protocol-version-2

jasongrout:
[...]

PR #3642, opened 2018-11-11
Add jupyter-widgets post message RXJS Subject utility.

jdfreder:
This will be used to communicate with the cross domain iframe that will
contain a rendered widget.

### DIFF
--- a/packages/jupyter-widgets/src/communication/message-subject.ts
+++ b/packages/jupyter-widgets/src/communication/message-subject.ts
@@ -1,0 +1,63 @@
+import {Subject, fromEvent} from "rxjs";
+import {filter, first, tap, takeUntil, map} from "rxjs/operators";
+import { Message, POST_MESSAGE_IDENTIFIER, PostMessageType } from "./message";
+
+/**
+ * Posts a message to a Window.
+ * This method does not return anything, it's a send and forget API.
+ * @param destinationWindow
+ * @param message
+ */
+function postMessage<T>(destinationWindow: Window, message: Message<T>) {
+  destinationWindow.postMessage(message, '*');
+}
+
+/**
+ * Makes a Subject which can be used to send and receive post messages.
+ * Note that this posts messages to a single `destinationWindow` but recieves
+ * messages from any window.
+ * @param thisWindow window of the current context.
+ * @param destinationWindow window to post messages to.
+ */
+export function MakePostMessageSubject<T>(
+    thisWindow: Window, destinationWindow: Window): Subject<T> {
+  // Create an observe which will send the correct message given the Observable
+  // event. e.g. an error will trigger an error message to be posted.
+  const outboundMessages = {
+    next: (payload: T) => postMessage<T>(destinationWindow, {
+      id: POST_MESSAGE_IDENTIFIER,
+      type: PostMessageType.REGULAR_MESSAGE,
+      payload,
+    }),
+    error: (error: any) => postMessage<T>(destinationWindow, {
+      id: POST_MESSAGE_IDENTIFIER,
+      type: PostMessageType.ERROR_MESSAGE,
+      error,
+    }),
+    complete: () => postMessage<T>(destinationWindow, {
+      id: POST_MESSAGE_IDENTIFIER,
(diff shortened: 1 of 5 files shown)


jdfreder:
... I also added jest config to the local package so that I can run the tests for this package alone.

jdfreder:
Related to #1385

rgbkrk:
> I also added jest config to the local package so that I can run the tests for this package alone.

To test a single file with jest you can run:

```
yarn test message-subject-spec
```

packages/jupyter-widgets/src/communication/message-subject.ts
```diff
@@ -0,0 +1,63 @@
...
+          throw new Error(message.error);
+        }
+      }),
+      // Complete when a completion message is received.
+      takeUntil(closeMessages.pipe(first())),
+      // Extract the payload of regular messages.
+      map(message => message.payload));
+
+  return Subject.create(outboundMessages, inboundMessages);
+}
```
rgbkrk, line 63: This is a slick little subject setup.

packages/jupyter-widgets/src/communication/message-subject.ts
```diff
@@ -0,0 +1,63 @@
...
+  const closeMessages = inboundPostMessages.pipe(
+      filter((message) => message.type === PostMessageType.COMPLETE_MESSAGE));
+  const inboundMessages = inboundPostMessages.pipe(
+      // Only accept widget post messages.
+      filter(message => message && message.id === POST_MESSAGE_IDENTIFIER),
+      // Throw an error locally if an error message is received.
+      tap(message => {
+        if (message.type === PostMessageType.ERROR_MESSAGE) {
+          throw new Error(message.error);
+        }
```
rgbkrk, line 55: Is it expected that the subscriber of this will have to do their own `catchError` handling?

packages/jupyter-widgets/package.json
```diff
@@ -35,5 +35,25 @@
...
+    "testURL": "http://localhost",
+    "moduleFileExtensions": [
+      "js",
+      "jsx",
+      "ts",
+      "tsx"
+    ],
+    "testMatch": [
+      "**/__tests__/**/*.[tj]s?(x)",
+      "**/?(*.)+(spec|test).[tj]s?(x)"
```
rgbkrk: I think we could probably move these up into the base test config so we can write more tests in typescript.

rgbkrk: Or perhaps you don't have to since it's getting tested as is.

rgbkrk (changes requested):
Please go ahead and remove the `jest` config from the widgets package. We've had a tough time remembering to update internal package setups that we like having just one top level config (as much as possible). `jest` has many ways to have it execute just one test suite for you, happy to point out some ways to make this easier.

packages/jupyter-widgets/src/communication/message-subject.ts
jdfreder: thanks

packages/jupyter-widgets/src/communication/message-subject.ts
jdfreder: For request/reply messages, I anticipate this being required, or a secondary message to communicate the failure. I wanted to make post message API transparent to the Subject's behavior, so that downstream you don't have to worry as much about what is happening upstream of you. 

e.g. Let's say the widget rendering code fails during rendering. The widget rendering normally results in a reaction which sends a resize request to nteract. In the failure case, the error would bubble up and on the nteract side of the fence the observable for resize requests would receive the error. With this, nteract could display an application level error message to the user.

packages/jupyter-widgets/package.json
jdfreder: 👍

jdfreder:
> To test a single file with jest you can run:

Ah thanks, that works great! 

I've removed the jest blurb and squashed per your request.

rgbkrk (approved):

rgbkrk:
Thank you for these bite sized chunks of widgets work. 😄  Been nice to read.

PR #3643, opened 2018-11-11
[data-explorer] Split up index into individual files in components folder

BenRussert:
I saw a discussion about splitting `index.js` up into a components folder so I thought I'd try it out as I read through the package. I think it's easier to follow this way, but would be interested in what you all think.

### DIFF
--- a/packages/transform-dataresource/src/components/Toolbar.js
+++ b/packages/transform-dataresource/src/components/Toolbar.js
@@ -1,0 +1,126 @@
+/* @flow */
+
+import * as React from "react";
+import { DatabaseOcticon } from "@nteract/octicons";
+
+import { chartHelpText } from "../docs/chart-docs";
+import {
+  TreeIcon,
+  NetworkIcon,
+  BoxplotIcon,
+  ScatterplotIcon,
+  LineChartIcon,
+  BarChartIcon,
+  HexbinIcon,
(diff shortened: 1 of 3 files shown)


rgbkrk:
If you run `yarn lint --fix` it'll address the couple of lint errors.

lgeiger:
> If you run yarn lint --fix it'll address the couple of lint errors.

👍 Probably we should add this to our pre-commit hook too.

BenRussert:
Woops, I'll fix those errors later today.

packages/transform-dataresource/src/components/Toolbar.js
```diff
@@ -0,0 +1,126 @@
...
+  BoxplotIcon,
+  ScatterplotIcon,
+  LineChartIcon,
+  BarChartIcon,
+  HexbinIcon,
+  ParallelCoordinatesIcon
+} from "../icons";
+import type View from "../index.js";
+
+import { IconButton } from "./IconButton";
```
BenRussert, line 19: I amended and force pushed to fix the eslint error. The only change was moving the type import above the `IconButton` import and adding an empty line between them.

emeeks (approved):

emeeks:
Thanks for handling this, Ben!

PR #3644, opened 2018-11-11
Migrate presentational components to typescript

jdetle:


packages/presentational-components/package.json
```diff
@@ -3,23 +3,21 @@
   "version": "2.0.0-alpha.0",
   "description": "pure presentational components for nteract",
   "main": "lib/index.js",
-  "nteractDesktop": "src/index.js",
+  "types": "lib/index.d.ts",
+  "nteractDesktop": "src/index.ts",
   "scripts": {
     "prepare": "npm run build",
-    "prepublishOnly": "npm run build && npm run build:flow",
+    "prepublishOnly": "npm run:ild",
```
rgbkrk: In the future you may want to copy the lines directly from the typescript conversion doc in whole, just to make sure we stay consistent.

packages/presentational-components/package.json
```diff
@@ -35,5 +33,9 @@
     "components"
   ],
   "author": "Kyle Kelley <rgbkrk@gmail.com>",
-  "license": "BSD-3-Clause"
+  "license": "BSD-3-Clause",
+  "devDependencies": {
+    "@types/react-syntax-highlighter": "^0.0.8",
+    "@types/styled-jsx": "^2.2.7"
```
rgbkrk: We're supposed to put the types packages in the root level package right?

packages/presentational-components/src/components/cells.tsx
```diff
@@ -1,8 +1,7 @@
-// @flow
 import * as React from "react";
 
-type CellsProps = {
-  children: React.ChildrenArray<any>
+interface CellsProps {
+  children: React.ReactNode[];
```
rgbkrk: We can probably make this one be just a React.ReactNode so that it can also accept a single child.

packages/presentational-components/src/components/header-editor.tsx
```diff
@@ -170,13 +170,13 @@ export class HeaderEditor extends React.Component<
...
                   editable
-                    ? () => {
+                    ? (e: React.MouseEvent<HTMLButtonElement>, props: ITagProps) => {
                         onChange({
                           ...headerData,
                           tags: headerData.tags.filter(p => p !== t)
                         });
                       }
-                    : null
+                    : undefined
```
rgbkrk: This is new to me. I thought React preferred null instead of undefined as the return.

packages/presentational-components/src/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "lib"
+  },
+  "include": ["src"]
+}
```
rgbkrk: We should only need the config at packages/presentational-components

packages/presentational-components/src/components/source.tsx
```diff
@@ -27,7 +26,7 @@ export class Source extends React.Component<SourceProps> {
       return (
         <Highlighter
           language={this.props.language}
-          className={this.props.className}
+          className={this.props.className != null ? this.props.className : ''}
```
rgbkrk: This seems like a good case for using a default prop.

rgbkrk (commented):
I left some comments you might already be addressing since this is WIP. Feel free to ignore if you're already on the case!

jdetle:
This should be good to go. Tested `yarn docs` and it compiled with an error. @rgbkrk  do you have any tips on getting styleguidist to work?

jdetle:
`yarn docs` definitely isn't working anymore.

jdetle:
Fixed it

rgbkrk:
Artifacts are looking good! Well done John!

https://3715-37496521-gh.circle-artifacts.com/0/styleguide/index.html

packages/presentational-components/src/components/prompt.tsx
```diff
@@ -44,10 +24,30 @@ type PromptProps = {
   /**
    * Create a prompt without the `[]`. Used with markdown cells.
    */
-  blank: boolean
+  blank: boolean,
+  /**
+   * Show that execution is currently happening related to this prompt
+   */
```
rgbkrk: This description belongs above `counter`.

rgbkrk (approved):

rgbkrk:
One little comment that needs to be moved, otherwise this is lookin' good!

PR #3646, opened 2018-11-12
Fix SaveAs issue when notebook is unsaved

BenRussert:
- filename has to be updated before save
- use concatMap when order matters


BenRussert:
I ran into a weird bug last night when saving an unsaved notebook. Using concatMap seems to fix the issue. [Here is an explanation](https://www.learnrxjs.io/operators/transformation/concatmap.html#example-1-demonstrating-the-difference-between-concatmap-and-mergemap)

Separate issue: if you try to save as from the alt menu the inital filename is an empty string and it gets messy again.

lgeiger (approved):
Thanks for catching this 👍

PR #3657, opened 2018-11-14
Add default props to transform components to be compatible with new output components in @nteract/outputs

gnestor:
I was playing with the new output components (`Output`, `DisplayData`, `RichMedia`) and although @nteract/outputs exports new "media" components that may be meant to replace the components in @nteract/transforms, the only thing preventing the transforms components from working with these new output components is a default `mediaType` prop. This adds default props to the transform components and also clean up both the transforms and media components to be more consistent with each other and prefer `React.PureComponent` over `shouldComponentUpdate`.


packages/transforms/__tests__/html-spec.js
```diff
@@ -8,17 +8,7 @@ describe("HTMLDisplay", () => {
     const component = mount(<HTMLDisplay data={"<b>woo</b>"} />);
     expect(component.html()).toEqual("<div><b>woo</b></div>");
   });
-  it("correctly chooses to update with data changing", () => {
-    const wrapper = mount(<HTMLDisplay data={"<b>woo</b>"} />);
-
-    const component = wrapper.instance();
-    expect(component.shouldComponentUpdate({ data: "<b>woo</b>" })).toBeFalsy();
```
gnestor, line 15: I can't figure out a good way to test whether a component re-renders or not using enzyme/jest. Now that these components are using `React.PureComponent`, we know that they won't re-render unless the new props or state do not equal the previous. 

> React.PureComponent’s shouldComponentUpdate() only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences. Only extend PureComponent when you expect to have simple props and state, or use forceUpdate() when you know deep data structures have changed. Or, consider using immutable objects to facilitate fast comparisons of nested data.

https://reactjs.org/docs/react-api.html#reactpurecomponent

rgbkrk: Since they're now pure components, we don't have to test the `sCU`.

rgbkrk:
Since I know there's going to be a conflict between #3592, I'm going to leave @captainsafia to review + merge this since we're currently transitioning all the output components to typescript (along with everything else).

captainsafia:
I’ll be able to review this tonight!

packages/outputs/src/components/media/html.js
```diff
@@ -27,12 +27,14 @@ export function createFragment(html: string): Node {
   return fragment;
 }
 
-export class HTML extends React.Component<Props> {
+export class HTML extends React.PureComponent<Props> {
```
captainsafia, line 30: Good call on using PureComponent on these components.

packages/outputs/src/components/media/html.js
```diff
@@ -27,12 +27,14 @@ export function createFragment(html: string): Node {
...
   el: ?HTMLElement;
+
   static defaultProps = {
-    mediaType: "text/html",
-    data: null
+    data: '',
```
captainsafia, line 34: Any particular reason you've changed the default `data` prop to an empty string here and elsewhere? I'm hesitant to introduce a non-nil default value here.

captainsafia (commented):
Thanks for making this change! Good call on unifying the API's for the transform and output components.

One thing I'd like to square away in this PR is standardizing what the `data` prop is set to by default. I'm leaning towards making it an explicit `null` in all cases to avoid shenanigans with truthy-values passed to the component.

packages/outputs/src/components/media/html.js
gnestor: Some were empty strings and some were null, so I decided to go with empty strings/objects because the prop types aren't expecting a null value. Currently the transform components are not doing null checks before rendering, but they could in which case defaulting to `null` would be better. In other words, if `data` defaults to `null` then components should handle null values.

captainsafia: That makes sense.

You can go ahead and merge this. Once our TypeScript migration is complete, we can run through and figure out whether or not we have sensible defaults set on our props.

Thanks! :shipit:

rgbkrk:
Based on your approval @captainsafia I'll go ahead and merge!

packages/outputs/src/components/media/html.js
gnestor: Thanks for reviewing!

PR #3658, opened 2018-11-14
Add styled-jsx as a dependency to packages that depend on it

gnestor:
I've run into issues using certain nteract packages outside of nteract because package may depend on styled-jsx but it doesn't include it in its dependencies. Some packages list it in `peerDependencies` but I don't think this makes sense because it assumes that the app consuming it depends on styled-jsx. Some of these packages may not be intended to use outside of nteract, in which case styled-jsx should be `peerDependencies`.

### DIFF
--- a/packages/connected-components/package.json
+++ b/packages/connected-components/package.json
@@ -34,12 +34,12 @@
     "babel-runtime": "^6.26.0",
     "rc-menu": "6.2.11",
     "react-redux": "^5.0.7",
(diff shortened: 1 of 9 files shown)


rgbkrk:
Alright, let's go for it. I intentionally left it as a peerDep so that people can be on whatever version of next.js or styled-jsx they want to, without the negative effect of multiple styled jsx on the page (mostly a flash of semi-styled content while it resolves).

PR #3659, opened 2018-11-15
Use package name as styleguidist header

captainsafia:
Realized while working on the play.nteract.io tutorial that it would make more sense for these titles to match up with the package names so they are easier to use in reference context.


lgeiger (approved):
I like the new naming 👍

PR #3660, opened 2018-11-15
Quit app when all windows closed on non-mac platforms

### LINKED ISSUE #3656: There is no way to completely close nteract
On linux, when I close a nteract window, the application remains open with an icon in the tray area. This icon doesn't have a "close" option.
If I open another document with nteract, it opens another instance, with another icon in the tray.
The only way I find to close nteract is to force quit it from System Monitor.

I am using Ubuntu 18.10 amd64

BenRussert:
For non-mac platforms:
- If user closes last window, `window-all-closed` is called, followed by `before-quit`.
- If user quits via menu/ctrl-q, `before-quit` is called, each window is closed, then`window-all-closed` fires.

Fixes #3656

### DIFF
--- a/applications/desktop/src/main/index.js
+++ b/applications/desktop/src/main/index.js
@@ -233,6 +233,7 @@ windowAllClosed.pipe(skipUntil(appAndKernelSpecsReady)).subscribe(() => {
     process.platform !== "darwin" ||
     store.getState().get("quittingState") === QUITTING_STATE_QUITTING
   ) {
+    store.dispatch(setQuittingState(QUITTING_STATE_QUITTING));
     app.quit();
   }
 });


applications/desktop/src/main/index.js
```diff
@@ -233,6 +233,7 @@ windowAllClosed.pipe(skipUntil(appAndKernelSpecsReady)).subscribe(() => {
     process.platform !== "darwin" ||
     store.getState().get("quittingState") === QUITTING_STATE_QUITTING
   ) {
+    store.dispatch(setQuittingState(QUITTING_STATE_QUITTING));
     app.quit();
```
jameskoch, line 237: The logic around quitting is really hard to follow, I've been trying to make it more sane, but haven't made great progress. I'm tempted to blame Electron.

Judging by your change, would it be accurate to modify your commit msg to "If user closes last window, `window-all-closed` is called, followed by `before-quit` *which is then canceled because BrowserWindow.getAllWindows() thinks there are still open windows*."? It seems like that might be the key diff between platforms here, and worth documenting a bit.

BenRussert: Wow, I did this quickly last night without realizing that point about the window count. This is pretty confusing.

jdetle:
@BenRussert, what are you thinking on this one? Do we want to merge and iterate, try to polish, or close and revisit later?

BenRussert:
I'll close this for now, I was hoping this was a quick fix, but realized there is more to it and haven't had time to properly debug.

PR #3662, opened 2018-11-15
Remove unused .npmrc files

lgeiger:
We're using `yarn` workspaces to install our dependencies. This makes the `.npmrc` files useless.


rgbkrk:
It does make it explicit what we don't want written out if anyone tries to use `npm` locally. Anywho, worth a clean up since it's not getting used.

PR #3663, opened 2018-11-15
Upgrade immutable.js

lgeiger:
This brings in quite some improvements, including Typescript 3 support: https://github.com/facebook/immutable-js/releases

It has been over one year since RC 9 was released so let's get this updated.

This also removes the deprecation warnings of `@types/immutable`.


rgbkrk:
Super fan here.

PR #3664, opened 2018-11-15
Persist metadata in Data Explorer

emeeks:
* Honor dx metadata
* Fire an onMetadataChange fn when metadata changes
* Rudimentary tests to ensure the component works without metadata


emeeks:
Whoops looks like I forgot to update the test snapshots.

packages/transform-dataresource/src/index.js
```diff
@@ -190,23 +220,21 @@ class DataResourceTransform extends React.Component<Props, State> {
...
-  //CHART CHOICES
-  //DERIVED DATA
-
-  shouldComponentUpdate(): boolean {
-    return true;
+  componentWillMount() {
+    // This is necessary to render any charts based on passed metadata because the grid doesn't result from the updateChart function but any other view does
+    if (this.state.view !== "grid") {
+      this.updateChart(this.state);
+    }
```
BenRussert, line 239: Is there a reason we need grid to always be the default initial view? I wondered in the past if the view should just be a prop with "grid" being default. Updating after mounting will essentially do the same thing, but I'm wondering if I understand why grid has to be the initial view in the constructor.

packages/transform-dataresource/src/index.js
```diff
@@ -20,17 +20,18 @@ type dataProps = {
...
 
+type AreaType = "hexbin" | "heatmap" | "contour";
+
+type SummaryType = "violin" | "joy" | "histogram" | "heatmap" | "boxplot";
+
+type PieceType = "bar" | "point" | "swarm" | "clusterbar";
+
+type HierarchyType = "dendrogram" | "treemap" | "partition";
+
+type NetworkType = "force" | "sankey" | "arc" | "matrix";
```
BenRussert, line 33: I really like having these types declared up here personally.

packages/transform-dataresource/src/index.js
```diff
@@ -42,25 +43,51 @@ export type View =
...
+};
+
+type Props = {
+  data: dataProps,
+  metadata: { dx: dxMetaProps },
+  theme?: string,
+  expanded?: boolean,
+  height?: number,
+  mediaType: "application/vnd.dataresource+json",
+  onMetadataChange?: Function
```
BenRussert: `onMetaDataChange?: ({ dx: dxMetaProps }) => void`, right?

BenRussert (approved):
This is really nice, and we even have tests now! I put some other comments/questions below.

packages/transform-dataresource/src/index.js
emeeks: Oh I like the whole prop with grid as default, I'll add that.

packages/transform-dataresource/src/index.js
emeeks: Yeah, I'm still getting used to the whole "More than me needs to understand this stuff". The "arc" and "matrix" support is coming soon BTW but it's not in there yet.

packages/transform-dataresource/src/index.js
emeeks: Oh good point. Excellent!

PR #3669, opened 2018-11-16
Filters For Columns

emeeks:
* Use `filterable` in react-table
* Add simple string and numeric (<>=) filters
* Hide/Show button

itt_muse_3

### DIFF
--- a/packages/transform-dataresource/src/charts/grid.js
+++ b/packages/transform-dataresource/src/charts/grid.js
@@ -1,32 +1,177 @@
 import * as React from "react";
 import ReactTable from "react-table";
 import withFixedColumns from "react-table-hoc-fixed-columns";
+import { Button, InputGroup, Tooltip } from "@blueprintjs/core";
+import { blueprintCSS, blueprintSelectCSS } from "@nteract/styled-blueprintjsx";
 
 import ReactTableStyles from "../css/react-table";
 
 const ReactTableFixedColumns = withFixedColumns(ReactTable);
 
-export const DataResourceTransformGrid = ({
-  data: { data, schema },
-  height
-}) => {
-  const tableColumns = schema.fields.map(field => ({
-    Header: field.name,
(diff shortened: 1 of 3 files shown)


packages/transform-dataresource/src/charts/grid.js
```diff
@@ -1,32 +1,174 @@
...
+          data={data}
+          columns={tableColumns}
+          style={{
+            height: `${height}px`
+          }}
+          className="-striped -highlight"
+          filterable={showFilters}
+        />
+        <style jsx>{ReactTableStyles}</style>
+        <style jsx>{blueprintCSS}</style>
```
rgbkrk, line 168: I think, though I could be very wrong, that you can put the blueprint CSS at the top level component because of how it's done as global for the top level component. Then you don't have to include it on each of these.

packages/transform-dataresource/src/index.js
```diff
@@ -414,6 +414,8 @@ class DataResourceTransform extends React.Component<Props, State> {
 
     if (view === "grid") {
       display = <DataResourceTransformGrid {...this.props} />;
+      console.log("DataResourceTransformGrid", DataResourceTransformGrid);
+      console.log("display", display);
```
rgbkrk: Probably want to delete these. ✂️

emeeks: Oh whoops!

packages/transform-dataresource/src/charts/grid.js
emeeks: That worked

rgbkrk:
Rebuilding now since CI timed out.

rgbkrk:
Oddly enough, Azure pipelines has been running for > 47 minutes. Something is up in the repo...

rgbkrk:
Rebased this to see if all is passing once synced up with `master`.

PR #3672, opened 2018-11-19
Some data-explorer fixes

BenRussert:
These are some changes to fix various console errors. More info in the commit comments.

### DIFF
--- a/packages/transform-dataresource/src/VizControls.js
+++ b/packages/transform-dataresource/src/VizControls.js
@@ -92,10 +92,9 @@ const renderMenuItem = (item, { handleClick, modifiers }) => {
     <MenuItem
       active={modifiers.active}
       disabled={modifiers.disabled}
-      label={text}
       key={text}
       onClick={handleClick}
-      text={menuText => menuText}
+      text={text}
     />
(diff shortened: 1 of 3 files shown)


emeeks:
I thought doing the state update in componentWillMount rather than componentDidMount meant that we had one fewer render--I didn't realize that setState sometimes doesn't work before a component mounts because I thought the whole idea behind that lifecycle event was to get in one last state update before actual rendering.

Otherwise everything looks great, so I'm going to merge it.

BenRussert:
> I thought doing the state update in componentWillMount rather than componentDidMount meant that we had one fewer render

It does, render will be called twice here. Though the [docs for `componentDidMount`](https://reactjs.org/docs/react-component.html#componentwillunmount) say that this happens before the browser updates the screen, so the user won't notice.

All we would need to eliminate the need for either lifecycle method is to initialize all state in the constructor. I have [a branch where I did this as an example](https://github.com/nteract/nteract/compare/master...BenRussert:example-initializing-all-state-in-constructor), but without some more refactoring I thought this looked worse than just using `componentDidMount`.

PR #3673, opened 2018-11-19
Create raw cells

kelleyblackmore:
Since we can take in raw cells from notebooks, It would be nice to be able to create them as well if needed. Here is my implementation.


kelleyblackmore:
Circle popped up an issue with the changes. I'll fix it.

kelleyblackmore:
I'm gonna close this and submit the new one

PR #3681, opened 2018-11-21
Create @nteract/epics package

captainsafia:


rgbkrk:
Coordinating with Safia now, I'm going to continue on with this PR. I'll be force pushing since there are conflicting files to resolve with a rebase.

package.json
```diff
@@ -128,6 +128,7 @@
     "@babel/preset-typescript": "^7.0.0",
     "@babel/runtime": "^7.0.0",
     "@babel/runtime-corejs2": "^7.0.0",
+    "@types/file-saver": "^2.0.0",
     "@types/immutable": "^3.8.7",
```
rgbkrk: It just dawned on me these types are likely wrong. We're on immutable 4.0.0-rc.12

rgbkrk: Ah we can delete it: https://www.npmjs.com/package/@types/immutable

> This is a stub types definition for Facebook's Immutable (https://github.com/facebook/immutable-js). Facebook's Immutable provides its own type definitions, so you don't need @types/immutable installed!

captainsafia:
It's ready for review!

image

package.json
rgbkrk: This'll still need to be deleted.

packages/actions/src/actions.ts
```diff
@@ -724,7 +724,7 @@ export function executeCanceled(payload: {
 
 export function executeFailed(payload: {
   error: Error;
-  contentRef: ContentRef;
+  contentRef?: ContentRef;
```
rgbkrk, line 727: If this ends up getting to be a messy PR for review, you might want to move this to a new commit on a new branch + PR.

packages/epics/README.md
```diff
@@ -0,0 +1,11 @@
+# `@nteract/epics`
+
+> TODO: description
+
+## Usage
+
+```
+const epics = require('@nteract/epics');
+
+// TODO: DEMONSTRATE API
```
rgbkrk: Likely example to pull from is how desktop stitches together epics.

packages/epics/__tests__/epics.test.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+describe("@nteract/epics", () => {
+  it("needs tests", () => {
+    expect(true).toEqual(true);
```
rgbkrk: I'm assuming you'll bring over the tests for the current epics (assuming we have those...)

packages/epics/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@nteract/epics",
+  "version": "1.0.0",
+  "description": "Epics for interfacing with the Jupyter ecosystem",
```
rgbkrk: "Redux-Observable epics for nteract apps"

packages/epics/src/comm.ts
```diff
@@ -0,0 +1,102 @@
...
+ * An epic that emits comm actions from the backend kernel
+ * @param  {ActionsObservable} action$ Action Observable from redux-observable
+ * @param  {redux.Store} store   the redux store
+ * @return {ActionsObservable}         Comm actions
+ */
+export const commListenEpic = (action$: ActionsObservable<NewKernelAction>) =>
+  action$.pipe(
+    ofType(LAUNCH_KERNEL_SUCCESSFUL),
+    switchMap(commActionObservable)
+  ); // We have a new channel
```
rgbkrk: This comment belongs above the `ofType` operator since it's describing that we launched a new kernel and are `switchMap`ing to kill off old comm listeners.

packages/epics/src/contents.ts
```diff
@@ -0,0 +1,372 @@
...
+  payload: { [error: string]: Error };
+};
+
+export function fetchContentEpic(
+  action$: ActionsObservable<
+    | actions.FetchContent
+    | actions.FetchContentFailed
+    | actions.FetchContentFulfilled
+  >,
+  state$: StateObservable<AppState>
```
rgbkrk, line 26: We may want to make `AppState` a bit loose (maybe this already handles it) so that this epic only has to recieve our base state. Mostly so that any apps that with additional properties can still use this typed epic.

packages/epics/src/contents.ts
```diff
@@ -0,0 +1,372 @@
...
+            // (no dummy or directory)
+            content => content.type === "file" || content.type === "notebook"
+          )
+          .keys()
+      );
+
+      return contentRef$;
+    }),
+    // TODO: Once we're switched to the coming redux observable 1.0.0 release,
+    // we should use the state$ stream to only save when the content has changed
```
rgbkrk: We're now on redux observable 1.0.0 so we can use the `state$` stream. Probably good to update this TODO and possibly leave it as a note instead.

packages/messaging/src/messages.ts
```diff
@@ -252,7 +252,7 @@ export function executeResult(content: {
 /**
  * Creates an error message to indicate when an exception has occurred during
  * code execution.
- * 
+ *
```
rgbkrk, line 255: This is all definitely highlighting that something is consistent about our prettier eslint setup. Not sure what's going on.

packages/messaging/src/messages.ts
```diff
@@ -171,7 +171,7 @@ export function executeRequest(
       ...options
     },
     channel,
-    buffers: []
+    buffers: new Uint8Array()
```
rgbkrk, line 174: Could probably move this out to another PR, as well as any other `messaging` changes, then do an interactive rebase.

rgbkrk (changes requested):
Excellent work getting this migrated!

Since I think this is really close, I wouldn't let my comments about moving some of the changes to a new PR hold this back. Focus on the other ones like removing that invalid `@types/immutable` include.

packages/epics/__tests__/epics.test.js
captainsafia: Yes, we do! Bringing them in now.

packages/epics/package.json
captainsafia: Done.

packages/epics/src/contents.ts
captainsafia: Hm. I guess we can change this type definition to `Partial<AppState>` then. This way, it will accommodate objects that are missing some of the AppState fields and those that have extra fields.

packages/messaging/src/messages.ts
captainsafia: I recently enabled format on save in my VSCode so I think it caught something that hadn't been formatted before.

rgbkrk (approved):
This is looking great, thanks for persevering through this one!

packages/messaging/src/messages.ts
captainsafia: I think it makes more sense to keep it here. The problem stemmed from the fact that our actions definitions had the `Uint8Array` type defined for all buffers. However, our custom type declarations for JMP used `any[]` as the type for buffers. Since `messaging` was converted before the actions and epics, it was set up to match up with the loosely-typed declarations we had. This makes sure that messages and JMP type declarations are using the stricter definitions everywhere.

PR #3680, opened 2018-11-21
Create @nteract/selectors package

captainsafia:
Working on splitting up @nteract/core into smaller TypeScript packages.

PR #3683, opened 2018-11-22
Create @nteract/types package

captainsafia:


rgbkrk:
Luckily it now _is_ a record, todo bot!

captainsafia:
Hurrah! This is ready for review and merge.

I basically split off everything that was in nteract/core/state into this package.

Had to sprinkle some typecasts in the `makeXRecord` functions because Immutable's flow definitions and typescript definitions differ a little bit.

packages/types/src/communication/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import { KernelspecsCommunicationRecordProps } from "./kernelspecs";
+import { makeContentsCommunicationRecord } from "./contents";
+import { makeKernelsCommunicationRecord } from "./kernels";
+import { makeKernelspecsCommunicationRecord } from "./kernelspecs";
+
+export * from "./contents";
+export * from "./kernels";
+export * from "./kernelspecs";
+
+export type CommunicationRecordProps = {
```
jdetle, line 14: are we makings props interfaces or types?

packages/types/src/communication/contents.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type ContentCommunicationRecordProps = {
+  loading: boolean,
+  saving: boolean,
+  error?: object | null;
+};
+
+export const makeContentCommunicationRecord: Immutable.Record.Factory<
+  ContentCommunicationRecordProps
+> = Immutable.Record({
```
rgbkrk: The `Immutable.Record` is a generic, so you can swap this over to:

```js
export const makeContentCommunicationRecord = Immutable.Record<ContentCommunicationRecordProps>({
...
})
```

Same thing for all the below ones as well. This also makes sure our default params match.

rgbkrk: For reference, check out [how commutable does it for outputs now](https://github.com/nteract/nteract/blob/master/packages/commutable/src/outputs.ts#L93).

rgbkrk: Then for all the cases in the rest of this PR where you had to `as` it, you can just rely on the generics in the `Record<T>()` call

captainsafia: Oooooh! Perfect. Thanks.

packages/types/src/communication/index.ts
captainsafia: I was originally going for interfaces, not types because interfaces are extendable. As it turns out, in recent versions of TypeScript, types are also extendable. I just kept them as types here to minimize the diff.

packages/types/src/communication/contents.ts
captainsafia: Done.

packages/types/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+export type CommsRecord = Immutable.RecordOf<CommsRecordProps>;
+
+export const makeCommsRecord = Immutable.Record({
+  targets: Immutable.Map(),
+  info: Immutable.Map(),
+  models: Immutable.Map()
+});
+
+// Pull version from our package.json
+const version: string = require("../package.json").version;
```
rgbkrk, line 67: Now this one is going to be interesting for the long term.

captainsafia: Yeah. I didn't know what to do here. I figured we should just version the types separately in the split-out packages.

rgbkrk:
Any way that we can hook this up to the current `@nteract/core` exports so that this is in use right away?

packages/types/src/index.ts
rgbkrk: I think at least for jupyter extension and desktop we make sure to fill this in, it's just interesting to have our default value here.

captainsafia:
> Any way that we can hook this up to the current @nteract/core exports so that this is in use right away?

Sure! I was originally planning on bringing this in directly to the @nteract/selectors PR but I'll try putting it into the existing code as well.

captainsafia:
Plan: I'm going to merge this now. Then rebase it and bring it into https://github.com/nteract/nteract/pull/3680. Then bring it into our current core package.

PR #3690, opened 2018-11-24
Migrate Outputs to TypeScript

captainsafia:
Closes https://github.com/nteract/nteract/issues/3592.


types/nteract__records/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module "@nteract/records" {
+    export type MediaBundle = any;
+    export type StreamOutput = any;
+    export type OutputType = any;
+}
```
rgbkrk, line 5: Works for now I suppose.

rgbkrk (approved):

types/nteract__records/index.d.ts
captainsafia: Yep. I posted a note in https://github.com/nteract/nteract/pull/3690 that that PR should be rebased once this is merged.

PR #3691, opened 2018-11-24
Migrate Editor to TypeScript

captainsafia:


packages/editor/src/mode/ipython.ts
```diff
@@ -7,7 +6,7 @@ import "codemirror/mode/python/python";
 
 CodeMirror.defineMode(
```
captainsafia, line 7: The code that we have here doesn't match the definition of `defineMode` in the CodeMirror library. We're passing 3 arguments when this function only expects two.

Also, there is no `defineMIME` function in the CodeMirror library.

packages/editor/src/index.tsx
```diff
@@ -86,11 +86,11 @@ class CodeMirrorEditor extends React.Component<
     channels: null
   };
 
-  constructor(props: CodeMirrorEditorProps): void {
+  constructor(props: CodeMirrorEditorProps) {
     super(props);
-    (this: any).hint = this.completions.bind(this);
-    (this: any).tips = this.tips.bind(this);
-    (this: any).deleteTip = this.deleteTip.bind(this);
+    this.hint = this.completions.bind(this);
```
captainsafia: Why is the binding different from the function name here? It is causing some errors in the component.

packages/editor/src/index.tsx
```diff
@@ -233,10 +233,11 @@ class CodeMirrorEditor extends React.Component<
     const completionResults = mergedCompletionEvents.pipe(
```
captainsafia: This is giving me a TypeError that I am not sure how to solve.

```
Argument of type 'OperatorFunction<CodeCompletionEvent, {}>' is not assignable to parameter of type 'OperatorFunction<{}, {}>'.
  Type '{}' is not assignable to type 'CodeCompletionEvent'.
    Property 'editor' is missing in type '{}'. 
```

packages/editor/src/index.tsx
rgbkrk: I think it can be switched since it's internal. I'm guessing it's based on the fact that this used to rely on a different codemirror component and now we use codemirror directly here.

packages/editor/src/mode/ipython.ts
rgbkrk: Uh oh, this is probably a result of old cargo culting from jupyter land. I'm not sure anymore.

packages/editor/src/index.tsx
rgbkrk: According to https://stackoverflow.com/questions/50857298/how-to-use-rxjs6-partition-in-typescript:

> The typings for pipe are currently not flexible enough to support partition as a pipeable operator

packages/editor/src/mode/ipython.ts
rgbkrk: This code has been fairly old, though it looks like @Carreau may have made this API in codemirror much better since https://github.com/codemirror/CodeMirror/issues/3971 and PRs that resulted.

rgbkrk:
I wonder if `defineMIME` was a Codemirror 2 thing: https://codemirror.net/2/doc/manual.html

packages/editor/src/mode/ipython.ts
rgbkrk: From convo (adjust if I have this wrong), it seems the codemirror type definitions are not complete.

captainsafia: Correct. There are `defineMime` functions exported from the CodeMirror codebase but they are not in the type definitions.

packages/editor/src/index.tsx
```diff
@@ -86,13 +86,12 @@ class CodeMirrorEditor extends React.Component<
...
+  constructor(props: CodeMirrorEditorProps) {
     super(props);
-    (this: any).hint = this.completions.bind(this);
-    (this: any).tips = this.tips.bind(this);
-    (this: any).deleteTip = this.deleteTip.bind(this);
+    this.hint = this.hint.bind(this);
+    this.tips = this.tips.bind(this);
+    this.deleteTip = this.deleteTip.bind(this);
     // $FlowFixMe: weirdness in the codemirror API
-    this.hint.async = true;
```
rgbkrk, line 95: We still need this though right?

packages/editor/src/index.tsx
```diff
@@ -215,9 +215,10 @@ class CodeMirrorEditor extends React.Component<
     this.completionSubject = new Subject();
 
     // $FlowFixMe: Somehow .pipe is broken in the typings
-    const [debounce, immediate] = this.completionSubject.pipe(
-      partition(ev => ev.debounce === true)
+    const pipeFunc: any = partition(
+      (ev: CodeCompletionEvent) => ev.debounce === true
     );
+    const [debounce, immediate] = this.completionSubject.pipe(pipeFunc) as any;
```
rgbkrk: The way you can fix this typing is to call partition directly (it takes an observable, as all the pipeable operators do).

Write it as:

```js
const [debounce, immediate] = partition(ev => ev.debounce === true)(this.completionSubject);
```

packages/editor/src/index.tsx
```diff
@@ -230,13 +231,14 @@ class CodeMirrorEditor extends React.Component<
       )
     );
 
-    const completionResults = mergedCompletionEvents.pipe(
-      switchMap((ev: CodeCompletionEvent) => {
+    const completionResults: Observable<any> = mergedCompletionEvents.pipe(
```
rgbkrk: You might no longer have to do this declaration as `Observable<any>` with the code change above.

packages/editor/src/index.tsx
rgbkrk: Yep, this will need to be set again, otherwise codemirror won't have the `callback` property on the event down below.

captainsafia:
I think this is good to go now.

rgbkrk:
Alright, I'll give it a spin!

rgbkrk:
Working perfectly from my testing.

PR #3694, opened 2018-11-24
Add TypeDoc configuration for monorepo

captainsafia:
Huzzah! Finally figured this out.

Here's an example of the deployment for rx-jupyter and commutable: https://packages.nteract.io.

screen shot 2018-11-23 at 10 28 31 pm

We'll need to pass in the name of each package that we'd like to generate docs for to the `package:docs` command and we'll also need to add a `@module package-name` comment to the top of every file in a package but this documents all our functions and types.

### DIFF
--- a/typedoc.js
+++ b/typedoc.js
@@ -1,0 +1,20 @@
+module.exports = {
+  name: "nteract",
+  mode: "modules",
+  out: "nteract-packages/dist",
+  theme: "default",
+  ignoreCompilerErrors: "true",
+  preserveConstEnums: "true",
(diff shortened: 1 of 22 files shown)


rgbkrk (approved):
Oh my gosh, this is super cool. I look forward to having all this put together well. 😄

typedoc.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+  name: "nteract",
+  mode: "modules",
```
rgbkrk, line 3: 😎

PR #3698, opened 2018-11-25
Update transform-vdom to work with vdom's event handlers

gnestor:
This depends on https://github.com/nteract/vdom/pull/76

This adds support for event handlers in transform-vdom. See https://github.com/nteract/vdom/pull/44 for more details on how it works.

Jupyter clients must provide a `onVDOMEvent` prop to `<VDOM>` now with the following signature:

```ts
onVDOMEvent(targetName: string, event: Attributes) => void
```

that sends a comm message to a comm channel given a target name and a body of `JSON.stringify(event)`. 

A reference implementation for JupyterLab: https://github.com/jupyterlab/jupyterlab/pull/5670

### DIFF
--- a/packages/transform-vdom/src/event-to-object.ts
+++ b/packages/transform-vdom/src/event-to-object.ts
@@ -1,0 +1,186 @@
+import * as React from "react";
+
+import { Attributes } from "./object-to-react";
+
+export function serializeEvent<T>(event: React.SyntheticEvent<T>): Attributes {
+  switch (event.type) {
+    // Clipboard events
+    case "copy":
+    case "cut":
+    case "paste":
+      const clipboardEvent = event as React.ClipboardEvent<T>;
+      return { clipboardData: clipboardEvent.clipboardData };
+    // Composition events
+    case "compositionend":
+    case "compositionstart":
+    case "compositionupdate":
+      const compositionEvent = event as React.CompositionEvent<T>;
+      return { data: compositionEvent.data };
+    // Keyboard events
+    case "keydown":
+    case "keypress":
+    case "keyup":
+      const keyboardEvent = event as React.KeyboardEvent<T>;
+      return {
+        altKey: keyboardEvent.altKey,
+        charCode: keyboardEvent.charCode,
+        ctrlKey: keyboardEvent.ctrlKey,
+        key: keyboardEvent.key,
+        keyCode: keyboardEvent.keyCode,
+        locale: keyboardEvent.locale,
+        location: keyboardEvent.location,
+        metaKey: keyboardEvent.metaKey,
+        repeat: keyboardEvent.repeat,
+        shiftKey: keyboardEvent.shiftKey,
+        which: keyboardEvent.which
+      };
+    // Focus events
+    case "focus":
+    case "blur":
+      return {};
+    // Form events
+    case "change":
+    case "input":
+    case "invalid":
+    case "submit":
+      const formEvent = event as React.ChangeEvent<T>;
+      return { value: (formEvent.target as any).value };
+    // Mouse events
+    case "click":
+    case "contextmenu":
+    case "doubleclick":
+    case "drag":
+    case "dragend":
+    case "dragenter":
+    case "dragexit":
+    case "dragleave":
+    case "dragover":
+    case "dragstart":
+    case "drop":
+    case "mousedown":
+    case "mouseenter":
+    case "mouseleave":
+    case "mousemove":
+    case "mouseout":
+    case "mouseover":
+    case "mouseup":
+      const mouseEvent = event as React.MouseEvent<T>;
+      return {
+        altKey: mouseEvent.altKey,
+        button: mouseEvent.button,
+        buttons: mouseEvent.buttons,
+        clientX: mouseEvent.clientX,
+        clientY: mouseEvent.clientY,
+        ctrlKey: mouseEvent.ctrlKey,
+        metaKey: mouseEvent.metaKey,
+        pageX: mouseEvent.pageX,
+        pageY: mouseEvent.pageY,
+        screenX: mouseEvent.screenX,
+        screenY: mouseEvent.screenY,
+        shiftKey: mouseEvent.shiftKey
+      };
+    // Pointer events
+    case "pointerdown":
+    case "pointermove":
+    case "pointerup":
+    case "pointercancel":
+    case "gotpointercapture":
+    case "lostpointercapture":
+    case "pointerenter":
+    case "pointerleave":
+    case "pointerover":
+    case "pointerout":
+      const pointerEvent = event as React.PointerEvent<T>;
+      return {
+        pointerId: pointerEvent.pointerId,
+        width: pointerEvent.width,
+        height: pointerEvent.height,
+        pressure: pointerEvent.pressure,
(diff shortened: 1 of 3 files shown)


packages/transform-vdom/src/index.tsx
```diff
@@ -1,33 +1,30 @@
...
 
-import { objectToReactElement, VDOMEl } from "./object-to-react";
+import { objectToReactElement, VDOMEl, Attributes } from "./object-to-react";
+import { serializeEvent } from "./event-to-object";
 
 interface Props {
   mediaType: "application/vdom.v1+json";
   data: VDOMEl;
-};
+  onVDOMEvent: (targetName: string, event: Attributes) => void;
```
captainsafia: Is there a reason the `event` is typed as `Attributes` and not `React.SyntheticEvent`?

captainsafia (commented):
Looks good to me from a code perspective! I left one question in-line.

packages/transform-vdom/src/index.tsx
rgbkrk: I think we're sending a reduced subset based on https://github.com/nteract/vdom/pull/76, though it does seem that based on the `Attributes` we're losing a lot of nice built in types here.

gnestor: Yes, the `serializeEvent` function from event-to-object.js serializes synthetic events according to [the spec](https://github.com/nteract/vdom/blob/master/docs/event-spec.md). We are losing some good types, so we could do something like this:


```ts
export type SerializedEvent<T> =
  | Partial<React.ClipboardEvent<T>>
  | Partial<React.CompositionEvent<T>>
  | Partial<React.KeyboardEvent<T>>
  | Partial<React.ChangeEvent<T>>
  | Partial<React.MouseEvent<T>>
  | Partial<React.PointerEvent<T>>
  | Partial<React.TouchEvent<T>>
  | Partial<React.UIEvent<T>>
  | Partial<React.WheelEvent<T>>
  | Partial<React.AnimationEvent<T>>
  | Partial<React.TransitionEvent<T>>
  | {};

export function serializeEvent<T>(
  event: React.SyntheticEvent<T>
): SerializedEvent<T> {
  ...
}
```

rgbkrk: Oooooh, I'm loving `Partial`s. This seems like the right kind of thing for jupyter messages that haven't been validated yet too.

packages/transform-vdom/src/event-to-object.ts
```diff
@@ -0,0 +1,200 @@
...
+  | Partial<React.KeyboardEvent<T>>
+  | Partial<React.ChangeEvent<T>>
+  | Partial<React.MouseEvent<T>>
+  | Partial<React.PointerEvent<T>>
+  | Partial<React.TouchEvent<T>>
+  | Partial<React.UIEvent<T>>
+  | Partial<React.WheelEvent<T>>
+  | Partial<React.AnimationEvent<T>>
+  | Partial<React.TransitionEvent<T>>
+  | {};
```
rgbkrk, line 15: 👌

packages/transform-vdom/src/index.tsx
```diff
@@ -49,4 +46,21 @@ export default class VDOM extends React.Component<Props> {
       );
     }
   }
+
+  // Merge event handlers (if any) into attributes as callback functions that
+  // serialize the event object and call `this.props.onVDOMEvent` with the
+  // comm target name and serialized event so that the Jupyter client can
+  // send a comm message for the kernel to handle the event.
+  mergeEventHandlers = (obj: VDOMEl): VDOMEl => {
```
rgbkrk: This seems more like `replaceEventHandlerPlaceholder`.

I think it also belongs in `object-to-react`'s setup, passing in the `onVDOMEvent` function to `object-to-react`.

gnestor: Ok, I moved this to object-to-react and just put it inline with the other property checks (e.g. `if (obj.attributes.style == null) { ... }`. I also moved `arrayToReactChildren` within `objectToReactElement`'s scope so that it can access `onVDOMEvent` arg.

rgbkrk: Ok cool I'll check that out now.

packages/transform-vdom/src/index.tsx
```diff
@@ -1,34 +1,31 @@
 import * as React from "react";
 import { cloneDeep } from "lodash";
 
-import { objectToReactElement, VDOMEl } from "./object-to-react";
+import { objectToReactElement, VDOMEl, Attributes, SerializedEvent } from "./object-to-react";
 
 interface Props {
   mediaType: "application/vdom.v1+json";
   data: VDOMEl;
+  onVDOMEvent: (targetName: string, event: SerializedEvent<any>) => void;
```
rgbkrk, line 9: We need to make this optional for the moment (with checking in the functions that use it) or provide a default prop for `onVDOMEvent`, so it doesn't break the current nteract app.

packages/transform-vdom/src/index.tsx
```diff
@@ -1,34 +1,31 @@
...
-  shouldComponentUpdate(nextProps: Props): boolean {
-    return nextProps.data !== this.props.data;
-  }
-
   render(): React.ReactElement<any> {
     try {
       // objectToReactElement is mutatitve so we'll clone our object
       var obj = cloneDeep(this.props.data);
-      return objectToReactElement(obj);
+      return objectToReactElement(obj, this.props.onVDOMEvent);
```
rgbkrk, line 31: 💯

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -54,13 +61,59 @@ export interface VDOMEl {
...
+  obj: VDOMEl,
+  onVDOMEvent: (targetName: string, event: SerializedEvent<any>) => void
+): React.ReactElement<any> {
+  /**
+   * Convert an array of items to React children.
+   *
+   * @param  {Array} arr - The array.
+   * @return {Array}     - The array of mixed values.
+   */
+  function arrayToReactChildren(
```
rgbkrk: Instead of moving the function in, I think we should just pass the `onVDOMEvent` as the second argument just like is done for `objectToReactElement`

gnestor:
Here's a working implementation in JupyterLab: https://github.com/jupyterlab/jupyterlab/pull/5670

And binder demo: https://github.com/gnestor/jupyterlab/tree/vdom-events-demo/packages/vdom-extension

packages/transform-vdom/src/index.tsx
```diff
@@ -18,7 +18,10 @@ export default class VDOM extends React.PureComponent<Props> {
   static MIMETYPE = mediaType;
 
   static defaultProps = {
-    mediaType
+    mediaType,
+    onVDOMEvent: () => {
+      console.log("This app doesn't support vdom events ☹️ See @nteract/transform-vdom for more info: https://github.com/nteract/nteract/tree/master/packages/transform-vdom");
```
rgbkrk, line 23: Ha, good edition for now.

rgbkrk:
Excellent, let's get this out there. I like that it'll work for now while we wait on https://github.com/nteract/vdom/pull/76

PR #3699, opened 2018-11-25
[WIP] Added a minimal draft.js markdown component

betatim:
Explorations in using draft.js for markdown cells. This gives you WYSIWYG instead of having a place to write markdown "code" that then gets rendered as output.

kapture 2018-11-25 at 21 31 24

Currently this doesn't do much except let you type and mash some buttons to change the style of the text. Need to work out how to pipe in source from a notebook and extract it again. As well as ... lots of other stuff.

My plan is to use https://github.com/Rosey/markdown-draft-js to handle the MD -> draft.js representation and draft.js representation -> markdown. Pointers and thoughts welcome.

Most of the code and CSS comes from https://github.com/facebook/draft-js/tree/master/examples/draft-0-10-0/rich

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,6 +155,7 @@
     "electron": "3.0.7",
     "electron-builder": "20.31.2",
     "electron-context-menu": "^0.10.0",
+    "electron-devtools-installer": "^2.2.4",
     "electron-log": "^2.2.13",
     "electron-updater": "^3.0.0",
     "enzyme": "^3.3.0",
--- a/packages/notebook-app-component/src/markdown-draft.js
+++ b/packages/notebook-app-component/src/markdown-draft.js
@@ -1,60 +1,55 @@
 import React from "react";
-import {Editor, EditorState, RichUtils, convertFromRaw} from 'draft-js';
 
-import { markdownToDraft } from 'markdown-draft-js';
+import {
+  Outputs,
+  PromptBuffer,
+  Input
+} from "@nteract/presentational-components";
 
+import {
+  TrashOcticon,
+  TriangleRightOcticon
+} from "@nteract/octicons";
+
+import { Editor, EditorState, RichUtils, getDefaultKeyBinding,
+         convertFromRaw } from "draft-js";
+
+import { markdownToDraft } from "markdown-draft-js";
 
 class StyleButton extends React.Component {
   constructor() {
     super();
-    this.onToggle = (e) => {
+    this.onToggle = e => {
       e.preventDefault();
       this.props.onToggle(this.props.style);
     };
   }
   render() {
-    let className = 'RichEditor-styleButton';
     if (this.props.active) {
-      className += ' RichEditor-activeButton';
+      //className += " RichEditor-activeButton";
     }
     return (
-      <React.Fragment>
-        <span className={className} onMouseDown={this.onToggle}>
-          {this.props.label}
+        <span className="octicon" onMouseDown={this.onToggle}>
+          <TrashOcticon />
         </span>
-        <style jsx>{`
-          .RichEditor-styleButton {
-            color: #999;
-            cursor: pointer;
-            margin-right: 16px;
-            padding: 2px 0;
-            display: inline-block;
-          }
-
-          .RichEditor-activeButton {
-            color: #5890ff;
-          }
-          `}
-          </style>
-        </React.Fragment>
     );
   }
 }
 
 const BLOCK_TYPES = [
-  {label: 'H1', style: 'header-one'},
-  {label: 'H2', style: 'header-two'},
-  {label: 'H3', style: 'header-three'},
-  {label: 'H4', style: 'header-four'},
-  {label: 'H5', style: 'header-five'},
-  {label: 'H6', style: 'header-six'},
-  {label: 'Blockquote', style: 'blockquote'},
-  {label: 'UL', style: 'unordered-list-item'},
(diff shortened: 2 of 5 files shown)


packages/notebook-app-component/src/notebook-app.js
```diff
@@ -27,6 +27,7 @@ import DraggableCell from "./draggable-cell";
 import CellCreator from "./cell-creator";
 import StatusBar from "./status-bar";
 import MarkdownPreviewer from "./markdown-preview";
+import MyMarkdown from "./markdown-draft";
```
rgbkrk, line 30: Looks like a `git add markdown-draft.js` didn't happen here.

betatim: Oups!

betatim:
Making slowish but steady progress here. I understand how to add new octoicons to use to replace the text, I think I understand the focus/unfocus story. The one thing I am a bit bamboozled by is how to get up-/down-arrow keys to do the right thing. This will take a bit of understanding how to get the cursor position in draft.js I think. Then we can call `focusAbove()` when the up arrow is pressed when at the very top left of the cell.

The one thing I haven't even looked at is how to make it run-time configurable to switch between the old markdown cell and this one. I think having something like that would mean this can get merged earlier/when it isn't yet ready for prime time. Maybe even someone wants to help make it pretty afterwards ;)

This means that my todo list is:
* [ ] get arrow key navigation right
* [ ] run time config of this vs old markdown
* [ ] other things you think I should do (edit me to add please)

It feels like there is a lot to do and it would take forever and three days to do it all in this PR :)

gnestor:
I have used draft js extensively and decided after much frustration to switch to slate js. Slate has a simpler API and there is a more active community around it. There aren't as many plugins for slate in the wild, but I've created quite a few from scratch that I can package up and share. Specifically, I've created a markdown shortcuts plugin that will automatically convert Markdown to rich text as you type. There is a toolbar plugin and there are lots of examples at https://www.slatejs.org.

betatim:
You think it is worth switching? I have no educated opinion on this, I picked draft.js because I found a plugin that seems to do a good job of taking draft.js state and converting it to markdown as well as markdown -> draft.js state.

What I am looking for is https://www.slatejs.org/#/rich-text so that as a user I don't have to know anything about markdown. The "markdown cell" looks just like a friendly, old fashioned WYSIWYG text editor. But because you want notebooks to be compatible we still have to store the cell contents as markdown (hence wanting the roundtrip plugin)

betatim:
So if there is a way to roundtrip markdown I am all for switching especially if you are keen on slate :)

rgbkrk:
I think it's ok to rely on draft.js nowadays, as the community has been better enabled to move it forward.

As for customization, I think we could toggle based on it being in the nteract (desktop) config. You can set a config key with the `setConfigAtKey` action: https://github.com/nteract/nteract/blob/master/packages/core/src/reducers/config.js. It's then in the state tree at `state.config.keyName`.

rgbkrk:
Possible name for the key -- `richMarkdownEditor`, make it a boolean.

gnestor:
As long as the use case is simple and doesn't require must custom behavior, then draft is fine. If you want to do very custom things with the editor, then I think switching to slate is worth it.

betatim:
I've spent a few hours trying to get the whole focus/unfocus thing to work but it only kinda works :-/ Going through the cell from below doesn't properly focus a cell above. I've given up on trying to work out why :)

If someone else can spot what it is I am missing that would be great. Also feel free to take over the PR as I probably won't work on this again for the next two weeks. My react-fu isn't strong enough :(

package.json
```diff
@@ -155,6 +155,7 @@
     "electron": "3.0.7",
     "electron-builder": "20.31.2",
     "electron-context-menu": "^0.10.0",
+    "electron-devtools-installer": "^2.2.4",
```
betatim, line 158: Remove me again. Tried to get the devtools installed/opened. This used to "just work" but now I think you'd also have to call the devtools from somewhere to make it work.

packages/notebook-app-component/src/markdown-draft.js
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+  render() {
+    return (
+      <React.Fragment>
+        <Outputs>
+          <div
+            className="RichEditor-root"
+            onDoubleClick={this.openEditor}
+            onKeyDown={this.editingKeyDown}
```
betatim: With or without this I don't seem to be able to receive any key events when you just single click on a markdown cell. Needed so that arrow up/down and "return to open editor" work.

rgbkrk:
Would you like to join the nteract meetings and demo this on Monday for everyone?

betatim:
That would be cool and could work if I can arrive a bit (10minutes) late. I have to get off a train right when the meeting starts.

rgbkrk:
My thinking on this is that this:

* gets enabled on a config option via our `nteract.json` file that's in the jupyter config dir
* only ships with desktop (as draft.js is fairly heavy).

As it stands though, this now requires a refactor to get it in line with the TypeScript setup. Either a rebase or a new PR that re-does it. Sorry @betatim! I'm going to go ahead and close the PR to be re-done or rebased. Let me know if you'd like to proceed (and if you'd want to maintain this).

betatim:
That's fine. I'll return to this at some point but if someone else wants to work on this: please do!

PR #3703, opened 2018-11-26
WIP fake opaque types for {Host|Kernel|Session|Cell}Id

rgbkrk:
Following the post @Deteam linked within [a tweet thread about these types](https://twitter.com/TimurAmirov/status/1066979897250390016), this implements what we previously had as  "opaque type"s in flow and is called a newtype in Haskell.

This takes the intersection type approach so that current code matches up well with little change (had to fix one type in the selectors package).


rgbkrk:
Closing, I'll revisit this some other day when we really want this level of safety.

PR #3706, opened 2018-11-27
Update TSDoc strings on commutable

captainsafia:


rgbkrk:
Oh, I just noticed this one is good to go. Merging! Thanks for showing the way for documenting these. I was getting confused by VSCode suggesting to remove JSDoc types so I've been deleting them when it looks like I should have been adapting them to this format.

Issue #3713, opened 2018-11-27
TypeScript Monorepo Regressions


rgbkrk:
I've noticed that TypeScript is resolving based on built packages rather than what's in `src`. For example, I changed `@nteract/commutable` to export `ImmutableCell` so that I could rely on it in `@nteract/selectors`:

<img width="780" alt="screen shot 2018-11-27 at 3 34 59 pm" src="https://user-images.githubusercontent.com/836375/49118657-0b939580-f25a-11e8-9983-ff933cd8e798.png">

The message above makes it pretty clear there too, it's resolving from `commutable/lib/index`. With flow we had a mapping to have it resolve from `src` when in monorepo land. Is there a way for us to handle that in typescript so you don't have to (re)build packages?

willingc:
@rgbkrk Can this be closed now?

rgbkrk:
I think we're good here now. Thanks @willingc!

PR #3714, opened 2018-11-28
Resolve packages for typescript via src

rgbkrk:
I was going to refactor the `selectors` module to make sure to use generics on Immutable data structures when I ran into #3713. So long as you don't run `yarn build`, you'll see an issue in the typescript setup of `ImmutableCell` not being an exported member of commutable. If you build, it will be available. This is because typescript is resolving based on what's in `lib`, because that's what our `main` points to in each `package.json`. When we were on flow, we remapped the locations like this: https://github.com/nteract/nteract/blob/e9fa8be823b92dc36f2d1b3624d181c30f55620f/.flowconfig#L8

I'd love to do the same.

### DIFF
--- a/packages/transforms/src/index.tsx
+++ b/packages/transforms/src/index.tsx
@@ -9,9 +9,7 @@ import LaTeXDisplay from "./latex";
 import SVGDisplay from "./svg";
 import { PNGDisplay, JPEGDisplay, GIFDisplay } from "./image";
 
-type Transform = {
-  MIMETYPE: string;
-};
+type Transform  = {MIMETYPE: string} & React.ComponentType
 
 type ObjectType = object;
 export interface Transforms extends ObjectType {
--- a/packages/transforms/src/index.tsx
+++ b/packages/transforms/src/index.tsx
@@ -9,7 +9,7 @@ import LaTeXDisplay from "./latex";
 import SVGDisplay from "./svg";
 import { PNGDisplay, JPEGDisplay, GIFDisplay } from "./image";
 
-type Transform  = {MIMETYPE: string} & React.ComponentType
+type Transform = { MIMETYPE: string } & React.ComponentType;
 
 type ObjectType = object;
 export interface Transforms extends ObjectType {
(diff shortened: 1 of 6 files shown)


rgbkrk:
```
../transform-vdom/src/object-to-react.ts(86,3): error TS2322: Type 'string' is not assignable to type 'never'.
../transform-vdom/src/object-to-react.ts(87,3): error TS2322: Type 'Attributes' is not assignable to type 'never'.
../transform-vdom/src/object-to-react.ts(95,9): error TS2322: Type 'null' is not assignable to type 'never'.
../transform-vdom/src/object-to-react.ts(99,7): error TS2322: Type 'string' is not assignable to type 'never'.
../transform-vdom/src/object-to-react.ts(101,7): error TS2322: Type 'ReactElement<any>' is not assignable to type 'never'.
../transform-vdom/src/object-to-react.ts(129,19): error TS2345: Argument of type 'ReactNodeArray' is not assignable to parameter of type 'never'.
../transform-vdom/src/object-to-react.ts(144,19): error TS2345: Argument of type 'ReactElement<any>' is not assignable to parameter of type 'never'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

These look like things @gnestor fixed up in #3698

rgbkrk:
Oh wait, the above error is happening when `@nteract/display-area` is built, which never uses that code directly.

charlesjolley:
Did you add an index.ts to the package root? Unless we changed processcthats how I got around the build issue.

gnestor:
@rgbkrk Anything I do here?

rgbkrk:
No process changed, I was hoping to figure out a way to make packages resolve from `src` when we're in the monorepo (so we don't have to build each individually).

rgbkrk:
> @rgbkrk Anything I do here?

I'm assuming you're saying "Anything I can do here", in which case I'll ping you when I'm stuck again.

Apologies for everyone being added as a reviewer before this was really ready. I didn't realize what else would start failing.

packages/transforms/src/index.tsx
```diff
@@ -38,7 +36,7 @@ const tfs = [
   TextDisplay
 ];
 
-export const standardTransforms: Transforms = {};
+export const standardTransforms: any = {};
```
rgbkrk, line 39: This seems like a bit of a backwards step.

PR #3715, opened 2018-11-28
Change styleguide config.

jdetle:
@captainsafia showed me how to change the styleguide config. This adds the google analytics tag.


captainsafia (approved):
Thanks!

PR #3717, opened 2018-11-28
Convert nteract actions to typescript

rgbkrk:
* [x] Perform cursory conversion
* [x] Hook up the old tests
* [x] Remove no longer valid TODOs and FlowFixMes

### DIFF
--- a/packages/actions/__tests__/actions-spec.js
+++ b/packages/actions/__tests__/actions-spec.js
@@ -1,0 +1,513 @@
+// @flow
+import * as actions from "../src/actions";
+import * as actionTypes from "../src/actionTypes";
+
+describe("setLanguageInfo", () => {
+  test("creates a SET_LANGUAGE_INFO action", () => {
+    const langInfo = {
+      codemirror_mode: { name: "ipython", version: 3 },
+      file_extension: ".py",
+      mimetype: "text/x-python",
+      name: "python",
+      nbconvert_exporter: "python",
+      pygments_lexer: "ipython3",
+      version: "3.5.1"
+    };
+
+    expect(actions.setLanguageInfo({ langInfo })).toEqual({
+      type: actionTypes.SET_LANGUAGE_INFO,
+      payload: { langInfo }
+    });
+  });
+});
+
+describe("unhideAll", () => {
+  test("allows being called with sets defaults for outputHidden and inputHidden", () => {
+    expect(
+      actions.unhideAll({ outputHidden: true, inputHidden: false })
+    ).toEqual({
+      type: actionTypes.UNHIDE_ALL,
+      payload: {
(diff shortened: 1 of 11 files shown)


rgbkrk:
CircleCI had a partial outage:

> Partially Degraded Service - components affected: CircleCI 2.0, macOS 2.0 Builds
> Updated 12 min ago - 2.0 Jobs not running: Service has been restored and we are processing the backlog of queued jobs. Our engineers are monitoring the system for performance and stability. We will provide an update

packages/actions/__tests__/actions-spec.js
```diff
@@ -0,0 +1,513 @@
+// @flow
```
captainsafia: I've been converting the tests to TypeScript as well so we might as well do that here too.

packages/actions/src/actionTypes.ts
```diff
@@ -0,0 +1,872 @@
+import {
+  CellID,
+  CellType,
+  MimeBundle,
+  Output,
+  JSONObject,
+  ImmutableJSONType
```
captainsafia, line 7: Some of these things are defined in `@nteract/types`. Do we want to import them from there to showcase that they are not unique to commutable?

packages/actions/src/actionTypes.ts
```diff
@@ -0,0 +1,872 @@
...
+  RemoteKernelProps
+} from "@nteract/types";
+
+export type ErrorAction<T extends string> = {
+  type: T;
+  payload: Error;
+  error: true;
+};
+
+export const OPEN_MODAL = "CORE/OPEN_MODAL";
```
captainsafia, line 28: You can choose to address this in this PR or in a follow-up but I think it would be very helpful if we split these actions up into different files based on what they are for.

captainsafia (approved):
Looks great! Left some comments inline.

packages/actions/__tests__/actions-spec.js
rgbkrk: This is done now, mostly thanks to @CeaMonster.

packages/actions/src/actionTypes.ts
rgbkrk: Since I realize this is going to conflict with my work in #3718, I'm going to go ahead with this.

packages/actions/src/actionTypes.ts
rgbkrk: Definitely a follow up since I think we shouldn't be doing too many refactors amidst the typescript refactor, especially since they're not hooked in the main apps yet.

rgbkrk:
I'm going to merge this one so that rebases aren't too scary for other PRs.

Issue #3654, opened 2018-11-14
Commuter does not run on macOS


nikhilkuria:
I followed the instructions to install it 
`yarn global add @nteract/commuter` 

Afterwards, when i try to run it using 
`COMMUTER_LOCAL_STORAGE_BASEDIRECTORY=<~/dev/notebooks> commuter`

it fails with the following error
```
CONFIGURED DISCOVERY none
internal/modules/cjs/loader.js:589
    throw err;
    ^

Error: Cannot find module '../../babel.flow.config'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:587:15)
    at Function.Module._load (internal/modules/cjs/loader.js:513:25)
    at Module.require (internal/modules/cjs/loader.js:643:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/nkose/.config/yarn/global/node_modules/@nteract/webpack-configurator/index.js:12:25)
    at Module._compile (internal/modules/cjs/loader.js:707:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:718:10)
    at Module.load (internal/modules/cjs/loader.js:605:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:544:12)
    at Function.Module._load (internal/modules/cjs/loader.js:536:3)
```

DemonR:
I tried both npm and yarn install, they all produce the same error. It only works when starting as development version from the repo

andriuspl:
Yes, same here.

willingc:
Thanks folks for the reports. We will investigate this along with #3666 when contributors are available.

Development version appears to work per @DemonR.

weldpua2008:
Same 
```
# NODE_ENV=production COMMUTER_STORAGE_BACKEND=s3 COMMUTER_BUCKET=ci /usr/local/nodejs-binary-8.15.0/bin/commuter server
CONFIGURED DISCOVERY none
module.js:550
    throw err;
    ^

Error: Cannot find module './config/babel.flow.config'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/usr/local/nodejs-binary-8.15.0/lib/node_modules/@nteract/commuter/node_modules/@nteract/webpack-configurator/index.js:12:25)
    at Module._compile (module.js:653:30)
[...]

PR #3728, opened 2018-11-29
Copy config to package lib directory on build

jdetle:
I think this should fix the commuter config issue.

### DIFF
--- a/packages/webpack-configurator/copy_config.sh
+++ b/packages/webpack-configurator/copy_config.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+mkdir config
+cp ./../../babel.config.js ./config/babel.config.js
+cp ./../../babel.flow.config.js ./config/babel.flow.config.js
+cp ./../../babel.typescript.config.js ./config/babel.typescript.config.js
--- a/packages/webpack-configurator/package.json
+++ b/packages/webpack-configurator/package.json
@@ -3,8 +3,12 @@
   "version": "2.0.0-alpha.0",
   "main": "index.js",
   "description": "a common webpack configurator for nteract applications",
+  "files": ["./config/babel.config.js", "./config/babel.flow.config.js", "./config/babel.typescript.config.js"],
   "scripts": {
-    "test": "echo 'write a basic integration test please' && node -e 'process.exit(1)'"
+    "test": "echo 'write a basic integration test please' && node -e 'process.exit(1)'",
+    "copyConfig": "bash ./copy_config.sh",
+    "build": "yarn copyConfig",
+    "prepublish": "yarn copyConfig"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/webpack-configurator/config/babel.config.js
+++ b/packages/webpack-configurator/config/babel.config.js
@@ -1,0 +1,35 @@
+module.exports = function(api) {
+  if (api) {
+    const env = api.env(); // eslint-disable-line no-unused-vars
+  }
+
+  const config = {
+    presets: ["@babel/preset-env", "@babel/preset-react"],
+    plugins: [
+      "styled-jsx/babel",
+      "@babel/plugin-syntax-dynamic-import",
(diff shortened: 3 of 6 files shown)


jdetle:
I think I would need to compile this config to get it to work, which I'm not sure makes sense.

jdetle:
Taking off the WIP tag to let someone know they should look at it, but this is a gross hack that I'm not even sure works. If someone could help me publish a test commuter and configurator update, I would be appreciative.

chrnorm:
@jdetle I'm a quick look at this to see if I can help, I'm not too familiar with the nteract codebase yet but here are some things that I noticed:

- The project-wide `babel.config.js` [actually has overrides for flow and typescript already (see hyperlink)](https://github.com/nteract/nteract/blob/024e7960aec8262fbb03112463a1987602cc244f/babel.config.js#L21-L31), if the overrides work properly there is probably no need for separate `babel.flow.config.js` and `babel.typescript.config.js` files.
- The babel docs specify that for a monorepo `babel.config.js`, you can pass `rootMode: "upward"` as an option to Webpack which will make babel find the config file in the root directory without the need for `require`ing any config files ([more info here](https://babeljs.io/docs/en/config-files#webpack)).

If we added `rootMode: "upward"` to the webpack config and just used `babel.config.js` instead of the two typescript/flow config files, it might solve the problem of finding the config file.

I can open a PR with these changes to see if it would help, but did you have any idea how to test them? As mentioned in #3654, when starting commuter from the git repo the bug isn't there.

jdetle:
Hi @chrnorm, thanks for your comment. I think you're on to something with respect the first bullet point. If you can open a PR that successfully removes extraneous config that would be warmly welcomed.  The second bullet point seems like it won't actually work for fixing the issue at hand, which is that when the package is built and shipped, the root level config files can't exist by design.

There are a few issues that are getting in the way of having a clean environment to test this. We should be able to build a production version of commuter with the production version of all our local packages that it depends on and it should "just work", but this typescript conversion seems to be causing some issues. I have some work to get to before I can really deep dive on assisting in recovery efforts. 

With respect to the commuter config issues: I think using `files` in the package.json of webpack-configurator with the webpack config paths in the package.json is probably a better solution than all of my work that I put in here.

packages/webpack-configurator/copy_config.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+mkdir lib
+cp ./../../babel.flow.config.js ./lib/babel.flow.config.js
+cp ./../../babel.typescript.config.js ./lib/babel.typescript.config.js
```
rgbkrk: This seems reasonable as a step to perform.

packages/webpack-configurator/package.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "@nteract/webpack-configurator",
   "version": "2.0.0-alpha.0",
-  "main": "index.js",
+  "main": "lib/index.js",
   "description": "a common webpack configurator for nteract applications",
+  "files": ["../../babel.flow.config", "../../babel.flow.config"],
```
rgbkrk: I don't think it'll be possible to grab and drop a file to a relative directory below. With the copy config above, have it run in a `prepare` script.

rgbkrk: Oh I just realized this is in the webpack configurator...

packages/webpack-configurator/copy_config.sh
jdetle: I'm thinking that I should delete this with the `files` portion in package.json, but I'm pretty out of my element here.

packages/webpack-configurator/package.json
jdetle: I need to look into this more.

jdetle:
@rgbkrk, @lgeiger I think this is ready to go, but I'm still not confident on how to test the distributed commuter app locally.

rgbkrk:
I'll go ahead and bring this in, then we can figure out a good integration test.

PR #3729, opened 2018-11-29
Add READMEs for markdown and iron-icons packages

chrnorm:
I added basic READMEs with install instructions to `@nteract/markdown` and `@nteract/iron-icons`, to tackle some of #1577.

For the `markdown` package, I renamed @captainsafia's great `examples.md` file to be the README and added install instructions to it - so more people will see it on github and npm.

- [x ] I have read the [Contributor Guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md)

PR #3730, opened 2018-11-30
Fix imports in Media components

captainsafia:
Follow-up on https://github.com/nteract/nteract/pull/3729.

Fixes up some import errors in Media components so that we can do a redeploy.


lgeiger (approved):
Thanks for the catch 👍

PR #3734, opened 2018-12-02
Set up docs for @nteract/actions

captainsafia:
I think the types and function names combined with the way TypeDoc structures this are good enough documentation ATM.

### DIFF
--- a/packages/actions/src/actions/index.ts
+++ b/packages/actions/src/actions/index.ts
@@ -1,7 +1,7 @@
 /**
  * @module actions
  */
-import { CellID, MimeBundle, Output, JSONObject } from "@nteract/commutable";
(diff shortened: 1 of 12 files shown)


packages/actions/src/actions/index.ts
```diff
@@ -0,0 +1,240 @@
...
+  ContentRef,
+  HostRef,
+  KernelRef,
+  LanguageInfoMetadata
+} from "@nteract/types";
+
+export * from "./cells";
+export * from "./contents";
+export * from "./kernels";
+export * from "./kernelspecs";
```
rgbkrk, line 18: Thank you for grouping all of these.

rgbkrk (approved):
Looking great, I say go ahead and merge this after rebase.

PR #3738, opened 2018-12-03
Add README for @nteract/actions

captainsafia:
Starting to work on adding READMEs to match our [template](https://content-guide.nteract.io/#/templates).


rgbkrk (approved):

packages/actions/README.md
```diff
@@ -0,0 +1,49 @@
...
+$ yarn add @nteract/actions
+```
+
+```
+$ npm install --save @nteract/actions
+```
+
+## Usage
+
+The Redux reducer below shows how we can leverage the actions and action types in this package to create a reducer for managing the status of a save event.
```
rgbkrk, line 17: I'm somewhere between this being too verbose and maybe not enough (no use of dispatch with an action creator).

rgbkrk (approved):

packages/actions/README.md
captainsafia: Got it. In the future, we'll hopefully be able to link the @nteract/play tutorial to show people how to use the package in more full-fledged situations.

PR #3740, opened 2018-12-03
Add README for @nteract/messaging

captainsafia:
I also went ahead and removed some unneeded babel dependencies from the package.json while I was at it.


packages/messaging/README.md
```diff
@@ -0,0 +1,39 @@
...
+```javascript
+import { createMessage } from "@nteract/messaging";
+
+const message = createMessage("inspect_request", {
+  code: "string.for",
+  cursor_pos: 10,
+  detail_level: 1
+});
+```
+
```
rgbkrk, line 28: So great to have this package documented.

rgbkrk (approved):
Thank you!

PR #3743, opened 2018-12-03
Update README for @nteract/commutable to match format

captainsafia:
Updates the README to match the new format we are using throughout our packages.

### DIFF
--- a/packages/commutable/README.md
+++ b/packages/commutable/README.md
@@ -2,7 +2,7 @@
 
 This is a package for Jupyter Notebook operations, helping to enable history stored as a series of immutable notebooks.
 
-This package follows the principles below, based on [Tom MacWright](http://www.macwright.org/2015/05/18/practical-undo.html)'s outline, for immutable data.
+This package follows the principles below, based on [Tom MacWright](http://www.macwright.org/2015/05/18/practical-undo.html)'s outline for practical undo.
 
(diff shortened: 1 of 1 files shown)


packages/commutable/README.md
```diff
@@ -1,32 +1,48 @@
...
 
-> /kəˈmyo͞otəbəl/
->
-> 1. (of a place or journey) allowing regular commuting to and from work.
-> 2. **capable of being exchanged or converted.**
+This is a package for Jupyter Notebook operations, helping to enable history stored as a series of immutable notebooks.
 
-`commutable` is a library for Jupyter Notebook operations, helping to enable
-history stored as a series of immutable notebooks.
+This package follows the principles below, based on [Tom MacWright](http://www.macwright.org/2015/05/18/practical-undo.html)'s outline, for immutable data.
```
rgbkrk: "for allowing for practical undo"

I realize the absurdity of calling this out since we don't have it. 😢

captainsafia: Yeah. That's why I was meh about using that phrase.

Ah well. Maybe we can document it into existence?

rgbkrk: > Maybe we can document it into existence?

Oooh good plan. I mean this does lay down the foundations so we don't have mutable gunk leading to absolutely no undo.

PR #3760, opened 2018-12-04
Add README for @nteract/transform-vega

captainsafia:


rgbkrk (approved):
Thanks for adding examples to each of these along the way!

PR #3764, opened 2018-12-05
Sticky Header

stormpython:
Adds sticky header behavior.

sticky-header


stormpython:
The `NotebookMenu` dropdown is broken. This is probably because they are positioned `absolute`.

stormpython:
This PR provides a sticky header and the functionality works. However, there is some refactoring needed around the layout of the header and the body (div with cells).

I will open another issue for that and link to it from here.

packages/connected-components/src/notebook-menu/styles.tsx
```diff
@@ -17,6 +17,10 @@ const Wrapper = (props: Props) => (
           border: 1px solid var(--theme-app-border);
           box-shadow: none;
           color: var(--theme-menu-fg);
+          position: fixed;
+          top: 40px;
```
rgbkrk: This 40px is the title header above I assume?

packages/presentational-components/src/components/sticky-header.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <header>
+        {this.props.children}
+        <style jsx>{`
+          header {
+            position: fixed;
+            top: 0;
+            left: 0;
+            width: 100%;
+            // TODO: Need to remove this once header is in a separate div
+            z-index: 2;
```
rgbkrk, line 19: Can this be removed now?

packages/connected-components/src/notebook-menu/styles.tsx
stormpython: Yes, I can change this so that is not needed, but I think it's best to make that change to the layout when I work on the layout for the header and the body.

packages/presentational-components/src/components/sticky-header.tsx
stormpython: No, not until I separate the header from the body.

rgbkrk:
Since there's a dependency on the items being a certain height within the sticky header, I think we shouldn't put it in the presentational components (yet).

packages/connected-components/src/notebook-menu/styles.tsx
stormpython: ok fixed it!

stormpython:
@rgbkrk, I removed the hard coded top positioning, can you take a look again?

rgbkrk:
Yeah, let's get this in and go ahead with iterating.

PR #3773, opened 2018-12-05
Migrate @nteract/notebook-app-component to TypeScript

captainsafia:
Closes https://github.com/nteract/nteract/issues/3574.


packages/actions/index.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * @module actions
+ */
+
+export * from "./src";
```
rgbkrk: It seems awfully suspect that there's an index.ts in the root, are you sure this is being resolved here? Wouldn't it be using `main`?

captainsafia: Hang on! I think this might have been an artifiact from me trying different things to fix the error that ended up being a result of having a stale lib directory.

rgbkrk:
Nice! Is this running well on both desktop and jupyter-extension too?

captainsafia:
- [x] `yarn test:flow` passes
- [x] `yarn test` passes
- [x] `yarn build:packages` passes

Test on desktop and Jupyter extension
- [x] Can drag and drop cell
- [x] Can execute cell
- [x] Can change kernel
- [x] Can create and delete cells

PR #3779, opened 2018-12-05
Move @nteract/core/dummy to its own TypeScript package

captainsafia:
I moved it into its own package and converted to TypeScript.

Also renamed it.

This needs to be reviewed and merged after the reducers are pulled out of core.


rgbkrk:
Fixtures! Thanks for renaming this!

PR #3783, opened 2018-12-06
Switch to react-timeago

jdetle:


packages/timeago/README.md
```diff
@@ -1,5 +1,8 @@
 ![TimeAgo](http://naman.s3.amazonaws.com/react-timeago.png)
 
+## Warning:
```
rgbkrk, line 3: Probably want to point out folks should use `react-timeago` instead.

rgbkrk (approved):

PR #3812, opened 2018-12-08
Update @nteract/core to use TypeScript packages

captainsafia:
Relies on https://github.com/nteract/nteract/pull/3779 being merged.


rgbkrk:
Merged #3779, looking forward to this!

package.json
```diff
@@ -77,7 +77,6 @@
     ],
     "moduleNameMapper": {
       "^@nteract/styles$": "<rootDir>/packages/styles/lib",
-      "^@nteract/core/dummy$": "<rootDir>/packages/core/src/dummy",
```
rgbkrk, line 80: Glad to be rid of this!

applications/desktop/__tests__/renderer/epics/kernel-launch-spec.js
```diff
@@ -1,5 +1,5 @@
 import { ActionsObservable } from "redux-observable";
-import { actions, actionTypes, makeStateRecord } from "@nteract/core";
+import { actions as actionsModule, makeStateRecord } from "@nteract/core";
```
rgbkrk, line 2: Since everywhere else uses this as actions we should do that here too.

packages/core/__tests__/config-spec.js
```diff
@@ -1,30 +0,0 @@
...
-    const initialState = new Map();
-
-    const config = { theme: "dark" };
-    const state = reducers(initialState, {
-      type: actionTypes.MERGE_CONFIG,
-      config
-    });
-    expect(state.get("theme")).toBe("dark");
-  });
-});
```
rgbkrk, line 30: Did each of these specs not get git mv'ed?

Oh wait they're in core so they got copied from the other PRs into individual packages.

rgbkrk (approved):

applications/desktop/__tests__/renderer/epics/kernel-launch-spec.js
captainsafia: I renamed this one because there was a collision in variables [here](https://github.com/nteract/nteract/pull/3812/files/42ef382e1b589c6d9110c2c925e0c3c421948ea2#diff-8db203c4826c0907a533e3ef6f5d1c1eL121).

rgbkrk:
Taking over this PR to push it through to completion. This pretty well finishes off the main typescript fun. 😅

PR #3814, opened 2018-12-09
Remove margin, remove position fixed

jdetle:


stormpython:
@jdetle, this change stops the header from sticking to the top of the page when scrolling down the notebook.

jdetle:
Ah right okay. This is a little vexing. We could try using flexbox? What do you think?

stormpython:
Yes. I think flexbox is the right solution for this problem.

PR #3818, opened 2018-12-10
stop running package builds in parallel

rgbkrk:
Currently our setup can lead to running out of memory (empirically, on both mine and @captainsafia's laptops).


captainsafia (approved):
My CPU thanks you. 🙏🏽

PR #3825, opened 2018-12-11
Update url-search-params dependency

captainsafia:
Closes https://github.com/nteract/nteract/issues/3813.


captainsafia:
Mmmmm. Nevermind. This deprecation looks kinda fishy.

PR #3827, opened 2018-12-11
Update types on executeCell action creator

captainsafia:
Closes https://github.com/nteract/nteract/issues/3739.

### DIFF
--- a/packages/actions/src/actions/kernels.ts
+++ b/packages/actions/src/actions/kernels.ts
@@ -10,7 +10,7 @@ import {
 } from "@nteract/types";
 
 import * as actionTypes from "../actionTypes";
-import { ExecuteRequest } from "@nteract/messaging/src";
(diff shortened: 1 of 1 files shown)


packages/actions/src/actions/kernels.ts
```diff
@@ -10,6 +10,7 @@ import {
 } from "@nteract/types";
 
 import * as actionTypes from "../actionTypes";
+import { ExecuteRequest } from "@nteract/messaging/src";
```
rgbkrk: This shouldn't import from src

rgbkrk (changes requested):

packages/actions/src/actions/kernels.ts
captainsafia: Bleh! Good catch. VSCode's autocompleter pulled it from here. I wonder why...

rgbkrk (approved):

PR #3828, opened 2018-12-11
Add type to kernelspec in newNotebook action creator

captainsafia:
Closes https://github.com/nteract/nteract/issues/3724.


rgbkrk (approved):
Thank you for these clean ups!

PR #3833, opened 2018-12-11
bump required engine to match electron-builder

rgbkrk:
With the new electron builder in place, it appears the required node version is higher. Our repo should follow suit.


jdetle:
I caused this error, sorry about that, don't know how it happened.

rgbkrk:
No prob, it's cool. It's just a warning.

PR #3837, opened 2018-12-12
remove old types

rgbkrk:
Looks like some of these type stubs got created during our typescript migration and need to be deleted since (most of) our packages are properly typescript.


captainsafia (approved):

rgbkrk:
I guess this deleting of `types/nteract__` uncovered some bad gunk. I've got to step away for a bit and come back this evening.

packages/commutable/src/v3.ts
```diff
@@ -214,7 +214,7 @@ export const fromJS = (notebook: Notebook) => {
     cellMap: ImmutableMap().asMutable()
   };
 
-  const cellStructure = [].concat.apply(
+  const cellStructure = ([] as CellStructure[]).concat.apply(
```
rgbkrk, line 217: Wow this turned out to be a simpler fix than I thought.

packages/transform-vdom/src/object-to-react.ts
```diff
@@ -66,7 +66,7 @@ export function objectToReactElement(
   onVDOMEvent: (targetName: string, event: SerializedEvent<any>) => void
 ): React.ReactElement<any> {
   // Pack args for React.createElement
-  let args: any[] = [];
+  let args: any = [];
```
rgbkrk, line 69: One day I should rework how this all gets called & set up.

PR #3819, opened 2018-12-10
Editable Title (first step)

stormpython:
Adding an editable title in place of the static nav title. However, the title will still remain static until I finish the redux logic.

PR #3838, opened 2018-12-12
Add redux to the editable title

stormpython:
This PR will complete the logic for getting the editable title to change the filename, and is related to #3819. It is almost complete. 

editable-title

There are some hiccups and some hacky things to get the url to change and refreshing the browser manually. There is probably a better way to do this.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.tsx
@@ -138,7 +138,7 @@ const mapStateToDirectoryProps = (
   }
 
   const contents: LightDirectoryEntry[] = [];
-  content.model.items.map((entryRef: string) => {
+  content.model.items.map((entryRef: ContentRef) => {
     const row = selectors.content(state, { contentRef: entryRef });
     if (!row) {
       return {
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.tsx
@@ -138,7 +138,7 @@ const mapStateToDirectoryProps = (
   }
 
   const contents: LightDirectoryEntry[] = [];
-  content.model.items.map(entryRef => {
+  content.model.items.map((entryRef: string) => {
     const row = selectors.content(state, { contentRef: entryRef });
     if (!row) {
       return {
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -36,17 +36,6 @@ export type ChangeContentNameFailed = {
   };
 };
 
-export const CHANGE_TITLE_AND_HISTORY_EPIC = "CORE/CHANGE_TITLE_AND_HISTORY_EPIC";
-export type ChangeTitleAndHistoryEpic = {
-  type: "CORE/CHANGE_TITLE_AND_HISTORY_EPIC";
-  payload: {
-    basepath: string;
-    filepath: string;
-    prevFilePath: string;
-    contentRef: ContentRef;
-  };
-};
-
 export const FETCH_CONTENT = "CORE/FETCH_CONTENT";
 export type FetchContent = {
   type: "CORE/FETCH_CONTENT";
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -10,6 +10,16 @@ export type UpdateContent = {
     filepath: string;
     kernelRef: KernelRef;
     contentRef: ContentRef;
+    kernelRef: KernelRef;
+  };
+};
+
+export const UPDATE_CONTENT_FULFILLED = "CORE/UPDATE_CONTENT_FULFILLED";
+export type UpdateContentFulfilled = {
+  type: "CORE/UPDATE_CONTENT_FULFILLED";
+  payload: {
+    filepath: string;
+    contentRef: ContentRef;
   };
 };
 
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -17,6 +17,7 @@ export const CHANGE_CONTENT_NAME_FULFILLED = "CORE/CHANGE_CONTENT_NAME_FULFILLED
 export type ChangeContentNameFulfilled = {
   type: "CORE/CHANGE_CONTENT_NAME_FULFILLED";
   payload: {
+    basepath: string;
     filepath: string;
     prevFilePath: string;
     contentRef: ContentRef;
@@ -27,9 +28,21 @@ export const CHANGE_CONTENT_NAME_FAILED = "CORE/CHANGE_CONTENT_NAME_FAILED";
 export type ChangeContentNameFailed = {
   type: "CORE/CHANGE_CONTENT_NAME_FAILED";
   payload: {
+    basepath: string;
     filepath: string;
     prevFilePath: string;
     error: Error;
+    contentRef: ContentRef;
+  };
+};
+
+export const CHANGE_TITLE_AND_HISTORY_EPIC = "CORE/CHANGE_TITLE_AND_HISTORY_EPIC";
+export type ChangeTitleAndHistoryEpic = {
+  type: "CORE/CHANGE_TITLE_AND_HISTORY_EPIC";
+  payload: {
+    basepath: string;
+    filepath: string;
+    prevFilePath: string;
     contentRef: ContentRef;
   };
 };
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -8,6 +8,7 @@ export type ChangeContentName = {
   type: "CORE/CHANGE_CONTENT_NAME";
   payload: {
     filepath: string;
+    prevFilePath: string;
     contentRef: ContentRef;
   };
 };
@@ -17,6 +18,7 @@ export type ChangeContentNameFulfilled = {
   type: "CORE/CHANGE_CONTENT_NAME_FULFILLED";
   payload: {
     filepath: string;
+    prevFilePath: string;
     contentRef: ContentRef;
   };
 };
@@ -26,6 +28,7 @@ export type ChangeContentNameFailed = {
   type: "CORE/CHANGE_CONTENT_NAME_FAILED";
   payload: {
     filepath: string;
+    prevFilePath: string;
     error: Error;
     contentRef: ContentRef;
   };
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -16,6 +16,11 @@ export type ChangeContentName = {
 export const CHANGE_CONTENT_NAME_FULFILLED = "CORE/CHANGE_CONTENT_NAME_FULFILLED";
 export type ChangeContentNameFulfilled = {
   type: "CORE/CHANGE_CONTENT_NAME_FULFILLED";
+  payload: {
+    filepath: string;
+    prevFilePath: string;
+    contentRef: ContentRef;
+  };
 };
 
 export const CHANGE_CONTENT_NAME_FAILED = "CORE/CHANGE_CONTENT_NAME_FAILED";
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -3,6 +3,16 @@
  */
 import { ContentRef, KernelRef, KernelspecInfo } from "@nteract/types";
 
+export const UPDATE_CONTENT = "CORE/UPDATE_CONTENT";
+export type UpdateContent = {
+  type: "CORE/UPDATE_CONTENT";
+  payload: {
+    filepath: string;
+    kernelRef: KernelRef;
+    contentRef: ContentRef;
+  };
+};
+
 export const FETCH_CONTENT = "CORE/FETCH_CONTENT";
 export type FetchContent = {
   type: "CORE/FETCH_CONTENT";
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -16,12 +16,6 @@ export type ChangeContentName = {
 export const CHANGE_CONTENT_NAME_FULFILLED = "CORE/CHANGE_CONTENT_NAME_FULFILLED";
 export type ChangeContentNameFulfilled = {
   type: "CORE/CHANGE_CONTENT_NAME_FULFILLED";
-  payload: {
-    basepath: string;
-    filepath: string;
-    prevFilePath: string;
-    contentRef: ContentRef;
-  };
 };
 
 export const CHANGE_CONTENT_NAME_FAILED = "CORE/CHANGE_CONTENT_NAME_FAILED";
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -23,6 +23,17 @@ export type UpdateContentFulfilled = {
   };
 };
 
+export const UPDATE_CONTENT_FAILED = "CORE/UPDATE_CONTENT_FAILED";
+export type UpdateContentFailed = {
+  type: "CORE/UPDATE_CONTENT_FAILED";
+  payload: {
+    filepath: string;
+    error: Error;
+    contentRef: ContentRef;
+    kernelRef: KernelRef;
+  };
+};
+
 export const FETCH_CONTENT = "CORE/FETCH_CONTENT";
 export type FetchContent = {
   type: "CORE/FETCH_CONTENT";
--- a/packages/actions/src/actionTypes/contents.ts
+++ b/packages/actions/src/actionTypes/contents.ts
@@ -3,34 +3,31 @@
  */
 import { ContentRef, KernelRef, KernelspecInfo } from "@nteract/types";
 
-export const UPDATE_CONTENT = "CORE/UPDATE_CONTENT";
-export type UpdateContent = {
-  type: "CORE/UPDATE_CONTENT";
-  payload: {
-    filepath: string;
-    kernelRef: KernelRef;
-    contentRef: ContentRef;
-    kernelRef: KernelRef;
-  };
-};
-
-export const UPDATE_CONTENT_FULFILLED = "CORE/UPDATE_CONTENT_FULFILLED";
-export type UpdateContentFulfilled = {
-  type: "CORE/UPDATE_CONTENT_FULFILLED";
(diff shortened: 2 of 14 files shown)


stormpython:
The issue seems to be with `@nteract/epics`. I will dig into that today. Also, need to change the logic around adding titles to make it optional to add the `.ipynb` extension.

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
```diff
@@ -45,6 +46,16 @@ type FileProps = {
...
 export class File extends React.PureComponent<FileProps, *> {
+  constructor(props: FileProps) {
+    super(props);
+
+    this.state = {
+      // Removes the `.ipynb` extension 
+      // title: props.displayName.split(".")[0]
+      title: props.displayName
+    };
+  }
```
rgbkrk: You don't want to be using `this.state` since this should be coming from the `props.displayName` exclusively via the redux store. Otherwise you have to handle many more lifecycle methods.

packages/actions/src/actionTypes/contents.ts
```diff
@@ -3,6 +3,37 @@
...
 import { ContentRef, KernelRef, KernelspecInfo } from "@nteract/types";
 
+export const UPDATE_CONTENT = "CORE/UPDATE_CONTENT";
+export type UpdateContent = {
+  type: "CORE/UPDATE_CONTENT";
+  payload: {
+    filepath: string;
+    kernelRef: KernelRef;
+    contentRef: ContentRef;
+    kernelRef: KernelRef;
```
rgbkrk: We shouldn't need a kernelRef for updating the name of the file.

As for the name of this action, I'd call it `CHANGE_CONTENT_NAME` or something similar (otherwise it sounds like we're updating any number of other things).

packages/actions/src/actions/contents.ts
```diff
@@ -5,6 +5,34 @@ import { ContentRef, KernelRef, KernelspecInfo } from "@nteract/types";
 
 import * as actionTypes from "../actionTypes";
 
+export const updateContent = (payload: {
+  filepath: string;
+  kernelRef: KernelRef;
+  contentRef: ContentRef;
+  kernelRef: KernelRef;
```
rgbkrk: Really only need the `contentRef` and `filepath` here.

packages/epics/src/contents.ts
```diff
@@ -11,12 +11,96 @@ import { Action } from "redux";
...
+
+      const state = state$.value;
+
+      const host = selectors.currentHost(state);
+      if (host.type !== "jupyter") {
+        // Dismiss any usage that isn't targeting a jupyter server
+        return empty();
+      }
+      const serverConfig: ServerConfig = selectors.serverConfig(host);
+      const fileName: string = window.frames.location.pathname.split("/")[3];
```
rgbkrk: This should be coming from the action rather than the current window's location / path name. The `contentRef` allows you to grab the content record and the new filepath should be in the action.

packages/epics/src/contents.ts
```diff
@@ -11,12 +11,96 @@ import { Action } from "redux";
...
+            window.history.replaceState(
+              {}, 
+              filepath, 
+              `${path}${filepath}`
+            );
+            // TOTAL HACK!!! This needs to be replaced. 
+            // Refreshes browser after confirming a title
+            // change. Due to the client requesting the 
+            // notebook file before the file change is finished. 
+            window.location.reload(true);
```
rgbkrk: We definitely don't want to do a reload, as we'll lose out on anything the user currently has in progress. What was this used for?

packages/reducers/src/core/entities/contents/index.ts
```diff
@@ -22,6 +23,31 @@ import { file } from "./file";
 
 const byRef = (state = Immutable.Map(), action: Action) => {
   switch (action.type) {
+    case actionTypes.UPDATE_CONTENT:
+      const updateContentAction = action as actionTypes.UpdateContent;
+      const { filepath, contentRef, kernelRef } = updateContentAction.payload;
+
```
rgbkrk: You can use `state.setIn` to alter just the filepath in the reducer, instead of setting up a new object on that `contentRef`.

packages/rx-jupyter/src/contents.ts
```diff
@@ -38,6 +38,7 @@ export type Payload = {
   mimetype: string;
   content: string | Notebook;
   format: string;
+  kernel: any
```
rgbkrk: Is this coming from the API now?

packages/actions/src/actionTypes/contents.ts
stormpython: In Jupyter Classic, the kernelRef is a key on the object that gets sent.
```javascript prettyprint
{
  "path":"Almond.ipynb",
  "type":"notebook",
  "name":"",
  "kernel":{
    "id":"88cfcc1b-ae48-4eb7-9d5f-dc4fed4af3f1",
    "name":"praline"
  }
}
```

packages/epics/src/contents.ts
stormpython: It used because after editing the title, the contents of the notebook disappear.

packages/rx-jupyter/src/contents.ts
stormpython: Yes. This is an example from Jupyter Classic when renaming:

```javascript
{
  "path":"Almond.ipynb",
  "type":"notebook",
  "name":"",
  "kernel":{
    "id":"88cfcc1b-ae48-4eb7-9d5f-dc4fed4af3f1",
    "name":"praline"
  }
}
```

stormpython: I need to change that from any and create a type for it

packages/epics/src/contents.ts
rgbkrk: That's because of the reducer for `UPDATE_CONTENT`. Right now this PR is completely wiping out the prior state of the content instead of just updating the filepath.

packages/reducers/src/core/entities/contents/index.ts
```diff
@@ -22,6 +23,31 @@ import { file } from "./file";
...
+      return state.set(
+        contentRef,
+        {
+          filepath: filepath || "",
+          kernelRef: kernelRef
+        }
```
rgbkrk: This is effectively deleting all of the content -- it's replacing the original content record with an object (not to mention an object without all the fields).

packages/actions/src/actionTypes/contents.ts
stormpython: Nope, you were right.

applications/desktop/src/notebook/epics/index.js
```diff
@@ -51,6 +52,7 @@ const epics = [
   publishEpic,
   saveEpic,
   saveAsEpic,
+  updateContentEpic,
```
rgbkrk: +1, glad you figured out where to add new epics in

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
```diff
@@ -7,7 +7,9 @@ import { selectors } from "@nteract/core";
 import type { ContentRef, AppState } from "@nteract/core";
 import { LoadingIcon, SavingIcon, ErrorIcon } from "@nteract/iron-icons";
 import { connect } from "react-redux";
-import { EditableText } from "@blueprintjs/core";
+import { FormGroup, classNames } from "@blueprintjs/core";
+import { Classes } from "@blueprintjs/core";
+import { actions } from "@nteract/core";
```
rgbkrk: Now that we have things broken up, you can import the actions from `@nteract/actions`. You may need to build core in this case as well `yarn build:only @nteract/core`

packages/actions/src/actions/contents.ts
```diff
@@ -5,6 +5,31 @@ import { ContentRef, KernelRef, KernelspecInfo } from "@nteract/types";
...
+
+export const changeContentNameFailed = (payload: {
+  filepath: string,
+  error: Error,
+  contentRef: ContentRef,
+}): actionTypes.ChangeContentNameFailed => ({
+  type: actionTypes.CHANGE_CONTENT_NAME_FAILED,
+  payload
+});
+
```
rgbkrk, line 38: I'm liking the naming structure now.

packages/epics/src/contents.ts
```diff
@@ -11,12 +11,78 @@ import { Action } from "redux";
...
+            name: "",
+            path: filepath.slice(1),
+            type: "notebook",
+            writable: true,
+            created: "",
+            last_modified: "",
+            mimetype: "",
+            content: "",
+            format: "",
+            kernel: {}
```
rgbkrk: I'm totally surprised it needs all of these. Then again, perhaps it has to hit the sessions API as well to update that.

packages/reducers/src/core/entities/contents/index.ts
```diff
@@ -22,6 +21,24 @@ import { file } from "./file";
 
 const byRef = (state = Immutable.Map(), action: Action) => {
   switch (action.type) {
+    case actionTypes.CHANGE_CONTENT_NAME:
+      const changeContentNameAction = action as actionTypes.ChangeContentName;
+      const { filepath } = changeContentNameAction.payload;
+      state.setIn(["filepath"], filepath);
+    case actionTypes.CHANGE_CONTENT_NAME_FULFILLED:
+      const changeContentNameFulfilledAction = action as actionTypes.ChangeContentNameFulfilled;
+      console.log("SUCCESS");
```
rgbkrk: When testing, you probably want this to return at least just `state` so it stays consistent.

rgbkrk: Right now this is passing through the action below (falling through to the next case statement).

packages/epics/src/contents.ts
stormpython: I probably don't need all of these, so I will reduce this to just what is needed.

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
```diff
@@ -91,11 +96,28 @@ export class File extends React.PureComponent<FileProps, *> {
...
-                disabled={true}
-                placeholder={"Enter Title..."}
-                value={this.props.displayName}
-              />
+              <FormGroup
+                disabled={false}
+                intent={"primary"} 
+              >
+                <input 
+                  className={Classes.EDITABLE_TEXT_INPUT}
```
rgbkrk: What prompted you to switch to using blueprint's classes instead of `<EditableText>`?

packages/epics/src/contents.ts
```diff
@@ -17,6 +17,69 @@ import * as selectors from "@nteract/selectors";
...
+              throw new Error(xhr.response);
+            }
+          }),
+          map(() => {
+            /* 
+             * Modifying the url's file name in the browser. 
+             * Effects back button behavior.
+             * Is there a better way to accomplish this?
+             */
+            window.history.replaceState({}, filepath, `/nteract/edit${filepath}`);
```
rgbkrk: We may want to put this in an epic that listens for `changeContentNameFulfilled`'s action, so that it's a side effect we're opting in to only for the jupyter extension.

```
const changeTitleAndHistoryEpic = action$.pipe(
  ofType(actions.ChangeContentNameFulfilled),
  switchMap(action => { window.history.replaceState(...); return empty(); })
)
```

There's also one extra thing that needs to be handled here -- the path needs to include the `basePath` `${basePath}/nteract/edit/${filepath}` so that in a multi-user environment it keeps their username `/users/kylek/nteract/edit/filepath`. It's in the serverConfig -- I can't recall the name of the property off the top of my head.

applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.js
stormpython: I was switching the input from a controlled to uncontrolled component. And I couldn't get the `EditableText` component to work as an uncontrolled input. So, I just used a standard HTML input element and used `ref={input => this.inputRef = input}` to grab hold of the value in the input DOM element, and the value is only stored when the input loses focus and therefore allows users to continue inputing characters without the component re-rendering until after the input has lost focus.

stormpython: The blueprintjs classes were an attempt to get the input styled exactly the way that the `editableText` component get styled. This is obviously not the case, and I will need to add some other classes to the `editableText` component.

stormpython: I replaced the `input` element with the `EditableText` element and used the `defaultValue` prop instead of the `value` prop to make the `EditableText` component an uncontrolled component.

stormpython:
@rgbkrk, this is ready to be reviewed again. Here is a view of the new functionality:

editable-title

stormpython:
I need to rebase this against master.

rgbkrk:
I'm on the branch now and seeing this:

screen shot 2018-12-21 at 2 59 36 pm

I think I'll play with this one again when I'm at home with monitors and not handling other stuff. 😬

stormpython:
> I'm on the branch now and seeing this:
> 
> screen shot 2018-12-21 at 2 59 36 pm
> 
> I think I'll play with this one again when I'm at home with monitors and not handling other stuff. 😬

This is an error that is also occurring for me in master.

rgbkrk:
> This is an error that is also occurring for me in master.

Oh no. This seems like part of the typescript migration fun.

packages/reducers/src/core/entities/contents/index.ts
```diff
@@ -28,6 +28,23 @@ const byRef = (
...
+    case actionTypes.CHANGE_CONTENT_NAME_FULFILLED:
+      const changeContentNameFulfilledAction = action as actionTypes.ChangeContentNameFulfilled;
+      const { contentRef, filepath } = changeContentNameFulfilledAction.payload;
+
+      /*
+       * Modifying the url's file name in the browser.
+       * This effects back button behavior.
+       * Is there a better way to accomplish this?
+      */
+      window.history.replaceState({}, filepath, `/nteract/edit${filepath}`);
```
rgbkrk: This should be done in an epic instead of a reducer for the following reasons:

* For the sake of clean invariants, no side effects should occur in reducers
* The path _must_ include the user's baseURL

applications/jupyter-extension/nteract_on_jupyter/app/contents/directory.tsx
```diff
@@ -138,7 +138,7 @@ const mapStateToDirectoryProps = (
   }
 
   const contents: LightDirectoryEntry[] = [];
-  content.model.items.map(entryRef => {
+  content.model.items.map((entryRef: string) => {
```
stormpython: Should the type here be `ContentRef` and not `string`?

rgbkrk: Probably a good idea, though because of TypeScript's type aliasing it's still a string at the end of the day.

packages/epics/src/contents.ts
```diff
@@ -9,13 +9,73 @@ import FileSaver from "file-saver";
...
+              throw new Error(xhr.response);
+            }
+          }),
+          map(() => {
+            /*
+            * Modifying the url's file name in the browser.
+            * This effects back button behavior.
+            * Is there a better way to accomplish this?
+            */
+            window.history.replaceState({}, filepath, `/nteract/edit${filepath}`);
```
rgbkrk:
```suggestion
            window.history.replaceState({}, filepath, urljoin(host.basePath, `/nteract/edit${filepath}`));
```

packages/epics/src/contents.ts
```diff
@@ -9,13 +9,73 @@ import FileSaver from "file-saver";
...
-import { toJS, stringifyNotebook, Notebook } from "@nteract/commutable";
+import { toJS, stringifyNotebook } from "@nteract/commutable";
+import { Notebook } from "@nteract/commutable";
 
 import * as actions from "@nteract/actions";
 import * as selectors from "@nteract/selectors";
 import { ContentRef, AppState } from "@nteract/types";
 import { AjaxResponse } from "rxjs/ajax";
 
+const urljoin = require("url-join");
```
rgbkrk:
```suggestion
import urljoin from "url-join";
```

stormpython:
@rgbkrk I updated the repo with your changes. I needed to fix an issue with errors in the console due to an action being passed improperly from an epic. Hopefully all the checks pass!

packages/epics/src/contents.ts
```diff
@@ -9,13 +9,78 @@ import FileSaver from "file-saver";
...
+              throw new Error(xhr.response);
+            }
+          }),
+          map(() => {
+            /*
+            * Modifying the url's file name in the browser.
+            * This effects back button behavior.
+            * Is there a better way to accomplish this?
+            */
+            window.history.replaceState({}, filepath, urljoin(host.basePath, `/nteract/edit${filepath}`));
```
rgbkrk, line 62: 👍

packages/rx-jupyter/src/contents.ts
```diff
@@ -72,6 +72,18 @@ export interface IContent<FT extends FileType = FileType>
...
+  path: string;
+  type: "directory" | "file" | "notebook";
+  writable: boolean;
+  created: string;
+  last_modified: string;
+  mimetype: string;
+  content: string | Notebook;
+  format: string;
+  kernel: any;
+};
```
rgbkrk: I _think_ `Payload` can be `IFileContent<FileType>`. Changes were made to make the epics and actions using these to be less `any`-prone, which led to the creation of `IFileContent`. I'm happy to do a follow on PR to rectify this, as this PR has been through the paces long enough.

packages/actions/src/actions/contents.ts
```diff
@@ -7,11 +7,41 @@ import * as actionTypes from "../actionTypes";
...
+  contentRef: ContentRef;
+  prevFilePath: string;
+}): actionTypes.ChangeContentNameFulfilled => ({
+  type: actionTypes.CHANGE_CONTENT_NAME_FULFILLED,
+  payload
+});
 
-export const fetchContent = (
-  payload: actionTypes.FetchContent["payload"]
-): actionTypes.FetchContent => ({
```
rgbkrk: I'm guessing you didn't mean to change `fetchContent` as part of this, must have been an accidental happening on a rebase.

stormpython: yeah, this happened with the last rebase. Let me correct this.

packages/rx-jupyter/src/contents.ts
stormpython: Which file is the IFileContent interface in? I am assuming the `I` stands for `interface`.

rgbkrk: Same file, above `Payload` here.

rgbkrk (approved):
Thanks for taking this through to completion, stoked to have this.

PR #3839, opened 2018-12-12
target es2017 for compilation

rgbkrk:
We were defaulting to producing ES3 output (since this wasn't set) when we can be on ES2017 (for node 8+, current browsers).


willingc (approved):
Thanks @rgbkrk

PR #3841, opened 2018-12-12
Persist metadata after UI controls changes in DX

emeeks:
* Actions

### DIFF
--- a/packages/actions/src/actionTypes/cells.ts
+++ b/packages/actions/src/actionTypes/cells.ts
@@ -246,12 +246,13 @@ export type UpdateCellStatus = {
   };
 };
 
-export const UPDATE_CELL_METADATA = "UPDATE_CELL_METADATA";
+export const UPDATE_OUTPUT_METADATA = "UPDATE_OUTPUT_METADATA";
 export type UpdateOutputMetadata = {
-  type: "UPDATE_CELL_METADATA";
+  type: "UPDATE_OUTPUT_METADATA";
   payload: {
     id: CellId;
     contentRef: ContentRef;
     metadata: Object;
+    index: number;
   };
 };
--- a/packages/actions/src/actionTypes/cells.ts
+++ b/packages/actions/src/actionTypes/cells.ts
@@ -1,7 +1,7 @@
 /**
  * @module actions
  */
-import { CellId, CellType } from "@nteract/commutable";
+import { CellId, CellType, JSONObject } from "@nteract/commutable";
 import { ContentRef } from "@nteract/types";
 
 export const TOGGLE_TAG_IN_CELL = "CORE/TOGGLE_TAG_IN_CELL";
@@ -252,7 +252,7 @@ export type UpdateOutputMetadata = {
   payload: {
     id: CellId;
     contentRef: ContentRef;
-    metadata: Object;
+    metadata: JSONObject;
     index: number;
   };
 };
--- a/packages/actions/src/actionTypes/cells.ts
+++ b/packages/actions/src/actionTypes/cells.ts
@@ -245,3 +245,13 @@ export type UpdateCellStatus = {
     contentRef: ContentRef;
   };
 };
+
+export const UPDATE_CELL_METADATA = "UPDATE_CELL_METADATA";
+export type UpdateCellMetadata = {
+  type: "UPDATE_CELL_METADATA";
+  payload: {
+    id: CellId;
+    contentRef: ContentRef;
+    metadata: Object;
+  };
+};
--- a/packages/actions/src/actionTypes/cells.ts
+++ b/packages/actions/src/actionTypes/cells.ts
@@ -253,5 +253,6 @@ export type UpdateCellMetadata = {
     id: CellId;
     contentRef: ContentRef;
     metadata: Object;
+    mimeType: string;
   };
 };
--- a/packages/actions/src/actionTypes/cells.ts
+++ b/packages/actions/src/actionTypes/cells.ts
@@ -247,12 +247,11 @@ export type UpdateCellStatus = {
 };
 
 export const UPDATE_CELL_METADATA = "UPDATE_CELL_METADATA";
-export type UpdateCellMetadata = {
+export type UpdateOutputMetadata = {
   type: "UPDATE_CELL_METADATA";
   payload: {
     id: CellId;
     contentRef: ContentRef;
     metadata: Object;
-    mimeType: string;
   };
 };
--- a/packages/actions/src/actions/cells.ts
+++ b/packages/actions/src/actions/cells.ts
@@ -342,3 +342,19 @@ export function changeCellType(payload: {
     payload
   };
 }
+
+export function updateCellMetadata(payload: {
+  id: CellId;
+  metadata: Object;
+  contentRef: ContentRef;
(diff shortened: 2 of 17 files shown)


emeeks:
CI is failing but the functionality is working, would love comments and review

itt_muse_3

packages/display-area/src/richest-mime.tsx
```diff
@@ -103,13 +104,26 @@ export default class RichestMime extends React.Component<
...
     const data: any = this.props.bundle[mimetype];
     const metadata: any = this.props.metadata[mimetype];
+    //    console.log("this.props", this.props);
+    //    console.log("this.props.metadata", this.props.metadata);
+    const onMetadataChangeScoped = (scopedChanges: any) =>
+      this.props.onMetadataChange &&
+      this.props.onMetadataChange(
+        {
+          ...this.props.metadata,
+          ...scopedChanges
```
rgbkrk: Passing mimetype as one of the arguments is making this more complex than necessary. I'd go for bubbling up the metadata like this:

```js
this.props.onMetadataChange({
  ...this.props.metadata,
  [mimetype]: scopedChanges
})
```

so you don't have to bubble the mimetype up and re-work it above.

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -125,6 +127,8 @@ const mapStateToCellProps = (
 
   const pager = model.getIn(["cellPagers", id]) || Immutable.List();
 
+  const metadata = (cell.getIn(["metadata"]) || Immutable.Set()).toJS();
```
rgbkrk: This should be an `Immutable.Map`.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -656,6 +656,19 @@ function updateCellStatus(
...
+  state: NotebookModel,
+  action: actionTypes.UpdateCellMetadata
+) {
+  const { id, metadata, mimeType } = action.payload;
+  console.log("updateCellMetadata", action);
+  console.log("store before update", state);
+  return state.setIn(
+    ["notebook", "cellMap", id, "metadata", mimeType],
+    metadata
+  );
```
rgbkrk, line 669: This looks like it's setting the metadata on the cell rather than the output, where the metadata belongs.

emeeks: The reason I was passing mimeType through was because I didn't think to pass the metadata through to the output and mimetype lets me do that cleaner `setIn`. How would I set the metadata of the output?

packages/display-area/src/richest-mime.tsx
emeeks: This is the right way to go once I figure out how to send it to the output rather than the metadata of the cell.

packages/actions/src/actionTypes/cells.ts
```diff
@@ -245,3 +245,13 @@ export type UpdateCellStatus = {
     contentRef: ContentRef;
   };
 };
+
+export const UPDATE_CELL_METADATA = "UPDATE_CELL_METADATA";
```
rgbkrk: Should switch this over to `UPDATE_OUTPUT_METADATA`

packages/display-area/src/output.tsx
```diff
@@ -14,6 +14,8 @@ type Props = {
   theme: string;
   models: Object;
   channels?: Subject<any>;
+  metadata: Object;
+  onMetadataChange?: Function;
```
rgbkrk: Instead of typing this as `Function`, you can make this explicit with a type `(metadata: Object) => void`.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -656,6 +656,23 @@ function updateCellStatus(
...
+function updateOutputMetadata(
+  state: NotebookModel,
+  action: actionTypes.UpdateOutputMetadata
+) {
+  const { id, metadata } = action.payload;
+  const currentOutputs = state.getIn(["notebook", "cellMap", id, "outputs"]);
+
+  const updatedOutputs = currentOutputs.update(
+    currentOutputs.findIndex(
+      (item: any) => item.get("output_type") === "execute_result"
```
rgbkrk: There can be more than one `execute_result` in a collection of outputs and the data explorer can also exist on output type `display_data`. Since we don't have IDs for outputs, you'll have to pass the index of the output that got the on metadata change callback.

rgbkrk (approved):

rgbkrk (changes requested):
Whoops, meant to mark this as needing changes, not approved. 😄

packages/actions/src/actionTypes/cells.ts
emeeks: Weird, not sure how I missed this one.

packages/actions/src/actionTypes/cells.ts
```diff
@@ -245,3 +245,14 @@ export type UpdateCellStatus = {
...
   };
 };
+
+export const UPDATE_OUTPUT_METADATA = "UPDATE_OUTPUT_METADATA";
+export type UpdateOutputMetadata = {
+  type: "UPDATE_OUTPUT_METADATA";
+  payload: {
+    id: CellId;
+    contentRef: ContentRef;
+    metadata: Object;
```
captainsafia: You'll probably want to replace this with `JSONObject`. In TypeScript, `Object` represents all JavaScript objects, not objects like {}.

packages/actions/src/actions/cells.ts
```diff
@@ -342,3 +342,15 @@ export function changeCellType(payload: {
     payload
   };
 }
+
+export function updateOutputMetadata(payload: {
+  id: CellId;
+  metadata: Object;
```
captainsafia: Ditto here as well.

packages/actions/src/index.ts
```diff
@@ -1,5 +1,8 @@
 /**
  * @module actions
  */
+
+import * as ActionTypes from "./actionTypes/index";
```
rgbkrk: Was this import necessary?

emeeks: Nope! Just left from some logging an issue and accidentally left it in.

packages/display-area/src/display.tsx
```diff
@@ -40,7 +40,7 @@ export default class Display extends React.PureComponent<Props> {
         >
           {outputs
             ? outputs.map((output, index) => (
-                <Output key={index} output={output} {...props} />
+                <Output key={index} index={index} output={output} {...props} />
```
captainsafia, line 43: What is this new `index` prop being used for?

captainsafia (commented):
Looks great! I left some comments related to the types inline.

packages/actions/src/actionTypes/cells.ts
emeeks: Didn't realize, thanks!

packages/display-area/src/display.tsx
emeeks: It's used to know in the updateOutputMetadata in `output.tsx` function which output in the arrays of outputs to update.

rgbkrk:
Thank you!

PR #3843, opened 2018-12-13
Use camelCase for values in commutable

captainsafia:
Updates commutable to use camel case property names and updates tests and code to refer to them.

### DIFF
--- a/packages/commutable/src/v4.ts
+++ b/packages/commutable/src/v4.ts
@@ -132,7 +132,7 @@ export const createImmutableOutput = (output: Output): ImmutableOutput => {
   switch (output.output_type) {
     case "execute_result":
       return makeExecuteResult({
-        execution_count: output.execution_count,
+        executionCount: output.execution_count,
         data: createImmutableMimeBundle(output.data),
         metadata: immutableFromJS(output.metadata)
       });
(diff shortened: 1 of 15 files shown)


rgbkrk (changes requested):
Commutable's toJS functions will need to take the `camelCase` -> `snake_case` into account since it has to match the jupyter notebook format. This also highlights that we need some conformance tests based on commutable output, having it run against the nbformat json schema.

captainsafia:
>Commutable's toJS functions will need to take the camelCase -> snake_case into account since it has to match the jupyter notebook format. This also highlights that we need some conformance tests based on commutable output, having it run against the nbformat json schema.

Is there a particular `toJS` function you are referring? I believe I updated all the cell related ones.

packages/commutable/src/v4.ts
```diff
@@ -285,29 +285,29 @@ const mimeBundleToJS = (immMimeBundle: ImmutableMimeBundle): MimeBundle => {
...
-  switch (output.output_type) {
+  switch (output.outputType) {
     case "execute_result":
       return {
-        output_type: output.output_type,
-        execution_count: output.execution_count,
+        output_type: output.outputType,
+        execution_count: output.executionCount,
         data: mimeBundleToJS(output.data),
         metadata: output.metadata.toJS()
```
rgbkrk, line 294: I stand corrected, you've totally got this covered.

rgbkrk (approved):

rgbkrk:
This turned out really nicely. Do we also need to update the reducers and epics that generate outputs?

rgbkrk:
Ah crap, we do need to update display area and others before we can have this. I'm going to revert so we can have `master` in a working state.

rgbkrk:
I'll make a new PR that uses all your commits and also making sure we adapt to the new APIs, so it'll be a one-shot PR.

PR #3847, opened 2018-12-13
Fixes padding in nteract desktop and flexes the web app

stormpython:
This PR fixes the margin I introduced in #3764. I also wrap the web app components in flex-box divs.

### DIFF
--- a/packages/notebook-app-component/src/notebook-app.tsx
+++ b/packages/notebook-app-component/src/notebook-app.tsx
@@ -608,7 +608,6 @@ export class NotebookApp extends React.PureComponent<NotebookProps> {
         />
         <style jsx>{`
           .cells {
-            margin-top: 80px;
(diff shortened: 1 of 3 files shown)


packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -608,7 +608,6 @@ export class NotebookApp extends React.PureComponent<NotebookProps> {
         />
         <style jsx>{`
           .cells {
-            margin-top: 80px;
```
jdetle, line 611: Was this originally nonexistent or 20px?

stormpython: Not sure, let me check the history.

stormpython: In the history before my commit, there was no margin set. That is, it was nonexistent.

Issue #2387, opened 2018-01-17
Errors inside reducers are not being reported up


rgbkrk:
Our error middleware is not noticing issues. The only way I was able to notice was via the `DEBUG=true` mode.

jameskoch:
Which app?  I just noticed that in jupyter_extension we don't wire in the errorMiddleware. PR coming shortly for that.

captainsafia:
I'm gonna go ahead and close this one because it seems to be resolved.

PR #3848, opened 2018-12-13
Wire errorMiddleware in jupyter_extension similarly to how it's wired in desktop

jameskoch:
Possibly related to #2387.

On macOS I'm seeing occasional errors in the UI. I had to add some
logging manually to track down that they're coming from saveContentEpic.
With this change they are reported to the console, albeit TWICE.
First first logging happens the the errorMiddleware itself here - https://github.com/nteract/nteract/blob/6d9e66cd78395a395b9e80a346b4d73a58261422/packages/core/src/middlewares.ts#L21
The message is then propagated to the notifactionSystem, which also logs
here - https://github.com/nteract/nteract/blob/255761ed1f9d640e47740da6fbe3fd035f4bd7bc/packages/types/src/index.ts#L143

Any thoughts on further sanitizing this? I'm happy to leave it double-
reporting; better that than not at all.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/store.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/store.js
@@ -2,7 +2,7 @@
 import { combineReducers, createStore, applyMiddleware, compose } from "redux";
 import { createEpicMiddleware, combineEpics } from "redux-observable";
 import type { AppState } from "@nteract/core";
-import { reducers, epics as coreEpics } from "@nteract/core";
+import { reducers, epics as coreEpics, middlewares as coreMiddlewares } from "@nteract/core";
 
 const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Double reporting for now is fine.

applications/jupyter-extension/nteract_on_jupyter/app/store.js
```diff
@@ -17,15 +17,16 @@ export default function configureStore(initialState: AppState) {
...
 
   const store = createStore(
     rootReducer,
     initialState,
-    composeEnhancers(applyMiddleware(middlewares))
+    composeEnhancers(applyMiddleware(...middlewares))
   );
 
-  middlewares.run(rootEpic);
+  epicMiddleware.run(rootEpic);
```
rgbkrk, line 29: Wow this is much cleaner now, I shouldn't have done `run` against an object called middlewares since the usual practice was to have an array of middleware as middlewares. Thanks for both cleaning this up and adding in the error middleware.

rgbkrk (approved):

Issue #3849, opened 2018-12-13
Unforeseen styled-jsx issues in typescript land


rgbkrk:
`styled-jsx` inline CSS _must_ be transpiled using the babel config. Now that (many) components are migrated to typescript, this isn't happening. Now they're ending up in one of two bad spots -- they're being created as regular `style` elements with a useless `jsx` property or they're getting used as template literals which generate errors upon use:

```js
> css`a { color: green; }`
Error: styled-jsx/css: if you are getting this error it means that your `css` tagged template literals were not transpiled.
    at notTranspiledError (/Users/kylek/code/src/github.com/nteract/nteract-ext/node_modules/styled-jsx/css.js:2:9)
    at css (/Users/kylek/code/src/github.com/nteract/nteract-ext/node_modules/styled-jsx/css.js:9:3)
```

I've done a bit of digging prior to opening this issue to see if we can migrate our current components to effectively using the transpiled API. Effectively that comes out to something like this:

```jsx
import Style from "styled-jsx/style";

export const Cell = props => {
  const children = props.children;
  const styleId = generateIdFromSomewhere();

  const jsxClassName = `.jsx-${styleId}`;

  return (
    <div
      className={`${jsxClassName} cell ${props.isSelected ? "focused" : ""} ${
        props._hovered ? "overrideHover" : ""
      }`}
    >
      <Style
        styleId={styleId}
        css={`
          .cell${jsxClassName} {
            position: relative;
            background: var(--theme-cell-bg, white);
            transition: all 0.1s ease-in-out;
          }
          ...more
        `}
      />
      {children}
    </div>
  );
};
```

I haven't yet figured out how I could turn the css string into the prefixed one here (there must be some internal styled-jsx API I could use).

All this to say... We need to figure out a direction now. Open questions:

* Should we switch back to babel, yet using the typescript plugin for babel, so we can run the styled-jsx transform?
* Should we migrate to styled-components or typestyle?
* What's another direction we could take?

Apologies for not seeing this one in advance and not having a clear direction for us to go yet.

rgbkrk:
This is now done. We've switched to `styled-components`!

PR #3852, opened 2018-12-15
Use styled-components in presentational-components

rgbkrk:
Switches `@nteract/presentational-components` to using `styled-components` 💅 

Explorative approach to handling #3849.

### DIFF
--- a/packages/presentational-components/package.json
+++ b/packages/presentational-components/package.json
@@ -9,7 +9,7 @@
     "prepare": "npm run build",
     "prepublishOnly": "npm run build",
     "build": "npm run build:clean && npm run build:lib",
-    "build:clean": "tsc -b --clean",    
+    "build:clean": "tsc -b --clean",
     "build:lib": "tsc -b",
     "build:lib:watch": "tsc -b --watch",
     "build:watch": "npm run build:clean && npm run build:lib:watch"
@@ -19,6 +19,7 @@
     "@blueprintjs/select": "^3.2.0",
     "@nteract/styled-blueprintjsx": "^1.1.1",
     "react-syntax-highlighter": "^9.0.0",
+    "styled-components": "^4.1.2",
     "styled-jsx": "^3.1.0"
   },
   "peerDependencies": {
@@ -34,5 +35,6 @@
   "author": "Kyle Kelley <rgbkrk@gmail.com>",
   "license": "BSD-3-Clause",
   "devDependencies": {
+    "@types/styled-components": "^4.1.4"
   }
 }
--- a/packages/presentational-components/src/components/cell.tsx
+++ b/packages/presentational-components/src/components/cell.tsx
@@ -1,75 +1,103 @@
 import * as React from "react";
 
+import styled from "styled-components";
+
+import { Prompt } from "./prompt";
+
 interface CellProps {
   /**
    * Indicates if a cell is selected
    */
-  isSelected?: boolean;
+  isSelected: boolean;
   /**
    * Indicates if hovering over a cell
    */
-  _hovered?: boolean;
+  _hovered: boolean;
   /**
    * Style children when a cell is selected or hovered over
    */
   children?: React.ReactNode;
 }
 
-/** @component */
-export const Cell = (props: CellProps) => {
-  const children = props.children;
-
-  return (
-    <div
-      className={`cell ${props.isSelected ? "focused" : ""} ${
-        props._hovered ? "overrideHover" : ""
-      }`}
-    >
-      <style jsx>{`
-        .cell {
-          position: relative;
-          background: var(--theme-cell-bg, white);
-          transition: all 0.1s ease-in-out;
-        }
-
-        .cell.overrideHover,
-        .cell:hover {
-          box-shadow: var(
-            --theme-cell-shadow-hover,
-            1px 1px 3px rgba(0, 0, 0, 0.12),
-            -1px -1px 3px rgba(0, 0, 0, 0.12)
-          );
-        }
-
-        .cell.focused {
-          box-shadow: var(
-            --theme-cell-shadow-focus,
-            3px 3px 9px rgba(0, 0, 0, 0.12),
-            -3px -3px 9px rgba(0, 0, 0, 0.12)
-          );
-        }
-
-        .cell.overrideHover :global(.prompt),
-        .cell:hover :global(.prompt),
-        .cell:active :global(.prompt) {
-          background-color: var(--theme-cell-prompt-bg-hover, hsl(0, 0%, 94%));
-          color: var(--theme-cell-prompt-fg-hover, hsl(0, 0%, 15%));
-        }
-
-        .cell:focus :global(.prompt),
-        .cell.focused :global(.prompt) {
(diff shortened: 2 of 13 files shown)


rgbkrk:
I am so in love with styled-components now, extremely happy with the API and how a lot of our components become smaller while also being super easy to override their styles. ❤️ 

Not that you _want_ to do this, but this is now possible:

```
  <Cell style={{backgroundColor: "yellow"}}>
    <Input>
      <Prompt style={{backgroundColor: "pink"}} />
      <Source language="python">{`# Level 0 - Flat`}</Source>
    </Input>
    <Outputs>
      <pre>Output</pre>
    </Outputs>
  </Cell>
```

screen shot 2018-12-15 at 5 39 32 pm

packages/presentational-components/src/components/cell.tsx
```diff
@@ -15,61 +19,77 @@ interface CellProps {
...
-          position: relative;
-          background: var(--theme-cell-bg, white);
-          transition: all 0.1s ease-in-out;
-        }
+    box-shadow: ${(props: CellProps) =>
+      props.isSelected
+        ? levels.SELECTED
+        : props._hovered
+        ? levels.HOVERED
+        : levels.FLAT};
```
captainsafia: Would it be possible to create a variable to store the levels property in and not use a nested ternary here?

packages/presentational-components/package.json
```diff
@@ -34,5 +35,6 @@
   "author": "Kyle Kelley <rgbkrk@gmail.com>",
   "license": "BSD-3-Clause",
   "devDependencies": {
+    "@types/styled-components": "^4.1.4"
```
captainsafia, line 38: I think we probably wanna move this to our top-level debDependencies?

packages/presentational-components/src/components/prompt.tsx
```diff
@@ -46,38 +48,42 @@ export function promptText(props: PromptProps): string {
...
 }
 
-export class Prompt extends React.Component<PromptProps, {}> {
-  static defaultProps = {
-    counter: null,
-    running: false,
-    queued: false,
-    blank: false
-  };
+/** @component */
```
captainsafia: Is this code comment doing something special with regard to styled-component?

packages/presentational-components/src/components/prompt.tsx
```diff
@@ -46,38 +48,42 @@ export function promptText(props: PromptProps): string {
...
-      </React.Fragment>
-    );
+    color: var(--theme-cell-prompt-fg, black);
+    background-color: var(--theme-cell-prompt-bg, #fafafa);
   }
-}
 
-export const PromptBuffer = () => <Prompt blank />;
+  &:after {
+    content: ${(props: PromptProps) =>
```
captainsafia: This style seems a little confusing to me. Can the content be rendered in the component instead of being set with CSS?

packages/presentational-components/src/components/cell.tsx
```diff
@@ -15,61 +19,77 @@ interface CellProps {
   children?: React.ReactNode;
 }
 
+// Note: no happy syntax highlighting if I include the generic
+//
+//   export const Cell2 = styled.div<CellProps>`
+//
+
+const levels = {
```
captainsafia: Style nit: maybe call this “boxShadowLevels”?

captainsafia:
Looks neat. I left some comments in line. Can you elaborate on what you think is good about the API?

captainsafia (commented):
Looks neat! Can you elaborate on what you like about the API?

packages/presentational-components/src/components/prompt.tsx
rgbkrk: This was something I thought we needed for the docs pages. I saw it on `cell.tsx` and applied it here as well.

packages/presentational-components/src/components/prompt.tsx
rgbkrk: Yep, can do. I can make this a functional component and then wrap it like this: `const Prompt = styled(BarePrompt)`.

packages/presentational-components/src/components/cell.tsx
rgbkrk: Good idea.

packages/presentational-components/src/components/cell.tsx
rgbkrk: Yeah, in fact what I should do is just make a function that returns the css string literal here, like is used in `Prompt`.

rgbkrk:
> Can you elaborate on what you like about the API?

The primary things I like about the API:

* We can write regular CSS \*\*
* The CSS is done as a template literal so you can embed expressions inline
* You can refer to other styled-components by name, like was done here with `Prompt`. With styled-jsx we had to use that weird `global()` selector that confused people to no end
* Get to treat your components like regular elements that have default styles. `styled.button` is still a button with all the regular DOM attributes including `onClick`, etc. -- this is probably my favorite part of the API -- that the component API is the regular DOM API -- they're regular components with 🌟 style 🌟

\*\* the `&`, like in sass, is used to indicate the current component's generated class name in `styled-components`.

Network effects I like:

* `styled-components` is the [most downloaded, most forked, and most starred of the CSS-in-JS libraries](https://www.npmtrends.com/aphrodite-vs-emotion-vs-glamorous-vs-jss-vs-radium-vs-styled-components-vs-styled-jsx)
* [Syntax Highlighting and IntelliSense plugin for VSCode](https://github.com/styled-components/vscode-styled-components)
* Good typescript definitions, active community around them

All the above applies to [`emotion`](https://emotion.sh/docs/styled) as well. One interesting thing to note with emotion is that typescript will check/validate css properties when using the object syntax. I should see if the same applies to `styled-components`.

rgbkrk:
Problems we're going to be running into:

* https://github.com/styled-components/styled-components/issues/1914
* https://github.com/DefinitelyTyped/DefinitelyTyped/issues/29540

packages/presentational-components/src/components/cell.tsx
```diff
@@ -1,76 +1,104 @@
...
+    --theme-cell-shadow-focus,
+    3px 3px 9px rgba(0, 0, 0, 0.12),
+    -3px -3px 9px rgba(0, 0, 0, 0.12)
+  );`
 };
 
+const selectedPrompt = `
+  background-color: var(--theme-cell-prompt-bg-focus, hsl(0, 0%, 90%));
+  color: var(--theme-cell-prompt-fg-focus, hsl(0, 0%, 51%));
+`;
```
stormpython, line 39: Is there a difference between using `styled.div` as a prefix to styled components and not using a prefix?

rgbkrk: This is just a regular [string literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

rgbkrk:
Hooray, all is passing! Sadly this does make all our `props` take optional arguments that ignore `defaultProps` due to DefinitelyTyped/DefinitelyTyped#29540.

captainsafia (approved):

PR #3859, opened 2018-12-17
Use styled-components in outputs

captainsafia:
Tested on styleguide docs and Jest tests.

### DIFF
--- a/packages/outputs/package.json
+++ b/packages/outputs/package.json
@@ -19,6 +19,7 @@
     "@nteract/mathjax": "^3.0.1",
     "ansi-to-react": "^4.0.0-alpha.0",
     "react-json-tree": "^0.11.0",
+    "styled-components": "^4.1.3",
     "styled-jsx": "^3.1.0"
   },
   "peerDependencies": {
--- a/packages/outputs/src/components/kernel-output-error.tsx
+++ b/packages/outputs/src/components/kernel-output-error.tsx
@@ -54,3 +54,5 @@ export const KernelOutputError = styled(PlainKernelOutputError)`
     white-space: pre-wrap;
   }
 `;
+
+KernelOutputError.displayName = "KernelOutputError";
--- a/packages/outputs/src/components/kernel-output-error.tsx
+++ b/packages/outputs/src/components/kernel-output-error.tsx
@@ -22,13 +22,7 @@ type Props = {
   traceback: Array<string>;
 };
 
-const KernelOutputErrorDiv = styled.div`
-  & code {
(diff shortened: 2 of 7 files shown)


packages/outputs/src/components/kernel-output-error.tsx
```diff
@@ -40,14 +47,9 @@ export const KernelOutputError = (props: Props) => {
 
   return (
     <React.Fragment>
-      <div className="kernel-output-error">
+      <KernelOutputErrorDiv>
```
rgbkrk: The Fragment wrapped around this can be deleted.

Additionally, for the sake of Hydrogen and others if they want to override the style here we could make this whole thing a styled component they can set things on.

rgbkrk:
```jsx
const _KernelOutputError = (props: Props) => {
  // ...
  return <div className={props.classname}><Ansi linkify={false}>{kernelOutputError.join("\n")}</Ansi></div>
}


export const KernelOutputError = styled(_KernelOutputError)`
  & code {
    white-space: pre-wrap;
  }
`
```

packages/outputs/src/components/media/image.tsx
```diff
@@ -16,17 +22,11 @@ export function Image(props: Props) {
...
         src={`data:${props.mediaType};base64,${props.data}`}
         {...size}
       />
-      <style jsx>{`
-        img {
-          display: block;
-          max-width: 100%;
-        }
-      `}</style>
     </React.Fragment>
```
rgbkrk: You can delete the React.Fragment here as well.

packages/outputs/package.json
```diff
@@ -19,6 +19,7 @@
     "@nteract/mathjax": "^3.0.1",
     "ansi-to-react": "^4.0.0-alpha.0",
     "react-json-tree": "^0.11.0",
+    "styled-components": "^4.1.3",
```
willingc, line 22: Is there a reason that the version is different here and in package.json at root?

willingc (commented):
Looks good.

packages/outputs/package.json
captainsafia: Yep. There was a minor fix to their type definitions that was pushed two weeks ago so it is one minor version ahead of the package.

rgbkrk (approved):

packages/outputs/package.json
willingc: Thanks @captainsafia for the explanation.

PR #3860, opened 2018-12-17
Use styled-components in dropdown-menu

captainsafia:
Tested on desktop app and tests.


packages/dropdown-menu/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -2,152 +2,441 @@
 
 exports[`DropdownMenu clicking dropdown content triggers the items callback and closes the menu 1`] = `
 <DropdownMenu>
-  <div
-    className="jsx-3444321073 dropdown"
-  >
-    <DropdownTrigger
-      key=".0"
-      onClick={[Function]}
+  <styled.div>
```
rgbkrk: We probably want to make sure that we apply `displayName`s to our components so that React tools show the name of the component.

rgbkrk (approved):

PR #3861, opened 2018-12-18
Use styled-components in directory-listing

captainsafia:


packages/directory-listing/src/components/name.tsx
```diff
@@ -1,35 +1,31 @@
...
-          }
-
-          .name :global(a:hover) {
-            text-decoration: underline;
-            outline-width: 0;
-          }
-        `}</style>
-      </td>
-    );
+    return <NameTD>{this.props.children}</NameTD>;
```
rgbkrk: Since no other props are used here, the `NameTD` can just be the `Name` component instead of having this extra class.

packages/directory-listing/src/components/listing.tsx
```diff
@@ -12,22 +26,11 @@ export class Listing extends React.Component<ListingProps> {
...
-            border-collapse: collapse;
-            border-radius: 2px;
-            border-spacing: 0;
-          }
-          .listing-root {
-            padding: 0px 0px 20px 0px;
-          }
-        `}</style>
+        </ListingRoot>
       </React.Fragment>
```
rgbkrk: The React Fragment can be removed here.

packages/directory-listing/src/components/listing.tsx
```diff
@@ -1,9 +1,23 @@
...
+  table {
+    width: 100%;
+    border-collapse: collapse;
+    border-radius: 2px;
+    border-spacing: 0;
+  }
+`;
+
+ListingRoot.displayName = "ListingRoot";
+
```
rgbkrk, line 17: Not sure if we need to but we could make this be a `styled.table` and switch to `margin` above (so it's outside the table).

```jsx
const ListingRoot = styled.table`
  margin: 0px 0px 20px 0px;
  width: 100%;
  border-collapse: collapse;
  border-radius: 2px;
  border-spacing: 0;
`;

ListingRoot.displayName = "ListingRoot";
```

packages/directory-listing/src/components/icon.tsx
```diff
@@ -1,11 +1,24 @@
...
 };
 
+const IconTD = styled.td`
+  padding-right: 2px;
+  padding-left: 10px;
+  width: 17px;
+  vertical-align: middle;
+  text-align: center;
+  opacity: 0.95;
+  color: #0366d6;
```
rgbkrk: You could let this override the color with the prop like this:

```suggestion
  color: ${props => props.color || "#0366d6"};
```

packages/directory-listing/src/components/name.tsx
```diff
@@ -1,35 +1,31 @@
...
-            padding: 8px;
-          }
+    return <td>{this.props.children}</td>;
+  }
+}
 
-          .name :global(a) {
-            text-decoration: none;
-          }
+export const Name = styled(PlainName)`
```
rgbkrk: You could make this `styled.td` and delete `PlainName` entirely.

rgbkrk (approved):

PR #3864, opened 2018-12-18
Add GitHub action for deploying component docs

captainsafia:
I think this is how this works?

### DIFF
--- a/.github/deploy-component-docs.workflow
+++ b/.github/deploy-component-docs.workflow
@@ -1,0 +1,9 @@
+workflow "Deploy component docs" {
+  on = "push"
+  resolves = ["GitHub Action for Zeit"]
+}
+
+action "GitHub Action for Zeit" {
+  uses = "actions/zeit-now@15fbbf2"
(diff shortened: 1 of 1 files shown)


.github/deploy-component-docs.workflow
```diff
@@ -0,0 +1,9 @@
+workflow "Deploy component docs" {
+  on = "push"
+  resolves = ["GitHub Action for Zeit"]
+}
+
+action "GitHub Action for Zeit" {
+  uses = "actions/zeit-now@15fbbf2"
+  runs = "yarn docs:deploy && yarn docs:promote"
```
rgbkrk, line 8: I feel like this has to run `yarn` too. Guess we'll see!

rgbkrk: Worst that can happen is we have to delete this workflow or fix it, so let's try it out.

jdetle: I think this needs a build step, just did this and needed to run yarn build first.

PR #3870, opened 2018-12-19
Add @nteract/fs-kernels package

captainsafia:
Woot!


captainsafia:
The tests and build work locally for me. We'll need to set up our CI to be configured for Python and Jupyter.

captainsafia:
Removing the WIP tag since this is just blocked on CI updates and ready for code review.

.circleci/config.yml
```diff
@@ -20,6 +20,8 @@ jobs:
             # fallback to using the latest cache if no exact match is found
             - yarn-packages-
 
+      - run: pip install --user setuptools
```
rgbkrk, line 23: I feel like @willingc would tell me to also upgrade `pip` right here while we're at it.

packages/fs-kernels/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+The example below shows how we can use this package to launch and connect to a new kernel and retrieve usage stats about the kernel's process.
+
+```javascript
+import Kernel from "@nteract/kernel-manager";
+
+export default () => {
+  const kernelName = "python3";
+  const kernel = new Kernel("python3");
+  console.log(kernel.getUsageStats());
```
rgbkrk, line 25: Nice!

packages/fs-kernels/src/jupyter-paths.ts
```diff
@@ -0,0 +1,268 @@
...
+    pushIfExists(paths, sysPathed);
+  }
+  return paths.concat(systemDirs);
+}
+
+function systemDataDirs() {
+  var paths: string[] = [];
+  // System wide for Windows and Unix
+  if (process.platform === "win32") {
+    const defaultProgramDataPath = "C:ProgramData";
```
rgbkrk: Needs more slashing.

packages/fs-kernels/src/jupyter-paths.ts
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * where the userland data directory resides
+ * includes things like the runtime files
+ * @return {string} directory for data
+ */
+function userDataDir() {
+  // Userland specific
+  if (process.platform === "darwin") {
+    return home("Library/Jupyter");
```
rgbkrk, line 186: I can tell this is now cleaner than the original `jupyter-paths` code now. ✨

packages/fs-kernels/src/jupyter-paths.ts
```diff
@@ -0,0 +1,268 @@
...
+    if (systemDirs.indexOf(sysPathed) === -1) {
+      pushIfExists(paths, sysPathed);
+    }
+  }
+  return paths.concat(systemDirs);
+}
+
+async function runtimeDir(opts?: {
+  askJupyter?: () => Promise<JupyterPaths>;
+}): Promise<string> {
```
rgbkrk, line 248: 👌

packages/fs-kernels/src/kernel.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    this.launchedKernel = launchedKernel;
+  }
+
+  async shutdown() {
+    if (this.launchedKernel) {
+      this.launchedKernel.spawn.kill();
+      cleanup(this.launchedKernel.connectionFile);
+      this.launchedKernel = undefined;
+    }
```
rgbkrk, line 40: The only scary thing in my mind about this is that if this fails at any point, the rest of the cleanup won't happen. A case where this matters: Process is already dead (died on its own), connection file still needs to be cleaned up.

In case this goes really poorly and someone needs to work around it, it'd be useful for `this.launchedKernel` to stick around.

Probably more alarming though are the various cases around kill:

> The ChildProcess object may emit an 'error' event if the signal cannot be delivered. Sending a signal to a child process that has already exited is not an error but may have unforeseen consequences. Specifically, if the process identifier (PID) has been reassigned to another process, the **signal will be delivered to that process instead which can have unexpected results**.

rgbkrk: We also need to make sure `stdin`, `stdout`, and `stderr` get closed (if opened). Perhaps we want the users to register their own `onCleanup` that we run?

packages/fs-kernels/src/spawnteract.ts
```diff
@@ -0,0 +1,242 @@
...
+function launch(
+  kernelName: string,
+  spawnOptions?: Options,
+  specs?: KernelResourceByName
+): Promise<LaunchedKernel> {
+  // Let them pass in a cached specs file
+  if (!specs) {
+    return findAll().then((sp: KernelResourceByName) =>
+      launch(kernelName, spawnOptions, sp)
+    );
```
rgbkrk, line 233: Makes me think we could cache the specs ourselves here.

packages/fs-kernels/src/kernel.ts
rgbkrk: It's `.end()` for the streams of stdin, stdout, and stderr. Note:

> if a child process waits to read all of its input, the child will not continue until this stream has been closed via end().

packages/fs-kernels/src/kernelspecs.ts
```diff
@@ -27,6 +27,10 @@ type KernelResource = {
 
 export type KernelResourceByName = { [name: string]: KernelResource };
 
+function flatten(array: Array<any>) {
```
rgbkrk, line 30: I believe the term is [_smoosh_](https://github.com/tc39/proposal-flatMap).

PR #3871, opened 2018-12-19
Converting connected components to styled components

stormpython:
Converts the `connected-components` package to styled components.


stormpython:
@captainsafia, how have you been handling snapshot (in test) failures? From what I can tell, the reason the test fails is because the class names do not match, which is expected with `styled components`.

captainsafia:
Yep! This is normal. The solution is to update the snapshots using the “yarn test -u” command. It’ll rewrite the snapshots with the new class names.

stormpython:
@captainsafia thanks! Updated snapshots and tests were passing locally.

packages/connected-components/src/new-notebook-navigation/index.tsx
```diff
@@ -27,6 +28,85 @@ export type AvailableNotebooks =
...
+  text-overflow: ellipsis;
+`;
+
+const DisplayNameLong = styled.p`
+  margin: 0;
+  color: var(--theme-primary-fg);
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`;
```
rgbkrk: In order to alleviate this error in the CI:

```
src/new-notebook-navigation/index.tsx(51,9): error TS2448: Block-scoped variable 'DisplayNameLong' used before its declaration.
src/new-notebook-navigation/index.tsx(51,9): error TS2454: Variable 'DisplayNameLong' is used before being assigned.
src/new-notebook-navigation/index.tsx(60,9): error TS2448: Block-scoped variable 'DisplayNameLong' used before its declaration.
src/new-notebook-navigation/index.tsx(60,9): error TS2454: Variable 'DisplayNameLong' is used before being assigned.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

move this up above the `NewNotebookDiv` declaration.

stormpython:
@rgbkrk ok, updated!

rgbkrk:
Looking good!

PR #3883, opened 2018-12-20
Add missing TypeScript package dependencies

rgbkrk:
Safia and I noticed some packages were missing dependencies. At the same time, we found strange behavior with the written out packages. For example, if you build `@nteract/epics`, you'll see this output in `lib`:

```
$ ls packages/epics/lib/
actions    commutable epics      messaging  rx-jupyter selectors  types
```


tsconfig.base.json
```diff
@@ -10,16 +10,6 @@
...
-    "moduleResolution": "node",
-    "paths": {
-      // Attempt to allow our namespace packages to get resolved
-      "@nteract/*": ["./packages/*/src"],
-      "@mybinder/*": ["./packages/*/src"],
-      "rx-jupyter": ["./packages/rx-jupyter/src"],
-      "rx-binder": ["./packages/rx-binder/src"],
-      "enchannel-zmq-backend": ["./packages/enchannel-zmq-backend/src"],
-      "fs-observable": ["./packages/fs-observable/src"]
-    }
```
rgbkrk: Let's remember to delete this commit before we merge this. I'll get going on the strangeness of deps here.

packages/notebook-render/src/index.tsx
```diff
@@ -6,7 +6,6 @@ import katex from "rehype-katex";
 import stringify from "rehype-stringify";
 import { InlineMath, BlockMath } from "react-katex";
 import styled, { createGlobalStyle } from "styled-components";
-import flush from "styled-jsx/server";
```
rgbkrk, line 9: I just realized I didn't look at this well in the other PR... We may want to include the styled-components server side rendering bits here (if necessary).

PR #3890, opened 2018-12-21
Adds browser-cache-circumvention to rx-jupyter.

### LINKED ISSUE #3850: Intermittent errors "open in another tab possibly..." in jupyter_extension
macOS
jupyter_extension

I am regularly able to repro this on a Scala kernel while re-running all cells in a notebook of about 5 cells w/ plotly chart outputs. I think an auto-save is initiated while the cells are calculating, and things get out of sync.

Once it happens it prevents all future saves.

I can exacerbate the issue by manually clicking the save menu item while recalculating.

I've been able to repro less frequently w/ Python, though it does happen on occasion.

jameskoch:
This is one of several ways we might fix #3850.

In this case we assume that Jupyter server will not change its ways
regarding lack of Cache-control headers on APIs. So we augment the
querystring of GET and HEAD requests w/ a timestamp to ensure the
request makes it to the server. This is how Jupyter notebook does it.

I mechanically went through https://github.com/jupyter/notebook/tree/master/notebook/static/services
and found usages of jquery's ajax method w/ its `cache` option set to
false. Notably this was occurring on several PUT/POST/DELETE calls;
jquery docs say that doesn't make sense except for IE8, and their
code actually NOOPs these cases, so I've ommitted them.

I altered createAJAXSettings to behave much like jquery's ajax().
This creates a familiar API (for many), and centralizes the cache-
busting logic. It's notable however that in many cases we are now
assembling a URL string, only to pass it into createAJAXSettings
where it is parsed, the cache-busting query section added, before
a full string is finally rebuilt.

### DIFF
--- a/packages/rx-jupyter/src/base.ts
+++ b/packages/rx-jupyter/src/base.ts
@@ -20,17 +20,23 @@ export interface ServerConfig extends Partial<AjaxRequest> {
  *
  * @param serverConfig Details about the server to connect to.
  * @param uri The URL to connect to, not including the base URL
- * @param opts A set of options to pass to the AJAX request.
+ * @param opts A set of options to pass to the AJAX request. We mimic jquery.ajax support of a cache option here.
  *
  * @returns A fully-configured AJAX request for connecting to the server.
  */
 export const createAJAXSettings = (
   serverConfig: ServerConfig,
   uri = "/",
-  opts: Partial<AjaxRequest> = {}
+  opts: Partial<AjaxRequest & { cache?: boolean }> = {}
 ): AjaxRequest => {
   const baseURL = normalizeBaseURL(serverConfig.endpoint || serverConfig.url);
-  const url = `${baseURL}${uri}`;
+  let url = `${baseURL}${uri}`;
+  if (opts.cache === false) {
+    const parsed = new URL(url);
+    parsed.searchParams.set("_", Date.now().toString());
+    url = parsed.href;
+  }
+
   // Use the server config provided token if available before trying cookies
   const xsrfToken = serverConfig.xsrfToken || Cookies.get("_xsrf");
   const headers = {
@@ -50,5 +56,6 @@ export const createAJAXSettings = (
     headers: { ...headers, ...opts.headers }
   };
   delete settings.endpoint;
+  delete settings.cache;
   return settings;
 };
--- a/packages/rx-jupyter/__tests__/contents-spec.ts
+++ b/packages/rx-jupyter/__tests__/contents-spec.ts
@@ -25,9 +25,9 @@ describe("contents", () => {
         "/walla/walla/bingbang.ipynb"
       ) as AjaxObservable;
       const request = content$.request;
-      expect(request.url).toBe(
-        "http://localhost:8888/api/contents/walla/walla/bingbang.ipynb"
-      );
+      expect(request.url).toEqual(expect.stringContaining(
+        "http://localhost:8888/api/contents/walla/walla/bingbang.ipynb?_="
+      ));
       expect(request.method).toBe("GET");
       expect(request.crossDomain).toBe(true);
       expect(request.responseType).toBe("json");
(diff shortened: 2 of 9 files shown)


packages/rx-jupyter/src/base.ts
```diff
@@ -20,17 +20,23 @@ export interface ServerConfig extends Partial<AjaxRequest> {
...
+  opts: Partial<AjaxRequest & { cache?: boolean }> = {}
 ): AjaxRequest => {
   const baseURL = normalizeBaseURL(serverConfig.endpoint || serverConfig.url);
-  const url = `${baseURL}${uri}`;
+  let url = `${baseURL}${uri}`;
+  if (opts.cache === false) {
+    const parsed = new URL(url);
+    parsed.searchParams.set("_", Date.now().toString());
+    url = parsed.href;
+  }
```
rgbkrk, line 38: At least this will make sense for folks within our ecosystem.

packages/rx-jupyter/src/base.ts
```diff
@@ -50,5 +56,6 @@ export const createAJAXSettings = (
     headers: { ...headers, ...opts.headers }
   };
   delete settings.endpoint;
+  delete settings.cache;
```
rgbkrk, line 59: 👍

rgbkrk:
In what instances should we opt in to caching? Should we default to not caching and not passing this param over in `rx-jupyter`?

rgbkrk:
Based on the test failures it looks like more of the `expect.stringContaining` logic needs to be sprinkled around (maybe as our own test helper?):

```
  ● sessions › get › creates an AjaxObservable for getting particular session info

    expect(received).toBe(expected) // Object.is equality

    Expected: "http://localhost:8888/api/sessions/uuid"
    Received: "http://localhost:8888/api/sessions/uuid?_=1545422769351"

      24 |       const session$ = sessions.get(serverConfig, "uuid") as AjaxObservable;
      25 |       const request = session$.request;
    > 26 |       expect(request.url).toBe("http://localhost:8888/api/sessions/uuid");
         |                           ^
      27 |       expect(request.method).toBe("GET");
      28 |       expect(request.crossDomain).toBe(true);
      29 |       expect(request.responseType).toBe("json");
```

jameskoch:
> In what instances should we opt in to caching? Should we default to not caching and not passing this param over in `rx-jupyter`?

Thinking only about browser-based apps for a second, I'm hopeful that over time these issues just get solved in Jupyter's server. It seems pretty easy to either mark non-cacheable content as such, or to make sure it sends proper headers so that Etags and/or last-modified dates are honored. 

Once such fixes land it would seem odd for rx-jupyter to ever add caching of its own; any alternative caching strategy probably involves accepting additional risk for out-of-date content, and so should live in a higher-order component w/ more knowledge about the overall system. If this all makes sense, then I would keep the code as-is, only opting out of caching in the cases where Jupyter gets it wrong; then we delete that code when they get it right.

Now I don't really know anything about nodejs and whether it applies standard caching behaviors to XMLHttpRequest. If not, maybe there's more to consider here.

I'm not super confident in any of this :)

jameskoch:
> Based on the test failures it looks like more of the `expect.stringContaining` logic needs to be sprinkled around (maybe as our own test helper?):
> 
> ```
>   ● sessions › get › creates an AjaxObservable for getting particular session info
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: "http://localhost:8888/api/sessions/uuid"
>     Received: "http://localhost:8888/api/sessions/uuid?_=1545422769351"
> 
>       24 |       const session$ = sessions.get(serverConfig, "uuid") as AjaxObservable;
>       25 |       const request = session$.request;
>     > 26 |       expect(request.url).toBe("http://localhost:8888/api/sessions/uuid");
>          |                           ^
>       27 |       expect(request.method).toBe("GET");
>       28 |       expect(request.crossDomain).toBe(true);
>       29 |       expect(request.responseType).toBe("json");
> ```

I took a quick look at doing a custom matcher, but [volume of code](https://stackoverflow.com/a/45745719) felt intense for something that we want to eventually go away.  So I just sprinkled around the code some more. Open to push-back on that of course.

rgbkrk:
> Now I don't really know anything about nodejs and whether it applies standard caching behaviors to XMLHttpRequest. If not, maybe there's more to consider here.

If `rx-jupyter` gets used server side, it won't have any effect to have the extra query params. That may depend on the `XMLHttpRequest` implementation, though it's outside the scope of this library.

rgbkrk:
> I took a quick look at doing a custom matcher, but volume of code felt intense for something that we want to eventually go away. So I just sprinkled around the code some more. Open to push-back on that of course.

Sprinkling it around for now is just fine.

PR #3891, opened 2018-12-22
Message collections

rgbkrk:
This aggregates collections of messages after a kernel is launched.

screen shot 2018-12-21 at 10 28 16 pm

Since I want to help get `master` unstuck for app builds, I'm just posting this for others to look at.

### DIFF
--- a/packages/kernel-relay/src/index.ts
+++ b/packages/kernel-relay/src/index.ts
@@ -18,12 +18,18 @@ const Types = gql`
     id: ID!
     status: String
   }
+
+  type Message {
+    id: ID!
+    payload: JSON
+  }
 `;
 
 const Query = gql`
   type Query {
     listKernelSpecs: [KernelSpec!]!
     running: [KernelSession!]!
+    messages: [Message!]!
   }
 `;
 
@@ -92,6 +98,9 @@ const resolvers = {
           ...kernelspecs[key]
         };
       });
+    },
+    messages: () => {
+      return ([] as JupyterMessage[]).concat(...Object.values(messages));
     },
     running: () => {
       return Object.keys(kernels).map(id => ({ id, status: "pretend" }));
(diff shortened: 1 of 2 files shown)


packages/kernel-relay/src/index.ts
```diff
@@ -12,12 +18,18 @@ const Types = gql`
...
+    id: ID!
+    payload: JSON
+  }
 `;
 
 const Query = gql`
   type Query {
     listKernelSpecs: [KernelSpec!]!
     running: [KernelSession!]!
+    messages: [Message!]!
```
captainsafia, line 32: What do the two exclamation points mean here? You can have null value in the array?

packages/kernel-relay/src/index.ts
```diff
@@ -33,27 +45,65 @@ type StartKernel = {
...
+      const subscription = kernel.channels.subscribe(
+        (message: JupyterMessage) => {
+          messages[id].push(message);
+          console.log(message);
+        }
+      );
+
+      // THOUGHT: Should we track our own messages sent out too?
+      //          Hmmmm
+      kernel.channels.next(kernelInfoRequest());
```
captainsafia: Me thinks we should track outbound messages.

packages/kernel-relay/src/index.ts
rgbkrk: It means you can't have null values in the array and the array itself can't be null.

packages/kernel-relay/src/index.ts
rgbkrk: I agree. Oooh. This makes me realize that some of the magic that enchannel does we'll need to explicitly set here. Enchannel applies generated session IDs and other header info if not set.

rgbkrk: Done.

rgbkrk:
Addressed initial feedback, this is looking good again. 😄 

screen shot 2019-01-15 at 7 05 04 pm

captainsafia (approved):
Looks good! Very excited to get this in.

rgbkrk:
🎉

PR #3894, opened 2018-12-22
Add missing nteractDesktop fields

lgeiger:
The `nteractDesktop` main field is used by babel/webpack to resolve modules to `src` instead of requiring a from `lib`. This means we can hack on the apps without needing to constantly run typescript compilation.

This fixes the webpack build of the desktop app.


rgbkrk:
This might end up fixing the jupyter extension and others too.

lgeiger:
> This might end up fixing the jupyter extension and others too.

Yes.

rgbkrk:
Thank you!

PR #3895, opened 2018-12-22
Bump electron-{builder,updater}

lgeiger:
This fixes auto update warnings of the desktop app during development.


rgbkrk:
Oooh I was wondering about this.

package.json
```diff
@@ -161,10 +161,10 @@
     "cross-env": "^5.2.0",
     "css-loader": "^1.0.0",
     "electron": "3.0.13",
-    "electron-builder": "20.38.3",
+    "electron-builder": "^20.38.4",
```
rgbkrk, line 164: IIRC @jameskoch was pinning this to keep builds reproducible, we might want to keep this pinned.

PR #3896, opened 2018-12-22
include a root level tsconfig

rgbkrk:
Lets us run:

```
tsc -b
```

at the root. I've also noticed some various auditing tools like `rxjs-operator-counter` will use `tsconfig.json` at the root to do its thing, so this is here for ease of detection.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,6 @@
+{
+  "files": [],
+  "references": [
+    { "path": "packages" }
+  ]
+}


tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "files": [],
+  "references": [
+    { "path": "packages" }
```
rgbkrk, line 4: We can probably add applications to this setup shortly after types are all set.

rgbkrk:
This got done in another PR.

PR #3898, opened 2018-12-22
switch menu to using styled-components directly

rgbkrk:
This doesn't fix anything but does make us have less interfaces to maintain. 😄


lgeiger (approved):
Thanks for the cleanup 👍

PR #3899, opened 2018-12-22
GraphQL: lay foundation for rx-backed messaging

rgbkrk:
Extracted from #3891 so we can start using it in other contexts (@captainsafia is using it for shutdown message logic).

* provide channels on kernel object
* include JSON scalars for untyped jupyter payloads

### DIFF
--- a/packages/kernel-relay/src/index.ts
+++ b/packages/kernel-relay/src/index.ts
@@ -2,7 +2,13 @@ import { ApolloServer, gql } from "apollo-server";
 
 import { findAll, launchKernel, Kernel } from "@nteract/fs-kernels";
 
+import { JupyterMessage } from "@nteract/messaging";
+
+import GraphQLJSON from "graphql-type-json";
+
 const Types = gql`
+  scalar JSON
+
   type KernelSpec {
     id: ID!
     name: String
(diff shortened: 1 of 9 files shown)


packages/kernel-relay/src/index.ts
```diff
@@ -51,21 +67,25 @@ const resolvers = {
...
       const kernel = await launchKernel(args.name);
 
+      console.log("kernel launched", kernel);
+
+      // NOTE: we should generate IDs
+      // We're also setting a session ID within the enchannel-zmq setup, I wonder
+      // if we should use that
+      const id = kernel.connectionInfo.key;
+
+      messages[id] = [];
```
captainsafia: It looks like CI is failing on this because the messages global isn't defined.

rgbkrk: _/me whistles_

packages/kernel-relay/tsconfig.json
```diff
@@ -5,5 +5,9 @@
     "rootDir": "src"
   },
   "include": ["src"],
-  "references": [{ "path": "../fs-kernels" }]
+  "references": [
+    { "path": "../fs-kernels" },
+    { "path": "../enchannel-zmq-backend" },
+    { "path": "../messaging" }
```
captainsafia, line 11: Have these been added to package.json?

rgbkrk: All set now!

PR #3906, opened 2018-12-26
Add shutdownEpic to Kernel in fs-kernels

captainsafia:


packages/enchannel-zmq-backend/src/index.ts
```diff
@@ -230,7 +230,7 @@ export const createMainChannelFromSockets = (
       Object.keys(sockets).forEach(name => {
         const socket = sockets[name];
         socket.removeAllListeners();
-        socket.close();
+        if (socket.close) socket.close();
```
captainsafia, line 237: This is a bit of a code smell. The tests kept failing because `close` was not defined on the socket but it seemed like Jest wasn't fully mocking out the JMP Socket object.

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,115 @@
...
+
+    // Try to make a shutdown request
+    // If we don't get a response within X time, force a shutdown
+    // Either way do the same cleanup
+    const shutDownHandling = this.channels.pipe(
+      childOf(request),
+      ofMessageType("shutdown_reply"),
+      first(),
+      // If we got a reply, great! :)
+      map((msg: { content: { restart: boolean } }) => {
```
captainsafia, line 86: Other places in the code ([here](https://github.com/nteract/nteract/blob/master/applications/desktop/src/notebook/epics/zeromq-kernels.js#L358) and [here](https://github.com/nteract/nteract/blob/9ddcc138b7ad75144adec81382170a45aff53f26/packages/actions/src/actions/kernels.ts#L276))seem to assume that content is a string when it is really the object defined above. Is there a reason for this?

types/jmp/index.d.ts
```diff
@@ -44,5 +44,7 @@ declare module "jmp" {
     unmonitor(): this;
 
     removeAllListeners(): this;
+
+    close(): this;
```
captainsafia, line 48: The ambient type declaration for JMP was incorrect here.

packages/enchannel-zmq-backend/src/index.ts
rgbkrk: I wonder if this is related to how jmp provides an interface for direct ZeroMQ communication as well as an event emitter. I look forward to coming back to clean up the jmp bits, so no worries here.

packages/fs-kernels/src/kernel.ts
rgbkrk: I think you've captured it best here, which is that it should be an object. Looks like the old code was incorrect.

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,114 @@
...
+    this.id = createKernelRef();
     this.process = launchedKernel.spawn;
     this.connectionInfo = launchedKernel.config;
     this.kernelSpec = launchedKernel.kernelSpec;
     this.connectionFile = launchedKernel.connectionFile;
     this.channels = launchedKernel.channels;
   }
 
-  async shutdown() {
+  shutdownEpic() {
```
rgbkrk: We could make the timeout configurable by passing it in here.

captainsafia: Oooooh. Good idea!

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,114 @@
...
     cleanup(this.connectionFile);
     if (!this.process.killed && this.process.pid) {
       process.kill(this.process.pid);
     }
     this.process.removeAllListeners();
   }
 
+  async shutdown() {
+    const observable = this.shutdownEpic();
+    return observable.pipe(toArray()).toPromise();
```
rgbkrk: Slick. I think you can switch this to:

```
return observable.pipe(first()).toPromise();
```

Then we can declare the promise as returning either the success object or the error object types here.

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,114 @@
...
+      first(),
+      // If we got a reply, great! :)
+      map((msg: { content: { restart: boolean } }) => {
+        return {
+          content: msg.content,
+          id: this.id
+        };
+      }),
+      // If we don't get a response within 2s, assume failure :(
+      timeout(1000 * 2),
```
rgbkrk: This seems comfortable to provide the timeout as a parameter to this function, defaulting to the (totally made up by me) 2s.

packages/fs-kernels/src/kernel.ts
rgbkrk: Oh I see, nevermind. There are follow on events there.

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,114 @@
...
+      catchError(err => of({ error: err, id: this.id })),
+      mergeMap(async action => {
```
rgbkrk: Probably good to include a comment here stating: "Whether the `shutdown_request` got a response or not, go ahead and clean up the rest".

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,114 @@
...
+        // End all communication on the channels
+        this.channels.complete();
+
+        await this.shutdownProcess();
+
+        return merge(
```
rgbkrk: This all can be a

```js
from([
  action,
  {id: this.id},
  {status: "shutting down", id: this.id}
])
```

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,114 @@
...
+          // Pass on our intermediate action (whether or not kernel ACK'd shutdown request promptly)
+          of(action),
```
rgbkrk: `action` isn't defined here (which I assume is a cargo cult from the original shutdown epic). Should it just be omitted?

captainsafia: Oh darn it! This is totally my bad. Hang on let me clean this bit up.

packages/fs-kernels/__tests__/kernel.spec.ts
```diff
@@ -1,24 +1,44 @@
+import { toArray } from "rxjs/operators";
+
 import { launchKernel } from "../src/kernel";
+import { DebugLogger } from "builder-util";
```
rgbkrk: I'm guessing this was a VS Code autoimport to remove.

captainsafia: Yes.

packages/fs-kernels/src/kernel.ts
```diff
@@ -4,36 +4,108 @@
...
+       */
+      mergeMap(async () => {
+        // End all communication on the channels
+        this.channels.complete();
+        await this.shutdownProcess();
+        return of({ status: "shutdown", id: this.id });
+      }),
+      catchError(err =>
+        // Catch all, in case there were other errors here
+        of({ error: err, id: this.id, status: "error" })
```
rgbkrk: I think we should get rid of all the `kernelRef` logic out of here for a few reasons:

* `ref`s are for the frontend to denote resources that may still be in creation (prior to a POST call)
* The user of this API is calling this API directly to shutdown and already has a reference to our kernel object itself. They can choose how to route the response messages.
* For a user of `ref`s, they expect to be able to create it themselves and re-use it

packages/fs-kernels/src/kernel.ts
rgbkrk: Now I realize since there are multiple events and that we'd want to expose them for transparency, we should either return the observable or possibly make an async iterator(?).

packages/fs-kernels/src/kernel.ts
```diff
@@ -26,14 +34,67 @@ export class Kernel {
...
+       * If we don't get a response within timeoutMs, then throw an error.
+       */
+      timeout(timeoutMs),
+      catchError(err => of({ error: err })),
+      /**
+       * Even if we don't receive a shutdown_reply from the kernel to our
+       * shutdown_request, we will go forward with cleaning up the RxJS
+       * subject and killing the kernel process.
+       */
+      mergeMap(async () => {
```
rgbkrk: In this function we should probably propagate the error from above, even though we're also going to force shut down fully. Something like:

```js
mergeMap(async (event) => {
  this.channels.complete();
  await this.shutdownProcess();

  const finalResponse = {status: "shutdown"};
  if(message.error) {
    finalResponse.error = message.error;
  }
  return of(finalResponse);
}); 
```

Not the cleanest, but does provide transparency for failure here.

packages/fs-kernels/src/kernel.ts
```diff
@@ -26,14 +34,84 @@ export class Kernel {
...
+      ),
+      catchError(err =>
+        // Catch all, in case there were other errors here
+        of({ error: err, status: "error" })
+      )
+    );
+
+    // On subscription, send the message
+    return Observable.create((observer: Observer<any>) => {
+      const subscription = shutDownHandling.subscribe(observer);
```
rgbkrk, line 134: The `send` is missing here now.

rgbkrk:
Let's keep this moving and start using it since `kernel-relay` is just getting started.

Potential follow ups I see for `fs-kernels`:

* Integrate it into the desktop app for the zeromq-kernels epic
* Integrate it into CoCalc's code for spawning kernels (note: this one requires a release from us)

PR #3909, opened 2018-12-27
de-any save file contents

rgbkrk:
More strictly type `SAVE_FULFILLED` reducer.


rgbkrk:
I've got another commit on the way that's a bit bigger and hairier, I'd love to merge this first though.

captainsafia (approved):

PR #3911, opened 2018-12-27
Set up TSLint checking

captainsafia:
This adds tslint to our codebase.

I migrated our eslint checks and used the tslint:recommended option elsewhere.

This enforces things like alphabetized members within objects, using interfaces instead of types for multi-key type definitions, combining multiple imports from a module into one, favoring `T[]` instead of `Array<T>` as a type for simple array types, not using `Object` types anywhere, and unnecessary type assertions. 

If everyone is cool with this configuration, I can start fixing up the problems it found.


jdetle:
Should we add a circleci command to lint? I'm in favor of this now that I understand better what TSLint is providing.

captainsafia:
> Should we add a circleci command to lint?

Yeah. I can add it. It's all set to warn right now so it would break CI but I can figure out what things would be good to error on.

CONTRIBUTING.md
```diff
@@ -7,23 +7,23 @@ dinner, but most importantly: read it thoroughly!
 
 ### JavaScript side
 
-You'll need knowledge of JavaScript (ES6), React, RxJS, Redux, and Flow to
+You'll need knowledge of JavaScript (ES6), React, RxJS, Redux, and TypeScript to
```
rgbkrk, line 10: 🎉

tslint.json
```diff
@@ -0,0 +1,22 @@
...
+  ],
+  "jsRules": {},
+  "rules": {
+    "no-unnecessary-type-assertion": true,
+    "no-debugger": true,
+    "no-duplicate-variable": true,
+    "ordered-imports": false,
+    "no-duplicate-imports": true,
+    "max-line-length": false,
+    "trailing-comma": false,
```
rgbkrk, line 17: As long as these match our prettier config I'm happy. Oh wait, add in [`tslint-config-prettier`](https://www.npmjs.com/package/tslint-config-prettier).

rgbkrk (approved):
I'm in favor of adopting this as it will allow people to clean up things as small PRs to get comfortable with the code base at large.

tslint.json
rgbkrk: I'm going to merge so we can start using this, feel free to make a fast follow.

PR #3912, opened 2018-12-27
include tslint-config-prettier and disable member ordering

rgbkrk:
Keeps our [tslint in line with our prettier](https://www.npmjs.com/package/tslint-config-prettier).

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,6 +219,7 @@
     "styled-jsx": "^3.1.0",
     "ts-jest": "^23.10.4",
     "tslint": "^5.12.0",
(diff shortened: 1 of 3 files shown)


rgbkrk:
Ok, I've disabled member-ordering for now to eliminate drastic changes with a `tslint` fix. It's ok for someone to bring this back in. The only place we need to keep things in a particular order is when writing jupyter formats (cells, outputs, notebooks) to disk. That helps minimize diffs of notebooks.

rgbkrk:
My local change to `jsx-boolean-value` didn't seem to take effect so I won't push that up as part of this.

PR #3913, opened 2018-12-27
Create dynamic type for jupyter contents API payloads

rgbkrk:
Definitely a work in progress, trying to lock down the types for contents API responses here.

### DIFF
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -53,7 +53,7 @@ export function fetchContentEpic(
         .pipe(
           tap(xhr => {
             if (xhr.status !== 200) {
-              throw new Error(xhr.response as string);
+              throw new Error(xhr.response.toString());
             }
           }),
           map(xhr => {
@@ -244,9 +244,7 @@ export function saveContentEpic(
 
         // This could be object for notebook, or string for files
         let serializedData: Notebook | string;
-        let saveModel: Partial<
-          contents.IContent<"file" | "notebook", true>
-        > = {};
+        let saveModel: Partial<contents.IContent<"file" | "notebook">> = {};
         if (content.type === "notebook") {
           const appVersion = selectors.appVersion(state);
 
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -53,7 +53,7 @@ export function fetchContentEpic(
         .pipe(
           tap(xhr => {
             if (xhr.status !== 200) {
-              throw new Error(xhr.response);
+              throw new Error(xhr.response as string);
             }
           }),
           map(xhr => {
@@ -244,7 +244,9 @@ export function saveContentEpic(
 
         // This could be object for notebook, or string for files
         let serializedData: Notebook | string;
-        let saveModel: Partial<contents.Payload> = {};
+        let saveModel: Partial<
+          contents.IContent<"file" | "notebook", true>
+        > = {};
         if (content.type === "notebook") {
           const appVersion = selectors.appVersion(state);
 
@@ -314,8 +316,14 @@ export function saveContentEpic(
               mergeMap(xhr => {
                 // TODO: What does it mean if we have a failed GET on the content
                 if (xhr.status !== 200) {
-                  throw new Error(xhr.response);
+                  throw new Error(xhr.response.toString());
                 }
+                if (typeof xhr.response === "string") {
+                  throw new Error(
+                    `jupyter server response invalid: ${xhr.response}`
+                  );
+                }
+
                 const model = xhr.response;
 
                 const diskDate = new Date(model.last_modified);
--- a/packages/rx-jupyter/src/base.ts
+++ b/packages/rx-jupyter/src/base.ts
@@ -1,8 +1,9 @@
+/** tslint:disable:jsdoc-format */
 /**
  * @module rx-jupyter
  */
 import Cookies from "js-cookie";
-import { AjaxRequest } from "rxjs/ajax";
+import { AjaxRequest, AjaxResponse } from "rxjs/ajax";
 
 export const normalizeBaseURL = (url = "") => url.replace(/\/+$/, "");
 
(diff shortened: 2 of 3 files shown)


packages/rx-jupyter/src/base.ts
```diff
@@ -59,3 +60,25 @@ export const createAJAXSettings = (
...
+ *
+ * This interface is meant to be internal, used as:
+ *
+ * ```
+   ajax('/api/contents') as Observable<JupyterAjaxResponse<Contents>>
+   ```
+ * or as
+ * ```
+   ajax('/api/contents').map(resp => {
+     if(resp.response.type === "notebook")
```
rgbkrk, line 75: Need to come back to finish this comment, got a little distracted trying to figure out how to tell`tslint` how to respect inline code blocks in tsdoc (it should accept GFM, and VS Code will render this inline).

packages/rx-jupyter/src/contents.ts
```diff
@@ -4,18 +4,53 @@
...
+   * Name of file or directory, equivalent to the last part of the path ,
+   */
+  name: string;
+  path: string;
+  type: FT;
+  writable: boolean;
+  created: string;
+  last_modified: string;
+  mimetype: string;
+  content: Contentful extends false
```
rgbkrk: This is a [conditional type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) from typescript that can check that something is a particular type or even a literal (as is done here by checking for `false`). Down below we check for string literals for `FT` in order to determine the content model type. Note that `type` in this interface above is also `FT`. This reduces boilerplate yet makes this a bit lengthy. I'm curious what people think.

The Jupyter API will return contents with the `content` field set to `null` in these cases:

* When `?content=0` is set as a query param on a `GET`
* When the content is an entry in a directory listing

Otherwise the match up goes like this:

* `notebook` -> Notebook JSON
* `file` -> raw string of file
* `directory` -> `Content[]`

packages/rx-jupyter/src/contents.ts
```diff
@@ -4,18 +4,53 @@
...
+    ? null
+    : FT extends "file"
+    ? string
+    : FT extends "notebook"
+    ? Notebook
+    : FT extends "directory"
+    ? Array<IContent<FileType>>
```
rgbkrk, line 56: Typescript (or tslint?) wouldn't let me write `IContent<FileType>[]`, saying that I needed to use `Array` for more complex types.

jasongrout:
In case it's useful, here are some of the typings we did a long time ago for the contents api. There may be better ways of doing the typings now with newer versions of ts, and it'd be good to see if there was anything we missed as well: https://github.com/jupyterlab/jupyterlab/blob/4797856baf212dd3c62d325a7f6958de0bb4cdff/packages/services/src/contents/index.ts#L37-L183 (validation functions are in https://github.com/jupyterlab/jupyterlab/blob/4797856baf212dd3c62d325a7f6958de0bb4cdff/packages/services/src/contents/validate.ts).

It may be good to have a separate server rest api typings library jlab and nteract can both share.

packages/epics/src/contents.ts
```diff
@@ -53,7 +53,7 @@ export function fetchContentEpic(
         .pipe(
           tap(xhr => {
             if (xhr.status !== 200) {
-              throw new Error(xhr.response);
+              throw new Error(xhr.response as string);
```
captainsafia: Is it better to use the `toString` method here as in L319 below?

packages/epics/src/contents.ts
```diff
@@ -244,7 +244,9 @@ export function saveContentEpic(
 
         // This could be object for notebook, or string for files
         let serializedData: Notebook | string;
-        let saveModel: Partial<contents.Payload> = {};
+        let saveModel: Partial<
+          contents.IContent<"file" | "notebook", true>
```
captainsafia: Is there a way to change the second generic to an enum or something to avoid a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)?

Edit: I see how this works with the description of conditional typing you wrote below. It still feels weird to read without knowing what the "true" is doing but I guess people can figure that out on Code through the type overlays. I still feel meh about it.

packages/rx-jupyter/src/contents.ts
captainsafia: Yep. That's tslint. We have a rule that requires you to use Array<> for complex type definitions and [] for simple type definitions.

packages/rx-jupyter/src/contents.ts
```diff
@@ -95,20 +122,21 @@ export const get = (
  *
  * @returns An Observable with the request response
  */
-export const update = (
+export function update<FT extends FileType>(
```
captainsafia, line 136: :+1: to this because it makes our TypeDoc setup easier.

captainsafia (commented):
Coolio. Posted some comments in-line.

packages/epics/src/contents.ts
rgbkrk: This is great to point out, especially the article. I'll make an enum or use a literal for this.

packages/epics/src/contents.ts
rgbkrk: Yeah let's keep it consistent.

rgbkrk:
> It may be good to have a separate server rest api typings library jlab and nteract can both share.

@jasongrout - I agree, that'd be great. It would be wonderful if it's just the typing of the responses and error payloads (so we can continue using RxJS's `ajax` and others can use whatever they want, including `fetch`).

captainsafia (approved):

jasongrout:
> It would be wonderful if it's just the typing of the responses and error payloads (so we can continue using RxJS's `ajax` and others can use whatever they want, including `fetch`).

Absolutely.

jasongrout:
> It would be wonderful if it's just the typing of the responses and error payloads (so we can continue using RxJS's `ajax` and others can use whatever they want, including `fetch`).

I think it should include both typings for the responses and errors, as well as runtime validation functions so you can actually check the messages before casting them to a specific type.

PR #3917, opened 2018-12-29
correct the tsconfig for jupyter extension

rgbkrk:
This sets up the right path for the jupyter extension. However, type checking will fail now (even though we don't hook it up to CI yet). To run it, do:

```
npx tsc -b applications/jupyter-extension
```

41 errors, we can take them one at a time with new PRs. 😄


stormpython:
LGTM!

stormpython (commented):
LGTM!

PR #3923, opened 2018-12-29
correct the types for the config merge action

rgbkrk:
This makes Desktop run properly again. 😄


rgbkrk:
I'm going to go ahead and merge so we're in a little better of a state this weekend.

PR #3927, opened 2018-12-30
remove null state generic for tsx

rgbkrk:
Since I've noticed `connect`ed components run into a problem with `null` being declared for a component's type of state frequently, I'm trying out removing all of them from our code base. This was a flow quirk that wanted the `null` at the end and we don't need it any more.


rgbkrk:
```
🤖 Application Error Report 🤖

Jupyter Extension: 25
Desktop:           13
Play:              28
```

PR #3928, opened 2018-12-30
Fix catchAndRetry parameter types

captainsafia:
Ended up chasing a lot of red herrings with this one but the problem ended up being that the type on `catchAndRetry` didn't match up with the type definitions for `catchError`.

### DIFF
--- a/applications/desktop/src/notebook/epics/index.ts
+++ b/applications/desktop/src/notebook/epics/index.ts
@@ -1,3 +1,4 @@
+import { Observable } from "rxjs";
 import { catchError, startWith } from "rxjs/operators";
 import { epics as coreEpics } from "@nteract/core";
 import { DesktopNotebookAppState } from "../state";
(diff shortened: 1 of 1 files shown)


applications/desktop/src/notebook/epics/index.ts
```diff
@@ -26,10 +27,7 @@ import { closeNotebookEpic } from "./close-notebook";
 
 import { Actions } from "../actions";
 
-export function retryAndEmitError(
-  err: Error,
-  source: ActionsObservable<Actions>
-) {
+export function retryAndEmitError(err: Error, source: Observable<Actions>) {
```
rgbkrk, line 30: Ohhhhhh

rgbkrk (approved):

rgbkrk:
This shows a stark contrast between our flow typings and our typescript typings.

rgbkrk:
Thank you, this was a toughie to figure out.

PR #3934, opened 2019-01-01
Update desktop tsconfig.json and fix types

captainsafia:
In conjunction with https://github.com/nteract/nteract/pull/3933, this should clear out all the errors in the desktop TS build.


rgbkrk:
Nice!

PR #3936, opened 2019-01-01
Fix some TS build errors in jupyter-extension

captainsafia:
There's 6 errors left to take care of. They are taking me a bit longer to work through so I'm submitting them in a separate PR.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/components/themed-logo.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/components/themed-logo.tsx
@@ -2,8 +2,8 @@ import * as React from "react";
 import { WideLogo } from "@nteract/logos";
 
 type ThemedLogoProps = {
-  height: number;
-  theme: "light" | "dark";
+  height?: number;
+  theme?: "light" | "dark";
 };
 
 const ThemedLogo = (props: ThemedLogoProps) => (
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
@@ -58,15 +58,16 @@ function main(rootEl: Element, dataEl: Node | null) {
   let config: JupyterConfigData;
 
   try {
-    config = JSON.parse(dataEl.textContent || "");
+    if (!dataEl.textContent) {
+      throw new Error("Unable to find Jupyter config data.");
+    }
+    config = JSON.parse(dataEl.textContent);
   } catch (err) {
     ReactDOM.render(<ErrorPage error={err} />, rootEl);
     return;
   }
 
   // Allow chunks from webpack to load from their built location
-  // eslint-disable-next-line no-unused-vars
-  var __webpack_public_path__: string;
   __webpack_public_path__ = urljoin(config.assetUrl, "nteract/static/dist/");
 
   const jupyterHostRecord = makeJupyterHostRecord({
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
@@ -21,6 +21,7 @@ import {
   makeJupyterHostRecord
 } from "@nteract/core";
 import { AppState } from "@nteract/core";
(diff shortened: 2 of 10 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/components/themed-logo.tsx
```diff
@@ -2,8 +2,8 @@ import * as React from "react";
 import { WideLogo } from "@nteract/logos";
 
 type ThemedLogoProps = {
-  height: number;
-  theme: "light" | "dark";
+  height?: number;
+  theme?: "light" | "dark";
```
rgbkrk, line 6: Oh yeah, that annoying lack of recognition of `defaultProps` from TypeScript. I wonder if we should use `!`.

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
```diff
@@ -57,15 +58,15 @@ function main(rootEl: Element, dataEl: Node | null) {
   let config: JupyterConfigData;
 
   try {
-    config = JSON.parse(dataEl.textContent);
+    config = JSON.parse(dataEl.textContent || "");
```
rgbkrk: If this doesn't validate, we definitely want to show an error page. Is this handling the case of `dataEl.textContent` being null? Order here should probably be something similar to:

```
const rawConfig = dataEl.textContent || "";
if(rawConfig === "") {
  throw new Error("Unable to find jupyter config data");
}
config = JSON.parse(rawConfig);
```

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
```diff
@@ -57,15 +58,15 @@ function main(rootEl: Element, dataEl: Node | null) {
...
   } catch (err) {
     ReactDOM.render(<ErrorPage error={err} />, rootEl);
     return;
   }
 
   // Allow chunks from webpack to load from their built location
   // eslint-disable-next-line no-unused-vars
-  declare var __webpack_public_path__: string;
+  var __webpack_public_path__: string;
```
rgbkrk: This was a flow workaround iirc while letting us set this global for webpack. In TypeScript we probably want:

```
declare var  __webpack_public_path__:string; 
```

set within a `globals.d.ts` [according to an SO post on setting the webpack public path dynamically in TypeScript](https://stackoverflow.com/questions/37186218/in-webpack-how-do-i-set-the-public-path-dynamically).

Amusingly, check out https://github.com/webpack/webpack/issues/842#issuecomment-323217716 too.

packages/connected-components/src/notebook-menu/index.tsx
```diff
@@ -45,7 +45,6 @@ type Props = {
   unhideAll?: (payload: object) => void;
   cutCell?: (payload: object) => void;
   copyCell?: (payload: object) => void;
-  notebook: Immutable.Map<string, any>;
```
rgbkrk, line 48: Wow that definitely needed to be deleted. 😅  I remember doing the refactor that led to not needing this.

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
rgbkrk: Otherwise, this is defining a new local var `__webpack_public_path__` and we won't affect the global variable we have to change in order to have nteract work on JupyterHub, tmpnb, etc.

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
```diff
@@ -57,15 +58,16 @@ function main(rootEl: Element, dataEl: Node | null) {
   let config: JupyterConfigData;
 
   try {
+    if (!dataEl.textContent) {
+      throw new Error("Unable to find Jupyter config data.");
+    }
```
rgbkrk, line 63: Way cleaner than my suggestion, thanks for this.

rgbkrk (approved):

rgbkrk:
Excellent!

rgbkrk:
Thank you! We are pretty much done with this migration.

PR #3937, opened 2019-01-02
switch over button groups to styled components

rgbkrk:
This switches over the button groups and buttons that were in place for the data explorer to styled-components. I _could_ further optimize some of this, but buttons were being handled in two distinct cases here. I'd love to be able to rely on styled-components' generated class names more here rather than having generic `button-text` class names. The `selected` logic we could do as a prop to the styled-component as well, which should make it more declarative (and type-safe) to use from the outside.

### DIFF
--- a/packages/transform-dataresource/src/components/button-group.js
+++ b/packages/transform-dataresource/src/components/button-group.js
@@ -1,6 +1,12 @@
-import css from "styled-jsx/css";
+// @flow
+import styled from "styled-components";
+import { ButtonGroup, Button } from "@blueprintjs/core";
 
-export default css`
+/**
+ * This ports over some of the overridden styles within the data
+ * explorer from the migration away from styled-jsx.
+ */
+export const StyledButtonGroup = styled(ButtonGroup)`
   .button-text {
     margin: 0 10px 10px 0;
     -webkit-appearance: none;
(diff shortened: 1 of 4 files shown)


packages/transform-dataresource/src/components/button-group.js
```diff
@@ -15,22 +21,24 @@ export default css`
...
-  .button-group .button-text.selected {
+  .button-text.selected {
     background: white;
     color: #1d8bf1;
     z-index: 1;
     position: relative;
   }
 `;
+
+export const StyledButton = styled(Button);
```
captainsafia, line 44: What's the point of wrapping this button as a styled component with no style?

captainsafia (commented):
Looks good from my end. Left one question inline.

packages/transform-dataresource/src/components/button-group.js
rgbkrk: I was going to use it to target in the button-group above the selector currently written as `button-text`. Something like this:

```jsx
export const StyledButtonGroup = styled(ButtonGroup)`
  ${StyledButton} {
    margin: 0 10px 10px 0;
    -webkit-appearance: none;
```

Since this button was being used in different ways, I opted not to refactor too much here.

rgbkrk: The only other reason I left this was to have one consistent import when refactoring this.

rgbkrk:
I think I'll make these changes target the data explorer TS conversion branch instead.

PR #3939, opened 2019-01-02
Resolve remaining type errors on desktop

captainsafia:
Alright! I'm getting no more type errors when I run `npx tsc -b applications/desktop/` on local off this branch.

### DIFF
--- a/applications/desktop/src/notebook/epics/close-notebook.ts
+++ b/applications/desktop/src/notebook/epics/close-notebook.ts
@@ -22,19 +22,28 @@ import {
   DESKTOP_NOTEBOOK_CLOSING_NOT_STARTED,
   DESKTOP_NOTEBOOK_CLOSING_READY_TO_CLOSE
 } from "../state";
-import * as actionTypes from "../actionTypes";
+import { CLOSE_NOTEBOOK, CloseNotebook } from "../actionTypes";
 import * as actions from "../actions";
 
 import { Actions } from "../actions";
 import { KernelRef, KernelRecord } from "@nteract/types";
 
 export const closeNotebookEpic = (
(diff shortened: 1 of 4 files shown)


rgbkrk:
🎉

applications/desktop/src/notebook/epics/close-notebook.ts
```diff
@@ -82,7 +91,15 @@ export const closeNotebookEpic = (
...
               action$.pipe(
-                ofType(
+                ofType<
+                  | CloseNotebook
+                  | coreActions.KillKernelFailed
+                  | coreActions.KillKernelSuccessful,
+                  ReturnType<
+                    | typeof coreActions.killKernelSuccessful
+                    | typeof coreActions.killKernelFailed
+                  >
```
captainsafia, line 101: This looks weird and is mostly a workaround to get the redux-observable to correctly narrow down the types in downstream clauses when filtering with `ofType`. [This issue](https://github.com/redux-observable/redux-observable/issues/437) has more info.

packages/actions/src/actions/kernels.ts
```diff
@@ -274,7 +274,7 @@ export function deleteConnectionFileSuccessful(payload: {
 }
 
 export function shutdownReplySucceeded(payload: {
-  text: string;
+  content: { restart: boolean };
   kernelRef: KernelRef;
```
rgbkrk, line 278: Are there reducers that are going to expect and use these?

captainsafia: From what I could tell, this action is only used as an intermediary message in the shutdown epic. The action that reducers process is the `KillKernelSuccessful` action which is sent when the kernel has sent a successful shutdown reply and the process has been killed.

PR #3941, opened 2019-01-02
Switch Data Explorer off styled-jsx and on to styled-components

rgbkrk:
Convert the following to styled components:

* [x] palette picker
* [x] grid
* [x] tooltip-content
* [x] viz controls
* [x] parallel coordinates controller
* [x] index


rgbkrk:
This is making me wonder if more of the components within data explorer should become part of a new package of well documented re-usable components. Maybe literally to be called `packages/data-explorer` and it would export the "big component" as well as the individual building blocks. We'd then document it as part of components.nteract.io.

rgbkrk:
Done with the styled components changes here.

rgbkrk:
Rebasing to see if all is fixed now.

packages/transform-dataresource/src/tooltip-content.tsx
```diff
@@ -1,74 +1,70 @@
...
 
-  const translateX = x < 100 ? "0px" : "calc(-50% + 7px)";
-  const translateY = y < 100 ? "10px" : "calc(-100% - 10px)";
+// A little "mixin" for picking the :before on a tooltip
+const beforeContent = (props: Props) =>
+  props.x < 100
+    ? null
+    : props.y < 100
+    ? `
+border-left: inherit;
```
captainsafia: Style nit: this would be much easier to understand if the contents of these string literals were assigned to variables and referenced here.

captainsafia (approved):
Looks good from my end. Left one small comment inline that you can do either here or as a fast follow.

packages/transform-dataresource/src/tooltip-content.tsx
rgbkrk: I was a little afraid to change the structure too much (I actually heavily modified this one then backed out of it). For this one though I'll at least break out of this nested ternary.

stormpython (approved):
This looks good. I'd prefer a better way of organizing the css styles, but that would take much more time than we currently have. In any case, follow up code refactoring in Data Explorer can also take care of this.

rgbkrk:
Alright, I'm merging this into our big Data Explorer TS PR.

PR #3943, opened 2019-01-02
Fix more Jext type errors

captainsafia:
Two more errors left.

Will submit a PR for those once I'm back home but I think these fixes are a good stopping point.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/components/last-saved.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/components/last-saved.tsx
@@ -81,13 +81,17 @@ class LastSaved extends React.PureComponent<LastSavedProps> {
   }
 }
 
+interface OwnProps {
+  contentRef: string;
+}
+
 /**
  * Create our state mapper using makeMapStateToProps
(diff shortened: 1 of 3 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/components/last-saved.tsx
```diff
@@ -102,9 +106,6 @@ const makeMapStateToProps = (
   return mapStateToProps;
 };
 
-const ConnectedLastSaved = connect(
-  makeMapStateToProps,
-  LastSaved
-);
-
-export default ConnectedLastSaved;
+export default connect<LastSavedProps, typeof LastSaved, OwnProps, AppState>(
```
rgbkrk, line 109: This is great to see what the definition needed to be here.

captainsafia: Yeah. This was the type that was inferred by TypeScript but for some reason I had to explicitly define it to get the places where it was used to correctly recognize what OwnProps was supposed to be.

PR #3947, opened 2019-01-02
dx: fix tsconfig references to other projects

rgbkrk:
Noticed some of the dependencies were missing references in the tsconfig.

### DIFF
--- a/packages/transform-dataresource/src/charts/xyplot.tsx
+++ b/packages/transform-dataresource/src/charts/xyplot.tsx
@@ -8,7 +8,7 @@ import TooltipContent from "../tooltip-content";
 
 import { sortByOrdinalRange } from "./shared";
 import * as Dx from "../types";
-import { JSONObject } from "@nteract/commutable/src";
+import { JSONObject } from "@nteract/commutable";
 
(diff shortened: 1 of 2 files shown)


rgbkrk:
The fact that we're using `@nteract/commutable` for the `JSONObject` type means to me we need a `@nteract/light-types` package. 🙃

packages/transform-dataresource/src/charts/xyplot.tsx
```diff
@@ -8,7 +8,7 @@ import TooltipContent from "../tooltip-content";
 
 import { sortByOrdinalRange } from "./shared";
 import * as Dx from "../types";
-import { JSONObject } from "@nteract/commutable/src";
+import { JSONObject } from "@nteract/commutable";
```
rgbkrk, line 11: Note: this isn't the only file in this package that uses `JSONObject` -- I just made it consistently use the top level package entry for this across all of them (only this offender was different).

rgbkrk:
Merging as this makes it more consistent and the main branch for the data explorer TS branch still has the same issue.

PR #3948, opened 2019-01-02
include dx in the top level packages setup

rgbkrk:
like it says on the tin


rgbkrk:
Little bit better, same error sticking around.

PR #3949, opened 2019-01-03
Fix action types in connected components

captainsafia:


packages/actions/src/actionTypes/kernels.ts
```diff
@@ -87,15 +87,15 @@ export const INTERRUPT_KERNEL = "INTERRUPT_KERNEL";
 export type InterruptKernel = {
   type: "INTERRUPT_KERNEL";
   payload: {
-    kernelRef: KernelRef;
+    kernelRef?: KernelRef | null;
```
rgbkrk, line 90: These actions definitely need a `kernelRef` though otherwise these actions shouldn't even be created.

Do we still have stuff that assumes there's only one kernel? Even on desktop we have to assume N kernels because of restarted kernels, switching kernels, etc.

packages/connected-components/src/notebook-menu/index.tsx
```diff
@@ -36,19 +36,38 @@ type Props = {
...
-  executeAllCellsBelow?: (payload: object) => void;
-  clearAllOutputs?: (payload: object) => void;
-  unhideAll?: (payload: object) => void;
-  cutCell?: (payload: object) => void;
-  copyCell?: (payload: object) => void;
-  pasteCell?: (payload: object) => void;
-  createCellBelow?: (payload: object) => void;
-  changeCellType?: (payload: object) => void;
-  setTheme?: (theme?: string) => void;
+  saveNotebook?: (payload: { contentRef: string }) => void;
```
rgbkrk, line 39: Something cool @stormpython and I noticed -- if you want to use _just_ the payload portion of an action, you can index into an `interface` or `type` like this `saveNotebook?: (payload: actions.Save["payload"]) => void;`

packages/connected-components/src/notebook-menu/index.tsx
```diff
@@ -57,11 +76,29 @@ type Props = {
...
     }
   ) => void;
-  restartKernel?: (payload: object) => void;
-  restartKernelAndClearOutputs?: (payload: object) => void;
-  restartKernelAndRunAllOutputs?: (payload: object) => void;
-  killKernel?: (payload: object) => void;
-  interruptKernel?: (payload: object) => void;
+  restartKernel?: (
+    payload: {
+      outputHandling: actions.RestartKernelOutputHandling;
```
rgbkrk, line 81: This seems a bit off, but hey if this all passes for now, I'm going to assume we can re-look at this after we're done with data explorer.

rgbkrk (approved):

packages/connected-components/src/notebook-menu/index.tsx
captainsafia: Yeah. I think all these actions/selector types need a good look through in a separate PR. I did this mostly to unblock data explorer.

PR #3961, opened 2019-01-03
[WIP] switch styleguide to an exclusively typescripted setup

rgbkrk:
Experimenting with trimming down our styleguide config. It's a bit of a mess here. If anyone wants to try this they have to, assuming they've also ran `yarn` at the root level, run `npx styleguidist server` inside the styleguide directory.


rgbkrk:
This tells me the styleguidist build should probably be part of our CI.

rgbkrk:
Closing in favor of a cleaner approach.

PR #3964, opened 2019-01-04
Remove @nteract/play from monorepo

captainsafia:
I'm working on moving it to https://github.com/nteract/play.


rgbkrk:
Thank you!

PR #3965, opened 2019-01-04
Convert the rest of Desktop to TypeScript

rgbkrk:
Completes the conversion to TypeScript for Desktop!

🎉 🌮 🎩 🎉

### DIFF
--- a/applications/desktop/src/main/menu.ts
+++ b/applications/desktop/src/main/menu.ts
@@ -20,7 +20,7 @@ function send(
 }
 
 function createSender(eventName: string, obj?: object | string | number) {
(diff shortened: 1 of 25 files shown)


applications/desktop/src/main/menu.ts
```diff
@@ -165,7 +169,7 @@ export const named = {
   ]
 };
 
-export function loadFullMenu(store: * = global.store) {
+export function loadFullMenu(store = global.store) {
```
rgbkrk, line 181: Wow I didn't realize these functions were cheating to get access to the global store rather than get passed it.

PR #3977, opened 2019-01-05
Switch Desktop to 100% TypeScript Loader

rgbkrk:
🎉

### DIFF
--- a/applications/desktop/tsconfig.json
+++ b/applications/desktop/tsconfig.json
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.base.json",
-  "compilerOptions": {},
+  "compilerOptions": {
+    "noEmit": true
+  },
   "include": ["src"],
   "references": [
     { "path": "../../packages/core" },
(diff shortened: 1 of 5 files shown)


applications/desktop/tsconfig.json
```diff
@@ -1,8 +1,6 @@
 {
   "extends": "../../tsconfig.base.json",
-  "compilerOptions": {
-    "noEmit": true
-  },
+  "compilerOptions": {},
```
rgbkrk: Since I had to disable `noEmit`, I wonder if I need to change how reporting works.

rgbkrk: Whoops. I sure do, otherwise the type checking generates `.d.ts` files.

rgbkrk: Ok, I'm wondering if I should just emit these to an untracked-by-git-location.

rgbkrk:
Alright, this is ready for a looksee.

Pretty happy with `ts-loader` thus far.

captainsafia (approved):

PR #3978, opened 2019-01-05
clean up remaining references to flow

rgbkrk:
Cleaned up some (now) old `$FlowFixMe`, `@flow`, and other references to flow bits.

### DIFF
--- a/packages/core/plan.md
+++ b/packages/core/plan.md
@@ -2,24 +2,24 @@ This document attempts to specify the goal for an nteract core package that can
 provide generic/core state management to all notebook-y applications. Right now
 this is focused on these apps:
 
-* Desktop
-* nteract on jupyter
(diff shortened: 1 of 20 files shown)


packages/core/plan.md
```diff
@@ -2,24 +2,24 @@ This document attempts to specify the goal for an nteract core package that can
...
 this is focused on these apps:
 
-* Desktop
-* nteract on jupyter
-* `@nteract/play`
-* notebook on next
+- Desktop
+- nteract on jupyter
+- `@nteract/play`
+- notebook on next
```
rgbkrk, line 8: My prettier markdown formatting snuck in here.

rgbkrk: Reminds me we should probably run `yarn prettify` in the root and make a PR.

PR #3980, opened 2019-01-05
remove all inner build & prepare scripts for pkgs

rgbkrk:
All the `prepare` and `prepublishOnly` scripts need to be taken out prior to publishing (I noticed this while in the middle of publishing), otherwise it builds every package individually a second time.


rgbkrk:
I... probably should have left the clean scripts. Maybe those can run at the top though.

rgbkrk:
Yep, `build:clean` does the work.

PR #3985, opened 2019-01-05
Use ts-loader in styleguidist config

captainsafia:
Still need to fix the `Host` and `Kernel` component rendering.


rgbkrk:
Looking better and better. I wonder if we need a custom tsconfig for the styleguide.

rgbkrk:
Hey it's passing!

rgbkrk:
Did you mean to leave this in WIP?

captainsafia:
I was gonna try to figure out why the Provider and Context types in the host-cache package weren’t being transpired but I guess that can be addressed later.

PR #3990, opened 2019-01-06
Cleanup old babel and styled-jsx remnants

rgbkrk:
This wipes out the final bits of `babel` usage inside of the nteract monorepo. I also set the `ts-jest` config to only transpile, so it won't check types merely transpiling. Similar to before, I've left a note about what we need to do to progress in the `package.json`.


rgbkrk:
Weird, the tests worked for me.

image

rgbkrk:
There were some inconsistencies between our mocks and the real jmp definition. I'll fix them up later tonight.

packages/enchannel-zmq-backend/__tests__/index_spec.ts
```diff
@@ -196,14 +203,14 @@ describe("createMainChannelFromSockets", () => {
     });
   });
 
-  test("propagates header information through", () => {
+  test("propagates header information through", async () => {
```
captainsafia: Do we need to invoke `done` here?

packages/notebook-app-component/__tests__/__snapshots__/toolbar.spec.tsx.snap
```diff
@@ -130,8 +130,8 @@ exports[`Toolbar View should be able to render a toolbar 1`] = `
                 <span
                   className="octicon"
                 >
-                  <TriangleRightOcticon>
-                    <SVGWrapper
+                  <Component>
+                    <Component
```
rgbkrk: I find it super weird these are coming up as `Component`. They're not `styled-components`, these are just functional components.

packages/enchannel-zmq-backend/__tests__/index_spec.ts
rgbkrk: Oh whoop, either that or I should take out `async`.

packages/notebook-app-component/__tests__/__snapshots__/toolbar.spec.tsx.snap
rgbkrk: I've manually set the display name here and it fixes it up within the jest snapshots. That's kind of gross to have to do for all these lightweight functional components.

rgbkrk: As pointed out to me by @captainsafia others have faced the same: https://stackoverflow.com/questions/51045532/ts-jest-component-name-disappears-in-snapshots

packages/enchannel-zmq-backend/__tests__/index_spec.ts
rgbkrk: Added the done in and called it at the bottom.

packages/notebook-app-component/__tests__/__snapshots__/toolbar.spec.tsx.snap
rgbkrk: Opting to just not fix these up and instead hope for resolution in the upstream tools (unless we take it upon ourselves to fix this up).

PR #3993, opened 2019-01-07
fix jupyter extension webpack build

rgbkrk:
This was a fun one.

We are importing `notebook` from the notebook reducer, by name.

Within one segment of a big switch we were creating a `const notebook`:

```jsx
case "notebook": {
          const notebook = fromJS(...)
}
```

Note the `{}` block scoping on that case statement. In a separate action entirely we use the `notebook` reducer. That turned out to be an undefined var instead (referencing this yet to be set `const notebook`). Switching over to typescript uncovered this, apparently TypeScript doesn't attempt to keep these block scoped with IIFEs or otherwise. Block scoping for switch cases is part of ES6 according to [MDN's guide to `switch`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch).

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
+++ b/applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
@@ -18,6 +18,7 @@ module.exports = {
   devServer: isProd
     ? {}
     : {
+        disableHostCheck: true,
         hot: true,
         headers: { "Access-Control-Allow-Origin": "*" }
       },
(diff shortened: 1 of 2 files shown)


applications/jupyter-extension/nteract_on_jupyter/webpack.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
   devServer: isProd
     ? {}
     : {
+        disableHostCheck: true,
```
rgbkrk, line 21: Note: this removes host checking (for now) until we upgrade webpack-dev-server. Otherwise we get spammed with lots of logs. I'll follow up with a fix for this by upgrading webpack-dev-server and related bits.

PR #4001, opened 2019-01-09
break up hint and codemirror css

### LINKED ISSUE #3872: Editor Popover styles broken
Just noticed on master that the editor styles are broken for tab completion. 😬 

I'll fix it, that came from my migration of editor to styled-components. Looks like the vendored codemirror css didn't get loaded properly.

screen shot 2018-12-18 at 8 08 18 pm

rgbkrk:
This fixes #3872 while also cleaning up our rendering of both hints and codemirror areas themselves. I refactored a bit of the completion handling and included more comments to help the next developer in this corner of the code.

screen shot 2019-01-08 at 9 28 22 pm

screen shot 2019-01-08 at 9 29 54 pm


rgbkrk:
My last commit(s) may break the build here, time to find out.

packages/editor/src/jupyter/complete.tsx
```diff
@@ -18,54 +26,31 @@ export function formChangeObject(cm: CMI, change: EditorChange) {
...
-      span.className += "completion-type completion-type-" + completion.type;
-      span.setAttribute("title", completion.type);
-      elt.appendChild(span);
-      elt.appendChild(text);
-    }
-  }))
-});
+type CompletionMatch =
+  | string
+  | {
```
captainsafia: Nit: But I think this interface type should be defined under its own variable.

captainsafia (approved):
Thanks for tackling this! I left one small style nit that you can fix here or ignore for now.

stormpython (approved):
LGTM!

packages/editor/src/jupyter/complete.tsx
rgbkrk: Addressed.

PR #4002, opened 2019-01-09
Raw cell

kelleyblackmore:
I'll be looking for some feedback on this one

### DIFF
--- a/applications/desktop/src/main/menu.ts
+++ b/applications/desktop/src/main/menu.ts
@@ -402,6 +402,11 @@ export function loadFullMenu(store = global.store) {
         click: createSender("menu:new-text-cell-below")
       },
       {
+        label: "Insert Raw Cell Below",
+        enabled: BrowserWindow.getAllWindows().length > 0,
+        click: createSender("menu:new-raw-cell-below")
+      },
+      {
         type: "separator"
       },
       {
--- a/applications/desktop/__tests__/renderer/menu.spec.ts
+++ b/applications/desktop/__tests__/renderer/menu.spec.ts
@@ -63,6 +63,25 @@ describe("dispatchCreateTextCellBelow", () => {
     );
   });
 });
+describe("dispatchCreateRawCellBelow", () => {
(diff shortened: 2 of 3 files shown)


rgbkrk:
This is looking good to me so far. You might want to change the title on this so folks know it's for the raw cell.

BenRussert:
Are you planning to add anything else to the ui for this, like a create cell below button in `notebook-app-component`? Otherwise, I'd add something to `USER_GUIDE.md` either way, but then this is ready to go!

kelleyblackmore:
I was going to put some other changes with this pr but I won't have time for it this week

applications/desktop/src/main/menu.ts
```diff
@@ -401,6 +401,11 @@ export function loadFullMenu(store = global.store) {
         enabled: BrowserWindow.getAllWindows().length > 0,
         click: createSender("menu:new-text-cell-below")
       },
+      {
+        label: "Insert Raw Cell Below",
+        enabled: BrowserWindow.getAllWindows().length > 0,
```
rgbkrk, line 406: Something I've pointed out in past PRs. There's a simple cleanup here that would apply to the whole menu. We can store `enabled in a variable before this segment is created:

```ts
const enabled = BrowserWindow.getAllWindows().length > 0;
```

Then use it in each of these

```ts
      {
        enabled,
        click: createSender("menu:new-text-cell-below")
      },
      {
        label: "Insert Raw Cell Below",
        enabled
     }
```

That can be done in another PR though.

captainsafia:
I'm gonna go merge this since the code feedback can be addressed in another PR.

> Are you planning to add anything else to the ui for this, like a create cell below button in notebook-app-component? Otherwise, I'd add something to USER_GUIDE.md either way, but then this is ready to go!

It would make sense to have this in the user guides on the nteract/docs repo.

rgbkrk:
In my (very) personal opinion I wouldn't make an icon for a raw cell in the cell creator as its one of the least used cells and has very different behavior dependent on what it's being used for. This implementation at least lets you create them if you need them.

Issue #2034, opened 2017-11-01
GeoJSON Transform needs styles to get brought in by webpack


rgbkrk:
Similar to the rest of our CSS setup, our GeoJSON transform needs to bring in CSS properly. Additionally, it looks like we need Leaflet to be able to load some image data:

https://github.com/nteract/nteract/blob/76493806330373bce633c7c904b9e6fdcdbba59e/packages/transform-geojson/src/index.js#L16

I am completely unaware of how we should handle that...

rgbkrk:
Totally forgot to note in this issue that our GeoJSON transform is currently _broken_ as a result.

josephofiowa:
leaftlet geojson renderer fails; Kyle and I are taking a look

captainsafia:
Working on this now. Now that we're on styled-components, @rgbkrk mentioned that we can use their global styles to get this set up so I'll give this a go now.

captainsafia:
I think we can go ahead and close this since GeoJSON rendering is working now.

PR #4003, opened 2019-01-09
Package LeafletCSS with styled-components

captainsafia:
Closes part of https://github.com/nteract/nteract/issues/2034. Still gotta figure out the images thing.

I ended up just copying the CSS as a string literal instead of doing a module import because that would require running our packages build with Webpack and using a CSS module loader.

### DIFF
--- a/packages/transform-geojson/src/leafletCss.ts
+++ b/packages/transform-geojson/src/leafletCss.ts
@@ -1,6 +1,6 @@
 import { createGlobalStyle } from "styled-components";
 
-const leafletCSS = `/* required styles */
+export default createGlobalStyle`/* required styles */
 
 .leaflet-pane,
 .leaflet-tile,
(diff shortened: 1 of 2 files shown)


packages/transform-geojson/src/leafletCss.ts
```diff
@@ -0,0 +1,639 @@
+import { createGlobalStyle } from "styled-components";
+
+const leafletCSS = `/* required styles */
```
rgbkrk: I think it's ok to just make this template string go ahead and use `createGlobalStyle` up here.

rgbkrk:
> I ended up just copying the CSS as a string literal instead of doing a module import because that would require running our packages build with Webpack and using a CSS module loader.

That's the pattern we want to stick with for the "third party" CSS, thanks for mirroring the pattern we use for CodeMirror, rc-menu, others.

rgbkrk (approved):

rgbkrk:
Great, this should fix things for each of the apps using it.

Issue #4004, opened 2019-01-09
Switch webpack back to babel-loader


rgbkrk:
As noted in https://github.com/gaearon/react-hot-loader/blob/master/README.md#typescript, in order to use `react-hot-loader` we have to use the `babel-loader`.

> As of version 4, React Hot Loader requires you to pass your code through Babel to transform it so that it can be hot-reloaded.

During the conversion process to TypeScript I thought that in order for us to have type checking, I thought we'd have to switch over to `ts-loader` (for the sake of handling project references). It seems that we can set things up in a dual mode: type checking gets done with a rigorous tsconfig that uses project references while our webpack setup is allowed to use `babel-loader` + `react-hot-loader`.

lgeiger:
Would switching back to babel would reintroduce #3594 due to https://github.com/babel/babel/issues/8361?

There's a discussion on the next.js repo about this problem related to HMR: https://github.com/zeit/next-plugins/issues/306

rgbkrk:
We took the next.js apps out of the repo, so we won't run into that anymore. They'll use the transpiled packages.

captainsafia:
We'll still need to do this.

The work to introduce babel-loader has been introduced to our test workflows during the addition of ipywidgets support. We'll need to follow this through to the source code.

captainsafia:
Closing this since we'll realistically not be able to handle this any time soon.

PR #4005, opened 2019-01-09
take out forced hot(module) setups

rgbkrk:
For the time being (until #4004 is addressed), this makes things more stable -- the page fully reloads instead of trying to hot swap out components.


stormpython (approved):
LGTM!

PR #4009, opened 2019-01-10
perf: switch connected components to makeMapStateToProps

rgbkrk:
Flow was preventing us from being able to switch to `makeMapStateToProps`. Now that we're on typescript it's an easy switch over. This will include a _lot_ of fixes for #3434.

* [x] jext's `contents/file`
* [x] jext's `contents/index`

General pattern here is:

* Move `mapStateToProps` into a function called `makeMapStateToProps`:

```js
const makeMapStateToProps = (initialState: AppState, initialProps: InitialProps) => {
  const mapStateToProps = (state: AppState) => {
    // can use initialProps in a closure here
  }
  return mapStateToProps;
}

connect(makeMapStateToProps, mapDispatchToProps)(Component);
```

Note that `mapDispatchToProps` also needs to move away from having `ownProps` as well.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/file/index.tsx
@@ -82,40 +82,50 @@ export class File extends React.PureComponent<FileProps> {
   }
 }
 
-const mapStateToProps = (
-  state: AppState,
-  ownProps: {
-    contentRef: ContentRef;
-    appBase: string;
-  }
-) => {
-  const content = selectors.content(state, ownProps);
-
-  if (!content || content.type === "directory") {
-    throw new Error(
-      "The file component should only be used with files and notebooks"
-    );
-  }
-
-  const comms = selectors.communication(state, ownProps);
-  if (!comms) {
-    throw new Error("CommunicationByRef information not found");
-  }
-
-  return {
-    appBase: ownProps.appBase,
-    baseDir: dirname(content.filepath),
-    contentRef: ownProps.contentRef,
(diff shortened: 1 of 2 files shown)


rgbkrk:
Oh gosh, I see how much work is involved for getting `ownProps` moved to `initialProps` within a `makeMapStateToProps` for both editor and notebook, it's a doozy. I'll want to do those as a separate PR so that other folks can follow the lead on this one. Probably going to go one PR at a time for the beefier components.

captainsafia (approved):
Coolio. Can you link to an article that explains why we're doing this for everyone's edification?

rgbkrk:
What's a little sad about this is that the documentation for `makeMapStateToProps` is buried.

Within [Sharing Selectors Across Components](https://redux.js.org/recipes/computing-derived-data#selectors-todoselectorsjs-1), it's noted that:

> If the `mapStateToProps` argument supplied to `connect` returns a function instead of an object, it will be used to create an individual `mapStateToProps` function for each instance of the container.

The first instance I heard of this came from [Dan Abramov on twitter](https://twitter.com/dan_abramov/status/719971882280361985?lang=en):

> If you only use ownProps in mapStateToProps() to read an ID, use this pattern instead and get perf a boost:
> https://github.com/mweststrate/redux-todomvc/pull/1/commits/e5d25f047ff62c1eddd2b3ad2a62a7b4638ee52e

The original PR: https://github.com/reduxjs/react-redux/pull/348

For us, basically every use of `contentRef` or `cellID` (or anything else by ID) should not be coming in as part of `ownProps`. It's going to stay the same for the same component and shouldn't be forced to deoptimize (what's currently happening).

PR #4011, opened 2019-01-10
Finalize typechecking on CI

rgbkrk:
* Addresses a couple new type errors in the jupyter extension
* Bypasses the error that was preventing us from reporting on type errors in jupyter extension
* Enables checking by default on CI


captainsafia (approved):
LGTM!

rgbkrk:
Wahoo, this makes me feel much better about changes going in now (even if not fully typesafe).

PR #4014, opened 2019-01-10
build all pkgs and apps with `tsc -b` at root

rgbkrk:
This moves our root level `tsc` build to handle all apps and packages. I've also switched up the Circle CI order so that we get the benefit of building all at once (since we're fully on project references).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,12 +31,9 @@ jobs:
 
       # run tests!
       - run: yarn run test:lint
-      - run: yarn run build:packages:ci
+      # build all packages and applications
+      - run: yarn build:all:ci
       - run: yarn run test --maxWorkers 1
-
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -31,13 +31,11 @@ jobs:
 
       # run tests!
       - run: yarn run test:lint
+      # build all packages and applications
+      - run: yarn build:all:ci
       - run: yarn run build:packages:ci
```
lgeiger: I think the `build:packages:ci` command can now be removed since, since `build:all:ci` handles everything.

lgeiger (commented):
Nice 🎉

.circleci/config.yml
rgbkrk: Ah yeah definitely.

PR #4022, opened 2019-01-12
 Fix error when a kernel info dir does not exist

### LINKED ISSUE #4021: Non-existing 'kernels' directory behaviour inconsistent with Jupyter
My system has a `kernels` directory in `/usr/share/jupyter` (used by notebook), but not in `/usr/local/share/jupyter` (used by lab). This causes a bunch of tests to fail with

  `Error: ENOENT: no such file or directory, scandir '/usr/local/share/jupyter/kernels'`

This seems to be a bug rather than a problem with my setup as Jupyter explicitly checks for non-existing dirs in [kernelspec.py](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/kernelspec.py#L86):

```python
def _list_kernels_in(dir):
    """Return a mapping of kernel names to resource directories from dir.
    If dir is None or does not exist, returns an empty dict.
    """
    if dir is None or not os.path.isdir(dir):
        return {}
    ...
```

CrystallineCat:
Fixes #4021.

### DIFF
--- a/packages/fs-kernels/src/kernelspecs.ts
+++ b/packages/fs-kernels/src/kernelspecs.ts
@@ -63,17 +63,25 @@ function getKernelResources(kernelInfo: KernelInfo): Promise<KernelResource> {
 
 /**
  * Gets a list of kernelInfo objects for a given directory of kernels
- * @param  {string}   directory path to a directory full of kernels
- * @return {Promise<Object[]>}  Promise for an array of kernelInfo objects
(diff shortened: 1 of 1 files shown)


captainsafia:
Oh gosh! Thank you for catching this. I implemented what I thought was a fix for this issue but I wrote the logic in `jupyter-paths.ts` and didn't catch the fact it was doing a read here too.

PR #4023, opened 2019-01-12
Use @nteract/outputs in @nteract/notebook-app-component

### LINKED ISSUE #3342: <Output> Picker

captainsafia:
Closes https://github.com/nteract/nteract/issues/3342.


packages/jupyter-widgets/src/index.tsx
```diff
@@ -19,7 +19,7 @@ type Props = {
  * Even though it may appear to be pure, since it doesn't have react state, this
  * component's iframe maintains it's own state in communication with the kernel.
  */
-export class WidgetDisplay extends React.Component<Props> {
+export class WidgetDisplay extends React.Component<Partial<Props>> {
```
captainsafia, line 22: Here and elsewhere, I had to require that transform components take a Partial as a prop since the picker passes the props to the components using `React.cloneElement` and the props aren't set on these components when they are children of `Output`.

packages/outputs/src/components/kernel-output-error.tsx
```diff
@@ -42,17 +42,17 @@ const PlainKernelOutputError = (props: Props) => {
...
-  traceback: []
-};
-
 export const KernelOutputError = styled(PlainKernelOutputError)`
   & code {
     white-space: pre-wrap;
   }
 `;
 
+KernelOutputError.defaultProps = {
```
captainsafia, line 49: I had to set the props on the styled component because styled-components clobbers the defaultProps initialization for some reason.

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -252,34 +268,69 @@ class AnyCell extends React.PureComponent<AnyCellProps> {
...
+                  <DisplayData>
+                    <Media.HTML />
+                    <Media.Image />
+                    <Media.JavaScript />
+                    <Media.Json />
+                    <Media.LaTeX />
+                    <Media.Markdown />
+                    <Media.Plain />
+                    <Media.SVG />
+                    <DataResourceTransform />
```
rgbkrk:
```suggestion
                    <DataResourceTransform onMetadataChange={this.props.updateOutputMetadata} />
```

packages/jupyter-widgets/src/index.tsx
rgbkrk: Oh... Interesting. I wonder if there's a good paradigm for having a version that can take a partial and another that must have the props set. We're probably safer if this is `Partial` and checked below though. One thing though... I feel like we have to require `mediaType` even though we set it as a `defaultProp` so that the `Output` component can use it.

rgbkrk:
First off, fantastic! I'm so happy to switch to this more compound API.

I'm going to re-iterate some stuff that was in slack messages, for the benefit of everyone.

We don't want to include all the "advanced" transforms in our default app component. The component should be flexible enough to set exactly which media components to use for each media type (and in which order). The main reasons are:

* Ability to not include media types or components deemed unsafe
* Enable media types & components custom to a deployment (customized version of Plotly, some special media type)
* Ability to code split and load transforms dynamically (improve initial render / user experience)
  * Jupyter extension does this currently here: https://github.com/nteract/nteract/blob/master/applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx#L76-L82

Safia said it best the problem we need to solve:

> Now that media transforms are passed as child components instead of props we need a way to configure which transforms will be rendered on the page.

Our current way of passing the transforms and displayOrder as an `ownProp` is exacerbating our perf problem (#3434), so we need to move them to somewhere else. My proposal: create redux actions for registering media types and set them on the state tree likely under `state.core.entities.media`.

The _wonderful_ thing about this is that it would make it easier for all the different segments that use these same media transform components to pluck them off in their `mapStateToProps` connected components easily. This includes the editor, display data, execute result, and pagers.

What I'm stepping past though is how the custom media types would get access to things to `models`, `onMetadataChange` (maybe it's fine if they all get it?), etc.

rgbkrk:
We could do a separate PR that does the switch over to storing the transforms in the redux store first.

* Expect transforms and displayOrder as they currently do to be on `state.core.entities.media.(transforms|displayOrder)`
* Desktop can literally set all the ones it's using as part of the `createStore` instead of dynamically load them
* Jupyter Extension can `dispatch` to add/set transforms

Alternatively, that could be part of this PR and just go straight after our final implementation. I'll leave that up to you @captainsafia.

packages/outputs/src/components/execute-result.tsx
```diff
@@ -45,7 +45,7 @@ export const ExecuteResult = (props: Props) => {
 };
 
 ExecuteResult.defaultProps = {
-  outputType: "pyout",
+  output_type: "execute_result",
```
rgbkrk, line 48: Good catch on this incorrect `output_type`. That's the old messaging spec there. 😬

packages/outputs/src/components/kernel-output-error.tsx
```diff
@@ -42,17 +42,17 @@ const PlainKernelOutputError = (props: Props) => {
...
 };
 
-PlainKernelOutputError.defaultProps = {
-  outputType: "error",
-  ename: "",
-  evalue: "",
-  traceback: []
-};
-
 export const KernelOutputError = styled(PlainKernelOutputError)`
```
rgbkrk, line 45: I just noticed that we're missing the passing of a `className` into the `<Ansi` child of `PlainKernelOutputError`. `styled-components` won't actually style this component as a result.

packages/outputs/src/components/media/image.tsx
```diff
@@ -4,15 +4,15 @@ import styled from "styled-components";
 interface Props {
   data: string;
   mediaType: "image/png" | "image/jpeg" | "image/gif";
-  metadata: any;
+  metadata?: { height: number; width: number };
```
rgbkrk: Each of these props should also be optional `height?: number` in case `metadata` is defined.

packages/outputs/src/components/rich-media.tsx
```diff
@@ -98,6 +103,7 @@ export class RichMedia extends React.Component<RichMediaProps, State> {
     let chosenOne: React.ReactChild | null = null;
 
     const data = this.props.data;
+    const mimetype = richestMimetype(data, displayOrder, transforms);
```
rgbkrk: Could probably name this var `mediaType` to match down below.

rgbkrk: Oh wait. We don't need this function any more.

The order is picked below by the order of when the children appear.

rgbkrk: This is documented well in the markdown in https://components.nteract.io/#richmedia yet not in the component.

> The `<RichMedia />` component accepts the whole media bundle from a kernel via the `data` prop and all the elements for rendering media types as children. The order of the children states their richness from highest to lowest.

packages/outputs/src/components/rich-media.tsx
```diff
@@ -109,7 +115,7 @@ export class RichMedia extends React.Component<RichMediaProps, State> {
       if (
         childElement.props &&
         childElement.props.mediaType &&
-        childElement.props.mediaType in data
```
rgbkrk: This original check basically made sure that the `mediaType` of a child matched that of one in `data`, checking in order.

```jsx
<RichMedia>
  <Plain />
  <HTML />
</RichMedia>
```

means that Plain is preferred over HTML whereas 

```jsx
<RichMedia>
  <HTML />
  <Plain />
</RichMedia>
```

means that HTML is preferred.

packages/records/__tests__/appendOutput.spec.ts
```diff
@@ -3,46 +3,46 @@ import appendOutput, { mutate } from "../src/outputs/append-output";
...
-        outputType: "error",
+        output_type: "error",
         ename: "well",
         evalue: "actually",
         traceback: []
       }
     ];
     const newOutputs = appendOutput(outputs, {
-      outputType: "displayData",
+      output_type: "displayData",
```
rgbkrk: Glad I saw this. The `output_type` name needs to match the jupyter spec so something (before this PR) was erroneously using the `camelCase` `displayData` instead of `display_data`.

packages/records/__tests__/index.spec.ts
```diff
@@ -61,7 +61,7 @@ describe("display_data output", () => {
       })
     ).toEqual(
       displayData({
-        outputType: "display_data",
+        output_type: "display_data",
```
rgbkrk, line 64: Yet this one has the write casing for `display_data`. Must just be the test above.

packages/transform-dataresource/src/index.tsx
```diff
@@ -486,7 +490,7 @@ class DataResourceTransform extends React.Component<Props, State> {
 
     return (
       <div>
-        <MetadataWarning metadata={this.props.metadata} />
+        <MetadataWarning metadata={this.props.metadata!} />
```
rgbkrk, line 493: Good way of handling this for now.

styleguide.config.js
```diff
@@ -74,16 +74,17 @@ module.exports = {
...
       fs: "empty",
       child_process: "empty",
-      net: "empty"
-    },
-    resolve: {
-      extensions: [".ts", ".tsx"]
-    },
-    module: {
+      net: "empty",
+      canvas: "empty"
```
rgbkrk, line 82: A+

packages/records/__tests__/appendOutput.spec.ts
captainsafia: Ooooh! Good catch. Thanks!

packages/outputs/src/components/rich-media.tsx
captainsafia: Oh! My bad. I totally missed this documentation (which is funny because didn't I write it?). I thought we needed to keep this `richestMime` logic back. This actually simplifies a whole lot.

packages/outputs/src/components/media/image.tsx
captainsafia: Great call!

packages/outputs/src/components/rich-media.tsx
rgbkrk: @alexandercbooth and I wrote this one like 4-5 months ago. Since we had not done the integration work (like you've done here), we never got to kill off the old `richestMime`.

captainsafia: OK! That makes me feel a little bit better about missing this. Code is actually much cleaner without it. Feels good to not rely on all the displayOrder stuff.

captainsafia:
Thanks for posting that summary, @rgbkrk!

Based on the problem definition that I mentioned, I thought up a possible solution we could use to address this. This solution uses the [react-loadable](https://www.npmjs.com/package/react-loadable) package which I quite liked for (1) its API, (2) documentation, and (3) decent adoption rate. With React loadable, we can define as Loadable component for each transform as such. Here's some example code that shows how we can use this component and also how we might be able to use the `render` API within react-loadable to pass props like onMetadataChange to components.

```
import Loadable from "react-loadable";

export default class Notebook extends React.Component {
  DataExplorer = Loadable({
    loader: () => import("@nteract/transform-dataresource"),
    loading: NullDataExplorerTransform,
    render(loaded, props) {
      this.props.dispatchAddTransforms(
        <loaded.export
          onMetadataChange={this.props.dispatchOnMetadataChange})
          )}
        />
      );
    }
  });

  render() {
    return <App contentRef={this.props.contentRef} transforms={this.props.transformsFromState}/>;
  }
}

const makeMapStateToProps = () => {
    return (state) => {
        return { transformsFromState: state.transforms }
    }
}

const makeMapDispatchToProps = () => {
  return dispatch => {
    return {
      dispatchAddTransforms: dispatch(actions.addTransforms),
      dispatchOnMetadataChange: dispatch(actions.onMetadataChange)
    };
  };
};

const ConnectedNotebook = connect(
  makeMapStateToProps,
  makeMapDispatchToProps
)(Notebook);

```

If this approach sounds good, to keep the velocity moving, I'm in favor of merging this and implementing this logic in a separate PR.

rgbkrk:
I'm in slight agreement. Let me look over the current state.

applications/desktop/src/notebook/index.tsx
```diff
@@ -122,8 +122,6 @@ export default class App extends React.PureComponent {
           <NotebookApp
             // The desktop app always keeps the same contentRef in a browser window
             contentRef={contentRef}
-            transforms={transforms}
-            displayOrder={displayOrder}
```
rgbkrk, line 126: 👍

packages/notebook-app-component/package.json
```diff
@@ -17,10 +17,18 @@
...
     "@nteract/octicons": "^1.0.4-alpha.0",
+    "@nteract/outputs": "^1.0.4-alpha.0",
     "@nteract/presentational-components": "^2.0.4-alpha.0",
     "@nteract/selectors": "^1.0.4-alpha.0",
+    "@nteract/transform-dataresource": "^5.0.4-alpha.0",
+    "@nteract/transform-geojson": "^4.0.4-alpha.0",
+    "@nteract/transform-model-debug": "^4.0.4-alpha.0",
+    "@nteract/transform-plotly": "^4.0.4-alpha.0",
+    "@nteract/transform-vdom": "^3.0.4-alpha.0",
+    "@nteract/transform-vega": "^4.0.4-alpha.0",
```
rgbkrk, line 31: I still don't want these as direct dependencies within the notebook app component

packages/notebook-preview/__tests__/__snapshots__/index.test.tsx.snap
```diff
@@ -223,19 +223,19 @@ print('hey')
...
-                "image/gif": [Function],
-                "image/jpeg": [Function],
-                "image/png": [Function],
-                "image/svg+xml": [Function],
-                "text/html": [Function],
-                "text/latex": [Function],
-                "text/markdown": [Function],
-                "text/plain": [Function],
-              }
+              Array [
```
rgbkrk: There must be either a prop ready for deletion or a defaultProp that needs to delete these transforms.

packages/outputs/__tests__/__snapshots__/media.spec.tsx.snap
```diff
@@ -1055,13 +1055,175 @@ exports[`HTML renders direct HTML 1`] = `
...
     }
-  />
+  >
+    <StyledComponent
+      dangerouslySetInnerHTML={
+        Object {
+          "__html": "<b>cats are best</b>",
+        }
+      }
+      forwardedComponent={
```
rgbkrk, line 1071: I wonder why this is all being written out here...

packages/outputs/src/components/kernel-output-error.tsx
```diff
@@ -39,20 +39,24 @@ const PlainKernelOutputError = (props: Props) => {
...
-  return <Ansi linkify={false}>{kernelOutputError.join("\n")}</Ansi>;
-};
-
-PlainKernelOutputError.defaultProps = {
-  outputType: "error",
-  ename: "",
-  evalue: "",
-  traceback: []
+  return (
+    <Ansi className="stacktrace" linkify={false}>
```
rgbkrk: The className should be `this.props.className`. That's what `styled-components` is going to pass through as the `className`(`s`)

packages/outputs/src/components/kernel-output-error.tsx
```diff
@@ -39,20 +39,24 @@ const PlainKernelOutputError = (props: Props) => {
...
+      {kernelOutputError.join("\n")}
+    </Ansi>
+  );
 };
 
 export const KernelOutputError = styled(PlainKernelOutputError)`
-  & code {
+  & .stracktrace {
```
rgbkrk: Typo here, though I think you can now just convert this whole thing to:

```css
white-space: pre-wrap;
```

with no extra qualifiers.

packages/transforms/src/index.tsx
```diff
@@ -80,7 +80,7 @@ export function richestMimetype(
       .sort((a, b) => order.indexOf(a) - order.indexOf(b))[0]
   );
 }
-export const transforms = standardTransforms;
+export const transforms = tfs;
```
rgbkrk: Doesn't this need to stay as `standardTransforms`?

captainsafia: Yep. You're right. Since they are no longer passed to the NotebookApp component, this can be `standardTransforms`. I had changed it to deal with a type error that no longer exists.

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -16,19 +15,33 @@ import {
...
 
+import PlotlyTransform, {
+  PlotlyNullTransform
+} from "@nteract/transform-plotly";
+import GeoJSONTransform from "@nteract/transform-geojson";
+import ModelDebug from "@nteract/transform-model-debug";
+import DataResourceTransform from "@nteract/transform-dataresource";
+import { VegaLite1, VegaLite2, Vega2, Vega3 } from "@nteract/transform-vega";
+import VDOMDisplay from "@nteract/transform-vdom";
+import { WidgetDisplay } from "@nteract/jupyter-widgets";
```
rgbkrk: Since we're talking about merging this in sooner rather than later, maybe we should take out the custom transforms for now and only stick in the default media ones.

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -252,34 +263,69 @@ class AnyCell extends React.PureComponent<AnyCellProps> {
...
-                  displayOrder={this.props.displayOrder}
-                  transforms={this.props.transforms}
-                  theme={this.props.theme}
-                  models={this.props.models}
-                  channels={this.props.channels}
-                  onMetadataChange={this.props.updateOutputMetadata}
-                  index={index}
-                  metadata={metadata}
-                />
+                <Output output={output.toJS()} key={index}>
```
rgbkrk: I also just noticed this... Which is definitely going to cause some perf thrashing. It's going to create the object each time. I _think_ because our records have the same structure now due to how you aligned everything on `camel_case` we can just pass the record `output`.

captainsafia: Yep. Passing just record works now. I can't recall what error I was addressing when I added this. I think it might have had something to do with a type error.

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
```diff
@@ -11,10 +11,22 @@ import {
   makeEntitiesRecord,
   makeHostsRecord,
   makeJupyterHostRecord,
-  makeStateRecord
+  makeStateRecord,
+  makeTransfromsRecord
```
rgbkrk:
```suggestion
  makeTransformsRecord
```

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
```diff
@@ -104,6 +116,20 @@ function main(rootEl: Element, dataEl: Node | null) {
               filepath: config.contentsPath
             })
           )
+        }),
+        transforms: makeTransfromsRecord({
```
rgbkrk:
```suggestion
        transforms: makeTransformsRecord({
```

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -252,34 +253,69 @@ class AnyCell extends React.PureComponent<AnyCellProps> {
...
-                <RichestMime
-                  metadata={{ expanded: true }}
-                  className="pager"
-                  displayOrder={this.props.displayOrder}
-                  transforms={this.props.transforms}
-                  bundle={pager}
-                  theme={this.props.theme}
-                  key={key}
-                />
+                <DisplayData {...pager}>
```
rgbkrk: What's up with the splat prop on the pager? Should this just be explicit about `data={pager.data}`, etc.?

packages/outputs/src/components/transform-media.tsx
```diff
@@ -0,0 +1,79 @@
+import React from "react";
```
rgbkrk: We should put these connected components within `notebook-app-component` since nothing else in this outputs package is a connected component. The `@nteract/outputs` components shouldn't have to rely on `core`, `types`, `redux`, etc. as we want Hydrogen to be willing to adopt it. The main barrier would be that this would install a _lot_ of things on atom plugin install that isn't even used.

packages/outputs/src/components/transform-media.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { connect } from "react-redux";
+import { Dispatch } from "redux";
+
+import { JSONObject } from "@nteract/commutable";
+import { actions, selectors } from "@nteract/core";
+import { AppState, ContentRef } from "@nteract/types";
+
+interface OwnProps {
+  id: string;
```
rgbkrk: Is the `id` being put in here for the future of having output IDs? If so, we should remove it for now so we're not leaving clutter that we have to clean up later (in case we don't get to revisit it).

captainsafia: This is the cellId. It's used to pull out the cell so that the outputs can be pulled from that. I realize that we modeled this to take an output, not a cell, so I will update.

packages/outputs/src/components/transform-media.tsx
captainsafia: Makes sense. I'll move the file.

packages/notebook-app-component/src/notebook-app.tsx
captainsafia: Yeah. We can be explicit here.

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
captainsafia: Oh gosh! I'm such a goof. I spent way too long trying to figure this out and didn't see that.

packages/outputs/src/components/transform-media.tsx
rgbkrk: Ahhhh! I suppose you need it for the action though so it's probably fine. Should just call it `cellId`

applications/desktop/src/notebook/index.tsx
```diff
@@ -59,6 +59,73 @@ const store = configureStore({
...
+          "text/html",
+          "text/markdown",
+          "text/latex",
+          "image/svg+xml",
+          "image/gif",
+          "image/png",
+          "image/jpeg",
+          "text/plain"
+        ]),
+        byId: Immutable.Map({
```
rgbkrk, line 87: Are `handlers` and `byId` overlapping in data? Do we only need `byId`?

applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
```diff
@@ -141,12 +77,34 @@ export default class Notebook extends React.PureComponent<
...
+            mediaType: transform.MIMETYPE,
+            component: transform
+          })
+        );
+      }
+    };
+  };
+  return mapDispatchToProps;
+};
+
```
rgbkrk, line 106: I'm really liking how you've done the `addTransform` here. This is solid.

packages/actions/src/actionTypes/index.ts
```diff
@@ -167,13 +167,19 @@ export interface MergeConfigAction {
 }
 
 export const LOAD_CONFIG = "LOAD_CONFIG";
-export interface LoadConfigAction { type: "LOAD_CONFIG" }
+export interface LoadConfigAction {
+  type: "LOAD_CONFIG";
+}
```
rgbkrk, line 172: I look forward to when all the tslint and prettier stuff is automated so we don't sneak in accidental changes to review. 😅

packages/actions/src/actionTypes/index.ts
```diff
@@ -217,5 +223,23 @@ export interface PublishGist {
...
+
+export const REMOVE_TRANSFORM = "REMOVE_TRANSFORM";
+export interface RemoveTransform {
+  type: "REMOVE_TRANSFORM";
+  payload: {
+    mediaType: string;
+    component: any;
+  };
+}
```
rgbkrk, line 242: Not sure when we'd use this but it makes sense.

packages/actions/src/actions/index.ts
```diff
@@ -240,6 +240,26 @@ export function publishGist(payload: {
   };
 }
 
+export function removeTransform(payload: {
```
rgbkrk: You don't have to do it in this PR if you don't want, thought I'd point out though that you can use `payload: actionTypes.RemoveTransform["payload"]` to tie this to the action's payload type directly.

packages/notebook-app-component/package.json
rgbkrk: With the refactor you've done here I _think_ you can delete these from the `package.json` and the `tsconfig.json`

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -251,34 +253,41 @@ class AnyCell extends React.PureComponent<AnyCellProps> {
...
-                  theme={this.props.theme}
-                  models={this.props.models}
-                  channels={this.props.channels}
-                  onMetadataChange={this.props.updateOutputMetadata}
-                  index={index}
-                  metadata={metadata}
-                />
+                <Output output={output} key={index}>
+                  <TransformMedia
+                    output_type={"display_data"}
```
rgbkrk, line 274: This is at least cleaner than before and now we have a nice connected component to work with. Well done.

packages/notebook-app-component/src/transform-media.tsx
```diff
@@ -0,0 +1,108 @@
...
+    outputData.keys((mediaType: string) => {
+      if (handlers[mediaType] && order.includes(mediaType)) {
+        return mediaType;
+      }
+    })
+  );
+  return validMediaTypes
+    .sort((a: string, b: string) => order.indexOf(a) - order.indexOf(b))
+    .get(0);
+};
```
rgbkrk, line 56: I enjoy that this is clean enough we can optimize it later nicely.

packages/notebook-app-component/src/transform-media.tsx
```diff
@@ -0,0 +1,108 @@
...
+  };
+  return mapDispatchToProps;
+};
+
+const TransformMedia = connect(
+  makeMapStateToProps,
+  makeMapDispatchToProps
+)(PureTransformMedia);
+
+export default TransformMedia;
```
rgbkrk, line 108: I'm really happy about this overall setup! It simplifies our rendering of rich outputs while optimizing how we work with them. One thing I feel like with this partial rewrite though is that we're missing some of the comments in either the old display-area bits or the `rich-media` component docs that provide context about selecting the richest. A few of these functions look eminently testable that we _might_ have had prior tests for(?).

packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -251,34 +253,41 @@ class AnyCell extends React.PureComponent<AnyCellProps> {
...
+                <DisplayData data={pager.data} metadata={pager.metadata}>
+                  <Media.Json />
+                  <Media.JavaScript />
+                  <Media.HTML />
+                  <Media.Markdown />
+                  <Media.LaTeX />
+                  <Media.SVG />
+                  <Media.Image />
+                  <Media.Plain />
+                </DisplayData>
```
rgbkrk: Since pagers haven't been showing anything much richer than HTML or Plain most of the time, I think this is a good solution. Also it's ok to use `<RichMedia>` from `@nteract/outputs` here.

packages/notebook-app-component/tsconfig.json
```diff
@@ -17,6 +17,14 @@
...
-    { "path": "../types" }
+    { "path": "../types" },
+    { "path": "../outputs" },
+    { "path": "../transform-plotly" },
+    { "path": "../transform-geojson" },
+    { "path": "../transform-model-debug" },
+    { "path": "../transform-dataresource" },
+    { "path": "../transform-vega" },
+    { "path": "../transform-vdom" },
+    { "path": "../jupyter-widgets" }
```
rgbkrk: Yup, looks like you can take `transform-*` and `jupyter-widgets` out of here since they're not getting used in `notebook-app-component` anymore.

captainsafia: Oh. Good call.

packages/notebook-app-component/src/notebook-app.tsx
captainsafia: Makes sense.

packages/notebook-app-component/src/transform-media.tsx
captainsafia: Hm. Yeah. I'll have to look into this as well as updating the component docs.

packages/actions/src/actions/index.ts
captainsafia: Ah yes. Fixed.

applications/desktop/src/notebook/index.tsx
captainsafia: Yeah. We only need byId so I'll remove handlers.

rgbkrk (approved):
Alright, looking great. The docs + tests can come in a follow up PR.

Thanks for carrying this through to completion!

PR #4024, opened 2019-01-13
tslint fix interface-over-type-literal

Shinigami92:

Automatic fix via tslint for the rule **interface-over-type-literal**


captainsafia (approved):
Thanks for the fix!

PR #4032, opened 2019-01-14
[WIP] Spread props for editor

rgbkrk:
Since a good number of things have changed in the repo since I kicked this PR off, I'm just going to post this so I have a nice easy way to read the changes and then I can make some new PRs.

The main goal here was to switch the codemirror editor to taking all the codemirror options as a direct prop instead of getting an object that is usually created on every render.

In the process I noticed we had a custom codemirror type instead of the one from `@types/codemirror`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
   },
   "private": true,
   "devDependencies": {
+    "@types/codemirror": "^0.0.71",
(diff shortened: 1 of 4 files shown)


rgbkrk:
Closing in favor of #4114.

PR #4036, opened 2019-01-15
Kick Off Styled Components Perf!

rgbkrk:
This begins tackling all the styled components that are using `props` to set up style. Since this effort is eminently parallelizable, I've carved out just a few and while doing searching internally removed another reference to `styled-jsx`.

Apparently we ended up creating lots of new entries in the `<head>` of the document each time we updated `props` on a component. For more info on the perf issue, check out: https://github.com/styled-components/styled-components/issues/134 Admittedly, I think we really only need to hit particular hot path items. With issues listed in #3434 still in effect, this all causes a cascade of performance issues.

## Guide to addressing this perf optimization for styled-components

Dynamic properties affected by `props` are the _only_ parts we have to change. Basically if there's a component using:

```js
const Box = styled.div`
  color: ${props => props.color};
  background-color: white;
`;
```

we should switch it to:

```js
const Box = styled.div.attrs(props => ({
    style: {
      color: props.color
    }
  })
)`
  background-color: white;
`;
```

Note: it's also ok to use class names like this:

```js
const Box = styled.div.attrs(props => ({
    className: props.focused ? "focused" : null
  })
)`
  &.focused {
    border: 2px solid grey
  }
`;
```

A pattern of which would have helped what I did in presentational components immensely. The final `className` will end up including what is set in the `attrs` here as well as styled-components generated `className` so it works out perfectly.

-------

Anything that's a constant is fine. It's only functions inside tagged template literals that are the ones to address.

### DIFF
--- a/applications/desktop/static/index.html
+++ b/applications/desktop/static/index.html
@@ -1,329 +1,322 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="UTF-8">
+    <meta charset="UTF-8" />
     <style>
-    /* The following is an excerpt from normalize.css v7.0.0, MIT License */
-    /*! normalize.css v7.0.0 | MIT License | github.com/necolas/normalize.css */
+      /* The following is an excerpt from normalize.css v7.0.0, MIT License */
+      /*! normalize.css v7.0.0 | MIT License | github.com/necolas/normalize.css */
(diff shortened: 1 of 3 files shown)


stormpython (approved):
LGTM!

PR #4038, opened 2019-01-15
tslint fix no-var-keyword and prefer-const

Shinigami92:

Automatic fix via tslint for the rules **no-var-keyword** and **prefer-const**
These are part of **tslint:recommended**


rgbkrk:
I'm going to assume the ones that got turned into `const` didn't need to be `let` and that the tslint setup would have figured that out.

Shinigami92:
@rgbkrk yea, I have to apologize. I didn't made this with lint:fix instead I run lint and made the needed changes by hand. I only wanted to lineup the description with my other PRs :P

PR #4039, opened 2019-01-15
More Styled components perf cleanup and tslint cleanup

stormpython:
Continuing the Styled components perf cleanup from #4036. Also cleaning up tslint warnings.

### DIFF
--- a/packages/connected-components/src/modal-controller/about-modal.tsx
+++ b/packages/connected-components/src/modal-controller/about-modal.tsx
@@ -87,7 +87,7 @@ const ModalContentOverlay = styled.div`
   z-index: 10000;
 `;
 
-class PureAboutModal extends React.Component<Props> {
+class PureAboutModal extends React.PureComponent<Props> {
   handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
     const { closeModal } = this.props;
     const { key, metaKey, altKey, ctrlKey, repeat, shiftKey } = event;
--- a/packages/connected-components/src/modal-controller/about-modal.tsx
+++ b/packages/connected-components/src/modal-controller/about-modal.tsx
@@ -87,7 +87,7 @@ const ModalContentOverlay = styled.div`
   z-index: 10000;
 `;
 
-class PureAboutModal extends React.PureComponent<Props> {
+class PureAboutModal extends React.Component<Props> {
   handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
     const { closeModal } = this.props;
     const { key, metaKey, altKey, ctrlKey, repeat, shiftKey } = event;
--- a/packages/directory-listing/src/components/icon.tsx
+++ b/packages/directory-listing/src/components/icon.tsx
@@ -1,7 +1,6 @@
 import { Book, FileDirectory, FileText } from "@nteract/octicons";
 import * as React from "react";
 import styled from "styled-components";
-import { ProgressPlugin } from "webpack";
 
 interface IconProps {
   color: string;
@@ -50,6 +49,6 @@ export class Icon extends React.PureComponent<IconProps> {
         icon = <FileText />;
     }
 
-    return <IconTD style={styles}>{icon}</IconTD>;
+    return <IconTD>{icon}</IconTD>;
   }
 }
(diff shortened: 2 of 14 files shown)


packages/connected-components/src/modal-controller/about-modal.tsx
```diff
@@ -87,7 +87,7 @@ const ModalContentOverlay = styled.div`
   z-index: 10000;
 `;
 
-class PureAboutModal extends React.Component<Props> {
+class PureAboutModal extends React.PureComponent<Props> {
```
rgbkrk: I should see if this tslint setting is actually checking for pure immutable props... otherwise it may be suggesting the wrong thing in some cases.

packages/directory-listing/src/components/icon.tsx
```diff
@@ -41,6 +46,6 @@ export class Icon extends React.Component<IconProps> {
         icon = <FileText />;
     }
 
-    return <IconTD style={{ color: this.props.color }}>{icon}</IconTD>;
+    return <IconTD style={styles}>{icon}</IconTD>;
```
rgbkrk: Whoa this shows we were duplicating. Do we need this still?

packages/dropdown-menu/src/index.tsx
```diff
@@ -85,7 +85,8 @@ const DropdownTriggerDiv = styled.div`
 
 DropdownTriggerDiv.displayName = "DropdownTriggerDiv";
 
-export class DropdownTrigger extends React.Component<{
+// tslint:disable max-classes-per-file
```
rgbkrk, line 88: Smart to do this for now since we can re enable this later.

rgbkrk:
I just realized I had drafted a review. Sending those off for now since I know you're about to rebase, then I can come back to it when you're ready.

packages/directory-listing/src/components/icon.tsx
stormpython: No, I don't think so. I will remove it.

stormpython:
@rgbkrk finally got the tests to pass, but I needed to revert a styled components change due to a type error. If you take a look at my last commit, you can see where I made changes.

rgbkrk:
Ok cool i'll look at it now.

packages/presentational-components/src/components/outputs.tsx
```diff
@@ -18,19 +18,16 @@ interface OutputWrapperProps {
   expanded?: boolean;
 }
 
-const OutputWrapper = styled.div.attrs<OutputWrapperProps>(props => ({
-  style: {
-    maxHeight: props.expanded ? "100%" : null
-  }
-}))`
```
rgbkrk: For now, the way to handle this is with an `as StyledComponent<"div", any, OutputWrapperProps, never>;` after the component declaration. I had to do this in https://github.com/nteract/nteract/pull/4036 as well.

rgbkrk:
Let's go ahead and merge this. I'll fix the outputs one after.

packages/directory-listing/src/components/icon.tsx
```diff
@@ -1,6 +1,7 @@
 import { Book, FileDirectory, FileText } from "@nteract/octicons";
 import * as React from "react";
 import styled from "styled-components";
+import { ProgressPlugin } from "webpack";
```
rgbkrk: This looks like an erroneous import.

packages/connected-components/src/modal-controller/about-modal.tsx
stormpython: Changing back to `Component`.

packages/presentational-components/src/components/outputs.tsx
stormpython: Will let @rgbkrk add a follow up PR for this change.

packages/directory-listing/src/components/icon.tsx
stormpython: Removed with new commit.

packages/presentational-components/src/components/outputs.tsx
rgbkrk: 👍

rgbkrk:
Ok great, now for my follow up PR. Thanks @stormpython!

PR #4042, opened 2019-01-16
yarn dev on kernel-relay now watches ts src

rgbkrk:
Now when hacking on `kernel-relay` you can run:

```
yarn dev
```

within the kernel-relay package to run both the typescript build in watch mode and the server in auto-reload.


captainsafia (commented):
Looks good but don't we need to add concurrently as a dev dependency to get this working?

captainsafia (approved):
Nevermind. I just saw that we already have concurrently as a developer dependency.

PR #4043, opened 2019-01-16
Small styled components perf cleanup

stormpython:
A styled components perf cleanup.

### DIFF
--- a/packages/transform-dataresource/src/tooltip-content.tsx
+++ b/packages/transform-dataresource/src/tooltip-content.tsx
@@ -45,7 +45,14 @@ const beforeContent = (props: Props) => {
   `;
 };
 
-const TooltipContent = styled.div`
+const TooltipContent = styled.div.attrs((props: Props) => ({
+  style: {
+    transform: translate(
+      props.x < 100 ? "0px" : "calc(-50% + 7px)",
+      props.y < 100 ? "10px" : "calc(-100% - 10px)"
+    )
+  }
+}))`
   color: black;
   padding: 10px;
   z-index: 999999;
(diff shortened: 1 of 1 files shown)


packages/transform-dataresource/src/tooltip-content.tsx
```diff
@@ -45,7 +45,14 @@ const beforeContent = (props: Props) => {
   `;
 };
 
-const TooltipContent = styled.div`
+const TooltipContent = styled.div.attrs((props: Props) => ({
+  style: {
+    transform: translate(
```
rgbkrk: The translate here needs to be wrapped in quotes. Likely:

```ts
const TooltipContent = styled.div.attrs((props: Props) => ({
  style: {
    transform: `translate(${props.x < 100 ? "0px" : "calc(-50% + 7px)"}, ${
      props.y < 100 ? "10px" : "calc(-100% - 10px)"
    })`
  }
```

packages/transform-dataresource/src/tooltip-content.tsx
```diff
@@ -55,11 +62,7 @@ const TooltipContent = styled.div`
...
   position: relative;
 
-  transform: translate(
-      ${(props: Props) => (props.x < 100 ? "0px" : "calc(-50% + 7px)")},
-      ${(props: Props) => (props.y < 100 ? "10px" : "calc(-100% - 10px)")}
-    )
-    & p {
+  & p {
     font-size: 14px;
   }
```
rgbkrk, line 68: Below here there's one more styled-components cleanup in `&:before` that is passing a function in called `beforeContent`.

rgbkrk:
With this file, #4045, #4046, and #4047 we are all done with the styled-components `props`athon cleanup.

stormpython:
Ok, I will fix this and then rebase.

packages/transform-dataresource/src/tooltip-content.tsx
rgbkrk: This still remains to fix, since it's basically a mixin that needs to move up into our props section. May take a bit of thinking to refactor up.

PR #4047, opened 2019-01-16
optimize styling of <Cell /> in notebook app

rgbkrk:
Another styled-components optimization. Fun thing I learned here is that `attrs` goes after the `styled(SomeComponent)` bit when it's a component you're (re)-styling. This one feels a little redundant since we also set this class on selected otherwise, but I feel like we have to ensure it for the cell toolbar. Working well in the app at least. 🤷‍♂️ 

## What's this perf enhancement all about?

Summary from https://github.com/nteract/nteract/pull/4036:

> Apparently we were creating lots of new entries in the `<head>` of our document on each render of a component. With issues listed in #3434 still in effect, this all causes a cascade of performance issues. Folks started noticing a huge performance regression after we introduced styled-components so here we take the liberty of optimizing the way folks suggest in styled-components/styled-components#134

### DIFF
--- a/packages/notebook-app-component/src/notebook-app.tsx
+++ b/packages/notebook-app-component/src/notebook-app.tsx
@@ -60,16 +60,15 @@ function getTheme(theme: string) {
   }
 }
 
-const Cell = styled(PlainCell)`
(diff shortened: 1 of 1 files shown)


captainsafia (approved):
Thanks for providing the implementation!

PR #4048, opened 2019-01-16
Polish fs-kernel comments to reflect new package structure

willingc:
- Remove references to the former kernelSpec package
- Add clarifying comments about the 0MQ connection
- Clarify return of JupyterConnectionInfo object (@captainsafia please check this closely)

### DIFF
--- a/packages/fs-kernels/README.md
+++ b/packages/fs-kernels/README.md
@@ -1,6 +1,7 @@
 # @nteract/fs-kernels
 
-This package contains functions for launching, getting stats on, and shutting down a Jupyter kernel.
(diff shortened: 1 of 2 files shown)


captainsafia (approved):
This looks good!

The `createConnectionInfo` does indeed return a JupyterConnectionInfo object. Our JupyterConnectionInfo definition doesn't include a version number so I modified the return definition here. We probably just want to make the version property an optional property in the JupyterConnectionInfo definition and avoid the whole `&` business here.

willingc:
Thanks @captainsafia

PR #4049, opened 2019-01-16
Edit readme definition of kernel-relay

willingc:


packages/kernel-relay/src/index.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @module kernel-relay
+ */
```
captainsafia, line 3: Good catch!

captainsafia (approved):
Looks good! Thank you.

PR #4051, opened 2019-01-16
refactor package commutable

Shinigami92:

- Make const pointers to functions
(this should not break anything, because there is no use of `this`)
- Renamed Notebook interfaces to reflect the version in it's name
This improves typescript usage of them
- Created typeguards for `NotebookV3` and `NotebookV4`
- Some additional small minor improvements


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules\\typescript\\lib"
```
captainsafia: What is this doing? And will it work on non-Windows machines?

packages/commutable/src/v4.ts
```diff
@@ -84,9 +79,9 @@ export interface RawCell {
 
 export type Cell = CodeCell | MarkdownCell | RawCell;
 
-export interface Notebook {
+export interface NotebookV4 {
   cells: Cell[];
-  metadata: Object;
+  metadata: object;
```
captainsafia: I think we can actually type this as a `JSONObject`.

captainsafia (commented):
Looks good overall. TypeDoc is rendering these under functions correctly now.

.vscode/settings.json
Shinigami92: woops, this file was not intended ^^
this could be added to .gitignore

Shinigami92:
I think it wants `metadata: plainNotebook.metadata.toJS()` to be `metadata: plainNotebook.metadata.toJS() as JSONObject` now?

packages/commutable/src/notebook.ts
```diff
@@ -34,10 +34,11 @@ export const makeNotebookRecord = Record<NotebookRecordParams>({
...
   Readonly<NotebookRecordParams>;
 
-const freezeReviver = <T extends JSONType>(_k: string, v: T) =>
-  Object.freeze(v) as T;
+function freezeReviver<T extends JSONType>(_k: string, v: T) {
+  return Object.freeze(v);
+}
 
-export type Notebook = v4.Notebook | v3.Notebook;
+export type Notebook = v4.NotebookV4 | v3.NotebookV3;
```
rgbkrk, line 41: Thanks for declobbering this namespace. 😄

packages/commutable/src/v3.ts
```diff
@@ -232,4 +238,13 @@ export const fromJS = (notebook: Notebook) => {
     nbformat: 4,
     metadata: immutableFromJS(notebook.metadata)
   });
-};
+}
+
+export function isNotebookV3(value: any): value is NotebookV3 {
```
rgbkrk, line 243: OMG. You have opened my eyes to a typescript feature I have always needed. Thank you.

rgbkrk (approved):

packages/commutable/src/v3.ts
Shinigami92: This is part of the FULL power of typescript

packages/commutable/src/notebook.ts
Shinigami92: If we make a complete rewrite I hope we can build functions like `fromJS` generic
So we can use in example fromJS<NotebookV4> where NotebookV4 extends from abstract/interface Notebook
But this would be another PR much later

Then we dont need v3. and v4. anymore

packages/commutable/src/v3.ts
rgbkrk: [I can not handle all this power.](https://www.youtube.com/watch?v=nyXu9rqtaDg)

[image](https://www.youtube.com/watch?v=nyXu9rqtaDg)

PR #4055, opened 2019-01-16
correct extension for editor spec

rgbkrk:
@Shinigami92 noticed this while cleaning up packages. `spec..tsx` just didn't have the right ring to it. 😅


rgbkrk:
CI done!

Kind of making me wonder if we should cache built typescript from prior builds (ensuring old timestamps so it rebuilds changed things).

PR #4057, opened 2019-01-16
Remove display-area and transforms

captainsafia:
This removes the packages and updates all the places they were used to use outputs.


applications/desktop/src/notebook/index.tsx
```diff
@@ -85,25 +94,17 @@ const store = configureStore({
...
+          "application/vnd.dataresource+json": DataResourceTransform,
+          "application/vnd.jupyter.widget-view+json": WidgetDisplay,
+          "application/vnd.vegalite.v1+json": VegaLite1,
+          "application/vnd.vegalite.v2+json": VegaLite2,
+          "application/vnd.vega.v2+json": Vega2,
+          "application/vnd.vega.v3+json": Vega3,
+          "application/vdom.v1+json": VDOMDisplay,
           "application/json": Media.Json,
           "application/javascript": Media.JavaScript,
           "text/html": Media.HTML,
```
rgbkrk, line 110: That's looking cleaner to lose the indirection. 👍

applications/desktop/tsconfig.json
```diff
@@ -7,9 +7,16 @@
...
-    { "path": "../../packages/fs-observable" }
+    { "path": "../../packages/fs-observable" },
+    { "path": "../../packages/jupyter-widgets" },
+    { "path": "../../packages/transform-dataresource" },
+    { "path": "../../packages/transform-geojson" },
+    { "path": "../../packages/transform-model-debug" },
+    { "path": "../../packages/transform-plotly" },
+    { "path": "../../packages/transform-vdom" },
+    { "path": "../../packages/transform-vega" },
+    { "path": "../../packages/outputs" }
```
rgbkrk, line 20: 😎

applications/jupyter-extension/nteract_on_jupyter/package.json
```diff
@@ -36,7 +36,6 @@
     "@nteract/transform-model-debug": "^4.0.4-alpha.0",
     "@nteract/transform-plotly": "^4.0.4-alpha.0",
     "@nteract/transform-vega": "^4.0.4-alpha.0",
-    "@nteract/transforms": "^5.0.4-alpha.0",
     "@nteract/types": "^3.1.4-alpha.0",
```
rgbkrk, line 39: Hmmmm... Is jupyter extension missing `@nteract/outputs` or is it not relevant here?

applications/jupyter-extension/package.json
```diff
@@ -15,5 +15,8 @@
   "private": true,
   "repository": "https://github.com/nteract/nteract/tree/master/applications/jupyter-extension",
   "author": "Kyle Kelley <rgbkrk@gmail.com>",
-  "license": "BSD-3-Clause"
+  "license": "BSD-3-Clause",
+  "dependencies": {
+    "@nteract/outputs": "^1.0.4-alpha.0"
```
rgbkrk: Oh I see. This dep needs to be moved into `applications/jupyter-extension/nteract_on_jupyter/package.json`

rgbkrk: In case you're wondering, the two package.json are here because of how python modules are always one extra nesting and we ship the assets with the package.

packages/editor/src/index.tsx
```diff
@@ -390,7 +390,7 @@ class CodeMirrorEditor extends React.PureComponent<
         const expanded = { expanded: true };
         const tipElement = ReactDOM.createPortal(
           <Tip className="CodeMirror-hint">
-            <RichestMime bundle={bundle} metadata={expanded} />
+            <RichMedia data={bundle} metadata={{ expanded }} />
```
rgbkrk: Does `RichMedia` come with default `children` to handle media types?

packages/notebook-preview/src/index.tsx
```diff
@@ -16,20 +23,15 @@ import {
...
+
 import * as React from "react";
 import styled, { createGlobalStyle } from "styled-components";
 
 import { PapermillView } from "./papermill";
 
 interface Props {
-  displayOrder: string[];
   notebook: any;
-  transforms: Transforms;
```
rgbkrk, line 32: We need to do one of two things with `notebook-preview`. Either we deprecate it entirely (and people are expected to build their own notebook with our other components) or we have to allow accepting some kind of prop for the media stuff. It could be a `children` prop that we re-use wherever we need it. We'd iterate through the children to see if there's a `RichMedia` defined within:

```tsx
<NotebookPreview>
  <RichMedia>
    <FancyTable />
    <Media.HTML />
  </RichMedia>
</NotebookPreview>
```

I'm not sure I'm real worried about making _more_ APIs so maybe deprecation of `notebook-preview` is fine?

packages/notebook-render/src/index.tsx
```diff
@@ -71,8 +74,6 @@ const Themes = {
 
 export default class NotebookRender extends React.PureComponent<Props, State> {
   static defaultProps = {
-    displayOrder: defaultDisplayOrder,
-    transforms: defaultTransforms,
```
rgbkrk, line 75: Same comment I made about `notebook-preview` applies here too.

rgbkrk (approved):
Much cleaner!

applications/jupyter-extension/package.json
captainsafia: Got it. Thanks.

packages/editor/src/index.tsx
captainsafia: Oh welp. Yeah. This needs to be set. I think in this case it is OK to just have the JSON transform as a child since the bundle is a JSON bundle.

rgbkrk: Typically we get back text/plain for this message type on the overall bundle. Trigger it with ctrl-shift-space on a module or keyword I believe

rgbkrk (approved):

Issue #3997, opened 2019-01-08
Fully migrate tests to TypeScript


captainsafia:
As part of the TypeScript conversion, we "migrated" all our tests from Flow to TypeScript. However, they are currently being transpiled only and not type checked. Before we enough type checking for our tests, we'll need to be sure that they pass the typechecker.

There are 60+ test files that need to be modified to pass the typechecker. Anyone can help with this conversion. To do so, you'll need to do the following.

Follow the steps in the [contributor guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md) to get a copy of the nteract codebase on your local machine.

1. Open package.json and find the `"isolatedModules": "true"` line and remove it.
2. Run `yarn` to install project dependencies and `yarn build` to run the monorepo packages.
3. Run `yarn test` and select one of the failing test files to fix.
4. Run `npx jest ${filename}` to run the tests for that single file.
5. Fix the test failures until `npx jest ${filename}` passes.
6. Commit and push the changes to the test file (or files if you did the process above for more than one file), not package.json.

If you have any questions, please post them below or on the [nteract Slack](https://slack.nteract.io/).

**Update** See message below: https://github.com/nteract/nteract/issues/3997#issuecomment-455378397

guntbert:
Hmm - gonna take you up on that "anyone" part :smile_cat: and jump in with no knowledge about the project.

`yarn` expects a file name as an argument but there is no file `test` to be found. Did I miss something?

captainsafia:
> Hmm - gonna take you up on that "anyone" part 😸 and jump in with no knowledge about the project.

Yes! All are welcome.

> yarn expects a file name as an argument but there is no file test to be found. Did I miss something?

`yarn` expects an argument that is one of the scripts defined in the package.json at the root of the project. Before you run `yarn test` make sure that you are running the command inside the nteract directory you cloned. Does that work?

guntbert:
Sorry, **what** command? To be clear: I tried to run `yarn test` in the nteract directory and got 

> No such file or directory: 'test'

captainsafia:
Hm. What operating system are you on and how did you install yarn?

guntbert:
Ubuntu, I used `sudo apt install cmdtest`

captainsafia:
Got it! In this case, can you try the process outlined in [this StackOverflow post](https://stackoverflow.com/a/47680872). It seems like it might be an issue with the yarn package. Installing it through npm might fix it.

[...]

PR #4058, opened 2019-01-17
Update test config

willingc:
Addresses #3997


scripts/test-shim.ts
```diff
@@ -1,8 +1,18 @@
+/**
+ * Mocks the window object
+ *
+ * Jest doesn't have direct access to the window object like the browser does.
+ * Mocking the window object allows Jest tests to interact with the window
+ * object.
```
rgbkrk: Probably worth clarifying that we mock / fake implement properties on the window object because JSDOM either doesn't support them or has only a partial implementation.

scripts/test-shim.ts
```diff
@@ -16,13 +26,15 @@ const createContextualFragment = html => {
 Range.prototype.createContextualFragment = html =>
   createContextualFragment(html);
 
+// tslint:disable-next-line:no-empty
```
rgbkrk: At the top of this file I think we can disable these 3 rules for the whole file:

```javascript
/* tslint:disable:no-empty no-console only-arrow-functions */
```

scripts/test-shim.ts
```diff
@@ -1,8 +1,18 @@
...
+
 // For some reason, this property does not get set above.
 global.Image = global.window.Image;
 
+// tslint:disable-next-line:no-empty
 global.Range = function Range() {};
 
+// tslint:disable-next-line:only-arrow-functions
 global.Blob = function(content, options) {
   return { content, options };
```
rgbkrk: I wonder if the reason we defined these as regular `function` is so it doesn't have a bound `this` (?)

willingc:
@rgbkrk @captainsafia Please take a look when you get a chance.

rgbkrk:
Wowsers, thanks! Checking it out now!

packages/transform-vega/__tests__/vega.test.tsx
```diff
@@ -71,7 +71,7 @@ describe("VegaEmbed", () => {
     expect(element.shouldComponentUpdate({ data: "324" })).toEqual(true);
   });
 
-  it("embeds vega and handles updates", () => {
+  it.skip("embeds vega and handles updates", () => {
```
rgbkrk, line 74: Do we want all these skipped?

tsconfig.base.json
```diff
@@ -10,6 +10,8 @@
     "target": "es2017",
     "module": "commonjs",
     "moduleResolution": "node",
+    "skipLibCheck": true,
```
rgbkrk, line 13: 👍

packages/transform-vega/__tests__/vega.test.tsx
willingc: Ultimately, no. For now, they are causing an error.

willingc: Figured we can open issues re: clean up for the 3 files that have skips

rgbkrk: Alright, that works for me.

willingc:
Thanks for the review and merge @rgbkrk

PR #4056, opened 2019-01-16
refactor with tslint

Shinigami92:
- [x] I have read the [Contributor Guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md)

_Please dont merge this know_

Fix tslint errors (currently especially errors!)

Ignoring errors for rules:

- `class-name`
- `jsx-no-lambda-props`
- `jsx-no-lambda`
- `jsx-no-object-literal-props`
- `no-empty`
- `react-a11y-event-has-role`
- `react-component-classes-should-implement-scu`
- `react-pure-components-have-simple-attributes`

Also ignored some `ban-types` and `strict-type-predicates`

I will create a follow-up PR for the tslint rule `typedef`, this will boost the projects type-definitions

PR #4059, opened 2019-01-17
Fix typescript warnings in fs-kernel

CrystallineCat:


rgbkrk (approved):
Great fixes, and happily this removes a bunch of variable name shadowing. Thank you!

packages/fs-kernels/src/spawnteract.ts
```diff
@@ -63,16 +63,16 @@ function createConnectionInfo(
...
 ): { version: number } & JupyterConnectionInfo {
   return {
-    version: 5,
-    key: uuid.v4(),
-    signature_scheme: "hmac-sha256",
-    transport: "tcp",
-    ip: "127.0.0.1",
-    hb_port: ports[0],
     control_port: ports[1],
+    hb_port: ports[0],
```
willingc: One small request for change:

Can we reorder to the original order instead of alphabetical? The original order kept the four ports needed for the 0MQ communication protocol together making a bit easier to scan if all four are returned.

willingc (commented):
This looks great. Thanks for taking the time to put it together. I have one small request to the PR. Thanks 👍

packages/fs-kernels/src/spawnteract.ts
rgbkrk: For this file we should add `// tslint:disable:object-literal-sort-keys` to the top so that tslint doesn't warn when it's for files we write to disk that we want to match the order jupyter writes them in.

CrystallineCat:
Thank you for the reviews! :) However, I've just noticed that PR #4056 now changes the same files (it didn't yesterday), and does many (although not all) of the same changes.

Should I still proceed with this, or close it in favour of #4056?

rgbkrk:
Go ahead and proceed with this and we can rebase #4056 since it has more remaining to fix. @Shinigami92 is waiting on some changes from me in #4061 which I'll have wrapped up shortly.

Shinigami92:
@CrystallineCat Yes please proceed, my #4056 is much more huge and I will manage to rebase my PR

packages/fs-kernels/src/spawnteract.ts
Shinigami92: Even better: implement an interface that holds the correct order and it will match with object-literal-sort-keys's match-declaration-order
The problem here is that `{ version: number } & JupyterConnectionInfo` is falling back to alphabetically because it is not one interface anymore

CrystallineCat: Ah, thank you! I'll look up how to do this.

rgbkrk:
Last thing it looks like is to remove the `WIP` tag in the title if you're all set. I'll do a quick look over the changes now.

rgbkrk (approved):

CrystallineCat:
Been trying to do this without the `tslint:disable`, as @Shinigami92 suggested above, but can't get it to work. I'll remove the WIP.

CrystallineCat:
Thank you all :)

rgbkrk:
All good, we can keep iterating on improving our types and making this all well covered and easy to work with.

PR #4061, opened 2019-01-17
unify commutable and records media bundle types

rgbkrk:
Media bundles are now frozen objects as part of `display_data`, `execute_result`, etc. rather than untyped `Map`s or overly restrictive records (since they don't allow additional keys). We thank `@nteract/records` for cleaning up the interfaces between `OnDisk` and in memory, and let it be free (I've deleted the package).


rgbkrk:
I totally expect this to fail at first since I have to adapt use of commutable by other packages.

rgbkrk:
This relieves a lot of stress for me around our partial venture into immer backed records (I guess this is remorse) and mismatching types amidst our components and commutable. @BenRussert -- commutable now uses records and our new output components can take either Immutable.js records or plain objects (as we were aiming for Hydrogen).

BenRussert:
Seems like a good plan here. And if we are still looking to try out immer someplace, I was thinking of trying it out on the data explorer component's internal state at some point to see if we can get a performance boost or other benefits.

rgbkrk:
There's an issue on master I'm going to fix. From chat (summarized):

```
rgbkrk [2:28 PM]:
I am currently flummoxed. It's like `React.cloneElement` is only passing props through to some of our components... oh hey this issue is on master too and relates to the outputs refactor, based on what I've figured out. Does `cloneElement` not like our record? Only stream and error output fail. Transform media is passed the output directly so it has no issue.

captainsafia [2:31 PM]
Can you give an example?

rgbkrk [2:31 PM]
run `print("hello")` in a cell

captainsafia [2:31 PM]
What the poop.

rgbkrk [2:31 PM]
transform media is passed the output directly so it has no issue

captainsafia [2:31 PM]
I tested the changes against the intro example notebook.

rgbkrk [2:32 PM]
maybe we only have display data in there?

captainsafia [2:32 PM]
Lemme try again.

rgbkrk [2:33 PM]
I wonder if a `yarn build:packages` was necessary
I'm on a freshly built setup
I mean... I suppose I could just make sure that my PR is good
since this is on master
and all the reducer stuff I changed is working out ok
then I can tackle this as a new thing to fix up

captainsafia [2:35 PM]
I’m working on verifying on my master.
But running a yarn clean and rebuilding everything.
Just to be sure.

rgbkrk [2:36 PM]
hmmmm
it's like I expected
and I think I see a fix
React is doing something special with records when they're the actual props
I think it must conflict with naming
and so we should pass `output` directly
as a property

captainsafia [2:37 PM]
To what?
The child components?

rgbkrk [2:37 PM]
yeah
so `<StreamOutput output={output} />`

captainsafia [2:37 PM]
That ends up borking the picker logic.
Then it defaults to the first child component since it clobbers the output type within those components.

rgbkrk [2:38 PM]
It would be the last step, as the `cloneElement`. The only badly gross aspect of this to me is that all our docs need some changing :face_with_rolling_eyes:
I liked the old interface

Here's the issue manifested in the chrome debugger:
```

screen shot 2019-01-17 at 2 40 14 pm

rgbkrk:
Ok this fixes up the issue on `master` with not all outputs being able to render.

rgbkrk:
This is worth a review now. I ended up fixing the output rendering issue that was on `master` as part of this since my changes here ripped out `@nteract/records` problems too.

yarn.lock
```diff
@@ -6629,10 +6637,10 @@ immer@1.7.2:
...
   integrity sha512-4Urocwu9+XLDJw4Tc6ZCg7APVjjLInCFvO4TwGsAYV5zT6YYSor14dsZR0+0tHlDIN92cFUOq+i7fC00G5vTxA==
 
-immer@^1.5.0:
-  version "1.9.3"
-  resolved "https://registry.yarnpkg.com/immer/-/immer-1.9.3.tgz#e88f8cbea730d2cde0f5e7e763000a8608bccdae"
-  integrity sha512-bUyz3fOHGn82V7h4oVgJGmFglZt53JWwSyVNAT4sO0d7IovHLwLuHbh14uYKY0tewFoDcEdiQW7HuL0NsRVziw==
+immutable@^3.8.1:
+  version "3.8.2"
+  resolved "https://registry.yarnpkg.com/immutable/-/immutable-3.8.2.tgz#c2439951455bb39913daf281376f1530e104adf3"
+  integrity sha1-wkOZUUVbs5kT2vKBN28VMOEErfM=
```
rgbkrk: These are wrong... I'm not sure what they're doing here.

rgbkrk: I wonder if it's from `redux-immutable` implicitly. We specify `4.0.0-rc.12` everywhere.

rgbkrk:
Rebased and now I'm seeing this on CI (yet not locally):

```
[12:22:19 AM] Building project '/home/circleci/nteract/packages/fs-kernels/tsconfig.json'...

packages/fs-kernels/src/kernel.ts:37:9 - error TS4053: Return type of public method from exported class has or is using name 'Stat' from external module "/home/circleci/nteract/node_modules/@types/pidusage/index" but cannot be named.

37   async getUsage() {
           ~~~~~~~~
```

packages/outputs/package.json
```diff
@@ -9,11 +9,16 @@
...
-    "@nteract/records": "^1.0.4-alpha.0",
     "ansi-to-react": "^4.0.4-alpha.0",
     "react-json-tree": "^0.11.0",
     "styled-components": "^4.1.3"
   },
+  "note:devDeps": [
+    "commutable is a development dependency as we should _only_ need it for types"
+  ],
+  "devDependencies": {
+    "@nteract/commutable": "^1.0.4-alpha.0"
```
rgbkrk: Whoops!

This should be the version in the repo itself.

rgbkrk:
Fixed the issue that's on CI in this PR: https://github.com/nteract/nteract/pull/4063

I'll rebase this when that's merged.

packages/commutable/src/outputs.ts
```diff
@@ -9,177 +9,86 @@ import {
...
-    ? s
-    : s.split(/(.+?(?:\r\n|\n))/g).filter(x => x !== "");
-}
-
-export function isJSONKey(key: string) {
-  return /^application\/(.*\+)?json$/.test(key);
-}
+import {
+  createFrozenMediaBundle,
+  demultiline,
```
willingc, line 14: demultiline and remultiline? Is this a toggle? Just curious.

packages/commutable/src/outputs.ts
```diff
@@ -231,17 +144,17 @@ export type Output = ExecuteResult | DisplayData | StreamOutput | ErrorOutput;
  *
  * @returns ImmutableOutput An immutable representation of the same output.
  */
-export function createImmutableOutput(output: Output): ImmutableOutput {
+export function createImmutableOutput(output: OnDiskOutput): ImmutableOutput {
```
willingc, line 147: Do we need to specify the return type is readonly?

packages/notebook-app-component/src/transform-media.tsx
```diff
@@ -3,66 +3,100 @@ import React from "react";
...
-        return mediaType;
-      }
-    })
-  );
-  return validMediaTypes
-    .sort((a: string, b: string) => order.indexOf(a) - order.indexOf(b))
-    .get(0);
+  const outputData = output.data;
+
+  // Find the first mediaType that we both support with handlers and is in the output data
```
willingc: Super nitpicky comment since this short circuits on the &&

Find the first media type in the output data that we support with a handler

packages/outputs/package.json
```diff
@@ -9,11 +9,16 @@
   "dependencies": {
     "@nteract/markdown": "^3.0.5-alpha.0",
     "@nteract/mathjax": "^3.0.5-alpha.0",
-    "@nteract/records": "^1.0.4-alpha.0",
     "ansi-to-react": "^4.0.4-alpha.0",
     "react-json-tree": "^0.11.0",
     "styled-components": "^4.1.3"
   },
+  "note:devDeps": [
+    "commutable is a development dependency as we should _only_ need it for types"
```
willingc: types for TypeScript? or something else?

packages/outputs/src/components/display-data.tsx
```diff
@@ -1,51 +1,35 @@
...
-   *
-   */
-  data: MediaBundle;
-  /**
-   * custom settings, typically keyed by media type
-   */
-  metadata: {};
-  /**
-   * React elements that accept mimebundle data, will get passed data[mimetype]
+   * React elements that accept media bundle data, will get passed data[mimetype]
```
willingc: mimetype or media type

packages/outputs/src/components/execute-result.tsx
```diff
@@ -8,44 +8,28 @@ interface Props {
...
-   *
-   */
-  data: MediaBundle;
-  /**
-   * custom settings, typically keyed by media type
-   */
-  metadata: {};
-  /**
-   * React elements that accept mimebundle data, will get passed data[mimetype]
+   * React elements that accept media bundle data, will get passed data[mimetype]
```
willingc: See prior comment

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -42,21 +43,29 @@ type KeyPaths = Immutable.List<KeyPath>;
  * @return {Immutable.List<Object>} updated-outputs - Outputs + Output
  */
 export function reduceOutputs(
-  outputs: Immutable.List<any> = Immutable.List(),
-  output: Output
+  outputs: Immutable.List<ImmutableOutput> = Immutable.List(),
```
willingc, line 46: Might be clearer to pull all the comments out for the function and make it a comment of the function.

willingc (commented):
Looks great @rgbkrk. A few clarifying questions - mostly for my benefit or new contribs.

packages/commutable/src/outputs.ts
rgbkrk: You know how our the `.ipynb` files on disk show source code like this:

```
      "source": [
        "import IPython\n",
        "IPython.display.display({'application/geo+json': geojson}, raw=True)"
      ]
```

`demultiline` turns that into:

`"import IPython\nIPython.display.display({'application/geo+json': geojson}, raw=True)"`

and `remultiline` turns it back into the array of strings.

packages/commutable/src/outputs.ts
rgbkrk: In this case we don't because they're `Immutable.Record`s

packages/notebook-app-component/src/transform-media.tsx
rgbkrk: Much better read, thank you.

packages/outputs/package.json
rgbkrk: I'm going to submit a slight change to this wording, though I'm not sure how much better I made it.

packages/commutable/src/outputs.ts
willingc: Maybe `unpackMultiline` `packMultiline`?

packages/outputs/src/components/display-data.tsx
rgbkrk: Good catch.

[Media types everywhere](https://www.iana.org/assignments/media-types/media-types.xhtml)

packages/reducers/src/core/entities/contents/notebook.ts
rgbkrk: I'm not understanding what to do here.

willingc: More like a docstring instead of individual comments of the code

willingc: The comments sort of distract from what is happening in the code

willingc (approved):
Awesome!

packages/commutable/src/outputs.ts
```diff
@@ -251,14 +164,24 @@ export function createImmutableOutput(output: Output): ImmutableOutput {
...
         traceback: ImmutableList(output.traceback)
       });
     default:
-      throw new TypeError(`Output type ${output.output_type} not recognized`);
+      // Since we're well typed, output is never. However we can still get new output types we don't handle
+      // and need to fail hard instead of making indeterminate behavior
+      const unknownOutput = output as any;
+      if (unknownOutput.output_type) {
+        throw new TypeError(
+          `Output type ${(output as any).output_type} not recognized`
```
BenRussert, line 180: Why not use `unknownOutput` here?

packages/commutable/src/primitives.ts
```diff
@@ -31,3 +31,173 @@ export type ImmutableJSONType =
...
+  "text/vnd.plotly.v1+html"?: MultiLineString | {};
+  "application/vnd.plotly.v1+json"?: {};
+  "application/geo+json"?: {};
+  "application/x-nteract-model-debug+json"?: {};
+  "application/vnd.vega.v2+json"?: {};
+  "application/vnd.vega.v3+json"?: {};
+  "application/vnd.vegalite.v1+json"?: {};
+  "application/vnd.vegalite.v2+json"?: {};
+
+  [key: string]: string | string[] | {} | undefined | undefined;
```
BenRussert, line 69: `undefined | undefined;`

packages/commutable/src/primitives.ts
```diff
@@ -31,3 +31,173 @@ export type ImmutableJSONType =
...
+  return s.split(/(.+(:\r\n|\n))/g).filter(x => x !== "");
+}
+
+function isJSONKey(key: string) {
+  return /^application\/(.*\+)json$/.test(key);
+}
+
+export function deepFreeze(object: string | { [key: string]: any }) {
+  if (typeof object === "string") {
+    return Object.freeze(object);
```
BenRussert, line 128: In my mind strings are immutable, so I'm just wondering if this does anything I might be completely unaware of?

packages/commutable/src/primitives.ts
```diff
@@ -31,3 +31,173 @@ export type ImmutableJSONType =
...
+  }
+
+  // Retrieve the property names defined on object
+  const propNames = Object.getOwnPropertyNames(object);
+
+  // Freeze properties before freezing self
```
BenRussert, line 134: Good advice. I love the unusual combinations of English words like this that make perfect sense programming :rofl:

packages/outputs/src/components/rich-media.tsx
```diff
@@ -28,13 +29,13 @@ interface RichMediaProps {
    * for more detail.
    *
    */
-  data: MediaBundle;
+  data: Readonly<MediaBundle>;
   /**
    * custom settings, typically keyed by media type
    */
-  metadata: { [mediaType: string]: object };
+  metadata: Readonly<{ [mediaType: string]: object }>;
```
BenRussert, line 36: I wanted to check my understanding of `Object` vs `object`. 

From the official docs, `object`, used here, refers to "any thing that is not number, string, boolean, symbol, null, or undefined."

So `metadata` here could be an array, etc.

packages/reducers/__tests__/document.spec.ts
```diff
@@ -91,65 +96,75 @@ describe("reduceOutputs", () => {
...
+          output_type: "stream"
+        }),
+        makeStreamOutput({
+          name: "stderr",
+          text: "errors are",
+          output_type: "stream"
+        })
+      ])
     );
+    const x = 2;
```
BenRussert, line 150: :fire:

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -65,28 +74,22 @@ export function reduceOutputs(
     return text;
   }
 
+  // Invariant: size > 0, outputs.last() exists
+  if (last.name === streamOutput.name) {
+    return outputs.updateIn([outputs.size - 1, "text"], appendText);
+  }
+
+  // Check if there's a separate stream to merge with
+  const nextToLast = outputs.butLast().last(null);
```
BenRussert, line 83: :sunglasses: :sparkles:

BenRussert (commented):
Great cleanup here. Nice work!

packages/outputs/src/components/rich-media.tsx
rgbkrk: Oh darn this should be switched to that JSON item we use in other places, to include the other primitives.

packages/reducers/__tests__/document.spec.ts
rgbkrk: Ha. Now everyone knows how I check to see if typescript is really checking things.

packages/commutable/src/primitives.ts
rgbkrk: I honestly borrowed this from MDN, so I would hope there's good reason. I'm with you on the assumption strings are immutable. Maybe it's properties aren't?

packages/commutable/src/primitives.ts
rgbkrk: 0_____0

packages/commutable/src/outputs.ts
rgbkrk: Whoops that's what was meant to be used.

packages/commutable/src/primitives.ts
rgbkrk: Wow I think this is a result of multiple iterations of flow and typescript that led to this happening. Our code is _close_ to what's on [MDN for deep freeze](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze). This seems like something I can clean up and provide a better `readonly` type for it.

PR #4063, opened 2019-01-18
expose explicit usage api for kernels

rgbkrk:
Ran into this issue on `master`. The type return for `getUsage` is flaky for our end users because it depends on `@types/pidusage`. Our typescript config now errors on this so I had to export an interface. Luckily if the types change we'll know what's missing and what we're expecting to expose to our own users.

### DIFF
--- a/packages/fs-kernels/src/kernel.ts
+++ b/packages/fs-kernels/src/kernel.ts
@@ -13,7 +13,7 @@ import { cleanup, launch, LaunchedKernel, launchSpec } from "./spawnteract";
 
 interface Usage {
   /**
-   * percentage (from 0 to 100*vcore)
+   * percentage (from 0 to 100*number of cores)
    */
   cpu: number;
 
@@ -23,27 +23,27 @@ interface Usage {
   memory: number;
 
   /**
-   * PPID
+   * PPID - Parent PID
    */
   ppid: number;
 
   /**
-   * PID
+   * PID - Process ID
    */
   pid: number;
 
   /**
-   * ms user + system time
+   * ms user + system time (in local time)
    */
   ctime: number;
 
   /**
-   * ms since the start of the process
+   * ms since the start of the process (local time)
    */
   elapsed: number;
 
   /**
-   * ms since epoch
+   * ms since epoch (local time)
    */
   timestamp: number;
 }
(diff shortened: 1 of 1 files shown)


packages/fs-kernels/src/kernel.ts
```diff
@@ -11,6 +11,43 @@ import { JupyterConnectionInfo } from "enchannel-zmq-backend";
 import { KernelSpec } from "./kernelspecs";
 import { cleanup, launch, LaunchedKernel, launchSpec } from "./spawnteract";
 
+interface Usage {
+  /**
+   * percentage (from 0 to 100*vcore)
```
willingc: vcore? Is this the percentage of CPU used including all cores

packages/fs-kernels/src/kernel.ts
```diff
@@ -11,6 +11,43 @@ import { JupyterConnectionInfo } from "enchannel-zmq-backend";
...
+   */
+  cpu: number;
+
+  /**
+   * bytes
+   */
+  memory: number;
+
+  /**
+   * PPID
```
willingc: PPID?

packages/fs-kernels/src/kernel.ts
```diff
@@ -11,6 +11,43 @@ import { JupyterConnectionInfo } from "enchannel-zmq-backend";
...
+   */
+  ppid: number;
+
+  /**
+   * PID
+   */
+  pid: number;
+
+  /**
+   * ms user + system time
```
willingc: local time?

willingc (commented):
A few questions to clarify the usage elements. Otherwise +1

rgbkrk:
Look at where copy + paste takes me. I guess I better more fully document what pidusage exposes. 😆

packages/fs-kernels/src/kernel.ts
rgbkrk: Since I think all that @captainsafia was trying to incorporate here was `cpu` and `memory`, perhaps I should tidy this up to only surface those two properties.

packages/fs-kernels/src/kernel.ts
rgbkrk: ¯\\\_(ツ)\_/¯

rgbkrk: Oh I bet it's parent PID.

willingc: That makes sense

willingc (approved):
Beautiful 👍

PR #4064, opened 2019-01-18
Add shim to suppress vega test errors and remove test skips

willingc:
Addresses #3997

### DIFF
--- a/scripts/test-shim.ts
+++ b/scripts/test-shim.ts
@@ -49,3 +49,6 @@ process.on("unhandledRejection", (error, promise) => {
   console.error(error.stack);
   promise.catch(err => console.error("promise rejected", err));
 });
+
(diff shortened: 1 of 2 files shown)


scripts/test-shim.ts
```diff
@@ -49,3 +49,6 @@ process.on("unhandledRejection", (error, promise) => {
   console.error(error.stack);
   promise.catch(err => console.error("promise rejected", err));
 });
+
+// HACK: To avoid error of NotImplemented in vega tests
+HTMLCanvasElement.prototype.getContext = jest.fn();
```
rgbkrk, line 54: Nice! That turned out to a be a nice tidy change to the test shim for these.

rgbkrk (approved):

PR #4066, opened 2019-01-18
Provide deep frozen type and cleanup related parts

rgbkrk:
Cleans up #4061

cc @BenRussert


captainsafia (approved):
This is a good and noble PR.

PR #4067, opened 2019-01-18
correct the kernel status bar logic

rgbkrk:
On `master` I noticed (really a while ago, during the typescript conversion) that the status bar isn't reporting the kernel properly. It looks like the `null` check was reversed. We're back to having good status:

screen shot 2019-01-18 at 6 12 17 am


rgbkrk:
cc @captainsafia

captainsafia (approved):
Typical Safia move for me not to have noticed this when I was looking at it. Thank you for fixing this!

rgbkrk:
> Typical Safia move for me not to have noticed this when I was looking at it.

Nah, I don't think it was about you. 🙃  There's a good amount of indirection with the `selectors` and `state` that I'm not surprised at this. We've got to simplify more of this code base while keeping the types nice and tidy.

rgbkrk:
One more for this effort: #4073

PR #4073, opened 2019-01-18
pluck kernelRef from content model

rgbkrk:
Hopefully this is the last vestige of the old global kernel setup. Any `contentRef` that identifies a notebook should pluck the kernel from the notebook model directly.

This fixes the status bar on desktop.


captainsafia (approved):
Ah! I see what was going on here. Thanks for putting in the fix.

PR #4075, opened 2019-01-18
remove indirection around creating cells

rgbkrk:
We've done enough cleanup within the code base that we can remove the `renderCell` and `createCellElement` helpers. Check out how clean this is now!


captainsafia (approved):
This is a wholesome, FDA-approved PR.

PR #4077, opened 2019-01-18
Edit docs and comments for commutable

willingc:
Commutable is an important package in our pipeline architecture. This PR edits the doc prose and attempts to provide more context on the package for a new contributor.


rgbkrk:
Thank you.

PR #4079, opened 2019-01-18
Remove dependency on canvas-prebuilt

### LINKED ISSUE #4070: yarn install fails in node 11 (ABI v67) due to missing canvas-prebuilt@2.0.0-alpha.14 at that ABI
The install of `canvas-prebuilt@2.0.0-alpha.14` (dependency added in #4058 by @willingc) fails in Node v11.6.0:

    node-pre-gyp ERR! stack Error: 404 status code downloading tarball https://github.com/chearon/node-canvas-prebuilt/releases/download/v2.0.0-alpha.14/canvas-prebuilt-v2.0.0-alpha.14-node-v67-linux-glibc-x64.tar.gz

It seems `canvas-prebuilt@2.0.0` (proper) is the first version to support Node 11 (ABI v67), but that release (and future releases) are no longer directly available via npm:

> Actually you can just `npm install canvas@2.0.1` and it downloads prebuilds from here! I won't be publishing `canvas-prebuilt@2.x` since I don't think there's a reason to anymore.

_Originally posted by @chearon in https://github.com/node-gfx/node-canvas-prebuilt/issues/58#issuecomment-430808805_

I'm a bit confused by the commit message of the commit that added the dependency an `canvas-prebuilt` (f7ed180878112878ea847da5d775a2fb436b39d1). The message refers to the removal of a dependency on `canvas`, which neither the commit nor the PR appear to have done. The modified `package.json` doesn't/didn't contain such a dependency.

CrystallineCat:
Fixes #4070 (probably; didn't actually test a stable install; passes all tests in LTS)

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,6 @@
     "@types/styled-components": "^4.1.4",
     "@types/styled-jsx": "^2.2.7",
     "@types/webfontloader": "^1.6.29",
-    "canvas-prebuilt": "^2.0.0-alpha.14",
     "concurrently": "^4.0.0",
     "cpy-cli": "^2.0.0",
     "cross-env": "^5.2.0",


willingc (approved):
Thanks @CrystallineCat for the PR and the catch earlier :cake:

PR #4080, opened 2019-01-18
Deprecate @nteract/gatsby-transformer-ipynb package

captainsafia:


rgbkrk:
Will this move out to `nteract/gatsby-transformer-ipynb`? Guess that's up to @benjaminabel

captainsafia:
For now, I think it makes sense for people to use the package under [gatsby-contrib](https://github.com/gatsby-contrib/gatsby-transformer-ipynb) since that is the one being actively maintained.

rgbkrk:
Fantastic.

PR #4081, opened 2019-01-18
Add The Option to Use Classes [pkg:ansi-to-react]

bnbarak:
# Background
The `anser` returns the rgb value.
There is an option to return class names instead, making the component more configurable. 

# How
Pass the option `use_classes` to `ansiToJson()`
```javascript
function ansiToJSON(input: string, use_classes = false) {
  input = escapeCarriageReturn(input);
  return ansiToJson(input, {
    json: true,
    remove_empty: true,
    use_classes
  });
}
```
and then, render the span with the class(es)
```javascript
const style = useClasses ? null : createStyle(bundle);
const className = useClasses ? createClass(bundle) : null;

  if (!linkify) {
    return React.createElement(
      "span",
      { style, key, className },
      bundle.content
    );
  }
```

# Example
```javascript
<Ansi useClasses>
    {'\u001b[34mhello world'}
</Ansi>;
```
Will render 
```javascript
<code>
    <span class="ansi-blue">hello world</span>
</code>
```

Feel free to provide me with feedback, changes.

Barak

### DIFF
--- a/packages/ansi-to-react/README.md
+++ b/packages/ansi-to-react/README.md
@@ -13,7 +13,7 @@ $ npm install --save @nteract/ansi-to-react
 ```
 
 ## Usage
-
+### Basic
 The example below shows how we can use this package to render a string with ANSI escape codes.
 
 ```javascript
@@ -21,10 +21,42 @@ import Ansi from "@nteract/ansi-to-react";
 
 export function () => {
   return <Ansi>
-    {'\u001b[34mnode_modules\u001b[m\u001b[m'}
+    {'\u001b[34mhello world'}
   </Ansi>;
 };
 ```
+Will render
+```javascript
+<code>
+    <span style="color:rgb(0, 0, 187)">hello world</span>
+</code>
+```
+
+### Classes
+Style with classes instead of `style` attribute.
+```javascript
+<Ansi useClasses>
+    {'\u001b[34mhello world'}
+</Ansi>;
+```
+Will render
+```javascript
+<code>
+    <span class="ansi-blue">hello world</span>
+</code>
+```
+
+#### Class Names
+|Font color| Background Color
+|---|---|
+|ansi-black|ansi-bright-black
+|ansi-red|ansi-bright-red
+ansi-green|ansi-bright-green
+ansi-yellow|ansi-bright-yellow
+ansi-blue|ansi-bright-blue
+ansi-magenta|ansi-bright-magenta
+ansi-cyan|ansi-bright-cyan
+ansi-white|ansi-bright-white
 
 ## Documentation
 
--- a/packages/ansi-to-react/__tests__/index.spec.ts
+++ b/packages/ansi-to-react/__tests__/index.spec.ts
@@ -1,4 +1,4 @@
-import { shallow } from "enzyme";
+import { shallow, render } from "enzyme";
 import * as React from "react";
 
 import Ansi from "../src/index";
@@ -102,5 +102,67 @@ describe("Ansi", () => {
     expect(el.text()).toBe(
       "<module 'something' from '/usr/local/lib/python2.7/dist-packages/something/__init__.pyc'>"
     );
+  });
+
+  describe("use classes", () => {
+    test("color only background", () => {
+      const el = shallow(
+        React.createElement(
+          Ansi,
+          { useClasses: true },
+          `hello ${GREEN_FG}world`
+        )
+      );
+      expect(el).not.toBeNull();
+      expect(el.text()).toBe("hello world");
+      expect(el.html()).toBe(
+        "<code><span>hello </span><span class=\"ansi-green\">world</span></code>"
+      );
+    });
+
+    test("useClasses option", () => {
+      const el = shallow(
+        React.createElement(
+          Ansi,
+          { useClasses: true },
+          `hello ${YELLOW_BG}world`
+        )
+      );
+      expect(el).not.toBeNull();
+      expect(el.text()).toBe("hello world");
+      expect(el.html()).toBe(
+        "<code><span>hello </span><span class=\"ansi-yellow\">world</span></code>"
+      );
+    });
+
+    test("color text and background", () => {
+      const el = shallow(
+        React.createElement(
+          Ansi,
+          { useClasses: true },
+          `hello ${GREEN_FG}${YELLOW_BG}world`
+        )
+      );
+      expect(el).not.toBeNull();
+      expect(el.text()).toBe("hello world");
+      expect(el.html()).toBe(
+        "<code><span>hello </span><span class=\"ansi-yellow ansi-green\">world</span></code>"
+      );
+    });
+
+    test("useClasses with linkify", () => {
+      const el = shallow(
+        React.createElement(
+          Ansi,
+          { linkify: true, useClasses: true },
+          `${GREEN_FG}this is a link: https://nteract.io/`
+        )
+      );
+      expect(el).not.toBeNull();
+      expect(el.text()).toBe("this is a link: https://nteract.io/");
+      expect(el.html()).toBe(
+        "<code><span class=\"ansi-green\">this is a link: <a href=\"https://nteract.io/\" target=\"_blank\">https://nteract.io/</a></span></code>"
+      );
+    });
   });
 });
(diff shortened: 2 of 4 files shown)


packages/ansi-to-react/README.md
```diff
@@ -13,18 +13,50 @@ $ npm install --save @nteract/ansi-to-react
...
+|Font color| Background Color
+|---|---|
+|ansi-black|ansi-bright-black
+|ansi-red|ansi-bright-red
+ansi-green|ansi-bright-green
+ansi-yellow|ansi-bright-yellow
+ansi-blue|ansi-bright-blue
+ansi-magenta|ansi-bright-magenta
+ansi-cyan|ansi-bright-cyan
+ansi-white|ansi-bright-white
```
rgbkrk, line 59: I suppose we should export these too.

packages/ansi-to-react/__tests__/index.spec.ts
```diff
@@ -103,4 +103,66 @@ describe("Ansi", () => {
...
+          `${GREEN_FG}this is a link: https://nteract.io/`
+        )
+      );
+      expect(el).not.toBeNull();
+      expect(el.text()).toBe("this is a link: https://nteract.io/");
+      expect(el.html()).toBe(
+        "<code><span class=\"ansi-green\">this is a link: <a href=\"https://nteract.io/\" target=\"_blank\">https://nteract.io/</a></span></code>"
+      );
+    });
+  });
```
rgbkrk, line 167: Tests! ❤️

packages/ansi-to-react/src/index.ts
```diff
@@ -11,34 +11,80 @@ const LINK_REGEX = /(https?:\/\/(?:www\.|(?!www))[^\s\.]+\.[^\s]{2,}|www\.[^\s]+
...
+
+  if (!bundle.bg && !bundle.fg) {
+    return null;
+  }
+  if (bundle.bg) {
+    classNames += bundle.bg + " ";
+  }
+  if (bundle.fg) {
+    classNames += bundle.fg + " ";
+  }
```
rgbkrk, line 41: This puts the `bundle.bg` as the className. Does this... only make sense when `useClasses` is true? Would this end up putting `rgb(0,0,0)` as a `className` with the right conditions?

packages/ansi-to-react/src/index.ts
```diff
@@ -11,34 +11,80 @@ const LINK_REGEX = /(https?:\/\/(?:www\.|(?!www))[^\s\.]+\.[^\s]{2,}|www\.[^\s]+
...
 ) {
-  const style: { backgroundColor?: string; color?: string } = {};
-  if (bundle.bg) {
-    style.backgroundColor = `rgb(${bundle.bg})`;
-  }
-  if (bundle.fg) {
-    style.color = `rgb(${bundle.fg})`;
-  }
+  const style = useClasses ? null : createStyle(bundle);
+  const className = useClasses ? createClass(bundle) : null;
```
rgbkrk, line 80: Answered my own question just now. It's either the inline style from the bundle or the classes.

packages/ansi-to-react/src/index.ts
```diff
@@ -11,34 +11,80 @@ const LINK_REGEX = /(https?:\/\/(?:www\.|(?!www))[^\s\.]+\.[^\s]{2,}|www\.[^\s]+
...
 
+/**
+ * Create a class string.
+ * @name createClass
+ * @function
+ * @param {AnserJsonEntry}.
+ * @return {String} class name(s)
+ */
+function createClass(bundle: AnserJsonEntry) {
+  let classNames: String = "";
```
BenRussert:
```ts
let classNames: string = "";
```

packages/ansi-to-react/src/index.ts
```diff
@@ -69,21 +115,23 @@ function convertBundleIntoReact(
...
 
 export default function Ansi(props: Props) {
+  const { className, useClasses, children, linkify } = props;
   return React.createElement(
     "code",
-    { className: props.className },
-    ansiToJSON(props.children).map(
-      convertBundleIntoReact.bind(null, props.linkify)
+    { className },
+    ansiToJSON(children, useClasses).map(
```
BenRussert: If you do 
```ts
ansiToJSON(children, !!useClasses).map(...
```  
the `!!` will coerce `useClasses` to a bool instead of undefined to make typescript happy.

BenRussert (approved):
This looks great @bnbarak. I left  a couple comments that will hopefully get CI to pass.

BTW, I'd love to see your project if you are using or looking at using this!

packages/ansi-to-react/README.md
bnbarak: You mean let the user pick the class names? I thought about adding a prefix prop to replace the `ansi`. Otherwise, how do you suggest we go about it?

packages/ansi-to-react/src/index.ts
bnbarak: Thank, first time using TypeScript lol

packages/ansi-to-react/src/index.ts
bnbarak: sweet, thanks

bnbarak:
Thanks for the comments @BenRussert @rgbkrk 

This is how we in Comet using your component https://www.comet.ml/cometpublic/parameter-space-exploration/0b05c7d070f14e36acd34353d892d717/stdout

Let me know if you have any other parts that you guys need help with. Seems like a cool opensource project that I will be glad to help with!

bnbarak:
Tests fails on an old snapshot due to the new prop `classNames`

```
2019-01-21T21:16:14.7922837Z FAIL packages/outputs/__tests__/media.spec.tsx
2019-01-21T21:16:14.7923510Z   ● Plain › Should render markdown
2019-01-21T21:16:14.7923552Z 
2019-01-21T21:16:14.7923650Z     expect(value).toMatchSnapshot() 
```

rgbkrk:
> Tests fails on an old snapshot due to the new prop `classNames`

To update snapshots, run `yarn test --updateSnapshot`

rgbkrk:
> Let me know if you have any other parts that you guys need help with. Seems like a cool opensource project that I will be glad to help with!

Welcome! We're happy to have you!

packages/ansi-to-react/README.md
rgbkrk: As in export the names that will be used, so someone can read/use them programmatically or at the very least be able to test against them.

packages/ansi-to-react/__tests__/index.spec.ts
```diff
@@ -103,4 +103,66 @@ describe("Ansi", () => {
...
+  describe("useClasses options", () => {
+    test("can add the font color class", () => {
+      const el = shallow(
+        React.createElement(
+          Ansi,
+          { useClasses: true },
+          `hello ${GREEN_FG}world`
+        )
+      );
+      expect(el).not.toBeNull();
```
rgbkrk, line 116: It'd be alright to use jest snapshots here too. I do realize some of the old tests had some anti-patterns. In the future it'd be cool to make the tests something like:

```tsx
import React from 'react';
import renderer from 'react-test-renderer';

it('can render with classnames', () => {
  const tree = renderer
    .create(<Ansi useClasses>{`hello ${GREEN_FG}world`}</Ansi>)
    .toJSON();
  expect(tree).toMatchSnapshot();
});
```

However I wouldn't worry about it for this PR. Would be a nice follow up if you've got the time.

packages/ansi-to-react/src/index.ts
```diff
@@ -69,21 +115,23 @@ function convertBundleIntoReact(
...
 }
 
 export default function Ansi(props: Props) {
+  const { className, useClasses, children, linkify } = props;
   return React.createElement(
     "code",
-    { className: props.className },
-    ansiToJSON(props.children).map(
-      convertBundleIntoReact.bind(null, props.linkify)
+    { className },
```
rgbkrk, line 132: Now that I see the snapshot below, it's probably best we don't include `className` if we're not using it.

rgbkrk:
Ok I think even with my last comment about when `useClasses` is `false` (and `className` ends up `null`), I think we can call this a minor release. Going ahead with merging. I think we're also close enough on other parts of the monorepo that we can have a real release instead of the alpha releases we've gated ourselves on while doing the typescript and package migration.

rgbkrk:
That being said, I think I'd want to aim for shipping on Thursday.

rgbkrk:
Welcome to the nteract party @bnbarak! 🎉

PR #4082, opened 2019-01-19
Add plugin for styled-components

Shinigami92:
## Before:

before

## After:

after

## Also documentation!

docs color
docs padding

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -7,7 +7,8 @@
     "editorconfig.editorconfig",
     "eamodio.gitlens",
     "esbenp.prettier-vscode",
-    "wayou.vscode-todo-highlight"
+    "wayou.vscode-todo-highlight",
+    "jpoissonnier.vscode-styled-components"
   ],
   // List of extensions recommended by VS Code that should not be recommended for users of this workspace.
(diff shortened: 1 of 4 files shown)


.vscode/extensions.json
```diff
@@ -7,7 +7,8 @@
     "editorconfig.editorconfig",
     "eamodio.gitlens",
     "esbenp.prettier-vscode",
-    "wayou.vscode-todo-highlight"
+    "wayou.vscode-todo-highlight",
+    "jpoissonnier.vscode-styled-components"
```
rgbkrk, line 11: Nice! Yeah I've been using this one with much developer happiness.

tsconfig.base.json
```diff
@@ -12,7 +12,10 @@
     "moduleResolution": "node",
     "skipLibCheck": true,
     "noImplicitAny": true,
-    "plugins": [{ "name": "typescript-tslint-plugin" }],
+    "plugins": [
+      { "name": "typescript-styled-plugin" },
```
rgbkrk, line 16: Wow I didn't know about this. I wonder if this means I should disable `csslint` in my vscode setup.

rgbkrk (approved):

Issue #4078, opened 2019-01-18
yarn app:commuter fails on MacOS


keitakurita:
**Application or Package Used**
commuter

**Describe the bug**
Running `yarn app:commuter` returns the message 'error Command "app:commuter" not found'

**To Reproduce**

1. Clone repo
2. Run `yarn install` on master branch
3. Run `yarn app:commuter`

**Expected behavior**
According to #3654 the development version is supposed to work. However, I cannot seem to run commuter via the commands provided. 

**Desktop (please complete the following information):**

- OS: macOS
- Version 10.13.4

**Additional context**
node version: 10.15.0
yarn version: 1.13.0

Thanks!

willingc:
Hi @keitakurita,

Thanks for the report, and sorry  for the confusion. We are in the process of refactoring portions of the code base. Commuter functionality can be found in the commutable package.

I've opened a PR to give a message to those that type the command. We should update docs too.

keitakurita:
@willingc 

Thanks for the update! 
I'm sorry that I'm not well versed in how node and packaging around it works, but if I just want to start a commuter server after cloning the nteract repo on Mac OS (as I'm sure other people will want to as well) would you mind kindly telling me what sequence of commands I would need to run? 

Best,
Keita Kurita

BenRussert:
@keitakurita you might be looking for [commuter](https://github.com/nteract/commuter), it was recently moved out of the monorepo here.

keitakurita:
@BenRussert  Oh I see! Sorry for the confusion! Thanks!

willingc:
Thanks @BenRussert. @keitakurita here's the PR that moved stuff: https://github.com/nteract/nteract/pull/3956

keitakurita:
@willingc I see, thanks!

PR #4085, opened 2019-01-19
add refactor message for commuter 

willingc:
Addresses #4078


BenRussert:
We could link to https://github.com/nteract/commuter instead. 

`commuter` is the app for exploring/viewing static notebooks from disk, while `commutable` is a utility package that takes a json representation of a notebook and makes various notebooks operations available. `insertCellAfter`, `createMarkdownCell`, `stringifyNotebook`, etc.

Issue #4019, opened 2019-01-11
Data Explorer: rename and refactor


stormpython:
**Is your feature request related to a problem? Please describe.**
Create and refactor the `Data Explorer` package.

**Describe the solution you'd like**
The `data-transform` should be renamed to `data-explorer`. There are some performance issues with the Typescript and Styled Components transition that need to be fixed. As well as some general code maintenance and cleanup.

[WIP] The following issues need to be created:
- [x] **(Merged)** [PR #4065] Rename `transform-dataresource` to `data-explorer`.
- [x] **(Merged)** [PR #4091] Fix styled components/CSS perf issues.
- [x] [PR #4113] General CSS cleanup

stormpython:
@BenRussert @emeeks feel free to add to the list above. This can be the main issue around the data explorer refactor and we can create sub issues from here. I have more items listed, but these were the top 3 issues to get this part done quickly and we can do more general refactoring later.

BenRussert:
The perf issue is definitely coming from `@nteract/styled-blueprintjsx`, but I have no idea how to fix it without switching to a css loader.

I put together a create-react-app prototyping package that renders the dx component with the iris data set. It is hosted here if you want to check it out: https://data-explorer-web.firebaseapp.com/. If you check devtools, I passed a logger to `onMetadataChange`  so we can see the changes in the console.

The project relies on a monorepo branch where I removed `blueprint-styledjsx` to make the app responsive. [Here is a link to the project source ](https://github.com/benrussert/data-explorer-web).

stormpython:
All relevant PRs have been merged, so closing.

PR #4091, opened 2019-01-21
Remove blueprint styled-component from data-explorer

BenRussert:
Based on the discussion at today's weekly meeting, and #4019 

This would require docs to explain how to load blueprint stylesheets with a css loader (or other method) [Here is an example using the component from a project bootstrapped with create-react-app](https://github.com/BenRussert/data-explorer-web/blob/fe0ec69024c982f50bba31fd0b2f820d8129037b/src/App.tsx#L5-L9)

Here are some more [docs on this from blueprint](https://blueprintjs.com/docs/#blueprint.quick-start)

### DIFF
--- a/packages/data-explorer/package.json
+++ b/packages/data-explorer/package.json
@@ -23,7 +23,6 @@
     "@blueprintjs/core": "^3.7.0",
     "@blueprintjs/select": "^3.2.0",
     "@nteract/octicons": "^1.0.4-alpha.0",
-    "@nteract/styled-blueprintjsx": "^1.1.5-alpha.0",
     "d3-collection": "^1.0.7",
     "d3-scale": "^2.1.2",
     "d3-shape": "^1.2.2",
--- a/packages/data-explorer/src/VizControls.tsx
+++ b/packages/data-explorer/src/VizControls.tsx
@@ -1,6 +1,5 @@
 import { Button, Code, IconName, MenuItem } from "@blueprintjs/core";
 import { Select } from "@blueprintjs/select";
-import { BlueprintCSS, BlueprintSelectCSS } from "@nteract/styled-blueprintjsx";
 import * as React from "react";
 
 import { StyledButtonGroup } from "./components/button-group";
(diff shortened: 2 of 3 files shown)


rgbkrk:
I suppose we just have to stick these into the very root of both jupyter extension and desktop next (or use a webpack css loader like you said).

BenRussert:
I think that's worth a shot. I think it would make the dx component more flexible if we uncouple the style component and provide as a separate option. We could outline potential style solutions in the docs.

rgbkrk:
That's cool with me. Would you want to make a follow up PR that provides the style for data explorer to use to both jupyter extension and desktop?

BenRussert:
> Would you want to make a follow up PR that provides the style for data explorer to use to both jupyter extension and desktop?

Should be able to do this week, hopefully tonight.

rgbkrk:
Ok I'm happy to do the follow on work for this to move the blueprint styles into the right spot for the apps. We just won't publish until we've got the README and other bits situated for #4019. Sound like a plan?

rgbkrk:
Alright, I've got a branch in the works that I'll submit in a few minutes here after testing how it goes.

rgbkrk:
Yeeha. Same problem for codemirror styles.

I can't believe how much better `styled-jsx` was at handling injected global styles.

rgbkrk:
😢 Components with global style we now have to extract out: 

* All global codemirror styles for the editor
* `connected-components`' menu
* `data-explorer`'s `react-table`
* `presentational-components`' `header-editor`*  
* `transform-geojson`'s `leafletCSS`

We can export these from the packages, but people _should_ move them up to the app level. I'll open my PR now.

rgbkrk:
I probably need to update the way css variables for themes are handled as well.

BenRussert:
> I probably need to update the way css variables for themes are handled as well.

The variables used in dx rely on declarations from other nteract components. That's why I loaded the `nteract.css` file in the web app above.


> I can't believe how much better styled-jsx was at handling injected global styles.

Agreed. We should probably open an issue if there isn't one.

rgbkrk:
https://github.com/nteract/nteract/pull/4093 is the first pass, handling both data explorer and the editor. I'll open some issues after lunch.

PR #4065, opened 2019-01-18
Renames @nteract/transform-dataresource to @nteract/data-explorer

stormpython:
This replaces #4020. It turns out that much of the `transform-dataresource` package is related to `data-explorer` code. Based on a meeting I had today with @emeeks. So, it's best to simply rename the package and then do a refactor of the package in another PR.

PR #4092, opened 2019-01-21
Data explorer rename

stormpython:
Renaming all instances of `DataTransformResource` to `DataExplorer`. This is a follow up to PR #4065.


rgbkrk (approved):
Beautiful. 😄

PR #4083, opened 2019-01-19
Typedef

Shinigami92:
~~I'm going to do this PR first to get the automation tests done~~
~~So I can check if I have broken something~~

You can also follow my progress and see what I do
And if you want you can add comments to the code with the review tool

PR #4093, opened 2019-01-21
Move global styles to root for perf

rgbkrk:
This moves a bunch of the `createGlobalStyle` styles up to the root. Apparently re-rendering a global style is _not_ a no-op and ends up causing a significant lag. Since this basically made the app entirely non-useful, I've refactored a few of these to require the styles to be placed at the root.

### DIFF
--- a/applications/desktop/src/notebook/index.tsx
+++ b/applications/desktop/src/notebook/index.tsx
@@ -8,6 +8,8 @@ import { GlobalCSSVariables } from "@nteract/presentational-components";
 import { BlueprintCSS, BlueprintSelectCSS } from "@nteract/styled-blueprintjsx";
 import { createGlobalStyle } from "styled-components";
 
+import { CodeMirrorCSS, ShowHintCSS } from "@nteract/editor";
+
 import DataExplorer from "@nteract/data-explorer";
 import { WidgetDisplay } from "@nteract/jupyter-widgets";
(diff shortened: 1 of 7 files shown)


rgbkrk:
The app is now back to a normal sane state. Wow.

stormpython:
I get the following error in the console when I type anything into a cell.

screen shot 2019-01-21 at 1 20 14 pm

stormpython:
Outside of the issue above, it looks like the styles are only being rendered once according to the profiler output.

rgbkrk:
Is that in the jupyter extension? Admittedly, I've only tried desktop so far.

stormpython:
Yes, thats in the Jupyter extension.

rgbkrk:
Booting it up now, thanks for pointing it out to me.

stormpython:
Same thing happens on desktop as well.

rgbkrk:
Weird. Jupyter extension is working perfectly for me.

rgbkrk:
Have you run `yarn && yarn build:packages`? I did change up the tsconfig.

rgbkrk:
Oh wait, I see it sometimes. I know where that's happening and it's unrelated to the CSS. I think that's likely from #4083

stormpython:
Alright, then this LGTM!

PR #4109, opened 2019-01-23
Add shutdownMutation and GraphQL tests

captainsafia:
Trying to sort out an issue with an async process in the tests not finishing.

I think this might have something to do with why I wasn't able to get the shutdownEpic working properly.


captainsafia:
Nevermind. The problem was the main that was being invoked in the Apollo server code. I added logic to only call it if the file is run, not imported.

packages/kernel-relay/__tests__/index.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { gql } from "apollo-server";
+import { createTestClient } from "apollo-server-testing";
+import { server } from "../src";
+
```
rgbkrk, line 8: Should we be providing some jest mocks for kernelspecs and launching, or just assume we want full integration tests here?

captainsafia: I was going for full integration tests. I'm still investigating the best options for unit testing our graphql setup. I think it makes sense to add mocks in there.

packages/kernel-relay/__tests__/index.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    const LIST_KERNELSPECS = gql`
+      query GetKernels {
+        listKernelSpecs {
+          name
+        }
+      }
+    `;
+    const response = await query({ query: LIST_KERNELSPECS });
+    expect(response).not.toBeNull();
+    expect(response.data.listKernelSpecs.length).toBeGreaterThan(0);
```
rgbkrk: We could use jest snapshots here! Then, coupled with the mock for the raw call this would come out consistently. We'd assume our `fs-kernels` does the heavyweight lifting and real testing there.

packages/kernel-relay/src/index.ts
```diff
@@ -150,4 +169,6 @@ process.on("exit", () => {
   });
 });
 
-main();
+if (require.main === module) {
+  main();
```
rgbkrk: I'm assuming this is for testing? Can you leave a comment about this?

packages/kernel-relay/src/index.ts
```diff
@@ -115,26 +133,27 @@ const resolvers = {
...
+        endpoint: "",
+        query: ``
+      }
+    ]*/
+  },
+  resolvers: resolvers as any,
+  typeDefs
+});
+
 async function main() {
```
rgbkrk, line 3: One of the reasons I put the server inside the main is for if/when we need to create other resources asynchronously.

packages/kernel-relay/src/index.ts
captainsafia: Oh yeah. I posted the comment above but I can elaborate on this.

> The problem was the main that was being invoked in the Apollo server code. I added logic to only call it if the file is run, not imported.

Basically, we don't want to run `main` if the file is being imported, as is the case when we use it in tests.

packages/kernel-relay/src/index.ts
rgbkrk: I think, instead of exporting `server` that we export a function called `createServer` for the inevitable day when we have to pass options. Otherwise we're exposing ourselves to having the same singleton class in every use of this `server`.

packages/kernel-relay/src/index.ts
rgbkrk: Should we have a `cli.js` and a `server.js` file?

captainsafia: Yeah. I think I know how to refactor our file setup based on all the input you provided.

rgbkrk (approved):
Solid ways of making this testable here.

PR #4110, opened 2019-01-24
Refactor: reducers minimize getIn

Shinigami92:
This improved some typeings

### DIFF
--- a/packages/reducers/src/comms.ts
+++ b/packages/reducers/src/comms.ts
@@ -36,7 +36,7 @@ function processCommMessage(
 ): CommsRecord {
   const { data, comm_id } = action;
 
-  const commInfo = state.getIn(["info", comm_id]);
+  const commInfo = state.info.get(comm_id);
   if (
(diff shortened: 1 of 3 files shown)


Shinigami92:
What .-.
`notebook` is not defined everywhere?
But how did it work before?

rgbkrk:
> notebook is not defined everywhere?
> But how did it work before?

`getIn` will return `undefined` if that path doesn't exist at any point in that array of paths.

rgbkrk:
Is this all the reducer tests for document failing or just certain ones that aren't fully set up properly?

Shinigami92:
Ok, then I will put this branch in a trash can

rgbkrk:
Might be worth just doing one of these and seeing if the tests pass locally (or more until you find one failing), in case it's the test that needs updating.

Shinigami92:
I do not know if it's worth it
We should maybe typedef the lhs instead

PR #4111, opened 2019-01-24
types: export all interfaces

Shinigami92:
Export all interfaces
So, if `PayloadMessage` is used somewhere, you have the option of using `PagePayloadMessage`, for example


rgbkrk (approved):

rgbkrk:
Smart. These seem like ones we'd keep consistent, API wise.

PR #4112, opened 2019-01-24
only import some things from immutable

Shinigami92:
This should improve tree-shaking

### DIFF
--- a/packages/reducers/src/config.ts
+++ b/packages/reducers/src/config.ts
@@ -1,4 +1,4 @@
-import { Map as ImmutableMap } from "immutable";
+import { Map } from "immutable";
 
 import { MergeConfigAction, SetConfigAction } from "@nteract/actions";
 import { ConfigState } from "@nteract/types";
@@ -19,7 +19,7 @@ export function mergeConfig(state: ConfigState, action: MergeConfigAction) {
 }
 
 export default function handleConfig(
-  state: ConfigState = ImmutableMap(),
+  state: ConfigState = Map(),
   action: ConfigAction
 ) {
   switch (action.type) {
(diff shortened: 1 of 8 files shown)


packages/reducers/src/config.ts
```diff
@@ -1,4 +1,4 @@
-import { Map as ImmutableMap } from "immutable";
```
rgbkrk, line 1: The reason we were using `ImmutableMap` as the name was to make sure we weren't shadowing the ES6 `Map`.

packages/reducers/src/core/entities/contents/index.ts
```diff
@@ -162,9 +162,9 @@ const byRef = (
               model: makeDocumentRecord({
                 notebook: immutableNotebook,
                 savedNotebook: immutableNotebook,
-                transient: Immutable.Map({
-                  keyPathsForDisplays: Immutable.Map(),
-                  cellMap: Immutable.Map()
+                transient: Map({
```
rgbkrk, line 165: Not that you have to here, but `Map` takes a generic. Perhaps typescript is autodetecting this one properly.

packages/reducers/src/config.ts
Shinigami92: Because I imported `import { Map } from "immutable";` typescript takes this

packages/reducers/src/core/entities/contents/index.ts
Shinigami92: Typescript is very smart :)
But I didnt changed the code here in any way

packages/reducers/src/config.ts
rgbkrk: Yeah, it's less of a typescript issue and more of a human-issue. Since we get random contributors, it makes it explicit which is which within here.

rgbkrk (approved):

PR #4113, opened 2019-01-24
Clean and move CSS to styled-component style

emeeks:
* Move things to styled-components

Minor bits:
* Clean up some component->pureComponent based on lint complaints
* Add reset to select on metric/dim selectors
* Disable scatterplot annotations for now (wait to come back to this with a better strategy and also some weird edge bugginess needs to get ironed out)


rgbkrk (approved):
This is looking nice and cleaned up. The new styled elements give me a bit more insight into the various UI components here. Thanks for all the new vocab and naming!

stormpython (commented):
These changes look good to me! This PR is failing mainly due to Typescript issues. Hopefully there aren't too many to clean up. But I think once they are fixed, this should be good to merge. @rgbkrk thoughts?

rgbkrk:
Yeah looks like it's all the `setColor` usage that needs to be cleaned up for types. If you want to check these locally @emeeks, you can run `npx tsc -b packages/data-explorer`

stormpython:
I am seeing this error in the console when using the data explorer.

screen shot 2019-01-25 at 4 42 03 pm

There is also some weird css issues with the panel below the chart. Is this expected or was this caused by the style component changes?

emeeks:
@stormpython What CSS issues? The label warning isn't a result of styled-component as far as I know, though maybe I didn't key one of the tooltip components, I'll dig into it.

stormpython:
@emeeks nevermind, the css issues were an artifact of something with my browser. Disregard the last comment.

rgbkrk:
Since it looks like its only Azure failing and Circle runs the comprehensive tests, I think I'll merge this after doing a quick overview.

packages/data-explorer/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -372,124 +344,6 @@ exports[`DataExplorerNoMetadata creates a data explorer with metadata 1`] = `
               },
             }
           }
-          annotations={
```
rgbkrk, line 375: Is it ok that these annotations aren't in the snapshots anymore (should they be here)?

packages/data-explorer/src/types.ts
```diff
@@ -13,7 +13,7 @@ export interface ChartOptions {
   networkType: NetworkType;
   hierarchyType: HierarchyType;
   primaryKey: string[];
-  setColor: (color: string) => void;
+  setColor: (color: string[]) => void;
```
rgbkrk, line 16: Wow that's quite a change in definitions. I'll hope for correct here now. 😅

PR #4114, opened 2019-01-24
[perf] take codemirror options as direct props

rgbkrk:
Reopening of #4032 with a fresh look.

This is now a semi-clean wrapper on top of Codemirror where you can do this:

```jsx
<Editor tabSize={8} />
```

directly instead of passing in an options object for the codemirror options. _However_, not every option is configurable because of the way we use the codemirror classes and set keys for notebook usage. For instance, `theme`, `hintOptions`, and `extraKeys` cannot be overridden since we need to be able to bind them to internal methods.

### DIFF
--- a/packages/editor/__tests__/editor.spec.tsx
+++ b/packages/editor/__tests__/editor.spec.tsx
@@ -141,11 +141,11 @@ describe("editor.hint CodeMirror callback", () => {
 
 describe("Editor", () => {
   it("handles cursor blinkery changes", () => {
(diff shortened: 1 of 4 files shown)


rgbkrk:
Ok this is ready for review. @captainsafia or @stormpython -- can you check this out?

stormpython:
I should be able to review this in a couple hours, latest being sometime this evening (PST).

stormpython (approved):
LGTM!

PR #4116, opened 2019-01-25
clean up unused properties and generated objects

rgbkrk:
Finding more areas where performance can be improved without large refactors. In this edition, I got rid of

* `metadata` that was being passed through as a `toJS`ed  object from the `mapStateToProps` function for our cells
* empty Immutable lists and sets are created once before usage (whenever needed)


lgeiger (approved):
Nice cleanup :+1:

PR #4119, opened 2019-01-25
catch errors on output failures

rgbkrk:
I've noticed on `master` that data explorer has a bug or two causing a breakdown in the component tree. I've set up some `componentDidCatch` logic in our `<Output />` picker component so that it doesn't kill the app and so we can show information about the outputs as necessary.

kapture 2019-01-25 at 8 37 24

This got missed in the switch over to a connected rich media component. Now we're back to disallowing outputs from breaking the app.


rgbkrk:
cc @emeeks @stormpython -- incidentally found a bug in data explorer ^^

captainsafia (approved):
What a heart-healthy PR!

rgbkrk:
This helps me sleep at night.

PR #4122, opened 2019-01-26
Editable header publish button

stormpython:
Adds the `EditableHeader` component to the `Nav`. Also adds the publish button which will allow users to publish their notebooks to a specified endpoint. This PR requires work on the `book store` API to complete the publisher experience feature.


rgbkrk:
From in person convo:

* We should validate bookstore on first render (from jupyter server) of nteract's index.html (this can then be memoized)
* This PR should be split apart into:
  * Editable header should be able to work without publishing (just a nice way to set notebook metadata)
  * Publishing and bookstore enablement checking we can go after in a separate PR

stormpython:
Ok, @captainsafia 😌 . I have this in working order, but I just want to say, I am sorry for giving you such a massive PR. I completely underestimated the amount of work needed for this PR, and I had a lot to learn about `Redux` and `Immutable.js`.

With all that said, the build for desktop is failing from what I believe is an issue with `AppState` since I modified it in `nteract_on_jupyter`. But I am lost as to how to narrow the discrepancies down. Would you mind helping me debug when you have a chance?

Other than that, I think the PR is ready for review. The functionality is there and working. In order to merge this one though, we need a couple other pull requests, including one in `Bookstore`.

packages/connected-components/src/header-editor/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+  private onAuthorsRemove = (t: any) => (
+    evt: React.MouseEvent<HTMLButtonElement>,
+    props: ITagProps
+  ): void => {
+    const { editable, headerData, onChange } = this.props;
+
+    if (editable === true && onChange) {
+      onChange({
+        ...this.props.headerData,
+        authors: Array.from(headerData!.authors).filter(p => {
```
captainsafia: I notice that there's a few of these non-null assertions whenever the header data is referenced. Is it something to do with the way the type is defined? Could we possibly restructure the code to avoid having the `!` whenever we access properties inside `headerData`?

packages/connected-components/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "outDir": "lib",
     "rootDir": "src"
   },
-  "include": ["src"],
+  "include": ["src", "src/header-editor/index.tsx"],
```
captainsafia: Hm. We shouldn't need to add this here. Was there a problem with the compilation?

packages/reducers/src/core/entities/hosts.ts
```diff
@@ -12,6 +13,12 @@ import {
 const byRef = (state = Map(), action: Action) => {
   let typedAction;
   switch (action.type) {
+    case actions.PUBLISH_TO_BOOKSTORE:
```
captainsafia, line 19: Hm. Is there no client-side state associated with publishing a notebook to bookstore (like a last published time) that would need to be stored here?

packages/connected-components/src/header-editor/index.tsx
```diff
@@ -0,0 +1,359 @@
...
+import {
+  AuthorTag,
+  Container,
+  EditableAuthorTag,
+  EditableTag,
+  MarginContainer
+} from "./styled";
+
+// Type Definitions
+export interface AuthorObject {
```
captainsafia: I see this type definition is also defined in the types package. Could we import it from there instead of redefining it here?

captainsafia (commented):
Looks great! I've posted some comments to follow-up on inline.

Edit: Messaging you on Slack about pairing on the AppState stuff.

packages/reducers/src/core/entities/hosts.ts
stormpython: You are right. We should be storing a last published time. I'll write an issue an add it in a separate PR.

packages/connected-components/src/header-editor/index.tsx
stormpython: I'll remove this one.

packages/connected-components/tsconfig.json
stormpython: I honestly don't remember. I was probably having an issue and tried many different things at once and forgot to revert back.

packages/connected-components/src/header-editor/index.tsx
stormpython: Absolutely, we can!

packages/reducers/src/core/entities/hosts.ts
stormpython: Ok, I am going to pull adding this functionality out of this Project, and open a issue to do it at another time.

stormpython:
Before this can be merged, the `object` that gets sent to the `Bookstore` published endpoint needs to be refactored to match the [spec](https://jupyter-notebook.readthedocs.io/en/stable/extending/contents.html).

captainsafia:
> Before this can be merged, the object that gets sent to the Bookstore published endpoint needs to be refactored to match the spec.

Do you still need to make this change @stormpython? Or is this ready for review.

stormpython:
I do. I didn't get a chance to finish it today.

stormpython:
@captainsafia this is ready for review again.

packages/reducers/src/core/index.ts
```diff
@@ -41,11 +41,16 @@ const currentKernelspecsRef = (state = "", action: Action) => {
   }
 };
 
+// const bookstore = (state = {}, action: Action) => {
+//   return state;
+// };
+
```
captainsafia: Might want to delete this code all together instead of commenting it out.

captainsafia (approved):
Looks good! I posted one comment about cleaning up some comments (no pun intended).

mpacer:
Is this in a state where I can test against the current version of bookstore?

stormpython:
@mpacer this is ready for testing with `Bookstore`!

stormpython:
Checklist for QA:

- [ ] Publish button is hidden and nteract works when bookstore is not installed
- [ ] Publish button works and publishes to bookstore
- [ ] ...

mpacer:
@stormpython and I worked through this with a bookstore-enabled server to verify that it all works together (without hand-tuning)!

mpacer (approved):
As said above, this works as far as I can tell.

In the long run it'd be nice to have an integration test for this in our CI setup.

willingc:
@stormpython Is this ready to merge? If so, just press the green button :wink:

PR #4123, opened 2019-01-26
mock fs-kernels

rgbkrk:
In order to test `kernel-relay` properly, we need to mock `@nteract/fs-kernels`. Right now on master if you update snapshots in general you'll erroneously encode your own list of kernelspecs. Here's some basic mocking for some of the functions in `fs-kernels`.

Due to our monorepo setup, I couldn't write `__mocks__/@nteract/fs-kernels` next to `kernel-relay` because the package is resolved from the root.


captainsafia (approved):
Cool! I think having the mock within the test file is alright.

Issue #4104, opened 2019-01-23
Keybindings: Add file shortcuts


stormpython:
**Is your feature request related to a problem? Please describe.**
Add `File` shortcuts to nteract.

- [x]  save => `ctrl+s`
- [x] [#4126] save as => `ctrl+shift+s`
~- [ ] open => `ctrl+o`~

This issue is a child of #4101.

**Describe the solution you'd like**
- [x] ~*Before this can be solved, #4054 needs to be merged.*~

For information on how to add shortcuts to nteract, please see [4102](https://github.com/nteract/nteract/issues/4102#issuecomment-457790015). See the list of [nteract shortcut-key](https://docs.nteract.io/#/desktop/shortcut-keys).

stormpython:
I am going to create another issue for the `open` keybinding since more work is needed with this one.

glitch401:
Hi ! can I take up this issue?

stormpython:
@indranilbiswas7, thanks for volunteering, but this particular issue has already been implemented. But there are other issues within Keybinding that can be picked up here: #4101.

PR #4126, opened 2019-01-28
Keybindings: Adds `save as`

stormpython:
Adds `save as` keybindings to nteract.

`ctrl+shift+s` mapped to redux `SAVE` action type.

This PR is part of issue #4104.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/index.tsx
@@ -26,6 +26,7 @@ interface IContentsProps {
   contentRef: ContentRef;
   displayName: string;
   error?: object | null;
+  filepath: string | undefined;
   lastSavedStatement: string;
   loading: boolean;
   mimetype?: string | null;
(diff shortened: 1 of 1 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/contents/index.tsx
```diff
@@ -146,7 +148,9 @@ const makeMapStateToProps: any = (
 
 const mapDispatchToProps = (
   dispatch: Dispatch,
-  initialProps: { contentRef: ContentRef }
+  initialProps: {
+    contentRef: ContentRef;
+  }
```
rgbkrk, line 165: I probably should have noticed this before, but this should be in a `makeMapDispatchToProps` since the content ref should remain the same and we shouldn't re-create these on every render.

rgbkrk: Feel free to come back to this in a new PR (unless you're already making the change).

rgbkrk (approved):

rgbkrk:
There's a merge conflict on this one now.

stormpython:
Ok, let me take a look.

stormpython:
@rgbkrk, the conflict should be resolved now.

rgbkrk (approved):

PR #4136, opened 2019-01-29
Remove icon not supported warning and add todo

BenRussert:
This line generates warnings the user doesn't need to see in production. I thought I'd add a todo though, since I'm not 100% sure if we are handling properly.

### DIFF
--- a/packages/data-explorer/src/VizControls.tsx
+++ b/packages/data-explorer/src/VizControls.tsx
@@ -118,7 +118,8 @@ const getIcon = (title: string) => {
   if (title === "X" || title === "Y" || title === "Size" || title === "Color") {
     return iconHash[title];
   } else {
-    console.warn("Icon title not supported");
+    // TODO: Verify if we are handling icon title properly
+    // console.warn("Icon title not supported");
     return title as IconName;
   }
 };


packages/data-explorer/src/VizControls.tsx
```diff
@@ -118,7 +118,8 @@ const getIcon = (title: string) => {
   if (title === "X" || title === "Y" || title === "Size" || title === "Color") {
     return iconHash[title];
   } else {
-    console.warn("Icon title not supported");
+    // TODO: Verify if we are handling icon title properly
```
rgbkrk, line 121: Is title only ever going to be the constants above or is it really a string?

rgbkrk: I'd probably leave the TODO out or reference dx since todo bot will immediately make an issue after we merge it.

rgbkrk:
Guh, I had left my two comments in a draft review instead of submitting as comments. 😞 

Sorry to hold this one up @BenRussert

BenRussert (commented):
I wasn't sure if this was was worth an issue, but thought an issue would be better than a console warning if we just didn't want to forget about it.

rgbkrk:
Alright, let's go with it then.

PR #4137, opened 2019-01-29
switch jext's tsconfig back to composite: true

rgbkrk:
(it's inherited from our base config)

I don't know if there's much I can optimize on the webpack build for now due to https://github.com/TypeStrong/ts-loader/issues/851


rgbkrk:
Oh no, we have a phantom Azure now.

PR #4145, opened 2019-01-30
bring back positive lookahead regex for newline splitting

rgbkrk:
A regression worked its way into `remultiline` during the last weeks' refactor. All the `?` got deleted, making this not a positive lookahead regex with a non-matching group...

cc @stormpython @captainsafia


rgbkrk:
Safia is adding some tests to this PR now. 😄

BenRussert:
:shipit:

PR #4147, opened 2019-01-30
include more tests for remultiline

rgbkrk:
As @captainsafia and I noticed... there are no tests for commutable, even though there [_used to be_](https://github.com/nteract/commutable/tree/ae3edf1dfd7fbb344585667b7361984d70949926/test).

Since this is much of the backbone of our entire notebook app state, I'm going to make at least one new test a day for `commutable`. We test some of this via our reducers, but a lot of little functions are missed for the trees as a result.


stormpython (approved):
LGTM!

PR #4159, opened 2019-01-30
Bring back chunks for jext's ts-loader

rgbkrk:
This doesn't _fully_ make our setup faster at reloads, though it does bring back chunks

screen shot 2019-01-30 at 3 41 41 pm

If we want this to load smoother I think we need to change how our initial app loads (relying on `import()` on app load), investigate the size of our bundle, and likely cleanup interdependence amongst some of the big packages (I have a feeling `@nteract/types` is a bit of a bug bear in this area).


applications/jupyter-extension/nteract_on_jupyter/tsconfig.json
```diff
@@ -0,0 +1,35 @@
+{
```
rgbkrk, line 1: Guh, I did `git mv`, but apparently I changed too much for this to be counted as the original version of this file. If anyone wants to see this history together, let me know and I'll re-do this as new commits.

rgbkrk:
Data explorer relying on `commutable` and `types` is a big issue for allowing chunks. I just went through ripping `commutable` and `types` out of its dependencies (even though it's only used for types) and the bundle for data explorer went from a few kilobytes to MB (telling me that it's actually getting its deps co-located now). This is probably the case for other packages too.

I'm hesitant to make too many of the optimizations in one PR so I won't add them on to here.

emeeks (approved):

PR #4163, opened 2019-01-31
remove excessively protective tslint typedef rules

rgbkrk:
The `typedef` rule has been bugging me across the codebase. We shouldn't have to declare types that should be able to be inferred:

image

Now, we don't have to. We can write `const y = "something"` and it will let us leave `y` as a string there. .______.


stormpython:
👏 HOORAY!!!

PR #4167, opened 2019-02-01
Asset path fixes for jext

rgbkrk:
Thanks to @jameskoch for finding out that our chunk assets weren't being loaded properly. This happened because `__webpack_public_path__` must be set synchronously on start. To make it so this is less likely to happen I broke things up into `config.tsx` for the config reading (all synchronous) and moved our initialization to a new file called `bootstrap.tsx`.


stormpython (approved):
LGTM!

rgbkrk:
Ok, I'll publish again. I've also made this more testable so that we can test the ancillary stuff around it now.

PR #4168, opened 2019-02-01
Minor adjustments

### LINKED ISSUE #4129: Data explorer table filters
The data-explorer table filters are not working as expected:

dx-filter-bug

Here is a snippet to reproduce like in the gif:
```python
import pandas as pd
pd.options.display.html.table_schema = True
pd.options.display.max_rows = None
df = pd.read_csv('https://raw.githubusercontent.com/uiuc-cse/data-fa14/gh-pages/data/iris.csv')
```

emeeks:
This should resolve #4129

* Filter value seems to be passing a string now instead of a number (for the numerical filter) so cast to number
* Return the results of the filtering
* Update types to reflect that the data object will be passing numbers to the  number filter (even though the filter is passing a string)

### DIFF
--- a/packages/data-explorer/src/charts/grid.tsx
+++ b/packages/data-explorer/src/charts/grid.tsx
@@ -38,7 +38,6 @@ const GridWrapper = styled.div`
 const NumberFilter = (props: NumberFilterProps) => {
   const { filterState, filterName, updateFunction, onChange } = props;
   const mode = filterState[filterName] || "=";
-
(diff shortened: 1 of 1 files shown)


stormpython (approved):
LGTM!

BenRussert (approved):
Makes a lot of sense, thanks!

rgbkrk:
Cool, let's do another release later today.

Issue #4169, opened 2019-02-01
[jext] Webpack Dev Server not loading chunks properly


rgbkrk:
The first request for `app.js` is ok, as its hitting the webpack dev server:

<img width="1031" alt="screen shot 2019-01-31 at 7 42 56 pm" src="https://user-images.githubusercontent.com/836375/52101435-85186e80-2590-11e9-9bed-22092a02b887.png">

All requests for chunks however are resolving using the `__webpack_public_path__` when they shouldn't be in dev:

<img width="1205" alt="screen shot 2019-01-31 at 7 44 21 pm" src="https://user-images.githubusercontent.com/836375/52101484-b4c77680-2590-11e9-8f17-3f8cd6c7146c.png">

rgbkrk:
Walking my way through webpack loading and I see the chunk URLs are formed by webpack like so:

<img width="396" alt="screen shot 2019-01-31 at 9 12 23 pm" src="https://user-images.githubusercontent.com/836375/52103943-fd852c80-259c-11e9-8b3e-6984572dae8d.png">

PR #4170, opened 2019-02-01
[WIP] allow chunks with webpack dev server + nteract --dev

rgbkrk:
Attempting #4169.

* [x] run `prettier` and `black` over jupyter extension to keep changes minimal
* [ ] When in dev mode, don't overwrite `__webpack_public_path__`

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/__init__.py
+++ b/applications/jupyter-extension/nteract_on_jupyter/__init__.py
@@ -7,7 +7,6 @@ from .extension import load_jupyter_server_extension
 
 EXT_NAME = "nteract_on_jupyter"
 
+
 def _jupyter_server_extension_paths():
-    return [{
-        "module": EXT_NAME
-}]
+    return [{"module": EXT_NAME}]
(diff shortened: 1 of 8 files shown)


rgbkrk:
Even though I'm running `jupyter nteract --dev` it looks like `load_jupyter_server_extension` is still running. This is probably because  it _is_ also enabled as a server extension on my machine so it probably gets loaded twice :laughing:

As a result, this means `config.dev_mode` gets set to `False` unconditionally here: 
https://github.com/nteract/nteract/blob/3e9cc9ef09b00cd83c41d490f7fa250e66b3da1a/applications/jupyter-extension/nteract_on_jupyter/extension.py#L35 Not sure what that was for, so this section of code is probably ready for some cleanup.

All that being said, it looks like even if I was able to get `devMode` to propagate through, it doesn't matter since it's not the `__webpack_public_path__` that is mucking up the chunk loading. Seems more like the chunk plugin isn't compatible with the webpack dev server. I'll keep investigating.

rgbkrk:
My little bits of digging are showing me that we likely shouldn't use chunk hashes in development.

rgbkrk:
Closing in favor of #4171

PR #4171, opened 2019-02-01
Cleanup asset paths

### LINKED ISSUE #4169: [jext] Webpack Dev Server not loading chunks properly

rgbkrk:
Fixes #4169, bringing `jupyter nteract --dev` back to working.

Turns out that we were kind of _over-configuring_ the webpack setup here. When in dev mode, setting `publicPath` to a default asset path was causing `/nteract/static/dist` to be set for `__webpack_require__.p` (computed by webpack later on) regardless of `__webpack_public_path__` being set to what we needed. By setting the webpack dev server to host its `publicPath` at `/nteract/static/dist` explicitly while _not_ setting `publicPath` for our main `output`, we've finally got the right setup.

If I recall correctly, I had set this default asset path so that the app would work without the setting of `__webpack_public_path__` in the simple case of no `--NotebookApp.base_url`. That ended up causing more issue than it was worth.

I've also verified this works when the dev server is not running with `yarn build:asap` followed by a plain `jupyter nteract`.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
@@ -14,13 +14,12 @@ if (!rootEl || !dataEl) {
 } else {
   const config: JupyterConfigData = readConfig(rootEl, dataEl);
 
+  const webpackPublicPath = urljoin(config.assetUrl, "nteract/static/dist/");
   // Allow chunks from webpack to load from their built location
   // NOTE: This _must_ run synchronously before webpack tries to load other
-  // chunks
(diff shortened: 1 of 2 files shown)


rgbkrk:
Alright, that's my last commit for now. This puts `master` back on track and we can further optimize our dev setup later.

PR #4178, opened 2019-02-01
Blueprint as css for desktop

rgbkrk:
This turns the style loader on for desktop and allows us to load the base CSS directly instead of these global styles.

### DIFF
--- a/applications/desktop/src/notebook/index.tsx
+++ b/applications/desktop/src/notebook/index.tsx
@@ -5,8 +5,10 @@
 import * as MathJax from "@nteract/mathjax";
 
 import { GlobalCSSVariables } from "@nteract/presentational-components";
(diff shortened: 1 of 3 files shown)


lgeiger (approved):
If desktop loads everything in dev and prod, this is far more painless then I expected 💯

rgbkrk:
The prod question is the big one. Let me run a build and check. Only dev is good so far.

lgeiger:
> The prod question is the big one. Let me run a build and check. Only dev is good so far.

I guess it *should* be fine 🤞

rgbkrk:
Looking good!

screen shot 2019-02-01 at 3 21 33 pm

With that, I think it's time for the great _move-all-global-css-back-into-big-css-files_ migration.

PR #4179, opened 2019-02-01
[WIP] start making a blueprint backed declarative react menu

rgbkrk:
Slightly kicking off this menu re-write so that we're all-in on blueprint for the UI components, to reduce our overall dependencies (streamlining our build, reducing load times) as well as unify our UI direction.

image

This will end up being a pretty straightforward PR to rip out our old menu with this now that I've done the start of it. A lot of indirection should get cleaned up as a result. The new setup will end up being a declarative React structure for the menu with some bound dispatch on nested keys. The old setup was 
declarative JSON structure getting re-hydrated into a React structure, along with a lot of key + constant redirection. It was not code I liked touching.

Since I want a more rapid way of developing on this, I'm going to put this PR on hold while I rip out the rest of the global style components (moving blueprint up already was a big win).

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/index.tsx
@@ -19,7 +19,9 @@ import { ConnectedDirectory } from "./directory";
 import { default as File } from "./file";
 import { ConnectedFileHeader as FileHeader, DirectoryHeader } from "./headers";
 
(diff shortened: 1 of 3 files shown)


rgbkrk:
This is looking really good now though I've run into https://github.com/palantir/blueprint/issues/3352

captainsafia:
I believe this can be closed now that @and0111's work on the menu is merged into master.

rgbkrk:
Woo!

PR #4180, opened 2019-02-02
Fix sending DX metadata to output

emeeks:
Not sure what's changed upstream but this seems to expect metadata to be a map of media types and what's actually being passed is a map from the metadata object.

This minor change makes things work but I doubt it's the right change. If anyone has any context on what's happening upstream where this has changed let me know (or fix it :D )

### DIFF
--- a/packages/notebook-app-component/src/transform-media.tsx
+++ b/packages/notebook-app-component/src/transform-media.tsx
@@ -43,7 +43,7 @@ const PureTransformMedia = (props: MappedProps & DispatchProps) => {
     <Media
       {...mediaActions}
       data={output.data[mediaType]}
-      metadata={output.metadata.get(mediaType)}
+      metadata={output.metadata.get(mediaType) || output.metadata.toJS()}
     />
   );
 };


packages/notebook-app-component/src/transform-media.tsx
```diff
@@ -43,7 +43,7 @@ const PureTransformMedia = (props: MappedProps & DispatchProps) => {
     <Media
       {...mediaActions}
       data={output.data[mediaType]}
-      metadata={output.metadata.get(mediaType)}
+      metadata={output.metadata.get(mediaType) || output.metadata.toJS()}
```
rgbkrk, line 46: Ah, looks like we're passing in metadata as an `Immutable.Map`, which is great for performance and not very good for components to deal with.

emeeks: No it's not that. It's that the map expected is a map of mediaTypes, when the map that's passed is just a toMap version of the metadata object that's sent back from Data Explorer. My assumption is that the corresponding entry in the map is supposed to be a plain js object, or if not is something that we should convert at this point.

stormpython (approved):
You found it! :)

emeeks:
Okay I took a look through the actions and code around it and I don't see any reference to `mediaType` anywhere in `UpdateOutputMetadata` or related functions, so I think this was either aspirational or mistaken, in which case I think my fix should be fine to merge.

PR #4181, opened 2019-02-02
create a styles package to load from

rgbkrk:
Creates a styles package that has all our global CSS in one spot (CSS variables, overrides for codemirror). Brings time to first interactive to < 2s (better, not the best).

Usage:

```ts
import "@blueprintjs/core/lib/css/blueprint.css";
import "@blueprintjs/select/lib/css/blueprint-select.css";

// For Data Explorer

import "codemirror/lib/codemirror.css";
import "codemirror/addon/hint/show-hint.css";

import "@nteract/styles/app.css";
import "@nteract/styles/global-variables.css";
import "@nteract/styles/editor-overrides.css";
```


rgbkrk:
Ok, reloading nteract on jupyter in dev mode now takes < 2s (better, but still a little long...).

packages/styles/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@nteract/styles",
+  "version": "2.0.0",
```
rgbkrk, line 3: Note: this is 2.0.0 because we used to have an `@nteract/styles` package that we aren't using.

packages/styles/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "include": ["index.js"]
+}
```
rgbkrk, line 4: This might all be optional since we are only using this package for css. Figured it didn't hurt to put some ts boilerplate in to make it part of the project references.

rgbkrk:
Questions for me to pose -- should we delete our `createGlobalStyle` elements with switching over to pure CSS for global settings?

BenRussert (approved):
This is going to work great for the issues I had been running into with styles in apps outside the monorepo. [Like this work around for example]( https://github.com/BenRussert/data-explorer-web/blob/master/src/App.tsx#L9).

[LGMT!](https://twitter.com/jafrog/status/1091386402544279552)

rgbkrk:
To quote @benlesh

> [**l**et's **g**i**t**'e**m**](https://twitter.com/BenLesh/status/1091809838596796416)

BenRussert:
My new favorite is: "Let's Gamble Try Merging"

Issue #4160, opened 2019-01-31
Reduce bundle size for nteract web


rgbkrk:
<img width="1552" alt="screen shot 2019-01-31 at 9 46 34 am" src="https://user-images.githubusercontent.com/836375/52073754-2b3a8900-253d-11e9-96c8-5c93343ff2b3.png">

<img width="1234" alt="screen shot 2019-01-31 at 9 48 06 am" src="https://user-images.githubusercontent.com/836375/52073833-5624dd00-253d-11e9-8c5a-970b0e9a75ca.png">

A few easy wins I see:

* Move to CodeMirror CSS and Blueprint CSS as static css files
* Move all the store setup to be asynchronously loaded with `import`
* Show the nteract animation on start

I'm seeing a few other areas in the bundle analyzer that are suspect:

* What's using `refractor`?
* What needs `moment` on startup?
* Does the interdependence on `@nteract/types` and `@nteract/commutable` make it hard to break the chunks up?

## Try it yourself!

```
cd applications/jupyter-extension/nteract_on_jupyter
```

Generate statistics with:

```
npx webpack --profile --json > stats.json
```

Generate HTML page with:

```
npx webpack-bundle-analyzer stats.json --mode static
```

rgbkrk:
If you want to see the current bundle, I've also uploaded the report statically: https://report-kmhandtxkg.now.sh/

lgeiger:
> What needs `moment` on startup?

I can tackle that on Saturday (maybe even earlier).
It's only used in the header component of the jupyter extension. We can remove it and use the same approach as in the desktop status bar which uses `date-fs` and is already included in the bundle.

rgbkrk:
Note: with the latest changes on master that diagram should be updated.

lgeiger:
> What's using `refractor`

It's used by `react-syntax` highlighter which we only need for the static version of the notebook.

captainsafia:
Here's the latest data on this.

<img width="1177" alt="Screen Shot 2019-12-08 at 6 22 55 PM" src="https://user-images.githubusercontent.com/1857993/70402325-d5016680-19e7-11ea-8a5c-1cd512db20f5.png">

Our biggest culprits at the moment are Plotly.js and the Monaco editor.

captainsafia:
Closing this since we will be deprecating the Jupyter extension soon.

PR #4183, opened 2019-02-02
Remove moment.js from the webpack bundle

lgeiger:
This PR removes the last-saved display in the header of the jupyter extension. This information is already present in the status bar and in my optinion doesn't need to be duplicated in the UI.

This partially adresses #4160 by removing moment.js from the webpack bundle.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/components/last-saved.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/components/last-saved.tsx
@@ -11,74 +11,13 @@
  *
  */
 
-import { AppState, ContentRef, selectors } from "@nteract/core";
-import moment from "moment";
+import { AppState, selectors } from "@nteract/core";
+import { LastSaved } from "@nteract/directory-listing";
 import * as React from "react";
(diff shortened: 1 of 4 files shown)


rgbkrk:
We may need to take a different direction on this since this last saved component is also used for text files (python scripts, etc.). The notebook status bar is a weird thing on the web when people are used to the status at the top. I'd like to instead just switch to the same date utils in this last saved component or dynamically load moment. Speaking of which... what are we using in the directory listing? React timeago?

lgeiger:
I switched the `LastSaved` component to use `react-timeago` from `@nteract/directory-listing`. The status bar currently relies on `date-fns/distance_in_words_to_now`.

rgbkrk:
Thank you! That's the perfect approach.

rgbkrk:
I felt bad telling you to go another direction with this, I'm glad you found the right way to go about this.

lgeiger:
> I felt bad telling you to go another direction with this, I'm glad you found the right way to go about this.

No need to feel bad. I just like deleting code ;)

PR #4184, opened 2019-02-02
[WIP] Connect to existing kernels

CrystallineCat:
Trying to implement `--existing`.


rgbkrk:
This is a cool idea!

A couple things to note that might be helpful --

* with the command line argument for `existing` we don't have to detect all running kernels, only combine the path with the runtime dir on startup. I suppose this would be a new way to launch, so we'd have to do some coordination between main and renderer

* we could potentially update the menu for running kernels by continuously watching the runtime dir while launched, updating the menu entry for it. I'm not sure of the perf & cross platform impact of this though.

CrystallineCat:
Thank you for your thoughts!

> * with the command line argument for `existing` we don't have to detect all running kernels, only combine the path with the runtime dir on startup. I suppose this would be a new way to launch, so we'd have to do some coordination between main and renderer

Yeah, I guess I've got a bit of scope creep. :smirk:

Just connecting to the given kernel was the initial intention, but the immediate use case I have involves a kernel whose output I don't see, so I'd have to list the kernels in a terminal (or find the relevant process ID) anyway. I think not knowing the connection file name isn't that uncommon.

Additionally, the implementation in `jupyter_client` has far more complicated semantics:
* the argument is a fileglob
* if the argument doesn't contain `*`, it will be surround with `*`
* if nothing is supplied, it uses `kernel-*.json` 
* it looks for the file in the cwd as well as the runtime dir
* if multiple files match, it uses the most recently accessed file

For compatibility I'd like to match that implementation (with the possible exception of the last point).

[jupyter/jupyter_client · jupyter_client/connect.py#L169](https://github.com/jupyter/jupyter_client/blob/44980c13680f4e4226cf25f199ce4e4bb6e11296/jupyter_client/connect.py#L169) _(for my own reference)_

> * we could potentially update the menu for running kernels by continuously watching the runtime dir while launched, updating the menu entry for it. I'm not sure of the perf & cross platform impact of this though.

I've thought about this, but right now I don't know how to do it cleanly. I'll probably attempt it at some point in the future.

rgbkrk:
Uhoh! Looks like an interactive rebase went poorly. Let me know if you'd like help rectifying it.

captainsafia:
Hi @CrystallineCat! How are things going with this PR?

CrystallineCat:
Closing this for now.

PR #4189, opened 2019-02-05
remove kernel-relay from monorepo

rgbkrk:
[`kernel-relay`](https://github.com/nteract/kernel-relay) is now out of the monorepo and in its own.


captainsafia (approved):
This is a heart-healthy and wholesome pull request.

PR #4192, opened 2019-02-06
Store and retrieve output metadata based on mimetype

emeeks:
Metadata is stored and accessed via the mimetype as key on the output metadata


stormpython:
It seems as though there are a bunch of type errors which are preventing the build. Mostly around the `mediaType` key.

stormpython (approved):
LGTM!

stormpython:
@emeeks, can you rebase your branch and then I will merge it.

PR #4199, opened 2019-02-08
Make keybindings respect meta key

### LINKED ISSUE #4196: Keybindings: nteract doesn't respect the meta key
nteract web

**Describe the bug**
In the keybindings, `ctrl` is respected, however the `meta` key is not. Users are typically used to using the `meta` key on macOS instead of explicitly the `ctrl` key.

**To Reproduce**

1. Go to a notebook
2. Hit `⌘` + `s` to save.
3. The save as HTML dialog comes up

**Expected behavior**
The file should be saved to the notebook server

stormpython:
Keybindings will respect both `meta` and `ctrl` keys.

Closes #4196.


willingc (approved):
Looks good to me @stormpython.

PR #4201, opened 2019-02-09
Purify our Cell Toolbar

### LINKED ISSUE #4154: Cell toolbar is missing
nteract desktop, nteract web, nteract on jupyter

**Describe the bug**
The cell toolbar is missing

**To Reproduce**

1. Open a notebook file in nteract
2. Click on a cell
3. See error

**Expected behavior**
Cell toolbar should be present.

**Screenshots**
image

rgbkrk:
Fixes #4154 by taking into account `cellFocused` in addition to if source is hidden.

This also makes our little notebook app toolbar more pure. It now takes only the handlers it needs instead of hooking up to `dispatch` and computing a bunch of nested state.

### DIFF
--- a/packages/notebook-app-component/src/toolbar.tsx
+++ b/packages/notebook-app-component/src/toolbar.tsx
@@ -5,7 +5,6 @@
 // TODO: Fix up a11y eslint here
 // TODO: All the `<li>` below that have role button should just be `<button>` with proper styling
 
-import * as actions from "@nteract/actions";
 import {
   DropdownContent,
   DropdownMenu,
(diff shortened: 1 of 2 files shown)


packages/notebook-app-component/src/toolbar.tsx
```diff
@@ -86,10 +85,19 @@ export const CellToolbar = styled.div`
...
+
+export const CellToolbarMask = styled.div.attrs<CellToolbarMaskProps>(
+  props => ({
     style: {
-      display: props.sourceHidden ? "block" : "none"
+      display: props.cellFocused
+        ? "block"
+        : props.sourceHidden
+        ? "block"
+        : "none"
```
rgbkrk, line 100: This is the majority of the fix we needed, though it's buried under the necessity of passing `cellFocused` through. Since plucking that property off _again_ was needlessly complex I just refactored the components in this file to take props.

captainsafia (approved):
What a pure and crisp glass of water this PR is.

PR #4209, opened 2019-02-13
Fixes Kernel Keybindings

### LINKED ISSUE #4176: Keybindings: Kernel keybindings fail to work
All applications

**Describe the bug**
Keybinding `alt+r, a` fails to restart the kernel

**To Reproduce**

1. Open notebook file
2. Press `alt+r`, then  `a`
3. Open the console

**Expected behavior**
The kernel restarts and runs each cell in the notebook.

**Screenshots**
screen shot 2019-02-01 at 9 47 15 am

stormpython:
Currently, the keybindings for kernel actions does not work. This PR address this bug by passing a `faked` kernelRef to the kernel `actions`.

Closes #4176.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
@@ -147,7 +147,7 @@ export async function main(config: JupyterConfigData, rootEl): Promise<void> {
   ReactDOM.render(
     <React.Fragment>
       <Provider store={store}>
-        <App contentRef={contentRef} />
+        <App contentRef={contentRef} kernelRef={kernelRef} />
       </Provider>
     </React.Fragment>,
     rootEl
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/headers.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/headers.tsx
@@ -51,7 +51,7 @@ export interface State {
 }
 
 class FileHeader extends React.PureComponent<FileHeaderProps, State> {
-  static defaultProps: { children: any } = {
+  static defaultProps: { children: React.ReactNode } = {
     children: null
   };
 
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/headers.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/headers.tsx
@@ -1,7 +1,7 @@
 import { H4 } from "@blueprintjs/core";
 import * as actions from "@nteract/actions";
 
-import { ContentRef } from "@nteract/core";
+import { ContentRef, KernelRef } from "@nteract/core";
 import { ErrorIcon, LoadingIcon, SavingIcon } from "@nteract/iron-icons";
 import * as React from "react";
 import { connect } from "react-redux";
@@ -51,9 +51,10 @@ export interface State {
 }
 
 class FileHeader extends React.PureComponent<FileHeaderProps, State> {
-  static defaultProps = {
+  static defaultProps: { children: any } = {
     children: null
   };
+
   constructor(props: FileHeaderProps) {
     super(props);
 
--- a/applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
@@ -1,8 +1,10 @@
+import { CellType } from "@nteract/commutable";
+import { actions, ContentRef } from "@nteract/core";
 import * as React from "react";
+import { HotKeys, KeyMap } from "react-hotkeys";
 import { connect } from "react-redux";
 import { Dispatch } from "redux";
-
-import { actions, ContentRef } from "@nteract/core";
+import urljoin from "url-join";
 
 // Show nothing while loading the notebook app
 const NotebookPlaceholder = (props: any) => null;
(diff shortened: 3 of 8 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
```diff
@@ -13,98 +15,171 @@ interface State {
...
 }
 
-interface InitialProps {
-  contentRef: ContentRef;
-}
+const mapDispatchToProps = (dispatch: Dispatch, ownProps: NotebookProps) => {
+  const { contentRef } = ownProps;
+  // Creates a fake kernelRef for kernel actions that allow
+  // for interrupt, restart, and kill behaviors.
+  const kernelRef = createKernelRef();
```
stormpython: Is this ok to do? I am not sure. The other option: I can require a `kernelRef` be passed to the `<Notebook />` component.

stormpython:
This PR also moves the `HotKey` functionality from the `.../contents/index.tsx` file to the `.../contents/notebook.tsx` file. This is moved because hotkeys are really only valid for the notebooks and not files or directories.

applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
rgbkrk: Wait, so is the kernelRef here even used? I'm assuming it's looked up via the `contentRef` instead?

applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
```diff
@@ -13,98 +15,171 @@ interface State {
...
+    "@nteract/transform-vega").then(module => {
+      this.props.addTransform(module.VegaLite1);
+      this.props.addTransform(module.VegaLite2);
+      this.props.addTransform(module.Vega2);
+      this.props.addTransform(module.Vega3);
+    });
 
-    // TODO: The geojson transform will likely need some work because of the basemap URL(s)
+    // TODO: The geojson transform will likely need some work
+    // because of the basemap URL(s)
```
rgbkrk: This TODO can be removed so that the TODO bot doesn't add this (again). @captainsafia handled this for the GeoJSON transform IIRC.

applications/jupyter-extension/nteract_on_jupyter/app/contents/notebook.tsx
rgbkrk: The way I'd do this is:

```ts
const makeMapDispatchToProps = (dispatch: Dispatch, ownProps: NotebookProps) => {
  const { contentRef } = ownProps;
  return function mapDispatchToProps(state: AppState) {
    const content = selectors.content(state, { contentRef });

    if (!content || content.type !== "notebook") {
      return { 
        // likely just return some unbound segment of these?
      };
    }

    const kernelRef = content.model.kernelRef; // this is the active kernelRef

    return {
      addTransform: ...,
      INTERRUPT_KERNEL: () => dispatch(actions.interruptKernel({ kernelRef })),
      // ...the rest of them
  }

}
```

rgbkrk: Alternatively, we can set the `kernelRef` in the `props` via `mapStateToProps` then pass them to the respective handlers (?).

This leads me to believe we probably shouldn't use state in `mapDispatch` though -- https://github.com/reduxjs/react-redux/issues/535#issuecomment-257692617

stormpython: The inner `mapDispatchToProps` function doesn't get `state` passed to it. It is the `dispatch` and `ownProps` that are passed.

stormpython: It seems that `state` isn't passed into `mapDispatchToProps` at all. And, since I was setting the `kernelRef` as a prop from `mapStateToProps`, the initial value is undefined. And even after its updated, the `mapDispatchToProps` function still holds onto the initial value because the `dispatch` functions are not reattached unless `ownProps`, i.e. values passed into the component, changes. So, the `kernelRef` is never updated.

rgbkrk: Whoop!

Yeah we'd have to weave in `kernelRef` via `mapStateToProps` via the state, then pass the `kenrelRef` into the functions for `dispatch`.

rgbkrk (changes requested):
We'll still need to handle the `kernelRef` here: https://github.com/nteract/nteract/pull/4209/files#r257295699

stormpython:
Ok, I've added the logic for `kernelRef` to be passed down to the `File` component in order to have the keybindings work properly. However, there are two concerns that I have about the current implementation of keybindings.

1. Pressing `Hotkeys` are not being registered by the UI, possibly because of where the blueprintjs `Hotkeys` component is placed.
2. The Notebook Menu uses the `currentKernelRef` prop but the `kernelRef` which is passed down from the `App` component is used by the `File` component. I am not sure if these two things are the same. My guess is that the `currentKernelRef` is computed because the `NotebookMenu` is a connected component and is pulling the `currentKernelRef` from state in the `mapStateToProps`.

applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
```diff
@@ -147,7 +147,7 @@ export async function main(config: JupyterConfigData, rootEl): Promise<void> {
   ReactDOM.render(
     <React.Fragment>
       <Provider store={store}>
-        <App contentRef={contentRef} />
+        <App contentRef={contentRef} kernelRef={kernelRef} />
```
captainsafia, line 150: I would use `mapStateToProps` to get the kernelRef. This passes the first generated kernelRef to the app, if the user restarts the kernel using the menu this prop won't be updated correctly.

applications/jupyter-extension/nteract_on_jupyter/app/contents/headers.tsx
```diff
@@ -51,9 +51,10 @@ export interface State {
 }
 
 class FileHeader extends React.PureComponent<FileHeaderProps, State> {
-  static defaultProps = {
+  static defaultProps: { children: any } = {
```
captainsafia: `children` can be typed as `React.ReactNode` here.

applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
stormpython: @captainsafia Can we talk about this because I don't have a good understanding of how the `kernelRef` life cycle works. For example, when I refresh the browser, I get a new kernelRef that doesn't match the previous kernelRef. And when I was using the kernelRef from `mapStateToProps` it was complaining that the kernelRef I was using to interrupt or shutdown was not valid.

applications/jupyter-extension/nteract_on_jupyter/app/contents/headers.tsx
stormpython: I will change this.

willingc:
@stormpython Is this good to merge?

willingc:
Oops needs a rebase.

rgbkrk:
Since this is way out of sync with the current `master` and I know Shelby is working on other things at the moment, I'm going to go ahead and close. Thanks Shelby!

PR #4234, opened 2019-02-22
clean up directory listing components

rgbkrk:
Tore out styles that were nested into components that should be done by the `<Entry>` container. Now the timeago component is reusable outside of the directory listing. This fixes the console error you'd sometimes see in the dev console about having a `td` inside of a non-table element.

While I was at it I made the ever incrementing counter for a new save not show "0 seconds ago", "1 seconds ago", ..., "59 seconds ago". Now it just says "less than a minute".


stormpython (approved):
Awesome! LGTM!

PR #4235, opened 2019-02-22
Fix the notebook app component readme

rgbkrk:
Noticed this while creating a simplified notebook component for commuter, as I wanted to see how we were working with the notebook app component.


stormpython (approved):
LGTM!

PR #4236, opened 2019-02-22
clean up all errors and warnings in styleguide

rgbkrk:
# Style guide cleanup!

`yarn docs` for everyone!

I went after cleaning up every single error currently present in the style guide, from `webpackConfig` to the use of `<Media.Plain >` to dangling `<td>`s.

## Migrate to `webpackConfig` key in style guide config

Before:

```
$ yarn docs
yarn run v1.13.0
$ styleguidist server
Warning: No webpack config found. You may need to specify "webpackConfig" option in your style guide config:
https://react-styleguidist.js.org/docs/webpack.html

You can now view your style guide in the browser:

  Local:            http://localhost:6060/
  On your network:  http://192.168.113.55:6060/

 DONE  Compiled successfully!
```

After:

```
$ yarn docs
yarn run v1.13.0
$ styleguidist server
You can now view your style guide in the browser:

  Local:            http://localhost:6060/
  On your network:  http://192.168.113.55:6060/

 DONE  Compiled successfully!
```

## Use `<Media.Plain />` wherever possible instead of assuming `<Plain />`

There were errors for `<Plain />` not being found

## Clean up the `<td>` errors in the component style guide

Make each of the leaf node elements for Directory Listing (Name, Icon, Last Saved) not have to be `<td>`.

### DIFF
--- a/packages/directory-listing/src/components/entry.tsx
+++ b/packages/directory-listing/src/components/entry.tsx
@@ -1,14 +1,9 @@
 import * as React from "react";
-// react-hot-loader uses proxies to the original elements so we need to use
-// their comparison function in case a consumer of these components is
-// using hot module reloading
-import { areComponentsEqual } from "react-hot-loader";
 import styled from "styled-components";
 
-import { Icon } from "./icon";
-import { LastSaved } from "./lastsaved";
-import { Name } from "./name";
+import { areComponentsEqual } from "react-hot-loader";
 
+import { Icon } from "./icon";
 interface EntryProps {
   children: React.ReactNode;
 }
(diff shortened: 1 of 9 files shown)


packages/directory-listing/src/components/entry.tsx
```diff
@@ -42,22 +47,12 @@ export class Entry extends React.PureComponent<EntryProps> {
...
-              Name
-            )
-          ) {
-            return childElement;
-          } else {
-            return <td>{child}</td>;
+        {/* Wrap each child in a `<td>` */}
+        {React.Children.map(this.props.children, child => {
+          if (child && areComponentsEqual((child as any).type, Icon)) {
+            return <td style={{ width: "25px" }}>{child}</td>;
```
stormpython, line 53: I am not so sure about this width being set here. What happens when users resize the window?

stormpython (approved):
Other than the `width` being set to a constant `25px`, this LGTM! I don't think the width being a constant is a deal breaker for merge. But if this is a constant we will want for other `<td>` elements, which should maybe think about setting value of this somewhere else and importing it here.

rgbkrk:
> Other than the `width` being set to a constant `25px`, this LGTM! I don't think the width being a constant is a deal breaker for merge. But if this is a constant we will want for other `<td>` elements, which should maybe think about setting value of this somewhere else and importing it here.

That was a workaround for the Icon so that it was a fixed size table element. 😬 

screen shot 2019-02-22 at 3 18 52 pm

We probably have some structure to rework here. Like, the icon should just be attached / next to the name. I'm not real sure how to shape the API on this one.

PR #4237, opened 2019-02-22
Fix remultiline for consecutive newlines

### LINKED ISSUE #1552: Notebook Markdown import/export bug with consecutive newlines
Running the export on
```py
# %% markdown
# Hello world
#
# Test
# %%
print('hi')
```
gives cells of 
```json
{
  "cells": [
    {
      "cell_type": "markdown",
      "source": [
        "Hello world\n",
        "\nTest"
      ],
      "metadata": {}
    },
    {
      "cell_type": "code",
      "source": [
        "print('hi')"
      ],
      "outputs": [],
      "execution_count": null,
      "metadata": {
        "collapsed": false,
        "outputHidden": false,
        "inputHidden": false
      }
    }
  ],
  "metadata": {},
  "nbformat": 4,
  "nbformat_minor": 4
}
```
The issue is that there's a leading `\n`. So when it gets imported back into Atom, the newline is for some reason stripped and no comment symbol is included:
```py
# %% markdown
# Hello world
# 
Test
# %%
print('hi')
```

I think it's standard in Jupyter notebook files to not have leading newlines; I think they're usually appended to the end. But not sure if that's a requirement or just common.

Either way current Hydrogen behavior is bad on import or export or both.

kylebarron:
As [mentioned here](https://github.com/nteract/nteract/commit/4445a69a18cdb96e570734bd183fbd272a035afb#r32423015) the current `remultiline` acts differently than Jupyter notebook does when converting a cell source string to multiple lines when there are consecutive newlines in the string.

With this change, 
```
test


this

out
```
is converted to
```js
["test\n", "\n", "\n", "this\n", "\n", "out"]
```
with no preceding newlines. 

- [ ] I was expecting code to be auto-formatted as a commit hook. Do I have to run prettier myself?

Fixes: https://github.com/nteract/hydrogen/issues/1552

### DIFF
--- a/packages/commutable/src/primitives.ts
+++ b/packages/commutable/src/primitives.ts
@@ -115,8 +115,8 @@ export function remultiline(s: string | string[]): string[] {
     // Assume already multiline string
     return s;
   }
-  // Use positive lookahead regex to split on newline and retain newline char
-  return s.split(/(.+?(?:\r\n|\n))/g).filter(x => x !== "");
+  // Split on newline and retain newline char
+  return s.split(/(.*?(?:\r\n|\n))/g).filter(x => x !== "");
 }
 
 function isJSONKey(key: string) {
--- a/packages/commutable/__tests__/primitive.spec.ts
+++ b/packages/commutable/__tests__/primitive.spec.ts
@@ -21,8 +21,6 @@ describe("remultiline", () => {
     expect(
       remultiline("test\n\n\nthis\n\nout\n\n\n\n\n\nwhat").map(flagNewlines)
     ).toEqual(
-      // This shows the super weird case with ours which is a bunch of
-      // newlines before the last line...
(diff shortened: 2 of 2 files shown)


packages/commutable/src/primitives.ts
```diff
@@ -115,8 +115,8 @@ export function remultiline(s: string | string[]): string[] {
     // Assume already multiline string
     return s;
   }
-  // Use positive lookahead regex to split on newline and retain newline char
-  return s.split(/(.+?(?:\r\n|\n))/g).filter(x => x !== "");
```
kylebarron, line 119: `?:` isn't a positive lookahead, it's a non-capturing group

rgbkrk: O_____O

rgbkrk:
Thank you so much for this cleanup. Let's get this shipped today along with the other fixes.

BenRussert:
I almost requested @kylebarron's review on the original PR a few weeks ago...

PR #4241, opened 2019-02-26
expose bookstore validation info to front-end via settings dict

mpacer:
This makes #4144 irrelevant. 

This relies on the webapp.settings object as a means of communication between the server extensions.

All of the load_server_extensions should be run by the time this handler can respond to the request.

This avoids all of the strange config finagling that we had to do in #4144. 

I still think as @willingc suggested we should reevaluate how we're using config in the jupyter-extension but for now at least this doesn't require changing any of that.

- need to declare that the content is safe for quotes to be not escaped 
(which is necessary for it to be read as JSON)


rgbkrk (approved):

stormpython (approved):
LGTM!

Issue #3491, opened 2018-10-22
Rendering of "application/vnd.dataresource+json" broken


davidanthoff:
As far as I can tell [this](https://gist.github.com/davidanthoff/e4a97e08b0f77b4be47ecf5c39b85a64) notebook should render properly. Note that it has a ``application/vnd.dataresource+json`` payload.

When I run it or open it in nteract, I get:
![image](https://user-images.githubusercontent.com/1036561/47321269-1f1f6100-d609-11e8-9890-777f13372ce1.png)

This is on nteract 0.11.9 on Windows.

willingc:
Thanks for the report @davidanthoff. 

The current version of nteract is 0.12.3. If you wish to try it and report the results, that would be helpful.

Next step for maintainers: Try the reported notebook with a Julia 1.0 kernel and the latest release.

willingc:
Hi @davidanthoff,

I had a chance to try your example notebook with nteract 0.12.3 and the IJulia 1.0.3 kernel.

I had similar behavior as you did until I set the dataset to the variable in the nteract notebook:

![screenshot 2018-12-18 15 25 47](https://user-images.githubusercontent.com/2680980/50189279-796f3200-02d9-11e9-97da-b1d36b5e5e44.png)

The same notebook displays the dataset in Jupyter notebook as you found.

![screenshot 2018-12-18 15 29 42](https://user-images.githubusercontent.com/2680980/50189448-10d48500-02da-11e9-95c5-dec14543a779.png)
![screenshot 2018-12-18 15 30 08](https://user-images.githubusercontent.com/2680980/50189465-2cd82680-02da-11e9-9275-bb799b5d304b.png)

@rgbkrk This leads me to believe this is not a kernel issue but instead how we are rendering dataset in nteract. Any ideas on the cause?

willingc:
Here's the JS console output of the dataset cell
![screenshot 2018-12-18 15 38 04](https://user-images.githubusercontent.com/2680980/50189640-f8b13580-02da-11e9-8366-4829ff161532.png)

willingc:
Going to try on master as a next step

davidanthoff:
I think ``print(p)`` essentially means that you only get a text MIME output, not the ``application/vnd.dataresource+json`` MIME version. So, yes, I agree, I think this is a probably a bug in how nteract processes that MIME type.

willingc:
[...]

PR #4256, opened 2019-03-06
Fix grid breaking if no primary keys are passed

emeeks:
Simple two-line fix for when the grid breaks if you don't have primary keys at all in your schema.


BenRussert:
Nice, thanks! Are there other order to fix #3491 completely? For the other charts?

If we could change this line:
https://github.com/nteract/nteract/blob/706dae4f8f9eaa5936e49d93607ff71d7da9d4b8/packages/data-explorer/src/types.ts#L27


to make `primaryKey` an optional property I think that would be ideal.

PR #4259, opened 2019-03-07
Update CSS to fix problems

### LINKED ISSUE #4258: Data Explorer plots do not always resize
Data Explorer

**Describe the bug**
Dynamically resizing the window or introducing sidebars does not resize the plots, causing the Data Explorer to overflow horizontally. This is a significant usability issue because the plot selector becomes hidden since it is on the right.

* The plots do resize when _increasing_ the available space, but not when decreasing it
* This is not an issue with the grid view

**To Reproduce**

1. Go to https://t.co/AoPtdG6BFM (nteract notebook containing the Data Explorer)
2. Use the plot selector on the right side of the Data Explorer to select a different view than the grid
3. Resize the window to make it less wide
4. See overflow issue

**Expected behavior**
The Data Explorer plots should ideally resize when the window does, as well as when other content changes around it (e.g., sidebars are toggled open/closed).

At the very least, if there is a mechanism to trigger a reflow, I could call it manually when toggling sidebars.

emeeks:
This should fix #4258 & #4258 @dylanross3 if you get a chance could you take a look?

### DIFF
--- a/packages/data-explorer/src/charts/grid.tsx
+++ b/packages/data-explorer/src/charts/grid.tsx
@@ -30,7 +30,7 @@ interface NumberFilterProps {
 }
 
 const GridWrapper = styled.div`
-  width: calc(100vw - 150px);
+  width: calc(100% - 150px);
 `;
 
 const NumberFilter = (props: NumberFilterProps) => {
--- a/packages/data-explorer/src/charts/grid.tsx
+++ b/packages/data-explorer/src/charts/grid.tsx
@@ -30,7 +30,7 @@ interface NumberFilterProps {
 }
 
 const GridWrapper = styled.div`
-  width: calc(100% - 150px);
+  width: 100%;
 `;
 
 const NumberFilter = (props: NumberFilterProps) => {
--- a/packages/data-explorer/src/index.tsx
+++ b/packages/data-explorer/src/index.tsx
@@ -155,7 +155,7 @@ const FlexItem = styled.div`
 `;
 
 const SemioticWrapper = styled.div`
-  width: "100%";
+  width: 100%;
   .html-legend-item {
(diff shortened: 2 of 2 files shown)


packages/data-explorer/src/charts/grid.tsx
```diff
@@ -30,7 +30,7 @@ interface NumberFilterProps {
 }
 
 const GridWrapper = styled.div`
-  width: calc(100vw - 150px);
+  width: calc(100% - 150px);
```
dylanross3: This adds 150px of white space between the grid and the chart icons. You can probably simplify to just `width: 100%` which looks right to me in my application and also in nteract.

dylanross3:
I can confirm this fixes #4258 and #4259. Thanks for jumping on this so quickly!

emeeks:
@dylanross3 Just changed the two wrappers to width:100% without any calcs, let me know if you see any weirdness if you get a chance, it looked fine to me.

packages/data-explorer/src/index.tsx
```diff
@@ -151,10 +151,11 @@ const FlexWrapper = styled.div`
 
 const FlexItem = styled.div`
   flex: 1;
+  min-width: 0;
 `;
 
 const SemioticWrapper = styled.div`
-  width: "calc(100vw - 200px)";
+  width: "100%";
```
dylanross3: I don't think there should be quotes around this. The Chrome developer tools inspector showed that this style was invalid (both the old and new). Simply `width: 100%` looks fine to me, no issues.

emeeks: Ugh you're right, good catch. I'm js-in-css-in-js-in-css-in-js-css always catches me like this.

PR #4264, opened 2019-03-09
Reducer refactoring

stormpython:
Added return `types` for all the reducer files. 

It helps to see the types of what is being returned, so that we can compare implementations.

### DIFF
--- a/packages/reducers/src/core/entities/hosts.ts
+++ b/packages/reducers/src/core/entities/hosts.ts
@@ -1,6 +1,7 @@
 // Vendor modules
 import * as actions from "@nteract/actions";
 import {
+  HostRecord,
   makeHostsRecord,
   makeJupyterHostRecord,
   makeLocalHostRecord
@@ -9,7 +10,10 @@ import { List, Map } from "immutable";
 import { Action, Reducer } from "redux";
 import { combineReducers } from "redux-immutable";
 
-const byRef = (state = Map(), action: Action): Map<{}, {}> => {
+const byRef = (
+  state = Map() as Map<string, HostRecord>,
+  action: Action
+): Map<string, HostRecord> => {
   let typedAction;
   switch (action.type) {
     case actions.ADD_HOST:
@@ -50,7 +54,7 @@ const refs = (state = List(), action: Action): List<string> => {
 
 export const hosts: Reducer<
   {
-    byRef: Map<{}, {}>;
+    byRef: Map<string, HostRecord>;
     refs: List<string>;
   },
   Action<any>
--- a/packages/reducers/src/core/entities/hosts.ts
+++ b/packages/reducers/src/core/entities/hosts.ts
@@ -1,15 +1,15 @@
-import { List, Map } from "immutable";
-import { Action } from "redux";
-import { combineReducers } from "redux-immutable";
-
+// Vendor modules
 import * as actions from "@nteract/actions";
 import {
   makeHostsRecord,
   makeJupyterHostRecord,
   makeLocalHostRecord
 } from "@nteract/types";
+import { List, Map } from "immutable";
+import { Action, Reducer } from "redux";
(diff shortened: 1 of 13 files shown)


packages/reducers/src/core/entities/hosts.ts
```diff
@@ -1,15 +1,15 @@
...
   makeHostsRecord,
   makeJupyterHostRecord,
   makeLocalHostRecord
 } from "@nteract/types";
+import { List, Map } from "immutable";
+import { Action, Reducer } from "redux";
+import { combineReducers } from "redux-immutable";
 
-const byRef = (state = Map(), action: Action) => {
+const byRef = (state = Map(), action: Action): Map<{}, {}> => {
```
captainsafia: Hm. Shouldn't this return type be something like `Map<string, HostRecord>`?

packages/reducers/src/core/entities/hosts.ts
```diff
@@ -37,7 +37,7 @@ const byRef = (state = Map(), action: Action) => {
   }
 };
 
-const refs = (state = List(), action: Action) => {
+const refs = (state = List(), action: Action): List<any> => {
```
captainsafia: Can we make this `List<string>`?

packages/reducers/src/core/entities/hosts.ts
```diff
@@ -48,4 +48,10 @@ const refs = (state = List(), action: Action) => {
   }
 };
 
-export const hosts = combineReducers({ byRef, refs }, makeHostsRecord as any);
+export const hosts: Reducer<
+  {
+    byRef: Map<{}, {}>;
+    refs: List<any>;
```
captainsafia: Ditto the comments about making the types more specific here.

packages/reducers/src/core/entities/transforms.ts
```diff
@@ -22,7 +22,7 @@ const byId = (state = Map(), action: Action) => {
   }
 };
 
-const displayOrder = (state = List(), action: Action) => {
+const displayOrder = (state = List(), action: Action): List<any> => {
```
captainsafia: Can we make this `List<string>`?

captainsafia (commented):
Cool! This is very useful. I posted comments inline about making some of the types more specific.

packages/reducers/src/core/entities/hosts.ts
stormpython: Because a `map` is instantiated as a default if no `state` object is supplied, the type returned may not only be `Map<string, HostRecord>`.

stormpython: Whoops, I figured out how to prevent the type warnings when making this more specific. Thanks @captainsafia!

stormpython:
@captainsafia this is ready for review again

packages/reducers/src/core/entities/transforms.ts
```diff
@@ -1,11 +1,11 @@
...
-
+// Vendor modules
 import * as actions from "@nteract/actions";
 import { makeTransformsRecord } from "@nteract/types";
+import { List, Map } from "immutable";
+import { Action, Reducer } from "redux";
+import { combineReducers } from "redux-immutable";
 
-const byId = (state = Map(), action: Action) => {
+const byId = (state = Map(), action: Action): Map<{}, {}> => {
```
captainsafia: Can the Map be typed as `Map<string, React.ComponentClass>`?

captainsafia (approved):
Looks good. I left one bit of feedback.

By the way, are the formatting changes a result of us not having run prettier on these files or is the editor you are using also autoformatting?

stormpython:
I think its caused by auto-formatting in my editor. I will fix it.

captainsafia:
I'm gonna go ahead and merge this. The editor related stuff is small enough to be addressed in a follow-on. Thanks for all the hard work!

PR #4268, opened 2019-03-10
Use cross-env when building Jext on Windows

captainsafia:
The Jext build fails on WIndows because cross-env isn't used in the build command.


BenRussert (approved):
Thanks!

Issue #4210, opened 2019-02-14
Make Data Explorer a Fully Connected Component


emeeks:
Currently Data Explorer updates the state optimistically and then emits a metadata change event. Instead, Data Explorer should use local state if no onMetadataChange function is passed to it but otherwise wait for metadata to be passed back down and then update using `getDerivedStateFromProps`

PR #4271, opened 2019-03-11
Separate DataExplorer into Viz and Toolbar components 

### LINKED ISSUE #4204: Separate `data-explorer` components and export to allow composition
The data explorer is made up of several sub-components (charts/controls) that could be exported and composed to fit the needs of different front-ends/applications.

I think we should start with the toolbar. Here is a rough sketch taken from @rgbkrk's comment on slack:

```jsx
<DataExplorer
  data={}
  metadata={}
/>

... Elsewhere

<Toolbar {...actions} />
```

BenRussert:
This allows for usage like this:

```jsx
 <DataExplorer {...dataExplorerProps}>
    <Viz />
    <Toolbar />
</DataExplorer>
```
to resolve #4204

Allows external package users to customize the position/existence of the toolbar. Down the line, you should be able to create a custom toolbar, but I thought we should address #4210 first.

Previously I tried another approach that broke out the visualization controls as well. I wasn't happy with the diff and how the code looked after. After stepping away from it a while, I thought of the approach below. It has a much smaller diff and is intended to be expanded on in later PRs. To separate the controls, disable unwanted `view`s, add error boundaries, etc.

Todos before we can merge:

- [x] Decide how exports work
- should there be a default single component like currently along with individual named exports for optional customization?
- [x] Update README with usage once we agree on how to handle that
- [x] Ensure compatibility with `RichMedia`
- [x] Integrate into jext and desktop


packages/data-explorer/src/Display.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+function PlaceHolder() {
+  return <div>This should be a display element!</div>;
+}
+
+Display.defaultProps = { componentType: "display", children: <PlaceHolder /> };
+
+export function Display({ children, componentType }: Props) {
+  return <FlexItem>{children}</FlexItem>;
+}
```
BenRussert: The `Display` component will eventually be an error boundary and handle the different `view` components as children, but that would be in another PR.

Also, I'm thinking about calling this `<Viz />` or something other than `Display`, since we've already had a `Display` component in the nteract ecosystem. What do you think?

emeeks: I like that a lot. I probably won't get a chance to review this until tomorrow morning.

BenRussert:
I modified a branch of the `semiotic-docs` and `yarn link`'ed to my locally built `@nteract/data-explorer` so I could see the component running live in localhost.  [You can check out the now deployment here](https://semiotic-docs-94pfq0m67.now.sh/guides/data-explorer)

If you want to know how to set up the semiotic docs like I did, I added instructions to the [readme on this branch of my `semiotic-docs` fork.](https://github.com/BenRussert/semiotic-docs/tree/data-explorer-example#use-for-dataexplorer-development)

emeeks (approved):

BenRussert:
I have the other bullets covered  in a local branch I will push later. There won't be required changes for apps that use the current component because a complete data explorer will still be the default export.

stormpython (approved):
LGTM!

packages/data-explorer/src/index.tsx
```diff
@@ -518,22 +517,51 @@ class DataExplorer extends React.PureComponent<Partial<Props>, State> {
...
+  );
+}
+
+DataExplorerDefault.defaultProps = {
+  mediaType
+};
+DataExplorerDefault.displayName = "DataExplorerDefault";
+
+// For the jupyter extension to load MIMETYPE must be present.
+DataExplorerDefault.MIMETYPE = mediaType;
```
BenRussert, line 569: I was stuck for a while tonight because I couldn't get jext to render a data-explorer, but it was due to this line. I got deep into the jext code in the process though, so I learned some things in the process.

captainsafia (approved):
Thanks for the tests and docs updates!

PR #4288, opened 2019-03-21
Fix area chart rendering

emeeks:
My earlier PR introduced an issue with how hexes in the area chart were styled causing it not to render. This corrects it.


BenRussert:
Build failing was unrelated due to timeout. Looks good to merge.

PR #4291, opened 2019-03-22
Remove Blueprint Controls

### LINKED ISSUE #4261: Data Explorer blueprintjs styles affect global styles
Data Explorer

**Describe the bug**
Following the instructions on the Data Explorer package [README](https://github.com/nteract/nteract/tree/master/packages/data-explorer), I am including the following imports in my application.

```
// Toolbars and other UI elements
import "@blueprintjs/core/lib/css/blueprint.css";
import "@blueprintjs/select/lib/css/blueprint-select.css";

// CSS for the grid view on the data explorer
```

The blueprintjs styles are affecting the rest of my application.

**Expected behavior**
The imported styles should only affect the Data Explorer.

**Additional context**
Might be related to #3489.

As a side note, it's not a good practice to have the consumer import a transitive dependency of your package. Alternatives include importing the styles in the Data Explorer package itself or declaring it as a peer dependency.

emeeks:
Blueprint CSS has had some issues for external partners so this just removes the Blueprint components and replaces them with simple HTML components. It still looks very functional but seems to maintain all the functionality. Should resolve #4261 

* Replaces Blueprint components with HTML components
* Leaves the blueprint code (commented out) so that someone can take a look at it down the line

### DIFF
--- a/packages/data-explorer/README.md
+++ b/packages/data-explorer/README.md
@@ -42,10 +42,6 @@ Note: you also need to include the CSS for a few dependencies as well
 
 ```
 
-// Toolbars and other UI elements
-import "@blueprintjs/core/lib/css/blueprint.css";
-import "@blueprintjs/select/lib/css/blueprint-select.css";
-
 // CSS for the grid view on the data explorer
 import "react-table/react-table.css";
 ```
--- a/packages/data-explorer/package.json
+++ b/packages/data-explorer/package.json
@@ -20,8 +20,6 @@
   ],
   "license": "BSD-3-Clause",
   "dependencies": {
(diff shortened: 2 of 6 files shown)


BenRussert (commented):
Just a couple notes, this looks pretty good overall and actually not a whole lot of noticeable difference for the user!

- Some "digital hoarding" going on, think we could :fire: most of the lines that are commented out.
- Maybe unrelated, but Hexbin isn't working.

emeeks:
Good points, @BenRussert, I'll get rid of the commented bits. Hexbin should be fixed as of #4288 and this was branched while that was in flight. I'll make sure to check, though. Thanks!

PR #4311, opened 2019-03-31
Use loading state within contentRef object

captainsafia:
Not sure why we were setting the loading state under the path, I recall @rgbkrk  and I having a conversation about this at some point, but can't recall the final conclusion.

The loading state is set under the file path but lingers once the content has been loaded creating a rather confusing state tree.


BenRussert:
Merging since the ci fail is unrelated. (kernel spawning test that always seems fine locally, but times out sometimes on ci)

PR #4312, opened 2019-04-01
Directory listing fixups

BenRussert:


packages/directory-listing/__tests__/index.spec.tsx
```diff
@@ -6,16 +6,12 @@ import { Entry, Icon, LastSaved, Listing, Name } from "../src";
 
 describe("Listing", () => {
   it("accepts props and renders entries in directory in a table", () => {
-    const component = (
+    const component = shallow(
```
BenRussert, line 9: The snapshot was just `undefined` since `component` in `toJson(component)` wasn't an enzyme object

packages/directory-listing/__tests__/index.spec.tsx
```diff
@@ -29,11 +25,7 @@ describe("Entry", () => {
       <Entry>
         <Icon fileType={"directory"} />
         <Name>{"linky jr"}</Name>
-        <LastSaved
-          last_modified={
-            "Fri Jun 22 2018 00:15:55 GMT-0400 (Eastern Daylight Time)"
-          }
-        />
+        <LastSaved lastModified={new Date("2018-06-22 00:15:55 GMT")} />
```
BenRussert, line 28: TypeScript insisted on this change. This prop wants a `Date` object or only certain types of strings.

packages/directory-listing/src/components/icon.tsx
```diff
@@ -30,7 +24,6 @@ export class Icon extends React.PureComponent<IconProps> {
   };
 
   render() {
-    const styles = { color: this.props.color };
```
BenRussert, line 33: This needs to get passed down to `IconWrapper` or all icons will be default blue regardless of props. This is the bug I saw in our components page that led to the PR.

BenRussert (commented):
Before:
with-color-bug

After:

after-color-prop

captainsafia (approved):
Thank you for the fixes!

PR #4324, opened 2019-04-07
Update RELEASING file with more information

captainsafia:
These updates:

- Establish a weekly release cadence for packages
- Establish a biweekly release cadence for the desktop app
- Clarify some instructions in the releasing doc for desktop releases

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,6 +1,18 @@
 # :package: Let's package a release!
 
-## All platform precursor
+## Release Schedule
+Unless there is an emergency bug or security fix, packages are released on a weekly cadence on Mondays. The desktop app is released on a bi-weekly cadence on Mondays. So, every Monday a package release is cut and every other Monday a desktop release is cut.
+
+## Releasing the Packages
+In order to release the JavaScript packages to npm, you will need to run the `npx lerna publish` command in the root directory. This will
+
+## Releasing the Desktop App and Packages
+
+### A special note for macOS
+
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -1,21 +1,36 @@
...
+```
 
-In order to build a signed copy with working auto-update, you will need to join the Apple developer program and get a certificate. The [Electron docs have a document on submitting your app to the app store](https://github.com/electron/electron/blob/master/docs/tutorial/mac-app-store-submission-guide.md), you only have to get through to the certificate step.
+Note that `setx` will apply the environment variable in subsequence Command Prompt sessions.
 
-## Release Process
 **Note: Make sure that your remote origin is set as `git@github.com:nteract/nteract.git`.**
 
-1.  Make sure the release is working by running `yarn dist` and testing the built app inside the `./applications/desktop/dist/` folder. You can build for all platforms using `yarn dist:all`.
+1.  Make sure the release is working by running `yarn dist` and testing the built app inside the `./applications/desktop/dist/` folder. You can build for all platforms using `yarn dist:all`. Note that you can only build for all platforms on macOS. macOS builds cannot be built on Windows and Linux.
```
MSeal, line 33: > macOS builds cannot be built on Windows and Linux.

:(

MSeal: But a good doc addition as I could see this leading someone astray if they didn't know.

willingc:
A bit off topic: do we build binaries nightly for nteract desktop? Might make sense to do so and keep somewhere.

BenRussert:
> do we build binaries nightly 

Not that I know of, but I think nightlies would be a good idea @willingc.

captainsafia:
We don't have nightly builds at the moment although we could configure them in a CircleCI build. I made an issue for it in https://github.com/nteract/nteract/issues/4328.

stormpython (approved):

PR #4326, opened 2019-04-07
Dynamically set base path for notebook

### LINKED ISSUE #1242: Markdown cells don't render local images

captainsafia:
Closes https://github.com/nteract/nteract/issues/1242.

Relative paths within a notebook are now relative to the location of the notebook file.

For new notebooks, this will continue to default to the location of the install. If we want, we can do something like default it to their home directory.

### DIFF
--- a/packages/notebook-app-component/__tests__/notebook.spec.tsx
+++ b/packages/notebook-app-component/__tests__/notebook.spec.tsx
@@ -32,6 +32,22 @@ describe("NotebookApp", () => {
     expect(component).not.toBeNull();
   });
 
+  describe("NotebookApp", () => {
+    test("contains a base attribute", () => {
+      const component = shallow(
+        <NotebookApp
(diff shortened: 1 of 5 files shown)


mpacer:
What is react helmet doing?

I hadn’t realized this about relative paths wasn’t being already taken into account. Could you give a concrete example of where this comes up?

mpacer:
Ah from #1242 


it only affects markdown cells.

mpacer:
Is there a good place to put tests for this behaviour?

BenRussert (approved):

captainsafia:
> Is there a good place to put tests for this behaviour?

Hm. Depends on what this behavior is referring to. If we just wanted to check that the `<base>` attribute was set properly, we could do that in Jest. I'll add those tests to this PR.

For checking that the image was rendered properly, we would probably need to set up some integration tests.

PR #4334, opened 2019-04-10
Don't set headers if values don't exist

captainsafia:
If there is no authorization or xsrf token data, don't send them in the header instead of sending invalid values.


BenRussert:
Looks good to me. So I can understand better, was this causing an issue someplace?

PR #4337, opened 2019-04-11
Add support for display_id in execute_result

### LINKED ISSUE #1380: Execute Result can also have display_id
Our current support of the update display mechanics only support `display_data` (and `update_display_data`). We should also support `execute_result`.

I partially started this in #1198 and would love to see it to completion. I probably won't come back to this one so if anyone else wants to take a try at it I'm more than happy to advise.

captainsafia:
Closes https://github.com/nteract/nteract/issues/1380.

This is basically the same stuff that @rgbkrk started in https://github.com/nteract/nteract/pull/1198 except I fixed the failing test.

Since it was reducing over the `keyPaths` every time an output was appended, it was overriding each output with the last output that was appended.

I removed the reduction over `keyPaths` and used a single `updateIn` to update the most recently inserted output as `appendOutput` was called.

### DIFF
--- a/packages/reducers/src/core/entities/contents/notebook.ts
+++ b/packages/reducers/src/core/entities/contents/notebook.ts
@@ -211,11 +211,14 @@ function appendOutput(
   const output = action.payload.output;
   const cellId = action.payload.id;
 
-  // If it's display data and it doesn't have a display id, fold it in like non
-  // display data
+  /**
+   * If it is not a display_data or execute_result with
+   * a display_id, then treat it as a normal output and don't
+   * add its index to the keyPaths.
+   */
   if (
-    (output.output_type !== "execute_result" ||
-      output.output_type !== "display_data") &&
+    output.output_type !== "execute_result" &&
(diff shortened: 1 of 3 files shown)


packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -214,7 +214,8 @@ function appendOutput(
   // If it's display data and it doesn't have a display id, fold it in like non
   // display data
   if (
-    output.output_type !== "display_data" ||
+    (output.output_type !== "execute_result" ||
+      output.output_type !== "display_data") &&
     !has(output, "transient.display_id")
   ) {
```
BenRussert, line 225:
```js
(output.output_type !== "execute_result" ||
      output.output_type !== "display_data") 
```
Should the `||` be `&&`?

captainsafia: .....Yes, I believe you are correct. Thanks for catching this!

BenRussert (approved):

PR #4338, opened 2019-04-11
Add removeHost action for removing hosts

captainsafia:
We've got an `addHost` action that allows users to inset a new host record into the state, but no corresponding `removeHost` action to remove host records that are no longer needed (for example, in the case of disconnecting from a remote Jupyter server).

### DIFF
--- a/packages/reducers/src/core/entities/hosts.ts
+++ b/packages/reducers/src/core/entities/hosts.ts
@@ -36,6 +36,9 @@ const byRef = (
             `Unrecognized host type "${typedAction.payload.host.type}".`
           );
       }
+    case actions.REMOVE_HOST:
+      typedAction = action as actions.RemoveHost;
+      return state.remove(typedAction.payload.hostRef);
     default:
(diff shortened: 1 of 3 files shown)


packages/reducers/src/core/entities/hosts.ts
```diff
@@ -36,17 +36,22 @@ const byRef = (
...
-  let typedAction;
   switch (action.type) {
     case actions.ADD_HOST:
-      typedAction = action as actions.AddHost;
-      return state.push(typedAction.payload.hostRef);
+      return state.push((action as actions.AddHost).payload.hostRef);
+    case actions.REMOVE_HOST:
+      return state.filter(
+        hostRef => hostRef !== (action as actions.RemoveHost).payload.hostRef
+      );
```
BenRussert, line 54: I wish typescript could figure out action types based on `action.type` so we didn't need to cast :man_shrugging:

BenRussert (approved):

PR #4339, opened 2019-04-11
Pin react-table and vendorize stylesheet deps

### LINKED ISSUE #4306: Data Explorer should not require importing react-table styles
Data Explorer

**Describe the bug**
Currently, the Data Explorer package README explains I must import the CSS for `react-table`.

```
// CSS for the grid view on the data explorer
```

It goes against best practices in the NPM ecosystem to import transitive dependencies. This is because where NPM chooses to place a package depends on the whole dependency graph, so `react-table` may not be hoisted to the top level if two different packages depended on different versions of it for instance, meaning the consumer of the Data Explorer package may have to do `import "@nteract/data-explorer/node_modules/react-table/react-table.css"` to workaround. This scenario could easily occur to a consumer when they simply update some packages.

**Expected behavior**
The `react-table` dependency should either be listed as a peer dependency, meaning the consumer of the Data Explorer must bring it in, or the Data Explorer package should import the styles itself.

BenRussert:
This is a quick and dirty copy/paste of the react-table and react-table-hoc-fixed-columns stylesheet dependencies.

In the future, instead of hard pinning these react-table dependencies, we could set up a script to produce the vendorized css files to account for any css updates (see @nteract/styled-blueprintjsx as an example). I thought I would keep it simple for now and iterate since this would have no visible difference to the user.

Fixes #4306

### DIFF
--- a/applications/desktop/src/notebook/index.tsx
+++ b/applications/desktop/src/notebook/index.tsx
@@ -7,10 +7,8 @@ import * as MathJax from "@nteract/mathjax";
 import "@blueprintjs/core/lib/css/blueprint.css";
 import "@blueprintjs/select/lib/css/blueprint-select.css";
 
-import "codemirror/lib/codemirror.css";
 import "codemirror/addon/hint/show-hint.css";
-
-import "react-table/react-table.css";
+import "codemirror/lib/codemirror.css";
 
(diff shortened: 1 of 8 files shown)


jdetle:
Looks good!

PR #4348, opened 2019-04-14
Custom styles and themes for react-table

BenRussert:
Add custom react-table styling with support for light/dark theme props
- These styles are just a start so we can iterate
- Fixes a couple styles issues from #4339
- No nteract internal css variables are used so it can work on external apps

Most of our output component already expect an optional theme prop, but we weren't passing down in our apps previously, so I've added that support.


BenRussert:
This will move all grid view (react-table) related styles into a `CustomReactTableStyles` styled component that takes a `type theme: "dark" | "light"` prop. This prop is used by the various nteract apps, but can also be used externally `<DataExplorer data={data} theme={theme}/>`

data-explorer table styles with this PR mp4

emeeks (approved):

PR #4349, opened 2019-04-17
Allow users to pass down custom headers to rx-jupyter

captainsafia:
Adds support for passing custom headers to rx-jupyter's base HTTP request object from the epics and actions.

### DIFF
--- a/packages/actions/__tests__/actions-spec.ts
+++ b/packages/actions/__tests__/actions-spec.ts
@@ -192,7 +192,7 @@ describe("launchKernelByName", () => {
         kernelRef,
         contentRef,
         selectNextKernel: false,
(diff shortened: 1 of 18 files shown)


packages/actions/__tests__/actions-spec.ts
```diff
@@ -191,7 +191,8 @@ describe("launchKernelByName", () => {
         cwd: ".",
         kernelRef,
         contentRef,
-        selectNextKernel: false
+        selectNextKernel: false,
+        opts: { headers: { "Contennt-Type": "application/json" } }
```
BenRussert: "Content-Type" has an extra "n"

BenRussert (approved):
Looks solid to me.

PR #4370, opened 2019-04-24
Add support for input_request/input_reply messages

### LINKED ISSUE #741: Support STDIN

captainsafia:
Closes https://github.com/nteract/nteract/issues/741.


packages/commutable/src/cells.ts
```diff
@@ -21,6 +21,7 @@ export interface CodeCellParams {
   execution_count: ExecutionCount;
   source: string;
   outputs: ImmutableList<ImmutableOutput>;
+  prompt?: ImmutableMap<string, any>;
```
rgbkrk: Based on my reading of the rest of the code, I think this can be made into a record with the fields `prompt` and `password`, then make this able to be either the `PromptRecord` or `null`.

rgbkrk: I am a little hesitant about putting things in the cell state for commutable that aren't serialized to the document. How do you feel about making this part of the notebook structure at a higher level? I realize it's easier when we can pluck it from a cell within `notebook.cellMap[cellId].prompt` but maybe it's ok to have `notebook.transientCellData[cellId].prompt`, kind of like we have pagers in a separate setup?

packages/outputs/src/components/prompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+          value={this.state.value}
+          onChange={this.handleChange}
+        />
+        <input type="submit" />
+      </form>
+    );
+  }
+}
+
+export default PromptRequest;
```
rgbkrk, line 57: For styling, it might be worth checking out how the `input` box is styled here: http://www.htmlfivewow.com/demos/terminal/terminal.html

packages/commutable/src/cells.ts
captainsafia: Done. Moved things into a separate cellPrompts state.

packages/types/src/entities/contents/notebook.ts
```diff
@@ -84,6 +85,7 @@ export const makeDocumentRecord = Immutable.Record<DocumentRecordProps>({
     keyPathsForDisplays: Immutable.Map()
   }),
   cellPagers: Immutable.Map(),
+  cellPrompts: Immutable.Map(),
```
rgbkrk, line 88: 👍

packages/types/src/entities/contents/notebook.ts
```diff
@@ -70,6 +70,7 @@ export interface DocumentRecordProps {
   // right now it's keypaths and then it looks like it's able to handle any per
   // cell transient data that will be deleted when the kernel is restarted
   cellPagers: any;
+  cellPrompts: Immutable.Map<CellId, InputRequestMessage>;
```
rgbkrk, line 73: 👍

rgbkrk (approved):

rgbkrk:
It looks like jest snapshots need updating.

PR #4380, opened 2019-04-26
Add support for AJAX and WS options in ServerConfig

### LINKED ISSUE #4373: allow to pass protocol into websocket construction
In order to authenticate the websocket connection, we need to attach authorization token in the headers of the websocket request, using the Sec-WebSocket-Protocol as mentioned in https://stackoverflow.com/questions/22383089/is-it-possible-to-use-bearer-authentication-for-websocket-upgrade-requests/35108078#35108078

However currently in nteract/kernels.ts (https://github.com/nteract/nteract/blob/8f86648c6e4523f1dd5dd532e9c8f2c1105d6827/packages/rx-jupyter/src/kernels.ts#L184) we only pass in an URL, there is no way to pass down the extra headers we need.

One idea is to add additional config parameter in the HostAppRecord for websocket construction, and we pass that down to the rxjs/webSocket

captainsafia:
This PR:
- Reverts changes introduced in https://github.com/nteract/nteract/pull/4349
- Allows users to pass custom options to configure Ajax and Websocket requests via the server config
- Closes https://github.com/nteract/nteract/issues/4373

**Example**

A JupyterHostRecord configured with the following properties will send a From request header on all headers going outbound from rx-jupyter and will add the "ws_session; blah" string to the `Sec-WebSocket-Protocol` header.

```
const host = makeJupyterHostRecord({
    origin: "https://example.com",
    ajaxOptions: { headers: { From: "email@example.com" } },
    wsProtocol: ["ws_session", "blah"]
});
```

### DIFF
--- a/applications/desktop/src/notebook/menu.ts
+++ b/applications/desktop/src/notebook/menu.ts
@@ -145,7 +145,7 @@ export function promptUserAboutNewKernel(
 
           store.dispatch(
             actions.launchKernelByName({
-              kernelSpecName: kernel.kernelSpecName || "",
+              kernelSpecName: kernel.kernelSpecName,
               cwd,
               selectNextKernel: true,
               kernelRef,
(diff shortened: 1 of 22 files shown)


stormpython (approved):
LGTM!

Issue #4379, opened 2019-04-25
DX Line chart Performance is Bad with 5000 datapoints


emeeks:
Introduce user controllable sampling (smooth every n points or base it on screen resolution)

Implement canvas rendering for line charts

Tooltip bug with traffic failover dataset

Issue #4375, opened 2019-04-25
Data Explorer: Reliance on CSS variables


dylanross3:
**Application or Package Used**
Data Explorer

**Describe the bug**
When I include the Data Explorer component in my application, the toolbar icons do not look the same as they do in the context of an nteract notebook. In particular, the Dendrogram icon is not visible (see screenshot).

**Expected behavior**
- At the very least, I expect the Dendrogram icon to be visible.
- It would also be nice if the icons looked like they do in nteract without requiring me to define the CSS variables myself

**Screenshots**
![image](https://user-images.githubusercontent.com/2413692/56702392-9cab5300-66b8-11e9-8f39-2227e15b30d5.png)

Issue #4369, opened 2019-04-23
Data Explorer table not displaying booleans


dylanross3:
**Application or Package Used**
Data Explorer

**Describe the bug**
Boolean values are not displayed.

**To Reproduce**
Execute the following code in a notebook that supports the Data Explorer and look at the table view.

```python
import pandas as pd
pd.options.display.html.table_schema = True
pd.options.display.max_rows = None

Data = {'animal':  ['cat', 'lion', 'dog', 'tiger'],
        'feline': [True, True, False, True]
        }

df = pd.DataFrame (Data, columns = ['animal', 'feline'])
df
```

Alternatively, render the Data Explorer component directly with the following data/schema:
https://pastebin.com/raw/LduMB3Ce

**Expected behavior**
I expect to see a representation of boolean fields instead of empty cells.

**Screenshots**
![image](https://user-images.githubusercontent.com/2413692/56597110-a5136900-65a6-11e9-846d-370274010f29.png)

**Additional context**
Seems related to #3105

Issue #4303, opened 2019-03-28
Data Explorer onMetadataChange does not fire when switching to the grid view


dylanross3:
**Application or Package Used**
Data Explorer

**Describe the bug**
Using the Data Explorer component in my React application, when switching the chart type to a view other than the grid view, the `onMetadataChange` prop is called as expected, but this does not occur when switching to the grid view.

**Expected behavior**
I expect the `onMetadataChange` prop to be called when switching to the grid view.

PR #4382, opened 2019-04-27
Dx improvements

emeeks:
This fixes a few outstanding minor issues in Data Explorer: #4369, #4379, #4375, #4303

### DIFF
--- a/packages/data-explorer/src/icons.tsx
+++ b/packages/data-explorer/src/icons.tsx
@@ -4,8 +4,8 @@ import styled from "styled-components";
 
 const IconPath = styled.path`
   & {
-    fill: var(--theme-app-bg);
-    stroke: var(--theme-app-fg);
+    fill: var(--theme-app-bg, #ccc);
+    stroke: var(--theme-app-fg, #2b2b2b);
   }
 `;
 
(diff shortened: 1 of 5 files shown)


packages/data-explorer/src/icons.tsx
```diff
@@ -4,8 +4,8 @@ import styled from "styled-components";
 
 const IconPath = styled.path`
   & {
-    fill: var(--theme-app-bg);
-    stroke: var(--theme-app-fg);
+    fill: var(--theme-app-bg, #ccc);
+    stroke: var(--theme-app-fg, #2b2b2b);
```
BenRussert, line 8: TIL! We probably have other places in this and other packages to use default values.

packages/data-explorer/src/index.tsx
```diff
@@ -471,7 +465,51 @@ class DataExplorer extends React.PureComponent<Partial<Props>, State> {
...
+          }
+        },
+        mediaType
+      );
+    }
+  }
+
   setGrid = () => {
+    this.updateMetadata({ view: "grid" })
     this.setState({ view: "grid" });
```
BenRussert, line 513: Maybe for another PR, but I think this might be easier to understand and keep track of if `onMetadataChange` was called once per render cycle, whenever state changes, and all in one place. Like in `componentDidUpdate`. Might be some factors I haven't thought of though.

BenRussert (approved):
Looks good to me. We should add tests for `onMetadataChange` :soon:.

PR #4383, opened 2019-04-27
[WIP] Fix editor tooltip

BenRussert:
Extracts the editor tooltip to a separate component and makes it easier to destroy

### DIFF
--- a/packages/editor/src/components/tooltip.tsx
+++ b/packages/editor/src/components/tooltip.tsx
@@ -13,6 +13,7 @@ const TipButton: StyledComponent<"button", never> = styled.button`
 `;
 
 const Tip: StyledComponent<"div", never> = styled.div`
+  position: absolute;
   padding: 20px 20px 50px 20px;
   margin: 30px 20px 50px 20px;
   box-shadow: 2px 2px 50px rgba(0, 0, 0, 0.2);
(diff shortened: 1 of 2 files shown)


BenRussert:
I'm still having a little styles trouble I can look at later, but this is an improvement. Take this branch for a spin if you have a chance and try the `ctrl+,` tooltip.

The style issue is that long tooltips don't wrap or scroll if they extend off screen.

packages/editor/src/components/tooltip.tsx
```diff
@@ -0,0 +1,84 @@
...
+  box-shadow: 2px 2px 50px rgba(0, 0, 0, 0.2);
+  white-space: pre-wrap;
+  background-color: var(--theme-app-bg);
+  z-index: 9999999;
+`;
+
+export interface ToolTipProps {
+  editor: EditorFromTextArea;
+  toolTipRef: React.RefObject<HTMLDivElement>;
+  bundle: any;
```
captainsafia, line 28: Should this be typed as a MimeBundle from the types library?

packages/editor/src/components/tooltip.tsx
```diff
@@ -0,0 +1,84 @@
...
+  deleteTip: () => void;
+}
+
+export interface ToolTipState {}
```
captainsafia, line 32: If there is no internal state, then we can just remove this interface definition.

captainsafia (commented):
Good optimization!

I'm trying to set this up locally but for some reason `npm start` fails for me on this PR. I don't think it is related to this code change though...

BenRussert:
I will come back to this, but didn't want to leave hanging open now. 

One interesting thing I found was that the codemirror `addWidget` method seems to break the normal way that `React.createPortal` works. This is why even on master when we try to portal to `document.body` the element stays embedded in he editor component.

PR #4384, opened 2019-04-28
Update README

captainsafia:
I made a few updates to the README based on recent changes.

Also removed references to @rgbkrk's email. I think we have a maintainer's Google Group somewhere that we should use instead so any incoming emails get surfaced to everyone.

I'll either have to dig up that Google group or make one.


rgbkrk:
Looks like the `fs-kernels` stuff is failing tests unrelated to this README change. I'll go ahead and merge this for now.

PR #4387, opened 2019-05-03
Use verbs Quit/Cancel in quit prompt insted of Yes/No

jarmokivekas:
When quitting the Nteract desktop app, I always need to read the prompt to know which button to press. It would be nice to have verbs on the button labels to save a second or two. 

It's a stupidly small change. I know.

Go from this:

Before

To this:

after


captainsafia (approved):
This is a great edit! Thank you!

Issue #4389, opened 2019-05-04
Nteract desktop statusbar exported to PDF


jarmokivekas:
**Application or Package Used**

Nteract desktop, version 0.14.0 installed in Ubuntu using .deb installer from Github releases

**Describe the bug**

The status bar at the bottom of the Nteract screen is included at the bottom of every page in the PDF export. Sometimes the status bar overlaps with notebook content, making the text unreadable.

This is a regression introduced sometime after version 0.12.x release. I updated yesterday from that version to 0.14.0 and this bug is a new behavior after the update

**To Reproduce**

1. Open new notebook, add a code cell
2. Run something that produces losts of output, e.g.
```python
for i in range(0,100):
   print(i)
```
1. File -> Export to PDF
5. Open the exported PDF file, bottom of first the page will look something like this:

![Screenshot from 2019-05-04 10-01-53](https://user-images.githubusercontent.com/3074453/57175358-ac166300-6e53-11e9-9a8f-329bf3441072.png)

**Expected behavior**

The status bar should be hidden/removed. The number 45 in the screenshot should be visible.

**Desktop (please complete the following information):**

- OS: Ubuntu 18.10
- Browser: does not apply, using desktop
- Version: 0.14.0 release using .deb installer. Occurs also on 0.14.1 from up-to-date master (commit f63a01c40)

jarmokivekas:
I downgraded to previous versions. Seems v0.12.1 does not have this bug, but v0.13.0 does.

jarmokivekas:
Selecting the status bar in the developer tools and running `$0.style.visibility= "hidden"` from the developer console before exporting works as a work-around (pdf looks fine).

On a related note, also tab-completetion dropdown lists are exported to pdf, although just pressing ESC before export gets rid of them.

![image](https://user-images.githubusercontent.com/3074453/57176328-8000df00-6e5f-11e9-857b-010321480839.png)

Might be worthwhile to see if adding  `@media print` css rules to these elements fixes this issue, similar to pull #1407.

captainsafia:
This can be closed via https://github.com/nteract/nteract/pull/4390.

Issue #4346, opened 2019-04-13
Remove kernel, status, and savetime from PDF export


eigenvivek:
I am exporting a notebook as a PDF to I can share it with others. There is a bar at the bottom that contains information that I don't want to share, but I can't figure out how to hide this information.

Is there a way to hide this? If not, I would love if I could remove it! 

Here's an example of my problem:
![streaming_bayes](https://user-images.githubusercontent.com/29757116/56082609-c9be5280-5de8-11e9-8089-89c91275b802.png)

captainsafia:
Hello @v715! Thanks for opening this issue and great notebook!

@jdetle Do you know if we can customize the `printToPDF` function to print only certain parts of the DOM?

jdetle:
Yeah I can do this! All we need to do is add a media query in the css sheets.

jdetle:
Looked into this today, it looks like some media queries from before styled components haven't got brought in yet. I can find the relevant components and move them over there.

jdetle:
Closed by #4390

PR #4390, opened 2019-05-04
fix(ui): hide status bar during pdf export

jarmokivekas:
Fixes issue #4389 by adding a missing class attribute to the top-level div of the status bar. 

The css class rule for hiding the status bar during pdf export already exists from before, but the div lost its class attribute at some point. This PR puts it back. 
 
https://github.com/nteract/nteract/blob/f63a01c4067b05541ebddd3932345d7102ea3e5b/applications/desktop/static/index.html#L276-L282

### DIFF
--- a/applications/desktop/.gitignore
+++ b/applications/desktop/.gitignore
@@ -1,0 +1,2 @@
+# the default file name when exporting unsaved notebook to pdf from the desktop app
+..pdf
--- a/applications/desktop/static/index.html
+++ b/applications/desktop/static/index.html
@@ -279,12 +279,6 @@
           box-shadow: none !important;
         }
       }
-
-      @media not print {
-        .cells {
-          padding-bottom: calc(100vh - 110px);
-        }
-      }
     </style>
   </head>
   <body>
--- a/applications/desktop/static/index.html
+++ b/applications/desktop/static/index.html
@@ -278,6 +278,9 @@
           /*remove the box-shadow around of cells when exporting to pdf */
           box-shadow: none !important;
         }
+        .notifications-wrapper {
+          display: none;
+        }
       }
     </style>
   </head>
--- a/applications/desktop/static/index.html
+++ b/applications/desktop/static/index.html
@@ -277,9 +277,6 @@
         * {
           box-shadow: none !important;
         }
-        .status-bar {
-          display: none !important;
-        }
         .notifications-wrapper {
           display: none !important;
         }
--- a/applications/desktop/static/index.html
+++ b/applications/desktop/static/index.html
@@ -275,15 +275,8 @@
 
       @media print {
         * {
+          /*remove the box-shadow around of cells when exporting to pdf */
           box-shadow: none !important;
-        }
-        .cell.focused {
-          border: none;
-          background: var(--theme-cell-bg, white) !important;
-        }
-        .cell:focus .prompt,
-        .cell.focused .prompt {
-          background: var(--theme-pager-bg, #fafafa) !important;
         }
       }
 
--- a/applications/desktop/static/index.html
+++ b/applications/desktop/static/index.html
@@ -277,15 +277,6 @@
         * {
           box-shadow: none !important;
         }
-        .notifications-wrapper {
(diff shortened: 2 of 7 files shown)


captainsafia (changes requested):
Thanks for opening this PR! Given that we've recently moved to styled-components, can you try fixing this the other way around, by bringing the media query style on the `Bar` styled-component defined in this file. Essentially something like this

```
export const Bar = styled.div`
   ...
   @media print {
       display: none;
    }
`
```

captainsafia:
This will also close https://github.com/nteract/nteract/issues/4346 when complete.

jarmokivekas:
Thanks for the pointer @captainsafia. Having not used styled-component before I didn't realize this was an option. I've removed the old rule for `.status-bar`, which is not used anywhere, and replaced with a rule in `Bar` itself.

This way the export looks as expected, with no staus bar:
image

jdetle:
There are a few more items listed here that also need to be addressed at some point: 
```
        .notifications-wrapper {
          display: none !important;
        }
        .cell-toolbar {
          display: none !important;
        }
        .cell-creator {
          display: none !important;
        }
        .cell.focused {
          border: none;
          background: 
		}
```

jdetle (changes requested):
I left a comment, if you could address the rest of the changes that need to be made it would be much appreciated.

jarmokivekas:
- `.notification-wrapper` , `.cell-toolbar` , `.cell-toolbar-mask`, `.cell-creator`, classes is not used anywhere anymore.
- notification popups behave well, and aren't visible in pdf exports
- cell toolbars are shown on the active/selected cell.
- active code cell has darker background color in the cell execution order number (left silde of the cell)
image

jarmokivekas:
The css classes `notifications-wrapper`, `cell-toolbar`, `cell-creator`, `draggable-cell`, `cell-drag-handle`, `cell-toolbar-mask`, `invisible` aren't used by anything anymore, so I removed them.

- notification popups and the cell creator buttons were already well behaved, and aren't visible in pdf exports (even if you try to deliberately export pdf while they are visible in the application).
- cell toolbars are shown when a cell is focused, I added `@media print` rule to the component to hide them.
- selected code cell has darker background in the left silde of the cell (area with the `[ ]`)

PDF export in current master: 

image


These rules used to deal with the active cell color during print, but don't anymore (obsolete css rules):
https://github.com/nteract/nteract/blob/f63a01c4067b05541ebddd3932345d7102ea3e5b/applications/desktop/static/index.html#L292-L298

 It's not obvious to me how to fix the focused cell color in the components.

---

I also noticed that the z-order of the notification wrappers and the cell toolbar aren't configured quite right, but didn't look into it yet. Althoug it's not a problem when exporting, since both are hidden in the PDF.

Screenshot from 2019-05-06 19-46-16

jdetle:
> It's not obvious to me how to fix the focused cell color in the components.

Maybe you can wrap that in `@media only screen`?

jarmokivekas:
I think figured out the color of the focused Cell components.

With current changes:
- [x] no notification popups
- [x] no cell creator buttons
- [x] focused/hovered cell `[ ]` prompt color is correct
- [x] cell toolbars are hidden
- [x] no status bar displayed

Test notebook and exported pdf:

Screenshot from 2019-05-06 23-16-13

I'm happy with the the way the pdf is exported now. Comments?

jarmokivekas:
There are tests  in `packages/notebook-app-component/__tests__/toolbar.spec.tsx` that fail because of changes in this PR. I guess the new toolbar no longer matches the snapshot the tests compare it against? If someone can give some pointers on how to update the failing tests, it would be much appreciated.

_Edit_: I think I figured it out. So many new frameworks...

captainsafia:
@jarmokivekas It looks like you figured out how to update the tests. Sorry I couldn't respond sooner. Is this PR good for a re-review?

jarmokivekas:
Yes, this PR is good for a re-review @captainsafia @jdetle

jdetle (approved):

Issue #4134, opened 2019-01-29
Markdown editor styling


rgbkrk:
The codemirror editor is missing the prompt buffer on markdown cells.

![screen shot 2019-01-28 at 4 21 08 pm](https://user-images.githubusercontent.com/836375/51875167-bd157c80-2318-11e9-8381-b0833085c6e5.png)

willingc:
@rgbkrk I'm not sure what needs to addressed in this issue. Do we want the markdown cell input sentence to be in line with the start of print and hey above it?

rgbkrk:
Yeah, somehow we're missing the `<PromptBuffer />` component styling that was on the left.

In the past all the text would be able to be read going down the page following an imaginary line:

![screen shot 2019-02-11 at 9 10 09 am](https://user-images.githubusercontent.com/836375/52580266-da4c4f80-2ddc-11e9-8bd1-946a57b5a95f.png)

willingc:
Perfect. Thanks for the clarification @rgbkrk

jarmokivekas:
The prompt buffer in markdown cells would also have practical use. Currently the part of the text editor where the prompt buffer should be is a drag handle for the cell, and you can't position the text editor cursor to the left edge of the editor by clicking.

![image](https://user-images.githubusercontent.com/3074453/57274578-5f759680-70a4-11e9-9c14-e113048ad9f2.png)

jarmokivekas:
Closing as fixed by #4396

willingc:
Awesome! Thanks again @jarmokivekas

PR #4396, opened 2019-05-07
fix missing PromptBuffer in markdown cells

jarmokivekas:
This PR is a fix to #4134. 

- [x] focused markdown cells can be dragged by dragging from the prompt area
- [x] navigation using a mouse work as expected in the markdown editor

The cause for  #4134 is the incorrect rendering of this prompt:

https://github.com/nteract/nteract/blob/d0fd5ba0cef3080b4ee1a095c9d5cdc739339863/packages/notebook-app-component/src/markdown-preview.tsx#L170

The `<Prompt>` component, and by extension `<PromptBuffer>`, doesn't behave exactly according to the documentation comments in `interface PromptProps`. With the changes in this PR a `<Prompt blank>` and `<PromptBuffer>` are actually rendered, but the prompt has no text/`[ ]` in it, which is the expected behavior according to comments in the `PromptProps` interface. On current master, the prompt doesn't render any elements at all, which is the reason for #4134.

Changes in this PR conform to the comments in this interface:
https://github.com/nteract/nteract/blob/4967874e56669131352958399d7478f0adcc00c7/packages/presentational-components/src/components/prompt.tsx#L5

---

Only the prompt for active markdown cells is changed by this PR:

**Before**
Screenshot from 2019-05-07 10-31-25

**After**

Screenshot from 2019-05-07 10-00-55

---

I also removed a CSS rule for `color` because it was defined twice for the `Prompt` component.

### DIFF
--- a/packages/presentational-components/src/components/prompt.md
+++ b/packages/presentational-components/src/components/prompt.md
@@ -1,7 +1,7 @@
 The `Prompt` component is typically a child of the `Input` component and serves two purposes.
 First, it serves as a visual indicator of input entry for end users. Secondly, it allows us to display relevant information about the state of a cell to the end user.
 
-```js
+```jsx static
 import { Prompt } from "@nteract/presentational-components"
 ```
 
--- a/packages/presentational-components/src/components/prompt.md
+++ b/packages/presentational-components/src/components/prompt.md
@@ -1,7 +1,7 @@
 The `Prompt` component is typically a child of the `Input` component and serves two purposes.
 First, it serves as a visual indicator of input entry for end users. Secondly, it allows us to display relevant information about the state of a cell to the end user.
 
-```jsx static
+```js
 import { Prompt } from "@nteract/presentational-components"
 ```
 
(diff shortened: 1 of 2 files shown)


willingc:
Thanks @jarmokivekas for the PR. We appreciate it and will review it as soon as possible. Thanks.

jarmokivekas:
I should maybe point out in this PR that this is not only a cosmetic change but fixes some unwanted behaviour caused by the missing prompt buffer. That is the real reason for this PR. I explained it in the issue: 

https://github.com/nteract/nteract/issues/4134#issuecomment-489918073 
> The prompt buffer in markdown cells would also have practical use. Currently the part of the text editor where the prompt buffer should be is a drag handle for the cell, and you can't position the text editor cursor to the left edge of the editor by clicking.
>

BenRussert (approved):

BenRussert:
Nice catch!

willingc:
Thanks @jarmokivekas for the PR and @BenRussert for the review :smile:

PR #4405, opened 2019-05-11
Use cardinality to get the best default for the hierarchical view

### LINKED ISSUE #4374: Data Explorer: No default nesting category for Dendrogram chart
Data Explorer

**Describe the bug**
When I select the Dendrogram chart type, there is no default nesting category selected, so there is no visualization shown until the user selects a nesting category. This is a confusing user experience.

**To Reproduce**

1. Go to [here](https://t.co/AoPtdG6BFM) to load the "Happiness" example notebook
2. Click on the Dendrogram icon on the Data Explorer toolbar
4. See there is no visualization

**Expected behavior**
I expect to see a visualization when I select the Dendrogram chart type.

emeeks:
This adds simple cardinality calculation in order to better guess at which dimension should be used for the hierarchical default. Cardinality could be used elsewhere to improve first guesses for other views.
Resolves #4374 but also lays a primitive groundwork for decorating dims and metrics with metadata to help with automatic chart generation & annotation.

### DIFF
--- a/packages/data-explorer/src/index.tsx
+++ b/packages/data-explorer/src/index.tsx
@@ -372,8 +372,7 @@ class DataExplorer extends React.PureComponent<Partial<Props>, State> {
       barGrouping,
       colors,
       primaryKey,
-      data: stateData,
-      largeDataset
+      data: stateData
     } = { ...this.state, ...updatedState };
 
     if (!this.props.data && !this.props.metadata && !this.props.initialView) {
@@ -659,7 +658,7 @@ const DataExplorerDefault: React.FunctionComponent<Props> & {
   return (
     <DataExplorer {...props}>
       <Viz />
-      <Toolbar largeDataset={false} />
+      <Toolbar />
     </DataExplorer>
   );
 };
--- a/packages/data-explorer/src/index.tsx
+++ b/packages/data-explorer/src/index.tsx
@@ -62,6 +62,7 @@ export interface Props {
 }
 
 interface State {
+  largeDataset: boolean;
   view: View;
   colors: string[];
   metrics: Dx.Field[];
@@ -256,7 +257,7 @@ class DataExplorer extends React.PureComponent<Partial<Props>, State> {
           field.type === "boolean" ||
           field.type === "datetime"
       )
-      .map(field => ({ ...field })) as Dx.Dimension[];
+      .map(field => ({ ...field, cardinality: 0 })) as Dx.Dimension[];
 
     // Should datetime data types be transformed into js dates before getting to this resource?
     const data = props.data.data.map((datapoint, datapointIndex) => {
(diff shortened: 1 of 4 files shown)


packages/data-explorer/src/index.tsx
```diff
@@ -272,6 +274,22 @@ class DataExplorer extends React.PureComponent<Partial<Props>, State> {
...
 
+    const cardinalityHash: { [key: string]: { [key: string]: true } } = {};
+    dimensions.forEach(dim => {
+      cardinalityHash[dim.name] = {};
+      data.forEach(datapoint => {
+        const dimValue = datapoint[dim.name];
+        cardinalityHash[dim.name][dimValue] = true;
+      });
+
+      dim.cardinality = Object.entries(cardinalityHash[dim.name]).length;
```
BenRussert: I thought maybe a util function would be simpler here and found we already have the [`d3-collection`](https://github.com/d3/d3-collection#sets)  dep.

What do you think of:

```js
dimensions.forEach(dim => {
  dim["cardinality"] = d3.set(data, d => d[dim.name]).size();
});


or similar?

```

packages/data-explorer/src/index.tsx
```diff
@@ -272,6 +274,22 @@ class DataExplorer extends React.PureComponent<Partial<Props>, State> {
...
+    });
+
+    const selectedDimensions = dimensions
+      .sort((a, b) => a.cardinality - b.cardinality)
+      .filter((data, index) => index === 0)
+      .map(dim => dim.name);
```
BenRussert: Maybe worth another `d3` utility `max` etc.

packages/data-explorer/src/index.tsx
BenRussert: Is it crazy to worry if some ugly datasets out there will cause perf issues when loading? If so, we could consider only doing the cardinality calculation if it is needed i.e. if `selectedDimension` is empty.

BenRussert (commented):
Thanks @emeeks! Just a few questions/comments below.

Side note: 
In the future, I see benefit in moving data processing and related metadata enrichment logic to it's own class or set of functions outside the component (or at least out of the constructor). 

One benefit of not processing in the constructor is that a new dataset could be passed as props and everything would sync up without mounting coming into play. Also, easier testing etc.

packages/data-explorer/src/index.tsx
emeeks: I tested it on 50k line datasets and since I don't see Data Explorer being able to do much with larger sets, especially in hierarchical, I could see limiting it. In fact, I think if you have a dataset above 50k then you should only get access to summary view, scatterplot, line chart and maybe parallel coordinates, since network/hierarchical and bar would be useless anyway.

emeeks: I'm less convinced about the use of d3.set in that situation, is there a reason not to use `Set`? I haven't done any perf tests, though, so we should probably use whatever works fastest.

BenRussert: `Set` would be good too. I was thinking a packaged/utility solution like this would be easier to read down the line if nothing else.

packages/data-explorer/src/index.tsx
emeeks: How would you use `max` here? Also this reminds me that we could use D3's (or the equivalent) `extent` and n-tiles to do the same thing we're doing with dimensions but with metrics.

BenRussert: Maybe I meant `min`. I may be reading wrong, is this part taking he name of the lowest cardinality dimension?

packages/data-explorer/src/index.tsx
emeeks: I ended up lowering the bar to 5000 which is still absurd for a bar chart but fine-ish for a network or hierarchical diagram, maybe, in some cases.

packages/data-explorer/src/index.tsx
emeeks: Oh yeah, `min`. This is taking the lowest cardinality dimension but I'm not sure how the `min` is going to make this code cleaner. I feel like `d3-array` utilities aren't very commonly known and would tend to prefer vanilla js if possible.

Issue #4408, opened 2019-05-16
Cell default height too big in nteract web


giux78:
**Application or Package Used**
nteract web, nteract on jupyter https://github.com/nteract/nteract/blob/master/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx

**Describe the bug**
The cell default height is too big with too many lines. When for example I want to create a chart of a dataframe normally i use cell of one or few lines.

**To Reproduce**
All cells on web apps are affected both new and old.

**Screenshots**
![Screen Shot 2019-05-14 at 08 02 10](https://user-images.githubusercontent.com/2806256/57836131-5ed5b200-77c0-11e9-8b93-876a273ac417.png)

**Expected behavior**
Cell should be flexible on the numbers of lines starting from one line

![Screen Shot 2019-05-16 at 09 54 40](https://user-images.githubusercontent.com/2806256/57836284-a65c3e00-77c0-11e9-966b-2dc582ead8a7.png)

- OS: [e.g. macOS]
- Browser [all]
- Version [e.g. last]

jameskoch:
It looks like this regressed in https://github.com/nteract/nteract/commit/7fcb71fcbe1e81f35805e7229174fd2b20119b14#diff-ac4acad2081a1ced22ecae1d968c8ca7L9

PR #4409, opened 2019-05-16
Bugfix issue #4408 - Update CSS overrides order

and0111:
During my last `nteract/master` alignment merge on my fork, I noticed the notebook's cells that contains even a few lines (1-4 lines) was taking almost half of my screen height. Since I noticed also the notebook's cells that contains a lot of lines (more than 15 lines) had a scrollbar!? I ended up by thinking that may was a problem. With the help of @giux78, I am opening a PR to solve the unexpected described _UI_ behavior.

**Cause**
`@nteract/styles/editor-overrides.css` styles classes get overridden by "codemirror" imports

**Solution**
`import "@nteract/styles/editor-overrides.css"` has to be placed down "codemirror" imports because as the name suggest they are and act as overrides for "codemirror" styles so overrides should be placed at the end of these.
This because _Webpack_ is going to parse all of these CSS imports in the same order as written in [applications/jupyter-extension/nteract_on_jupyter/app/index.tsx](https://github.com/nteract/nteract/blob/master/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx)

- Updates vendor modules imports order in "nteract_on_jupyter"
entrypoint to get CSS overrides work as expected
- Fix issue #4408 

**Before Fix**
BeforeFix

**After Fix**
AfterFix

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
@@ -6,11 +6,11 @@
 import "@blueprintjs/core/lib/css/blueprint.css";
 import "@blueprintjs/select/lib/css/blueprint-select.css";
 
+import "codemirror/addon/hint/show-hint.css";
+import "codemirror/lib/codemirror.css";
 import "@nteract/styles/app.css";
 import "@nteract/styles/editor-overrides.css";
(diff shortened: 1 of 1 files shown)


BenRussert (approved):
Awesome, great job debugging this!

PR #4414, opened 2019-05-20
Add support for SaveAs action and epics

captainsafia:
This PR:
- Adds action types and creators for a SaveAs action
- Adds a saveAsContentEpic for PUTing new files to the Jupyter server
- Refactors shared logic between the saveContentEpic and the saveAsContentEpic
- Adds tests for the saveAsContentEpic
- Fixes a type error in the `stringifyNotebook` function

### DIFF
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -14,10 +14,19 @@ import { catchError, map, mergeMap, switchMap, tap } from "rxjs/operators";
 
 import * as actions from "@nteract/actions";
 import * as selectors from "@nteract/selectors";
-import { AppState, ContentRef } from "@nteract/types";
+import {
+  AppState,
+  ContentRef,
+  DirectoryContentRecordProps,
+  DummyContentRecordProps,
+  FileContentRecordProps,
+  NotebookContentRecordProps
+} from "@nteract/types";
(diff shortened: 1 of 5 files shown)


packages/epics/src/contents.ts
```diff
@@ -454,3 +475,69 @@ export function saveContentEpic(
...
+        action: actions.SaveAs
+      ):
+        | Observable<actions.SaveAsFailed | actions.SaveAsFulfilled>
+        | Observable<never> => {
+        const state = state$.value;
+
+        const host = selectors.currentHost(state);
+        if (host.type !== "jupyter") {
+          // Dismiss any usage that isn't targeting a jupyter server
+          return empty();
```
BenRussert: Is it important to that these failures are silent instead of returning some sort of failure action?

BenRussert (approved):
This looks great. The tests are really helpful!

packages/epics/src/contents.ts
captainsafia: Good point! I was thinking about this as I was writing these. I think we should opt for explicit error actions instead of doing nothing. I'll fix this up later.

rgbkrk:
Thanks for fixing this up.

PR #4416, opened 2019-05-20
Fix cm editor props

BenRussert:
- Call `fullOptions` before trying to spread into options object
- Update test to guard against this issue

### DIFF
--- a/packages/editor/__tests__/editor.spec.tsx
+++ b/packages/editor/__tests__/editor.spec.tsx
@@ -141,10 +141,10 @@ describe("editor.hint CodeMirror callback", () => {
 
 describe("Editor", () => {
   it("handles cursor blinkery changes", () => {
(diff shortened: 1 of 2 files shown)


packages/editor/__tests__/editor.spec.tsx
```diff
@@ -141,10 +141,10 @@ describe("editor.hint CodeMirror callback", () => {
 
 describe("Editor", () => {
   it("handles cursor blinkery changes", () => {
-    const editorWrapper = mount(<Editor cursorBlinkRate={530} />);
+    const editorWrapper = mount(<Editor cursorBlinkRate={531} />);
```
BenRussert, line 144: The updated test here fails before this fix. Since 530 is the default, the test was passing anyway.

rgbkrk (approved):

rgbkrk:
Thanks Ben!

PR #4418, opened 2019-05-21
Increase jest timeout limit for kernel launch tests

BenRussert:
- Otherwise test often fails on CI due to unrelated timeout


rgbkrk:
I suppose this is alright for now, though I think the kernel tests should be moved to integration tests. cc @captainsafia

PR #4431, opened 2019-06-05
Nteract BlueprintJS NotebookMenu Implementation

and0111:
During a weekly meeting my attention was caught on the possibility of refactoring 
_NotebookMenu_ _UI_ using _BlueprintJS_ components.

We started by rewriting the `render` method using _BlueprintJS_ components but without making too aggressive changes in comparison to the previous render implementation.
Once developed a working `render`, we created a new click handler on top of oldest one in order to make the previous works as before.
We also removed some unused parts like the `StickyMenu` in order to a bit clean up the package sources.

Last and most important, we had to get tests pass. This took a little while during development process since replacing `render` method breaks all _Enzyme_ selectors that are used to simulate, check the correct click handling.
Now we are glad to propose a possible fully working solution.

**This PR will:**

- Reimplements notebook-menu `render` method using BlueprintJS
components

- Implements handleActionClick method in order to reuse old
`handleClick` method istead of reimplement completely

- Adds `MENU_ITEM_LABELS` in "constants.ts"

- Updates test snapshot and "calls appropriate handlers on click" test 

- Removes "rc-menu" dependency with all related sources like
"notebook-menu.css", "styles.tsx" and rc-menu's types

- Removes `MENUS` in "constants.ts"

- Removes "StickyMenu" component and "styled" import from
`PureNotebookMenu`

- Updates "@nteract/connected-components" package version from "6.2.1" to
"6.3.0"

**Before Merge** (_rc-menu_ implementation)
old_rc-menu_implementation

**After Merge** (_BlueprintJS_ implementation)
new_BlueprintJS_implementation


rgbkrk:
This is super fantastic to be able to move towards. I look forward to reviewing this in earnest.

applications/jupyter-extension/nteract_on_jupyter/app/index.tsx
```diff
@@ -16,9 +16,6 @@ import urljoin from "url-join";
 
 // Local modules
 import { JupyterConfigData, readConfig } from "./config";
-// Until we're switched to blueprint for the menu, we have our own custom css
-// for the rc-menu style menu
-import "./notebook-menu.css";
```
rgbkrk, line 21: 👍

packages/connected-components/__tests__/notebook-menu.spec.tsx
```diff
@@ -25,10 +25,30 @@ describe("PureNotebookMenu ", () => {
...
+        SET_THEME_LIGHT,
+        SET_THEME_DARK,
+        SAVE_NOTEBOOK,
+        OPEN_ABOUT,
+        INTERRUPT_KERNEL,
+        RESTART_KERNEL,
+        RESTART_AND_CLEAR_OUTPUTS,
+        RESTART_AND_RUN_ALL_OUTPUTS,
+        KILL_KERNEL
+      } = MENU_ITEM_LABELS;
```
rgbkrk, line 50: Reminder for myself within the review -- I wonder if we can get rid of the indirection around the labels.

packages/connected-components/__tests__/notebook-menu.spec.tsx
```diff
@@ -53,100 +73,84 @@ describe("PureNotebookMenu ", () => {
...
         currentContentRef: "fake-content-ref",
-        currentKernelRef: "fake-kernel-ref",
-
-        // menu props, note that we force all menus to be open to click.
-        defaultOpenKeys: Object.values(MENUS)
+        currentKernelRef: "fake-kernel-ref"
       };
-      const wrapper = mount(<PureNotebookMenu {...props} />);
+      const wrapper = shallow(<PureNotebookMenu {...props} />);
+      const eventMock = { preventDefault: () => null };
```
rgbkrk, line 79: It would be good to denote why we have to do this for all the calls to `simulate` below.

packages/connected-components/src/index.tsx
```diff
@@ -7,4 +7,4 @@ export { default as HeaderEditor } from "./header-editor";
 export { default as NotebookMenu } from "./notebook-menu";
 export { default as ModalController } from "./modal-controller";
 
-export { GlobalMenuStyle } from "./notebook-menu/styles";
+// export { GlobalMenuStyle } from "./notebook-menu/styles";
```
rgbkrk, line 10: Might as well just delete this.

packages/connected-components/package.json
```diff
@@ -14,7 +14,7 @@
     "@nteract/commutable": "^7.1.2",
     "@nteract/core": "^10.5.0",
     "@nteract/types": "^4.2.1",
-    "rc-menu": "6.2.11",
+    "@blueprintjs/core": "^3.7.0",
```
rgbkrk, line 17: 👍

packages/connected-components/src/notebook-menu/PureNotebookMenu.tsx
```diff
@@ -0,0 +1,577 @@
...
+              >
+                {currentKernelspecs &&
+                  currentKernelspecs.byName
+                    .keySeq()
+                    .map((name: string) => [
+                      name,
+                      currentKernelspecs.byName.getIn([name, "displayName"])
+                    ])
+                    .toArray()
+                    .map(([name, displayName]: string[]) => {
```
rgbkrk, line 544: This all makes me think we should normalize this before we pass it into the menu. I'm assuming this came from the old code though and so doesn't need to be part of this refactor. Is that correct that it came from the old menu code?

packages/connected-components/src/notebook-menu/PureNotebookMenu.tsx
```diff
@@ -0,0 +1,577 @@
...
+            contentRef: currentContentRef,
+            kernelSpecName: args[0]
+          });
+        break;
+      case PUBLISH_TO_BOOKSTORE:
+        onPublish && onPublish({ contentRef: currentContentRef });
+        break;
+      default:
+        console.log(`unhandled action: ${action}`);
+    }
```
rgbkrk, line 291: Recalling how we used to do this, I'm really happy with this pattern. It makes it cleaner checking for if a given `onAction` is even set. There's probably more to do to clean up indirection and make the menu more compositional (instead of being one big built component), but I think this works well. 😄

rgbkrk (approved):
I'm a fan, this at least moves us in the right direction off of `rc-menu` so that we're relying on less external UI libraries.

packages/connected-components/src/notebook-menu/constants.ts
```diff
@@ -29,18 +29,29 @@ export const MENU_ITEM_ACTIONS = {
...
+  SET_CELL_TYPE_MARKDOWN: "To Markdown",
+  TOGGLE_EDITOR: "Notebook Header",
+  SET_THEME_LIGHT: "Light",
+  SET_THEME_DARK: "Dark",
+  EXECUTE_ALL_CELLS: "Run All Cells",
+  EXECUTE_ALL_CELLS_BELOW: "Run All Cells Below",
+  CREATE_CODE_CELL: "Code Cell",
+  CREATE_MARKDOWN_CELL: "Markdown Cell",
+  CLEAR_ALL_OUTPUTS: "Clear All Outputs",
+  // CHANGE_KERNEL: "change-kernel",
```
captainsafia, line 49: Is there a reason this is commented out?

captainsafia (approved):

packages/connected-components/__tests__/notebook-menu.spec.tsx
and0111: This because `handleActionClick` uses `event.preventDefault()` just above calling older handler within the key. So since `handleActionClick` expect an event as first argument test breaks on calling `preventDefault` of undefined. Without going deep by thinking too much about the cause, I quickly troubleshooted this by using a mock for that event.
May now can be removed along with the `e.preventDefault()` in `handleActionClick`..

packages/connected-components/src/notebook-menu/PureNotebookMenu.tsx
and0111: Yeah, the "kernelspecs" loading logic comes entirely from old implementation with only the difference of using a && operator instead of a ternary, I do not wanted to be much aggressive on changes since the purpose of this PR born from the idea of only replacing the UI components used in "notebook-menu" with the _BlueprintJS_ ones.

packages/connected-components/src/notebook-menu/constants.ts
and0111: Hi @captainsafia :smile: 
The reason is that `MENU_ITEM_LABELS` are used mainly to get a straightforward and clear way of selecting "MenuItems" in test, but of course they are reused in render method. Thinking about the way I used those labels, I leaved as reminder for future a comment about the label correspondent action which does not have been tested. So the commented line act as only a reminder for next test/render updates. I would have done it better if I written just a bit more on that comment :+1:

packages/connected-components/src/notebook-menu/PureNotebookMenu.tsx
and0111: The first approach that I wanted to implement was realize a dedicated mapping function that takes an array of objects that contains only specifically needed props for a "MenuItem" (i.e. text or icon) and map each of those item object as a "MenuItem". So I chose not to be so aggressive in implementation compared to the previous one. Any way, refactors always come. I am verry happy about this solution since it is the first huge contribution that involves me as Nteract contributor :+1: :smile:

rgbkrk:
💥 

🎉  Welcome to the nteract party @and0111! 🎉

PR #4433, opened 2019-06-06
Data Explorer: Update react-table styles

### LINKED ISSUE #4401: Data Explorer: Unexpected high contrast table styles
Data Explorer

**Describe the bug**
The styles of the Data Explorer changed in #4339. The light theme now has very high contrast for striped rows, which looks unpleasant in my opinion. I understand there are contrast guidelines for usability, but this seems to be more contrast than necessary. The left-most column also has a border that I'm not sure is needed.

image

Expected:
image

Actual:
image

dylanross3:
This updates the background color of even table rows in the Data Explorer table view to the previous background color before #4339.

The color set by react-table css is actually `rgba(0,0,0,.03)` but a transparent color doesn't work well when other columns scroll underneath it (since it is fixed). `#f7f7f7` seems to be equivalent.

Fixes #4401.

screenshot

### DIFF
--- a/packages/data-explorer/src/css/index.tsx
+++ b/packages/data-explorer/src/css/index.tsx
@@ -22,7 +22,7 @@ export default styled.div<ThemeProps>`
 
   .ReactTable.-striped .rt-tr.-even > div {
     color: ${props => (props.theme === "dark" ? "#bbb" : "#111")};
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):
Awesome, thank you for addressing this!

BenRussert:
Thanks!

PR #4434, opened 2019-06-07
Allow multiple output_types

BenRussert:
This allows a single custom Output child component to handle multiple
output_types. e.g. ['display_data', 'execute_result']

A large part of the motivation for this is the upgrade work @wadethestealth took the lead on over on Hydrogen. [See this comment](https://github.com/nteract/hydrogen/pull/1655/files#r291652015) for more

### DIFF
--- a/packages/outputs/__tests__/output.spec.tsx
+++ b/packages/outputs/__tests__/output.spec.tsx
@@ -140,3 +140,60 @@ describe("Full Outputs usage", () => {
     expect(chosen.text()).toEqual("This is some HTML that WILL render");
   });
 });
+
+describe("Output with an array of output_types", () => {
+  const displayOutput = {
+    output_type: "display_data",
+    data: {
+      "text/html": "<p>This is some HTML that <b>WILL</b> render</p>",
+      "text/plain": "This is some plain text that WILL NOT render"
+    },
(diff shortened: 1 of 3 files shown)


BenRussert:
This would allow our main display component on hydrogen to look like this, specifically the `DisplayOrExecuteResult` component below:


```jsx
 function Display(props) {
  return (
    <Output output={toJS(props.output)}>
      <DisplayOrExecuteResult expanded>
        <Vega3 />
        <Vega2 />
        <PlotlyTransform />
        <VegaLite2 />
        <VegaLite1 />
        <Media.Json />
        <Media.JavaScript />
        <Media.HTML />
        <Media.Markdown />
        <Media.LaTeX />
        <Media.SVG />
        <Media.Image />
        <Media.Plain />
      </DisplayOrExecuteResult>
      <StreamText expanded />
      <KernelOutputError expanded />
    </Output>
  );
}

```

rgbkrk (approved):
I think this is a reasonable interface for making it easier to support execute_result and display_data at once. 👍

rgbkrk:
I wonder if we should default `DisplayData` to handling `execute_result` as well in this style. Then if someone wants one specific to execute result they can include that instead.

Issue #4397, opened 2019-05-07
Altair plots do not render


mattbit:
**Application or Package Used**
nteract desktop v0.14.2 (Linux AppImage)
altair version 3.0.1
(Note: I get the same with Atom+Hydrogen)

**Describe the bug**
Altair (Vega) plots do not render properly. Instead, I get the following message:
```
<VegaLite 3 object>

If you see this message, it means the renderer has not been properly enabled
for the frontend that you are using. For more information, see
https://altair-viz.github.io/user_guide/troubleshooting.html
```

**To Reproduce**
```python
import altair as alt
from vega_datasets import data

iris = data.iris()

chart = alt.Chart(iris).mark_point().encode(
    x='petalLength',
    y='petalWidth',
    color='species'
)

display(chart)
```

**Expected behavior**

Altair [should be working out of the box](https://twitter.com/nteractio/status/984257546985783297).

captainsafia:
Hello @mattbit! Thanks for opening this issue. I was able to reproduce this issue in the nteract desktop as well.

It appears that there is a bug in our output rendering with respect to Vega outputs that will need to be resolved before the next release.

Thanks again for submitting this bug report.

davidanthoff:
Any update on this? This is also affecting the julia side of things.

davidanthoff:
Sorry to bump again, but it would be really helpful for us to get a vague sense of when a version of nteract might ship that has this fixed. I'm exploring nteract as an option for some workshops that I'm giving, but this is blocking for that. So, just a short update on whether a fix might make it into the next release would already be fantastic :) Thanks!

Issue #4411, opened 2019-05-16
Julia vega-lite plot doesn't show


davidanthoff:
The notebook [here](https://gist.github.com/davidanthoff/b9ff6e31ac64788c1204e90cdda64ac1) shows correctly with a plot in jupyter notebook, jupyter lab and on github. But in nteract, I don't see a plot.

If you look into the notebook, it has the plot as a vega-lite JSON spec, a PNG or SVG MIME bundle, so I would assume that one of them should work :)

mattbit:
May be related to the problem with the output rendering in #4397.

davidanthoff:
Yes, that might well be the same issue!

captainsafia:
Thanks for your patience folks! Hopefully #4442 will be merged soon and we can close this and related issues.

CrystallineCat:
Should be fixed by 97721aa.

PR #4442, opened 2019-06-11
Allow vega v4/v5 and vega-lite v3 mime-types

CrystallineCat:
Attempt at fixing #4397.

Unfortunately causes an exception in vega-dataflow:
```
TypeError: Cannot add property Symbol(vega_id), object is not extensible
    at setid (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-dataflow/src/Tuple.js?:44)
    at ingest (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-dataflow/src/Tuple.js?:59)
    at Object.pulse (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-dataflow/src/ChangeSet.js?:86)
    at View.pulse (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-dataflow/src/dataflow/update.js?:74)
    at View.ingest (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-dataflow/src/dataflow/load.js?:25)
    at Context.add (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-runtime/src/context.js?:68)
    at Context.transform (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-runtime/src/context.js?:118)
    at __webpack_exports__.default (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-runtime/src/operator.js?:20)
    at eval (webpack:////home/mk/Projects/Webstorm/nteract/node_modules/vega-runtime/src/dataflow.js?:27)
    at Array.forEach (<anonymous>)
```

Will take a closer look at it tomorrow.


mattbit:
I confirm this fixes #4397 for me.

davidanthoff:
Awesome, can't wait to see this shipping!!

One question: when this renders an older vega or vega-lite MIME type, does it still use the latest version of the vega or vega-lite code? I.e. when there is a vega3 MIME type, does it use vega 5 to render it, or does it use vega3 to render that? I seems to me that the latter would be a better option, i.e. always use the latest minor version that corresponds to a given MIME type.

CrystallineCat:
Got VegaLite 1/2/3 rendering with the appropriate compiler and Vega 2/3/5 respectively:



Haven't tried Vega itself yet, and Vega 4 would render with Vega 5 at the moment.

davidanthoff:
@CrystallineCat could you also try to open the notebook I attached to https://github.com/nteract/nteract/issues/4411 and see whether that displays properly with your patch? It includes a vega-lite 2 plot that was created with the julia package [VegaLite.jl](https://github.com/queryverse/VegaLite.jl). I'll try to also create a notebook that embeds a vega-lite 3 spec created with that package in a moment.

davidanthoff:
[Here](https://gist.github.com/davidanthoff/21fc3db8b040fe945eaf4d5066787ce5) is a notebook file with an embedded vegalite3 spec from the julia side of things. Would be fantastic if you could try that one as well!

CrystallineCat:
@davidanthoff thank you for your examples. Both of the notebooks render fine in my current version:

davidanthoff:
Great!

captainsafia:
@CrystallineCat @davidanthoff Is this pull request ready for review or are there more changes incoming?

davidanthoff:
I don’t know, I’m just a very happy observer from the outside who can’t wait to see this shipping :)

CrystallineCat:
@captainsafia thanks for asking! I'm not entirely happy with the PR yet, but also a bit stuck. But it does fix the bug (on the desktop at least, I didn't touch `nteract_on_jupyter`, as I have never used it), so I should probably stop letting the perfect be the enemy of the good. :)

The things I'm not happy about revolve around the mechanism for supporting old versions of vega. I'd like to have a better solution than making a new repository for each one. I tried to setup `yarn`/`rollup` to make a single repository for all versions locally, but I don't understand the JS module system enough to make it work. :(

If we keep the current system, I'd still have to duplicate https://github.com/CrystallineCat/vega-embed-v3 into the @nteract namespace and update the reference (I think I have the permissions, but I didn't want to make a repository there without asking). It embeds `vega@3` and `vega-lite@2`, but uses `vega-embed@4`, so I'm not happy with the name, but I chose to use the `vega` version number, rather than vega-embed, as `vega-embed@4` appears to be completely neutral regarding the vega/vega-lite versions. (I just wish it'd let me hand the appropriate vega/vega-lite objects to the embed function, rather than importing them itself...)

Thank you again! :)

wadethestealth:
This PR closes #4481 and will help hydrogen contributors to close nteract/hydrogen#1700

captainsafia:
> The things I'm not happy about revolve around the mechanism for supporting old versions of vega. I'd like to have a better solution than making a new repository for each one. I tried to setup yarn/rollup to make a single repository for all versions locally, but I don't understand the JS module system enough to make it work. :(

Thanks for bring this up. Ideally, we would have a single repository for all versions of Vega. I agre that we shouldn't let perfect be the enemy of good and approach this incrementally and figure out how we want to manage Vega dependencies as part of another PR.

> If we keep the current system, I'd still have to duplicate https://github.com/CrystallineCat/vega-embed-v3 into the @nteract namespace and update the reference (I think I have the permissions, but I didn't want to make a repository there without asking). It embeds vega@3 and vega-lite@2, but uses vega-embed@4, so I'm not happy with the name, but I chose to use the vega version number, rather than vega-embed, as vega-embed@4 appears to be completely neutral regarding the vega/vega-lite versions. (I just wish it'd let me hand the appropriate vega/vega-lite objects to the embed function, rather than importing them itself...)

You should have permissions. If not, I can set you up with this. I think setting up a new repository is quick to do so we can implement this as a temporary solution while we figure out a simpler way to manage Vega versions.

packages/transform-vega/src/external.ts
```diff
@@ -0,0 +1,43 @@
+import { promisify } from "util";
+import { MEDIA_TYPES, VegaMediaType } from "./mime";
+
+export interface VegaOptions {
+  renderer: "canvas" | "svg";
+}
+
+/** Call the external library to do the embedding. */
+export async function doEmbedding(
```
captainsafia: Nit: Can we call this method `embed`?

packages/transform-vega/src/external.ts
```diff
@@ -0,0 +1,43 @@
...
+  spec: Readonly<{}>,
+  options: Partial<VegaOptions> = {},
+): Promise<any> {
+  const version = MEDIA_TYPES[mediaType];
+  const defaults = {
+    actions: false,
+    mode: version.kind,
+  };
+
+  if (version.vegaLevel <= 2) {
```
captainsafia: Nit: We might want to use a `switch` statement here.

applications/desktop/src/notebook/index.tsx
```diff
@@ -86,8 +86,11 @@ const store = configureStore({
       transforms: makeTransformsRecord({
```
captainsafia, line 86: With regard to your comment about nteract_on_jupyter, the same configuration will need to be made in the transforms. I don't know if we'll need to update the Webpack config in nteract_on_jupyter.

captainsafia (approved):
Looks good! Thanks for creating the tests. I left a few nit comments in line.

I think good next steps after merging this are:
- Add support for different Vega views in nteract_on_jupyter
- Create repository for vega-embed 4
- Figure out a way to clean up our management of Vega dependencies

applications/desktop/src/notebook/index.tsx
CrystallineCat: I added lines for the new mime types and bumped the version of the transform-vega dependency. It looks like all Vega/VegaLite transforms use NullTransform currently, but I'm not sure if that means "not supported" or "handled somewhere else".

CrystallineCat:
I created  https://github.com/nteract/vega-embed-v3, but can't publish it to npm, so the reference in `package.json` still points to GitHub. Apart from this, this should be ready to merge.

captainsafia:
@CrystallineCat What is your npm username? I can add you to the nteract organization.

Looks like the recent changes caused a failure in the nteract_on_jupyter build. Let me know if you need help debugging this.

CrystallineCat:
@captainsafia I've just joined npm as `crystalline-cat`. Do I need to do anything else but `yarn publish` and answer the questions?

The test failure is due to `vega@3` testing for `fs` thus:
```js
function fs() {
  var fs = typeof require === 'function' && require('fs');
  return fs && isFunction(fs.readFile) ? fs : null;
}
```
This ends up in the bundle and will trigger the configured `webpack.IgnorePlugin` on the `require('fs')`.

I'm unsure what the best way to fix this is; we need Vega to be able to load things from the filesystem for the desktop app, so that either leaves changing the check to be more robust via rollup or changing the webpack config to return undefined (or an empty object) for the require. No idea how difficult either would be right now, nor any idea what the side effects of the latter change would be.

CrystallineCat:
@captainsafia Thanks for adding me! I published the dependency and updated `package.json`. I also added `fs` to the webpack externals for `nteract_on_jupyter`. If I understand correctly, this makes this from a compile time error to a runtime error if anyone actually tries to import `fs` (e.g. tries to access a filesystem path in a vega spec), which should be okay.

Do you think this can be merged as it is?

captainsafia:
Yes, this can be merged as is. It looks like there are some conflicts in the package.json following the latest release. Feel free to merge once these changes are rebased. Thanks for all the hard work on this!

CrystallineCat:
Had a bit of a problem with the rebase, but I think I sorted it out.

Thank you for all the help!

mattbit:
Brilliant! Thank you very much, @CrystallineCat!

davidanthoff:
Hurray! This hasn't shipped, right? Any estimate when we might get a public release that includes this?

captainsafia:
@davidanthoff On Monday!

davidanthoff:
Great!

We have Juliacon starting on Monday, and I'd actually like to show some of this stuff off. In theory if it ships on Monday it should be good enough, but it does make me slightly nervous re timing. If there is a chance that it could ship earlier, it would be great. But I also totally understand that there is an existing shipping plan, so no problem if that doesn't work.

captainsafia:
@davidanthoff We can ship on Sunday but that depends on my/other's availability. Are you planning on having people install nteract or are you including it in your demo? If it is in your demo, you might be able to use a build of nteract off master.

davidanthoff:
It will just be me demoing. Don't worry about it, and I don't want you to ruin your weekend for this :) Lets go with the normal Mon release. If it all works, great, if not, I can just show a screenshot (the whole part would only take a few moments in any case, I don't have a ton of time).

captainsafia:
@davidanthoff  That's good to know. You should be able to get it working via master then. Thanks for sharing nteract in your presentation!

davidanthoff:
@captainsafia do you have a sense when a new version might ship?

willingc:
@davidanthoff Did the vega-lite v3 work for you with the latest release of nteract? I'm running into errors on Binder with v3 not rendering. Wondering if its an nteract issue or a Binder config issue. Thanks. Crossing my fingers that it worked for you. 

I heard JuliaCon was nice from some of the Iodide folks working with Julia :smile: Hope you enjoyed too.

davidanthoff:
Yes, I just tested it with 0.14.5, [VegaLite.jl](https://github.com/queryverse/VegaLite.jl) plots show, and they use the vega-lite 3 MIME type. [Here](https://gist.github.com/davidanthoff/40014c8c7e5261c7fc039fddf250e606) is the example notebook I just created to try.

I ended up not demoing anything live during my juliacon talk, it was too short so I resorted to normal slides. And yes, juliacon was great!

willingc:
Thanks @davidanthoff. I appreciate you checking :smile:

PR #4446, opened 2019-06-13
Add restartWebSocketKernel epic and tests

### LINKED ISSUE #4440: Expose epic for restarting a kernel via Jupyter server API
Currently, our `restartKernel` epic restarts the kernel by killing the kernel then launching the kernel. Amongst other reasons, this has precedence in the fact that we built a desktop-focused API for kernel lifecycle management that assumed that the kernel was a process adjacent to the front-end.

We should expose a `restartKernel` epic that operates under the assumption that the user is is connected to a Jupyter server and sends the `RESTART_KERNEL` action through the `api/kernels/{kernelId}/restart` endpoint.

This epic should live in our [websocket-kernel](https://github.com/nteract/nteract/blob/29879454aa96d972549c51e87e185716010beed6/packages/epics/src/websocket-kernel.ts) file colocated with the other epics that target kernels running in a server environment.

Validations for this new epic should test that kernel state is update appropriately and correct cell execution or clear events are emitted.

Tangent: We should move the `restartKernel` epic to the desktop application and expose just the `restartKernel` that will be created for this issue as part of our SDK.

Another tangent: We should only house context-agnostic (doesn't care if the kernel is running locally or in a container or on the moon) epics in the [kernel-lifecycle](https://github.com/nteract/nteract/blob/29879454aa96d972549c51e87e185716010beed6/packages/epics/src/kernel-lifecycle.ts) file.

captainsafia:
Closes https://github.com/nteract/nteract/issues/4440.

### DIFF
--- a/packages/epics/src/websocket-kernel.ts
+++ b/packages/epics/src/websocket-kernel.ts
@@ -12,8 +12,7 @@ import {
   filter,
   map,
   mergeMap,
-  switchMap,
-  withLatestFrom
+  switchMap
 } from "rxjs/operators";
 
 import * as actions from "@nteract/actions";
@@ -23,6 +22,7 @@ import { createKernelRef } from "@nteract/types";
 import { AppState } from "@nteract/types";
 import { RemoteKernelProps } from "@nteract/types";
 
+import { AjaxResponse } from "rxjs/ajax";
 import { extractNewKernel } from "./kernel-lifecycle";
 
 export const launchWebSocketKernelEpic = (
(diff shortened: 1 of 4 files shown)


packages/epics/src/websocket-kernel.ts
```diff
@@ -322,3 +323,102 @@ export const killKernelEpic = (
...
+            kernelRef,
+            contentRef
+          })
+        );
+      }
+
+      const id = kernel.id;
+
+      return kernels.restart(serverConfig, id).pipe(
+        map(() => {
```
rgbkrk: Since these are returning `of`s, this needs to be a `mergeMap`.

rgbkrk: Otherwise this overall epic here is forming a stream of stream of actions rather than a stream of actions.

captainsafia: This is fixed now!

rgbkrk (approved):
Updates look great, thank you!

PR #4451, opened 2019-06-14
Changing API for Bookstore

### LINKED ISSUE #4450: Bookstore API Change
The bookstore API is changing, see [here](https://github.com/nteract/bookstore/issues/132). In order to keep integration, we need to change what we look for on the call back from bookstore.

Bookstore will no longer send back `bookstore: True` since it is equivalent to getting a `200` back. With this change, we will need to remove the equivalent `bookstore` key in nteract.

N/A

**Additional context**
N/A

stormpython:
Closes #4450.

I removed the `bookstore` key from the payload, and I changed the way I am checking to see if bookstore is available. I have questions around the latter since I am not sure its the best solution.

### DIFF
--- a/applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
+++ b/applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
@@ -44,7 +44,7 @@ export async function main(
     token: config.token,
     origin: location.origin,
     basePath: config.baseUrl,
-    bookstoreEnabled: config.bookstore.enabled,
+    bookstoreEnabled: !!config.bookstore.version,
     showHeaderEditor: false
   });
 
--- a/packages/types/src/entities/hosts.ts
+++ b/packages/types/src/entities/hosts.ts
@@ -9,7 +9,6 @@ import { HostRef } from "../refs";
 
 export interface Bookstore {
   version: string;
-  enabled: boolean;
 }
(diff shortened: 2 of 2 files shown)


applications/jupyter-extension/nteract_on_jupyter/app/bootstrap.tsx
```diff
@@ -44,7 +44,7 @@ export async function main(
     token: config.token,
     origin: location.origin,
     basePath: config.baseUrl,
-    bookstoreEnabled: config.bookstore.enabled,
+    bookstoreEnabled: !!config.bookstore.version,
```
stormpython, line 47: I think there is a better way to do this. However, I am not sure what happens when we make a call to Bookstore and get a 404 back. Do we still get an object? Is that object empty?

Also I know the double bangs are not great to do stylistically. So I am happy to change this once I have a better understanding of what payload I get back.

captainsafia:
@mpacer @stormpython Should this PR still be considered active or can it be closed? Trying to do some clean-up on our PRs page.

stormpython:
I think this can be merged unless @mpacer you see any issues with the PR. That is, if this PR is still relevant to the Bookstore changes. Otherwise, we can close this.

rgbkrk:
I'm going to err on the side of assuming this is good to go.

Issue #4415, opened 2019-05-20
autocomplete with Julia


PaulSoderlind:
in nteract 0.14.2 (on Windows 10), using the Julia 1.1.1 kernel there seems to be no auto-complete. 

To reproduce:
type `ab` and nothing happens (with/without Ctrl-Space).

In contrast, with the python3 kernel, it works.

BenRussert:
I tried debugging this for a bit and didn't find a fix yet. The Julia inspection and completion messaging works in hydrogen with the same kernel so I think the issue is somewhere in the monorepo.

If I subscribe to the kernel observable in the console and log all the kernel messages in nteract I don't see any complete or inspect replies coming through. I do see a time out error message for completion requests.

My guess is we are mistakenly filtering out messages for some reason. The timeout limit seems to be much longer than even a slow reply would take so I don't think thats the issue.

captainsafia:
@BenRussert Is this issue resolved now that #4453 is in?

BenRussert:
@captainsafia unfortunately not. Both inspection and completion are not working. It seems to me that the inspection request messages sent to IJulia are ignored silently.

My guess was it's something about the format of the inspect_request message, but took a pretty close look and found no issues. Inspection works fine on hydrogen, but I'm not sure what we do differently. I was thinking of asking for debug suggestions on the IJulia repo next.

I'm really curious about this one, so let me know if you have debug suggestions or slack me if you want to pair up sometime.

captainsafia:
@BenRussert Got it! Thanks for clarifying. I'd be interested in helping out. I'll ping you on Slack to discuss more.

padmer:
I tried with Julia v1.2.0 and nteract 0.15.0 and still get the same issue. After Ctrl-. or Ctrl-Space it shows busy on the bottom left for some seconds then goes back to idle but autocomplete doesn't work for me.

willingc:
Closing and consolidating autocomplete under #4995

PR #4453, opened 2019-06-15
Fix up editor tooltip

### LINKED ISSUE #4421: Inline Documentation Does Not Render Correclty
nteract desktop

**Describe the bug**
When I try to look at the documentation using the keyboard shortcut (CTRL + .) the documentation renders, but it is not opaque. Meaning the notebook code blends in with the documentation code. Additionally, the "X" to close the documentation is not clickable, so it cannot be closed.

**To Reproduce**
Launch documentation with keyboard shortcut.

**Expected behaviour**
Documentation launches in a form that is readable.

**Screenshots**
image

**Desktop Version

- OS: 10.14.4 macOS 
- Version: 0.14.3 (0.14.3)

BenRussert:
Changes: 
- Esc destroys tooltip (from editor or tooltip itself)
- Ctrl-. destroys tooltip if already open (toggle)
- Tip position updates based on current cursor position
- Close #4421


packages/editor/src/index.tsx
```diff
@@ -246,6 +233,18 @@ export default class CodeMirrorEditor extends React.PureComponent<
 
     const { completion, editorFocused, focusAbove, focusBelow } = this.props;
 
+    // ensure a single tooltip holder exists on document.body
+    const tipHolder = document.getElementsByClassName(
+      "tip-holder"
+    )[0] as HTMLDivElement;
```
rgbkrk, line 239: Is this guaranteed to return a non-empty array? We probably need the bounds checked on this.

packages/editor/src/index.tsx
```diff
@@ -263,8 +262,16 @@ export default class CodeMirrorEditor extends React.PureComponent<
...
 
-    this.cm.on("topBoundary", focusAbove!);
-    this.cm.on("bottomBoundary", focusBelow!);
+    this.cm.on("topBoundary", (editor: Editor) => {
+      this.deleteTip();
+      focusAbove!(editor);
+    });
+    this.cm.on("bottomBoundary", (editor: Editor) => {
+      this.deleteTip();
+      focusBelow!(editor);
```
rgbkrk: Is this similar to the `?` operator from coffeescript or is this forcing `focusBelow` to be assumed non-null?

packages/editor/src/components/tooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <Tip
+      className="CodeMirror-hint"
+      tabIndex={0}
+      onKeyDown={e => {
+        e.key === "Escape" ? deleteTip() : null;
+      }}
+      cursorCoords={cursorCoords}
+    >
+      <RichMedia data={bundle} metadata={{ expanded: true }}>
+        <Media.Markdown />
```
BenRussert, line 52: By the way markdown wasn't here before. The IJulia kernel often sends back markdown docs from inspection requests, so we've supported it in Hydrogen for a while. (Julia kernel is another issue #4415, anybody let me know if you want to work on that I've been stumped a couple times now)

packages/editor/src/index.tsx
BenRussert: Yep, if the class name ("tip-holder" in this case) returns no results, the result is an empty iterable/array-like `HTMLDivElement`

packages/editor/src/index.tsx
BenRussert: Yes, non-null assertion operator. I carried this over from the previous code. It seems better to remove.

PR #4454, opened 2019-06-15
Pass down props for Markdown component

wadethestealth:
This allows properties to be passed down inside the markdown component allowing more control for customization of your provided component.

In particular, hydrogen needs to be able to set `className` prop in order to revert some style choices, but are unable to since properties aren't being passed down.

I am unsure if every component needs this, but markdown should have it.

### DIFF
--- a/packages/outputs/src/components/media/markdown.tsx
+++ b/packages/outputs/src/components/media/markdown.tsx
@@ -20,7 +20,8 @@ export class Markdown extends React.PureComponent<Props> {
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


BenRussert:
I'm wondering if it would be better to just create a custom markdown component for hydrogen. You could start by just copying over this file into a new component in the Hydrogen repo and as long as the media type default prop stays the same it will basically work the same.

wadethestealth:
That's probably the best way to do it

PR #4456, opened 2019-06-16
Fix the ordering of cell toolbar and notifications

### LINKED ISSUE #4404: Cell toolbar overlaps with notification item
The cell toolbar appears over the notification element when the user is prompted with a notification item.

rganardi:

Fixes #4404.

The cell toolbar appears over the notification element, so reorder the `z-index` so the elements appear in the right order

> notification
> toolbar
> cell


rgbkrk (approved):
Thank you, this is a pragmatic solution that I've punted on while waiting for us to come back to looking at our stacking contexts and an explicit z index hierarchy. 👌

PR #4476, opened 2019-06-29
Adjust z-index for leaflet component

### LINKED ISSUE #4316: Leaflet map overflows status bar
nteract Desktop

**Describe the bug**
The Leaflet map overflows and is above the status bar when scrolling on macOS.

**To Reproduce**

1. Open the GeoJSON example notebook under Examples > Python.
2. Scroll down.

**Expected behavior**
The map should remain behind the status bar.

**Screenshots**
Screen Shot 2019-04-01 at 10 12 04 PM

- OS: macOS
- Version nteract 0,13

rganardi:
Fixes #4316

The leaflet map appears over the status bar, so reorder the `z-index` so the elements appears in the right order.

### DIFF
--- a/packages/transform-geojson/src/leafletCss.ts
+++ b/packages/transform-geojson/src/leafletCss.ts
@@ -17,6 +17,7 @@ export default createGlobalStyle`
 	}
 .leaflet-container {
 	overflow: hidden;
+	z-index: 9;
 	}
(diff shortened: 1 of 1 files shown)


captainsafia:
Thanks for submitting this fix, @rganardi!

PR #4477, opened 2019-06-30
Add check for killed kernel to comms epics

captainsafia:
This PR:

- Updates the comms epics to stop listening when a kernel has been killed
- Updates the update display epic to stop listening when a kernel has been killed


mpacer:
The last time I tried to do something with comms they had stopped working as I had expected them to. 

Where did the issue come up that triggered this & how did you debug the comms parts?

rgbkrk (approved):
We have such a common pattern of "do this when kernel is launched then stop doing it once it's killed" that it makes me wonder if there's a common operator we should make. Something like `duringKernelLifecycle` that will do the `switchMap` on the new kernel action(s?) and then always do an "inner" `takeUntil` on the kill action(s).

That's not a blocker for this though, you've got the right pattern in place here.

Issue #993, opened 2016-10-12
Linkify URLs in plain text output


gnestor:
notebook supports this (but I noticed that jupyterlab doesn't yet).

Screenshot: 
![image](https://cloud.githubusercontent.com/assets/512354/19318998/ad0df8a6-905f-11e6-98fc-9bad00e71403.png)

gnestor:
@rgbkrk Does it make sense to add this feature to [ansi-to-react](https://github.com/nteract/ansi-to-react)?

rgbkrk:
I'm not sure where this would belong. I'd rather we finish off handling line feeds and other ANSI/terminal codes before we add automatic linkification.

gnestor:
I was looking at the anser library used in ansi-to-react and there is a linkify method which would require we use its `ansiToHtml` method vs. `ansiToJson` and then `dangerouslySetInnerHTML`.

rgbkrk:
While convenient to use the linkify, it comes at the cost of reducing what can be displayed in untrusted notebooks. All outputs from streams would not be shown on opening the notebook.

gnestor:
True. What if we just rolled our own?

``` js
function inlineBundleToReact(bundle, key) {
  // Very rough, use a URL regex and assuming that `anser.ansiToJson` parses URLs into separate items
  if (bundle.content.includes('http') return React.createElement('a', {
    href: bundle.content,
    style: bundle.style,
    key,
  }, bundle.content);
  return React.createElement('span', {
    style: bundle.style,
    key,
  }, bundle.content);
}
```

Would we be able to display this in untrusted notebooks?

rgbkrk:
Yeah, an approach that keeps content as `textContent`, can be displayed. When strings are the `children` of a React Element, they are done as `textContent` which is what you've done above.

There's going to be some amount of tripping on escape codes here. An admittedly naive approach that would keep the console consistent would be to do a second pass on the bundles returned by `ansiToJson`.

As for rolling our own, I recall ~Hyperterm~ Hyper.app talking about trying to do terminal state in a way where they could push down state into a React component more effectively. At a high level, the current state of the terminal is an intermediate reduction on the stream of character. Stepping out of that abstract way to say it, I wish we had a function with a signature like this:

```
function ttyReducer(state: ttyState, text): ttyState
```

Where ttyState is (in all likelihood) a string. This is important to me because we'd be able to store the last state in the notebook document rather than the entire stream of text from stdout.

[...]

PR #4480, opened 2019-07-10
Enable linkify in text-based outputs

### LINKED ISSUE #4479: Enable linkify in StreamText output renderer
Currently, the `StreamText` component sets the `linkify` flag to false. We should enable this by default or make it a prop that consumers of the component can pass in.

Note, that prior to enabling this, we should sanity check that our `linkify` component can sanitize data appropriately.

captainsafia:
Closes #4479.

### DIFF
--- a/packages/outputs/__tests__/__snapshots__/media.spec.tsx.snap
+++ b/packages/outputs/__tests__/__snapshots__/media.spec.tsx.snap
@@ -1520,3 +1520,40 @@ exports[`Plain Should render markdown 1`] = `
   </pre>
 </Plaintext>
 `;
+
+exports[`Plain should render urls as links 1`] = `
+<Plaintext
+  data="Let's go to https://google.com today!"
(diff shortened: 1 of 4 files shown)


BenRussert:
I did a quick search to try and figure out why this was disabled in the first place. [Found this conversation](https://github.com/nteract/nteract/issues/993#issuecomment-253286167), but I don't yet follow what the issue is/was with linkify.

PR looks good other than that.

BenRussert:
Oh, but snapshots :roll_eyes:

captainsafia:
@BenRussert The linkify changes mentioned in that conversation are implemented [here](https://github.com/nteract/nteract/blob/master/packages/ansi-to-react/src/index.ts#L73). I believe things got shuffled around when the ansi-to-react package got moved to the monorepo.

BenRussert (approved):
Thanks!

PR #4504, opened 2019-07-31
Set monospace default font for outputs

captainsafia:
Just a quick fix for something I noticed.


willingc (approved):
LGTM. Thanks @captainsafia

PR #4510, opened 2019-08-08
Add muted attribute to splash video so that autoplay works again

motin:
I noticed (in the source code) that the splash image was in fact a video. This PR sets the video free to do what videos do best - play.

Credit for the solution goes to https://stackoverflow.com/a/50742427/682317

PS First PR here, so:


captainsafia (approved):
Thanks for submitting this fix! I just saw that autoplay was not working for the splash video. Good catch!

PR #4512, opened 2019-08-09
Run webpack command as shell command on Windows

captainsafia:
When trying to run the Jupyter extension on Powershell, the `Popen` command fails to find the yarn executable. Setting `shell=True` spawns a command prompt with the environment variables for global Node modules set correctly so that the webpack command in our Jupyter extension development build can be run.


rgbkrk (approved):

rgbkrk:
Thanks!

PR #4528, opened 2019-08-26
Fix execute and shift focus flow on Markdown cells

### LINKED ISSUE #4524: Focus on next cell when executing code cell should not enter edit mode
**Problem**
Currently, when a user executes a code cell that comes before a markdown cell, the focus will shift to the markdown cell and enter edit mode. This is jarring and not the flow that users are used to in classic.

**Solution**
When shifting focus, we should check to see if the cell that focus is being shifted to is a markdown cell and not focus on the editor if so.

captainsafia:
Closes #4524 

Ctrl + Enter now shifts to focus on the editor only if the following cell is a code cell or a cell added at the end of the notebook. Ctrl + Enter will still focus on Markdown cells but not enter edit mode.

### DIFF
--- a/packages/notebook-app-component/src/notebook-app.tsx
+++ b/packages/notebook-app-component/src/notebook-app.tsx
@@ -482,8 +482,7 @@ const makeMapStateToProps = (
   const mapStateToProps = (state: AppState): NotebookStateProps => {
     const content = selectors.content(state, { contentRef });
     const model = selectors.model(state, { contentRef });
-    const focusedCell = selectors.notebook.cellFocused(model);
-    const cellMap = selectors.notebook.cellMap(model);
+
 
     if (!model || !content) {
       throw new Error(
@@ -507,6 +506,9 @@ const makeMapStateToProps = (
         "<Notebook /> has to have content & model that are notebook types"
       );
     }
+
+    const focusedCell = selectors.notebook.cellFocused(model);
+    const cellMap = selectors.notebook.cellMap(model);
 
     return {
       cellOrder: model.notebook.cellOrder,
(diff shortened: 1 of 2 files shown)


packages/notebook-app-component/src/notebook-app.tsx
```diff
@@ -595,9 +606,22 @@ export class NotebookApp extends React.PureComponent<NotebookProps> {
...
       focusNextCell({ id: undefined, createCellIfUndefined: true, contentRef });
-      focusNextCellEditor({ id: undefined, contentRef });
+
+      /** Only focus the next editor if it is a code cell or a cell
+       * created at the bottom of the notebook. */
+      if (
+        nextCell === undefined ||
+        (nextCell && nextCell.get("cell_type") === "code")
+      ) {
+        focusNextCellEditor({ id: focusedCell, contentRef });
```
motin: Out of curiosity, what does sending `focusedCell` do here rather than sending `undefined`?

motin (changes requested):
I tested the changes in the desktop app and indeed the new behavior is much more pleasant and matches the discussion in #4524 

I'd love to see a second test added under the `"detects a focus to next cell keypress"` test which presses shift+enter again and does not `expect(focusNextCellEditor).toHaveBeenCalled();`.

That and to resolve the linting errors :)

packages/notebook-app-component/src/notebook-app.tsx
captainsafia: Previously, if the `id` passed here was `undefined` it would use the value of the currently focused editor. It would then advance the focus to the editor following it.

However, there can be a situation where a user has CodeA, MDA, CodeB. The cell focus might be on MDA but the editor focus might be on CodeA. With the pre-existing behavior, Ctrl + Enter on MDA would have shifted the cell focus to CodeB and the editor focus to MDA.

Instead, we want the editor focus to always be the focused cell if it is a code cell.

Note that I added a `|| undefined` in my latest PR to resolve some type issues. Practically, the `focusedCell` should never be null but ImmutableJS's type definitions for `gets` constrain us here.

captainsafia:
@motin Thanks for the feedback, Frederik! I've added tests and fixed the lint errors.

motin (approved):
Looks super! Thanks @captainsafia :)

Issue #2765, opened 2018-04-15
Create example notebooks that work with the builtin Node.js kernel


rgbkrk:
From @Jaykul in #1513

> it would be really cool if at least one of the example-notebooks worked out of the box with the Node.js kernel 

The "Exploring Custom Revival with JSON.parse" notebook used to, back when it was able to reach the dependency on Immutable.js. We need to do several things here:

* [ ] Update the existing JSON.parse notebook to work out of the box
* [ ] Include more dependencies in our base kernel by adding to a [`NODE_PATH` env variable](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders)
* [ ] Group the collections of notebooks into language sections (node and python at the moment)

PR #4533, opened 2019-08-27
Add nteract node_modules as NODE_PATH

motin:
I found this a great idea, mentioned in https://github.com/nteract/nteract/issues/2765#issue-314426122

It allows the bundled Javascript kernel to access more modules out of the box.

### DIFF
--- a/applications/desktop/src/main/kernel-specs.ts
+++ b/applications/desktop/src/main/kernel-specs.ts
@@ -20,7 +20,8 @@ const KERNEL_SPECS: Kernelspecs = {
       display_name: "Node.js (nteract)",
       language: "javascript",
       env: {
-        ELECTRON_RUN_AS_NODE: "1"
+        ELECTRON_RUN_AS_NODE: "1",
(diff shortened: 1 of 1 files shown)


captainsafia (approved):
Thank you!

rgbkrk:
Does this mean though that if you load a notebook with a directory structure like this:

```
mine.ipynb
node_modules
package.json
```

That it won't use the `node_modules` from that package?

motin:
> Does this mean though that if you load a notebook with a directory structure like this:
> 
> ```
> mine.ipynb
> node_modules
> package.json
> ```
> 
> That it won't use the `node_modules` from that package?

First of all - thank you for hinting at that possibility! Being able to add dependencies to your notebook that way was something I didn't expect to be supported. Would love to see this in the Node.js Kernel docs. An example notebook showing this would also be useful.

As for the question, no it does not break that functionality, it merely additionally makes more node modules available to node.js-based notebooks.

rgbkrk:
Awesome, thank you!

PR #4534, opened 2019-08-27
Improve yarn run spawn debug

motin:
When nteract hangs with no apparent reason and no helpful console output, I want (to) enable a debug mode, so that I can see helpful console output to troubleshoot this.

As a new contributor, i interpreted `yarn spawn:debug` as running nteract in debug mode. This PR makes nteract behave more in according with expectations of a debug mode, with more useful debug messages being output to the console(s).

(Installing app deps before running spawn:debug prevents a follow-up use case where I had checked out a new clone of nteract and did not run `yarn spawn` before `yarn spawn:debug`. This resulted in failure for nteract to start.)

### DIFF
--- a/applications/desktop/package.json
+++ b/applications/desktop/package.json
@@ -30,6 +30,7 @@
     "start": "npm run spawn",
     "prespawn": "npm run install:app-deps",
     "spawn": "cross-env NODE_ENV=development electron .",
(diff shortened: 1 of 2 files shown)


captainsafia (approved):
Thanks!

PR #4536, opened 2019-08-28
Clean up folder structure for selectors

captainsafia:
**Changes in this PR**
- Moves selectors into a folder structure that mimics that of reducers
- Removes the dependency on the `identity` function. Note, I wasn't sure why this was used, it seemed like a relic from the Flow days. If it is necessary, I can bring it back.


packages/selectors/src/core/contents/index.ts
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @returns                             The kernel associated with a notebook
+ */
+export const kernelRefByContentRef = (
+  state: AppState,
+  ownProps: { contentRef: ContentRef }
+): KernelRef | null | undefined => {
+  const c = content(state, ownProps);
+  // TODO: When kernels can be associated on other content types, we'll
+  //      allow those too. For now, because of how flow works we have to
```
rgbkrk: We could probably get rid of the flow comment here (yes I realize this was copied over).

rgbkrk:
At some point I think we need to break away from the create selectors library since we're not actually memoizing (or incidentally doing it in a cache-bust-every-time fashion, since we always have to do `byRef` lookups)

rgbkrk (approved):

packages/selectors/src/core/contents/index.ts
captainsafia: Done.

PR #4542, opened 2019-08-30
Add more selectors for comms state

captainsafia:


packages/selectors/src/comms.ts
```diff
@@ -17,3 +17,46 @@ export const models = createSelector(
   comms,
   comms => comms.get("models")
 );
+
+/**
+ * Returns the registered comm targets that are stored in the nteract application state.
+ */
+export const targets = createSelector(
+  comms,
+  comms => comms.get("targets")
```
rgbkrk: Was comms a `Map` rather than a `Record`? If it was a record, we don't even need a selector since we can just go straight to the property.

captainsafia: You're right. It's a record so I can go straight to the property here. Good catch!

rgbkrk: What I mean is that use of this code within other parts of the frontend can just access the property instead of making these wrapper functions. This would reduce the testing and typing layers.

PR #4544, opened 2019-09-02
Spruce up package documentation

captainsafia:
**Changes in this PR**
- Update documentation page to use Now v2
- Update TypeDoc to use `@strictsoftware/typedoc-plugin-monorepo` plugin which:
    - Automagically handles mapping files to modules without the need for a `@module` comment
    - Includes the README in each folder within the contents of the package documentation

For an example of what this looks like now, check out [this example deployment](https://nteract-packages-captainsafia.nteract1.now.sh/).


blink1073:
Awesome, I had not heard of that plugin, nice find!

packages/types/src/refs.ts
```diff
@@ -1,6 +1,3 @@
-/**
- * @module types
- */
```
rgbkrk, line 3: Yay for deleting all this module redundancy!

PR #4551, opened 2019-09-05
Move ServerConfig definition from rx-jupyter to @nteract/types

languy:
Merge `ServerConfig` definition in `rx-jupyter` with one that already exists in `types/src/entities/hosts.ts`

When integrating nteract notebook into my application, it won't compile:
node_modules/@nteract/selectors/lib/index.d.ts(12,132): error TS2307: Cannot find module '../../rx-jupyter/lib/base'.

This change:
* Removes the `ServerConfig` definition in `packages/rx-jupyter/src/base.ts`
* Updates the `ServerConfig` definition in `types/src/entities/hosts.ts` with the additional fields found in `rx-jupyter`'s definition.
* Makes `crossDomain` and `token` optional fields (was: `boolean | null | undefined` before. There could be some subtle type semantic difference that I'm missing, but I wanted them to be consistent with the other definitions).
* Updates all packages import that rely on `ServerConfig`.

Notes:
* I found some similar definition of ServerConfig in [host-cache/src/host-storage.ts](https://github.com/nteract/nteract/blob/01d9338daf10ad2b3d9c1d0ee11cd95252db7cca/packages/host-cache/src/host-storage.ts#L20). I left it as-is. I'm open to suggestions as to how to eliminate the duplication.
* It seems that `ServerConfig.url` is redundant with `ServerConfig.endpoint`. Can `url` be removed?

### DIFF
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -5,7 +5,7 @@ import { sample } from "lodash";
 import { Action } from "redux";
 import { ofType } from "redux-observable";
 import { ActionsObservable, StateObservable } from "redux-observable";
-import { contents, ServerConfig } from "rx-jupyter";
+import { contents } from "rx-jupyter";
 import { empty, from, interval, Observable, of } from "rxjs";
 import {
   catchError,
@@ -26,7 +26,8 @@ import {
   DirectoryContentRecordProps,
   DummyContentRecordProps,
   FileContentRecordProps,
-  NotebookContentRecordProps
+  NotebookContentRecordProps,
+  ServerConfig
 } from "@nteract/types";
 import { AjaxResponse } from "rxjs/ajax";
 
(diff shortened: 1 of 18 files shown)


captainsafia:
> It seems that ServerConfig.url is redundant with ServerConfig.endpoint. Can url be removed?

I would keep it in for now since it's referenced in `createAJAXSettings` and some external users use the type within their use of nteract.

> I found some similar definition of ServerConfig in host-cache/src/host-storage.ts. I left it as-is. I'm open to suggestions as to how to eliminate the duplication.

We can import from the @nteract/types package in this scenario.

captainsafia (approved):
Approving with a suggestion to update the type reference in host-cache as well.

languy:
> > I found some similar definition of ServerConfig in host-cache/src/host-storage.ts. I left it as-is. I'm open to suggestions as to how to eliminate the duplication.
> 
> We can import from the @nteract/types package in this scenario.

The `ServerConfig `definition in `host-cache/src/host-storage.ts` enforces `crossDomain` to always be `true`: so we now import from `@nteract/types`, but extend in order to override `crossDomain`.

PR #4564, opened 2019-09-10
Fix broken nteract/markdown publish

### LINKED ISSUE #4563: nteract/mardown failing install and build
Markdown publish is breaking
Due to [this commit](https://github.com/nteract/nteract/commit/c1f1ca41081f9e569dda7da68ebaa8d51736922f#diff-2aefa3dbe60d43c8058f25804d2ad46a), the ci fails to find nteract/presentational-components used in the index. I believe the fix is to add nteract/presentational-components to the package.json as a dependency. Looking at the ci though you may also need nteract/outputs in there too.

cc: @captainsafia @BenRussert @rgbkrk

wadethestealth:
Fixes #4563

### DIFF
--- a/packages/markdown/package.json
+++ b/packages/markdown/package.json
@@ -18,6 +18,8 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "@nteract/mathjax": "^4.0.0",
+    "@nteract/presentational-components": "^3.0.0",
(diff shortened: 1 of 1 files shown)


BenRussert (approved):
Thank you for fixing this Thomas!

PR #4578, opened 2019-09-25
Apply fixes to some accessibility issues

captainsafia:
I ran nteract through the Deque Systems accessibly checker and the Lighthouse checker.

The good news: We rank at around 85% accessibility ranking in both at the moment. Not bad!

I applied some fixes to resolve the gnarliest class of issues: those related to color contrast. The big culprit here was our midnight shade which didn't have a wide enough contrast gradient across the light and lightest color variations of that shade.

I also updated our dark theme to use alabaster as the foreground color instead of midnight because it provided better contrasts against the dark background.

Following these changes, our Lighthouse accessibility score is 88%.

CodeMirror presents the gnarliest set of challenges in this regard. For one, the fact that it uses a [hidden input textarea](https://bgrins.github.io/codemirror-accessible/) causes some problems with screen readers. Also, since it is using a textarea is knocks off some accessibility points because there are no labels associated with that form element. Finally, the current color theme we use flags for accessibility for single numbers in the code cell. For example, if you have a cell with just the number 42. I started looking around the Codemirror themes page to see if they had any accessible themes but didn't find anything that passed the automated checkers.

Another class of issues comes from the kernel errors that we get back. The contrast on those is off too but that's set by the ANSI color codes and is out of our control there.

These automated checkers don't assess all things. Lighthouse provides a list of the things it doesn't check and from that list I think "Interactive controls are keyboard focusable" is one we need to work on.

### DIFF
--- a/packages/connected-components/src/notebook-menu/PureNotebookMenu.tsx
+++ b/packages/connected-components/src/notebook-menu/PureNotebookMenu.tsx
@@ -141,6 +141,7 @@ class PureNotebookMenu extends React.PureComponent<PureNotebookMenuProps> {
     // console.log(`::handleActionClick -> ${currentKey.key}`);
     this.handleClick(currentKey);
   };
+
   handleClick = ({ key }: { key: string }) => {
     const {
       saveNotebook,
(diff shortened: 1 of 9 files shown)


willingc (approved):
Looks good @captainsafia.

rgbkrk (commented):
Wow what a nice accessiblity overhaul!

PR #4581, opened 2019-09-29
Add support for deriving kernel ref from content ref in actions and epics

captainsafia:
**Changes in this PR**
- Apply some prettify fixes
- Add kernelByContentRef selector to get kernel from kernels record by ContentRef stored in content model
- Updates epics/reducers/actions that were using current kernel primarily to use kernelByContentRef first and currentKernel as a fallback


captainsafia:
I'm working on adding unit tests but I figured I would open this up for functionality review in the meanwhile.

cc: @rgbkrk @BenRussert @willingc

rgbkrk (approved):
Quite the overhaul!

Seems like maybe we should have done a `prettier` run as a separate PR. I'll just approve this as is for now.

captainsafia:
@rgbkrk Yeah, I think I will cherry-pick the prettify commit out of this.

Surprisingly, this was quite an easy change to implement. It involved doing the following.

1. Writing a new `kernelByContentRef` selector that extracted a kernel from the `kernels` entities list for a given ContentRef.
2. Replace all references to `currentKernel` with the selector from #1.
3. Added backwards-compatibility logic that used the currentKernel or the kernel referenced by the KernelRef in the action if it didn't exist.
4. Add ContentRef as a payload item in all the relevant actions.

Now, this PR is in a pretty much mergeable state. However, there is some follow-up work. It includes the following:

- We need to pass the ContentRef back to the user in all FULFILLED and FAILED actions.
- We need to add a deprecation notice to the `currentKernel` selector and, eventually, remove it all together and release a major-version of the core SDK.
- We need to create an appropriate test-bed for this, something like the multi-tab nteract application that @languy mentioned.

I'm inclined to merge as-is since this PR includes backwards compatibility and then iterate on the API experience for this moving forward.

I'll take out the prettify commit and write another round of unit tests and then set this in action.

Also, I will create task items for the follow-on work listed below.

applications/desktop/src/notebook/epics/zeromq-kernels.ts
```diff
@@ -195,11 +194,7 @@ export const launchKernelByNameEpic = (
             return actions.launchKernelFailed({
               contentRef: action.payload.contentRef,
               error: new Error(
-                `A kernel named ${
-                  action.payload.kernelSpecName
-                } does not appear to be available. Try installing the ${
-                  action.payload.kernelSpecName
-                } kernelspec or selecting a kernel from the runtime menu.`
+                `A kernel named ${action.payload.kernelSpecName} does not appear to be available. Try installing the ${action.payload.kernelSpecName} kernelspec or selecting a kernel from the runtime menu.`
```
languy: nit: I would phrase it to make it less specific to the nteract menu implementation (ie in our app, changing kernel is done from a different menu).

captainsafia: Good call!

PR #4582, opened 2019-09-29
tear reselect out of comms & kernelspecs selectors

rgbkrk:
We can just pluck the values out of our state tree, we don't need to memoize these operations. In some cases it's just silly (now that we have `Record`s, we can directly access properties). In other cases it's misguided because `reselect` memoizes one while we end up having to get different cells and other entities by ID.

This is a WIP PR while I tear it out from each file using `reselect` individually.

### DIFF
--- a/packages/selectors/src/comms.ts
+++ b/packages/selectors/src/comms.ts
@@ -1,7 +1,5 @@
 import { AppState } from "@nteract/types";
 
-import { createSelector } from "reselect";
-
 /**
  * Returns the Jupyter comms data for a given nteract application.
  */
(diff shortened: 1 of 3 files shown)


captainsafia (approved):
LGTM flagged an unused import. Other than that, good to go!

rgbkrk:
Oops! I forgot to force-push my updated commits yesterday. 

Screen Shot 2019-09-30 at 4 39 09 PM

rgbkrk:
Thanks for the merge! I'll try to get back on more of these tonight.

Issue #4461, opened 2019-06-19
Nteract Mathjax package has trouble with rendering in Atom


wadethestealth:
### Background Info
We have recently just upgraded all of hydrogen's old components to the new outputs package, which is a great improvement. However, it seems like the @nteract/mathjax does not like this new style of rendering components. When trying to use LaTeX through either python or our new markdown result bubbles, we are constantly getting a Content Security Policy Error.

I have done some debugging, and have tried to figure out the issue, but nothing seems to work.

### Abstract Model of our Result View Component
```html
<MathJax.Provider src={LocalMathjaxFile} > <!-- This is technically in a separate file wrapping our component -->
    <some random and nested divs>
        <Outputs>
            <DisplayData>
                <Media.Markdown />
            </DisplayData>
        </Outputs>
    </some random and nested divs>
</MathJax.Provider >
```

### Abstract Model of a Result View Component that gets rendered
```html
<MathJax.Provider src={LocalMathjaxFile} >
    <some random and nested divs>
        <Outputs>
            <DisplayData>
                <Media.Markdown >
                    <MathJax.Provider src={cdn-that-can't-be-loaded-in-atom} >
                        <MathJax.Node >
                        <MathJax.Node />
                    <MathJax.Provider />
                </Media.Markdown>
            </DisplayData>
        </Outputs>
    </some random and nested divs>
</MathJax.Provider >
```

### The Issues
It seems that the Media.Markdown component is not recognizing or doesn't care about the wrapping provider, and it simply creates a new Provider every time but tries to load from a cdn, which is why I get the CSP error. However, we need it to use the local Provider. Interestingly though, It will still render but only the first markdown cell you run. The second run cell will just be empty, like its waiting on the script to render it. If you use a python code cell, and use the ipython latex, it still gives you the CSP error, but it renders everytime, which I don't understand. On top of this, we are creating providers for every cell result, which results in a bunch of the same MathJax script for every result, and I was wondering if there is a way to fix these issues and only how one script being added to the head of the atom editor. Thanks for any insight you can give @rgbkrk

rgbkrk:
Oh no, that is troubling. It seems the provider logic is off at the moment. Maybe the right approach is to override the Markdown package for now within Hydrogen (set it up to handle the same media type). cc @BenRussert

wadethestealth:
[...]

PR #4586, opened 2019-10-01
Update node.tsx to add a delay for typsetting

wadethestealth:
This is a temporary fix to #4461 and has already been discussed between @rgbkrk and I inside #4461.

### DIFF
--- a/packages/mathjax/src/node.tsx
+++ b/packages/mathjax/src/node.tsx
@@ -28,7 +28,7 @@ class MathJaxNode_ extends React.Component<Props & MathJaxContextValue> {
    * Render the math once the node is mounted
    */
   componentDidMount() {
-    setTimeout(() => this.typeset(), 10);
+    setTimeout(() => this.typeset(), 0);
   }
 
   /**
--- a/packages/mathjax/src/node.tsx
+++ b/packages/mathjax/src/node.tsx
@@ -28,7 +28,7 @@ class MathJaxNode_ extends React.Component<Props & MathJaxContextValue> {
    * Render the math once the node is mounted
    */
   componentDidMount() {
-    this.typeset();
+    setTimeout(() => this.typeset(), 10);
   }
 
   /**


captainsafia (approved):
Change seems reasonable.

Is the 10ms the only value that will work or is it any value within a certain threshold?

Also, can you add a comment in this change to explain why it was made?

wadethestealth:
@captainsafia it seems that even if I set the value to 0 the extra overhead seems to still make it work, but if I remove the set timeout it goes back to only rendering the first node. The intuition behind the change is that Mathjax is supposedly async, and hydrogen was only ever able to render the first node ran, and it always left the first node in the processing state even though it was finished rendering. So I randomly tried adding a timeout so that I could easier debug and the issue went away completely. Again something wrong is happening here without the setTimeout, but I have tried to step by step debug it multiple times with beautified code, and I have seen no obvious errors. Again this is just a temporary fix to the issue and the real issue is written very in depth inside #4461

packages/mathjax/src/node.tsx
```diff
@@ -28,7 +28,7 @@ class MathJaxNode_ extends React.Component<Props & MathJaxContextValue> {
    * Render the math once the node is mounted
    */
   componentDidMount() {
-    this.typeset();
+    setTimeout(() => this.typeset(), 10);
```
rgbkrk: I wonder if this would even work with `setTimeout(fn, 0);`, since it still lets some DOM rendering catch up.

rgbkrk (approved):

packages/mathjax/src/node.tsx
wadethestealth: at least for hydrogen setting it to 0 still works, idk if we should leave it as 10 though for tolerance or not

rgbkrk: I'd probably go with 0 if it works as it basically just schedules the work after everything else in the queue. Background: https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful

wadethestealth:
Ready for merge? @rgbkrk and @captainsafia

rgbkrk:
🎉

Issue #4575, opened 2019-09-24
Restart kernel yields a 500


languy:
**Application or Package Used**
nteract on jupyter, nteract SDK.

**Describe the bug**
When issuing a restart kernel, either from the "Runtime" dropdown menu item in the nteract webapp or dispatching a `restartKernel` action, I'm observing:
a. a DELETE call on the current session 
b. a POST call on sessions
c. a POST call on /api/kernels/<current kernel id>/restart. This fails with a 500 response.

The jupyter console shows:
jupyter-lab     |         raise web.HTTPError(404, u'Kernel does not exist: %s' % kernel_id)
jupyter-lab     |     tornado.web.HTTPError: HTTP 404: Not Found (Kernel does not exist: 1ed91ae8-a636-41de-ad2b-4020987d1688)

I'm not the expert here, but I believe the DELETE call isn't necessary and causes the /restart call to fail, because the kernel has been previously removed by the DELETE call.

This problem isn't visible from the user's perspective (as the kernel is effectively restarted), but it always produces RESTART_KERNEL_SUCCESSFUL as well as RESTART_KERNEL_FAILED actions which show conflicting error/success messages to the user in my application.

**To Reproduce**

1. Open nteract on jupyter
2. Click on Runtime
3. Select Restart Kernel
4. Observe Network tab on developer console and Redux actions

**Expected behavior**
Jupyterlab just produces a single call to /api/kernels/<kernel id>/restart
KERNEL_RESTART followed by either KERNEL_RESTART_SUCCESSFUL or KERNEL_RESTART_FAILED.

captainsafia:
Thanks for opening this issue, @languy!

I'm tagging this one for our Hacktoberfest 2019 participation as it should be manageable for a new contributor.

For Hacktoberfest 2019 participants: the [restart epic](https://github.com/nteract/nteract/blob/master/packages/epics/src/kernel-lifecycle.ts#L214) will need to be updated to only dispatch a single call to the restart API when the kernel that we are connected to is a WebSocket kernel. You can do a search for "websocket" in the codebase to find if-statements where this casing is used.

If the kernel the application is connected to is a zeromq kernel, the restart epic should continue to function as is.

PR #4591, opened 2019-10-05
Fix issue #4575: Websocket kernel should only restart without deletin…

chrispanag:
Fixing issue #4575 

I've tried to make my fix as non-intrusive as possible. Of course feedback is welcome :)

### DIFF
--- a/packages/epics/src/kernel-lifecycle.ts
+++ b/packages/epics/src/kernel-lifecycle.ts
@@ -315,10 +315,8 @@ export const restartKernelEpic = (
         })
       );
 
-      const kernel = selectors.currentKernel(state);
-
       const actionsDispatched =
-        kernel.type !== "websocket" ? of(kill, relaunch) : of(relaunch);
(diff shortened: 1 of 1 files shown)


captainsafia (changes requested):
Thanks for submitting this pull request. Two points of feedback.

- Can you add a test to validate this change? The test will need to be added to [this file](https://github.com/nteract/nteract/blob/f52fb4e3347f4050e21d61a4bb0189af95ba2c7b/packages/epics/__tests__/kernel-lifecycle.spec.ts).
- For the WebSocket scenario, we'll actually need to finish the restart by sending a request to the `/api/kernels/{kernelId}/restart` endpoint on the Jupyter server. This way, we'll be restarting the existing kernel instead of creating a new one. If you have time to make this change in this PR, great! If not, just adding the tests is fine. I can answer any questions you have about what it would take to make that change.

captainsafia:
Hello @chrispanag! This has been resolved in another PR. Thanks for taking the time to make this change.

PR #4593, opened 2019-10-05
[tests] Add NodeJS Global Types for test-shim.ts

hydrosquall:
## Summary

- I started to look at https://github.com/nteract/nteract/issues/3997 with some potential contributors  earlier this week, and noticed that the app-specific test code wasn't even running when trying to test `vega.spec.ts` due to type errors on the `test-shim` file. (we can reproduce by running `yarn test`).

This change should free people up to focus on errors that are specific to the test files.

- To reproduce these errors, make sure this line is set to `false`.

https://github.com/nteract/nteract/blob/a400cf406ab1d39bc29850b3a4bfc27435e5b71c/package.json#L89

Here's the before + after result when patching our `@types/node` definition.

<a href="https://cl.ly/a5948bb7964d" target="_blank"></a>

### DIFF
--- a/scripts/test-shim.ts
+++ b/scripts/test-shim.ts
@@ -11,22 +11,23 @@
 // For some reason, this property does not get set above.
 global.Image = global.window.Image;
 
-global.Range = function Range() {};
+global.Range = function Range(): void {};
 
-global.Blob = function(content, options) {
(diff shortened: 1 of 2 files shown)


captainsafia (approved):
Good catch! Thanks for following through with a fix.

PR #4594, opened 2019-10-05
[tests] Update ts tests for transform-model-debug package

hydrosquall:
## Summary

- Goal: show a small example of how to help with test migration for #3997 
- I developed this on top of commits from #4593 since without those changes I was getting errors at the `test-shim.ts` stage, but I've kept those separate from this PR to avoid confusing things.

*Error message before this change was applied*

<a href="https://cl.ly/d47fe455ad8d" target="_blank"></a>

### DIFF
--- a/packages/transform-model-debug/__tests__/model-debug.test.tsx
+++ b/packages/transform-model-debug/__tests__/model-debug.test.tsx
@@ -10,7 +10,14 @@ describe("ModelDebug", () => {
     );
 
     const instance = modelDebugWrapper.instance();
-    expect(instance.shouldComponentUpdate()).toBeTruthy();
+
+    // shouldComponentUpdate is tested twice
+    // First time is for an informative Jest message
+    expect(instance.shouldComponentUpdate !== undefined);
+    // Second time is for typescript compiler to pass
+    if (instance.shouldComponentUpdate !== undefined) {
+      expect(instance.shouldComponentUpdate({}, {}, {})).toBeTruthy();
+    }
 
     expect(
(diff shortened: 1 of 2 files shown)


packages/transform-model-debug/__tests__/model-debug.test.tsx
```diff
@@ -10,7 +10,9 @@ describe("ModelDebug", () => {
     );
 
     const instance = modelDebugWrapper.instance();
-    expect(instance.shouldComponentUpdate()).toBeTruthy();
+    if (instance.shouldComponentUpdate !== undefined) {
```
hydrosquall, line 18: The reason for checking this is that not every react component instance will have a `shouldComponentUpdate` function. We could add a jest expectation that the instance have a `instance.shouldComponentUpdate` method as well, but I wasn't sure that was necessary since that feels like an internal implementation detail.

captainsafia: Hmmm. Since this is a test, I'm partial to checking if `shouldComponentUpdate` exists via an assertion.

hydrosquall: Ah, good point. I was a little bit hesitant about this test because it forces the component to be a class component, but I see your point - I've added an `expect` in front of this one. (Both are needed because otherwise the typescript compiler complains.)

captainsafia: Got it. Thanks for clarifying!

captainsafia (approved):

Issue #4583, opened 2019-09-30
SAVE_FAILED because of inaccurate last_modified property


languy:
**Application or Package Used**
nteract on jupyter, nteract SDK

**Describe the bug**
I'm running a jupyter server with the nteract-on-jupyter extension from a docker container.
The first save of a notebook succeeds, but subsequent save attempts result in SAVE_FAILED emitted.
The failures happen [here](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fnteract%2Fnteract%2Fblob%2F0da75faefc6cef1bcf3ad86847a454b8f88d388d%2Fpackages%2Fepics%2Fsrc%2Fcontents.ts%23L439&data=02%7C01%7CLaurent.Nguyen%40microsoft.com%7C4de5384c34d34a801b7b08d742d974bf%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637051374047481130&sdata=Q1mhQTtt39ZO%2FYL4COPd5ZrgbAMkg4RX%2Bt8QYULV4yw%3D&reserved=0): `diffDate` in my case is well over 600ms.

Looking into the issue, I found that the jupyter server returns an incorrect last_modified value on the PUT call to save content. I checked the jupyter server python source code and it seems correct AFAICT: it checks the file attribute after writing, but the last_modified field returned by PUT is definitely “earlier in time” compared to what File Explorer (or ls -la in my docker container) reports. Maybe the write is async and jupyter server is checking for last_modified before the write is finished. 
 
So it looks like this Math.abs(diffDate) > 600 test is a workaround for this jupyter problem. If we had an accurate last_modified date, we would be doing a strict === comparison to avoid any possible race condition. The first Save works, because fetching content gets the correct last_modified value and diff is 0ms.

[...]

PR #4596, opened 2019-10-08
Repeatedly check last-modified file flag in Save before sending save successful

languy:
Here's a PR to address this #4583

### DIFF
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -15,7 +15,8 @@ import {
   mergeMap,
   pluck,
   switchMap,
-  tap
+  tap,
+  take
 } from "rxjs/operators";
 
 import * as actions from "@nteract/actions";
@@ -27,7 +28,8 @@ import {
   DummyContentRecordProps,
   FileContentRecordProps,
   NotebookContentRecordProps,
-  ServerConfig
+  ServerConfig,
+  JupyterHostRecord
 } from "@nteract/types";
 import { AjaxResponse } from "rxjs/ajax";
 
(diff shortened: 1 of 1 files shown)


packages/epics/src/contents.ts
```diff
@@ -446,11 +448,48 @@ export function saveContentEpic(
...
 
               return contents.save(serverConfig, filepath, saveModel).pipe(
-                map((xhr: AjaxResponse) => {
-                  return actions.saveFulfilled({
-                    contentRef: action.payload.contentRef,
-                    model: xhr.response
-                  });
+                mergeMap((saveXhr: AjaxResponse) => {
+                  const pollIntervalMs = 500;
+                  const maxPollNb = 4;
```
captainsafia, line 453: Do you have any insights you can share about the intuition behind these values?

packages/epics/src/contents.ts
```diff
@@ -446,11 +448,48 @@ export function saveContentEpic(
...
+
+                  // Last_modified value from jupyter server is unreliable: https://github.com/nteract/nteract/issues/4583
+                  // Check last-modified until value is stable.
+                  return interval(pollIntervalMs)
+                    .pipe(take(maxPollNb))
+                    .pipe(
```
captainsafia, line 459: Curious: why is the `take` operator in its own subscription as opposed to in this one?

captainsafia (commented):
Can we add a unit test for this in `contents.spec.ts`? I'm imagining that using mocks we can implement a mocked `contents.get` function that returns a last_modified that either stabilizes into a new value or doesn't depending on the scenario.

```
const counter = 0;
jest.mock("rx-jupyter", () => ({
  contents: {
    get: (severConfig, filepath, model) => {
      while (counter <= 3) {
            return of({ response: {last_modified: 'some_value'} });
      }
      return of({ response: {last_modified: 'a_different_value'} });
     
    }
  }
}});
```

captainsafia:
Closing in favor of https://github.com/nteract/nteract/pull/4700..

Issue #4571, opened 2019-09-18
Expose API handles for disposing of a notebook


captainsafia:
**Problem**
Currently, we don't expose any Redux actions or epics that handle disposing of a notebook once we are done with it.

In both desktop and web app, we rely on the assumption that the user will either close the application window (or the browser tab) and hence dispose of the Redux store and any open WebSocket connections.

We cannot always rely on this assumption in multi-notebook scenarios, such as users who are rendering NotebookApp components within a tabbed or paneled view.

**Proposed Solution**
We need to expose a set of actions/epics for handling the proper closing and disposing of a notebook.

- A `CloseNotebook` action that takes a `ContentRef`
- A reducer that listens for the `CloseNotebook` action and clears the notebook with that ContentRef from the state model. It should also shutdown the kernel and clear out the kernel state for the the kernel associated with that ContentRef
- Update the epics to close out any open comms or kernel watchers

languy:
Attempt to fix here: #4597

PR #4597, opened 2019-10-08
Close notebook workflow

languy:
This change is an attempt to address #4571 
It introduces a CLOSE_NOTEBOOK action. An epic will trigger:
* DISPOSE_CONTENT: it deletes the contenRef from the state (which also prevents auto-save to keep saving the content ref).
* KILL_KERNEL: a `dispose` flag is added to the kill kernel action. If set, the the killer kernel epic sends a DISPOSE_KERNEL upon successful completion (in addition to the usual KILL_KERNEL_SUCCESSFUL).

DISPOSE_KERNEL deletes the kernelRef from the state.

### DIFF
--- a/packages/actions/src/actions/contents.ts
+++ b/packages/actions/src/actions/contents.ts
@@ -206,3 +206,14 @@ export function closeNotebook(payload: {
     }
   };
 }
+
+export function disposeContent(payload: {
+  contentRef: ContentRef;
+}): actionTypes.DisposeContent {
+  return {
+    type: actionTypes.DISPOSE_CONTENT,
+    payload: {
+      contentRef: payload.contentRef
+    }
+  };
+}
--- a/packages/actions/src/actions/contents.ts
+++ b/packages/actions/src/actions/contents.ts
@@ -195,3 +195,14 @@ export function updateFileText(payload: {
     }
   };
 }
+
+export function closeNotebook(payload: {
+  contentRef: ContentRef;
+}): actionTypes.CloseNotebook {
+  return {
+    type: actionTypes.CLOSE_NOTEBOOK,
+    payload: {
+      contentRef: payload.contentRef
+    }
+  };
+}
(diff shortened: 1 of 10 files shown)


languy:
#4571

packages/actions/src/actions/contents.ts
```diff
@@ -195,3 +195,25 @@ export function updateFileText(payload: {
     }
   };
 }
+
+export function closeNotebook(payload: {
```
captainsafia, line 199: Nit: Moving forward we want to adopt a pattern where we use TypeScript's interface accessor to refer to the payload here.

```
export function closeNotebook(payload: actionTypes.CloseNotebook["payload"])
```

Instead of re-implementing the interface.

packages/actions/src/actions/kernels.ts
```diff
@@ -93,6 +93,7 @@ export function kernelRawStderr(payload: {
 export function killKernel(payload: {
   restarting: boolean;
   kernelRef?: KernelRef | null;
+  dispose?: boolean;
 }): actionTypes.KillKernelAction {
```
captainsafia, line 97: Wanna help out with the work to move towards the new pattern for referring to action types by changing this to:

```
export function killKernel(payload: actionTypes.KillKernelAction["payload"])
```

Thanks!

packages/epics/src/websocket-kernel.ts
```diff
@@ -280,11 +280,12 @@ export const killKernelEpic = (
       const serverConfig: ServerConfig = selectors.serverConfig(host);
 
       const kernel = selectors.currentKernel(state);
+      const kernelRef = action.payload.kernelRef;
```
captainsafia, line 283: Note to self: once #4581 is merged, this will need to use `kerelRefByContentRef`.

captainsafia (approved):
Left some nit comments in-line.

captainsafia:
Closing in favor of https://github.com/nteract/nteract/pull/4698/ which contains the necessary follow-ups.

PR #4598, opened 2019-10-09
Add global error handler for redux-observable

captainsafia:
The `combineEpics` function in redux-observable merges a set of epics into a mega-epic.

This mega-epic lacks error handling by default. For more context on this, see https://github.com/redux-observable/redux-observable/issues/591.

This PR adds global error handling to the master epic in both the desktop application and the Jupyter extension.

This error handler logs the error and returns the re-subscribes to the source Observable stream.

### DIFF
--- a/applications/desktop/src/notebook/store.ts
+++ b/applications/desktop/src/notebook/store.ts
@@ -21,7 +21,7 @@ const rootEpic = (
 ) =>
   combineEpics(...epics)(action$, store$, dependencies).pipe(
     catchError((error: any, source: Observable<any>) => {
-      console.log(error);
+      console.error(error);
       return source;
     })
   );
--- a/applications/desktop/src/notebook/store.ts
+++ b/applications/desktop/src/notebook/store.ts
@@ -1,13 +1,31 @@
 import { middlewares as coreMiddlewares, reducers } from "@nteract/core";
 import { applyMiddleware, combineReducers, createStore, Store } from "redux";
-import { combineEpics, createEpicMiddleware } from "redux-observable";
+import {
+  combineEpics,
+  createEpicMiddleware,
+  ActionsObservable,
+  StateObservable
+} from "redux-observable";
+import { catchError } from "rxjs/operators";
 
 import { Actions } from "./actions";
 import epics from "./epics";
 import { handleDesktopNotebook } from "./reducers";
 import { DesktopNotebookAppState } from "./state";
+import { Observable } from "rxjs";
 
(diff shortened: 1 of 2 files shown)


applications/desktop/src/notebook/store.ts
```diff
@@ -1,13 +1,31 @@
...
+import { Observable } from "rxjs";
+
+const rootEpic = (
+  action$: ActionsObservable<any>,
+  store$: StateObservable<any>,
+  dependencies: any
+) =>
+  combineEpics(...epics)(action$, store$, dependencies).pipe(
+    catchError((error: any, source: Observable<any>) => {
+      console.log(error);
```
jayphelps: What are your thoughts on console.error perhaps? In case stderr is reported/stored differently?

captainsafia: Good call! I've updated this PR to use `console.error`.

By the way, what are your thoughts on adding a note on this approach to [this documentation page](https://redux-observable.js.org/docs/basics/SettingUpTheMiddleware.html#)? Happy to open a PR with an edit.

jayphelps: I definitely want to do _something_. The latest thoughts [are here](https://github.com/redux-observable/redux-observable/issues/94#issuecomment-532871810) but there was also talk about something [different today here](https://github.com/redux-observable/redux-observable/pull/674).

If you want to add this pattern to the docs in the mean time, I'd definitely welcome it, but we'd need to make sure it was clear that console.error isn't supported in IE 8/9 without the dev tools open and in that case they might want to use `setTimeout(() => { throw e; })` instead (or always use that pattern)

jayphelps: Also if you do PR add a note plz about that "restarting" epics isn't always safe since epics can be stateful and if you restart it, it lost its state, similar to the fact that an unhandled exception might mean your reducers or other app state is in a bad state too. For most people's apps, the risk is probably worth it to try and at least limp along as the web normally does, instead of having that native feel™

lol

```js
const rootEpic = (...args) => combineEpics(...args)(..epics).pipe(
  catchError((e) => {
    window.alert(e.message);
    window.close();
  })
);
```

captainsafia: To be honest, I'm just going to start calling `window.close` in all my error handling code.

Here's the PR, feel free to provide feedback there.

https://github.com/redux-observable/redux-observable/pull/675

rgbkrk (approved):
Thanks for this default handling.

PR #4605, opened 2019-10-10
Extract last_modified and created date from model

captainsafia:
For some reason, the "notebook" case in the FetchContentFulfilled reducer was attempting to read the initial values from properties that did not exist.

It should be reading `created` and `lastSaved` from the `created` and `last_modified` properties within the model, respectively.


zhenchaoni:
Thanks for the fix!

rgbkrk (approved):
Thank you!

PR #4610, opened 2019-10-14
Move the spread statement so MarkdownRender.renderers can override preset value.

### LINKED ISSUE #4609: Cannot assign a custom renderer in nteract/markdown
I want to add some extra spacing before and *after* rendered code fence. But I couldn't find a location where I can customize it.
```ts
const App: React.FC = () => {
  return (
    <div className="App">
      <div style={{ border: "1px solid red" }}>
        <MarkdownRender source={"test\n```ts\ncode fence\n```"} />
      </div>
    </div>
  );
}
```
image

The `margin: 0` is hard-coded in the generated HTML, so I actually cannot change it.
image

Simply moving L27 (the spread operator) to L30 should do. I will at least be able to place all the code block inside a `<div>` that gives it extra spacing. This is only possible if I can set the renderer from outside the `MarkdownRender`.
https://github.com/nteract/nteract/blob/17b124e512e65c0094adfa402b521c439f677ec8/packages/markdown/src/index.tsx#L26-L31

**Additional context**

chenxinyanc:

Fixes #4609.

### DIFF
--- a/packages/markdown/src/index.tsx
+++ b/packages/markdown/src/index.tsx
@@ -24,10 +24,10 @@ const MarkdownRender = (props: ReactMarkdown.ReactMarkdownProps) => {
     escapeHtml: false,
     plugins: [RemarkMathPlugin],
     renderers: {
-      ...props.renderers,
       math,
       inlineMath,
-      code
+      code,
+      ...props.renderers
     } as any
   };
 
--- a/packages/markdown/src/index.tsx
+++ b/packages/markdown/src/index.tsx
@@ -17,8 +17,8 @@ const code = (props: {
   language: string;
   value: string
 }): React.ReactElement<unknown> => (
-    <Source language={props.language}>{props.value}</Source>
-  );
+  <Source language={props.language}>{props.value}</Source>
+);
 
 const MarkdownRender = (props: ReactMarkdown.ReactMarkdownProps) => {
   const newProps: ReactMarkdown.ReactMarkdownProps = {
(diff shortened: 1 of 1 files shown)


packages/markdown/src/index.tsx
```diff
@@ -5,30 +5,33 @@ import ReactMarkdown from "react-markdown";
 
 import RemarkMathPlugin from "./remark-math";
 
-const math = (props: { value: string }): React.ReactNode => (
+const math = (props: { value: string }): React.ReactElement<unknown> => (
```
wadethestealth: shouldn't `unknown` be `any`, plus I don't really understand this change to begin with.

packages/markdown/src/index.tsx
```diff
@@ -5,30 +5,33 @@ import ReactMarkdown from "react-markdown";
...
 
-const code = (props: { language: string; value: string }): React.ReactNode => (
-  <Source language={props.language}>{props.value}</Source>
-);
+const code = (props: {
+  language: string;
+  value: string
+}): React.ReactElement<unknown> => (
+    <Source language={props.language}>{props.value}</Source>
+  );
```
wadethestealth: small adjustment, but this shouldn't be tabbed

wadethestealth (approved):

packages/markdown/src/index.tsx
wadethestealth: line 20-21 should be 2 spaces back

wadethestealth (approved):

wadethestealth:
We normally publish Mondays, so you should see this change soon.

Also Thanks 😄

Issue #4612, opened 2019-10-15
Client keeps reconnecting to the old kernel when there's a kernel switch


ghost:
**Application or Package Used**
nteract web, nteract on Jupyter

**Describe the bug**
The client keeps reconnecting to the old kernel when there's a kernel switch.

**Expected behavior**
The client should dispose the old socket connection and switch to the new kernel

captainsafia:
This should be resolved now. Please re-open if this is not the case.

PR #4614, opened 2019-10-16
Remove reconnecting WebSocket logic

captainsafia:
This bit of logic has admirable aspirations but is rather buggy, so I'm nuking it for now. Related to #4612.

My suspicion is that we will need to move this retry logic to higher up in the stack, probably as a kernel lifecycle epic or the like.

Placing it in a kernel lifecycle epic has the benefit of making it easier to test and debug, easier for consumers to disable, and easier to integrate with other kernel status/reliability features.

### DIFF
--- a/packages/rx-jupyter/src/kernels.ts
+++ b/packages/rx-jupyter/src/kernels.ts
@@ -151,35 +151,6 @@ export const connect = (
     protocol: serverConfig.wsProtocol
   });
 
-  wsSubject.pipe(
-    retryWhen(error$ => {
-      // Keep track of how many times we've already re-tried
-      let counter = 0;
-      const maxRetries = 100;
-
-      return error$.pipe(
-        tap(e => {
-          counter++;
-          // This will only retry on error when it's a close event that is not
-          // from a .complete() of the websocket subject
-          if (counter > maxRetries || e instanceof Event === false) {
-            console.error(
-              `bubbling up Error on websocket after retrying ${counter} times out of ${maxRetries}`,
-              e
-            );
-            throw e;
-          } else {
-            // We'll retry at this point
-            console.log("attempting to retry kernel connection after error", e);
-          }
-        }),
-        delay(1000)
-      );
-    }),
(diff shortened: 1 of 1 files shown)


captainsafia:
cc: @languy @virangai

ghost (approved):

captainsafia:
@languy and I have both noticed the interesting characteristic that the `kernel.channels` subscription will error out if the kernel is disconnected. For example, if you kill the Jupyter server you are connected to, you'll see a `UPDATE_DISPLAY_FAILED` action emitted because the `updateDisplay` epic runs an active listener on the WebSocket subject.

We can probably exploit this behavior to implement the retry logic -- although I'm not bull-ish on client-side checks for kernel state as they are vulnerable to network latency or client issues.

On the other hand, these vulnerabilities shouldn't rule out the implementation of client side checks. Just wanting to note that a robust solution will include both client-side and server/kernel-side status checks. We have flexibility to address the server/kernel-side with the fs-kernels and GraphQL kernel server work (cc: @rgbkrk).

All this in mind, we can do something like this to remedy this in the meanwhile. @languy proposed something like this a while back.

```
export const updateDisplayEpic = (
  action$: ActionsObservable<actions.NewKernelAction>
) =>
  // Global message watcher so we need to set up a feed for each new kernel
  action$.pipe(
    ofType(actions.LAUNCH_KERNEL_SUCCESSFUL),
    switchMap((action: actions.NewKernelAction) =>
      action.payload.kernel.channels.pipe(
            retryWhen(blah)
        ),
        takeUntil(action$.pipe(ofType(actions.KILL_KERNEL_SUCCESSFUL))),
        catchError(error =>
          of(
            actions.kernelIsReallyDeadYo({
              error,
              contentRef: action.payload.contentRef
            })
          )
        )
      )
    )
  );
```

However, the last time I tried this approach, it just....wasn't working. But that might be because the retry logic in the epic was interfering with the retry logic on the subject. Removing the retry logic in the subject might solve this.

Anyone have other thoughts on this?

Issue #4573, opened 2019-09-23
RFC: @nteract/jupyter-widgets Implementation Approach


captainsafia:
**Note:** I'll update this document as I get feedback and clarify more details of the design. You can [view a history of the edits via GitHub](https://help.github.com/en/articles/tracking-changes-in-a-comment).

## Problem Statement

Currently, the nteract project does not export a React component that renders ipywidgets. As a result, users who want to leverage the functionality of a comms-enabled interactive widget API cannot use the nteract desktop and web applications or the nteract core SDK.

## Goals

The nteract project should ship a React component that renders ipywidgets that adhere to the comms message protocol and the ipywidgets convention. Furthermore, React component should provide developers consuming the API with the ability to:

- Replace the default renderer for widget components with a custom renderer that allows developers to ship an experience where ipywidgets rendered within a notebook have a look and feel similar to that of the parent application. This is similar to the functionality support by the Google Colab project in their web-based Jupyter notebook application.
- Allow developers consuming the API to wrap the rendered widgets inside an iframe or another mechanism that provides an isolation boundary between the ipywidget and the notebook. This allows developers consuming the API to isolate ipywidgets, which have the potential to execute malicious JavaScript, behind a secure isolation boundary.

The initial ipywidget implementation that we will ship will not include these two features by default but will expose points were consumers can implement them.

## Design and Implementation

This component will be shipped as a single nteract transform that takes a prop which defines the `model_id` of the widget that will be rendered by the transform.

```
import { WidgetDisplay } from "@nteract/jupyter-widgets";

return <WidgetDisplay data={{'model_id': 'some_model_id_here', version_major: 2, version_minor: 0}}/>
```

Furthermore, we want to continue to support scenarios where a widget might be rendered outside the context of a notebook, such as in the nteract play application. In this scenario, the `WidgetDisplay` transform will take an optional kernel property that can be used to configure the comms setup with the component. This kernel property accepts an interface that matches that of a `KernelRecord` from the nteract core SDK.

```
import { WidgetDisplay } from "@nteract/jupyter-widgets";

return <WidgetDisplay data={{'model_id': 'some_model_id_here', version_major: 2, version_minor: 0}} kernel={remoteKernelInstance}/>
```

Under the hood, the `WidgetDisplay` component will leverage three subcomponents: a `Manager`, a `Renderer` and a `BackboneViewer`.

### The Manager

[...]

PR #4622, opened 2019-10-18
Add support for rendering ipywidgets

captainsafia:
This PR is ready for review at the moment. Note that it only contains the presentational aspects, @miduncan will submit a PR with the kernel/comms related bits.

There are some follow-up items here, including:
- Fixing the rendering of font-awesome icons
- Fixing the rendering of elements with child_views, like accordions and tabs

IT'S ALIVE!

image


rgbkrk:
Wow!

wadethestealth:
I am truly amazed :+1:

captainsafia:
Shoutout to @miduncan for working with me to make this possible!

applications/desktop/webpack.common.js
```diff
@@ -63,6 +63,10 @@ const rendererConfig = {
       {
         test: /\.tsx?$/,
         use: [configurator.tsLoaderConfig]
+      },
+      {
+        test: /\.(jpg|png|gif)$/,
+        use: "file-loader"
```
rgbkrk, line 69: How do these get packaged up for a release of desktop?

babel.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  presets: ["@babel/preset-env"],
+  plugins: [["@babel/plugin-transform-runtime"]]
+};
```
rgbkrk, line 4: What necessitates bringing babel back?

packages/jupyter-widgets/__tests__/index.spec.tsx
```diff
@@ -4,11 +4,34 @@ import WidgetDisplay from "../src/index";
...
+      max: 10,
+      min: 0,
+      orientation: "horizontal",
+      readout: true,
+      readout_format: "d",
+      step: 1,
+      style: "IPY_MODEL_2e810dab21354e9381f53582e51e9a79",
+      value: 7
+    };
+    expect(<WidgetDisplay data={data} model={model} />).not.toBeNull();
```
rgbkrk: Would a snapshot have a little more to illustrate than a non-null check?

packages/jupyter-widgets/package.json
```diff
@@ -26,9 +26,10 @@
     "react": "^16.3.2"
   },
   "dependencies": {
-    "backbone": "^1.4.0"
+    "@jupyter-widgets/controls": "^1.5.2",
+    "font-awesome": "^4.7.0",
+    "jquery": "^3.4.1",
+    "jquery-ui": "^1.12.1"
   },
```
rgbkrk, line 33: Exciting. The 00s have come full circle.

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,7 +1,46 @@
...
+        {this.props.model ? (
+          <BackboneWrapper
+            model={this.props.model.get("state").toJS()}
+            manager={this.state.manager}
+            model_id={this.props.model_id}
+            widgetContainerRef={this.widgetContainerRef}
+          />
+        ) : null}
+      </React.Fragment>
+    );
```
rgbkrk, line 52: 🎉 

React connected to the backbone connected ipywidgets!

packages/jupyter-widgets/src/manager/manager.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
+  el: HTMLElement;
+
+  constructor(el: HTMLElement) {
+    super();
+    this.el = el;
+  }
+
+  loadClass(className: string, moduleName: string, moduleVersion: string): any {
```
rgbkrk, line 39: Is `loadClass` a standard thing within `WidgetManager`s or is this some niftiness you've put in?

packages/jupyter-widgets/src/manager/manager.ts
```diff
@@ -0,0 +1,94 @@
+import * as base from "@jupyter-widgets/base";
+import { DOMWidgetModel, DOMWidgetView } from "@jupyter-widgets/base";
+import * as controls from "@jupyter-widgets/controls";
+import { KernelMessage } from "@jupyterlab/services";
+
```
rgbkrk, line 5: My only big ask here is if you can comment a bit more on the bits pretty new to our code base, what you think we should know about this interconnect with the ipywidgets.

It could include links to the ipywidgets code base(s), just enough for other people to be able to assist in the future.

packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
```diff
@@ -1,12 +1,66 @@
...
+require("jquery-ui/themes/base/menu.css");
+require("jquery-ui/themes/base/progressbar.css");
+require("jquery-ui/themes/base/resizable.css");
+require("jquery-ui/themes/base/selectable.css");
+require("jquery-ui/themes/base/selectmenu.css");
+require("jquery-ui/themes/base/slider.css");
+require("jquery-ui/themes/base/sortable.css");
+require("jquery-ui/themes/base/spinner.css");
+require("jquery-ui/themes/base/tabs.css");
+require("jquery-ui/themes/base/tooltip.css");
```
rgbkrk: This is... quite a bit of CSS. Is it all scoped well? Are there globally set things we need to not have? Are there any selector conflicts?

I know that's a big ask to figure out. 😬  It's also the most likely head-scratcher later.

packages/notebook-app-component/src/status-bar.tsx
```diff
@@ -58,7 +58,7 @@ export class StatusBar extends React.Component<Props> {
       <Bar>
         <RightStatus>
           {this.props.lastSaved ? (
-            <p> Last saved {distanceInWordsToNow(this.props.lastSaved)} </p>
+            <p> Last saved {formatDistanceToNow(this.props.lastSaved)} </p>
```
rgbkrk: Do you want to make a new PR for this one?

scripts/test-shim.ts
```diff
@@ -17,6 +17,15 @@ global.Blob = function(content: any[], options: object): any {
...
 };
 
+/**
+ * Mock jQuery to allow jupyter-widgets tests to run.
+ */
+import $ from "jquery";
+global.$ = global.jQuery = $;
+require("jquery-ui");
+require("jquery-ui/ui/widget");
+require("jquery-ui/ui/widgets/mouse");
```
rgbkrk, line 27: and away we go

rgbkrk (approved):

packages/notebook-app-component/src/status-bar.tsx
rgbkrk: Unless this is one you can rebase out now that I see the conflict?

captainsafia: I rebased it out. Just had to change `this.props.lastSaved` to `new Date(this.props.lastSaved)`.

packages/jupyter-widgets/src/manager/manager.ts
captainsafia: Good call! I'll add some for the relevant view bits here.

@miduncan I saw you had some comments in your branch so we can set up some more here.

packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
captainsafia: Looking through most of the CSS in these files they all use class-based selectors for the styles. I'm not sure if there is a way to scope it to just this output using styled-components -- that would be ideal.

Also, I think we can condense this to just an import of `all.css` to reduce the clutter here.

packages/jupyter-widgets/src/index.tsx
```diff
@@ -8,12 +12,29 @@ interface JupyterWidgetData {
 
 interface Props {
   data: JupyterWidgetData;
+  model: any;
```
miduncan: You can use WidgetModel here instead of any

packages/jupyter-widgets/__tests__/renderer/backbone-wrapper.spec.tsx
```diff
@@ -1,13 +1,40 @@
...
+      _model_name: "IntSliderModel",
+      _view_count: null,
+      _view_module: "@jupyter-widgets/controls",
+      _view_module_version: "1.5.0",
+      _view_name: "IntSliderView",
+      continuous_update: false,
+      description: "Test:",
+      description_tooltip: null,
+      disabled: false,
+      layout: "IPY_MODEL_db277618a5c443009a6aa5b07f6b1812",
```
miduncan, line 21: When we integrate our stuff together, I believe this will throw an error because this model wont actually exist. I would delete the layout and style from this.

packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
miduncan: This might not be in this PR, but I was thinking maybe we could load these in the loadClass function instead so we only load what we need.

packages/notebook-app-component/src/status-bar.tsx
```diff
@@ -53,12 +53,17 @@ export class StatusBar extends React.Component<Props> {
...
     const name = this.props.kernelSpecDisplayName || "Loading...";
-
+    console.log(this.props.lastSaved);
     return (
       <Bar>
         <RightStatus>
           {this.props.lastSaved ? (
-            <p> Last saved {formatDistanceToNow(this.props.lastSaved)} </p>
+            <p>
+              {" "}
```
miduncan: Nit: I think padding or margin would be a better solution here

miduncan: Or at least a non-breaking space

miduncan (approved):

packages/notebook-app-component/src/status-bar.tsx
captainsafia: I think this was a weird formatting thing. The space wasn't needed here so I removed it altogether.

packages/jupyter-widgets/src/index.tsx
captainsafia: Nice tip! Thanks.

babel.config.js
captainsafia: The @jupyter-widgets/control package uses ES6 exports which Jest does not support. To resolve this, I had to update our Jest config to use babel to transpile ES6 modules.

We can remove this once Jest brings in native support for ES6 modules. That functionality is being tracked in https://github.com/facebook/jest/issues/4842.

packages/jupyter-widgets/src/manager/manager.ts
captainsafia: It's a standard part of the `WidgetManager`. This one is actually directly copied from the WidgetManager examples in the ipywidgets repo. [Here](https://github.com/jupyter-widgets/ipywidgets/blob/master/examples/web2/manager.js#L11) is a link,

packages/jupyter-widgets/__tests__/index.spec.tsx
captainsafia: Good call!

packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
captainsafia: @miduncan That's a good idea! The mapping between the CSS file and the view name won't be directly related and we'll have to sort through what views are related to which CSS but that would reduce the damage caused by loading all the CSS on the page.

packages/jupyter-widgets/__tests__/renderer/backbone-wrapper.spec.tsx
captainsafia: Got it. I'll remove these for now.

motin:
Amazing!! Great job!!
Linking back to https://github.com/nteract/nteract/issues/4573 so that this PR shows up in that thread

Issue #4561, opened 2019-09-09
Non DRY redux action definitions


Kethku:
**Application or Package Used**
actions package

**Describe the bug**
Currently the redux actions in the nteract actions package have repeat code in two ways:

First, the type string constant is repeated twice making the likely hood of an error higher.
Second, the action type definition is defined along with the return value shape of the action constructors.

I think the first problem could be fixed by replacing the string literal in the action constructors with {actionType} as {actionType} where {actionType} is the variable name of the type constant. This way the string is specified just once, the action type in the return value is a constant string literal, and the type string is ensured to be the same.

The second problem could be fixed by using the [Return helper type](https://github.com/pirix-gh/ts-toolbelt/blob/master/src/Function/Return.ts) from [ts-toolbelt](https://github.com/pirix-gh/ts-toolbelt) or something similar.

For bonus points, the two files in the action package could be merged since the type definitions would take up much less space with this refactoring. This would make jump to definition WAY more useful as all of the information for a given action will be located in the same place.

I propose the actions be defined like this:

``` typescript
export const TOGGLE_TAG_IN_CELL = "CORE/TOGGLE_TAG_IN_CELL";
export function toggleTagInCell(payload: {
  id: CellId;
  contentRef: ContentRef;
  tag: string;
}): actionTypes.ToggleTagInCell {
  return {
    // expectation is that if a tag doesn't exist, it will set it
    // if the tag is already in the collection of tags it will delete it
    type: actionTypes.TOGGLE_TAG_IN_CELL,
    payload
  };
}
export type ToggleTagInCell = Function.Return<typeof toggleTagInCell>;
```

I am happy to create a PR for this, but I wanted to get feedback on whether this would be a good idea or not. Further it is very possible one of these changes is more appealing than the other or neither of them! Let me know if this is something you folks would like.

captainsafia:
Tagging this for Hacktoberfest.

Hacktoberfest participants: we've already begun the work of cleaning up the repeative nature of type definitions in our actions. The pattern that we've adopted leverages the fact that you can access properties within a TypeScript interface. For example: 

```
export const setAppHost = (
  payload: { host: HostRecord }
): actionTypes.SetAppHostAction => ({
  type: actionTypes.SET_APP_HOST,
  payload
});
```

would be come

```
export const setAppHost = (
  payload: actionTypes.SetAppHostAction["payload"]
): actionTypes.SetAppHostAction => ({
[...]

PR #4623, opened 2019-10-18
Cleanup action definitions in hosts.ts

alpaix:
This PR promotes the clean concise action definition style to avoid code repetition as proposed in #4561. The new pattern is only applied to action definitions in `hosts.ts`.

Question to reviewers:
This file contains four bookstore actions and one host action which is not consistent. Would it make sense to split it?

 [x] I have read the [Contributor Guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md)


captainsafia:
Thanks for submitting this PR, @alpaix!

The bookstore actions should be in a separate file, IMO, especially because I'm thinking there is a few more host action creators that we can list.

If you have the ability to, feel free to create a new `bookstore.ts` file and match up the appropriate exports in this PR.

captainsafia (approved):
Thanks!

PR #4624, opened 2019-10-18
Fix appendOutput reducer not updating previous cells with the same display id

### LINKED ISSUE #4514: The `display-update` Example notebook is not working as described - display() without update=True does not update previous display outputs
I have reproduced this on:
- nteract desktop
- nteract on jupyter

**Describe the bug**
The `display-update` Example notebook is not working as described - display() without update=True does not update previous display outputs

**To Reproduce**

1. Open the `display-update` Example notebook
2. Run all commands up until and including `h2 = display('spoiler alert: output updated in both', display_id='some_destination')`

**Expected behavior**
The output should be updated in both destinations.

**Actual behavior**
The output is only updated in the cell directly below the command.

**Desktop:**
- OS: macOS 10.14.6
- Browser: nteract 0.14.5
- Version 0.14.5

**Additional context**
The subsequent command that uses display() with update=True works as intended.

The notebook in its entirety works as intended in the jupyter notebook environment using the same kernel.

CrystallineCat:
Fixes #4514.

### DIFF
--- a/packages/reducers/__tests__/document.spec.ts
+++ b/packages/reducers/__tests__/document.spec.ts
@@ -875,6 +875,29 @@ describe("updateDisplay", () => {
           transient: { display_id: "1234" }
         }
       }),
+    ];
+
+    const state = actionArray.reduce(
+      (s, action) => reducers(s, action),
+      originalState
+    );
+
+    expect(state.getIn(["notebook", "cellMap", id, "outputs"])).toEqual(
+      Immutable.List([
+        makeDisplayData({
+          output_type: "display_data",
+          data: { "text/plain": "shennagins afoot" },
+          metadata: Immutable.Map({})
+        }),
+        makeExecuteResult({
+          output_type: "execute_result",
+          data: { "text/plain": "shennagins afoot" },
+          metadata: Immutable.Map({})
+        })
+      ])
+    );
+
+    const moreActionArray = [
       actions.updateDisplay({
         content: {
           output_type: "update_display_data",
@@ -884,12 +907,12 @@ describe("updateDisplay", () => {
       })
     ];
 
-    const state = actionArray.reduce(
+    const moreState = moreActionArray.reduce(
       (s, action) => reducers(s, action),
-      originalState
+      state
     );
 
-    expect(state.getIn(["notebook", "cellMap", id, "outputs"])).toEqual(
+    expect(moreState.getIn(["notebook", "cellMap", id, "outputs"])).toEqual(
       Immutable.List([
         makeDisplayData({
           output_type: "display_data",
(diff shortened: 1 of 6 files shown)


CrystallineCat:
Unsure what causes the compiler to fail on CircleCI.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -17,10 +17,11 @@ import {
   insertCellAt,
   makeCodeCell,
   makeMarkdownCell,
-  makeRawCell,
+  makeRawCell, OnDiskDisplayData, OnDiskExecuteResult, OnDiskMediaBundle,
   OnDiskOutput,
   OnDiskStreamOutput
 } from "@nteract/commutable";
+import { JSONObject } from "@nteract/data-explorer/lib/types";
```
captainsafia: This is causing the build to fail. Try importing `JSONObject` from `@nteract/commutable` instead.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -204,6 +205,53 @@ function clearAllOutputs(
     .set("transient", transient);
 }
 
+// Utility function used in two reducers below
+function updateAllDisplaysWithID(
+  state: NotebookModel,
+  content: OnDiskExecuteResult | OnDiskDisplayData | {
```
captainsafia: Can we move the last type to a named interface?

captainsafia (commented):
Left some feedback and a note inline about why the build is failing. Thanks for submitting this PR!

packages/reducers/src/core/entities/contents/notebook.ts
rgbkrk: Guess the name could be `OnDiskOutput` or something?

rgbkrk: Though I guess there are only these three and they maybe only matter as types here.

`UpdatableOutputContent` as a name?

packages/reducers/src/core/entities/contents/notebook.ts
CrystallineCat: Thank you, that fixed it!

packages/reducers/src/core/entities/contents/notebook.ts
CrystallineCat: I moved the type to an interface, and added an alias for the three types. I moved the interface to @nteract/messaging, since that seems to be where the data originated from.

rgbkrk (approved):
Love it!

packages/reducers/__tests__/document.spec.ts
```diff
@@ -875,6 +875,29 @@ describe("updateDisplay", () => {
...
+          data: { "text/plain": "shennagins afoot" },
+          metadata: Immutable.Map({})
+        }),
+        makeExecuteResult({
+          output_type: "execute_result",
+          data: { "text/plain": "shennagins afoot" },
+          metadata: Immutable.Map({})
+        })
+      ])
+    );
```
rgbkrk, line 898: Especially happy about extra tests. 😄

rgbkrk (approved):

PR #4628, opened 2019-10-21
webpack-dev-server package update

miduncan:


captainsafia (approved):
Thanks for the fix!

Context for others: the version of the webpack-dev-server we were using caused the dev server for Jext to fail due to incompatible dependencies in Webpack.

PR #4635, opened 2019-10-24
Fixed executionCounts operator

### LINKED ISSUE #4541: Execution count not updating when kernel does not send execute_input message
Currently, in implementations of the Q# and C# kernels for Jupyter, the kernel does not respond back with an `execute_input` message to execution requests, just an `execute_reply` message. Currently, the `executionCounts` operator in nteract only looks for execution counts set on the `execute_reply` message. The protocol seems to be ambiguous on which one is required so we should update the `executionCounts` operator to check for both.

ensigntepth:

Closes #4541

### DIFF
--- a/packages/messaging/__tests__/messaging-spec.ts
+++ b/packages/messaging/__tests__/messaging-spec.ts
@@ -315,7 +315,6 @@ describe("executionCounts", () => {
   });
 });
 
-
 describe("kernelStatuses", () => {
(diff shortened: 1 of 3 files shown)


captainsafia (approved):
Nice! Thanks for adding the tests as well!

rgbkrk:
Thank you!

PR #4636, opened 2019-10-24
Upgrade react-syntax-highlighter in presentational-components

akx:

---

# What?

This PR upgrades the [`react-syntax-highlighter`](https://github.com/conorhastings/react-syntax-highlighter) package from 9.x to 11.x, heeding the upgrade notes from 9.x to 10.x.

# Why?

To get rid of `babel-runtime` dependencies.

Our (@valohai) downstream project uses `@nteract/presentational-components` and `@nteract/outputs`, and we end up with an unnecessary `babel-runtime` in our bundle, which makes us, bandwidth bills and users with slow connections unnecessarily sad.

After this change, there's still a couple places where `babel-runtime` lurks: 
```
=> Found "babel-runtime@6.26.0"
info Reasons this module exists
   - "_project_#@nteract#outputs#react-json-tree" depends on it
   - Hoisted from "_project_#@nteract#outputs#react-json-tree#babel-runtime"
   - Hoisted from "_project_#@nteract#data-explorer#semiotic#roughjs-es5#babel-runtime"
```

As for that `react-json-tree` dep, their `master` already has a version depending on a newer Babel, but it's not released; I've raised an issue: https://github.com/reduxjs/redux-devtools/issues/490

As for `semiotic` using `roughjs-es5`, it looks like [Semiotic 2](https://github.com/nteract/semiotic/tree/semiotic-2) will be [making the Rough dependency optional](https://github.com/nteract/semiotic/commit/ebc2e270477f8d33f4e2dff5dd72e42f5d302876), so whenever that's done, no more `babel-runtime`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,8 @@
       "^@nteract/vega-embed-v3": "@nteract/vega-embed-v3",
       "^@nteract/([^/]*)$": "<rootDir>/packages/$1/src",
       "^(ansi-to-react|enchannel-zmq-backend|fs-observable|rx-binder|rx-jupyter)$": "<rootDir>/packages/$1/src",
-      "\\.css$": "<rootDir>/scripts/noop-module.ts"
+      "\\.css$": "<rootDir>/scripts/noop-module.ts",
+      "react-syntax-highlighter/dist/esm/styles/prism/.+": "<rootDir>/scripts/noop-module.ts"
     },
     "coveragePathIgnorePatterns": [
       "<rootDir>/scripts",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -104,7 +104,8 @@
       "^@nteract/vega-embed-v3": "@nteract/vega-embed-v3",
       "^@nteract/([^/]*)$": "<rootDir>/packages/$1/src",
       "^(ansi-to-react|enchannel-zmq-backend|fs-observable|rx-binder|rx-jupyter)$": "<rootDir>/packages/$1/src",
-      "\\.css$": "<rootDir>/scripts/noop-module.ts"
+      "\\.css$": "<rootDir>/scripts/noop-module.ts",
+      "react-syntax-highlighter/dist/esm/styles/prism/.+": "<rootDir>/scripts/noop-module.ts"
```
rgbkrk, line 108: 👍

rgbkrk (approved):
Thanks for doing this upgrade work!

wadethestealth:
This is causing people unable to install hydrogen nteract/hydrogen#1804. I need this to be rolled back until a solution is provided to allow users to install hydrogen.

PR #4639, opened 2019-10-24
Add support for kernels comms in ipywidgets

miduncan:
This adds the widget manager and the implemented comms interface that allows the widgets to talk to the back end. This allows for basic interaction with the widget and widget attributes, but there is still some work left to be done:

- Style/Layout settings
- Handling of certain comm messages (clear_output, display_data, etc.)


packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
```diff
@@ -29,14 +29,18 @@ interface Props {
 }
 
 export default class BackboneWrapper extends React.Component<Props> {
+  created = false;
```
miduncan, line 32: This will be changed in a later PR

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
-import { WidgetModel } from "@jupyter-widgets/base";
+import { connect } from "react-redux";
+import { Dispatch } from "redux";
+import { AppState, selectors, KernelNotStartedProps, LocalKernelProps, RemoteKernelProps } from "@nteract/core";
+import { commOpenAction, commMessageAction } from "@nteract/actions";
 
 interface Props {
-  model: WidgetModel;
-  model_id: string;
+  model: any;
```
captainsafia: Looks like the strict types here got reverted. Can we change these to `WidgetModel` and string respectively?

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
+  model_id: any;
+  model_lookup_by_id: (id: string) => any;
```
captainsafia: Major nit: Can we rename this to `modelById` everywhere? It follows the convention used elsewhere in the codebase.

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
+  kernel?: RecordOf<KernelNotStartedProps> | RecordOf<LocalKernelProps> | RecordOf<RemoteKernelProps> | null | undefined
```
captainsafia: Nit: We can remove the `undefined` at the end since we already have the `?`.

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
 
-/**
- * This component is is a wrapper component that initializes a
- * WidgetManager singleton and passes a model reference to the
- * BackboneModelWrapper. It's doing most of the heavy lifting with
- * respect to bridging the kernels comms that the WidgetManager provides,
- * our client-side state model, and the view.
- */
-export default class Manager extends React.Component<Props, State> {
+class Manager extends React.Component<Props, State> {
```
captainsafia: Can we keep the documentation string?

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
+  getManager(){
+    if(Manager.manager == undefined){
+      Manager.manager = new WidgetManager(this.props.kernel, this.props.model_lookup_by_id)
+    }else{
+      Manager.manager.update(this.props.kernel, this.props.model_lookup_by_id);
     }
+    return Manager.manager;
+  }
+
+  getManager2(){
```
captainsafia: Can this be removed?

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
+      manager: new WidgetManager(this.props.kernel, this.props.model_lookup_by_id)
     };
   }
 
-  componentDidMount() {
-    if (this.widgetContainerRef && this.widgetContainerRef.current !== null) {
-      this.setState({
-        manager: new WidgetManager(this.widgetContainerRef.current)
-      });
+  getManager(){
```
captainsafia: Can you add a doc string to this method explaining what it is doing?

packages/jupyter-widgets/src/manager/widget-comms.ts
```diff
@@ -0,0 +1,150 @@
...
+    kernel: any;
+
+    /**
+     * 
+     * @param comm_id 
+     * @param target_name 
+     * @param target_module 
+     * @param kernel 
+     */
+    constructor(comm_id: string, target_name:string, target_module: string, kernel: any){
```
captainsafia: Can you update the doc string to explain what each constructor parameter is used for?

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -0,0 +1,201 @@
...
+  get_model(model_id: string): Promise<WidgetModel> | undefined {
+    let model = super.get_model(model_id);
+    if(model === undefined){
+      let model_state = this.model_comm_lookup(model_id).state;
+      model = this.new_widget_from_state_and_id(model_state, model_id);
+    }
+    return model;
+  }
+
+  async new_model_from_state_and_id(state: any, model_id: string){
```
captainsafia: Can you add a doc string clarifying what this is used for?

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -1,54 +1,67 @@
...
+    if(Manager.manager == undefined){
```
captainsafia: I think this can just be `if (!Manager.manager)` to capture any non-truth values.

captainsafia (commented):
Left some feedback inline.

What kind of tests can we write for this?

packages/jupyter-widgets/src/manager/index.tsx
miduncan: I can do this here, but personally I don't think we should do that in the WidgetManager class because it already has a getModel(id) function which gives what it calls a model, so it would be super confusing.

packages/jupyter-widgets/src/manager/index.tsx
miduncan: Good catch

packages/jupyter-widgets/src/manager/index.tsx
captainsafia: Good point! Perhaps `stateModelById` would be better to indicate that this is coming from the nteract core state?

packages/jupyter-widgets/src/manager/index.tsx
miduncan: It's only meant to be a check for if the static variable hasn't been touched. If the type gets changed to be nullable in the future, I think we should not be overriding it

rgbkrk:
Reviewing now!

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -19,36 +34,52 @@ interface State {
...
-    };
   }
 
-  componentDidMount() {
-    if (this.widgetContainerRef && this.widgetContainerRef.current !== null) {
-      this.setState({
-        manager: new WidgetManager(this.widgetContainerRef.current)
-      });
+  /**
+   * Because the iPyWidgets keeps track of the widgets it creates as a
```
rgbkrk, line 46: Replace `iPyWidgets` with `IPyWidgets`

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -19,36 +34,52 @@ interface State {
...
+   * member variable, the WidgetManager needs to be treated like a singleton.
+   * However, we still need to be constantly updating the singleton with the most up
+   * to date modelById function, otherwise it will be searching a stale state for a
+   * model
+   */
```
rgbkrk, line 51: Thanks for adding the comments in here to explain this.

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -19,36 +34,52 @@ interface State {
...
-        {this.props.model ? (
-          <BackboneWrapper
-            model={this.props.model.get("state").toJS()}
-            manager={this.state.manager}
-            model_id={this.props.model_id}
-            widgetContainerRef={this.widgetContainerRef}
-          />
-        ) : null}
+        <BackboneWrapper
+          model={this.props.model.get("state").toJS()}
```
rgbkrk, line 68: Note -- this will cause a re-render every time because this creates a new object from the original immutable object every time. We should re-think what we're putting in the state tree or consume the immutable object more directly.

Thoughts @captainsafia? ^^

packages/jupyter-widgets/src/manager/widget-comms.ts
```diff
@@ -0,0 +1,197 @@
+import { IClassicComm } from "@jupyter-widgets/base";
+import {
+  createCommMessage,
+  createCommOpenMessage
+} from "@nteract/epics/lib/comm";
```
rgbkrk, line 5: This seems like something where we should put `createCommMessage` and `createCommOpenMessage` should be exported from the top so we don't have to dig into `lib`

rgbkrk:
I'm going to go ahead with merging, even with my comments and questions inline. Thanks for pushing this further and further along.

rgbkrk:
Whoops. Safia told me this wasn't meant to be merged yet. Didn't realize it was a WIP.

packages/jupyter-widgets/src/manager/index.tsx
captainsafia: Makes since. I think consuming the immutable object more directly would make since. This is actually a bit of code that I wrote so I've created and self-assigned an issue at https://github.com/nteract/nteract/issues/4643.

wadethestealth: You could have a comparison algorithm in react's should render if you prefer to use this how it is.

PR #4651, opened 2019-10-28
Fix error when receiving a display update on a non-existing keypath

### LINKED ISSUE #4515: The `display-update` Example notebook hangs when creating an instance of AutoupdatingProgressBar

I have reproduced this on:
- nteract desktop
- nteract on jupyter

**Describe the bug**

The `display-update` Example notebook hangs when creating an instance of AutoupdatingProgressBar

**To Reproduce**

1. Open the `display-update` Example notebook
2. Run all commands up until and including:
```
better_bar = AutoupdatingProgressBar(100)
better_bar.display()
```

**Expected behavior**
A progress bar should be displayed.

**Actual behavior**
The Python 3 kernel hangs indefinitely. It can't be killed nor restarted. Nteract needs to be shut down using force quit.

**Desktop:**
- OS: macOS 10.14.6
- Browser: nteract 0.14.5
- Version 0.14.5

**Additional context**
The notebook in its entirety works as intended in the jupyter notebook environment using the same kernel.

CrystallineCat:
Fixes #4515.

### DIFF
--- a/packages/reducers/__tests__/document.spec.ts
+++ b/packages/reducers/__tests__/document.spec.ts
@@ -853,6 +853,28 @@ describe("appendOutput", () => {
 });
 
 describe("updateDisplay", () => {
+  test("handles a non-existent keypath gracefully", () => {
+    const originalState = monocellDocument;
+
+    const id = originalState.getIn(["notebook", "cellOrder", 2]);
+
+    const action = actions.updateDisplay({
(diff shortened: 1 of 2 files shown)


captainsafia (approved):
Good catch! Was actually debugging an issue recently that might be related to this. We'll see if it resolve the problem I was running into.

PR #4652, opened 2019-10-28
Pass Immutable objects through React components

### LINKED ISSUE #4643: Rely on Immutable structure within BackboneWrapper
Follow-up to [this](https://github.com/nteract/nteract/pull/4639#discussion_r339391120).

Currently, we convert the Immutable model to a JavaScript object which forces unnecessary renderers. We should modify this.

captainsafia:
Closes https://github.com/nteract/nteract/issues/4643.

@miduncan @rgbkrk Thoughts on relying on Immutable structures more deeply in the WidgetManager implementation? One the one hand, we get the benefit of immutable structures throughout the manager class. On the other hand, we'd have to rewrite more methods that are currently implemented by the `ManagerBase` like `new_widget`.

### DIFF
--- a/packages/jupyter-widgets/src/manager/index.tsx
+++ b/packages/jupyter-widgets/src/manager/index.tsx
@@ -65,7 +65,7 @@ class Manager extends React.Component<Props> {
     return (
       <React.Fragment>
         <BackboneWrapper
-          model={this.props.model.get("state").toJS()}
+          model={this.props.model.get("state")}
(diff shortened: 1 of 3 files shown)


miduncan:
Immutability is nice, but I'm not sure it's worth sacrificing resiliency to updates by copy/pasting the methods and changing them to use immutables. If it's just new_widget, though, I think it is fine.

rgbkrk (approved):

PR #4654, opened 2019-10-29
Add default kernel config option

### LINKED ISSUE #620: Setup a default kernel based on user

### LINKED ISSUE #4608: Add an option to configure a default runtime
This is about the windows desktop app.

I'm a julia user, and use nteract only for julia. It would be great if I could make the julia kernel my default kernel, so that any new notebook I create on my machine will use that kernel and I don't have to manually switch things over to julia for every new notebook I create.

CrystallineCat:
Fixes #4608. Fixes #620.

### DIFF
--- a/applications/desktop/src/main/index.ts
+++ b/applications/desktop/src/main/index.ts
@@ -97,7 +97,9 @@ const fullAppReady$ = zip(electronReady$, prepareEnv).pipe(first());
 const jupyterConfigDir = join(app.getPath("home"), ".jupyter");
 const nteractConfigFilename = join(jupyterConfigDir, "nteract.json");
 
-const CONFIG = {};
+const CONFIG = {
+  defaultKernel: "python3",
+};
 
 const prepJupyterObservable = prepareEnv.pipe(
   mergeMap(() =>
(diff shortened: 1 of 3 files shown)


applications/desktop/src/main/index.ts
```diff
@@ -274,16 +282,19 @@ openFile$
...
+        const defaultKernel = CONFIG.defaultKernel;
 
-        if (argv.kernel in specs) {
-          kernel = argv.kernel;
-        } else if ("python2" in specs) {
-          kernel = "python2";
+        if (passedKernel && passedKernel in specs) {
+          kernel = passedKernel;
+        } else if (defaultKernel && defaultKernel in specs) {
+          kernel = defaultKernel;
```
rgbkrk, line 291: Seems like a decent fallback here.

applications/desktop/src/main/menu.ts
```diff
@@ -542,6 +547,15 @@ export function loadFullMenu(store = global.store) {
       {
         label: "Editor options",
         submenu: blink_menu
+      },
+      {
+        label: "Default kernel",
```
rgbkrk: I'm a little curious about making this "Set default kernel" as the label. That's not a blocker on merging though.

applications/desktop/src/notebook/menu.ts
```diff
@@ -555,8 +561,8 @@ export function dispatchNewNotebook(
   store: DesktopStore,
   event: Event,
   filepath: string | null,
-  kernelSpec: KernelSpec
-) {
+  kernelSpec?: KernelSpec,
```
rgbkrk: What makes this optional now?

applications/desktop/src/main/menu.ts
CrystallineCat: Yeah, that might be a better label. I changed it.

applications/desktop/src/notebook/menu.ts
CrystallineCat: Thanks for catching that! I tried putting the default kernel resolution in the reducer, with the caller passing undefined unless a specific kernel was requested at the command line. This is a holdover from that. I removed the question mark.

rgbkrk:
I think this looks good. There's a merge conflict you'll need to resolve in order to merge this. 😬

CrystallineCat:
Thank you! 😀

One day, I will successfully fix a merge conflict without needing to hard reset and force push... One day! 🤕

rgbkrk:
Force push is the only way to remedy it on your branch, it's a-ok. You're making new commits that match the updated history. 😄

Issue #4644, opened 2019-10-28
Widget styles and layouts not applied to ipywidgets


captainsafia:
In the Widget Styling example notebook from the ipywidgets repo, none of the layouts and styles applied to the widgets render on the page. We need to support layout and styling for widgets.

PR #4655, opened 2019-10-29
Fix for ipywidgets style and layout

miduncan:
Issue: https://github.com/nteract/nteract/issues/4644 
This change enables the style and layout attributes to be used on ipywidgets. For example:

```
import ipywidgets as widgets
w = widgets.IntSlider()
w.layout.border='1px solid red'
w.style.handle_color='blue'
display(w)
```
will now produce something like this

image

and update with comms


packages/jupyter-widgets/__tests__/manager/manager.spec.ts
```diff
@@ -40,12 +40,86 @@ describe("WidgetManager", () => {
...
+      );
+      const view = await manager.create_view(
+        widget,
+        {},
+        document.createElement("div")
+      );
+      manager.render_view(view);
+      // render_view sends out an event that the view waits for before it it sets the style and layout promise.
+      // this means we need to wait for the callbacks to finish before we check their value
+      setTimeout(async () => {
```
miduncan, line 118: Any ideas on a better way to do this? I need to wait for the callbacks to finish before I check the values. I could subscribe to the event, but I'm not sure I could guarantee that my callback won't come before or at the same time as the callback I need to let finish.

captainsafia: cc: @rgbkrk Any thoughts on this?

packages/jupyter-widgets/__tests__/manager/manager.spec.ts
```diff
@@ -40,12 +40,86 @@ describe("WidgetManager", () => {
...
+      const widget = await manager.new_widget_from_state_and_id(
+        model,
+        "test_model_id"
+      );
+      const view = await manager.create_view(
+        widget,
+        {},
+        document.createElement("div")
+      );
+      expect(view).not.toBeNull();
```
captainsafia, line 52: Can we add a snapshot test here as well?

packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
```diff
@@ -31,15 +31,20 @@ interface Props {
...
-          el: widgetContainerRef.current
-        });
-        widget.render();
+        const widget = await manager.new_widget_from_state_and_id(
+          model,
+          model_id
+        );
+        const view = await manager.create_view(
+          widget,
+          {},
```
captainsafia: Nit: Can you post a comment clarify that this empty object is the options.

packages/jupyter-widgets/__tests__/manager/manager.spec.ts
rgbkrk: In an ideal world, you could schedule the events like one would with the Rx test scheduler. Alas, you're at the whims of the ipywidgets events. What are the callbacks or events that need to be waited for?

It's ok to try following the callbacks, waiting for them all to resolve. It's ok for a test to fail because events don't fire -- that's par for the course in our async world.

packages/jupyter-widgets/src/renderer/backbone-wrapper.tsx
captainsafia: Also: Since create view can throw an error, can we add a try-catch to this?

captainsafia (commented):
Looks good overall!

Posted some comments about tests and better error handling in-line.

packages/jupyter-widgets/__tests__/manager/manager.spec.ts
miduncan: My code comment gives a brief overview, but essentially ipywidgets wants a "displayed" event fired after the widget is rendered, after which it will instantiate the layout and style widgets. I have to wait for their callback to finish before I check their values. To clarify, though, I'm the one firing this event and doing so unconditionally, so what I'm really trying to test is that the ipywidgets callbacks complete successfully.
When you say follow and wait for the callback, what does that entail code-wise?

packages/jupyter-widgets/__tests__/manager/manager.spec.ts
miduncan: I think the only value of a snapshot here would be to check that the state you passed made it to the end result, so I added a test that value is 7. I have a work item for writing ipywidgets tests, so we can revisit this

rgbkrk (approved):

rgbkrk:
I see the WIP tag is off now, so feel free to merge when you are ready.

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -129,9 +129,15 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
     view: base.DOMWidgetView,
     options: any
   ): Promise<base.DOMWidgetView> {
+    this.render_view(view);
     return Promise.resolve(view);
   }
 
+  render_view(view: base.DOMWidgetView): void {
+    view.render();
+    view.trigger("displayed");
```
captainsafia, line 138: Learned from @jasongrout that this is the only line in the PR we probably need since it invokes the displayed callback in the base widget model which then renders the styles and layout.

Can you test to see if only adding this will did the problem as opposed to the other changes?

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -129,9 +129,15 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
     view: base.DOMWidgetView,
     options: any
   ): Promise<base.DOMWidgetView> {
+    this.render_view(view);
```
captainsafia, line 132: Also learned that we shouldn’t be rendering the view in the create_view method so this will need to be moved elsewhere.

miduncan: This is not create_view, this is display_view

packages/jupyter-widgets/src/manager/widget-manager.ts
miduncan: This is the only change needed to trigger the callbacks, but the other change is still necessary. The change to create_view was needed because you had overloaded the method to also create a model, which was problematic because create_view was being called from within ipywidgets for the layout and style widgets who already had a model and only needed a view created.

miduncan: Aside from possibly the yarn.lock. That one updated on its own when I ran `yarn` after merging with master

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -173,40 +179,31 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
...
+    el?: HTMLElement | null
+  ): Promise<DOMWidgetView> {
+    return super
+      .create_view(model, options)
+      .then(view => {
+        if (el) {
+          //setting the element has to be done this way, not using setElement()
+          //otherwise it somehow breaks the interaction between mutliple views on
+          //the same widget model
+          view.el = el;
```
captainsafia, line 203: Can you clarify what is happening here a little more?

Also, we need to keep the reference to our custom `loadClass` here. The work to support Output displays requires us resolving our own OutputModel and OutputView from the `loadClass` method invoked via create_view.

miduncan: Essentially, this is just "extending" (overloaded term, but I think it applies) the `create_view` function in ManagerBase to assign the element. In the `ManagerBase.create_view`, it will still call our custom `loadClass` function.

I strongly believe we should use this function because it does some some stuff for us that we weren't doing like

- Assigning a reference to the view in the given model
- Deletes the view when a the model is destroyed (subscribes to a "destroyed" event in the model)

miduncan: I'll change the docstring to include more detail

captainsafia: Sounds good. Thanks for clarify and extending upon the points here.

Can you add the text you posted in this GitHub comment as documentation in the docstring for this method?

Other than that, this is good to merge.

PR #4660, opened 2019-10-30
Set HostRef in RemoteKernelRecord on kernel launch

### LINKED ISSUE #4627: Jupyter Server endpoint associated with kernel
We have a scenario where we need to be able to easily map the jupyter server endpoint of the host to the associated kernel. 
Scenario: pick which contentRef by the related kernelRef that had the endpoint of the selected host.

Right now there is no mapping maintained in the kernelRef, and a non-consistent workaround (within channels->_source).

Example of a state that would solve this scenario for us:
image

captainsafia:
Closes #4627.

**Changes in this PR**
- Sets HostRef in RemoteKernelRecord after launch of WebSocket kernel
- Adds selectors for accessing hosts and retrieving host record by host ref and vice versa


rgbkrk (approved):
Nice and simple.

Issue #4665, opened 2019-10-30
Support showing/hiding all cell inputs/outputs independently


alpaix:
This issue affects `@nteract/actions` package. Currently, `UNHIDE_ALL` action unconditionally resets all cell inputs and outputs visibility. There is no support for independently changing visibility of inputs or outputs.

**Expected behavior**
We need a method of only changing visibility of all cell inputs or outputs independently. For instance, when getting asked to hide all cell outputs, the `UNHIDE_ALL` action should keep visibility of cell inputs untouched.

captainsafia:
@alpaix This is now released in the latest version of @nteract/core.

PR #4667, opened 2019-11-01
Tweak unhideAll to change visibility of all inputs or outputs

alpaix:
The purpose of this PR is to allow `unhideAll` action to alter visibility of all cell inputs and outputs independently. 

Key changes are:
- `inputHidden` and `outputHidden` attributes will be optional
- `unhideAll` reducer will test attributes values to be not `undefined` before applying the change

Resoves #4665

### DIFF
--- a/packages/actions/src/actions/cells.ts
+++ b/packages/actions/src/actions/cells.ts
@@ -196,11 +196,9 @@ export function updateCellExecutionCount(payload: {
   return setInCell({ ...payload, path: ["execution_count"] });
 }
 
-export function unhideAll(payload: {
-  outputHidden: boolean;
-  inputHidden: boolean;
-  contentRef: ContentRef;
-}): actionTypes.UnhideAll {
+export function unhideAll(
+  payload: actionTypes.UnhideAll["payload"]
+): actionTypes.UnhideAll {
   return {
     type: "UNHIDE_ALL",
     payload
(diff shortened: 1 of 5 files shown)


packages/actions/src/actions/cells.ts
```diff
@@ -196,11 +196,9 @@ export function updateCellExecutionCount(payload: {
...
 }
 
-export function unhideAll(payload: {
-  outputHidden: boolean;
-  inputHidden: boolean;
-  contentRef: ContentRef;
-}): actionTypes.UnhideAll {
+export function unhideAll(
+  payload: actionTypes.UnhideAll["payload"]
+): actionTypes.UnhideAll {
```
captainsafia, line 201: Thanks for making this incremental clean-up!

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -655,15 +655,19 @@ function unhideAll(
   state: NotebookModel,
   action: actionTypes.UnhideAll
 ): RecordOf<DocumentRecordProps> {
+  const metadataMixin: any = {};
```
captainsafia: Nit: Can we add a strict type here? It'll probably be helpful as we move to resolve the to-do in this code to ensure that we are parsing input/output hidden properties from different front-ends correctly.

captainsafia (approved):
Looks good! Left one comment in-line that would be good to resolve before merging.

packages/reducers/src/core/entities/contents/notebook.ts
alpaix: Done.

PR #4664, opened 2019-10-30
Revert react-syntax-highlighter upgrade in #4636

captainsafia:
This reverts commit c34b1aedbd358e3b7a6229656c5365be742bcacd, reversing
changes made to d08dbe1664a8e5b6691d490ceb4d3a531310e0e6.

Issue #4663, opened 2019-10-30
@nteract/presentational-components ships with ES5 incompatible dependencies


mpacer:
**Application or Package Used**
@nteract/presentational-components

**Describe the bug**
@nteract/presentational-components 3.0.8 includes dependencies that are not ES5 compatible, specifically by including react-syntax-highlighter which has explicit `export` statements 
```
/apps/commuter/node_modules/react-syntax-highlighter/dist/esm/styles/prism/vs.js:1
export default {
^^^^^^

SyntaxError: Unexpected token export
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/apps/commuter/node_modules/@nteract/presentational-components/lib/syntax-highlighter/index.js:15:30)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
```

I think this may be related

https://github.com/nteract/nteract/blob/cbeb83b1abe8c0c3227945b8ab5c124e9489893e/package.json#L108

wadethestealth:
This should be solved by the latest patch.

PR #4669, opened 2019-11-02
Upgrade react-syntax-highlighter in presentational-components (again, with less problems, hopefully)

akx:
Ugh, sorry about the mess I brought on in #4636 (refs #4663, #4664).
I was actually bit by it personally too, now that the upgraded module was published. 😅 

This is essentially the same PR as #4636, but without ESM imports; I just vendored in those troublesome theme modules instead.


wadethestealth:
Before merging, I need to test that this would work in Hydrogen. ~I will update this comment with the results.~

~Results: https://github.com/nteract/nteract/pull/4669#discussion_r341831963~ My mistake.
Results: works and specs pass

packages/presentational-components/src/syntax-highlighter/index.tsx
```diff
@@ -1,11 +1,7 @@
 import * as React from "react";
-import SyntaxHighlighter from "react-syntax-highlighter/prism";
-import {
-  vs as lightTheme,
-  xonokai as darkTheme
-} from "react-syntax-highlighter/styles/prism";
-
-// import syntax from "./theme";
+import {Prism as SyntaxHighlighter} from "react-syntax-highlighter";
```
rgbkrk, line 2: One way of guaranteeing this working is by using the [direct cjs import the library authors recommend (though stated as non-ideal)](https://github.com/conorhastings/react-syntax-highlighter/blob/d58813a0f2e2fd4ba94e5c1122e1ddb9ca01f67f/CHANGELOG.MD#changed-1):

```
import { Prism as SyntaxHighlighter } from "react-syntax-highlighter/dist/cjs/prism";
```

wadethestealth, line 2: Currently, this dosn't even build. To test builds please run `tsc -b packages/presentational-components` to test bulding
```
packages/presentational-components/src/syntax-highlighter/index.tsx:2:9 - error TS2305: Module '"react-syntax-highlighter"' has no exported member 'Prism'.

2 import {Prism as SyntaxHighlighter} from "react-syntax-highlighter";
```

wadethestealth (changes requested):

packages/presentational-components/src/syntax-highlighter/index.tsx
akx: Wait, um, I'm not seeing that (and CI passes too)...

```shell
$ git rev-parse HEAD
b5ed3382ab2ea9085d203a1e6c85d2e82ba67150
$ ./node_modules/.bin/tsc -b packages/presentational-components
$
```

wadethestealth: The ci passes because this module has no tests

wadethestealth: I think that was my mistake though let me try again.

akx: `react-syntax-highlighter`'s `package.json` has 
```json
  "main": "dist/cjs/index.js",
  "module": "dist/esm/index.js",
```
so bundlers/runtimes will know to use the ESM build if they're ESM aware, and use the CJS build otherwise.

The problem was with the direct imports of `esm` style theme modules which didn't get transpiled and thus were not importable.

As for the theme modules, sure, they could be imported `cjs` style, so they wouldn't need to get bundled, at the likely expense of a little more boilerplate in the code here. I can do that, too, if that's preferred over vendoring the themes.

wadethestealth: Ok that was totally my fault. It does build, but we do need to be careful because the ci does not test this module

wadethestealth: I think importing a cjs rather then updating a vender is better.

akx: Okiedoke, changed that. Did mean I had to add an "ad-hoc" `.d.ts` file for the two modules, but it's explained in the commit message.

wadethestealth (approved):

akx:
Thanks! Here's hoping I don't break anything this time...

PR #4671, opened 2019-11-03
Check cwd for suitability before using it as document directory

### LINKED ISSUE #4607: Default working dir on Windows shouldn't be install dir
This is about the nteract Windows app. When I start it, it creates a new notebook, and the path for that is `C:\Program Files\nteract\Untitled.ipynb`. That is not great, because I of course don't have write access to that folder and it is in general a location where one shouldn't save user files.

I think the ideal solution would be that if I create a new file it doesn't have any path associated it, only once I save it does it get one. But if it does need one, then it should probably be a file in the current users Documents folder, not the install directory.

CrystallineCat:
Fixes #4607 (hopefully, don't have Windows to test).

### DIFF
--- a/applications/desktop/src/notebook/menu.ts
+++ b/applications/desktop/src/notebook/menu.ts
@@ -22,7 +22,7 @@ function isWriteable(pathToCheck: string): boolean {
 
 function getDocumentDirectory(): string {
   const cwd = path.normalize(process.cwd());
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):

rgbkrk:
Thank you @CrystallineCat!

PR #4674, opened 2019-11-04
Increase text contrast, add test, remove obsolete code

CrystallineCat:
- Adds a test that reads the CSS to check contrast for WCAG AAA (as far as possible).
- Changes theme styles to conform to WCAG AAA.
- Removes obsolete code (`GlobalCSSVariables`) that appears to have been moved to `global-variables.css` and doesn't appear to be referenced anywhere in the repo.


captainsafia (approved):
Great clean-up! Thanks,

PR #4678, opened 2019-11-06
Fix for Box based widgets

### LINKED ISSUE #4640: Box-based ipywidgets do not render
@nteract/jupyter-widgets

**Describe the bug**
The current implementation of the jupyter-widgets package does not render Box-based widgets like Box, VBox, and HBox.

**To Reproduce**

1. Create a new notebook.
2. Attempt to create a box widget. Note that no widget is not rendered.

```
import ipywidgets
items = [ipywidgets.Label(str(i)) for i in range(4)]
ipywidgets.Box(items)
```

**Expected behavior**
Layout-based widgets should render.

master

miduncan:
Fixes #4640 
```
import ipywidgets as widgets
items = [widgets.IntSlider() for i in range(4)]
widgets.Box(items)
```
now outputs
image

### DIFF
--- a/packages/jupyter-widgets/src/manager/widget-manager.ts
+++ b/packages/jupyter-widgets/src/manager/widget-manager.ts
@@ -1,6 +1,7 @@
 import { KernelMessage } from "@jupyterlab/services";
 import * as base from "@jupyter-widgets/base";
 import * as controls from "@jupyter-widgets/controls";
+import * as pWidget from "@phosphor/widgets";
 import {
   DOMWidgetView,
   WidgetModel,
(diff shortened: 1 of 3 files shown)


packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -124,18 +125,30 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
...
-  render_view(view: base.DOMWidgetView): void {
-    view.render();
-    view.trigger("displayed");
+  /**
+   * Render the given view to a target element
+   * @param view View to be rendered
+   * @param el Target element that the view will be rendered within
+   */
+  render_view(view: base.DOMWidgetView, el: HTMLElement): void {
+    pWidget.Widget.attach(view.pWidget, el);
```
captainsafia, line 151: Do we have to transfer all rendering control to Phosphor? I'd prefer if we kept as much of it in React as possible.

It looks like @williamstein started working on creating [a set of React controls](https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/jupyter/widgets/controls.ts) that are [used](https://github.com/sagemathinc/cocalc/blob/master/src/smc-webapp/jupyter/widgets/manager.ts#L396) to render the Box-based views. Can we take a similar approach?

miduncan: Unfortunately, the Box widget seems to be intentionally designed to not work any other way. It creates its own Phosphor element on initialization and has an explicit check in its `setElement` function that the element is not changed to anything else. Even if you pass the element in the constructor, it will still reach this check and throw an error.

We could look at designing our own Box based views (we would actually need to replace a bit more than that to completely get rid of the dependency), but I'm not sure it's a good investment of our time this sprint.

miduncan: I take that back-- I don't really know the whole picture. What negatives does the existing Phosphor dependency bring with it?

captainsafia (approved):
Let's settle for this then replace the Box-based layouts with React controls in a follow-up.

PR #4680, opened 2019-11-06
Process ipywidgets update messages in comms reducer

captainsafia:


packages/reducers/__tests__/comms.spec.ts
```diff
@@ -66,4 +67,30 @@ describe("comm reducers", () => {
       x: 5113
     });
   });
+  test("does a straight replacement for unknown comm messages", () => {
+    const state = stateModule.makeCommsRecord({
+      models: Immutable.Map({
+        101: {
+          state: {
+            x: 5112
```
miduncan: Let's add some property `y: "blahblah"` to this to make sure it gets preserved. Otherwise this test would still pass even if it is replacing instead of merging

miduncan (approved):

PR #4684, opened 2019-11-07
Clear cell prompts when clearing cell outputs and add support for multiple prompts per cell

captainsafia:
**Changes in this PR**
- Store multiple input prompts per cell in the state
- Send input replies to both WebSockets and 0mq-backed kernels
- Clear input prompts when cell outputs are cleared


miduncan (approved):
Code looks good. Are there any tests we could write that would catch this issue if it happens again, though?

PR #4686, opened 2019-11-07
Keep track of widgets in the creation process

### LINKED ISSUE #4681: Same widget model not synced across different views
In the `Widget Low Level.ipynb` example, the widget slider that is rendered in the notebook does not sync changes across the different views that are rendered. All widgets should update simultaneously.

widget-sync

miduncan:
Fixes #4681

### DIFF
--- a/packages/jupyter-widgets/src/manager/widget-manager.ts
+++ b/packages/jupyter-widgets/src/manager/widget-manager.ts
@@ -120,16 +120,23 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
    */
   new_widget(options: any, serialized_state: any = {}): Promise<WidgetModel> {
     const model_id = options.model_id;
-    if (super.get_model(model_id)) {
-      //if this widget is already created
-      return super.get_model(model_id)!; //we can assert that it's not undefined because we just checked
-    } else if (this.widgetsBeingCreated[model_id]) {
-      //if this widget is in the process of being created
+    //if this widget is already created
+    const existing_widget = super.get_model(model_id);
+    if (existing_widget) {
+      return existing_widget;
+    }
+    //if this widget is in the process of being created
+    else if (this.widgetsBeingCreated[model_id]) {
       return this.widgetsBeingCreated[model_id];
-    } else {
+    }
+    //otherwise create a new widget
+    else {
       let widget = super.new_widget(options, serialized_state);
       this.widgetsBeingCreated[model_id] = widget;
-      return widget;
+      return widget.then((new_widget: WidgetModel) => {
+        delete this.widgetsBeingCreated[new_widget.model_id];
+        return Promise.resolve(new_widget);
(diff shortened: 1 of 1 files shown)


packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -117,12 +119,18 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
    * @param  serialized_state - serialized model attributes.
    */
   new_widget(options: any, serialized_state: any = {}): Promise<WidgetModel> {
-    //first we check if the model was already created
-    let widget = super.get_model(options.model_id); //we need to use the super because we override get_model to create what it can't find
-    if (!widget) {
-      widget = super.new_widget(options, serialized_state);
+    const model_id = options.model_id;
+    if (super.get_model(model_id)) {
+      //if this widget is already created
```
captainsafia: Major nit: Can you move the documenting comments into a block comment above the code section instead of inline. Makes it easier to read.

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -117,12 +119,18 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
...
    */
   new_widget(options: any, serialized_state: any = {}): Promise<WidgetModel> {
-    //first we check if the model was already created
-    let widget = super.get_model(options.model_id); //we need to use the super because we override get_model to create what it can't find
-    if (!widget) {
-      widget = super.new_widget(options, serialized_state);
+    const model_id = options.model_id;
+    if (super.get_model(model_id)) {
+      //if this widget is already created
+      return super.get_model(model_id)!; //we can assert that it's not undefined because we just checked
```
captainsafia: Can we get around having to do this non-null coercion by storing the result in a variable. Something like this.

```
...
const existingModel = super.get_model(model_id);
if (existingModel) return existingModel;
...
```

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -117,12 +119,18 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
...
+    } else if (this.widgetsBeingCreated[model_id]) {
+      //if this widget is in the process of being created
+      return this.widgetsBeingCreated[model_id];
+    } else {
+      let widget = super.new_widget(options, serialized_state);
+      this.widgetsBeingCreated[model_id] = widget;
+      return widget;
```
captainsafia: Can we do something here to clear out the state of the `widgetsBeingCreated` dictionary once the matching widget has been made?

captainsafia (commented):
Posted some comments inline.

captainsafia (approved):
Looks good! Thanks for resolving the feedback.

Issue #4645, opened 2019-10-28
widget.observe handlers are not functional


captainsafia:
In the Widget Events example notebook from ipywidgets, the onClick event on the button is not functional.

miduncan:
Hey I was able to find the cause of this (although I don't think this solves all of our event issues). You just need to call `view1.delegateEvents()` after you create a WidgetView(in this example, named `view1`). I can put a PR for this once the layout/style stuff is merged into master, because I think it should go in the create_view function I made there.

captainsafia:
@miduncan Sure! I was able to get some events working by configuring the callbacks to use an Observable.

Shortly after getting the events to work, I ended up needing to solve the bug where stdout is not redirected to the correct output.

Feel free to submit a PR with your solution as an incremental change.

captainsafia:
I learned a little bit more about this and turns out that these issues have to do with the fact that we don't currently have an OutputArea implementation. It is related to #4650.

miduncan:
The `delegateEvents` change accidentally made it in #4655 since that the branch I was testing it on. Considering that the PR had a create_view rewrite, though, it wasn't that out of place. 

Let me know if there is anything else I can do to help here-- I think this is our biggest blocker for acceptance in example notebooks

captainsafia:
@miduncan I think we can go ahead and close this. The work in #4687 allows the callback in the `observe` to be actually triggered and I haven't been able to reproduce this issue since.

We can close this and if it comes up again we can re-open the issue.

PR #4687, opened 2019-11-07
Emit notebook actions in callbacks for ipywidget messages

captainsafia:
Fix for #4645 

Output-related events are now handled properly for widgets on observe. Note, the lag in the image below is due to my machine lagging, not an actual issue with the appending of outputs.

widget-output

To make this change, I had to pass in the cellId and contentRef to the rendered media component. This way, I could set up some dispatch actions that would invoke our appendOutput/clearOutputs functions on the appropriate notebook and cell.

### DIFF
--- a/packages/jupyter-widgets/src/manager/index.tsx
+++ b/packages/jupyter-widgets/src/manager/index.tsx
@@ -9,8 +9,10 @@ import {
   actions,
   KernelNotStartedProps,
   LocalKernelProps,
-  RemoteKernelProps
+  RemoteKernelProps,
+  ContentRef
 } from "@nteract/core";
+import { CellId } from "@nteract/commutable";
 import { WidgetModel } from "@jupyter-widgets/base";
 
 interface ConnectedProps {
@@ -21,12 +23,24 @@ interface ConnectedProps {
     | RecordOf<RemoteKernelProps>
     | null;
 }
+
+interface DispatchProps {
+  actions: {
+    appendOutput: (output: any) => void;
+    clearOutput: () => void;
+    updateCellStatus: (status: string) => void;
+    promptImputRequest: (prompt: string, password: boolean) => void;
+  };
+}
+
 interface OwnProps {
   model: WidgetModel;
   model_id: string;
+  id: CellId;
+  contentRef: ContentRef;
 }
 
-type Props = ConnectedProps & OwnProps;
+type Props = ConnectedProps & OwnProps & DispatchProps;
 
 /**
  * This component is is a wrapper component that initializes a
@@ -89,7 +103,7 @@ const mapStateToProps = (state: AppState, props: OwnProps): ConnectedProps => {
   };
 };
 
-const mapDispatchToProps = (dispatch: any, props: OwnProps): any => {
+const mapDispatchToProps = (dispatch: any, props: OwnProps): DispatchProps => {
   return {
     actions: {
       appendOutput: (output: any) =>
@@ -107,7 +121,7 @@ const mapDispatchToProps = (dispatch: any, props: OwnProps): any => {
             contentRef: props.contentRef
           })
         ),
-      updateCellStatus: status =>
+      updateCellStatus: (status: string) =>
         dispatch(
           actions.updateCellStatus({
             id: props.id,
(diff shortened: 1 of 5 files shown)


packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -82,4 +102,47 @@ const mapStateToProps = (state: AppState, props: OwnProps): ConnectedProps => {
...
+          })
+        ),
+      clearOutput: () =>
+        dispatch(
+          actions.clearOutputs({
+            id: props.id,
+            contentRef: props.contentRef
+          })
+        ),
+      updateCellStatus: (status: string) =>
```
miduncan, line 124: Double checking that there isn't a more specific type for this since status is usually an enum of some sort

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -82,4 +102,47 @@ const mapStateToProps = (state: AppState, props: OwnProps): ConnectedProps => {
...
+        dispatch(
+          actions.updateCellStatus({
+            id: props.id,
+            contentRef: props.contentRef,
+            status
+          })
+        ),
+      promptImputRequest: (prompt: string, password: boolean) =>
```
miduncan: Typo: I**m**put

miduncan: Applies to interface too

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -37,16 +38,19 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
     | RecordOf<LocalKernelProps>
     | RecordOf<RemoteKernelProps>
     | null;
+  actions: any;
 
-  constructor(kernel: any, stateModelById: (id: string) => any) {
+  constructor(kernel: any, stateModelById: (id: string) => any, actions: any) {
```
miduncan: I think we should move `stateModelById` into `actions`. Also I think we might as well make the type for `actions` now to help with Intellisense and catch some errors in the build

packages/jupyter-widgets/src/manager/widget-manager.ts
```diff
@@ -142,6 +146,33 @@ export class WidgetManager extends base.ManagerBase<DOMWidgetView> {
...
   }
 
+  /**
+   * The ManagerBase type definition for the callbacks method expects
+   * the message types to be as defined by the IMessage interface from
+   * @jupyterlab/services. It is typed as any here so that we can use
+   * our JupyterMessage types that are emitted from our kernel.channels
+   * pipeline.
+   */
+  callbacks(): any {
```
miduncan: It looks like this is called internally within ipywidgets, so I would be somewhat hesitant to change the contract by not adhering to their expected return type. Are we sure that this doesn't cause any problems?

miduncan (commented):
Left feedback inline

packages/jupyter-widgets/src/manager/index.tsx
captainsafia: We have a kernel status type but it encapsulates the types that are not the busy/idle that come from the kernel. The updateCellStatus action itself takes a string but we should probably create an enum for this and the other statuses that come back from the kernel.

Tracked this separately in https://github.com/nteract/nteract/issues/4688.

packages/jupyter-widgets/src/manager/index.tsx
captainsafia: Good catch! Thanks!

packages/jupyter-widgets/src/manager/widget-manager.ts
captainsafia: I was able to add a type for `actions`. Couldn't move the `stateModelById` method to the actions directory since it requires the state when it is initialized in mapStateToProps whereas the actions are created in `mapDispatchToProps`. We can probably rearrange this stuff later.

packages/jupyter-widgets/src/manager/widget-manager.ts
captainsafia: The typings shouldn't affect the functionality of the application. From examination though, both JupyterMessage and IMessage are the top-level message structs that are used so I think they will interop fine.

miduncan: I was thinking along the lines of `foo.bar is null or undefined` or `value is NaN` type errors. But if the interop fine together, I believe Typescript looks at the interface properties/methods and not just the names of the interfaces for determining a type error. If not, `any` works.

miduncan (approved):

PR #4702, opened 2019-11-13
Requests state from kernel if model is not found in redux store

### LINKED ISSUE #4675: Implement _get_comm_info function to rehydrate comms
We need to implement a `_get_comm_info` method that requests info about the active comms from the kernel so that we can rehydrate the state when we refresh the page without changing the kernel.

cc: @miduncan

miduncan:
Fixes #4675

### DIFF
--- a/packages/jupyter-widgets/src/index.tsx
+++ b/packages/jupyter-widgets/src/index.tsx
@@ -1,9 +1,18 @@
 import * as React from "react";
-import { selectors, AppState, ContentRef } from "@nteract/core";
-
+import { connect } from "react-redux";
+import {
+  selectors,
+  AppState,
+  ContentRef,
+  KernelNotStartedProps,
+  LocalKernelProps,
+  RemoteKernelProps
+} from "@nteract/core";
+import { fromJS, RecordOf } from "immutable";
 import Manager from "./manager";
 import { WidgetModel } from "@jupyter-widgets/base";
 import { CellId } from "@nteract/commutable";
+import { request_state } from "./manager/widget-comms";
 
 interface JupyterWidgetData {
   model_id: string;
@@ -13,8 +22,18 @@ interface JupyterWidgetData {
 
 interface Props {
   data: JupyterWidgetData;
+  modelById: (model_id: string) => Promise<WidgetModel>;
+  kernel?:
+    | RecordOf<KernelNotStartedProps>
+    | RecordOf<LocalKernelProps>
+    | RecordOf<RemoteKernelProps>
+    | null;
   id: CellId;
   contentRef: ContentRef;
+}
+
+interface State {
+  model?: WidgetModel;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -96,10 +98,21 @@ class Manager extends React.Component<Props> {
 }
 
 const mapStateToProps = (state: AppState, props: OwnProps): ConnectedProps => {
+  let currentKernel = selectors.currentKernel(state);
```
captainsafia: Nit: Use `selectors.kernelByContentRef` and pass down the ContentRef from the parent component so that this maps to the correct kernel.

packages/jupyter-widgets/src/manager/widget-comms.ts
```diff
@@ -84,6 +85,36 @@ export class WidgetComm implements IClassicComm {
...
+   * needs to be called before a Comm object is made
+   * @param kernel
+   * @param comm_id
+   */
+  static request_state(kernel: any, comm_id: string): Promise<any> {
+    return new Promise((resolve, reject) => {
+      const message = createCommMessage(comm_id, { method: "request_state" });
+      let replySubscription = kernel.channels
+        .pipe(childOf(message))
+        .subscribe((reply: any) => {
```
captainsafia: Nit: Can you type this as `reply: JupyterMessage`?

packages/jupyter-widgets/src/manager/widget-comms.ts
```diff
@@ -84,6 +85,36 @@ export class WidgetComm implements IClassicComm {
...
+          //if we get a comm message back, it is the state we requested
+          if (reply.msg_type === "comm_msg") {
+            replySubscription.unsubscribe();
+            return resolve(reply);
+          }
+          // otherwise, if we havent gotten a comm message and it goes idle, it wasn't found
+          else if (
```
captainsafia: Should we make this into an else and not an else-if or is this flow guranteed to always return either a `comm_msg` or return the kernel to idle?

packages/jupyter-widgets/src/index.tsx
```diff
@@ -41,16 +38,4 @@ export class WidgetDisplay extends React.Component<Props> {
...
-  return {
-    model: selectors.modelById(state, { commId: model_id })
-  };
-};
-
-export default connect(
-  mapStateToProps,
-  null
-)(WidgetDisplay);
+export default WidgetDisplay;
```
captainsafia: I'm a fan that this top-level component is no longer connected.

packages/jupyter-widgets/src/manager/index.tsx
```diff
@@ -82,12 +81,15 @@ class Manager extends React.Component<Props> {
   }
 
   render() {
+    const { modelById, model_id } = this.props;
     return (
       <React.Fragment>
         <BackboneWrapper
-          model={this.props.model.get("state")}
+          getModelAsync={() => {
```
captainsafia: Can we possibly call this in `componentDidMount` instead of inside the `render` method?

captainsafia (commented):
Passing a selector instead of the model to the manager seems fine.

Left some comments inline about how we are structuring this.

packages/jupyter-widgets/src/manager/widget-comms.ts
miduncan: So the flow will always end with a kernel to idle, so we want it to be an else-if.

packages/jupyter-widgets/src/manager/index.tsx
miduncan: So this is how we were selecting the kernel before to hand down to the manager. I'll create an issue and we can solve this in another PR, because if we are going to use this we need to also change the Manager to work with different widgets having different kernels

packages/jupyter-widgets/src/index.tsx
miduncan: Unfortunately this is no longer the case. I think it is better for testing to keep the Manager expecting a model rather than to use the modelById function

packages/jupyter-widgets/src/manager/widget-comms.ts
miduncan: It looks like this doesn't let have the msg_type field, which I need here.

packages/jupyter-widgets/src/index.tsx
```diff
@@ -14,39 +22,80 @@ interface JupyterWidgetData {
...
+    this.state = { model: undefined };
+  }
+
+  async componentDidMount() {
+    const {
+      modelById,
+      data: { model_id }
+    } = this.props;
+    const model = await modelById(model_id);
+    console.log(model);
```
captainsafia: Can you remove this?

captainsafia (approved):
Looks good!

Left one clean-up comment to address and looks like the LGTM bot has some feedback too.

packages/jupyter-widgets/src/index.tsx
miduncan: Oof! Good catch

Issue #4696, opened 2019-11-12
RFC: Stateful Components and Theming


captainsafia:
### Problem

Currently, nteract's component and theming story is only partial implement. We started development a component ecosystem in the [@nteract/ion project](https://github.com/nteract/ion) but parked that work while we iterated on presentational components in the nteract notebook app. Our component system has evolved and we've got quite a few pieces of feedback in issues and direct reports about ways to improve this component API. This RFC is an effort to push this through.

What exactly is the problem?

Currently, it's rather difficult for user's to compose their own nteract-based applications. From my observation, this is the result of a few things.

- Our @nteract/presentational-components are opinionated about what styles are applied to them. Not all users building on top of nteract might want to adopt these tiles.
- Our @nteract/notebook-app-component is very monolithic and contains quite a few pieces that would make more sense factored out into their own units such as the `TransformMedia`, `Cell`, `CellToolbar`, and `CellCreator` components.
- Our theming story is limited and it is difficult for community members and developers building on top of our API to theme nteract-based applications.

### Goals

We need to solve this problem! I believe our goals for this solution should as follows:

- Improve the maintainability of the nteract component ecosystem between the components that are under @nteract/presentational-component, @nteract/connected-components and @nteract/notebook-app-component, our component story is quite difficult to reason about for contributors and maintainers.
- Create more atomic and composable units for developers building nteract-based components
- Make it easy for developers and community members to theme nteract components via styled-components or stylesheets
- Create documentation for developer building nteract-based applications to make it easier for developers to build maintainable UIs based on the nteract component ecosystem.

### Proposed Solution

The proposed solution largely rests in cleaning up how we export and document the presentational components, how we manage and load our stylesheets, how we connect Redux to presentation components, and how we expose class names as interfaces that users can extend with their themes.

Currently, most of the presentational components that we export are styled dependent on our styled sheets. This is fine for our scenario but we should also export unstyled components that set the appropriate CSS classes so that component consumers can style them independently. The components should still maintain their semantic structure but should be more meaningful to decouple from state.

Also, there is a monolithic `AnyCell` component that acts as the connected component that wraps all individual presentation components. This pattern should be decomposed so that the state and actions for inputs are mapped separately from the state and actions for outputs and so on. Instead of a singular `mapStateToProps` and `mapDispatchToProps` function for the entire cell, we should decompose it into smaller pieces.

With all of this, I anticipate that we'll be able to use the components a little something like this. Within nteract we export the following.

```
[...]

PR #4703, opened 2019-11-13
Move TransformMedia component to @nteract/outputs

captainsafia:
Ref: #4696

### DIFF
--- a/packages/outputs/src/components/transform-media.tsx
+++ b/packages/outputs/src/components/transform-media.tsx
@@ -169,9 +169,7 @@ const makeMapDispatchToProps = (
   return mapDispatchToProps;
 };
 
-const TransformMedia = connect(
+export const TransformMedia = connect(
   makeMapStateToProps,
   makeMapDispatchToProps
 )(PureTransformMedia);
-
-export default TransformMedia;
--- a/packages/notebook-app-component/src/notebook-app.tsx
+++ b/packages/notebook-app-component/src/notebook-app.tsx
@@ -13,7 +13,8 @@ import {
   Output,
   PromptRequest,
   RichMedia,
-  StreamText
+  StreamText,
+  TransformMedia
 } from "@nteract/outputs";
 import {
   Cell as PlainCell,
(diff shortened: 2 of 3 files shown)


packages/outputs/src/components/transform-media.tsx
```diff
@@ -169,9 +169,7 @@ const makeMapDispatchToProps = (
   return mapDispatchToProps;
 };
 
-const TransformMedia = connect(
+export const TransformMedia = connect(
   makeMapStateToProps,
   makeMapDispatchToProps
 )(PureTransformMedia);
```
rgbkrk, line 175: The only strange thing I'm seeing here is that this isn't relying on `@nteract/outputs`' `TransformMedia` as `PureTransformMedia`, so this may not have done exactly what you needed?

houshuang:
@captainsafia @rgbkrk As someone who is relying on @nteract/outputs in a different project, using these components as pure components connected to a different state system than your Redux, I would rather not have seen this file in @nteract/outputs. It's pulling in @nteract/core (which is not declared in the package.json), which of course is in its turn pulling in a lot of stuff that I don't need. I know my use case is perhaps not so important for you as someone outside of @nteract, but given the loving care you people have put into making these packages very cleanly separated and documented (I love it!) I assume you have also thought about the needs of external users? 

I made an example repo here which show cases the outputs and representational components using Storybook and a bunch of example data snippets captured from actual nteract sessions: https://github.com/houshuang/nteract-showcase.

You can see that currently it works when I pinned @nteract/outputs to 2.4, but if I set it to 2.5, it refuses to start.

(PS If you want to use any of this code to showcase components, or give people examples about how to combine your components without using Redux etc, feel free to use it in any way you see fit).

wadethestealth:
@houshuang 2.5 is broken since the dependencies required aren't in the package json anyways

captainsafia:
This change was made as part of the work around #4696.

It probably makes sense to move this to @nteract/connected-components. Although, that package already has its own thing going on.

rgbkrk:
Thank you for speaking up @houshuang so it can be addressed. Thank you for using the package.

PR #4704, opened 2019-11-13
Move MarkdownPreviewer component to @nteract/markdown

### LINKED ISSUE #3638: New Markdown Preview Cell Package
**Motivating Problem**
Someone familiar with the classic notebook's dichotomy of markdown and input cells might expect a symmetric API with similarly prominent input and markdown cells, but the current one offers prominent input cells and relatively hidden markdown cell functionality. It would be nice to have a MarkdownPreviewCell component that lives with the other cell components.

**A Solution I'd Like**
*Interface*
The docs invite users to render a cell using the following pattern,
```
const { Prompt, Source } = require("@nteract/presentational-components");
<Input>
<Prompt/>
<Source/>
</Input>
```
which suggests that the API needs a sister component to Source named Markdown. Just as Source is a presentational component that renders syntax-highlighted source code, Markdown would be a presentational component that renders Markdown.
__
*Implementation*
@rgbkrk recommended in Slack chat to add something like a `packages/markdown-preview-cell` directory that includes a new MarkdownPreviewCell component, based on the 'markdown' case from the `AnyCell` component in `packages/notebook-app-component/src/markdown-preview.js`). It looks like the `MarkdownCell` component in `notebook-app-component/src/markdown-preview.js`)

**Considered Alternatives**
At first I was thinking of just adding a data prop to the Cell component, but I see from the docs that it's not supposed to do anything by itself, only in tandem with other components like Input and Source, so that's right out! 

On second thought, it might be easier to just check if the `<Source />` component's `language` prop is set to "markdown." If it is, it should be treated like a markdown cell. This makes more sense to me, as markdown is also pretty much a language, no? I took [a crack at drafting this solution](https://github.com/jefftrevino/nteract/blob/dev_markdown_preview_cell/packages/presentational-components/src/components/source.js), predicting that you all will soon explain to me why this is too simple a solution....

**Additional Context**
I started thinking about this when I went to add a markdown cell to the jupyter-demo-app project example and found it much more difficult and less documented than the task of adding an input cell.

captainsafia:
Closes #3638

Ref: #4696

### DIFF
--- a/packages/markdown/__tests__/markdown-preview.spec.tsx
+++ b/packages/markdown/__tests__/markdown-preview.spec.tsx
@@ -2,11 +2,11 @@ import { emptyMarkdownCell } from "@nteract/commutable";
 import { mount, shallow } from "enzyme";
 import React from "react";
 
-import MarkdownPreview from "../src/markdown-preview";
+import { MarkdownPreviewer } from "../src/markdown-previewer";
 
-describe("MarkdownPreview ", () => {
+describe("MarkdownPreviewer ", () => {
   test("can be rendered", () => {
-    const cell = shallow(<MarkdownPreview cell={emptyMarkdownCell} />);
+    const cell = shallow(<MarkdownPreviewer cell={emptyMarkdownCell} />);
     expect(cell).not.toBeNull();
   });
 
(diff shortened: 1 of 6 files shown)


rgbkrk (approved):
Seems alright to shuffle this over.

PR #4710, opened 2019-11-15
Upgrade from url-search-params to @ungap/url-search-params

### LINKED ISSUE #3813: Dependency deprecation warning: url-search-params (npm)
On registry `https://registry.npmjs.org`, the "latest" version (v1.1.0) of dependency `url-search-params` has the following deprecation notice:

`now available as @ungap/url-search-params`

Marking the latest version of an npm package as deprecated results in the entire package being considered deprecated, so contact the package author you think this is a mistake.

Affected package file(s): `packages/rx-jupyter/package.json`

If you don't care about this, you can close this issue and not be warned about `url-search-params`'s deprecation again. If you would like to completely disable all future deprecation warnings then add the following to your config:

```
"suppressNotifications": ["deprecationWarningIssues"]
```

captainsafia:
Closes #3813.

### DIFF
--- a/packages/rx-jupyter/package.json
+++ b/packages/rx-jupyter/package.json
@@ -20,6 +20,7 @@
   "dependencies": {
     "@nteract/commutable": "^7.1.4",
     "@nteract/messaging": "^6.1.7",
+    "@types/ungap__url-search-params": "^0.1.0",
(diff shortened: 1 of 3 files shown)


rgbkrk (approved):
Cool, I checked out the package while I was at it. Looks good.

PR #4712, opened 2019-11-17
remove typeahead from nteract

### LINKED ISSUE #3904: Add configuration options for auto-complete
When typing code, I don't always like the auto-complete. In many editors and languages it sometimes pops up when unneeded and gets in the way. Trying to make everyone happy, I would suggest the following options for people to chose from:

1. Auto-complete always on, or only after pressing a key (e.g. TAB)

2. Auto-complete only after N=? characters. Let the user choose.

rgbkrk:
This takes out typeahead-backed completions for nteract's codemirror editor.

Closes #3904.


rgbkrk:
Seems to work well on Desktop! `ctrl-space` is the only way to do completion now.

rgbkrk:
Woo! Ready for review after the last few fixups. 😅

captainsafia (approved):
Nifty.

PR #4716, opened 2019-11-20
Improve host types interface

captainsafia:
**Changes in this PR**
- Add `ofHostType` RxJS operator to make it easier to write epics that are only applicable to applications backed by certain hosts
- Cleaned up hosts types folder to make it easier to add new host types


rgbkrk (approved):

packages/types/src/entities/hosts/base.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * An RxJS operator for determining if the host that is currently
+ * set as the current host in the state is of a particular type.
+ *
+ * @param hostType  The host type to filter by
+ */
+export const ofHostType = (...hostTypes: Array<string | [string]>) => (
+  source: Observable<any>
```
rgbkrk, line 22: This `source` type can be tightened up here, but it can always be updated.

captainsafia: Yeah, I think I'll probably end up cleaning this up once I complete the refactor in #4717

PR #4717, opened 2019-11-21
[WIP] Clean up epics folder and epics TODOs

captainsafia:
**Changes in this PR**
- Moves Jupyter-related epics to their own folder
- Moves remote and local epics for Jupyter to their own folders
- Moves generic Jupyter epics to their own folder


packages/epics/__tests__/execute.spec.ts
```diff
@@ -128,11 +127,13 @@ describe("createExecuteCellStream", () => {
...
     );
     const actionBuffer = [];
-    observable.subscribe(x => actionBuffer.push(x), err => done.fail(err));
+    observable.subscribe(
+      x => actionBuffer.push(x),
+      err => done.fail(err)
+    );
     expect(actionBuffer).toEqual([
-      actions.sendExecuteRequest({
+      actions.executeCell({
```
rgbkrk, line 135: What led to this naming change?

packages/epics/src/jupyter/generic/execute.ts
```diff
@@ -200,9 +198,14 @@ export function createExecuteCellStream(
   );
 
   return merge(
-    // We make sure to propagate back to "ourselves" the actual message
-    // that we sent to the kernel with the sendExecuteRequest action
-    of(actions.sendExecuteRequest({ id, message, contentRef })),
```
rgbkrk, line 205: What takes the place of this former code?

rgbkrk: Ah, it seems like now we just issue the clear outputs as a result of this instead of handling the `execute` within the reducer.

captainsafia:
Closing this in favor of a more piece meal approach.

PR #4726, opened 2019-11-30
Call finalize on vega embed result or view upon unmounting

### LINKED ISSUE #4724: Fix potential memory leak in Vega visualizations
Vega-Embed and Vega may be holding on to memory after a visualization is no longer needed. To fix this, you can call `finalize` on a Vega-Embed result (since 6.2) or Vega view. 

When you know that a vis is no longer needed (i.e. the react component is unmounted), nteract should call `finalize`.

CrystallineCat:
Fixes #4724.


captainsafia (approved):
Thanks!

captainsafia:
I will go ahead and merge this since the Azure Pipelines failure is unrelated.

captainsafia:
Welp! Totally had a brain fart and accidently closed instead of merging. My bad!

domoritz:
Just a small change is needed here. Instead of only calling finalize on the view, you should call finalize on the embed result in embed 6.2. It removes some additional handlers that are otherwise not freed.

captainsafia:
Ah -- thanks for adding that.

Interested in submitting a PR for this @domoritz?

domoritz:
Sure! I'll take a stab at it.

CrystallineCat:
Thanks for reviewing and merging!

I tried upgrading to `vega-embed@6.2` (and had the call to `finalize()` of the embed object; see commit history), however the type definitions of that version seem to be incompatible with our version of typescript. `tsc` gave me errors about the `readonly` keyword being used on arrays, which needs `typescript@3.4` (according to the release notes), we are at `typescript@3.2.2`.

domoritz:
I see. Can you update TypeScript?

I could make another Vega-Embed release without the readonly in https://github.com/vega/vega-embed/blob/master/src/util.ts#L17. I'm not sure that is enough, though since we also use readonly in dependent projects. Can you check whether removing the readonly in https://github.com/vega/vega-embed/blob/master/src/util.ts#L17 is enough? I would prefer not to remove the `readonly` since it makes the types stronger.

PR #4728, opened 2019-11-30
[RFC] [WIP] Add a preferences window

CrystallineCat:
Collects all preferences in a window:
Screenshot from 2019-11-30 18-10-03


captainsafia:
@CrystallineCat Looks good! One thought is to make the configuration component it's own package so that it can be used in both the desktop and web apps.

CrystallineCat:
I need to make this into an issue and multiple smaller PRs, I think.

PR #4731, opened 2019-12-03
Add @nteract/stateful-components package

captainsafia:
This PR lays the groundwork for work around @nteract/stateful-componetns as outlined in https://github.com/nteract/nteract/issues/4696.

Namely, it does the following:
- Decomposes the elements that were previously in @nteract/notebook-app-component into their own individual bits
- Provides an interface for accessing React components that aren't connect to Redux
- Provides a set of exported connected components that developers can use out of the box
- Removes the assumptions about styles to pave the way for a more sensible theming story

This is a big PR, so I'm gonna go ahead and post comments inline as notation.

Some work still remains here, in particular:
- Adding the CellToolbar component to the CodeCell and MarkdownCell components
- Adding an InputPrompt component
- Adding more docs about the different ways this can be leveraged

I figure this is good to go in with the code, tests, and docs that exist in the moment and we can continue building.

### DIFF
--- a/packages/editor/src/index.tsx
+++ b/packages/editor/src/index.tsx
@@ -63,6 +63,7 @@ export type CodeMirrorEditorProps = {
   onChange?: (value: string, change: EditorChangeLinkedList) => void;
   onFocusChange?: (focused: boolean) => void;
   value: string;
+  editorType: "codemirror";
 } & Partial<FullEditorConfiguration>;
 
 interface CodeMirrorEditorState {
@@ -94,7 +95,8 @@ export default class CodeMirrorEditor extends React.PureComponent<
     matchBrackets: true,
     indentUnit: 4,
     lineNumbers: false,
-    cursorBlinkRate: 530
+    cursorBlinkRate: 530,
+    editorType: "codemirror"
   };
 
   textarea?: HTMLTextAreaElement | null;
--- a/packages/fixtures/src/index.ts
+++ b/packages/fixtures/src/index.ts
@@ -117,7 +117,8 @@ export const mockAppState = (config: JSONObject): AppState => {
                 cellFocused:
                   config && config.codeCellCount && config.codeCellCount > 1
                     ? dummyNotebook.get("cellOrder", Immutable.List()).get(1)
-                    : null
+                    : null,
+                kernelRef
               }),
               filepath:
                 config && config.noFilename ? "" : "dummy-store-nb.ipynb"
--- a/packages/stateful-components/README.md
+++ b/packages/stateful-components/README.md
@@ -1,11 +1,67 @@
-# `stateful-components`
+# @nteract/stateful-components
 
-> TODO: description
+This package exports a set of React components that are pre-connected to the Redux state as mapped by @nteract/core. This package is designed to reduced the hassle for building React-based applications for developers who are already building on top of the core SDK. It also exposes several interfaces for extending the `mapStateToProps`, `mapDispatchToProps`, and styles of the individual components.
+
+## Installation
+
+```
+$ yarn add @nteract/stateful-components
+```
+
+```
+$ npm install --save @nteract/stateful-components
+```
 
 ## Usage
 
-```
-const statefulComponents = require('stateful-components');
+You can import the top-leval Notebook component which contains the canonical setup for a Jupyter-backed notebook application.
 
-// TODO: DEMONSTRATE API
 ```
+import { Notebook }  from "@nteract/stateful-components"
+
+class MyNotebook extends React.Component {
+    render() {
+        return <Notebook contentRef={someContentRefHere}>
+    }
+}
+```
+
+Alternatively, you can assemble your own notebook-UI from the subcomponents. For example, to swap out the canonical `MarkdownCell` implementation.
+
+```
+import { Cells, Cell, CodeCell, RawCell } from "@nteract/stateful-components";
+
+class MyMarkdownCell extends React.Component {
+    static defaultProps = {
+        cell_type: "markdown"
+    }
+
+    render() {
+        return <p>My custom logic here</p>;
+    }
+}
+
+class MyNotebook extends React.Component {
+    render() {
+        return <Cells>
+            <Cell>
+                <MyMarkdownCell/>
+                <CodeCell/>
+                <RawCell>
+            </Cell>
+        </Cells>;
+    }
+}
+```
+
+## Documentation
+
+We're working on adding more documentation for this component. Stay tuned by watching this repository!
+
+## Support
+
+If you experience an issue while using this package or have a feature request, please file an issue on the [issue board](https://github.com/nteract/nteract/issues/new/choose) and add the `pkg:stateful-components` label.
+
+## License
+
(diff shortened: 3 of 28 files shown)


packages/editor/src/index.tsx
```diff
@@ -63,6 +63,7 @@ export type CodeMirrorEditorProps = {
   onChange?: (value: string, change: EditorChangeLinkedList) => void;
   onFocusChange?: (focused: boolean) => void;
   value: string;
+  editorType: "codemirror";
```
captainsafia, line 66: This is added to enable the selection of different editor types from the `Editor` stateful component. This paves the way for integrating Monaco into the nteract notebook app and making the chosen editor configurable.

packages/fixtures/src/index.ts
```diff
@@ -117,7 +117,8 @@ export const mockAppState = (config: JSONObject): AppState => {
                 cellFocused:
                   config && config.codeCellCount && config.codeCellCount > 1
                     ? dummyNotebook.get("cellOrder", Immutable.List()).get(1)
-                    : null
+                    : null,
+                kernelRef
```
captainsafia, line 121: This is needed to make the tests that use the new `kernelByContentRef` work nicely. That selector is used throughout several epics and expects a kernelRef to be set in the notebook model.

packages/stateful-components/README.md
```diff
@@ -0,0 +1,67 @@
...
+                <RawCell>
+            </Cell>
+        </Cells>;
+    }
+}
+```
+
+## Documentation
+
+We're working on adding more documentation for this component. Stay tuned by watching this repository!
```
captainsafia, line 59: I want to add a `docs` directory with more tutorial style docs. I'm debating trying to see if I can get a monorepo docs setup for more prose-like docs with Gitbook or something.

Regardless, I'm hoping the content will go there.

packages/stateful-components/src/cells/cell.tsx
```diff
@@ -0,0 +1,72 @@
...
+  id: string;
+  contentRef: string;
+}
+
+interface StateProps {
+  cell: Immutable.Map<string, any>;
+}
+
+export class Cell extends React.Component<ComponentProps & StateProps> {
+  render() {
```
captainsafia, line 17: I brought the same child-picking logic we currently use in the `Output` component to the `Cell` component. I like that it encourages (**cough** forces **cough**) developers to build declarative and semantic components hierarchies in their notebook based applications.

packages/stateful-components/src/cells/cells.tsx
```diff
@@ -0,0 +1,65 @@
...
+  cellOrder: Immutable.List<string>;
+}
+
+export class Cells extends React.Component<StateProps & ComponentProps> {
+  render() {
+    const { cellOrder, contentRef, children } = this.props;
+    return (
+      <div className="nteract-cells">
+        {cellOrder.map((id: string) =>
+          children ? (
```
captainsafia, line 27: This logic allows developers to override the child cell component with their own implementation instead of relying on the canonical implementation. Namely, it allows the consumer to avoid have to rewrite the logic (for iterating through a list of cells) in this container component if they just want to override the children.

packages/stateful-components/src/cells/code-cell.tsx
```diff
@@ -0,0 +1,84 @@
...
+          />
+          <TransformMedia
+            output_type={"execute_result"}
+            id={id}
+            contentRef={contentRef}
+          />
+          <KernelOutputError />
+          <StreamText />
+        </Outputs>
+        {/* <InputPrompts id={id} contentRef={contentRef} /> */}
```
captainsafia: This will need to be implemented but I don't want it to block this for now.

packages/stateful-components/src/inputs/editor.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+interface StateProps {
+  editorType: string;
+  editorFocused: boolean;
+  value: string;
+  channels: any;
+  kernelStatus: string;
+}
+
+export class Editor extends React.Component<ComponentProps & StateProps> {
```
captainsafia, line 21: This component implements the same child-picker logic and connects the state properties that are needed by both CodeMirror and Monaco (and generic to other editors) -- paving the way for multi-editor support in nteract.

packages/editor/src/index.tsx
rgbkrk: Would we not want to do something where we drop in the editor as a child component rather than a prop to change?

```
<EditorShell>
  <ConnectedCodeMirrorEditor />
<EditorShell />
```

packages/stateful-components/README.md
rgbkrk: Sounds good. Just happy to start with the content first, so this is always super welcome.

packages/editor/src/index.tsx
captainsafia: The `Editor` still allows passing editor components as children. But the `editorType` allows selecting between different editors. So the structure would look like this.

```
<EditorShell>
    <ConnectedCodemirrorEditor/>
    <ConnectedMonacoEditor/>
</EditorShell>
```

Then the editor to render is configurable by setting a particular `editorType` so that users can swap between CodeMirror, Monaco, and others as desired.

packages/stateful-components/src/cells/cell.tsx
rgbkrk: Hehe.

packages/stateful-components/src/cells/markdown-cell.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { MarkdownPreviewer } from "@nteract/markdown";
+import CodeMirrorEditor from "@nteract/editor";
+
+import Editor from "../inputs/editor";
+
+interface ComponentProps {
+  id: string;
+  contentRef: ContentRef;
+  cell: Immutable.Map<string, any>;
+  cell_type: "markdown";
```
rgbkrk, line 17: I didn't realize we used `snake_case` here now. I'm guessing that was a commutable change? I'm so behind.

rgbkrk (approved):

packages/stateful-components/src/cells/cells.tsx
rgbkrk: This is cool! The nice thing here is that they don't have to deal with the `id` and `contentRef` themselves while they can encode the structure of cells themselves. 😎

packages/stateful-components/src/cells/markdown-cell.tsx
captainsafia: Yes, the cell_type that is stored in the cellMap uses snake_case. A while back I submitted a PR to change to pascalCase but it ended up being more trouble than it was worth so I abandoned it.

captainsafia: Yes, it it still snake case in accordance with what is in nbformat.

A while back, I submitted a PR changing the snake case to pascal case but it turned out to be more trouble than it was worth so I have abandoned it.

rgbkrk: Oh yeah!

PR #4733, opened 2019-12-04
Only unsubscribe for kernel-listeners on if kernel action is triggered on same kernel

### LINKED ISSUE #4732: LAUNCH_KERNEL_BY_NAME breaks cell output display
We're using the jupyter_extension code base, we're opening multiple notebook at the same time (in different tabs in our app).

**Describe the bug**
The APPEND_OUTPUT action (which causes the cell output to update with the latest results) isn’t emitted anymore. APPEND_OUTPUT stops being emitted as soon as LAUNCH_KERNEL_BY_NAME is emitted (by the launchKernelWhenNotebookSetEpic, or by dispatching manually in the redux dev tool).
The [`executeCellStream()`](https://github.com/nteract/nteract/blob/0691e6d3dddb9e91489a52d6e85b1e268e8d4bba/packages/epics/src/execute.ts#L56) creates `cellStream` with `takeUntil()` with `actions.LAUNCH_KERNEL_BY_NAME` or `actions.KILL_KERNEL`. I believe that means we stop listening to the stream if these actions are received. This also explains why it stops updating when we close a notebook in our app (closing a notebook emits a KILL_KERNEL action).

**To Reproduce**
1.	Open notebook A, execute a long-running cell, for example this:
import time
for x in range(60):
    print(x)
    time.sleep(1)
2.	Open another notebook B.
3.	Go back to the previous notebook A and notice that now the long running cell doesn’t update anymore. It seems stop on running.

**Expected behavior**
Notebook A should still keep updating the cell output.

**Screenshots**

- OS: [e.g. macOS]

**Additional context**

captainsafia:
Closes #4732 

Following the changes made to introduce better handling for maintaining multiple notebooks in the nteract state each with their own kernel, this bug surfaced.

The `takeUntil` on the `executeCellStream` was pausing the execution follow-up Observable whenever a new kernel was launched or killed. This makes sense if there is only ever one notebook in the store, but when there are many, it results in weird behavior as each notebooks launch and kill events for the kernel affect the execution flows of the other notebook.

This PR fixes this by adding an additional filter on the `takeUntil` so that it only triggers if the kernel-related events are triggered by the same notebook that the execution handle is currently happening on.

**Update:** I went ahead and did a pass-through and updated all the places where we use `takeUntil` in a kernel-related context to filter by the correct content ref or kernel ref.

cc: @languy @virangai @rgbkrk

### DIFF
--- a/packages/epics/src/comm.ts
+++ b/packages/epics/src/comm.ts
@@ -2,14 +2,15 @@ import { ofMessageType } from "@nteract/messaging";
 import { ofType, StateObservable } from "redux-observable";
 import { ActionsObservable } from "redux-observable";
 import { merge } from "rxjs";
-import { map, switchMap, takeUntil } from "rxjs/operators";
+import { map, switchMap, takeUntil, filter } from "rxjs/operators";
 
 import {
   commMessageAction,
   commOpenAction,
   KILL_KERNEL_SUCCESSFUL,
   LAUNCH_KERNEL_SUCCESSFUL,
-  NewKernelAction
+  NewKernelAction,
+  KillKernelSuccessful
 } from "@nteract/actions";
 import * as selectors from "@nteract/selectors";
 import { AppState } from "@nteract/types";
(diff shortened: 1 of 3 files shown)


languy (approved):
Thanks for fixing the other `takeUntil()` calls.

rgbkrk (approved):

rgbkrk:
Since the azure pipelines setup is now disabled and old PRs have it still, I'm assuming this can just be merged.

PR #4735, opened 2019-12-04
Add missing message types to MessageType

### LINKED ISSUE #4694: Audit MessageType for types that are not represented
The `MessageType` does not currently contain a type for `complete_reply`. We should add it and audit the `MessageType` for any other missing types.

captainsafia:
Closes https://github.com/nteract/nteract/issues/4694.


rgbkrk (approved):
Mega thank you! I was just using these and I'm so happy to have the additional messages. ❤️

PR #4737, opened 2019-12-04
Update kernel info in notebook metadata on kernel change

### LINKED ISSUE #4576: Change kernels doesn't get saved
Using nteract on jupyter

**Describe the bug**
Changing kernel in notebook doesn't get saved.

**To Reproduce**

1. Open notebook
2. Select from the dropdown menu Runtime -> Change Kernel -> new kernel
3. Save notebook
4. Close notebook
5. Open notebook: the current kernel displayed at the bottom is not the newly selected kernel in step 2.

**Expected behavior**
I expect the opened notebook to show the newly selected kernel.
Changing on jupyterlab will display the new kernel.

**Screenshots**

- OS: [e.g. macOS]

**Additional context**
I debugged this and here's what I found out:
* In this [reducer](https://github.com/nteract/nteract/blob/0da75faefc6cef1bcf3ad86847a454b8f88d388d/packages/reducers/src/core/entities/kernels.ts#L92), CORE/SET_KERNEL_INFO sets `core/entities/kernels/byref/id/info`, but not `core/entities/contents/byRef/id/model/notebook/metadata/kernelspecs`
* The [SaveContentEpic](https://github.com/nteract/nteract/blob/0da75faefc6cef1bcf3ad86847a454b8f88d388d/packages/epics/src/contents.ts#L341) uses the `content`'s `metadata` field to persist to disk, so it will persist the wrong (old) `metatdata/kernelspecs`.
* When opening the notebook, the old kernel gets used.

captainsafia:
Closes #4576 

It turns out the `SetKernelspecInfo` action was deprecated. The `acquireKernelInfo` epic now epics a `SetKernelInfo` action. I've updated the reducers to modify the metadata of the notebook when this action is emitted as well.

### DIFF
--- a/packages/epics/src/kernel-lifecycle.ts
+++ b/packages/epics/src/kernel-lifecycle.ts
@@ -6,7 +6,7 @@ import {
   JupyterMessage,
   ofMessageType
 } from "@nteract/messaging";
-import { ActionsObservable, ofType } from "redux-observable";
+import { ActionsObservable, ofType, StateObservable } from "redux-observable";
 import { empty, merge, Observable, Observer, of } from "rxjs";
 import {
   catchError,
@@ -62,7 +62,8 @@ export const watchExecutionStateEpic = (
 export function acquireKernelInfo(
   channels: Channels,
   kernelRef: KernelRef,
-  contentRef: ContentRef
+  contentRef: ContentRef,
+  state: Appstate
 ) {
   const message = createMessage("kernel_info_request");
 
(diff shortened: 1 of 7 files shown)


captainsafia:
There's actually a bit more going on here. Marking as WIP while I figure out a robust way to do this set with the way we currently have our actions structured.

packages/epics/src/kernel-lifecycle.ts
```diff
@@ -101,6 +101,8 @@ export function acquireKernelInfo(
           })
         ];
       } else {
+        const kernelspec = selectors.kernelspecByName(state, { name: l.name });
+        const kernelInfo = { name: l.name, spec: kernelspec };
```
rgbkrk, line 105: 👀

rgbkrk (approved):
Quality propagation of the kernel metadata here.

PR #4739, opened 2019-12-05
Add keyboard-based shifting support to dropdown component

miduncan:
This adds the following keyboard shortcuts to dropdown menus for the sake of accessibility:
* "Arrow up" to move focus an item up
* "Arrow down" to move focus an item down
* "Enter" to click on an item
* "Esc" to hide the dropdown
temp


captainsafia (commented):
Are there any Jest tests that we can write for this? Perhaps by using enzymes event triggers on particular elements?

captainsafia (approved):
Thanks for adding the tests!

PR #4740, opened 2019-12-05
Fix pasting cells also pasting text

### LINKED ISSUE #3407: Fix: pasting cell also pastes text from clipboard
When I copy a cell (`Ctrl Shift C`) and paste it (`Ctrl Shift V`) while the cursor is still inside the cell (i.e. edit mode of Jupyter notebooks), then the text in my clipboard is also getting inserted along with the cell being pasted below

CrystallineCat:
Fixes #3407.

After several tries to intercept the event in the editor, all of which didn't work out, I decided to intercept the accelerator globally and then just trigger the menu item.

### DIFF
--- a/applications/desktop/src/main/menu.ts
+++ b/applications/desktop/src/main/menu.ts
@@ -6,7 +6,9 @@ import {
   BrowserWindow,
   dialog,
   FileFilter,
+  globalShortcut,
   Menu,
+  MenuItemConstructorOptions,
   shell,
   WebContents
 } from "electron";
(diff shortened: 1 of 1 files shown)


captainsafia:
@CrystallineCat I assume this also works when the paste happens via a keyboard shortcut?

CrystallineCat:
Yes, this works on `Ctrl-Shift-V`. I tested `Ctrl-C`, `Ctrl-V`, `Ctrl-Shift-C` as well (in various combinations), and they work as expected.

captainsafia:
@CrystallineCat Got it! Thanks for clarifying and thanks for putting in the fix.

captainsafia (approved):
:shipit:

rgbkrk (approved):

rgbkrk:
🎉 🍝

PR #4744, opened 2019-12-06
include the wire protocol in our messaging lib

rgbkrk:
This brings the wire protocol over to nteract as individual functions, with an `encode` and a `decode`. I wrote this code untyped originally within https://github.com/rgbkrk/trying-out-new-zeromq-with-jupyter.

~~One thing to possibly be wary of -- I'm not sure if my `require('crypto')` will impact builds of the frontend. You would hope that webpack will tree shake that right on out of there.~~ We're all set on this front.

* [x] Fully type
* [x] Test it all!
* [x] Determine if this impacts the frontend build(s)

-------

The main motivation here is to get us off of `jmp` and onto a library we control for the pretty simple use case of encoding and decoding wire protocol messages in node.js

```js
const message = encode(executeRequest(`print("hello")`));
const reply = decode(await shell_socket.send(message));
```

My follow on PR to this will be the `async` `await`-able ZeroMQ sockets we can incorporate in the desktop app.

### DIFF
--- a/packages/messaging/src/wire-protocol.ts
+++ b/packages/messaging/src/wire-protocol.ts
@@ -1,0 +1,175 @@
+const crypt = require("crypto");
+
+import { JupyterMessageHeader, MessageType } from "./types";
+
+const WIRE_PROTOCOL_DELIMITER = "<IDS|MSG>";
+
+// There must be at least 5 frames after the delimiter:
+// HMAC signature, serialized header, serialized parent header,
+// serialized metadata, serialized content
+const REQUIRED_NUMBER_OF_MESSAGE_FRAMES = 5;
+
+function toJSON(value: any) {
+  return JSON.parse(value.toString());
+}
+
+export interface RawJupyterMessage<
+  MT extends MessageType = MessageType,
+  C = any
+> {
+  header: JupyterMessageHeader<MT>;
+  parent_header:
+    | JupyterMessageHeader<any>
+    | {
+        msg_id?: string;
+      };
+  metadata: object;
+  content: C;
+  buffers: Array<Buffer>;
+  idents: Array<Buffer>;
+}
+
+function initializeMessage(
+  _message: Partial<RawJupyterMessage>
+): RawJupyterMessage {
+  const message = Object.assign(
+    {},
+    {
+      header: {},
+      parent_header: {},
+      metadata: {},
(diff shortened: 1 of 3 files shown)


rgbkrk:
Making sure the frontend build is a-ok using the jupyter extension now.

rgbkrk:
Frontend builds are looking good, webpack does the right thing.

packages/messaging/src/wire-protocol.ts
```diff
@@ -0,0 +1,153 @@
...
+      idents: [],
+      buffers: []
+    },
+    _message
+  );
+
+  return message;
+}
+
+function identifyHMACScheme(_scheme: string) {
```
captainsafia, line 58: Can we add docstrings to the methods defined in this file?

packages/messaging/src/wire-protocol.ts
```diff
@@ -0,0 +1,153 @@
...
+  const idents = [];
+  for (i = 0; i < messageFrames.length; i++) {
+    var frame = messageFrames[i];
+    if (frame.toString() === WIRE_PROTOCOL_DELIMITER) {
+      break;
+    }
+    idents.push(frame);
+  }
+
+  if (messageFrames.length - i < 5) {
```
captainsafia: Can you place `5` in a constant value that explains what it is?

packages/messaging/src/wire-protocol.ts
```diff
@@ -0,0 +1,153 @@
...
+    buffers: Array.prototype.slice.apply(messageFrames, [i + 6])
+  });
+
+  return message;
+}
+
+export function encode(
+  _message: RawJupyterMessage,
+  _scheme = "sha256",
+  key = ""
```
captainsafia: Is there a reason we are defaulting this to an empty string as opposed to `null` or `undefined`?

captainsafia (approved):
Looks good overall. Posted some comments inline.

packages/messaging/src/wire-protocol.ts
rgbkrk: This is a key sign I went from pure javascript to TypeScript. 😆 

Switching this to `key?: string` now.

rgbkrk:
Thank you for the review Safia!

PR #4746, opened 2019-12-06
Add initial implementation of display name-based overrides

captainsafia:
An idea I had after thinking through more extensibility scenarios for the new @nteract/stateful-components package.

This is an implementation but also a RFC. What are thoughts on this approach?

### DIFF
--- a/packages/stateful-components/__docs__/extensibility.md
+++ b/packages/stateful-components/__docs__/extensibility.md
@@ -2,7 +2,7 @@
 
 This suite of React components provides several points of extensibility for the UI. This document covers them here.
 
-## Extending editors
+## Extending Editors
 
 nteract ships with the CodeMirror editor in its desktop and Jupyter extensions. However, you might want to use an alternative editor in your own notebook based UI. To add your own editor to the nteract UI, create a React component that encompasses the editor.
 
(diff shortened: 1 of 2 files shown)


packages/stateful-components/__docs__/extensibility.md
```diff
@@ -68,3 +68,44 @@ class MyCells extends React.Component {
...
+Display name-based overrides are currently enabled for the following parent and child components.
+
+| Parent Component | Child Component |
+| ---------------- | --------------- |
+| CodeCell         | Pagers          |
+| CodeCell         | Prompt          |
+| CodeCell         | Outputs         |
+| CodeCell         | Editors         |
+
+Since this override interface depends on having the display name on the React component set to the name of the child component, be sure that the display names for your overrides are set correctly.
```
rgbkrk, line 111: Sweet docs!

packages/stateful-components/src/cells/code-cell.tsx
```diff
@@ -13,12 +13,15 @@ import Outputs from "../outputs";
 import Pagers from "../outputs/pagers";
 import InputPrompts from "../outputs/input-prompts";
 import Input from "../inputs/input";
+import React from "react";
+import React from "react";
```
rgbkrk: Let's only react once!

rgbkrk (approved):

packages/stateful-components/src/cells/code-cell.tsx
captainsafia: Whoops. Good catch. Can't have too much React.

PR #4751, opened 2019-12-07
Upgrade typescript 3.2.2 -> 3.7.3

CrystallineCat:
Upgrade typescript to 3.7.3 (latest).

This is required for compatibility with current vega* packages, as they are using newer features in their type definitions.

I chose to upgrade to latest instead of minimally for future compatibility with other projects using newer features; i. e. if we make bigger upgrade jumps, we will have to make fewer of them.

### DIFF
--- a/packages/data-explorer/src/charts/bar.tsx
+++ b/packages/data-explorer/src/charts/bar.tsx
@@ -149,7 +149,7 @@ export const semioticBarChart = (
         return (
           <TooltipContent x={hoveredDatapoint.x} y={hoveredDatapoint.y}>
             <div
-              style={{ heightMax: "300px", display: "flex", flexWrap: "wrap" }}
+              style={{ maxHeight: "300px", display: "flex", flexWrap: "wrap" }}
             >
               {combinedOptions.map((dim: { name: string }, index: number) => (
                 <div
--- a/packages/editor/src/mode/ipython.ts
+++ b/packages/editor/src/mode/ipython.ts
@@ -20,7 +20,6 @@ CodeMirror.defineMode(
     });
     return CodeMirror.getMode(conf, ipythonConf);
   },
-  "python"
 );
 
 // @ts-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ junit.xml
 
 # ignore notebooks created while running dev copy of the desktop app
 applications/desktop/*.ipynb
+
+# Typescript incremental build data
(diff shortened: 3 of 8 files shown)


CrystallineCat:
Ooooh, and 3.7 has *Optional Chaining*, I just noticed... shiny! Want! 😸😛

domoritz:
And the upgrade uncovered a bug in the style name 🎉.

captainsafia (approved):
Thanks!

packages/data-explorer/src/charts/bar.tsx
```diff
@@ -149,7 +149,7 @@ export const semioticBarChart = (
         return (
           <TooltipContent x={hoveredDatapoint.x} y={hoveredDatapoint.y}>
             <div
-              style={{ heightMax: "300px", display: "flex", flexWrap: "wrap" }}
+              style={{ maxHeight: "300px", display: "flex", flexWrap: "wrap" }}
```
rgbkrk, line 152: Slick find TS.

packages/editor/src/mode/ipython.ts
```diff
@@ -20,7 +20,6 @@ CodeMirror.defineMode(
     });
     return CodeMirror.getMode(conf, ipythonConf);
   },
-  "python"
```
rgbkrk, line 23: I'm assuming this is covered by the `name: "python"` above then?

rgbkrk (approved):

packages/editor/src/mode/ipython.ts
CrystallineCat: That depends what the parameter was supposed to be doing. Since the `defineMode()` method only has two parameters, it didn't do anything and the change is a no-op. The `name: "python"` retrieves the mode object for that mode, which the new mode is base on.

Looking closer, I'm curious why TS errored here despite the `@ts-ignore`. In any case, the methods do seem to have type definitions now, so I'm going to make a PR removing the outdated comment.

CrystallineCat: Actually, I misread the comment; the method does accept extra arguments for dependencies. This is both undocumented and deprecated, though, and all it seems to do per source is to `require()` the modes specified. Since we have `import "codemirror/mode/python/python"` anyway, I still think this is a no-op.

rgbkrk: Thank you @CrystallineCat!

PR #4752, opened 2019-12-07
Add ability to undo cell deletes for 10s after the action

### LINKED ISSUE #1038: Add "Undo" Cell delete
Hi guys,

I see this commit https://github.com/nteract/nteract/pull/627 but that's probably for general text undo
Is this feature on the radar (don't want to say this but i.e. like in jupyter menu edit > undo cell delete)?

Thanks

CrystallineCat:
Fixes #1038.

This will replace the cell with a note and an undo button. If that note is still displayed after 10s, it'll do the actual deletion.

Screenshot from 2019-12-07 16-12-47


packages/commutable/src/cells.ts
```diff
@@ -9,9 +9,21 @@ import {
...
+}
+
+const makeCellTransientParam = Record<TransientCellParam>({
+  deleting: false,
+});
+
 /* CodeCell Record Boilerplate */
 
-export interface CodeCellParams {
+export interface CodeCellParams extends TransientCellParams {
```
captainsafia: Is there a reason you opted to use a new `transient` model in each cell instead of the existing top-level `transient` property?

packages/notebook-app-component/src/decorators/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface CellAddress {
+  id: string;
+  contentRef: ContentRef;
+}
+
+export const cellAddress = ({ id, contentRef }: CellAddress) => (
+  { id, contentRef }
+);
+
+export const cellFromState = (
```
captainsafia: Can we move this into the `selectors` package?

packages/notebook-app-component/src/decorators/undoable-delete.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
```
captainsafia, line 1: Oh neat! I'm really a fan of this decorator pattern! It opens up lots of interesting patterns for us to build features.

packages/notebook-app-component/src/decorators/undoable-delete.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+class TimedUndoableDelete
+  extends React.PureComponent<Props> {
+
+  timer: NodeJS.Timeout | null = null;
+
+  componentDidMount(): void {
+    this.timer = setTimeout(
+      this.props.doDelete,
+      this.props.secondsDelay * 1000,
```
captainsafia, line 20: Nit: Any chance we can make this configurable with a default in much the same way the auto-save interval is currently configured?

captainsafia (commented):
Nifty implementation! Posted some comments inline.

packages/notebook-app-component/src/decorators/utils.ts
CrystallineCat: Done.

packages/notebook-app-component/src/decorators/undoable-delete.tsx
CrystallineCat: Done.

packages/commutable/src/cells.ts
CrystallineCat: You mean the one in `DocumentRecordProps`? I wasn't really aware of it, tbh. I was only looking in the cell types in `commutable`. Also, attaching it to the cell means that the cell's information is more self-contained and we don't have to worry about syncing metadata stored elsewhere.

Do you want me to move it?

captainsafia: I see. I think it makes more sense to put it in the transient state since this is more so an app-action then a document-related event.

rgbkrk:
Thanks @CrystallineCat, this is neat!

packages/commutable/src/cells.ts
CrystallineCat: Hmm... I can see your point, but I'm unsure how to do so cleanly. In order to skip delete-marked cells when saving, `commutable` needs to have access to the state, and if I move it, I'd have to introduce a reference to app state to commutable, or not skip marked cells upon save (which is surprising behaviour).

I could remove the cells from the document and then potentially re-insert them, but that means more bookkeeping, especially regarding cell order (e.g. cells get deleted, inserted, moved, undeleted...).

Of everything I can think of, I like the way it currently is best.

captainsafia: >  In order to skip delete-marked cells when saving, commutable needs to have access to the state, and if I move it

Ah I see! This hadn't crossed my mind when I was reviewing it. I was worried that the transient state was being serialized to the doc but I realize in looking closer at the PR that it is not.

In that case, can we move the `transient` property under the metadata of each cell and have it namespaced under the `nteract` scope. Something like.

```js
{
  source: "Some cell code here",
  metadata: {
    nteract: {
      transient: { deleting: true }
    }
  }
}
```

That should strike a balance between having access to the data in commutable and not adding anything extraneous to the `CellParams` properties.

Thoughts?

CrystallineCat: That seems reasonable! Thanks for your help!

captainsafia (approved):
Nice. Very excited to get this merged and start using the cell decorator pattern some more.

greenstick:
The problem with not having an undo button & an autosave option is that you can't just close and reload the notebook. Further, on a Mac, CMD + Shift + D (delete cell) is dangerously close to CMD + Shift + C (copy) and CMD + Shift + x (cut).

I'm glad this issue has been _somewhat_ addressed in newer versions. I seriously wish I had known about this and updated – I just lost hours of work due to there being no undo feature. I think my only question is, why 10 seconds? Is storing a history too memory intensive? Is it something to do with Electron? Another nice feature might be a warning dialogue that explains there will be only 10 seconds to reverse the deletion, or something to that effect, and allows the user to confirm they want to delete the cell .

PR #4756, opened 2019-12-09
More overrides and clean-up for @nteract/stateful-components

captainsafia:
- Adds class names to each component for styling
- Adds support for overrides at more levels
- Updates documents

### DIFF
--- a/packages/stateful-components/__docs__/examples.md
+++ b/packages/stateful-components/__docs__/examples.md
@@ -23,11 +23,19 @@ class MyApp extends React.Component {
 ```
 import Notebook, { CodeCell, MarkdownCell, RawCell } from "@nteract/stateful-components";
 
-class Editors extends React.Component {
-    static displayName = "Editors";
+class CodeEditor extends React.Component {
+    static displayName = "Editor";
 
     render() {
-        return <MyCustomEditor />;
+        return <div><Editor />{this.props.editorButtons.map(Button => <Button/>)}</div>;
+    }
+}
+
+class MDEditor extends React.Component {
+    static displayName = "Editor";
+
+    render() {
+        return <div><Editor />{this.props.editorButtons.map(Button => <Button/>)}</div>;
     }
 }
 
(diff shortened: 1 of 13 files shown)


packages/stateful-components/__docs__/examples.md
```diff
@@ -0,0 +1,112 @@
...
+class MyApp extends React.Component {
+    render() {
+        return <React.Fragment>
+            <MySiderbar />
+            <Notebook />
+            <MyLeftPanel />
+        </React.Fragment>;
+    }
+}
+```
```
rgbkrk, line 21: Slick writeup. I love the progression from basics to more advanced ways of composing the nteract components.

packages/stateful-components/src/cells/cells.tsx
```diff
@@ -9,6 +9,8 @@ import RawCell from "./raw-cell";
 import MarkdownCell from "./markdown-cell";
 import CodeCell from "./code-cell";
 
+import childWithDisplayName from "./markdown-cell";
```
rgbkrk: Considering how `childWithDisplayName` gets used, it seems to not belong in the `markdown-cell` module.

rgbkrk: Oh I see, there's a more canonical location. This was probably a VS Code autoimport thing.

captainsafia: Oh gosh! Good catch. This should be importing from the `pickers` directory.

packages/stateful-components/src/pickers/display-name.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const childWithDisplayName = (
```
rgbkrk, line 3: The pattern being used here is to key off of the `displayName` to pick a child? Interesting pattern, though I'm not sure how I feel about it. I suppose it means we should export a collection of the constants we're checking for in the `displayName`. 🤔

captainsafia: Yeah, this was a weird one. I was trying to go for something that would use built-in properties as much as possible.

Kyle linked me to [this article](https://medium.com/@martin_hotell/react-children-composition-patterns-with-typescript-56dfc8923c64) and I feel approach #3 linked in it might be a better approach to use for this overriding interface.

PR #4757, opened 2019-12-10
Add support for vega-lite v4

### LINKED ISSUE #4750: Add support for vega-lite v4
This probably involves:
- [ ] Maybe make an external repository for vega-lite@3?
- [ ] Update typescript (vega-lite@4 uses a newer version)
- [ ] Update vega/vega-lite/vega-embed

I'm going to take a look at this.

CrystallineCat:
Fixes #4750.


packages/transform-vega/src/react.tsx
```diff
@@ -62,14 +62,15 @@ export class VegaEmbed<T extends VegaMediaType>
...
   componentWillUnmount(): void {
-    if (
-      this.embedResult &&
-      this.embedResult.view &&
-      this.embedResult.view.finalize
-    ) {
-      this.embedResult.view.finalize();
-    }
+    if (this.embedResult) {
+      if (this.embedResult.finalize) {
```
domoritz, line 79: Thanks!

CrystallineCat:
How do I get my dependency tree to use a consistent version of `vega`/`vega-lite` in my [vega monorepo](https://github.com/nteract/any-vega)? I.e. not having sub-packages depend on a different version, like this:
```
any-vega > vg4vl2 > ...
... > vega: 4.4.0
... > vega-lite: 2.7.0
... > vega-embed > vega: 3.3.1
... > vega-embed > vega-lite: 1.3.1

any-vega > vg5vl3 > ...
... > vega: 5.9.0
... > vega-lite: 3.4.0
... > vega-embed > vega: 3.3.1
... > vega-embed > vega-lite: 1.3.1

any-vega > vg5vl4 > ...
... > vega: 5.9.0
... > vega-lite: 4.0.0
... > vega-themes > vega: 3.3.1
... > vega-themes > vega-lite: 1.3.1
```
Not sure if versions of other `vega-*` packages match, either. I tried using `resolutions`, but that didn't help. 

@domoritz @captainsafia Hoping one of you knows, cause I spent two days on this so far, and can't seem to get any further.

wadethestealth:
@CrystallineCat I think it could be because of a yarn or package.lock. If you delete that file and reinstall they should all just converge to their highest supported version.

CrystallineCat:
@wadethestealth Thanks for your reply, however that wasn't it. I deleted that file and the `node_modules` folders pretty religiously. :smirk_cat:

Any way, I got desperate enough to see if the v2 yarn branch does better, and it did. :smile_cat: But that broke rollup... :angry: ...so I switched to webpack... :smirk_cat: in the end, I got it working :smile_cat:

domoritz:
Wow, I’m sorry for all the trouble. If there is anything I can do from the Vega side, I’m happy to look into it.

CrystallineCat:
Ok, I think this is ready. :joy_cat:

All vega and vega-lite versions now render; error reporting is a bit enhanced (if the embedder function throws, it'll display an error instead of the graph); contents of vega mimetypes are now stored as strings in the MediaBundle, and only JSON parsed right before passing them to vega-embed, which skips recursively freezing specs with potentially large datasets, so hopefully performance is better now.

Had a new bunch of type errors all across the repo, for some reason, which I fixed hopefully correctly...

CrystallineCat:
@domoritz Thank you! I got it working in the end. :cat: Biggest problem was probably my insistence to have all vega/vega-lite versions in a single repository... :stuck_out_tongue: but I'm really glad now that I'm not going to have to maintain five different ones.

applications/desktop/src/notebook/epics/zeromq-kernels.ts
```diff
@@ -352,8 +352,8 @@ function killSpawn(spawn: ChildProcess): void {
   if (spawn.stdin && spawn.stdin.destroy) {
     spawn.stdin.destroy();
   }
-  spawn.stdout.destroy();
-  spawn.stderr.destroy();
+  spawn.stdout?.destroy();
+  spawn.stderr?.destroy();
```
captainsafia, line 356: Nifty!

packages/commutable/src/primitives.ts
```diff
@@ -181,7 +183,9 @@ export function createFrozenMediaBundle(
   const bundle: MediaBundle = {};
 
   for (const key in mediaBundle) {
-    if (typeof mediaBundle[key] === "string") {
+    if (/^application\/vnd.vega(.*\+)json$/.test(key)) {
```
captainsafia: Nit: Assign this regex to a variable.

packages/messaging/src/index.ts
```diff
@@ -172,7 +172,7 @@ export const ofMessageType = (
             return;
           }
 
-          if (messageTypes.indexOf(msg.header.msg_type) !== -1) {
+          if (messageTypes.includes(msg.header.msg_type as any)) {
```
captainsafia, line 175: Can we explicitly cast this as `MessageType`?

packages/presentational-components/src/components/error.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const Error = (props: ErrorProps) =>
+  props.error
+    ?
+      <p className={props.className}>
+        <b>{props.error.name ?? "Error"}:</b> {props.error.message}
+      </p>
+    : null;
+
+const StyledError = styled(Error)`
+  background: var(--jp-error-color3);
```
captainsafia: Can we add a CSS color variable in the themes and use that here instead of relying on the ones from the ipywidgets?

captainsafia (approved):
Looks good from my end overall. Left some comments and it looks like this needs a rebase now.

@domoritz Do you have any other Vega-specific feedback?

Thanks for getting this in!

packages/transform-vega/src/mime.ts
```diff
@@ -6,7 +6,9 @@ export type VegaMediaType =
   | "application/vnd.vega.v5+json"
   | "application/vnd.vegalite.v1+json"
   | "application/vnd.vegalite.v2+json"
-  | "application/vnd.vegalite.v3+json";
+  | "application/vnd.vegalite.v3+json"
+  | "application/vnd.vegalite.v4+json"
+  ;
```
domoritz, line 11: Is this intended?

packages/transform-vega/src/react.tsx
```diff
@@ -1,11 +1,11 @@
+import { Error } from "@nteract/presentational-components";
 import * as React from "react";
-import { Result } from "vega-embed";
 import { embed, VegaOptions } from "./external";
 import { VegaMediaType } from "./mime";
 
 /** Props needed for embedding a certain Vega (Lite) media type. */
```
domoritz:
```suggestion
/** Props needed for embedding a certain Vega(-Lite) media type. */
```

domoritz:
Looks great to me.

packages/messaging/src/index.ts
CrystallineCat: It's already `MessageType`, however the array is `Array<T | [T]>`, so each element is expected to be `T | [T]` with `T extends MessageType`, which I would expect to pass, but doesn't for some reason :woman_shrugging:. If I explicitly cast to `MessageType` it still fails.

It shouldn't really matter anyway, since `includes` should be able to handle any typed argument. The real fix would be to fix the function signature to simplify the typing (I don't see any reason for it to be this complex), but that would of course require fixing callers.

I'm going to leave it for now.

packages/presentational-components/src/components/error.tsx
CrystallineCat: Didn't realize where they were from; just looked for colours in the dev tools.

packages/transform-vega/src/mime.ts
CrystallineCat: I'm not sure what you mean?

CrystallineCat:
Thanks for the reviews, @captainsafia and @domoritz!

domoritz:
🎉 Thank you!

davidanthoff:
Awesome!

Is there a build somewhere that I could download to test this with the Julia VegaLite.jl package?

Also, just for my own understanding: the strategy in nteract is to ship vega3, vega4, vega5, vega-lite3 and vega-lite4 all in the box, and then use the version of the code that corresponds to the MIME version embedded, right? I.e. you are not falling back on vega-lite 4 to render a vega-lite 3 MIME type, instead you render that with vega-lite 3?

domoritz:
That's my understanding as well. @CrystallineCat solved the version issue with https://github.com/nteract/any-vega.

jasongrout:
@domoritz - just to be clear, what is the EOL maintenance window for older versions of vega/vega-lite? For example, if there is a security issue, how far back do you plan to release patch versions?

domoritz:
We will do that on a case-by-case basis. We can backport critical security patches to the latest minor version for each major version.

captainsafia:
@davidanthoff You can follow the dev instructions for the nteract desktop app in the README and try this out in a local dev build. Let me know if you need help with this.

davidanthoff:
Ok, that was easy :) I just checked, it all works.

PR #4760, opened 2019-12-11
Update composition pattern used in @nteract/stateful-components

captainsafia:
After getting some feedback from @rgbkrk, I've updated the stateful-components package to use named children to enable child overrides instead of the existing display name overrides.


packages/stateful-components/src/cells/code-cell.tsx
```diff
@@ -44,26 +50,20 @@ export default class CodeCell extends React.Component<ComponentProps> {
...
-    const EditorOverride = childWithDisplayName(children, "Editor");
-    const InputPromptsOverride = childWithDisplayName(children, "InputPrompts");
+    let editor, prompt, pagers, inputPrompts, outputs;
+    if (children) {
+      editor = children.editor;
+      prompt = children.prompt;
+      pagers = children.pagers;
+      inputPrompts = children.inputPrompts;
+      outputs = children.outputs;
+    }
```
rgbkrk, line 60: If `children` is set but e.g. `prompt` is not, where does the default come from?

packages/stateful-components/src/cells/code-cell.tsx
```diff
@@ -44,26 +50,20 @@ export default class CodeCell extends React.Component<ComponentProps> {
...
       <div>
         <Input id={id} contentRef={contentRef} className="nteract-cell-input">
-          {PromptOverride ? (
-            <PromptOverride
-              id={id}
-              contentRef={contentRef}
-              className="nteract-cell-prompt"
-            />
+          {prompt ? (
+            <React.Fragment>{prompt}</React.Fragment>
```
rgbkrk, line 66: Well that answers my question.

rgbkrk (approved):
Cool, I think I'm liking this approach.

PR #4761, opened 2019-12-11
Fixes `hidden` attribute default value

mscottford:

The documentation indicates that the default value for the `Input` component's `hidden` attribute was `true`, but then in a later paragraph said that you can also set it to true. I'm assuming from context that the actual default value for that attribute is `false`, so I've made a correction. I also tried to make it more clear which attribute is being talked about in the paragraph where it's mentioned that the value can be set to `true`. This is something that I was confused about because the attribute did not appear in the previous code sample.


captainsafia (approved):
Good catch! Thanks.

rgbkrk:
Thanks @mscottford!

PR #4764, opened 2019-12-11
Add warning when using npm inside repo

### LINKED ISSUE #4523: Prevent npm install (since Yarn is required)
Using yarn is required due to how lerna and yarn workspaces are used to support the mono-repo structure. Running `npm install` is still possible but leaves the working directory in a horribly broken state that can't be fixed just by running `yarn install` afterwards (starting over with a clean git clone is recommended instead).

See https://github.com/yarnpkg/yarn/issues/4895 for a couple of possible solutions.

captainsafia:
Closes https://github.com/nteract/nteract/issues/4523.

### DIFF
--- a/package.json
+++ b/package.json
@@ -253,7 +253,9 @@
     "yargs": "^14.0.0"
   },
   "engines": {
-    "node": ">=8.12.0"
+    "node": ">=8.12.0",
+    "npm": "please-use-yarn",
+    "yarn": ">= 1.17.0"
   },
   "workspaces": {
     "packages": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -253,7 +253,9 @@
     "yargs": "^14.0.0"
   },
   "engines": {
-    "node": ">=8.12.0"
+    "node": ">=8.12.0",
+    "npm": "please-use-yarn",
```
rgbkrk, line 257: Hah

package.json
```diff
@@ -253,7 +253,9 @@
...
+    "yarn": ">= 1.17.0"
```
rgbkrk: Make this `>= 1.16.0` to make this pass on CircleCI for now.

PR #4765, opened 2019-12-12
Save hidden states to name-spaced metadata

### LINKED ISSUE #4294: Update where hidden status is stored
https://github.com/nteract/nteract/blob/e7acc822f0a5bf58bd941e5cd02ce2bfaf7e5a7d/packages/reducers/src/core/entities/contents/notebook.ts#L637-L641

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in e7acc822f0a5bf58bd941e5cd02ce2bfaf7e5a7d when #4264 was merged. cc @stormpython.

captainsafia:
Closes https://github.com/nteract/nteract/issues/4294.

This makes it so that we are saving the hidden states to the namespaced values that exist in in the nbformat.

Should I remove the existing `outputHidden`/`inputHidden` stuff altogether or remove it in case people are using it downstream?

Removing it would require a major version release but maybe that's fine...


rgbkrk:
This is one of the annoying things where we really have to support _all_ the metadata flags for hidden because of old+new notebooks that will work across multiple frontends. This to me means two possible approaches:

* We maintain all keys for `*hidden` in our state tree
* We maintain only _one_ key and then normalize it on save to include all of them.

captainsafia:
> We maintain only one key and then normalize it on save to include all of them.

Personally, I'm partial to this approach since `source_hidden` and `outputs_hidden` are now officiallly part of the spec.

I debated removing our current `outputHidden` and `inputHidden` altogether but opted to include them so users can maintain compatibility across notebooks.

Perhaps I should add the normalization logic from our existing pattern to the one in the spec to commutable so this is more standardized.

I can update the PR to do that if that seems like a sensible approach.

captainsafia:
Alright! I've moved stuff over to use the namespaced hidden state properties that are defined in the Jupyter notebook spec.

We can merge this whenever it is ready.

rgbkrk (approved):

PR #4778, opened 2019-12-18
React + Redux Devtool

wadethestealth:
I used my previous experience working with atom devtools to create this local manual build of getting react and redux devtools working in electron.

My review of this PR will explain all that I know.

If anyone wants to build further on this to make it cross platform or at least windows system they may. I would suggest most of this being hidden behind a yarn script similar to the redux script you already have, but all this code works in my local environment using `yarn app:desktop`.

I haven't tried, but I am unsure of how reliable this setup is to the watch and dynamic build as I am unfamiliar with that tooling.

The purpose of this PR is not for me to add the tools to the main dev process, but to show case how to do it so that if someone wishes they can

### DIFF
--- a/applications/desktop/package.json
+++ b/applications/desktop/package.json
@@ -121,6 +121,7 @@
   },
   "devDependencies": {
     "electron": "5.0.12",
-    "file-loader": "^5.0.0"
+    "file-loader": "^5.0.0",
+    "remote-redux-devtools": "^0.5.16"
   }
 }
--- a/applications/desktop/src/main/index.ts
+++ b/applications/desktop/src/main/index.ts
@@ -87,6 +87,12 @@ ipc.on("show-message-box", (event: Event, arg: any) => {
 
 app.on("ready", initAutoUpdater);
 
+
+app.on("ready", () => {
+  BrowserWindow.addDevToolsExtension("C:\\Users\\Wade\\AppData\\Local\\Google\\Chrome\\User Data\\Default\\Extensions\\fmkadmapgofadopljbjfkapdkoienihi\\4.2.1_0");
+  BrowserWindow.addDevToolsExtension("C:\\Users\\Wade\\AppData\\Local\\Google\\Chrome\\User Data\\Default\\Extensions\\lmhkpmbekcpmknklioeibfkpmmfibljd\\2.17.0_0");
+});
+
 const electronReady$ = new Observable(observer => {
   app.on("ready", (event: Event) => observer.next(event));
 });
--- a/applications/desktop/src/main/launch.ts
+++ b/applications/desktop/src/main/launch.ts
@@ -33,6 +33,7 @@ export function launch(filename?: string) {
     show: false,
     webPreferences: { nodeIntegration: true }
   });
+  win.webContents.openDevTools();
 
   win.once("ready-to-show", () => {
     win.show();
--- a/applications/desktop/src/main/store.ts
+++ b/applications/desktop/src/main/store.ts
@@ -1,5 +1,6 @@
 import { middlewares as coreMiddlewares } from "@nteract/core";
 import { applyMiddleware, compose, createStore, Middleware } from "redux";
+import { composeWithDevTools } from 'remote-redux-devtools';
 
 import reducers from "./reducers";
 
@@ -11,5 +12,5 @@ if (process.env.DEBUG === "true") {
 }
 
 export default function configureStore() {
-  return createStore(reducers, compose(applyMiddleware(...middlewares)));
+  return createStore(reducers, composeWithDevTools({realtime: true, name:"main"})(applyMiddleware(...middlewares)));
 }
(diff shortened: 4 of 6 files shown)


applications/desktop/package.json
```diff
@@ -121,6 +121,7 @@
   },
   "devDependencies": {
     "electron": "5.0.12",
-    "file-loader": "^5.0.0"
+    "file-loader": "^5.0.0",
+    "remote-redux-devtools": "^0.5.16"
```
wadethestealth, line 125: This package is for monitoring the main electron store since it is not used with react or in a browser context. This is basically the only official solution I found to viewing outside a browser context.

applications/desktop/src/main/index.ts
```diff
@@ -87,6 +87,12 @@ ipc.on("show-message-box", (event: Event, arg: any) => {
 
 app.on("ready", initAutoUpdater);
 
+
+app.on("ready", () => {
+  BrowserWindow.addDevToolsExtension("C:\\Users\\Wade\\AppData\\Local\\Google\\Chrome\\User Data\\Default\\Extensions\\fmkadmapgofadopljbjfkapdkoienihi\\4.2.1_0");
+  BrowserWindow.addDevToolsExtension("C:\\Users\\Wade\\AppData\\Local\\Google\\Chrome\\User Data\\Default\\Extensions\\lmhkpmbekcpmknklioeibfkpmmfibljd\\2.17.0_0");
+});
```
wadethestealth, line 94: Electron is built on Chrome and electron exposes some sets of native chrome apis used by extensions, however, it is not a full coverage and therefore not all extensions work.

To add an extension to all renderer processes, we use an electron api of browser window called addDevToolsExtension this just loads in chrome extensions into your electron browser windows.

In nteract/nteract:
- Currently the only Redux store in a renderer process is that of `desktop/src/notebook`. 
- Obviously all React is used in a renderer process because react is for the dom (aka browser window).

To get the extensions, I opened my chrome web browser and downloaded the extensions I wanted from the chrome extension store.

After this I went to where my extensions are stored locally and copy pasted the path you see here. Note these above are only for my exact user profile running windows.

The first path is react devtools v4.2.1 and the second is redux devtools v2.17.0.

To see more about this step visit https://electronjs.org/docs/tutorial/devtools-extension#how-to-load-a-devtools-extension

applications/desktop/src/main/launch.ts
```diff
@@ -33,6 +33,7 @@ export function launch(filename?: string) {
     show: false,
     webPreferences: { nodeIntegration: true }
   });
+  win.webContents.openDevTools();
```
wadethestealth, line 36: this line automatically opens the console/chrome devtools for me. Its useful to me as I can set earlier break points in the source because I don't have to provide input to open it.

applications/desktop/src/main/store.ts
```diff
@@ -11,5 +12,5 @@ if (process.env.DEBUG === "true") {
 }
 
 export default function configureStore() {
-  return createStore(reducers, compose(applyMiddleware(...middlewares)));
+  return createStore(reducers, composeWithDevTools({realtime: true, name:"main"})(applyMiddleware(...middlewares)));
```
wadethestealth, line 15: Not much to this just hooks up a socket to monitor and alter a store. This is only required from the npm package since this is the store on the main electron process which does not have access to things like extensions or browser function.

Read the comment on this packages installation and its docs for more info

wadethestealth: To view this store you must go to this link in your browser `http://remotedev.io/local/` it will automatically detect the running store for you.

applications/desktop/src/notebook/store.ts
```diff
@@ -51,10 +51,13 @@ const rootReducer = combineReducers({
...
   initialState: Partial<DesktopNotebookAppState>
 ): DesktopStore {
+  const composeEnhancers = (window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
   const store = createStore(
     rootReducer,
     (initialState as unknown) as any,
-    applyMiddleware(...middlewares)
+    composeEnhancers(
+      applyMiddleware(...middlewares)
+    )
```
wadethestealth, line 60: Hooks the redux devtools extension into the package/notebook-app store. To view any of the renderer process stores you can use the redux tab in the chrome devtool. Each window will show its own store, and if you try to switch you may cause the extension to crash. also I recommend not viewing the state tab in raw form its laggy for large states like this one. If you cant see this store you may need to open the extension settings in chrome and enable the file:// viewing toggle switch

package.json
```diff
@@ -224,6 +224,7 @@
     "regenerator-runtime": "^0.13.0",
     "remark-parse": "^7.0.0",
     "remark-stringify": "^7.0.0",
+    "remote-redux-devtools": "^0.5.16",
```
wadethestealth, line 227: this may be unnecessary unsure how the dependencies cross since this is root level but the package that needs this is application/desktop.

wadethestealth (commented):
There isn't really much to say about react devtools as it just works out the box.

captainsafia:
Closing this for now. We can look back it for reference as needed.

We should try to get a cross-platform Redux/React devtool configuration on the desktop app sometimes soon.

PR #4784, opened 2019-12-19
Add monorepo-level documentation setup with mkdocs

### LINKED ISSUE #4782: Set up monorepo-level documentation generation
#### Context
Currently, we have the tutorial-style nteract docs published at docs.nteract.io from the nteract/docs repo. Unfortunately, this separation creates a lot of friction as we work towards improving our tutorial-style docs. I believe that we should move our tutorial-style docs to the monorepo, especially since a bulk of the packages and apps that we ship are located there. I'm largely inspired by the docs that I have started to setup for the new `@nteract/stateful-components` package ([link](https://github.com/nteract/nteract/tree/master/packages/stateful-components)).

#### Problem
Our tutorial-style documentation story can be improved. Normally, we can be better at:
- Writing showcase documentation for each of the individual packages
- Writing showcase documentation for implementations across individual packages
- "Here's how it's built" style documentation for our applications.

#### Solution
We can improve the user/developer experience here by setting up documentation generation at the monorepo level. I'm imagining something that would allow to write markdown in the `__docs__` directory under each subfolder in `packages` and `applications`. 

After much research, it appears that configuring [mkdocs](https://www.mkdocs.org/) with the [mkdocs-monorepo-plugin](https://spotify.github.io/mkdocs-monorepo-plugin/).

To get this setup we will need to:

1. Install mkdocs and configure a mkdocs config file at the root that builds the static site to a top-level `docs` directory.
2. Map over our standard zeit now deployment pattern from the component and package sites.
3. Install the mkdocs-monorepo-plugin and configure it to render the docs from @nteract/stateful-components.
4. Build and deploy with this config to now (without aliasing to the production URL).

Once this is done, we can start moving each of the items currently documenting in docs.nteract.io to their appropriate folder in the monorepo and wiring that up with the monorepo documentation setup.

Once that is done, we can officially make docs.nteract.io alias to the mkdocs-backed static site.

After that, the hard work begins, incrementally writing docs for each package.

captainsafia:
Closes #4782.

This is currently tagged as work-in-progress since I need to figure out how to hook up a deployment of this to zeit now. However, I was able to serve a local instance of the docs successfully.

EDIT: I realized that deploying this to zeit is pretty easy and also learned that we no longer have to do the hack where we keep a `now.json` file in an empty directory. Instead, we can rely on the project configuration that leaves on `zeit.co` and deploy our generated static content using `now directory-with-static-content --target production --name project-name`.


rgbkrk:
It's looking good and I'm looking forward to these docs!

Issue #4741, opened 2019-12-05
Make Jupyter-based completion epics editor agnostic


captainsafia:
This issue captures a shower-thought I had around the efforts to make the nteract notebook component more editor-agnostic.

Some of the work around stateful components (ref: #4731) but there's an another aspect to this that is not addressed: the completion epics that handle managing completion requests and replies from the Jupyter server.

Currently, this epics are embedded within the Codemirror component itself. As a result, when we want to light up new editor support in nteract, that code has to be copied over to each editor component (for example, the Monaco editor).

This kind of code duplication is not ideal. The "core" aspects of these epics include:

1. Sending the completion request to Jupyter.
2. Retrieving the completions response.
3. Handling timeouts and cancellations appropriately.

This functionality is editor agnostic and shouldn't be within the logic for an editor. The parts that are editor specific include:

1. How to fetch the current cursor position and editor content
2. How to display the completion results in the editor

The division of editor agnostic/specific codepaths related to completions makes me thing we should polish up the division here. I think we should do the following:

1. Move the editor-agnostic bits of completion epic to a `jupyterCompletionsEpic` that will live in our @nteract/epics package.
2. Dispatch events (or maybe Redux actions) from the editors when they need to send a completion request. This is how the editor will be responsible for sending it's cursor position and contents.
3. Dispatch events (or maybe Redux actions) out of the `jupyterCompletionsEpic` that each editor will listen to. It will contain the completion responses formatted in a certain pattern that each editor can then consume and process per its needs.

This approach reduces the amount of logic that editors need to implement in order to plug into nteract and reduces the amount of Jupyter-related work they do, making them easier to hook up with other completion providers if necessary or portable in environments where Jupyter completions are not supported.

captainsafia:
Chatted with @barryt2 about this and we realized a key thing to validate with this is any perf impact we might take with decoupling the completion fetch from the completion render. It might introduce some latency to the process. We should evaluate whether or not this latency is small enough to ignore when considering the maintainability benefits that we get with this.

PR #4788, opened 2019-12-19
Use @nteract/stateful-components in @nteract/notebook-app-component

captainsafia:
I'm leaving the WIP tag on this since I am still polishing things up. However, most of the core stuff is there and it is ready for review.

This pull request utilizes the @nteract/stateful-components package in the @nteract/notebook-app-component package. There are a couple of key things to note here:

1. The notebook-app-component package is no longer a single notebook app component. It contains a variety of implementations to showcase the flexibility of the decorator + render children overrides features in the application.
2. I'm now using render functions as the children of the overridable components. This struck a balance between creating a structured syntax that kept the props interface simple and made it easier to type.
3. Styles now happen via CSS instead of style components. They can still be migrated to styled-components but CSS is used to here to highlight uses for pure-CSS design.

I'll leave comments inline throughout the PR to explain certain things.


packages/styles/themes/default.css
```diff
@@ -0,0 +1,167 @@
+.nteract-cell-prompt {
```
captainsafia, line 1: This contains the styles from presentational-components ported into CSS.

packages/styles/themes/base.css
```diff
@@ -0,0 +1,144 @@
+.nteract-cell-prompt {
```
captainsafia, line 1: This contains the styles from presentational-components ported to CSS. These are only the styles that are layout related.

packages/stateful-components/src/notebook/status-bar.tsx
```diff
@@ -0,0 +1,81 @@
+import React from "react";
```
captainsafia, line 1: This is a new stateful container for the status bar. It's packaged as a ContextComponent in a pattern similar to the cell toolbar to allow API users flexibility in what kind of status bar UI they want to create.

packages/notebook-app-component/src/decorators/cell-creator/index.tsx
```diff
@@ -44,6 +42,7 @@ export const CellCreatorMenu = styled.div`
   pointer-events: all;
   position: relative;
   top: -5px;
+  z-index: 50;
```
MSeal, line 50: This looked a little suspicious -- I always had issues with hardcoded z-index in react components in the past. But given the other css rules inlined I am assuming this is always combined in a way that you want this in front?

packages/notebook-app-component/src/decorators/cell-creator/index.tsx
```diff
@@ -160,11 +165,33 @@ class CellCreator extends React.PureComponent<ConnectedProps> {
...
+  let first = false;
+
+  if (model && model.type === "notebook") {
+    const cellOrder = selectors.notebook.cellOrder(model);
+    const cellIndex = cellOrder.findIndex(cellId => cellId === id);
+    first = cellIndex === 0;
+  }
+
+  return {
+    first
```
MSeal: The return value seems odd given the function name. I'd either add a comment to the function explaining what the function does / returns or rename it to better match the in <-> out pattern here.

packages/notebook-app-component/src/decorators/hijack-scroll/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  focused: boolean;
+}
+
+interface DispatchProps {
+  selectCell: () => void;
+}
+
+type Props = ComponentProps & DispatchProps & StateProps;
+
+export class HijackScroll extends React.Component<Props> {
```
MSeal, line 26: Have we tested this on a few browser setups? Scroll hijacking is always finicky and prone to edge case issues.

packages/notebook-app-component/src/decorators/themer/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  theme: string;
+}
+
+export class Themer extends React.Component<ComponentProps & StateProps> {
+  render() {
+    if (this.props.theme === "dark") {
+      return (
+        <React.Fragment>
+          {this.props.children}
+          <DarkTheme />
```
MSeal, line 21: Not sure how the theme code applies in the rest of the codebase -- why the self-contained element here?

packages/stateful-components/src/inputs/editor.tsx
```diff
@@ -35,6 +47,15 @@ export class Editor extends React.Component<ComponentProps & StateProps> {
         // Already have a selection
         return;
       }
+
+      if (
```
MSeal, line 44: Might warrant a comment about what's being checked for here

packages/stateful-components/src/inputs/prompt.tsx
```diff
@@ -16,10 +16,26 @@ interface StateProps {
 
 export class Prompt extends React.Component<ComponentProps, StateProps> {
   render() {
-    const { children } = this.props;
     return (
       <div className="nteract-cell-prompt">
-        {React.cloneElement(children, this.props)}
+        {React.Children.map(this.props.children, child => {
```
MSeal, line 21: This also may warrant some comments on the logic being applied

packages/stateful-components/src/outputs/pagers.tsx
```diff
@@ -17,12 +17,24 @@ interface StateProps {
...
 export class Pagers extends React.PureComponent<ComponentProps & StateProps> {
   render() {
-    const { pagers, children } = this.props;
+    const { pagers } = this.props;
     return (
       <div className="nteract-cell-pagers">
         {pagers.map(pager => (
           <RichMedia data={pager.data} metadata={pager.metadata}>
-            {children}
+            {React.Children.map(this.props.children, child => {
```
MSeal, line 25: comment on logic here too

MSeal (commented):
I left a few generic comments and didn't try running the changes myself. But overall the changes look positive. I didn't notice there are very few test changes and I am guessing the test coverage here was low before? At some point we may want to invest in getting this logic unittested so it stays reliable, but I don't want to block getting changes out for it,

packages/notebook-app-component/src/decorators/hijack-scroll/index.tsx
rgbkrk: When this component was originally written it was only for electron so we used to be able to pretend it 💯 % checked across all supported browsers. A lot of these files appear to be moved files and I don't think Safia changed this one. The main modern browser feature used is `element.scrollIntoViewIfNeeded()`, which is feature checked before using as well as a fallback to `el.scrollIntoView()`

packages/notebook-app-component/src/decorators/themer/index.tsx
rgbkrk: `<DarkTheme />` is effectively a CSS Stylesheet with CSS variables: https://github.com/nteract/nteract/blob/aa9d92b33505732132394379dd2edfd908c8377c/packages/presentational-components/src/styles.tsx#L3

MSeal: Got it, thanks for catching me up some on react patterns

packages/notebook-app-component/src/decorators/hijack-scroll/index.tsx
MSeal: Fair, should we make an issue to test this across browsers and add any automated tests that can be easily added for later?

packages/notebook-app-component/src/decorators/cell-creator/index.tsx
captainsafia: Yes, this component represents the cell creator that resides within the notebook app. This cell creator is intedended to go on top of and in between cells.

Because of the refactor I did, the cell creator appears behind cells if it is at the top so this rule is to get around that. I'll add a comment for clarity.

packages/editor/src/index.tsx
```diff
@@ -56,14 +60,15 @@ export type CodeMirrorEditorProps = {
   // _Our_ theme, not the codemirror one we use
   theme: string;
   channels?: Channels | null;
-  // TODO: We only check if this is idle, so the completion provider should only
-  //       care about this when kernelStatus === idle _and_ we're the active cell
-  //       could instead call it `canTriggerCompletion` and reduce our current re-renders
```
rgbkrk, line 61: This should probably get turned into a different comment without a `TODO`, just stating it's current use.

packages/editor/src/index.tsx
```diff
@@ -481,3 +486,56 @@ export default class CodeMirrorEditor extends React.PureComponent<
...
+    };
+  };
+  return mapStateToProps;
+};
+
+const ConnectedCodeMirrorEditor = connect(makeMapStateToProps)(
+  CodeMirrorEditor
+);
+
+export default ConnectedCodeMirrorEditor;
```
rgbkrk: Should this also live in the stateful components instead of connecting it in the editor package?

packages/notebook-app-component/__tests__/cell-creator.spec.tsx
```diff
@@ -108,38 +108,13 @@ describe("CellCreatorProvider", () => {
...
-        resolve();
-      };
-      store.dispatch = dispatch;
-      const wrapper = setup(true, "test");
-      const createCell = wrapper
-        .findWhere(n => n.prop("createCell") !== undefined)
-        .prop("createCell");
-      createCell("code");
-    });
-  });
```
rgbkrk, line 135: Was this test just not working or was it getting in the way?

packages/notebook-app-component/src/decorators/cell-creator/index.tsx
rgbkrk: Wow do we need some design hierarchies here. 😅

packages/notebook-app-component/src/decorators/cell-creator/index.tsx
```diff
@@ -160,11 +168,33 @@ class CellCreator extends React.PureComponent<ConnectedProps> {
...
   render() {
     return (
-      <PureCellCreator above={this.props.above} createCell={this.createCell} />
+      <React.Fragment>
+        {this.props.isFirstCell && (
+          <PureCellCreator above={true} createCell={this.createCell} />
+        )}
+        {this.props.children}
+        <PureCellCreator above={false} createCell={this.createCell} />
+      </React.Fragment>
```
rgbkrk, line 179: This is much cleaner, I like it. It's easy to read this as:

```jsx
<CellCreator id={id} contentRef={cr}>
  <Cell ...>
<CellCreator />
```

packages/notebook-app-component/src/derived-components/status-bar.tsx
```diff
@@ -0,0 +1,64 @@
...
+                  {context.kernelStatus}
+                </p>
+              </LeftStatus>
+            </Bar>
+          )}
+        </StatusBarContext.Consumer>
+      </StatusBar>
+    );
+  }
+}
```
rgbkrk, line 64: Nifty.

packages/notebook-app-component/src/notebook-apps/base.tsx
```diff
@@ -0,0 +1,68 @@
...
+                cell_type="raw"
+              >
+                {{
+                  toolbar: () => (
+                    <CellToolbar id={props.id} contentRef={props.contentRef} />
+                  )
+                }}
+              </RawCell>
+            )
+          }}
```
rgbkrk, line 62: I feel silly saying this now, but are we going to run into perf issues when React sees this inline object as a new object every render?

packages/editor/src/index.tsx
```diff
@@ -56,14 +60,15 @@ export type CodeMirrorEditorProps = {
   // _Our_ theme, not the codemirror one we use
   theme: string;
   channels?: Channels | null;
-  // TODO: We only check if this is idle, so the completion provider should only
-  //       care about this when kernelStatus === idle _and_ we're the active cell
-  //       could instead call it `canTriggerCompletion` and reduce our current re-renders
   kernelStatus: string;
+  kernel?: any;
```
rgbkrk: This is just `kernel` info right? If this is a kernel object with `channels`, is our `channels` prop duplicative?

packages/stateful-components/src/outputs/transform-media.tsx
```diff
@@ -16,8 +16,8 @@ interface ComponentProps {
   output_type: string;
   id: string;
   contentRef: ContentRef;
-  index: number;
-  output: any;
+  index?: number;
+  output?: ImmutableDisplayData | ImmutableExecuteResult;
```
rgbkrk, line 20: 👍

packages/stateful-components/src/outputs/transform-media.tsx
```diff
@@ -150,7 +150,7 @@ const makeMapDispatchToProps = (
               id,
               contentRef,
               metadata,
-              index,
+              index: index || 0,
```
rgbkrk, line 153: 🤔

packages/stateful-components/src/notebook/status-bar.tsx
rgbkrk: I really like this pattern.

packages/stateful-components/src/inputs/input.tsx
```diff
@@ -33,7 +33,7 @@ const makeMapStateToProps = (
 
     let hidden = false;
 
-    if (model && model.type == "notebook") {
+    if (model && model.type === "notebook") {
```
rgbkrk: 😅

packages/stateful-components/src/cells/cells.tsx
```diff
@@ -23,52 +23,46 @@ interface StateProps {
...
+        />
+      ),
+      raw: (props: CellProps) => (
+        <RawCell id={props.id} contentRef={props.contentRef} cell_type="raw" />
+      )
+    };
+
+    const code = children?.code || defaults.code;
+    const markdown = children?.markdown || defaults.markdown;
+    const raw = children?.raw || defaults.raw;
```
rgbkrk, line 57: So clean with the `?.` operator!

packages/stateful-components/docs/styling.md
```diff
@@ -37,6 +37,15 @@ To style with stylesheets, you can use element and combinator-based CSS selector
...
+| Prompt       | `.nteract-cell-prompt`    |
+| Cells        | `.nteract-cells`          |                     |
+| Cell         | `.nteract-cell-container` | `.selected`         |
+| CodeCell     | `.nteract-code-cell`      |                     |
+| MarkdownCell | `.nteract-markdown-cell`  |                     |
+| CellToolbar  | `.nteract-cell-toolbar`   |                     |
+| Input        | `.nteract-cell-input`     |                     |
+| Pagers       | `.nteract-cell-pagers`    |                     |
+| InputPrompts | `.nteract-input-prompts`  |                     |
+| Outputs      | `.nteract-outputs`        | `.hidden .expanded` |
```
rgbkrk: What's our maintenance burden with class names being part of our "design API"?

rgbkrk (approved):

packages/editor/src/index.tsx
captainsafia: Good call. Although I still think this should be refactored at some point. Likely alongside https://github.com/nteract/nteract/issues/4741.

packages/editor/src/index.tsx
captainsafia: Yeah, this makes sense. I'll move it out.

packages/stateful-components/src/cells/cells.tsx
captainsafia: I know right! So glad that @CrystallineCat got us upgraded to the latest version of TypeScript.

packages/stateful-components/src/inputs/input.tsx
captainsafia: With strict equality for all!

packages/notebook-app-component/src/decorators/cell-creator/index.tsx
captainsafia: Indeed. I love this decorator pattern that @CrystallineCat has introduced.

packages/stateful-components/docs/styling.md
captainsafia: Great question!

I imagine that we'll be pretty static on the component-based class names (e.g., `nteract-cell-input`) but will probably evolve the state-based class names.

I think as long as we make sure that any styleable component state is exposed as a classname is fine.

I also imagine this will be part of a more formal theming API so we will have to take on the cost of keeping this classnames tidy but I don't imagine they'll change all too much.

We'll have a better sense of where this is at once the new redesign is underway.

packages/editor/src/index.tsx
captainsafia: Yep. This is just kernel info. I wanted to keep the `channels` prop for backwards compatability with other API users.

Although I realize that if I take your advice of moving the connected bits to stateful-components we won't have this issue.

packages/notebook-app-component/src/notebook-apps/base.tsx
captainsafia: I don't believe so. This pattern is using [functions as  children](https://codedaily.io/tutorials/6/Using-Functions-as-Children-and-Render-Props-in-React-Components) which is the same thing as render props so I don't think it plugs into the react lifecycle in the same way.

packages/notebook-app-component/src/derived-components/status-bar.tsx
captainsafia: Indeed! Loving all of these new patterns.

packages/notebook-app-component/__tests__/cell-creator.spec.tsx
captainsafia: A bit of both. I verified the functionality this tests manually. However I was having trouble making sure that the state was configured correctly here. Ideally, this would be rewritten to use the pure component as opposed to the connected one and validate functionality that way.

packages/notebook-app-component/src/decorators/hijack-scroll/index.tsx
captainsafia: I can create an issue for cross-browser compat.

rgbkrk:
Are there specific segments I should re-review or is this good to go?

rgbkrk:
Ok, I've looked at the new commits and I'm 👍

PR #4793, opened 2019-12-22
Fix wrong import in component docs

### LINKED ISSUE #4791: Components documentation import line is misleading
In the [nteract components documentation](https://components.nteract.io/#!/Plaintext) for e.g. the `Plain` component, the first line suggests that the component can be used as

```
import { Plain } from '@nteract/outputs';
```

However, this doesn't seem to actually work. Instead, the component needs to be used as `Media.Plain` (and is referenced as such elsewhere). It would be nice to either make the suggested import work or adjust the snippet to the correct usage.

CrystallineCat:
Replaces the import in the component docs with e. g. `@nteract/outputs/src/components/media/plain`. Fixes #4791.

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -13,20 +13,7 @@ module.exports = {
   defaultExample: false,
   propsParser: typescriptPropsParser,
   resolver: require("react-docgen").resolver.findAllComponentDefinitions,
-  getComponentPathLine: componentPath => {
-    const toPascalCase = string => {
-      return string
-        .match(/[a-z]+/gi)
-        .map(function(word) {
-          return word.charAt(0).toUpperCase() + word.substr(1).toLowerCase();
-        })
-        .join("");
-    };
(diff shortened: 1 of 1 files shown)


wadethestealth:
Replaces it with what?

CrystallineCat:
With the package name and path, for example `@nteract/outputs/src/components/media/plain`.

rgbkrk:
Seems reasonable on first glance.

captainsafia (commented):
Looks good overall. Can we bring back the `import X from` text before the component path?

For example, currently we display this:

```
import { Cell } from '@nteract/presentational-components';
```

And this PR changes it to display this:

```
@nteract/presentational-components/src/components/cell
```

Can we strike a balance between the two and do this?

```
import Cell from "@nteract/presentational-components/src/components/cell";
```

CrystallineCat:
@captainsafia Done.

PR #4794, opened 2019-12-23
Set content state on FetchContentFailed

### LINKED ISSUE #4785: ContentRecord Loading Status doesn't change on Fetch_Content_Failed action
I am implementing a scenario where we show a loading icon while the notebook content is loading, which relies on the loading status on the ContentRecord by contentRef.

After some investigation, I found the following: 
When FETCH_CONTENT_FAILED action occurs, the ContentRecord loading status remains true, and error is null. Additionally, the ContentRecord stays stuck as "dummy" type.

**I believe the correct logic would be for the loading status to be set to false when FETCH_CONTENT_FAILED** (and perhaps other failed actions too?), In addition, it seems that error would be set to something (though I may be misunderstanding the purpose of error) - this would allow our logic to take action on ContentRecord.error property instead of needing a workaround reducer.

Thanks

captainsafia:
Closes #4785.

### DIFF
--- a/packages/reducers/src/core/entities/contents/index.ts
+++ b/packages/reducers/src/core/entities/contents/index.ts
@@ -107,6 +107,14 @@ const byRef = (
           // TODO: we can set kernelRef when the content record uses it.
         })
       );
+    case actionTypes.FETCH_CONTENT_FAILED:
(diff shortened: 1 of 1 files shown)


rgbkrk (approved):
Thanks.

PR #4795, opened 2019-12-23
Some DRYing of @nteract/actions

CrystallineCat:
Progress on #4561.

Removes repeated parts of the action definitions. Also reorganizes the actions, and merges `actions` into `actionTypes` to keep related code closer together.

I turned off `tslint` long lines for the files, because I like the way this looks like a table of actions now. I think the ability to get an overview over all the actions in much better now.

There's still a lot of code that could be abstracted more, but the idea I have would require a change to the package's interface, so I'm going to make that separately.

I also removed all the deprecated actions.


CrystallineCat:
I'm having problems with the last build error. Even if I fix the `undefined`/`null` thing, `KernelspecProps` and `KernelspecMetadata` are incompatible.

There seem to be two entirely different type hierarchies for kernelspecs, in `@nteract/types/src/entities/kernelspecs.ts` and `@nteract/types/src/index.ts`. What is the purpose of that?

I could write a function to convert them, but I'm not sure if that's correct.

CrystallineCat:
Also, `tsc -b` locally is behaving weirdly, finding errors in unrelated code that `circleci` doesn't find, but not finding any of the real errors. :frowning_face: I already tried `yarn run build:clean`, but that doesn't help. Unsure what the issue is there.

Edit: After deleting all `lib` (there doesn't seem to be a command in the `package.json` for this?), everything works properly again.

captainsafia:
In this case, `KernelspecMetadata` is a type that refers to the kernelspec as it is stored within the notebook document when it is serialized.

`KernelspecProps` is the type that we store in the Redux state when we fetch the kernel specs.

For this scenario, we probably want to create a new type adjacent to `KernelspecInfo` that has the following schema.

```js
interface NamedKernelsepc {
    name: string;
    spec: KernelspecsRecord
};
```

CrystallineCat:
Thank you for the reply!

I don't think that would solve the problem, as `KernelspecProps` and `KernelspecMetadata` are not interchangeable, so I wouldn't just be able to use the new type as a drop-in replacement for `KernelspecInfo` in the action.

For example, the reducer accesses `display_name` of `KernelspecMetadata`, which is spelled `displayName` in `KernelspecProps`. In fact, it seems that up to now, the epic just gave the action the wrong type, and the reducer thus getting `undefined` values, which I think is a bug.

As I said, I could write a function to convert the `KernelspecProps` to `KernelspecMetadata`, if such a function doesn't exist yet, which might be the best approach.

captainsafia:
@CrystallineCat Just saw that you've submitted some updates to this. Is it ready for review?

CrystallineCat:
@captainsafia Thanks for asking, I had a problem with the merge, but fixed that now. It's now ready for review.

I noticed after the merge that there are some issues when testing nteract. Keyboard shortcuts for running cells don't work, the dev console shows some errors and updating previous outputs doesn't work. These issues all appear to be present in the master branch as well. Apart from that, everything appears to run fine.

captainsafia:
Keyboard shortcuts is fixed here: https://github.com/nteract/nteract/pull/4830

Can you open a bug for the issue around updating previous outputs so we can dig into that?

EDIT: To be clear, master is in a bit of a subpar state at the moment after the merge of #4788 but I am hoping to resolve any issues before the next production release at the end of the month.

rgbkrk:
Seems we can probably merge this after #4830, or should we consider fixing up `master` to be separate from bringing this in?

captainsafia:
I would consider fixing up master separate from this. I wanna follow-up on cleaning up master before this months production release as part of the follow on for the PR linked above.

CrystallineCat:
Please tell me once I can merge this (or if there are changes), I'd like to minimize rebases. :cat2:

captainsafia:
@CrystallineCat I think this is fine to merge after tomorrow's (Monday afternoon US time) release.

captainsafia:
OK, @CrystallineCat we are clear for a final rebase and merge on this! Thanks.

CrystallineCat:
Done. Thank you!

captainsafia:
Thanks so much for all your hard work on this, @CrystallineCat! A much needed and much-appreciated clean-up.

PR #4798, opened 2019-12-25
Update type used in setKernelspecInfo and fix bug in reducer

captainsafia:
Follow up to https://github.com/nteract/nteract/pull/4737.

- Updates the type passed to the `setKernelspecInfo` action to be a `KernelspecRecord` stored in the state
- Updates the tests to reflect this new type
- Removes the `legacy` comment about this type since we still need it
- Fixes bug where display_name was not set correctly in notebooks

### DIFF
--- a/packages/actions/src/actionTypes/kernelspecs.ts
+++ b/packages/actions/src/actionTypes/kernelspecs.ts
@@ -35,11 +35,11 @@ export interface FetchKernelspecsFailed {
   };
 }
 
-export const SET_KERNELSPEC_INFO = "SET_KERNELSPEC_INFO";
-export interface SetKernelspecInfo {
-  type: "SET_KERNELSPEC_INFO";
+export const SET_KERNEL_METADATA = "SET_KERNEL_METADATA";
+export interface SetKernelMetadata {
+  type: "SET_KERNEL_METADATA";
   payload: {
-    kernelInfo?: KernelspecRecord | null;
+    kernelInfo: KernelspecRecord;
     contentRef: ContentRef;
   };
 }
--- a/packages/actions/src/actionTypes/kernelspecs.ts
+++ b/packages/actions/src/actionTypes/kernelspecs.ts
@@ -1,8 +1,8 @@
 import {
   ContentRef,
   HostRef,
-  KernelspecInfo,
   KernelspecProps,
+  KernelspecRecord,
   KernelspecsRef
 } from "@nteract/types";
 
@@ -35,13 +35,11 @@ export interface FetchKernelspecsFailed {
   };
 }
 
-// "legacy" action that pushes kernelspec info back up
-// for the notebook document
 export const SET_KERNELSPEC_INFO = "SET_KERNELSPEC_INFO";
 export interface SetKernelspecInfo {
   type: "SET_KERNELSPEC_INFO";
   payload: {
-    kernelInfo: KernelspecInfo;
+    kernelInfo?: KernelspecRecord | null;
     contentRef: ContentRef;
   };
 }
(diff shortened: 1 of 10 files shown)


CrystallineCat (commented):
Thank you for this change. I hope it's ok for me to review this.

There's one other reference to `actions.setKernelspecInfo` that you need to update to the new parameters: https://github.com/nteract/nteract/blob/e1ecdb433ee0d7e0822206a6cefbeb1855677f3b/applications/desktop/src/notebook/epics/zeromq-kernels.ts#L116

I'll put other comments into the code.

packages/actions/src/actionTypes/kernelspecs.ts
```diff
@@ -35,13 +35,11 @@ export interface FetchKernelspecsFailed {
...
 }
 
-// "legacy" action that pushes kernelspec info back up
-// for the notebook document
 export const SET_KERNELSPEC_INFO = "SET_KERNELSPEC_INFO";
 export interface SetKernelspecInfo {
   type: "SET_KERNELSPEC_INFO";
   payload: {
-    kernelInfo: KernelspecInfo;
+    kernelInfo?: KernelspecRecord | null;
```
CrystallineCat: I'm not sure about the `undefined` and the `null` here. Either one will turn the action into a no-op, and I think it would be preferable to not emit the action in the first place in that case. That would allow us to have a stronger type here.

packages/actions/src/actionTypes/kernelspecs.ts
```diff
@@ -35,13 +35,11 @@ export interface FetchKernelspecsFailed {
   };
 }
 
-// "legacy" action that pushes kernelspec info back up
-// for the notebook document
 export const SET_KERNELSPEC_INFO = "SET_KERNELSPEC_INFO";
 export interface SetKernelspecInfo {
```
CrystallineCat: Nit: Since its payload is no longer a `KernelInfo`, would it be better to rename the action to reflect that?

packages/actions/src/actionTypes/kernelspecs.ts
captainsafia: Good call. I figure we should be more opinionated about requiring only valid properties to be set on actions that are emitted. I can update this.

packages/actions/src/actionTypes/kernelspecs.ts
captainsafia: Yeah. Perhaps, `SET_KERNEL_METADATA` is a better name for this action.

captainsafia:
@CrystallineCat Thanks for the review! Much appreciated. I've addressed your feedback. Let me know what you think.

packages/epics/src/kernel-lifecycle.ts
```diff
@@ -128,11 +127,13 @@ export function acquireKernelInfo(
             kernelRef,
             info
           }),
-          actions.setKernelspecInfo({
-            contentRef,
-            kernelInfo
-          })
-        ];
+          kernelspec
```
MSeal: This may be my internal python dev speaking, but I think it might read cleaner if the kernelspec is only added to result in another line if it's defined rather than adding undefined and then filtering it out. If the above actions can also be falsy then probably keep what's here.

MSeal (approved):
LGTM. Minor comment, up to you on if you want to change it.

packages/epics/src/kernel-lifecycle.ts
captainsafia: Yeah, I'll admit this is messy. This setup is purely because of some battles I was having with TypeScript when I was trying to get this working. I'll see if I can get TypeScript happy with the appropriate changes.

captainsafia: Fixed.

Issue #4768, opened 2019-12-13
Remove "Save Successful" toast on autosave?


dougdenn:
**Is your feature request related to a problem? Please describe.**
The "Save Successful" dialog pops up every time the notebook is auto-saved and it's really distracting, especially in dark mode where the toast has a white background.

**Describe the solution you'd like**
I think the toast makes sense on manual saves but not autosaves - for those I think the info in the bottom right (Last saved less than a minute) is good enough. Maybe it could say "Autosaved" to let new users know that autosave is implemented.

I saw the [issue here](https://github.com/nteract/nteract/issues/4567#issuecomment-561801711) that the auto save interval is going to be configurable, which is great, this might just be the last piece needed for that.

Would the team be interested in a PR for this?

![image](https://user-images.githubusercontent.com/30697943/70805128-d8954480-1d7d-11ea-9ef7-8e8f5c597e52.png)

captainsafia:
Hi @dougleville!

Thanks for opening this. Funny coincidence -- I was thinking about something similar just last night.

I agree that it should be removed and am partial to removing all notifications in general and relying on in-app cues to notify users of when important things have happened (like a kernel restarting or a save completed). This will be resolved with our upcoming redesign of nteract. Stay tuned for the interactive mockups of this so you can share feedback there.

In the meantime, would you be interested in removing the notification system components out of the desktop app?

dougdenn:
@captainsafia Sure thing 👍

PR #4802, opened 2019-12-28
Remove all notifications from desktop application

dougdenn:
- [✔] I have read the [Contributor Guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md)

This PR addresses #4768 - hopefully this is what you had in mind @captainsafia

### DIFF
--- a/applications/desktop/__tests__/renderer/menu.spec.ts
+++ b/applications/desktop/__tests__/renderer/menu.spec.ts
@@ -421,7 +421,8 @@ describe("dispatchInterruptKernel", () => {
         }
       });
     } else {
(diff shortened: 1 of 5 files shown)


captainsafia (changes requested):
Thanks for submitting this PR, @dougleville!

I realize in the issue I mentioned removing the notification system altogether. Considering it'll be a bit before the new UI is rolled out can we keep all uses of the notification system except the "Save Successful" one?

We already have save indicators in the status bar so I think it should be fine there.

captainsafia:
Closing this in favor of #4871.

Issue #4762, opened 2019-12-11
RFC: Lazy Kernel Launch


captainsafia:
**Context**

Over the past few years, we've tracked quite a few issues on nteract around improving the kernel launch process. This RFC is meant to capture all of these tracked issues and propose a definitive solution for all of them. Here are some of the issues that have been opened in the past:

* Caching of execution requests during slow kernel starts ( https://github.com/nteract/nteract/issues/1431)
* Buffer cell evaluations while kernel is starting ( https://github.com/nteract/nteract/issues/3368)

Also, @rgbkrk has taken at a stab at implementing buffering for execution requests in the past in this (now-closed) PR: .https://github.com/nteract/nteract/pull/2879.

**Problem**

nteract should support "lazy" kernel launches and improve the kernel launch API to be more robust to kernels that take a long time to launch. This is particular relevant for the following scenarios:

* Users who are using nteract with slow-to start kernels who want to be able to queue cells for execution while their kernel is still starting up.
* Developers who are building nteract-based applications and what to "delay" the launching of a kernel which might be hosted on price-y compute resources (like a GPU host).
* Users who want to open notebooks in view only mode without connecting to a kernel until needed -- similar to the pattern that Google's Colab notebook solution provides.

**Proposed Solution**

nteract should standardize the experience for launching kernels lazily and buffering execution messages as needed. I'd also like to take advantage of this RFC to do some other housekeeping work on the nteract kernel interaction to make it easier to do other things (like support swapping out the execution back-end used to other protocols). I think this work can be achieved by repurposing the `SendExecuteRequest` action to be the action that that execution epics will listen to when determining when to execute code. Currently, the `ExecuteCell` is directly mapped to the execution epics. Instead, I imagine the flow will look like this: 

`ExecuteCell`, `ExecuteFocusedCell`, `ExecuteCellsBelow`, and any other UI-specific execution requests will be mapped to a  `SendExecuteRequest` action.

The `SendExecuteRequest` action will be consumed by an epic that checks to see if a kernel is already available for the ContentRef the execution request is associated with. If so, it will send the action to the `executeCellStream`. If not, it will dispatch an action to launch the kernel and use RxJS's `bufferUntil` operator to buffer the remaining execute requests until the kernel has completely launched.

The interface for the `SendExecuteRequest` action should look something like this.

```javascript
{
	type: "SEND_EXECUTE_REQUEST",
	payload: {
		message: JupyterMessage;
		contentRef: ContentRef;
[...]

PR #4822, opened 2020-01-11
Lazy launch kernel

victor-meng:
- [✓] I have read the [Contributor Guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md)

The goal of this PR is to launch the kernel only when user attempts to execute a cell instead of immediately launching the kernel when user opens a notebook.

- `ExecuteFocusedCell`, `ExecuteAllCells`, and `ExecuteAllCellsBelow` are mapped to `ExecuteCell`.
- `ExecuteCell` now checks if the kernel exists:
  - If it doesn't, `launchKernelByName` and `enqueueAction` actions are emitted.
  - If it does, but the kernel status is not fully ready (e.g. the kernel status is `starting`), then `enqueueAction` is emitted.
  - If the kernel is fully ready, then `sendExecuteRequest` action is emitted which executes the cell like what `executeCell` used to do.
- When the kernel doesn't exist or isn't fully ready, the pending actions are stored in a message queue which is an Immutable.List of actions that represents a FIFO queue.
  - The message queue's path in the state object is: **state.core.entities.messages**
-  A new epic which listens to the `LAUNCH_KERNEL_SUCCESSFUL` action is added. It checks whether the kernel is ready and emits an `executeCell` action for each pending action in the message queue. The message queue is then cleared.

References:
 - RFC by @captainsafia :https://github.com/nteract/nteract/issues/4762
 - Action flowchart: image

### DIFF
--- a/packages/actions/src/actions/messages.ts
+++ b/packages/actions/src/actions/messages.ts
@@ -2,17 +2,17 @@ import { ContentRef } from "@nteract/types";
 import * as actionTypes from "../actionTypes";
 
 export function enqueueAction(payload: {
-    id: string;
-    contentRef: ContentRef;
+  id: string;
+  contentRef: ContentRef;
 }): actionTypes.EnqueueAction {
-    return {
-      type: actionTypes.ENQUEUE_ACTION,
-      payload
-    };
+  return {
+    type: actionTypes.ENQUEUE_ACTION,
+    payload
+  };
 }
   
 export function clearMessageQueue(): actionTypes.ClearMessageQueue {
-    return {
-      type: actionTypes.CLEAR_MESSAGE_QUEUE
-    };
+  return {
+    type: actionTypes.CLEAR_MESSAGE_QUEUE
+  };
 }
--- a/packages/actions/src/actions/messages.ts
+++ b/packages/actions/src/actions/messages.ts
@@ -1,0 +1,18 @@
+import { ContentRef } from "@nteract/types";
+import * as actionTypes from "../actionTypes";
+
+export function enqueueAction(payload: {
+    id: string;
+    contentRef: ContentRef;
+}): actionTypes.EnqueueAction {
+    return {
+      type: actionTypes.ENQUEUE_ACTION,
+      payload
+    };
+}
+  
+export function clearMessageQueue(): actionTypes.ClearMessageQueue {
+    return {
+      type: actionTypes.CLEAR_MESSAGE_QUEUE
+    };
+}
--- a/packages/epics/src/execute.ts
+++ b/packages/epics/src/execute.ts
@@ -314,13 +314,14 @@ export function lazyLaunchKernelEpic(
         content.type !== "notebook" ||
         content.model.type !== "notebook"
       ) {
-        return actions.launchKernelFailed({
-          contentRef,
-          error: new Error(
-            "Launch kernel failed because the source content is not a notebook"
-          ),
-          kernelRef
-        });
+        return of(
+          actions.launchKernelFailed({
+            error: new Error(
+              "Launch kernel failed because the source content is not a notebook"
+            ),
+            contentRef
+          })
+        );
       }
 
       const filepath = content.filepath;
--- a/packages/epics/src/execute.ts
+++ b/packages/epics/src/execute.ts
@@ -13,6 +13,7 @@ import {
   outputs,
   payloads
 } from "@nteract/messaging";
+import { AnyAction } from "redux";
 import { ofType } from "redux-observable";
 import { ActionsObservable, StateObservable } from "redux-observable";
 import { empty, merge, Observable, Observer, of, throwError } from "rxjs";
@@ -28,8 +29,10 @@ import {
   share,
   switchMap,
   takeUntil,
-  tap
+  tap,
+  withLatestFrom
 } from "rxjs/operators";
+import { extractNewKernel } from "./kernel-lifecycle";
 
 import * as actions from "@nteract/actions";
 import { CellId, OnDiskOutput } from "@nteract/commutable";
@@ -40,8 +43,7 @@ import {
   InputRequestMessage,
   PayloadMessage
 } from "@nteract/types";
-
-const Immutable = require("immutable");
+import { List } from "immutable";
 
 /**
  * Observe all the reactions to running code for cell with id.
@@ -134,15 +136,7 @@ export function executeCellStream(
 }
 
 export function createExecuteCellStream(
-  action$: ActionsObservable<
-    | actions.ExecuteCanceled
-    | actions.DeleteCell
-    | actions.LaunchKernelAction
-    | actions.LaunchKernelByNameAction
-    | actions.KillKernelAction
-    | actions.ExecuteCell
-    | actions.ExecuteFocusedCell
-  >,
+  action$: ActionsObservable<actions.SendExecuteRequest>,
   state: any,
   message: ExecuteRequest,
   id: string,
@@ -168,51 +162,10 @@ export function createExecuteCellStream(
     );
   }
 
-  const cellStream = executeCellStream(channels, id, message, contentRef).pipe(
-    takeUntil(
-      merge(
-        action$.pipe(
-          ofType(actions.EXECUTE_CANCELED, actions.DELETE_CELL),
-          filter(
-            (
-              action:
-                | actions.ExecuteCanceled
-                | actions.DeleteCell
-                | actions.LaunchKernelAction
-                | actions.LaunchKernelByNameAction
-                | actions.KillKernelAction
-                | actions.ExecuteCell
-                | actions.ExecuteFocusedCell
-            ) => (action as actions.ExecuteCanceled).payload.id === id
-          )
-        ),
-        action$.pipe(
-          ofType(
-            actions.LAUNCH_KERNEL,
-            actions.LAUNCH_KERNEL_BY_NAME,
-            actions.KILL_KERNEL
-          ),
-          filter(
-            (
-              action:
-                | actions.ExecuteCanceled
-                | actions.DeleteCell
-                | actions.LaunchKernelAction
-                | actions.LaunchKernelByNameAction
-                | actions.KillKernelAction
-                | actions.ExecuteCell
-                | actions.ExecuteFocusedCell
-            ) => action.payload.contentRef === contentRef
-          )
-        )
-      )
-    )
-  );
+  const cellStream = executeCellStream(channels, id, message, contentRef);
 
   return merge(
-    // We make sure to propagate back to "ourselves" the actual message
-    // that we sent to the kernel with the sendExecuteRequest action
-    of(actions.sendExecuteRequest({ id, message, contentRef })),
+    of(actions.clearOutputs({ id, contentRef })),
     // Merging it in with the actual stream
     cellStream
   );
@@ -237,7 +190,7 @@ export function executeAllCellsEpic(
           return empty();
         }
 
-        let codeCellIds = Immutable.List();
+        let codeCellIds = List();
 
         if (action.type === actions.EXECUTE_ALL_CELLS) {
           codeCellIds = selectors.notebook.codeCellIds(model);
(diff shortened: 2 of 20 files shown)


packages/epics/src/execute.ts
```diff
@@ -168,51 +162,10 @@ export function createExecuteCellStream(
     );
   }
 
-  const cellStream = executeCellStream(channels, id, message, contentRef).pipe(
```
captainsafia: Was this `takeUntil` moved elsewhere? We need it to ensure that the post-processing of cell execution stops when one of the outline events happens.

packages/epics/src/execute.ts
```diff
@@ -168,51 +162,10 @@ export function createExecuteCellStream(
...
-      )
-    )
-  );
+  const cellStream = executeCellStream(channels, id, message, contentRef);
 
   return merge(
-    // We make sure to propagate back to "ourselves" the actual message
-    // that we sent to the kernel with the sendExecuteRequest action
-    of(actions.sendExecuteRequest({ id, message, contentRef })),
+    of(actions.clearOutputs({ id, contentRef })),
```
captainsafia, line 216: This should also set the cell status to queued as the `setIn(["transient", "cellMap", id, "status"], "queued")` portion of the now-deleted logic in the reducer was doing.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -634,32 +634,6 @@ function acceptPayloadMessage(
   return state;
 }
 
-function sendExecuteRequest(
```
captainsafia, line 637: So glad this has been removed and we are now handling these follow-on actions properly in the epics. Thanks for cleaning this up!

packages/reducers/src/core/entities/messages.ts
```diff
@@ -0,0 +1,26 @@
+import * as actions from "@nteract/actions";
```
captainsafia, line 1: Nit: Same as the other new file added above. The formatting on this will need to be fixed.

packages/epics/src/websocket-kernel.ts
```diff
@@ -86,7 +86,8 @@ export const launchWebSocketKernelEpic = (
               sessionId
             ),
             kernelSpecName,
-            hostRef
+            hostRef,
+            status: session.kernel.execution_state
```
captainsafia, line 90: We'll need to update the `launchWebSocketKernelEpic` unit test in `websocket-kernel.spec.ts` to account for this.

packages/actions/src/actions/messages.ts
```diff
@@ -0,0 +1,18 @@
+import { ContentRef } from "@nteract/types";
```
captainsafia, line 1: Nit: It appears this file is using 4-spaces for indentation. The formatter should have configured it to 2-spaces. You can run prettier (prettier --write ${pathToThisFile}) to reformat it according to our formatting configuration.

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+      if (kernel.channels && (kernel.status === "idle" || kernel.status === "busy")) {
+        return of(actions.sendExecuteRequest(action.payload));
+      }
+
+      return of(actions.enqueueAction(action.payload));
+    })
+  )
+}
+
+export function executeCellAfterKernelLaunchEpic(
```
captainsafia, line 375: We'll need to add a test for this. The `updateDisplayEpic` test in `execute.spec.ts` should be good inspiration for this.

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+  state$: StateObservable<AppState>
+) {
+  return action$.pipe(
+    ofType(actions.EXECUTE_CELL),
     tap((action: actions.ExecuteCell) => {
       if (!action.payload.id) {
         throw new Error("execute cell needs an id");
       }
     }),
+    withLatestFrom(state$),
```
captainsafia, line 300: We'll have to update the unit tests for the `executeCellEpic` in `execute.spec.ts` to account for the changes in functionality.

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+ */
+export function sendExecuteRequestEpic(
+  action$: ActionsObservable<actions.SendExecuteRequest>,
+  state$: any
+) {
+  return action$.pipe(
+    ofType(actions.SEND_EXECUTE_REQUEST),
     // Split stream by cell IDs
-    groupBy((action: actions.ExecuteCell) => action.payload.id),
+    groupBy((action: actions.SendExecuteRequest) => action.payload.id),
```
captainsafia, line 422: Maybe I'm a little lost in the code but I am not seeing where ExecuteCell actions are mapped to SendExecuteRequest actions. Can you point this out?

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+          content.model.type !== "notebook"
+        ) {
+          return empty();
+        }
+
+        const filepath = content.filepath;
+        const notebook = content.model.notebook;
+        const { cwd, kernelSpecName } = extractNewKernel(filepath, notebook);
+
+        return of(
```
captainsafia, line 317: To avoid emitting duplicate `LaunchKernelByName` actions from this epic, let's try adding a [distinct](https://www.learnrxjs.io/operators/filtering/distinct.html) operator after the mergeMap operator in this epic.

captainsafia (commented):
Thanks for submitting this PR, @victor-meng! I've left some comments and questions inline.

Also, I really appreciate the diagram you put together. Super helpful for others reading this code in the future.

packages/epics/src/execute.ts
victor-meng: It's in line 285. SendExecuteRequest is emitted if the kernel exists and is ready to execute code.

packages/epics/src/execute.ts
captainsafia, line 317: Can we add an explicit if-statement around this for the scenarios where we do want to launch a new kernel? I think it'll make the code easier to read. For example,

```js
if (kernel.channels && kernel.status === "idle" || kernel.status === "busy") {
  // sendExecuteRequest
} else if (kernel && kernel.status == "starting") {
  // enqueue execution
} else {
  // launch and queue execution
}
```

captainsafia: Victor and I chatted off-GitHub and tried a couple of different ways of using the `distinct` operator but weren't able to get the desired affect. I suspect that this has something to do with the way our operators and actions are returned.

We tried a couple of things including:
- Having the `distinct` operator in its own pipe
- Adding a filter to the `distinct` operator in case it checked for shallow equality by default

Given this, I think there are a couple of things we can try to resolve the issue of too many LaunchKernelByName actions emitted when an ExecuteAllCells action is emitted.

- Splitting out the code that does the launching and the code that does the queuing into separate subscriptions and use [the first operator](https://www.learnrxjs.io/operators/filtering/first.html) to only run the launch flow the first time an ExecuteCell action happens
- Instead of placing the logic into separate subscriptions with `pipe`, split it up into two separate epics with one using the first operator

captainsafia (commented):
Left more comments in.

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+    withLatestFrom(state$),
+    filter(([action, state]) => {
+      if (selectors.messageQueue(state).size === 0) {
+        return false;
+      }
+
+      const contentRef = action.payload.contentRef;
+      const kernel = selectors.kernelByContentRef(state, { contentRef });
+      return !!(kernel && kernel.channels &&
+        (kernel.status === "idle" || kernel.status === "busy"));
```
languy, line 390: Similar expression is used line 283: consider making this a function.

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+    }),
+    concatMap(([, state]) => {
```
languy, line 392: nit: do you need the `,`?

packages/epics/src/execute.ts
```diff
@@ -282,22 +230,115 @@ export function executeCellEpic(
...
+}
+
+
+/**
+ * the execute cell epic processes execute requests for all cells, creating
+ * inner observable streams of the running execution responses
+ */
+export function sendExecuteRequestEpic(
+  action$: ActionsObservable<actions.SendExecuteRequest>,
+  state$: any
```
languy: Can we put the proper type here?

languy (approved):

packages/actions/src/actions/messages.ts
victor-meng: Done

packages/epics/src/execute.ts
victor-meng: Thanks for the explanation offline! I've added it back in my latest commit.

packages/epics/src/execute.ts
victor-meng: Done

packages/epics/src/execute.ts
victor-meng: I've splitted this logic into two epics:
- lazyLaunchKernelEpic emits the launchKernelByName action
- executeCellEpic emits the EnqueueAction action

The lazyLaunchKernelEpic uses the `first` operator so the launchKernelByName action is emitted at most once.

victor-meng: Since I moved the launch kernel logic into another epic, there should only be two cases: sendExecuteRequest and enqueueAction.

packages/epics/src/execute.ts
victor-meng: Added tests for both executeCellAfterKernelLaunch and lazyLaunchKernel epics.

packages/epics/src/execute.ts
victor-meng: All functions in this file are epics. I think it might be better to preserve this pattern.

packages/epics/src/execute.ts
victor-meng: Yes, the first element in the array is `action` but it's unused in this function.

packages/epics/src/execute.ts
victor-meng: Done

packages/epics/src/websocket-kernel.ts
victor-meng: Done.

packages/reducers/src/core/entities/contents/notebook.ts
victor-meng: No problem!

packages/epics/src/execute.ts
```diff
@@ -40,8 +44,7 @@ import {
   InputRequestMessage,
   PayloadMessage
 } from "@nteract/types";
-
-const Immutable = require("immutable");
+import { List } from "immutable";
```
captainsafia, line 47: Thanks for unifying us on ESM imports here!

packages/epics/src/execute.ts
```diff
@@ -282,22 +282,128 @@ export function executeCellEpic(
...
+      const kernelRef = selectors.kernelRefByContentRef(state, {
+        contentRef
+      });
+
+      if (
+        !kernelRef || !content ||
+        content.type !== "notebook" ||
+        content.model.type !== "notebook"
+      ) {
+        return empty();
```
captainsafia: Nit. Can we return a [LaunchKernelFailedAction](https://packages.nteract.io/interfaces/actions.launchkernelfailed.html) here with an error message saying the source content is not a notebook?

packages/epics/src/index.ts
```diff
@@ -58,9 +65,13 @@ const allEpics = [
 
 export {
   allEpics,
+  executeCellAfterKernelLaunchEpic,
```
captainsafia, line 68: Note to self: We'll need to document these new epics in the changelog for the next release.

packages/epics/src/execute.ts
```diff
@@ -282,22 +282,128 @@ export function executeCellEpic(
         return of(
           actions.executeCell({ id, contentRef: action.payload.contentRef })
         );
+    })
+  );
+}
+
+export function lazyLaunchKernelEpic(
```
captainsafia, line 294: Quick favor: Can you do a pass and add docstrings to all the *new* epics here?

I'm planning on passing through and adding doc strings to all the existing epics sometime later this month but figured we should start being proactive about doc strings on new ones.

captainsafia (approved):
Looks great, Victor! I've left a few last comments on the PR. Once those addressed this is good to merge. We'll shake any further issues with this as they come up.

Thanks for steering this through design, implementation, and testing!

PR #4825, opened 2020-01-13
Add support for auto-complete on Tab

### LINKED ISSUE #4817: Update CodeMirror interactions on Tab
Recently had a conversation with @julianchu1 about the various <kbd>Tab</kbd> related issues in the CodeMirror editor.

Currently, <kbd>Tab</kbd> is mapped to indent the current line or highlighted section.

We need to update the logic in the [executeTab](https://github.com/nteract/nteract/blob/49bb0a5cd4a5c0c756b73cc9e88529ed02b0a1ee/packages/editor/src/index.tsx#L436) method so that it:
- Indents if the cursor is at the beginning of the line or if content is highlighted
- Executes an autocomplete if the cursor is at the end of the line

Also, we need to add a tab-trap escape key. This [CodeMirror issue](https://github.com/codemirror/CodeMirror/issues/5444) has more information but long story short we need to add a keypress that will dispatch an action to unfocus the editor and ensure that tab indexes are applied correctly on notebook cells.

captainsafia:
Closes #4817 .

This allows Tab to be used for auto-complete when it is not at the beginning of the line for compatibility with Jupyter classic and other UIs.

### DIFF
--- a/packages/editor/src/index.tsx
+++ b/packages/editor/src/index.tsx
@@ -434,9 +434,14 @@ export default class CodeMirrorEditor extends React.PureComponent<
   }
 
   executeTab(editor: Editor & Doc): void {
-    editor.somethingSelected()
-      ? editor.execCommand("indentMore")
-      : editor.execCommand("insertSoftTab");
+    const { line, ch } = editor.getCursor();
+    if (editor.somethingSelected()) {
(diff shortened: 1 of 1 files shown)


rgbkrk:
Very trim handling here!

PR #4828, opened 2020-01-14
Abstract out rx-jupyter contents usage with IContentProvider

gitanuj:
The idea here is to replace usage of rx-jupyter contents API with an interface which can be passed as a dependency when creating the epic middleware. That way an Nteract application can plug in any kind of content provider (like GitHub etc). To start the discussion I've created the interface which mimics what Jupyter Contents API does. We need to discuss what level of abstraction is acceptable in this change because abstracting the responses of IContentProvider might be a big refactor in itself.

1. Added a class JupyterContentProvider which gives an inbox implementation for jupter contents API.
2. Added LocalContentProvider for desktop app which reads/writes from local storage.
3. Updated epics to use IContentProvider passed as a dependency instead.


packages/epics/src/contents.ts
```diff
@@ -125,15 +121,9 @@ export function fetchContentEpic(
 
       const state: any = state$.value;
       const host: any = selectors.currentHost(state);
-
```
captainsafia, line 136: Can you clarify why the host type check was removed?

packages/epics/src/contents.ts
```diff
@@ -38,7 +38,8 @@ import { existsSync } from "fs";
 
 export function updateContentEpic(
   action$: ActionsObservable<actions.ChangeContentName>,
-  state$: StateObservable<AppState>
```
captainsafia, line 42: The tests for this file are under `packages/epics/__tests__/contents.spec.ts`

packages/rx-jupyter/src/contents.ts
```diff
@@ -115,10 +171,10 @@ export function get(
  *
  * @returns An Observable with the request response
  */
-export function update<FT extends FileType>(
```
captainsafia, line 118: Let's continue to export these individual functions out for people who are using rx-jupyter and importing individual methods.

captainsafia (commented):
Great start! Looking forward to seeing how this fits in.

Something helpful that we can do to fully evaluate this content provider model is to move the content-related methods for the desktop app (which operate by writing/reading directly from disk instead of a Jupyter server) to this model and using one set of epics for both the desktop app and web-based apps.

packages/epics/src/contents.ts
```diff
@@ -329,17 +320,8 @@ export function saveContentEpic(
...
-      if (host.type !== "jupyter") {
-        return of(
-          actions.saveFailed({
-            error: new Error("Cannot save content if no host is set."),
-            contentRef: action.payload.contentRef
-          })
-        );
-      }
+      const state: any = state$.value;
+      const host: any = selectors.currentHost(state);
```
languy: nit: you probably don't need the `: any` type declaration as the compiler can infer them from the assignment and `const` won't be reassigned later.

packages/epics/src/contents.ts
```diff
@@ -478,14 +460,15 @@ export function saveContentEpic(
...
+  state$: StateObservable<AppState>,
+  dependencies: { contentProvider: contents.IContentProvider }
 ): Observable<actions.SaveAsFailed | actions.SaveAsFulfilled> {
   return action$.pipe(
     ofType(actions.SAVE_AS),
     mergeMap((action: actions.SaveAs):
       | Observable<actions.SaveAsFailed | actions.SaveAsFulfilled>
       | Observable<never> => {
-      const state = state$.value;
+      const state: any = state$.value;
```
languy: nit: `: any` probably not needed.

languy:
Thanks @gitanuj for doing this!

A couple of thoughts on the content api:
* is `path: string` appropriate for what we need? jupyter only knows about filesystem paths, but we probably will use uri's.
* Can (should?) we handle multiple content providers? For example, in our current resource tree, we have files on github next to the files on the notebook storage server (ie jupyter server right now).
* Does it enable us to implement the proper authentication specific to the content provider? (I think it does, but just want to make sure we have all fields we need in the api or `ServerConfig`).

packages/epics/src/contents.ts
gitanuj: Oops! That should be put back. Must have been left from when I was prototyping.

packages/epics/src/contents.ts
gitanuj: I will update those tests in my next update.

packages/rx-jupyter/src/contents.ts
gitanuj: Not exporting them helps me catch all the usage so that I can go over them and update to JupyterContentProvider wherever required. I will revert this once this change is near completion.

gitanuj:
> Thanks @gitanuj for doing this!
> 
> A couple of thoughts on the content api:
> 
> * is `path: string` appropriate for what we need? jupyter only knows about filesystem paths, but we probably will use uri's.
> * Can (should?) we handle multiple content providers? For example, in our current resource tree, we have files on github next to the files on the notebook storage server (ie jupyter server right now).
> * Does it enable us to implement the proper authentication specific to the content provider? (I think it does, but just want to make sure we have all fields we need in the api or `ServerConfig`).

- I'm currently using `path: string` as uri only. (like in our data explorer I'm using `github://` to denote the file exists on github. Do you think we should use a more stricter type to enforce everything is a uri?
- I think it'll be simpler to support only a single `ContentProvider` in Nteract. The way I'm handling multiple content providers in our code is by composing multiple providers into one provider (based on the uri scheme of path). But all of that is implementation detail. Is there any scenario you think supporting multiple content providers will be useful?
- Since this change gives us flexibility to plug in our own content provider, one can always build in auth logic in the content provider itself (like in our `GitHubContentProvider` we know how to authenticate with GitHub). One problem I faced was passing information about github account name and repo name along with the path to the content provider. I ended up creating my own uri scheme like `github://<account>:<repo>:<path>`. May be we can solve this in a better way?

gitanuj:
> Great start! Looking forward to seeing how this fits in.
> 
> Something helpful that we can do to fully evaluate this content provider model is to move the content-related methods for the desktop app (which operate by writing/reading directly from disk instead of a Jupyter server) to this model and using one set of epics for both the desktop app and web-based apps.

That's definitely a good idea to prove this model works. Expect an update with that soon.

languy:
> * I'm currently using `path: string` as uri only. (like in our data explorer I'm using `github://` to denote the file exists on github. Do you think we should use a more stricter type to enforce everything is a uri?

`string` type is fine. Ideally, we would rename `path` to `uri` since reusing that field/arg makes sense, but this would probably mean changing all the current code to pass jupyter file paths as `file://` instead of just the path to be correct. It might be too much of a change, so we can keep it as is for now I think.

> * I think it'll be simpler to support only a single `ContentProvider` in Nteract. The way I'm handling multiple content providers in our code is by composing multiple providers into one provider (based on the uri scheme of path). But all of that is implementation detail. Is there any scenario you think supporting multiple content providers will be useful?

Agreed, let's keep it simple (and single) if you have something that works for our use case.

> * Since this change gives us flexibility to plug in our own content provider, one can always build in auth logic in the content provider itself (like in our `GitHubContentProvider` we know how to authenticate with GitHub). One problem I faced was passing information about github account name and repo name along with the path to the content provider. I ended up creating my own uri scheme like `github://<account>:<repo>:<path>`. May be we can solve this in a better way?

There were two things that came to mind: 1) how to store credentials + address. Your scheme looks fine. My only minor comment is to consider reusing github's `https` or `ssh` connection string directly in the uri (the one used to clone the repo with git) if it can save you from writing transform/parsing code or reinventing a new scheme. 2) passing the creds/token/whatever when doing the request (as in injecting token in the header, etc). This part is in the implementation of the provider so we should be fine, too as the API doesn't restrict or dictate how the call will be made.

packages/epics/src/contents.ts
gitanuj: Actually I was just adding a `LocalContentProvider` for desktop app and found out that the app doesn't set host. So in order to use these epics for the desktop app we'll need to accept any kind of host. Is there any specific reason we make sure that it's a `jupyter` host in these epics?

packages/epics/src/contents.ts
```diff
@@ -378,38 +380,42 @@ export function saveContentEpic(
...
-              if (Math.abs(diffDate) > 600) {
-                return of(
-                  actions.saveFailed({
-                    error: new Error("open in another tab possibly..."),
-                    contentRef: action.payload.contentRef
-                  })
-                );
+              // Ignore if the file wasn't found. Since this is a save epic
+              // we don't expect file to be present yet.
+              if (xhr.status != 404) {
```
gitanuj, line 377: With this change we're using the same set of epics for desktop app too. In the app when we first save a notebook we expect to get 404. It made sense to ignore 404 here since this is a "save" epic (and the file isn't expected to be present at the first place).

gitanuj:
> > * I'm currently using `path: string` as uri only. (like in our data explorer I'm using `github://` to denote the file exists on github. Do you think we should use a more stricter type to enforce everything is a uri?
> 
> `string` type is fine. Ideally, we would rename `path` to `uri` since reusing that field/arg makes sense, but this would probably mean changing all the current code to pass jupyter file paths as `file://` instead of just the path to be correct. It might be too much of a change, so we can keep it as is for now I think.
> 
> > * I think it'll be simpler to support only a single `ContentProvider` in Nteract. The way I'm handling multiple content providers in our code is by composing multiple providers into one provider (based on the uri scheme of path). But all of that is implementation detail. Is there any scenario you think supporting multiple content providers will be useful?
> 
> Agreed, let's keep it simple (and single) if you have something that works for our use case.
> 
> > * Since this change gives us flexibility to plug in our own content provider, one can always build in auth logic in the content provider itself (like in our `GitHubContentProvider` we know how to authenticate with GitHub). One problem I faced was passing information about github account name and repo name along with the path to the content provider. I ended up creating my own uri scheme like `github://<account>:<repo>:<path>`. May be we can solve this in a better way?
> 
> There were two things that came to mind: 1) how to store credentials + address. Your scheme looks fine. My only minor comment is to consider reusing github's `https` or `ssh` connection string directly in the uri (the one used to clone the repo with git) if it can save you from writing transform/parsing code or reinventing a new scheme. 2) passing the creds/token/whatever when doing the request (as in injecting token in the header, etc). This part is in the implementation of the provider so we should be fine, too as the API doesn't restrict or dictate how the call will be made.

Definitely a good idea to use an existing path scheme for github (if there's one). I'll investigate and update here.

applications/desktop/src/notebook/local-content-provider.ts
```diff
@@ -0,0 +1,154 @@
...
+import * as path from "path";
+import { readFileObservable, statObservable, writeFileObservable } from "fs-observable";
+import { Observable, of } from "rxjs";
+import { AjaxResponse } from "rxjs/ajax";
+import { map, mergeMap, catchError } from "rxjs/operators";
+import { stringifyNotebook, Notebook } from "@nteract/commutable";
+import { Notification } from "react-notification-system";
+
+// A Content provider which reads/writes to local disk
+export class LocalContentProvider implements contents.IContentProvider {
```
captainsafia: I think it makes sense to move the IContentProvider interface to the `@nteract/types` package instead of having it inside `rx-jupyter` since it is agnostic to whether the user is using file system, Jupyter Server, GithHub, etc for providing contents.

packages/rx-jupyter/src/contents.ts
```diff
@@ -55,6 +55,62 @@ export interface IContent<FT extends FileType = FileType>
     : null;
 }
 
+/**
+ * Standardized interface to perform CRUD operations on a remote content.
+ */
+export interface IContentProvider {
```
captainsafia: Per the comment above, let's move this interface to the `@nteract/types` package in a new file.

Let's also move any dependencies it has like the `FileType` and `IContent` interfaces.

packages/epics/src/hosts.ts
```diff
@@ -49,9 +50,9 @@ export function publishToBookstore(
         }) as any;
       }
 
-      // Dismiss any usage that isn't targeting a jupyter server
-      if (host.type !== "jupyter") {
-        return empty();
+      // Dismiss any usage that isn't targeting a jupyter server
```
captainsafia: We'll probably want to keep the host check in the `publishToBookstore` epic.

For context, bookstore is another [nteract project](https://github.com/nteract/bookstore) that extends the Jupyter server API with some new endpoints for notebook versioning. We will need to be sure we are running against a Jupyter server in those scenarios.

packages/epics/src/contents.ts
captainsafia: I think this will be fine.

content.save in the JupyterContentProvider is mapped to a PUT request to the Jupyter contents API which does not check if a file exists and does a save with a name.

packages/epics/src/contents.ts
captainsafia: Yes, these epics were designed with the assumption that they would only be used in the scenario where the user is connected to the Jupyter server.

When the user is on the desktop app, the host type is set to `local`.

As you alluded to in your PR description, we will want to remove these host checks across all the content epics and ensure that they can operate across the different providers.

rgbkrk (commented):
This looks like a great abstraction for the contents, I look forward to checking this out more.

applications/desktop/src/notebook/local-content-provider.ts
```diff
@@ -0,0 +1,153 @@
...
+  // Needed to get to notificationSystem
+  private state?: AppState;
+
+  constructor() {}
+
+  setAppState(state: AppState) {
+    this.state = state;
+  }
+
+  remove(serverConfig: ServerConfig, path: string): Observable<AjaxResponse> {
```
languy: nit: declare these as `public`

languy (approved):

captainsafia:
Alright, @gitanuj! Finally taking this for a spin. Been busy this week with this month's production release of nteract and triaging follow-on user issues. I'll let you know how it goes.

Also, looks like coverage dropped with this PR. Have you had the chance to write tests for this? Now that our test coverage is in a good state, I'd like to keep it consistent by making sure that PRs are well tested before going in. Thanks in advance!

captainsafia:
Alright. Here is my validation matrix for things that I did.

| App | Scenario | Works? |
|---|---|---|
| Desktop | Open | ✅ |
| Desktop | Save | ✅ |
| Desktop | Save As | ✅ |
| Desktop | Reload | ✅ |
| Desktop | New | ✅ |
| Desktop | Save New | ✅ |
| Web | Open | ✅ |
| Web | Save | ✅ |
| Web | Download | ✅ |
| Web | Reload | ✅ |
| Web | New | ✅ |
| Web | Save New | ✅ |

Looking good from a QA perspective. Will do one last code review then we should be good to go after some tests.

applications/desktop/src/notebook/epics/index.ts
```diff
@@ -46,6 +45,9 @@ const epics = [
   coreEpics.executeAllCellsEpic,
   coreEpics.executeFocusedCellEpic,
   coreEpics.updateContentEpic,
+  coreEpics.saveContentEpic,
```
captainsafia, line 48: So satisfying that we are now on one set of epics for both the desktop and web app experiences! Will make it easier to ship improvements to both equally.

packages/types/src/index.ts
```diff
@@ -10,6 +10,7 @@ import {
 import { HostRecord } from "./entities/hosts";
 import { KernelRef, KernelspecsRef } from "./refs";
 
+export * from "./content-provider";
```
captainsafia, line 13: Great! This is a good place for this to go.

captainsafia (approved):
Looks good overall!

Some of the removed tests will need to be restructured under the new pattern.

Once this is merged, I'll update the documentation I created in https://github.com/nteract/nteract/pull/4887 to explain the IContentProvider model and showcase how folks can create their own content providers.

gitanuj:
> Alright, @gitanuj! Finally taking this for a spin. Been busy this week with this month's production release of nteract and triaging follow-on user issues. I'll let you know how it goes.
> 
> Also, looks like coverage dropped with this PR. Have you had the chance to write tests for this? Now that our test coverage is in a good state, I'd like to keep it consistent by making sure that PRs are well tested before going in. Thanks in advance!

Thanks Safia for taking time to test this change. I'll get some time tomorrow to write tests. Will send another update soon.

gitanuj:
@captainsafia I just added some new tests for `LocalContentProvider` and also updated `rx-jupyter` tests to maintain the desired coverage. The change is ready to merge! 🎉

captainsafia:
Splendid! Merging now. Let's get this show on the road!

PR #4830, opened 2020-01-15
Add back support for Ctrl + Enter keyboard shortcut

captainsafia:
Adds back support for Ctrl+Enter keyboard shortcut for cell execution as a component decorator.

And also some tests. 🎉 🌮


captainsafia:
Hm. This is failing in CI but not on my local machine. I'll have to investigate this...

CrystallineCat (commented):
I would replace the references to the deprecated `e.keyCode` and `13` with `e.key` and `"Enter"`, otherwise this looks good.

rgbkrk (approved):
Approved, conditional on addressing @CrystallineCat's feedback about `e.key` instead of `e.keyCode`.

PR #4832, opened 2020-01-15
Add tests and clean up host epics

captainsafia:
Ditto the title.


packages/epics/src/hosts.ts
```diff
@@ -15,36 +15,10 @@ import { RecordOf } from "immutable";
...
-    path: filepath,
-    type,
-    created: "",
-    last_modified:
-      lastSaved && lastSaved.toString() ? lastSaved.toString() : "",
-    content: notebook,
-    mimetype: mimetype || "",
-    format: "json"
-  };
-}
```
rgbkrk, line 46: Interesting, I didn't realize we had an unused function in here.

rgbkrk (approved):

PR #4833, opened 2020-01-15
Remove fs-kernels package from monorepo

captainsafia:
Removing this package from the monorepo. We'll pursue this work elsewhere.

If we need to reference this code, let's do so from the commit history.


rgbkrk (approved):

rgbkrk:
Extra notes: no one was using this package and we didn't push it all the way to completion and adoption.

PR #4834, opened 2020-01-15
Change acceptPayloadMessage to take JupyterMessage

javag97:
Before, the function was taking a custom payload object. Typing now allows
JupyterMessage to be accepted as payload.

### DIFF
--- a/packages/epics/src/execute.ts
+++ b/packages/epics/src/execute.ts
@@ -42,8 +42,7 @@ import {
   AppState,
   ContentRef,
   InputRequestMessage,
-  KernelStatus,
-  PayloadMessage
+  KernelStatus
 } from "@nteract/types";
 import { List } from "immutable";
 
@@ -80,7 +79,7 @@ export function executeCellStream(
 
   const cellAction$ = merge(
     payloadStream.pipe(
-      map((payload: PayloadMessage) =>
+      map((payload: JupyterMessage) =>
         actions.acceptPayloadMessage({ id, payload, contentRef })
       )
     ),
--- a/packages/messaging/src/index.ts
+++ b/packages/messaging/src/index.ts
@@ -1,4 +1,3 @@
-import { PayloadMessage } from "@nteract/types";
 import { from, Observable, Subscriber } from "rxjs";
 import { filter, map, mergeMap } from "rxjs/operators";
 import { executeRequest, message } from "./messages";
@@ -231,16 +230,12 @@ export const updatedOutputs = () => (source: Observable<JupyterMessage>) =>
  *     payloads()
  *   )
  */
-export const payloads = () => (
-  source: Observable<JupyterMessage>
-): Observable<PayloadMessage> =>
-  source.pipe(
-    ofMessageType("execute_reply"),
-    map(entry => entry.content.payload),
-    filter(p => !!p),
-    mergeMap((p: Observable<PayloadMessage>) => from(p))
-  );
-
+export const payloads = () => (source: Observable<JupyterMessage>):
+Observable<JupyterMessage> => source.pipe(
+  ofMessageType("execute_reply"),
+  mergeMap((p: Observable<JupyterMessage>) => from(p))
+);
+  
 /**
  * Get all the execution counts from an observable of jupyter messages
  */
--- a/packages/reducers/src/core/entities/contents/notebook.ts
+++ b/packages/reducers/src/core/entities/contents/notebook.ts
@@ -25,7 +25,7 @@ import {
   OnDiskOutput,
   OnDiskStreamOutput
 } from "@nteract/commutable";
-import { UpdateDisplayDataContent } from "@nteract/messaging";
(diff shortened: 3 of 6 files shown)


packages/actions/src/actionTypes/index.ts
```diff
@@ -111,11 +112,7 @@ export interface UpdateDisplayFailed {
 export const ACCEPT_PAYLOAD_MESSAGE = "ACCEPT_PAYLOAD_MESSAGE";
 export interface AcceptPayloadMessage {
   type: "ACCEPT_PAYLOAD_MESSAGE";
-  payload: {
-    id: CellId;
-    payload: PayloadMessage;
```
captainsafia: Ah to me more detailed, we only want this payload type to be a JupyterMessage. id and contentRef are still needed.

packages/epics/src/execute.ts
```diff
@@ -76,7 +75,7 @@ export function executeCellStream(
 
   const cellAction$ = merge(
     payloadStream.pipe(
-      map((payload: PayloadMessage) =>
+      map((payload: any) => //get this to work with a jupyter message
```
captainsafia: Let's change this to `JupyterMessage`. You will also need to update the `payloads` method in `packages/messaging/src/index.ts` which is used to generate the `payloadStream` object here.

packages/messaging/src/index.ts
```diff
@@ -233,12 +233,12 @@ export const updatedOutputs = () => (source: Observable<JupyterMessage>) =>
...
 export const payloads = () => (
   source: Observable<JupyterMessage>
-): Observable<PayloadMessage> =>
+): Observable<JupyterMessage> =>
   source.pipe(
     ofMessageType("execute_reply"),
     map(entry => entry.content.payload),
     filter(p => !!p),
-    mergeMap((p: Observable<PayloadMessage>) => from(p))
+    mergeMap((p: Observable<JupyterMessage>) => from(p))
```
captainsafia: There is one more change that needs to be made here.

Since we are returning a `JupyterMessage` and not a `PayloadMessage`, we will need to remove the `map` on line 238.

This map extracts the `content.payload` property from the message, but we don't want that anymore. Instead, we'll just return the entire message.

We'll also need to update the `acceptPayloadMessage` in `packages\reducers\src\core\entities\contents\notebook.ts` to take the updated format from the reducer.

javag97:
Status Update: 
It looks like I've gotten merge conflicts since this pull request has been open for a while. I can try to fix that on my feature branch and see how that goes. I am currently passing tests, although it looks like payload types `edit` and `ask_exit` are not being tested, so that will change.

captainsafia:
@javag97 The structure of the payload message is a little different. Here is an example of a this type of message.

```json
{
   "header":{
      "msg_id":"82f36e0e-17af3c883025ae499dc1f660",
      "msg_type":"execute_reply",
      "username":"username",
      "session":"3e37709b-9eae434d4f21a1641a80a792",
      "date":"2020-02-04T19:36:38.753971Z",
      "version":"5.3"
   },
   "msg_id":"82f36e0e-17af3c883025ae499dc1f660",
   "msg_type":"execute_reply",
   "parent_header":{
      "msg_id":"27ad5f9cecea41408e4ffea0fb126298",
      "username":"username",
      "session":"67f06f92bdf241728e2c602cc6d15bfe",
      "msg_type":"execute_request",
      "version":"5.2",
      "date":"2020-02-04T19:36:38.749972Z"
   },
   "metadata":{
      "started":"2020-02-04T19:36:38.749972Z",
      "dependencies_met":true,
      "engine":"9678b838-0f3c-4e59-ae18-0bd17bc15af3",
      "status":"ok"
   },
   "content":{
      "status":"ok",
      "execution_count":2,
      "user_expressions":{

      },
      "payload":[
         {
            "source":"set_next_input",
            "text":"TEST",
            "replace":false
         }
      ]
   },
   "buffers":[

   ],
   "channel":"shell"
}
```

This was generated by having a code cell with the following:

```python
def create_new_cell(contents):
    from IPython.core.getipython import get_ipython
    shell = get_ipython()

    payload = dict(
        source='set_next_input',
        text=contents,
        replace=False,
    )
    shell.payload_manager.write_payload(payload, single=False)
```

Then calling it like so.

```
create_new_cell("TEST")
```

And you will get the expected output above.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -552,35 +554,41 @@ function acceptPayloadMessage(
...
-        payload: {
-          cellType: "code",
-          // TODO: is payload.text guaranteed to be defined?
-          source: payload.text || "",
-          id,
-          contentRef: action.payload.contentRef
-        }
-      });
+  if(message.content.status === 'ok'){
+    const payload: any = message.content.payload[0];
```
captainsafia: This can be typed as a `PayloadMessage`.

Also why are we picking from a index here?

PR #4837, opened 2020-01-17
Fix desktop integration in AppImage package

### LINKED ISSUE #2661: Automatic file association doesn't work on ubuntu 17.10
I just download the AppImage from the website. Running the app for the first time I get asked whether I want to integrate it with the system. I said yes, and now `nteract` appears among the applications and in the context menu in nautilus on ipynb files. However, neither works (while directly executing the app file from the terminal works as intended).

Running the `nteract` launcher that is added in the dash reveals the possible reason: it gives the error 
>Failed to launch nteract
>Failed to change to directory "/opt/nteract/" (no such file or directory)

Indeed, `/opt/nteract` does not exist.

hakon-j-d-johnsen:
By removing the hardcoded reference to /opt/nteract in the configuration for electron-builder, this PR fixes #2661. It also fixes an issue with the .deb package, where Nteract would attempt to save unnamed notebooks to `/opt/nteract/Untitled.ipynb`, where the user does not have write permission, as described in https://github.com/nteract/nteract/issues/2661#issuecomment-575561001.

The hardcoded path existed in `build -> linux -> desktop -> Path` of `package.json`, and specifies the Path key  for the .desktop files generated by electron-builder. This Path key of .desktop files control the working directory where the application is launched, and is not necessary in our case because nteract does not need to be run with a specific working directory.

**Note:**
I have not replicated your build system in order to test this PR. I have, however, tested that removing the Path key from the .desktop files fixes AppImage installations, and improves .deb installations. It seems excessive to spend  lot of time setting up a correct build system for such a small PR, but feel free to let me know if it would be necessary before it could be accepted.

### DIFF
--- a/applications/desktop/package.json
+++ b/applications/desktop/package.json
@@ -90,7 +90,6 @@
         "Icon": "nteract",
         "MimeType": "application/x-ipynb+json;",
         "Name": "nteract",
-        "Path": "/opt/nteract/",
         "StartupNotify": "true",
         "Terminal": "false",
         "Type": "Application",


captainsafia:
Thanks for submitting this PR, @hakon-j-d-johnsen!

I'll validate this by building the desktop app locally to see if this doesn't cause any issues. Stay tuned for a follow-up.

PR #4860, opened 2020-01-25
Support "addCloseObserver" as configurable property in JupyterHostRecord

xuke444:

Add closeObserver pass to rxjs, so consumer application can detect web socket connection failure.

### DIFF
--- a/packages/epics/__tests__/websocket-kernel.spec.ts
+++ b/packages/epics/__tests__/websocket-kernel.spec.ts
@@ -41,12 +41,14 @@ describe("launchWebSocketKernelEpic", () => {
     const contentRef = "fakeContentRef";
     const kernelRef = "fake";
     const hostRef = "fakeHostRef";
+    const closeObserver = { next: () => {} }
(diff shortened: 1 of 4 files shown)


captainsafia (commented):
How did you validate this change?

xuke444:
> How did you validate this change?

That I actually need some help.
I cannot plug compiled code into my test project, so instead I changed compiled js in node_modules folder of my test project and test.

captainsafia:
Got it! The best place to validate this would be the nteract Jupyter extension since it is an nteract client connected to a Jupyter web server.

To set it up, you will need to follow [these instructions in the contribution docs](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md#how-do-i-make-a-contribution) and [these instructions in the nteract Jupyter extension](https://github.com/nteract/nteract/blob/master/applications/jupyter-extension/README.md).

Can you share the `closeObserver` function you used to test this? I can try the validation in my set up as well.

PR #4866, opened 2020-01-26
Change windows installer to allow change of directory

### LINKED ISSUE #4804: Optional install location
When attempting to install, it defaults to Program Files. Better would be to let the user define where to install, in case they want to install in user-space only. Users without admin access cannot install to C:/Program Files/*

Please provide an option at install to allow for users to decide where to install their programs.

CrystallineCat:
Fixes #4804. I don't have windows access, so couldn't test the change.

### DIFF
--- a/applications/desktop/package.json
+++ b/applications/desktop/package.json
@@ -69,7 +69,8 @@
     ],
     "nsis": {
       "perMachine": true,
-      "oneClick": false
+      "oneClick": false,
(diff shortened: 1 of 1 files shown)


tomrod-pcci:
Much obliged folks. This helps a lot!

captainsafia:
@tomrod-pcci No problem. This should go out in today's nteract release!

PR #4868, opened 2020-01-26
Some ansi-to-react fixes

### LINKED ISSUE #4815: <Ansi /> component should have optional linkify prop
Not following the issue template because I'm not quite sure how this fits in. I just wanted to point out that the `linkify` prop of the `Ansi` component (in `ansi-to-react`) should be optional in TypeScript (all of the examples show it as optional).

### LINKED ISSUE #4816: Ansi component doesn't return correct classes
Package: `ansi-to-react`

**Describe the bug**
The current implementation doesn't differentiate between foreground and background when using the `useClasses` option.

**To Reproduce**
```jsx
// \u001b[40 is the ANSI code for black background
<Ansi>
{"\u001b[40mhello\u001b[0m"}
</Ansi>
```
results in
```html
<span class="ansi-black">hello</span>
```

**Expected behavior**
Generated classes should have `-fg`/`-bg` suffixes (this is what Anser does when generating html with the `use_classes` option, but because you're using the JSON functionality, it only returns the colors).
```html
<span class="ansi-black-bg">hello</span>
```

This arguably might be an issue with Anser itself.

CrystallineCat:
Fixes #4815. Fixes #4816.

I also made `children` optional to fix the typescript errors in the test file.

### DIFF
--- a/packages/ansi-to-react/README.md
+++ b/packages/ansi-to-react/README.md
@@ -42,21 +42,29 @@ Style with classes instead of `style` attribute.
 Will render
 ```javascript
 <code>
-    <span class="ansi-blue">hello world</span>
+    <span class="ansi-blue-fg">hello world</span>
 </code>
 ```
 
 #### Class Names
 |Font color| Background Color
 |---|---|
-|ansi-black|ansi-bright-black
-|ansi-red|ansi-bright-red
-ansi-green|ansi-bright-green
-ansi-yellow|ansi-bright-yellow
-ansi-blue|ansi-bright-blue
-ansi-magenta|ansi-bright-magenta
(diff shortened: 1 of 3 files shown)


twavv:
Yay! I meant to get around to opening a PR for this but life happens.

As an aside, should this be published as a new major version of ansi-to-react as it is technically brealing?

rgbkrk (approved):

rgbkrk:
Yeah, we'll want to make this package be a major release bump due to the CSS class name change at the very least.

captainsafia:
I'll go ahead and merge this in advance of today's release.

Per the discussion in the issue, I'll tag this with a major version release.

PR #4869, opened 2020-01-26
Revert "Update dependency mkdirp to v1"

CrystallineCat:
Reverts nteract/nteract#4859 due to it breaking `desktop`.


CrystallineCat:
Merging to get `master` back into a runnable shape.

captainsafia:
Thanks for handling this, @CrystallineCat!

I've been heads down on adding tests to our codebase so I'll add some more to cover the fs-observables package so we can catch these kinds of regressions in the build.

PR #4871, opened 2020-01-26
Migrate from react-notification-system to blueprintjs's Toaster

### LINKED ISSUE #3503: Switch from react-notification-system to blueprint.js Toaster
Now that we're using blueprint.js, I'd like to make use of their [toasts](https://blueprintjs.com/docs/#core/components/toast). We'd probably switch to having a `Toaster` in the store like we do with the notification system. It's either that or a singleton.

### LINKED ISSUE #4768: Remove "Save Successful" toast on autosave?

### LINKED ISSUE #4870: Notification System problems
At the moment, `desktop` on `master` is generating a `SET_NOTIFICATION_SYSTEM` with a payload of
```typescript
{ notificationSystem: undefined }
```

This appears due to a mix of `ref` styles; changing `notebook/index.tsx` to
```tsx
<NotificationSystem ref={this.notificationSystem} />
```
will fix this issue. However, when a notification is then issued, I get errors in the console, which appears to be a often reported spurious bug, e.g.:
* https://github.com/igorprado/react-notification-system/issues/168
* https://github.com/igorprado/react-notification-system/issues/165
* https://github.com/igorprado/react-notification-system/issues/143

The core issue appears to be that when the elements inside the notification system's render function are created, the `owner` is set to undefined. I've stepped through with the debugger, and the static `ReactCurrentOwner.current` is set at the beginning of the render function, but is undefined at the time of the `createElement`-call.

The reason for this might be due to webpack splitting the object. It has several `ReactCurrentOwner`s, and I'm unsure how to trace that they all refer to the same object.

I guess this would prioritize #3503?

CrystallineCat:
Fixes #3503. Fixes #4870. Fixes #4768.

<s>ATM a hack to get notifications working in desktop again. Could be merged as-is if getting notifications working in desktop quickly is important.</s>


captainsafia:
Since the desktop app was released yesterday, we have an entire month to get this polished up before next month's release.

CrystallineCat:
Most of the work on the new notification system is done... :smile_cat: 

Peek 2020-02-06 17-57

rgbkrk:
Wow! That is looking _good_.

CrystallineCat:
Dark theme and `nteract_on_jupyter` support:

jupyter-extension

All that's left to do is some tests... :smile_cat:

tslint.json
```diff
@@ -40,10 +40,7 @@
     "no-var-requires": {
       "severity": "warn"
     },
-    "object-literal-sort-keys": {
-      "options": "match-declaration-order",
-      "severity": "warn"
-    },
+    "object-literal-sort-keys": false,
     "only-arrow-functions": [true, "allow-declarations"],
```
CrystallineCat, line 44: Changed this config setting 'cause all the warnings annoyed me, and they don't seem to really serve much of a purpose, and they don't work properly with `extends` AFAICT.

packages/types/src/index.ts
```diff
@@ -175,12 +152,3 @@ export interface AppState {
...
 }
-
-export type AppStateRecord = Immutable.RecordOf<AppState>;
-
-export const makeAppStateRecord = Immutable.Record<AppState>({
-  app: makeAppRecord(),
-  comms: makeCommsRecord(),
-  config: Immutable.Map<string, any>(),
-  core: makeStateRecord()
-});
```
CrystallineCat, line 186: Both of those appear to be completely unused.

packages/core/src/middlewares.ts
```diff
@@ -2,53 +2,9 @@
...
-import * as selectors from "@nteract/selectors";
-
-interface ErrorAction {
-  type: string;
-  error?: boolean;
-  payload?: any;
-}
-
-export const errorMiddleware = (store: any, console = global.console) => (
-  next: any
```
CrystallineCat, line 18: Handled by the notification system directly now.

CrystallineCat:
This PR adds two new packages:
- `myths`, (the beginnings of) a framework to manage dependencies between actions/reducers/epics/components better and reduce boilerplate code. A set of closely-related actions/reducers/epics/components is now known as a `Myth`, a package with myths as a `MythicPackage`.
- `mythic-notifications`, the notifications system using the above to centralize all the notifications-related actions/reducers/epics/components.
Neither of them depend on any other `@nteract/...` packages.

The latter exports three things:
- `notifications`, a `MythicPackage`, used to create the rootReducer/rootEpic.
- `NotificationRoot`, a connected component that initializes the store when mounted.
- `sendNotification`, a `Myth`, which can create dispatchable actions with `sendNotification.create`, be used in filtering with `ofType(sendNotification.type)` or used in typing with `typeof sendNotification.action`, `typeof sendNotification.props`, `typeof sendNotification.type`.

I added `NotificationRoot` to all three variations of `notebook-app-component`; added `notifications` to both applications' `configureStore`; and changed all calls to `addNotification` to `sendNotification.create` actions. These are all the dependencies on `mythic-notifications`. In particular, there are no dependencies from `@nteract/types`, `@nteract/core`, `@nteract/actions`, or `@nteract/reducers` onto the new package.

The current notification system object used to process the notifications is still part of the store, but now considered private state of the notifications package.

I don't really like the `mythic-` prefix on the package name, but I want all `MythicPackage`s to sort next to each other.

applications/desktop/src/notebook/epics/github-publish.ts
```diff
@@ -64,18 +60,17 @@ export const publishEpic = (
 
       const contentRef = action.payload.contentRef;
       if (!contentRef) {
-        return empty();
+        return EMPTY;
```
captainsafia, line 63: Neat! TIL: it is better to use `EMPTY` over `empty()`.

applications/desktop/src/notebook/menu.tsx
```diff
@@ -607,22 +613,47 @@ export function exportPDF(
         )
       );
 
-      const appNotificationSystem = state.app.get("notificationSystem");
-
       fs.writeFile(pdfPath, data, _error_fs => {
-        appNotificationSystem.addNotification({
+        const pdfPathParts = pdfPath.split("/");
+        const Spacer = styled.div`
```
captainsafia, line 620: Mind adding a comment to clarify that these are needed to style the contents of the notification bar specifically for PDF generation notifications?

packages/core/src/middlewares.ts
captainsafia: Does the notification system also handle logging? I guess I'll find out as I keep reviewing but it would be good to decouple logging for devs from notifications for end-users.

packages/epics/src/index.ts
```diff
@@ -1,37 +1,10 @@
...
-  executeCellAfterKernelLaunchEpic,
-  executeCellEpic,
-  executeFocusedCellEpic,
-  lazyLaunchKernelEpic,
-  sendExecuteRequestEpic,
-  sendInputReplyEpic,
-  updateDisplayEpic
-} from "./execute";
+import { autoSaveCurrentContentEpic, closeNotebookEpic, fetchContentEpic, saveAsContentEpic, saveContentEpic, updateContentEpic } from "./contents";
+import { executeAllCellsEpic, executeCellAfterKernelLaunchEpic, executeCellEpic, executeFocusedCellEpic, lazyLaunchKernelEpic, sendExecuteRequestEpic, sendInputReplyEpic, updateDisplayEpic } from "./execute";
```
captainsafia, line 3: Nit: I feel like the linter should put in new lines here.

packages/epics/src/kernel-lifecycle.ts
```diff
@@ -6,9 +6,10 @@ import {
   JupyterMessage,
   ofMessageType
 } from "@nteract/messaging";
+import { sendNotification } from "@nteract/mythic-notifications";
```
captainsafia, line 9: Question on how the myths pattern is used. If an application is using the `@nteract/epics` package with these changes, will they need to have the `@nteract/mythic-notifications` package registered in the `configureStore` function of their application? Can @nteract/epics still be used independently of @nteract/mythic-notifications?

packages/mythic-notifications/README.md
```diff
@@ -0,0 +1,73 @@
+# @nteract/mythic-notifications
+
+*N.B.: I don't really like the `mythic-` part of this name, but I like this and future `myths`-based packages to sort together...*
+
+This package implements a notification system based on `blueprintjs`.
```
captainsafia: Can we add a cross-link to the README of the @nteract/myths package and clarify that this package is based on the semantics of nteract myths?

packages/mythic-notifications/README.md
```diff
@@ -0,0 +1,73 @@
+# @nteract/mythic-notifications
+
+*N.B.: I don't really like the `mythic-` part of this name, but I like this and future `myths`-based packages to sort together...*
```
captainsafia: I think the myths name is quite clever!

Does `mythical-notifications` seem like a better name? I have no strong opinions on how myth packages are prefixed.

packages/mythic-notifications/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@blueprintjs/core": "^3.7.0",
+    "@nteract/myths": "^0.1.0"
+  },
+  "peerDependencies": {
+    "immutable": "^4.0.0-rc.12",
+    "react": "^16.3.2",
+    "react-dom": "^16.3.2",
+    "styled-components": ">= 4"
+  },
+  "author": "Magdalena Kiesel <magda.kiesel@gmx.net>",
```
captainsafia: Nit: We should probably standardize on a generic author label for here. "nteract contributors" is what we use in some of the packages so it makes sense to use here.

A lot of our other packages need to use a standard author name as well, as opposed to individual names and emails.

No need to address it here if you don't have time, we can do a pass and cover all the package.json files in the monorepo seperately.

packages/mythic-notifications/src/myths/send-notification.ts
```diff
@@ -0,0 +1,36 @@
+import { MythicAction } from "@nteract/myths";
+import { notifications } from "../package";
+import { NotificationMessage } from "../types";
+
+const titleFromAction = (action: MythicAction) => {
+  const text = action.type
+    .replace(/.*\//, "")
```
captainsafia: Nit: Can these regular expressions be moved to named variables for readability?

packages/myths/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+```typescript
+export const addToSum =
+  iCanAdd.createMyth("addToSum")<number>({
+    reduce: (state, action) =>
+      state.set("sum", state.get("sum") + action.payload),
+  });
+```
+
+A package can have any number of myths. To create an action based on a myth, use its create function. You can then dispatch this action normally:
```
captainsafia, line 43: Does this mean that a package can have multiple states? For example, I can have a "widgets" package that implements three different states each with their own reducers?

packages/myths/__tests__/myths.spec.ts
```diff
@@ -0,0 +1,72 @@
+import { createMythicPackage, makeConfigureStore } from "../src";
+
+describe("myths", () => {
+  test("can create a mythic package and use it to alter state", () => {
+    const iCanAdd = createMythicPackage("iCanAdd")<
```
captainsafia, line 5: Ah! I see from these tests how a myth can have multiple states. Does this end up looking like this in the final state?

```js
{
  core,
  app,
  config,
  iCanAdd: {
    sum
  }
}
```

packages/myths/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "keywords": [
+    "nteract"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+  },
+  "dependencies": {
+    "react-redux": "^6.0.0",
```
captainsafia, line 19: Should we make these peer-dependencies so that we can rely on the versions of the package that the user's applications are using?

packages/myths/__tests__/myths.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      },
+    });
+
+    const addToSum =
+      iCanAdd.createMyth("addToSum")<number>({
+        reduce: (state, action) =>
+          state.set("sum", state.get("sum") + action.payload),
+        epics: [
+          {
+            on: action => action.payload >= 10,
```
captainsafia, line 52: I'm having trouble understanding how these get mapped to epics.

I assume that `on` and `create` are the individual epics and the `action.payload >= 10` parts are the contents of the filter statement?

Maybe I'll figure this out as I continue reading through the code...

packages/myths/src/epics.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  for (const definition of definitions ?? []) {
+    if ("create" in definition) {
+      const def: CreateEpicDefinition<PROPS> = definition;
+
+      epics.push(
+        (action$: ActionsObservable<MythicAction>) =>
+          action$.pipe(
+            filter(def.on),
+            map(def.create),
```
captainsafia, line 25: Ah, I see! The `on` represents the filter and the `create` is the map. Does it make more sense to rename `create` to `do`?

packages/myths/src/myth.ts
```diff
@@ -0,0 +1,48 @@
+import { Action } from "redux";
+import { makeEpics } from "./epics";
+import { makeCreateConnectedComponent } from "./react";
+import { makeReducer } from "./reducer";
+import { Myth, MythDefinition, MythicAction, Myths } from "./types";
+
+export const makeCreateMyth =
```
captainsafia, line 7: Quick question: do you have ideas for other things in the nteract codebase that we can turn to myths to continue validating this design pattern?

packages/myths/src/myth.ts
```diff
@@ -0,0 +1,48 @@
...
+          name,
+          type: `${pkg}/${name}`,
+
+          // for use in typeof expressions:
+          props: undefined as unknown as PROPS,
+          state: undefined as unknown as STATE,
+          action: undefined as unknown as MythicAction<PKG, NAME, PROPS>,
+        };
+
+        mythWIP.create =
```
captainsafia, line 32: Can you add some comments to each of these properties to document what they are for?

packages/myths/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export { createMythicPackage } from "./package";
```
captainsafia, line 1: If you have time, it would be good to create a `docs` directory for this new package.

Check out the [actions package](https://github.com/nteract/nteract/tree/master/packages/actions) for an example of how this is structured with mkdocs.

packages/types/src/index.ts
captainsafia: Considering the other changes in this PR will warrant a major version release, I think it is fine to remove them and release them in that major version release as a breaking change.

captainsafia (commented):
This PR was a very informative read! I've left a few comments and questions inline.

applications/desktop/src/notebook/epics/github-publish.ts
CrystallineCat: Yeah, `empty()` is deprecated. My IDE shows that as <s>`empty()`</s>, which looks ugly, so I'm always prompted to replace them when I see them.. :joy_cat:

applications/desktop/src/notebook/menu.tsx
CrystallineCat: Done.

packages/epics/src/index.ts
CrystallineCat: At the moment our config for this is:
```json
    "max-line-length": {
      "options": {
        "limit": 80,
        "ignore-pattern": "^import |^export {(.*?)}"
      },
      "severity": "warn"
    },
```

My IDE removes them when I *Optimize import*, and I kinda prefer this over one item per line, but I can change that, if you really want me to. :cat:

packages/epics/src/kernel-lifecycle.ts
CrystallineCat: It can be used independently, though of course it'll pull the package into, e. g., webpack. If you don't register it in `configureStore`, all the happens is that the `notifications/sendNotification` actions will go unreduced and no notifications are emitted for errors.

packages/core/src/middlewares.ts
CrystallineCat: The default `consoleNotificationSystem` does, but it gets replaced when a `<NotificationRoot/>` is mounted.

Hmm... I have to think about the implications of this some more...

packages/mythic-notifications/README.md
CrystallineCat: Done.

packages/mythic-notifications/README.md
CrystallineCat: Thank you. :cat: Took a while to come up with a name that wasn't used on `npm`.

My dislike is about having a prefix at all, I'd rather not, but I also want the `packages` dir to be somewhat structured. I've removed the note.

packages/mythic-notifications/package.json
CrystallineCat: Yeah, I'd actually be more comfortable with that. I just did what the package I copied this from did. :stuck_out_tongue:

Done.

packages/mythic-notifications/src/myths/send-notification.ts
CrystallineCat: Done.

packages/myths/README.md
CrystallineCat: Every package has a single private state. Of course, it can reserve different parts of that state for different myths, if it wants, but each reducer will be passed the package's full private state.

This might change in the future, of course, if I find a use case where it would be best to change it. `myths`'s interface is still very much WIP.

packages/myths/__tests__/myths.spec.ts
CrystallineCat: Close!
```
{
  core,
  app,
  config,
  __private__: {
    iCanAdd: {
      sum
    }
  }
}
```

packages/myths/package.json
CrystallineCat: Hmm... all the other packages have it under `dependencies`, except for top-level, where it is `devDependencies`. That said, I really don't understand how the `package.json` stuff works in the monorepo; this weekend I added `@nteract/mythic-notifications` to `applications/desktop`, which caused React to get included in the bundle twice, which caused the store provider not to work, which wasn't fun at all to track down. :stuck_out_tongue_closed_eyes:

So, your call. :cat:

packages/myths/__tests__/myths.spec.ts
CrystallineCat: I'll address this below...

packages/myths/src/epics.ts
CrystallineCat: This part is very much unfinished. Essentially, I noticed that many epics have similar patterns in how they use the actions observable and tried to abstract that. The `{on: ..., create: ...}`-one is the only one implemented to far, because it was the only one need for the Notifications package.

I've considered making it a function, as in `createEpicFrom({on: ..., create: ...})` to make it clearer, but I'm not sure...

I have plans to revisit this when building upon `myths` in the future, so I'd like to postpone the issue for now.

packages/myths/src/myth.ts
CrystallineCat: Oh, do I ever! Here's my rough to do for the next while:
- Implement config centralization as `mythic-configuration`.
- Implement the preferences window; not sure if that will use `myths` past the configurations one.
- Implement search, which probably will end up as `mythic-search`, but haven't looked at it in detail yet.
- Clean up the mess that is `menu.ts(x)` in desktop, which will involve moving out GitHub publishing (into probably `mythic-publication`) and PDF exports.
- Implement connecting to existing kernels, which will likely mean that I'll turn some kernel behaviour into a myth, but not sure exactly yet.
- Turn the contents interface into a mythic package, which will require adding support for failed and fulfilled states on actions to `myths`.

I can probably find more, but that's what I've though of doing so far, and I think that's enough for now. :smile_cat:

packages/myths/src/myth.ts
CrystallineCat: Done.

packages/myths/src/index.ts
CrystallineCat: Hmm... I think it's a bit premature to document `myths` just yet, as I still expect it to change when I implement the next mythic packages. I will keep it in mind, though!

CrystallineCat:
Thank you for your review @captainsafia!

packages/myths/src/index.ts
captainsafia: Maybe a design doc or more code comments? Something to guide devs who are looking at this as we continue to explore this model more.

packages/myths/src/myth.ts
captainsafia: Great list! I like the idea of factoring out pieces of functionality into myths. Will make it a lot easier to test and swap out implementations as the project evolves.

> Implement connecting to existing kernels, which will likely mean that I'll turn some kernel behaviour into a myth, but not sure exactly yet.

Particularly interested in seeing how this will intersect with the proposal I laid out in https://github.com/nteract/nteract/issues/4902.

captainsafia (approved):
The build is passing and a light QA revealed one small issue -- all failures to save the file are surfaced as "Save failed...possibly open in another tab..."

@CrystallineCat If this is something quick to fix in this PR, it would be good to get it in.

packages/myths/src/index.ts
CrystallineCat: Sure, I'll write some more as part of the next mythic package.

CrystallineCat:
> The build is passing and a light QA revealed one small issue -- all failures to save the file are surfaced as "Save failed...possibly open in another tab..."

I'll take a look at that and address it as a separate PR, that error comes from the contents epics and I don't exactly understand the logic there yet...

Thank you for your detailed comments! I'll rebase and merge this. :cat:

PR #4875, opened 2020-01-27
Add ansi-{decoration} classes to Ansi component

twavv:

Followup to #4816.

This adds classes like `ansi-bold` to the output of the `<Ansi />` component when using CSS classes. The declaration for `decoration` is
```ts
export interface AnserJsonEntry {
  // ...
  decoration: null | 'bold' | 'dim' | 'italic' | 'underline' | 'blink' | 'reverse' | 'hidden' | 'strikethrough';
  // ...
}
```

### DIFF
--- a/packages/ansi-to-react/__tests__/index.spec.ts
+++ b/packages/ansi-to-react/__tests__/index.spec.ts
@@ -5,6 +5,7 @@ import Ansi from "../src/index";
 
 const GREEN_FG = "\u001b[32m";
 const YELLOW_BG = "\u001b[43m";
+const BOLD = "\u001b[1m";
 const RESET = "\u001b[0;m";
 
 describe("Ansi", () => {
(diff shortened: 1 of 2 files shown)


twavv:
I will add some tests too. :^)

captainsafia:
@travigd Great! If you are able to get around to this today we can bundle this change with the other change in today's release of nteract.

twavv:
@captainsafia Pushed some tests!

It's worth noting that the text decoration stuff isn't implemented when **not** using classes. It wasn't possible at all before, so this is definitely strictly an improvement, but might wanna address that someday. Some of the decorations don't really make sense (e.g. `reverse`?) so we'd probably only want to support the common ones. Something like that.

captainsafia:
@travigd Thanks! Do you mind opening an issue to track implementing the follow-ons?

twavv:
Opened #4879.

PR #4898, opened 2020-01-31
Add support for storing execution stats per nbformat

### LINKED ISSUE #2660: Support execution times in notebook per nbformat v4.4
As [stated in the spark roadmap](https://github.com/jupyter/roadmap/blob/master/spark.md#timing-data), we want to store how long it took to run a cell. The transient information for how long execution took is available in the message spec.

[JupyterLab has started adopting it with the same metadata the `execute_time` extension was using](https://github.com/jupyterlab/jupyterlab/pull/4161#issuecomment-371929527).

captainsafia:
Closes https://github.com/nteract/nteract/issues/2660.

### DIFF
--- a/packages/epics/src/execute.ts
+++ b/packages/epics/src/execute.ts
@@ -85,6 +85,54 @@ export function executeCellStream(
       )
     ),
 
+    /**
+     * Set the ISO datetime when the execute_input message
+     * was broadcast from the kernel, per nbformat.
+     */
+    cellMessages.pipe(
+      ofMessageType("execute_input"),
+      map(() =>
+        actions.setInCell({
+          id,
+          contentRef,
+          path: ["metadata", "execution", "iopub", "execute_input"],
+          value: new Date().toISOString()
+        })
+      )
+    ),
+
+    /**
+     * Set the ISO datetime when the execute_reply message
+     * was broadcast from the kernel, per nbformat.
+     */
+    cellMessages.pipe(
+      ofMessageType("execute_reply"),
+      map(() =>
+        actions.setInCell({
+          id,
+          contentRef,
+          path: ["metadata", "execution", "shell", "execute_reply"],
+          value: new Date().toISOString()
+        })
+      )
+    ),
+
+    /**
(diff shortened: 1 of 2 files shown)


packages/epics/src/execute.ts
```diff
@@ -85,6 +85,54 @@ export function executeCellStream(
...
+     * Set the ISO datetime when the execute_input message
+     * was broadcast from the kernel, per nbformat.
+     */
+    cellMessages.pipe(
+      ofMessageType("execute_input"),
+      map(() =>
+        actions.setInCell({
+          id,
+          contentRef,
+          path: ["metadata", "execution", "iopub", "execute_input"],
```
MSeal: FYi in the spec I believe the key name at the end should be `iopub.execute_input` and not another map with `iopub` key:

https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.4.schema.json#L212-L227

MSeal: Same for the other settesr

MSeal (dismissed):
Unless I am mistaken, while the way described here seemed more intuitive from a object store pov it differs from the spec slightly.

captainsafia:
> Unless I am mistaken, while the way described here seemed more intuitive from a object store pov it differs from the spec slightly.

Ah! This makes sense. I was referring to the Sphinx docs when I was making this PR and my JS-leaning brain interpreted the periods as indicators for a nested object. I can update this to be string keys though.

packages/epics/src/execute.ts
rgbkrk: What a strange way to name a key in a json spec. 🙄

rgbkrk (approved):
Since this matches the spec with the new commit, this looks good.

packages/epics/src/execute.ts
MSeal: Yeah, I didn't catch that when it was PR'd in nbformat -- I can see it making sense from a namespace -> key, value pair POV where you want a fixed depth of settings. But I agree it does make runtime path assignments awkward.

PR #4900, opened 2020-01-31
Remove check for kernelspecs at launch

### LINKED ISSUE #4894: Desktop 0.21.0 | Should not require kernels to be installed to view notebooks
nteract desktop app

**Describe the bug**
As of nteract 0.21, the default mode is view-only, which should allow the user to view notebooks but not executre them. Hence, it should not be an error if there are no kernels available.

**To Reproduce**

1. Open nteract desktop v0.21 on a system with no kernels in global environment
2. See error

**Expected behavior**
User should be able to view and edit ipynb files, but not execute notebooks even if there are no kernels in the global environment.

**Screenshots**
```
[Window Title]
No Kernels Installed

[Main Instruction]
No kernels are installed on your system.

[Content]
No kernels are installed on your system so you will not be able to execute code cells in any language. You can read about installing kernels at https://ipython.readthedocs.io/en/latest/install/kernel_install.html
Full error: UNKNOWN: unknown error, lstat 'C:\Users\MYAJ\AppData\Local\Microsoft\WindowsApps\python.EXE'

[OK]
```

- OS: Windows 10
- Browser: N/A
- Version 0.21

**Additional context**
I only install kernels in virtual environments (e.g. conda or venv). My global Python environment is as clean as possible.

There are many issues about nteract not correctly detecting kernels, but this is a different angle.

captainsafia:
Closes https://github.com/nteract/nteract/issues/4894.

Now that we have a view-only mode in the desktop app, people do not need to have kernels installed in order to be able to view notebooks.

### DIFF
--- a/applications/desktop/src/main/index.ts
+++ b/applications/desktop/src/main/index.ts
@@ -28,7 +28,8 @@ import {
   first,
   mergeMap,
   skipUntil,
-  takeUntil, tap
+  takeUntil,
+  tap
 } from "rxjs/operators";
 
 import {
@@ -98,7 +99,7 @@ const jupyterConfigDir = join(app.getPath("home"), ".jupyter");
 const nteractConfigFilename = join(jupyterConfigDir, "nteract.json");
 
 const CONFIG = {
-  defaultKernel: "python3",
+  defaultKernel: "python3"
 };
 
 const prepJupyterObservable = prepareEnv.pipe(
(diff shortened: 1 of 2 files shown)


applications/desktop/src/main/index.ts
```diff
@@ -330,54 +328,16 @@ fullAppReady$.subscribe(() => {
...
-            "https://ipython.readthedocs.io/en/latest/install/kernel_install.html" +
-            `\nFull error: ${err.message}`
-        },
-        index => {
-          if (index === 0) {
-            app.quit();
-          }
-        }
-      );
+      console.error("Unexpected error when fetching kernelspecs: ", err);
```
rgbkrk, line 341: This makes it hard for people to discover how to get kernels working though right? We probably need some way to link to them how to install kernels (maybe an option in the menu?).

rgbkrk (approved):

captainsafia:
> This makes it hard for people to discover how to get kernels working though right? We probably need some way to link to them how to install kernels (maybe an option in the menu?).

Good call. Menu option seems like the most discrete thing to get going here.

PR #4906, opened 2020-02-03
Fix performance issues in editor components

### LINKED ISSUE #4885: 0.21.0 Desktop | Very slow editing markdown and code cell blocks
nteract desktop 0.21.0 | macOS 10.15.2
 - upgraded from nteract desktop 0.15.0 to 0.21.0 using dmg file release

**Describe the bug**
Editing cells is very slow and sluggish.

For example: When I type in '3333333' just fast, it takes about four seconds to render it.

**Expected behavior**

- Rendering while editing should be fluid.

macOS 10.15.2 & macOS 10.15.3

captainsafia:
Closes https://github.com/nteract/nteract/issues/4885.

For more information on the changes in `packages/editor/src/index.tsx` and `packages/stateful-components/src/inputs/editor.tsx` please see [this comment](https://github.com/nteract/nteract/issues/4885#issuecomment-581191883).

The changes in `packages/stateful-components/src/cells/cell.tsx` were made to resolve an issue I noticed where the `Cell` parent component was re-rendered every time anything in the cell changed since the `Cell` component takes the entire ImmutableCell object as a prop.

### DIFF
--- a/packages/stateful-components/src/cells/cell.tsx
+++ b/packages/stateful-components/src/cells/cell.tsx
@@ -17,7 +17,7 @@ interface StateProps {
 
 type Props = ComponentProps & StateProps;
 
-export class Cell extends React.Component<ComponentProps & StateProps> {
+export class Cell extends React.Component<Props> {
   shouldComponentUpdate(nextProps: Props): boolean {
     return nextProps.selected !== this.props.selected;
   }
--- a/packages/stateful-components/src/cells/cell.tsx
+++ b/packages/stateful-components/src/cells/cell.tsx
@@ -15,7 +15,13 @@ interface StateProps {
   selected: boolean;
 }
 
-export class Cell extends React.PureComponent<ComponentProps & StateProps> {
+type Props = ComponentProps & StateProps;
+
+export class Cell extends React.Component<ComponentProps & StateProps> {
+  shouldComponentUpdate(nextProps: Props): boolean {
+    return nextProps.selected !== this.props.selected;
+  }
+
   render(): JSX.Element | null {
     // We must pick only one child to render
     let chosenOne: React.ReactNode | null = null;
--- a/packages/stateful-components/src/inputs/editor.tsx
+++ b/packages/stateful-components/src/inputs/editor.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { connect } from "react-redux";
 import { Dispatch } from "redux";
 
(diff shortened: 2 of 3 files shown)


packages/stateful-components/src/inputs/editor.tsx
```diff
@@ -63,7 +70,14 @@ export class Editor extends React.Component<ComponentProps & StateProps> {
 
     // Render the output component that handles this output type
     return React.cloneElement(chosenOne, {
-      ...this.props,
```
eimfach, line 66: Is it necessary to copy all this props ?

packages/stateful-components/src/cells/cell.tsx
```diff
@@ -15,7 +15,13 @@ interface StateProps {
   selected: boolean;
 }
 
-export class Cell extends React.PureComponent<ComponentProps & StateProps> {
+type Props = ComponentProps & StateProps;
+
+export class Cell extends React.Component<ComponentProps & StateProps> {
```
eimfach: I am not really experienced with TypeScript, but wouldn't placing the new type alias into the class signature be better ?
`export class Cell extends React.Component<Props>`

packages/stateful-components/src/inputs/editor.tsx
captainsafia: Yes, they are the props that are commonly required by editor components. Before, we were passing all the props. Now we are passing all of them, except `children`.

packages/stateful-components/src/cells/cell.tsx
captainsafia: Good catch! I will update this.

packages/stateful-components/src/inputs/editor.tsx
eimfach: Okay, and what if only `value` changed ? Wouldn't it be copying all those other props repeatedly on every keystroke ? Just a thought ...

eimfach: I understand that this is convenient to do like this ...

captainsafia: Hmmm. Yes, we should probably add a stronger `shouldComponentUpdate` check here as well. I'll have to try this out later and validate how we can optimize this.

rgbkrk (approved):

rgbkrk:
Thanks for the review @eimfach! I'm going to go ahead with merging this to continue iterations in follow on PRs.

PR #4908, opened 2020-02-04
Lazy kernel launch fixes

victor-meng:
This PR includes two fixes for my previous [lazy kernel launch](https://github.com/nteract/nteract/pull/4822) change:
1. In `lazyLaunchKernelEpic`, using the `first` operator prevents the kernel to be launched in multi-tab senario. Instead of using `first`, we should use `distinct` and compare the `contentRef` from the payload. This will still prevent the kernel from being launched more than once within the same notebook, but allow other tabs to lazy launch kernel as well.
2. In `executeCellAfterKernelLaunchEpic`, if the kernel status is still "starting", then the requests in the message queue will never be executed because the `LAUNCH_KERNEL_SUCESSFUL` action is only emitted once. Since we have existing logic that queues the execute requests if the kernel is still starting, we can relax the status check to only filter out "not connected" and "shutting down" status.

### DIFF
--- a/packages/epics/src/execute.ts
+++ b/packages/epics/src/execute.ts
@@ -20,8 +20,8 @@ import { empty, merge, Observable, Observer, of, throwError } from "rxjs";
 import {
   catchError,
   concatMap,
+  distinct,
   filter,
-  first,
   groupBy,
   map,
   mapTo,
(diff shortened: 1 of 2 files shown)


packages/epics/src/execute.ts
```diff
@@ -447,9 +447,8 @@ export function executeCellAfterKernelLaunchEpic(
       return !!(
         kernel &&
         kernel.channels &&
-        (kernel.status === KernelStatus.Idle ||
```
captainsafia, line 450: Note to self: as part of the IKernelProvider work, we should standardize and create selectors for things like `isKerneRunning`.

captainsafia (approved):
Thanks for the quick fix and for detailing the issues in the PR.

PR #4918, opened 2020-02-10
Move from cloneElement to child render props

captainsafia:
Move from `React.cloneElement` to render props for the selection of Editors and the override of the Prompt sub-contents.

This has the benefits of improving the experience when working with TypeScript and reducing aggressive renders.


rgbkrk:
`cloneElement` leads to more renders than child render props? I'm pretty surprised at this!

PR #4922, opened 2020-02-11
Fix error handling when saving

CrystallineCat:
Follow up to an issue raised in #4871.

I looked at the logic, and it seems like the "open in another tab" error is supposed to prevent saving when the local and on-disk modified timestamps mismatched by more than 600 ms. It was apparently added to prevent files in multiple tabs from interfering, but I think this is extremely surprising behaviour, and it can prevent saving a file completely, e.g. for the test I `chmod 000`-ed a file, then `chmod 666`-ed it back, and I still couldn't save over it. I removed the behaviour entirely.

Additionally, the save epic didn't check the contents of the fake-XHR for an error response, so the actual `EACCES` didn't get through. I added that check.

save failed

I also fixed warnings in the file, which lead to removal of several guards that were guaranteed to never fail by the typechecker. This led to tests failing that provided malformed actions to the epic. I removed those tests; I don't think they make much sense with the typechecker.

### DIFF
--- a/packages/epics/__tests__/kernelspecs.spec.ts
+++ b/packages/epics/__tests__/kernelspecs.spec.ts
@@ -1,8 +1,8 @@
+import { fetchKernelspecs } from "@nteract/actions";
+import { JupyterHostRecordProps } from "@nteract/types";
 import { ActionsObservable } from "redux-observable";
 import { toArray } from "rxjs/operators";
-
-import { fetchKernelspecs } from "@nteract/actions";
-import { fetchKernelspecsEpic } from "../src/kernelspecs";
+import { fetchKernelspecsEpic } from "../src";
 
 describe("fetchKernelspecsEpic", () => {
   test("calls kernelspecs.list with appropriate jupyter host config", done => {
@@ -16,16 +16,12 @@ describe("fetchKernelspecsEpic", () => {
         app: {
           host: {
             type: "jupyter",
-            kernelRef: null,
-            id: null,
-            selectedKernelRef: null,
-            kernelspecsRef: null,
-            defaultKernelName: null,
-            activeKernelRefs: [],
             token: "x-x-x",
-            serverUrl: "http://localhost:8888/",
-            crossDomain: false
-          }
+            origin: "http://localhost:8888/",
+            basePath: "",
+            bookstoreEnabled: false,
+            showHeaderEditor: false,
+          } as JupyterHostRecordProps
         }
       }
     };
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -1,42 +1,16 @@
-import { stringifyNotebook, toJS } from "@nteract/commutable";
-import { Notebook } from "@nteract/commutable";
-import FileSaver from "file-saver";
-import { Action } from "redux";
-import { ofType } from "redux-observable";
-import { ActionsObservable, StateObservable } from "redux-observable";
-import { empty, from, interval, Observable, of } from "rxjs";
-import {
-  catchError,
-  concatMap,
-  distinctUntilChanged,
-  map,
-  mergeMap,
-  pluck,
-  switchMap,
-  tap,
-  take
-} from "rxjs/operators";
-
 import * as actions from "@nteract/actions";
+import { Notebook, stringifyNotebook, toJS } from "@nteract/commutable";
 import * as selectors from "@nteract/selectors";
-import {
-  AppState,
-  ContentRef,
-  DirectoryContentRecordProps,
-  DummyContentRecordProps,
-  FileContentRecordProps,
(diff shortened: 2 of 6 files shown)


CrystallineCat:
Hmm... I'm puzzled about the test error.

captainsafia:
@CrystallineCat Dug around with the error and might've figured out why it is happening.

[This](https://github.com/facebook/jest/issues/2567) recommends that the import of the dependency be moved to inside the mock, like so.

```
jest.mock("rx-jupyter", () => {
  const { of } = require("rxjs");
  return {
    contents: {
      JupyterContentProvider: {
        save: (severConfig, filepath, model) => {
          return of({ response: {} });
        },
        update: (serverConfig, prevFilePath, object) => {
          return of({ status: 200, response: {} });
        },
        get: jest
          .fn()
          .mockReturnValue(
            of({
              status: 200,
              response: { last_modified: "some_stable_value" }
            })
          )
          .mockReturnValueOnce(
            of({
              status: 200,
              response: { last_modified: "one_value" }
            })
          )
          .mockReturnValueOnce(
            of({
              status: 200,
              response: { last_modified: "one_value" }
            })
          )
      }
    }
  };
});
```

Also, I found an old `__mocks__` directory in `core` that was conflicting with this. AFAIK, this is no longer needed since all the epics make their own internal mocks. I removed this.

After making these changes, your PR tests ran locally for me.

Can you try it and see if you get the same result?

CrystallineCat:
@captainsafia Thank you very much for your help! This got the tests working.

However, deleting `core/__mocks__` resulted in another test error when a test tried to fetch kernelspecs. I thus reverted that part of the change. That test also used a very strange input that was supposedly valid `JupyterHostRecordProps`, but was missing several properties and had several extra. I fixed that (hopefully correctly, the individual fields don't have any documentation) and typed it.

captainsafia:
> That test also used a very strange input that was supposedly valid JupyterHostRecordProps, but was missing several properties and had several extra. I fixed that (hopefully correctly, the individual fields don't have any documentation) and typed it.

This would make sense. We have an open issue tracked in #3997 to migrate our tests to TypeScript so there's quite a few typing errors sprinkled around the tests.

In any case, glad we got over this hiccup!

packages/epics/src/contents.ts
```diff
@@ -46,20 +20,8 @@ export function updateContentEpic(
   return action$.pipe(
     ofType(actions.CHANGE_CONTENT_NAME),
     switchMap(action => {
-      if (!action.payload || typeof action.payload.filepath !== "string") {
```
captainsafia, line 49: Hm. Is there a reason we are removing this check?

packages/epics/__tests__/kernelspecs.spec.ts
```diff
@@ -16,16 +16,12 @@ describe("fetchKernelspecsEpic", () => {
         app: {
           host: {
             type: "jupyter",
-            kernelRef: null,
```
captainsafia, line 19: Thanks for removing these. These look like properties that used to be on the HostRecord object from a very old version of our SDK.

packages/epics/src/contents.ts
```diff
@@ -118,17 +80,6 @@ export function fetchContentEpic(
   return action$.pipe(
     ofType(actions.FETCH_CONTENT),
     switchMap(action => {
-      if (!action.payload || typeof action.payload.filepath !== "string") {
```
captainsafia, line 121: Ditto here. Are we anticipating that the `catchError` will grab it?

captainsafia (approved):
Looks good overall.

Posted some questions about the safeguards we removed on filepath.

packages/epics/src/contents.ts
CrystallineCat: In both of these, `action.payload` is guaranteed to be an object, and `action.payload.filepath` is guaranteed to be a string by the type checker. This lead to `tslint` telling me that the conditions are always false, so I removed the code as unreachable.

Of course, at runtime, types might not hold. However, in that case we probably have a major problem anyway, and I don't think the guards would improve things by much. That said, I can see an argument for leaving them in.

In general, it would be good to know how much we should trust the type checker. I have been reluctant to fix up the typing in test because many tests are designed to feed wrong parameters to a function, which a type checker would catch, and I'm not sure what to do with the test in that situation.

captainsafia: Ah fair point! I was mostly alarmed by this from a visibility standpoint. This error was good to have because it was explicit about what was wrong but agree that we should rely on the typechecker for this.

In general, I think there are a lot of code-smells lingering in our codebase from our TypeScript conversion. We can continue to clean them up in passes as we come along.

CrystallineCat:
Thanks for reviewing and merging, @captainsafia!

PR #4928, opened 2020-02-14
Clean up auto-save epic

victor-meng:
In [this commit](https://github.com/nteract/nteract/commit/fac12544f9befa1cf3dc17e0b3d6bd9c453700bf), the line `const state = state$.value` is moved outside of the `pipe` operator. This means that we are no longer getting the latest value of state which will cause problems once multi-tab is supported. The state will not be updated when we open/close tabs so auto-save will not work properly.

The fix is to simply move the line back inside `pipe` so we are always getting the latest value of state.

### DIFF
--- a/packages/epics/__tests__/contents.spec.ts
+++ b/packages/epics/__tests__/contents.spec.ts
@@ -1,13 +1,37 @@
 import { stringifyNotebook } from "@nteract/commutable";
-import { actions, AppState, ContentRecord, createContentRef, createKernelspecsRef, makeAppRecord, makeCommsRecord, makeContentsRecord, makeDummyContentRecord, makeEntitiesRecord, makeHostsRecord, makeJupyterHostRecord, makeNotebookContentRecord, makeStateRecord, makeTransformsRecord } from "@nteract/core";
+import {
+  actions,
+  AppState,
+  ContentRecord,
+  createContentRef,
+  createKernelspecsRef,
+  makeAppRecord,
+  makeCommsRecord,
+  makeContentsRecord,
+  makeDummyContentRecord,
+  makeEntitiesRecord,
+  makeHostsRecord,
+  makeJupyterHostRecord,
+  makeNotebookContentRecord,
+  makeStateRecord,
+  makeTransformsRecord
+} from "@nteract/core";
 import { fixtureJSON, mockAppState } from "@nteract/fixtures";
 import FileSaver from "file-saver";
 import * as Immutable from "immutable";
 import { ActionsObservable, StateObservable } from "redux-observable";
 import { contents } from "rx-jupyter";
 import { of, Subject } from "rxjs";
-import { map, toArray } from "rxjs/operators";
-import { closeNotebookEpic, downloadString, fetchContentEpic, saveAsContentEpic, saveContentEpic, updateContentEpic } from "../src/contents";
+import { map, take, toArray } from "rxjs/operators";
+import {
+  autoSaveCurrentContentEpic,
+  closeNotebookEpic,
+  downloadString,
+  fetchContentEpic,
+  saveAsContentEpic,
+  saveContentEpic,
+  updateContentEpic
+} from "../src/contents";
 
 jest.mock("rx-jupyter", () => {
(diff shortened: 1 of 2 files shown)


captainsafia:
@victor-meng Thanks for catching this bug and submitting a PR!

This motivated me to write unit tests for the `autoSaveContentEpic`. It turned out to be rather cumbersome because this is the first instance of testing a never-ending Observable in our codebase but I finally got unit tests for it. Based on the unit tests, I pushed a few more changes to this epic.

* I used `switchMap` in favor of `concatMap` so that we only send a `SAVE` request for the most recent change that occurred within an interval instead of all
* I updated the distinct check to find the model using the selector instead of the `pluck` operator so that changes in metadata fields are processed as part of the flow
* Removed the `existsSync` check since the content provider support means that we can speak to a remote content source when using a local kernel

While doing the validation, I spotted another issue related to this bug report. The `duration` extracted at the start of the epic will always be set to the default when the configuration is loaded from the desktop app. Also, for applications that allow users to change their auto-save duration, this value won't be respected.

I think we'll have to tweak how this interval is used. I'll see if I can bundle that in this PR.

In the meantime, I've pushed the changes in the bulleted list above and more unit tests to the code.

captainsafia:
Alright! Did some research and pushed a commit that reads the latest interval from the StateObservable. If it changes, it uses a `switchMap` to update the interval in the stream with the new value.

packages/epics/src/contents.ts
```diff
@@ -153,27 +172,16 @@ export function autoSaveCurrentContentEpic(
...
-          "contents",
-          "byRef",
-          contentRef,
-          "model",
-          "notebook"
-        ),
+        map(state => selectors.model(state, { contentRef })),
         distinctUntilChanged(),
-        concatMap(() => of(actions.save({ contentRef })))
+        switchMap(() => of(actions.save({ contentRef })))
```
jdetle, line 184: Can you explain the necessity of refactoring from concatMap to switchMap? Is this taking the latest save action rather than combining all the saves?

jdetle: _answered in thread, oops_

packages/epics/src/contents.ts
```diff
@@ -153,27 +172,16 @@ export function autoSaveCurrentContentEpic(
              */
             content =>
               (content.type === "file" || content.type === "notebook") &&
-              content.filepath !== "" &&
-              selectors.isCurrentKernelZeroMQ(state)
```
jdetle, line 157: why is this code getting deleted? does anyone know what it was doing?

captainsafia: Yes, this was a guard that was checking to see if the file that we were about to save already existed on disk. It operated under the assumption that if you were using a local kernel, you were getting contents from the local filesystem as well.

However, with new improvements in our error notification handling and a content provider model that gives users the flexibility to fetch contents from different places regardless of where their kernel is, the check doesn't make sense.

The consequence is that auto-saves will error with a message for the user but I think this is fine given that the assumptions the code was making aren't accurate anymore.

jdetle (approved):
Left questions and tested locally, but everything related to this pr works great thanks @victor-meng and @captainsafia

captainsafia:
> Left questions and tested locally, but everything related to this pr works great thanks @victor-meng and @captainsafia

Alright! Let's :shipit: . Thanks to everyone for participating in this PR.

packages/epics/__tests__/contents.spec.ts
```diff
@@ -390,3 +418,97 @@ describe("updateContentEpic", () => {
...
+          kernels: {}
+        }
+      }
+    };
+    const state$ = new StateObservable(stateSubject$, state);
+    stateSubject$.next(state);
+    const results = await autoSaveCurrentContentEpic(action$, state$)
+      .pipe(take(2), toArray())
+      .toPromise();
+    expect(results.map(a => a.type)).toContain(actions.SAVE);
```
victor-meng, line 512: Should we still expect a save action even when there's no change in state?

PR #4930, opened 2020-02-14
Spruce up execution epics

captainsafia:
@jayphelps and I met yesterday and walked through the execution epics. Following his feedback, I cleaned up the execution epics in the epics package. The changes are as outlined:

- Splits up the monolithic execute.ts file into multiple files
- Applies the same splitting to the test files
- Removes `map` and `mergeAll` in `sendExecuteRequest` epic in favor of `mergeMap`
- Move all error checks in the core execution epics to the top-level `sendExecuteRequestEpic`
- Remove resubscribing to source in the `catchError` for `createExecuteCellStream`
- Use `cellStream$.pipe(startsWith()` to clear outputs and queue cells before executing
- Replaces all uses of `empty()` with `EMPTY`
- Adds documentation to execution-related epics
- Adds CellId to ExecuteFailed action where applicable for improved diagnostics
- Adds a set of error codes for identifying different types of execution and kernel launch related issues

BTW, @CrystallineCat I think we should update lineWidth config in the prettier file to avoid repeatedly undoing each others formatting. What line-width configuration do you have?


packages/epics/__tests__/execute/execute.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+    sent.subscribe(() => {
+      kernelMsgs.map(msg => received.next(msg));
+    });
+
+    const obs = executeCellStream(channels, "0", message, "fakeContentRef");
+
+    const emittedActions = [];
+    obs.subscribe(action => {
+      emittedActions.push(action);
+    });
```
jayphelps, line 81: You might consider using the TestScheduler and the auto-run `testScheduler.run(callback)` so that time is virtualized and all your tests are deterministic and fast.

I see that there are [tests already in nteract that use it](https://github.com/nteract/nteract/blob/d0fd5ba0cef3080b4ee1a095c9d5cdc739339863/applications/desktop/__tests__/renderer/epics/close-notebook.spec.ts#L201) which is great.

I don't immediately know if any of the tests you're including truly would benefit from virtualized time or not, so it's possible they're all already synchronous though I do see the usage of `done()` signifying potentially async code.

That said, writing marble diagrams and using TestScheduler is not easy. There's the official guide here: https://rxjs-dev.firebaseapp.com/guide/testing/marble-testing but it's still somewhat lacking.

Also important if you google for it, in RxJS v6 we added the that `testScheduler.run(callback)` thing so that it would automatically use the TestScheduler instance in all operators without you needing to do your own dependency injection crap. Nearly all tutorials don't reference this new way of doing it, and there are also some subtle but important differences in the semantics of the diagrams. In March I'm gonna be doing a talk on this at [RxJS Live Conf](https://www.rxjs.live/), hopefully it'll be helpful for folks.

packages/epics/src/execute/lazy-launch.ts
```diff
@@ -0,0 +1,147 @@
...
+        of(
+          ...selectors
+            .messageQueue(state)
+            .map((queuedAction: AnyAction) =>
+              actions.executeCell(queuedAction.payload)
+            )
+        ),
+        of(actions.clearMessageQueue())
+      );
+    })
```
jayphelps: `merge(of(...items), of(item))` could be done as `of(...items, item)`

packages/epics/src/execute/lazy-launch.ts
```diff
@@ -0,0 +1,147 @@
...
+        actions.launchKernelByName({
+          kernelSpecName,
+          cwd,
+          kernelRef,
+          selectNextKernel: true,
+          contentRef
+        })
+      );
+    }),
+    distinct(
```
jayphelps, line 125: FYI distinct keeps a record of all previously distinct items, in this case every previous distinct action's contentRef. Worth mentioning in case this could be unintentional and a memory leak.

packages/epics/__tests__/execute/execute.spec.ts
```diff
@@ -0,0 +1,364 @@
...
+import { catchError, share, toArray } from "rxjs/operators";
+
+import {
+  createExecuteCellStream,
+  executeCellEpic,
+  executeCellStream,
+  sendExecuteRequestEpic
+} from "../../src/execute";
+
+const Immutable = require("immutable");
```
jayphelps: Curious why this has to be CJS vs ESM?

packages/epics/__tests__/execute/lazy-launch.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+            })
+          }),
+          kernels: stateModule.makeKernelsRecord(),
+          messages: stateModule.makeMessagesRecord()
+        }),
+        kernelRef: "fake"
+      })
+    };
+    const state$ = new StateObservable(new Subject(), state);
+    const action$ = ActionsObservable.of(
```
jayphelps, line 156: If your code isn't using the prototype based `.ofType()` operator and only uses the pipe version, I would suggest not using ActionsObservable since it's going to be removed in v2 of redux-observable and it buys you nothing else. Though I'm not sure if your epics will have type errors if you don't give them an ActionsObservable...that might be the case, in which just ignore my suggestion for now

packages/epics/src/execute/lazy-launch.ts
```diff
@@ -0,0 +1,147 @@
...
+  state$: StateObservable<AppState>
+) {
+  return action$.pipe(
+    ofType(actions.EXECUTE_CELL),
+    filter((action: actions.ExecuteCell) => {
+      const state = state$.value;
+      const contentRef = action.payload.contentRef;
+      return !selectors.kernelByContentRef(state, { contentRef });
+    }),
+    mergeMap<
```
jayphelps: Does this need to be a mergeMap or do all code paths inside project to a single `of(action)` so this could just be a map?

packages/epics/src/execute/map-to-execute.ts
```diff
@@ -0,0 +1,141 @@
...
+) {
+  return action$.pipe(
+    ofType(actions.EXECUTE_FOCUSED_CELL),
+    mergeMap((action: actions.ExecuteFocusedCell) => {
+      const contentRef = action.payload.contentRef;
+      const state = state$.value;
+      const model = selectors.model(state, { contentRef });
+      // If it's not a notebook, we shouldn't be here
+      if (!model || model.type !== "notebook") {
+        return EMPTY;
```
jayphelps: Probably should be an error

packages/epics/src/execute/lazy-launch.ts
```diff
@@ -0,0 +1,147 @@
...
+      const contentRef = action.payload.contentRef;
+      const kernel = selectors.kernelByContentRef(state, { contentRef });
+      return !!(
+        kernel &&
+        kernel.channels &&
+        kernel.status !== KernelStatus.NotConnected &&
+        kernel.status !== KernelStatus.ShuttingDown
+      );
+    }),
+    concatMap(([, state]) => {
```
jayphelps: There's a few epics in this PR with 1-to-many actions. Since they are synchronous it doesn't matter if they use mergeMap, switchMap, concatMap, exhaustMap, etc but I recommend picking one as a team and being consistent. Here it is concatMap, below it is mergeMap

jayphelps: As a very general comment, epics which map actions to other actions without performing any side effects is sometimes a code smell, though not always. I'm not familiar enough with all the moving parts to know in this case.

The questions I ask myself are:

- should whatever this epic be doing done where I dispatched the original action? Or is  this the right place to do
it?
- can whoever is listening for action I'm mapping to instead listen for the original action or is it better to have them separate?

Again, general spider-sense and not a rule. And this is very much just one person's opinion.

jayphelps: What do you think about still giving the first param a name, and just underscore it so linters are okay with it? `concatMap(([_action, state]) => {`

The reason why is that I have indeed hit bugs that were caused by people not seeing the comma `[,` because it's one of those things thats easy to miss unless you already know its there, then you can't unsee it.

packages/epics/__tests__/execute/execute.spec.ts
captainsafia: Yeah, we use quite a few different Observable testing patterns in our codebase. I've always been a bit uneasy about this because it would be great to standardize on one thing so its easier to write new tests.

But speaking from experience, it's really hard to test Observables sometimes so I usually end up copying an existing pattern that works and modifying it to fit the test scenario.

I'll resolve this for now but will keep this in mind if we ever get the chance to do an overhaul on our tests here.

packages/epics/src/execute/lazy-launch.ts
captainsafia: The use here is intentional. For context, we want to launch a kernel associated for each unique content that is registered in the user's store.

Is there a recommended pattern for "run this epic for each unique instance of this"?

Assuming it only keeps unique instances of contentRef, this will grow with each new file that the user opens.

packages/epics/__tests__/execute/execute.spec.ts
captainsafia: No reason. I assume it lingered here from another commit. I cleaned it up.

packages/epics/src/execute/lazy-launch.ts
captainsafia: Sounds reasonable!

packages/epics/__tests__/execute/lazy-launch.spec.ts
captainsafia: I think I'll keep it to ActionsObservable for now, but we can run through and update everything once we upgrade redux-observable.

packages/epics/src/execute/lazy-launch.ts
captainsafia: This led me down a typing wormhole so I decided to leave it with the `merge` for now.

packages/epics/src/execute/lazy-launch.ts
captainsafia: I ended up standardizing this entire file to use `switchMap` per your other recommendation.

Using `map` sent me down a TypeScript wormhole so I decided to save my sanity for the type being. Maybe the next version of redux-observable will improve the experience here.

rgbkrk:
Thanks for the review @jayphelps!

PR #4931, opened 2020-02-15
Replace <li> in dropdown menu with buttons

barshana-banerjee:
Issue solve : All the `<li>` below that have role button should just be `<button>` #4499

Changed list to button objects and added css styles


captainsafia (approved):
It works well locally! Thanks.

PR #4937, opened 2020-02-18
Fix IContent definition for "directory" type

gitanuj:
According to current definition of `IContent<"directory">` the content property is of type `Array<IEmptyContent<"directory">>`. However this is not correct and the directory content can contain any type of file. So this change updates the content property type to `Array<IEmptyContent<FileType>>` instead.

### DIFF
--- a/packages/types/src/content-provider.ts
+++ b/packages/types/src/content-provider.ts
@@ -31,7 +31,7 @@ export interface IContent<FT extends FileType = FileType>
     : FT extends "notebook"
     ? Notebook
     : FT extends "directory"
-    ? Array<IEmptyContent<FT>>
+    ? Array<IEmptyContent<FileType>>
     : null;
 }
 


packages/types/src/content-provider.ts
```diff
@@ -31,7 +31,7 @@ export interface IContent<FT extends FileType = FileType>
     : FT extends "notebook"
     ? Notebook
     : FT extends "directory"
-    ? Array<IEmptyContent<FT>>
+    ? Array<IEmptyContent<FileType>>
```
captainsafia, line 34: Thanks for catching this!

Do you notice any  other places where we have the incorrect type?

captainsafia (approved):
Looks good overall. Left one comment inline.

packages/types/src/content-provider.ts
gitanuj: This was the only place where I found incorrect type. Found this only while writing a content provider which creates content array for a "directory" type.

PR #4938, opened 2020-02-20
[testMigration] Fix unhandleRejection in test-shim.ts

ramantehlan:
##  Before
DeepinScreenshot_select-area_20200220093314

## After
DeepinScreenshot_select-area_20200220092815

### DIFF
--- a/packages/mathjax/__tests__/provider.spec.tsx
+++ b/packages/mathjax/__tests__/provider.spec.tsx
@@ -4,8 +4,8 @@ import React from "react";
 import MathJaxContext from "../src/context";
 import MathJaxProvider from "../src/provider";
 
-describe("MathJaxProvider", () => {
-  it("renders without crashing", () => {
+describe("MathJaxProvider", () : void => {
+  it("renders without crashing", () :void => {
     const component = shallow(<MathJaxProvider>$x^2 + y = 3$</MathJaxProvider>);
     expect(component.find(MathJaxContext.Provider)).toHaveLength(1);
   });
--- a/packages/mathjax/__tests__/provider.spec.tsx
+++ b/packages/mathjax/__tests__/provider.spec.tsx
@@ -4,8 +4,8 @@ import React from "react";
 import MathJaxContext from "../src/context";
 import MathJaxProvider from "../src/provider";
 
-describe("MathJaxProvider", () : void => {
-  it("renders without crashing", () :void => {
+describe("MathJaxProvider", () => {
+  it("renders without crashing", () => {
     const component = shallow(<MathJaxProvider>$x^2 + y = 3$</MathJaxProvider>);
     expect(component.find(MathJaxContext.Provider)).toHaveLength(1);
   });
--- a/scripts/test-shim.ts
+++ b/scripts/test-shim.ts
@@ -38,8 +38,8 @@ const createContextualFragment = (html: string) => {
 Range.prototype.createContextualFragment = (html: string) =>
   createContextualFragment(html);
 
-  globalAny.window.URL.createObjectURL = function(): void {};
-  globalAny.window.focus = () => {};
+globalAny.window.URL.createObjectURL = function(): void {};
+globalAny.window.focus = () => {};
 
 // HACK: Polyfill that allows codemirror to render in a JSDOM env.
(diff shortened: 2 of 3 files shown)


scripts/test-shim.ts
```diff
@@ -36,11 +38,11 @@ const createContextualFragment = (html: string) => {
 Range.prototype.createContextualFragment = (html: string) =>
   createContextualFragment(html);
 
-global.window.URL.createObjectURL = function(): void {};
-global.window.focus = () => {};
+  globalAny.window.URL.createObjectURL = function(): void {};
```
captainsafia: Nit: Let's remove the indentation here.

scripts/test-shim.ts
```diff
@@ -54,7 +56,7 @@ global.window.document.createRange = function createRange() {
 // HACK: To test index.js
 document.querySelector = () => document.createElement("div");
 
-process.on("unhandledRejection", (error, promise) => {
+process.on("unhandledRejection", (error:any, promise:Promise<any>):void => {
```
captainsafia: Does this work if you type it as `error: any` or perhaps `Error & { stack: string }`? The `any` type isn't adding much value here.

Also for the `promise` type, what value is TypeScript resolving for it?

packages/mathjax/__tests__/provider.spec.tsx
```diff
@@ -4,8 +4,8 @@ import React from "react";
 import MathJaxContext from "../src/context";
 import MathJaxProvider from "../src/provider";
 
-describe("MathJaxProvider", () => {
-  it("renders without crashing", () => {
+describe("MathJaxProvider", () : void => {
```
captainsafia: I don't think we need to add this type here. TypeScript should be able to resolve it automagically. What happens if you remove it?

captainsafia (commented):
Left some comments inline.

packages/mathjax/__tests__/provider.spec.tsx
ramantehlan: Nothing happens, It's safe to remove.

scripts/test-shim.ts
ramantehlan: According to the docs(https://nodejs.org/api/process.html#process_event_unhandledrejection),  `error` aka `reason` can be of type `Error` or `any`, but when we type it as `Error` or don't type it, it throws the following error, only when I type it as `any` there is no error. 

DeepinScreenshot_select-area_20200220093314

For `promise`, there is no error if I type it or not. Should I remove it?

captainsafia: Yeah, I would remove it and let the type inference do the trick.

scripts/test-shim.ts
```diff
@@ -8,20 +8,22 @@
  * does not support them or has only a partial implementation.
  */
 
+const globalAny:any = global;
```
captainsafia: can we add an interface to type this? Typing it as `any` isn't pushing the needle on improving our types.

scripts/test-shim.ts
```diff
@@ -54,7 +56,7 @@ global.window.document.createRange = function createRange() {
 // HACK: To test index.js
 document.querySelector = () => document.createElement("div");
 
-process.on("unhandledRejection", (error, promise) => {
+process.on("unhandledRejection", (error:any, promise) => {
```
captainsafia, line 57: Does typing this as `Error & { stack: string }` not work?

captainsafia (commented):
Left some more comments inline. Ideally, we should be strictly typing things instead of relying on any.

scripts/test-shim.ts
ramantehlan: According to the docs, It can be of two types, `Error` and `any`, but it raises an error when typed `Error & { stack: string }` or just `Error`.

DeepinScreenshot_select-area_20200220084753


Ref: https://nodejs.org/api/process.html#process_event_unhandledrejection

scripts/test-shim.ts
ramantehlan: Done, interface already exist, but was missing some properties, I have added those properties.

captainsafia (approved):
Thanks for addressing the follow-ons!

PR #4942, opened 2020-02-21
Follow-on fixes for auto-save epic

captainsafia:
@victor-meng noticed one of the tests for the auto-save epics was improperly structured. This PR fixes that and a couple of other things:

* Auto-save now uses the `filter` operator and the `isDirty` selector. From the different patterns I tried, this struck the best balance between readability, performance, and testability.
* Auto-save epic tests now use fake-timers to simulate the passing of time that is caused by the `Observable.interval()` creator
* Fixed the notebook model created in the tests so that the isDirty selector would function as appropriate

In addition to the unit tests, I tested this on the desktop app with the following scenarios:
* auto-save every 1000 seconds (change content then don't change every other second and confirm the status bar updates)
* auto-save every 1000 seconds then change auto-save interval by dispatching `SET_CONFIG_AT_KEY` action and confirm everything still works


captainsafia:
Codecov, I just came here to fix bugs and have fun and I honestly feel so disrespected right now.

victor-meng (approved):
Thanks for fixing the tests! I love the idea of replacing `distinctUntilChanged` with `filter` + `isDirty` selector. Makes the code much easier to read and debug.

captainsafia:
@victor-meng Thanks for the review, Victor! Glad you like the new approach. Hopefully all these changes improve the experience for auto-save.

I added some tests to this file to appease Codecov and will merge after this check iteration.

PR #4946, opened 2020-02-22
[testMigration] Fix type in fs.ts 

ramantehlan:
## Before
DeepinScreenshot_select-area_20200222071637

DeepinScreenshot_select-area_20200222071709

## After
DeepinScreenshot_select-area_20200222071542


packages/fs-observable/__mocks__/fs.ts
```diff
@@ -1,4 +1,5 @@
-const fs = jest.genMockFromModule("fs");
+const fs:any = jest.genMockFromModule("fs");
```
captainsafia: Do we need this any type here?

packages/fs-observable/__tests__/fs.spec.ts
```diff
@@ -12,7 +12,7 @@ describe("unlinkObservable", () => {
     // File exists so we try to unlink it
     fs.existsSync.mockImplementation(() => true);
     // Unlink fails for some reason (permissions, etc.)
-    fs.unlink.mockImplementation((path, callback) =>
+    fs.unlink.mockImplementation((path: any, callback: (arg0: { message: string; }) => void) =>
```
captainsafia: It seems like path should be typed as a string, not any.

scripts/test-shim.ts
```diff
@@ -8,20 +8,22 @@
  * does not support them or has only a partial implementation.
  */
 
+const globalAny:any = global;
```
captainsafia: This PR contains changes from the other PR. You might want to remove the commit that brings in these changes from this PR or create a new branch off master for each new PR.

captainsafia (commented):
Left some comments inline.

packages/fs-observable/__mocks__/fs.ts
ramantehlan: without it, it raise the ` Object is of type 'unknown'.` error.

packages/fs-observable/__mocks__/fs.ts
```diff
@@ -1,4 +1,11 @@
-const fs = jest.genMockFromModule("fs");
+type Fs = {
```
ramantehlan, line 1: I know this is not the perfect solution, but I wasn't able to find any other solution.

captainsafia: This works for now.

captainsafia (approved):

captainsafia (approved):
Looks good! Thanks for making this fix!

Issue #4943, opened 2020-02-21
PPA please or Debian + Ubuntu repo


socialcyberdog:
I can not find an Ubuntu ppa or debian repo - it would be very welcomed to not have to manually update / download things. I am used to "apt update && apt upgrade" for too long, manual _nteraction_ for software updates feels so backwards - using Linux for a reason!

ramantehlan:
Thank for opening this issue @socialcyberdog! 

**nteract** recently published its first [**Snap**](https://snapcraft.io/nteract) to [Snapcraft](https://snapcraft.io), which should work with all the major Linux distros including Debian based distros. If you are new to Snaps, you can learn more about it from [here](https://snapcraft.io/docs/getting-started), and download it for your distro from [here](https://snapcraft.io/docs/installing-snapd). 

For Debian based distro:
```
sudo apt update
sudo apt install snapd
```

Once you have installed it, you can install `nteract` by:
```
sudo snap install nteract --edge
```

> Our **next release is on 30th March 2020** and since **Snaps are updated automatically**, you don't need to worry about updating it once downloaded.

willingc:
Thanks for the question @socialcyberdog and response on using snap @ramantehlan.

Issue #4913, opened 2020-02-07
arm64 debian package and/or appimage.


AdrianKoshka:
**Is your feature request related to a problem? Please describe.**
I'm a user of an arm64 chromebook, and I use crostini to run linux applications. Sadly, nteract doesn't have a debian package or appimage for arm64 systems.

**Describe the solution you'd like**
An arm64 debian package and/or appimage, for use on arm64 systems.

**Describe alternatives you've considered**
Not using nteract.

**Additional context**
N/A

captainsafia:
Thanks for opening this issue, @AdrianKoshka!

Supposedly, electron-builder can be [configured to support arm64](https://www.electron.build/cli) but we don't have this configuration enabled at the moment.

Would you be interested in submitting a contribution to add this support? It will require [editing the build configuration](https://github.com/nteract/nteract/blob/master/applications/desktop/package.json#L47) for the desktop app and validating that the nteract desktop app works on your OS.

Happy to help with the setup/testing for this contribution since I don't have an arm64 machine myself.

AdrianKoshka:
Sorry about accidentally closing the issue. How would one install/setup electron-builder?

captainsafia:
You'll have to fork and clone the repository per our CONTRIBUTING.md first. Once you've done that, you can install the dependency by running `yarn` in the root.

The `dist` command in the `package.json` for the desktop app runs the electron-builder command line using the configuration within that package.json.

[Here's a link to the package.json](https://github.com/nteract/nteract/blob/958b7bfef7820e2781124c5ea4641a150b64b484/applications/desktop/package.json).

Let me know if you have any follow-up questions.

ramantehlan:
@AdrianKoshka, this seems to be answered in the issue [4943](https://github.com/nteract/nteract/issues/4943) | [Comment-5920084248](https://github.com/nteract/nteract/issues/4943#issuecomment-592084248)
 
> **nteract** recently published its first [**Snap**](https://snapcraft.io/nteract) to [Snapcraft](https://snapcraft.io), which should work with all the major Linux distros including Debian based distros. If you are new to Snaps, you can learn more about it from [here](https://snapcraft.io/docs/getting-started), and download it for your distro from [here](https://snapcraft.io/docs/installing-snapd).
> 
> For Debian based distro:
> 
> ```
> sudo apt update
> sudo apt install snapd
> ```
> 
> Once you have installed it, you can install `nteract` by:
> 
> ```
[...]

PR #4947, opened 2020-02-22
[Snap packaging] add snap config to package.json 

ramantehlan:
This PR is to add support for `amd64` `.snap` and `arm64` `.deb`. 

Snap can be published as a [snap package](https://snapcraft.io/nteract), which will allow Linux and Mac users to install it using snap CLI. 

[Get it from the Snap Store](https://snapcraft.io/nteract)

Example:
DeepinScreenshot_select-area_20200223012013

This will address the issue  #4913 and #4943.


applications/desktop/package.json
```diff
@@ -80,10 +93,9 @@
     },
     "linux": {
       "maintainer": "nteract contributors <jupyter@googlegroups.com>",
+      "synopsis": "The nteract desktop application allows you to quickly view, edit, and publish notebooks from your desktop. It  is a great tool for first-time and veteran notebook users, alike.",
       "target": [
-        "deb",
-        "AppImage",
-        "tar.gz"
+        "snap"
```
ramantehlan: I have removed other target because I don't feel we need them after the snap package but if required, we can add them again.

applications/desktop/package.json
```diff
@@ -35,6 +35,7 @@
     "prebuild": "rimraf lib",
     "build": "webpack --config webpack.dev.js --progress --colors",
     "build:clean": "rimraf lib dist",
+    "build:lin": "electron-builder --linux snap",
```
BenRussert: I think we may be ok without adding a separate script for this, but if we do add one I think we should use something  like `dist:linux`. The term "`build`" in these scripts is generally for webpack/typescript.

applications/desktop/package.json
BenRussert: I see your point, but think we should leave them:

1. So we don't throw off users who are used to these install methods and may not know about snap or how to use it on their distro 

2. Appimage has some cool features like the ability to run the app from a single binary without system installing, ([it's also a community driven project :smile: ](https://github.com/AppImage/AppImageKit))

BenRussert (commented):
Wow, thanks! I left some comments below, but overall awesome!

BenRussert:
Once the snap is built, how does it get published to the snap repository? Can we use a shared nteract organization snapcraft account for this?

applications/desktop/package.json
ramantehlan: I agree we don't need a separate script for this, will remove this. :)

ramantehlan:
Once the snap is built, we can either automate the publishing with CI/CD or can publish it manually. To publish it, you will need to login with [snapcraft CLI](https://snapcraft.io/docs/snapcraft-overview), which use [Ubuntu one account](https://login.ubuntu.com/).
```
$ snapcraft login
$ snapcraft push --release=edge nteract_0.22.0-alpha.0_amd64.snap
```

`--release` here is edge because **nteract** doesn't have a stable version yet, but maintainers can decide which release(`stable`, `candidate`, `beta`, `edge`) they wanna pick. The biggest difference it will make is when a user will download it, he will need to explicitly mention the release if it's not stable.

--- 

Right now, I used my personal account to publish the snap which is published [here](https://snapcraft.io/nteract), but yes we should create a shared nteract organization account for it.

captainsafia:
@ramantehlan Thanks for working on this! I'll look into creating an official nteract org account on the Snap store.

Also, it looks like this is something that we will be able to ship into production as part of our March release.

ramantehlan:
Release Process, please correct if anything is missing or wrong. :) 

DeepinScreenshot_select-area_20200228025321

README.md
```diff
@@ -3,6 +3,9 @@
 <h1 align="center">the interactive computing suite for you</h1>
 
 <p align="center">
+<a href="https://snapcraft.io/nteract">
```
captainsafia: Nit: since this is specific to Linux builds, I would keep it out of the README.

A while back I did a clean-up of the badges on the README to reduce the clutter on our README so  I'd like us to keep it clean for the next while.

captainsafia (approved):
Looks good overall! I left one comment small nit in-line. Otherwise, this is good to go!

PR #4960, opened 2020-02-27
[Fix] Fix the outgrow of DragHandle

bryanchen-d:

**Problem**
The DragHandle area expand outside of the drag area, see screenshot below (the DragHandle is colored as red, and the cell's DragArea is colored in blue)
image

The cause is that DragHandle is set to use absolute position and 100% height, so it finds the nearest base - the DragArea, and set its height to be DragArea's height subtract border width. However the 10px padding on DragArea push down the DragHandle to outgrow the DragArea.

**Solution**
Add a layer of div called DragHandleAnchor with position set to relative. So DragHandle's position and height is based on it, DragHandleAnchor doesn't have padding so DragHandle won't outgrow it. 
After the fix:
image

### DIFF
--- a/packages/notebook-app-component/src/decorators/draggable/index.tsx
+++ b/packages/notebook-app-component/src/decorators/draggable/index.tsx
@@ -106,9 +106,13 @@ const DragArea = styled.div.attrs<DragAreaProps>(props => ({
         : "3px transparent solid"
   }
 }))`
-  position: relative;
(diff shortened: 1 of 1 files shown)


captainsafia (approved):
Thanks for submitting this fix, @bryanchen-d!

Merging to release for the 3/1 alpha release.

PR #4965, opened 2020-02-28
Auto-adds http prefix to www. links and doesn't hyperlink partial link matches

### LINKED ISSUE #4964: Links starting with "www" or surrounded in quotes are treated as relative links
ansi-to-react

**Describe the bug**
Links starting with "www" or surrounded in quotes are treated as relative links

**To Reproduce**

1. In code cell, type `print('www.google.com')`
2. Run cell
3. Click link
4. Observe URL is now http://example.com/www.google.com

1. In code cell, type 
```
c = 'http://www.google.com'
c
```
2. Run cell
3. Click link
4. Observe URL is http://example.com/'http://www.google.com'

miduncan:
Closes #4964

### DIFF
--- a/packages/ansi-to-react/src/index.ts
+++ b/packages/ansi-to-react/src/index.ts
@@ -2,7 +2,7 @@ import Anser, { AnserJsonEntry } from "anser";
 import { escapeCarriageReturn } from "escape-carriage";
 import * as React from "react";
 
-const LINK_REGEX = /(https?:\/\/(?:www\.|(?!www))[^\s.]+\.[^\s]{2,}|www\.[^\s]+\.[^\s]{2,})/;
+const LINK_REGEX = /^(https?:\/\/(?:www\.|(?!www))[^\s.]+\.[^\s]{2,}|www\.[^\s]+\.[^\s]{2,})$/;
 
 /**
  * Converts ANSI strings into JSON output.
@@ -15,7 +15,7 @@ const LINK_REGEX = /(https?:\/\/(?:www\.|(?!www))[^\s.]+\.[^\s]{2,}|www\.[^\s]+\
  */
 function ansiToJSON(
   input: string,
-  use_classes: boolean = false,
+  use_classes: boolean = false
 ): AnserJsonEntry[] {
   input = escapeCarriageReturn(input);
   return Anser.ansiToJson(input, {
@@ -89,7 +89,7 @@ function convertBundleIntoReact(
   linkify: boolean,
   useClasses: boolean,
   bundle: AnserJsonEntry,
-  key: number,
+  key: number
 ): JSX.Element {
   const style = useClasses ? null : createStyle(bundle);
   const className = useClasses ? createClass(bundle) : null;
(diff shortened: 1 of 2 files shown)


packages/ansi-to-react/src/index.ts
```diff
@@ -116,21 +117,20 @@ function convertBundleIntoReact(
         return words;
       }
 
+      const href = word.startsWith("www.") ? `http://${word}` : word;
```
captainsafia, line 122: This won't capture links like `google.com`. Can we invert the way this works and instead have  it check to see if `http://` is a substring of the word and add it to the beginning if not?

Also nit: Let's use `http://` instead of `https://`.

EDIT: The above should be be Let's use **`https://`** instead of **`http://`**.

captainsafia: Also, let's add a comment to explain why we are doing this,

packages/ansi-to-react/src/index.ts
```diff
@@ -116,21 +117,20 @@ function convertBundleIntoReact(
...
       words.push(
         React.createElement(
           "a",
           {
             key: index,
-            href: word,
+            href: href,
```
captainsafia: Nit: this can just be `href: href` instead of `href`.

captainsafia (commented):
Welp! My review of this was stuck in the draft.

I left some comments inline.

packages/ansi-to-react/src/index.ts
miduncan: I think that is out of scope of this PR, as we need a broader discussion of if those types of links (like `google.com`) should be supported. Currently they are not, the regular expression expects a `http` or `www` to start. Personally I think we shouldn't-- most things that try to help you out with linking don't support them, plus then we have to keep up with all of the domain endings and not false positive on things like "ASP.NET"

I'm assuming you meant use `https` since it already uses `http` right now. The problem is that we can't assume a website supports https, many don't.

I can definitely add a comment, though.

captainsafia: > I think that is out of scope of this PR, as we need a broader discussion of if those types of links (like google.com) should be supported. Currently they are not, the regular expression expects a http or www to start. Personally I think we shouldn't-- most things that try to help you out with linking don't support them, plus then we have to keep up with all of the domain endings and not false positive on things like "ASP.NET"

Seems reasonable.

> I'm assuming you meant use https since it already uses http right now. The problem is that we can't assume a website supports https, many don't.

I think the opposite is true. Considering most browsers will through a warning if a user tries to navigate to an unsecured domain, we should be prudent and redirect to HTTPS when possible.

miduncan: It will show an icon that it isn't secure, but the warning page is for when you go to an https link that doesn't have a trusted certificate so it needs to fall back to http.

For comparison, Microsoft Word and Outlook also only add http
image

Usually if the website requires https, the http link will redirect to https.

captainsafia (approved):

Issue #4956, opened 2020-02-26
The road to @nteract/web beta


captainsafia:
Following up on the work introduced in #4925, I wanted to draft a roadmap that outlined the goals and future of this project.

**Goals**
- Provide a web-based nteract experience that is freely available to folks at an nteract-based domain
- Improve the experience for front-end development on nteract as users no longer need to set up local Jupyter instances
- Eventually move towards a common app codebase for both the web and desktop app, enabled by frameworks like [nextron](https://github.com/saltyshiomix/nextron)
- Provide a flexible, compute protocol agnostic application that folks can build interactive computing apps on top off
- Deprecate the nteract Jupyter extension in favor of the nteract web app

**Tasks**
- [x] Implement sidebar component for browsing multiple files
- [x] Implement Binder config component for configuring target Binder instance
- [ ] Deploy beta of nteract web app to zeit now
- [ ] Implement KernelProvider model as outlined in #4902 
- [ ] Unify nteract web and desktop app code bases

willingc:
I've gone ahead and pinned this issue @captainsafia as a strategic issue :smile:

jabbera:
What does this mean for those of use who use nteract as part of a jupyterhub installation (specifically on a k8s cluster)? Will there be a similar offering?

captainsafia:
> What does this mean for those of use who use nteract as part of a jupyterhub installation (specifically on a k8s cluster)? Will there be a similar offering?

The existing published version of the nteract Jupyter extension will continue to work although it is no longer supported. The current plan is to consolidate nteract web and the desktop app. There's no plans to bring back the Jupyter extension although it should be possible to repurpose the existing code and the new model to get something working.

Let me know if you have any follow-up questions/

jabbera:
Thanks! Hopefully there is a comparable offering at some point. Hosted notebook environments that scale horizontally based on active users is key to our use case.

ramantehlan:
@captainsafia Do you think the nteract web beta is ready for release? 
I think we can release it and let the community start using it.

captainsafia:
@ramantehlan I think we probably want to do a round of testing/QA before that and also figure out some sort of CI for the deployment.

We can track that work in a separate issue. Would you be interested in creating it and identifying some scenarios to test?

ramantehlan:
@captainsafia Sure, I will create the issue for the testing and for deployment pipeline.

PR #4966, opened 2020-02-28
[nteract/web] Alpha webapp layout

ramantehlan:
This PR is for issue #4956.

Changes made so far by this PR.
- Implemented following components
     - Sidebar
     - BinderMenu
     - StatusLine
     - Console
- Implemented GitHub API to fetch the files and contents.

### Demo
final


ramantehlan:
@captainsafia, can you review this? Not able to add reviewers. :P

willingc:
@ramantehlan Thanks for the PR. I've gone ahead and requested a review from Safia.

captainsafia:
@ramantehlan Thanks for working on this! I'll spend some time taking a look at this over the weekend since there is quite a lot here.

applications/web/components/App.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  loadFile(fileName){
+    const octokit = new Octokit()
+    octokit.repos.getContents({
+      owner: this.state.org,
+      repo: this.state.repo,
+      path: fileName
+    }).then(({data}) => {
+      console.log(data)
+      this.setState({fileContent: atob(data["content"])})
```
captainsafia: When I was trying this locally, this line of code threw an exception when I clicked on a folder (not a file).

```
Unhandled Rejection (InvalidCharacterError): Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
```

applications/web/components/App.tsx
```diff
@@ -0,0 +1,236 @@
...
+            </FilesListing>
+            
+        </Side>
+        <Body>
+          {this.state.fileContent}
+          { this.state.contentRef ? (<NotebookApp contentRef={this.state.contentRef} />) : "Wating"}
+        </Body>
+        <Footer>
+          { this.state.showConsole && <Console></Console> }
+            <StatusLine toggleConsole={this.toggleConsole}></StatusLine>
```
captainsafia: I am unable to see this toggle in the footer when I launch the application. Do you need to do something special to enable it?

applications/web/components/App.tsx
```diff
@@ -0,0 +1,236 @@
...
+            <FilesListing 
+                  loadFile={this.loadFile} 
+                  org={this.state.org} 
+                  repo={this.state.repo}
+                  gitRef={this.state.gitRef}>
+            </FilesListing>
+            
+        </Side>
+        <Body>
+          {this.state.fileContent}
```
captainsafia: Do we want this here? It ends up causing the contents of the selected file to render on top of the notebook which seems unideal.

captainsafia (approved):
Looks good to move forward with this. I left some comments inline.

applications/web/components/App.tsx
ramantehlan: Github API returns an `object` when for file type and `array of objects` for folder type. I have added the logic to handle it differently for both the cases.

For file type, it stores it to `fileContent`, and for folder type, it should display the list of sub files and folders, which I will prefer to do in a separate PR. I have added a TODO in detail.

applications/web/components/App.tsx
ramantehlan: I have checked its working. I guess what happens is that when the `NotebookApp` is rendered, it overlaps the footer implemented in this PR. 

I am not sure how to fix it; I will have to look at how the `NotebookApp` footer is implemented.

applications/web/components/App.tsx
ramantehlan: Ideally, we will render the files only through Github API and only send the changes to Mybinder instance.
Rendering can have two cases:

- We want to render a notebook; It can be edited.
- We want to render any other file type; I will suggest initially we keep it read/view only, and maybe in the future, we can think of supporting the editing. It arguable, but for me, the reason for that is to keep the focus on the notebooks only and perhaps role the editing other files if required.

This again should be done in a separate PR; for now, I will only console it out and add a TODO for it.

ramantehlan: I have raised an issue for this: https://github.com/nteract/nteract/issues/5042

captainsafia (approved):

PR #4972, opened 2020-03-03
Fixed toggling output exanded

### LINKED ISSUE #4971: "Toggle Output Expanded" does not do anything
stateful-components
nteract Desktop

**Describe the bug**
"Toggle Output Expanded" does not do anything

**To Reproduce**

1. Run cell with following code
```
for i in range(0,50):
  print(i);
```
2. Observe in inspector (output has "expanded" class)
3. Toggle Output Expanded
4. Observe in inspector. Notice "expanded" class is still there. No change.

miduncan:
Closes #4971

### DIFF
--- a/packages/commutable/src/cells.ts
+++ b/packages/commutable/src/cells.ts
@@ -25,6 +25,7 @@ export const makeCodeCell = Record<CodeCellParams>({
   execution_count: null,
   metadata: ImmutableMap({
     collapsed: false,
+    outputExpanded: true,
     jupyter: ImmutableMap({
       source_hidden: false,
       outputs_hidden: false
--- a/packages/reducers/src/core/entities/contents/notebook.ts
+++ b/packages/reducers/src/core/entities/contents/notebook.ts
@@ -477,7 +477,7 @@ function unmarkCellAsDeleting(
 
 function deleteCellFromState(
   state: NotebookModel,
-  action: actionTypes.DeleteCell,
+  action: actionTypes.DeleteCell
 ): RecordOf<DocumentRecordProps> {
   const id = action.payload.id ? action.payload.id : state.cellFocused;
   if (!id) {
@@ -860,10 +860,15 @@ function toggleOutputExpansion(
   return state.updateIn(
     ["notebook", "cellMap"],
     (cells: Map<CellId, ImmutableCell>) =>
-      cells.setIn(
-        [id, "metadata", "outputExpanded"],
-        !cells.getIn([id, "metadata", "outputExpanded"])
-      )
+      cells
+        .setIn(
+          [id, "metadata", "collapsed"],
+          !cells.getIn([id, "metadata", "collapsed"])
+        )
+        .setIn(
+          [id, "metadata", "outputExpanded"],
+          !cells.getIn([id, "metadata", "outputExpanded"])
+        )
   );
 }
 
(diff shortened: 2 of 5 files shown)


miduncan:
Any ideas on how to get the code coverage bot to run again? It seems to have not updated.

captainsafia:
> Any ideas on how to get the code coverage bot to run again? It seems to have not updated.

Hmmmm...something must've gotten stuck in the part of our build configuration that passes coverage stats generated in the CircleCI test step to the Codecov system. I will see if re-running the build for this PR will resolve this issue.

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -477,7 +477,7 @@ function unmarkCellAsDeleting(
 
 function deleteCellFromState(
   state: NotebookModel,
-  action: actionTypes.DeleteCell,
+  action: actionTypes.DeleteCell
```
captainsafia, line 480: Let's add a test for these reducer changes as well.

packages/stateful-components/src/outputs/index.tsx
```diff
@@ -58,7 +58,7 @@ export const makeMapStateToProps = (
         expanded =
           cell.cell_type === "code" &&
           (cell.getIn(["metadata", "outputExpanded"]) ||
-            !cell.getIn(["metadata", "collapsed"]));
+            cell.getIn(["metadata", "collapsed"]) === false);
```
captainsafia: Nit: Since the `metadata.collapsed` property is now supported in the notebook format, let's move that one first.

packages/commutable/src/cells.ts
```diff
@@ -25,6 +25,7 @@ export const makeCodeCell = Record<CodeCellParams>({
   execution_count: null,
   metadata: ImmutableMap({
     collapsed: false,
+    outputExpanded: true,
```
captainsafia, line 28: Since `metadata.collapsed` is the officially supported attribute for toggling expansion for the notebook format, let's set that in commutable instead of the `outputExpanded` property.

This will set us up to remove support for the `outputExpanded` property and instead rely on the `metadata.collapsed` property.

We went through a similar transition for the `jupyter.source_hidden` and `jupyter.outputs_hidden` metadata properties listed below.

captainsafia (approved):
No glaring issues worthy of a block.

I left some comments about adding more tests and setting us up to deprecate `outputExpanded` in favor of the official `metadata.collapsed` property.

Thanks for submitting this change.

packages/commutable/src/cells.ts
miduncan: If we are going to look at both fields (I assume for compatibility), I think we really should be setting both fields. Otherwise we set ourselves up for bugs where `outputExpanded` is true but our button just toggles `collapsed`, in which case the cell will always be expanded because it looks at both.

I think the best move eventually is to read the both values, but if it is `outputExpanded` to delete that field and replace it with `expanded`. I think that work should come when we want to deprecate, though, not this.

packages/reducers/src/core/entities/contents/notebook.ts
miduncan: I'll update the existing one. Since it doesn't add any new logic or code paths, though, I don't think a new test would increase any coverage here or provide any new information.

willingc:
Ping @captainsafia. I tried to merge this and reran the tests but the PR will not let me override and merge.

willingc (approved):

PR #4983, opened 2020-03-10
Don't pass kernelRefGenerator via dependency injection

### LINKED ISSUE #4952: Error "n is not a function" when trying to restart kernel
nteract desktop Windows

**Describe the bug**
When trying to restart the kernel (Python 3) I get a notification Error n is not a function. And the kernel does not seem to restart

**To Reproduce**
1. Start a Python notebook
2. execute some commands
3. Menu Runtime, choose Restart

**Expected behavior**
Restarting the Kernel

**Screenshots**
https://i.vgy.me/ZdEJma.png

- OS: Windows
- Browser -
- Version 0.22.0

**Additional context**
If you kill the kernel there seems to be no way to restart it (without restarting nteract or reopening the file)

captainsafia:
Closes #4952 

Now that we pass the contentProvider in via dependency injection, the value of the kernelRefGenerator passed in to the restartEpic is overriden by the dependency injection set in the root epic.

This PR removes the `kernelRefGenerator` and instead relies on the `createKernelRef` method directly.

### DIFF
--- a/packages/epics/__tests__/kernel-lifecycle.spec.ts
+++ b/packages/epics/__tests__/kernel-lifecycle.spec.ts
@@ -566,9 +566,6 @@ describe("restartKernelEpic", () => {
     });
   });
   test("emits no action for remote kernel", async () => {
-    const contentRef = "contentRef";
-    const newKernelRef = "newKernelRef";
-
     const state = {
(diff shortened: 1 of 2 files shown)


willingc:
LGTM. Thanks @captainsafia

PR #4984, opened 2020-03-11
[testMigration] Add missing properties in <Manager> test

ramantehlan:
For issue #3997. This is to resolve the test error in `manager` test.

### Before

DeepinScreenshot_select-area_20200311062835

### After 

DeepinScreenshot_select-area_20200311063424


captainsafia (approved):
Thanks for getting this in! Let's continue chipping away at this.

Issue #4587, opened 2019-10-02
Smarter selection of current working directory


nathancarter:
**Is your feature request related to a problem? Please describe.**
In the nteract.app on Mac OS X (version 0.15.0), I'm always frustrated when I do File > New > Python 3 then File > Save As, but the Python runtime in the notebook is somehow in my home folder.  I never asked it to initialize its cwd to my home folder, I can't imagine a situation in which I would want that.  The only way to fix it is to close my new file and reopen it again from where I saved it.  Even Runtime > Restart and Run All Cells doesn't update the Python runtime to the correct folder.

**Describe the solution you'd like**
Two things:
 1. When saving a freshly created notebook for the first time, and no cells have ever been run, immediately change its Python runtime's working directory to the directory in which the notebook was just saved.
 2. When restarting the kernel, update its current working directory to the directory in which the notebook is saved.  (I thought #919 fixed this, but I guess not?)

**Describe alternatives you've considered**
Show cwd in status bar and let user click there to change it?

**Additional context**
Imagine also that someone would do a lot of work in the notebook, creating and saving files, not realizing that the output folder were their home folder.  Later, they re-open the notebook and it behaves differently, because now it has a different working directory than the last time they had it open.  An entirely different pain point from this same problem.  Also, #386 seems to address a similar issue, but it did not describe how it resolved it.

captainsafia:
Thanks for opening this issue, @nathancarter!

Triage note: I think this is a dupe of an existing issue, other than #386 mentioned above, but cannot find it at the moment.

> When saving a freshly created notebook for the first time, and no cells have ever been run, immediately change its Python runtime's working directory to the directory in which the notebook was just saved.

When you launch the notebook that was just saved again, is the cwd set to the correct path?

> When restarting the kernel, update its current working directory to the directory in which the notebook is saved. (I thought #919 fixed this, but I guess not?)

The change in #919 doesn't affect the restart path so this is still a relevant issue.

I'm a little hesitant to take the approach in #1 as it requires us to implement state-dependent logic.

#2 should be resolved though and will likely need to be done in conjunction with #4575.

nathancarter:
Hi, @captainsafia -- thanks for the quick reply!

> When you launch the notebook that was just saved again, is the cwd set to the correct path?

The cwd is correct only if you close the notebook entirely in nteract, then re-open it; in that case it is correct.  Just saving the file for the first time does not correctly set the cwd.

[...]

Issue #721, opened 2016-09-11
[Request for Design] New or renamed file should ask about launching new kernel on save


rgbkrk:
Right now if we rename a file (or save a new one in some location), the kernel is using the previous working directory. We likely want to launch a brand new kernel.

How should we give users a choice?

captainsafia:
I think using a dialog box should do the trick.

```
It looks like you've saved your notebook file to a new location. The kernel 
executing your code thinks your notebook is still in the old location. Would 
you like to relaunch the kernel to synchronize it with the new location 
of your notebook?

[Don't Relaunch Kernel]     [Relaunch Kernel]
```

What are your thoughts on the explanatory text in this case?

rgbkrk:
I like that text!

rgbkrk:
Maybe swap:

> relaunch the kernel to synchronize it with the new location of your notebook?

with

> launch a new kernel to synchronize it with the new location of your notebook?

captainsafia:
> launch a new kernel to synchronize it with the new location of your notebook?

Good point! Especially since it is not actually relaunching a kernel.

rgbkrk:
Aiming for simple language, another revision:

> launch a new kernel to match it with the location of your notebook?

rgbkrk:
> launch a new kernel to match the directory with that of your notebook?

captainsafia:
I think "launch a new kernel to match it with the **new** location of your notebook?" works best.

rgbkrk:
ok awesome, that looks great

captainsafia:
Great! PR inbound!

PR #4989, opened 2020-03-11
Restore Save-As kernel prompt

### LINKED ISSUE #4968: New document remains as Untitled after a "Save As"
nteract desktop

**Describe the bug**
When starting nteract or when creating new file, after "saving as" with a proper name and in a specific folder, the file in nteract remains the "Untitled" file. Subsequent saves do not save in the "saved as file"

**To Reproduce**

1. Create new File (I used Python)
2. Save as... (enter a valid filename)
3. The file in the window title remains unchanged (Untitled.ipynb)

**Expected behavior**
To edit the file under the new file name. Subsequent saves should store the content under the "Saved as" filename

- OS: Windows
- Browser -
- Version 0.22.0

nathancarter:

This PR does the following things:
 - Issue #4587 asks for a feature that was added way back in #721 but that somehow disappeared in the intervening work; this PR puts that feature back in.
 - That feature had previously used a popup dialog, but @captainsafia said to use `sendNotification()` this time instead, so that's a small upgrade.
 - In order to do that stuff, it was also necessary to fix #4968, so this PR fixes that one, too.
 - The code in `applications/desktop/src/notebook/menu.tsx` had some vestigial stuff that needed to be cleaned out; I did that while I was in there.

### DIFF
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -460,8 +460,7 @@ export function saveAsContentEpic(
       const host = selectors.currentHost(state) as JupyterHostRecord;
       const serverConfig = selectors.serverConfig(host);
 
-      const kernelRef = selectors.kernelRefByContentRef(
-        store.getState(), { contentRef } );
+      const kernelRef = selectors.kernelRefByContentRef( state, { contentRef } );
       const kernel = selectors.kernel( state, { kernelRef } );
       if ( kernel ) {
         const cwd = filepath
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -460,11 +460,17 @@ export function saveAsContentEpic(
       return dependencies.contentProvider
         .save(serverConfig, filepath, saveModel)
         .pipe(
-          map((xhr: AjaxResponse) => {
-            return actions.saveAsFulfilled({
-              contentRef: action.payload.contentRef,
-              model: xhr.response
-            });
+          mergeMap((xhr: AjaxResponse) => {
+            return of(
+              actions.changeFilename({
+                contentRef: action.payload.contentRef,
+                filepath
+              }),
+              actions.saveAsFulfilled({
+                contentRef: action.payload.contentRef,
+                model: xhr.response
+              })
+            );
           }),
           catchError((error: Error) =>
             of(
--- a/packages/epics/src/contents.ts
+++ b/packages/epics/src/contents.ts
@@ -13,6 +13,8 @@ import {
   NotebookContentRecordProps,
   ServerConfig
 } from "@nteract/types";
+import { sendNotification } from "@nteract/mythic-notifications";
+import { createKernelRef } from "@nteract/core";
 import FileSaver from "file-saver";
 import { RecordOf } from "immutable";
 import { Action } from "redux";
@@ -30,6 +32,7 @@ import {
   tap
 } from "rxjs/operators";
 import urljoin from "url-join";
+import * as path from "path";
 
 export function updateContentEpic(
   action$: ActionsObservable<actions.ChangeContentName>,
@@ -456,6 +459,42 @@ export function saveAsContentEpic(
 
       const host = selectors.currentHost(state) as JupyterHostRecord;
       const serverConfig = selectors.serverConfig(host);
+
+      const kernelRef = selectors.kernelRefByContentRef(
+        store.getState(), { contentRef } );
+      const kernel = selectors.kernel( state, { kernelRef } );
+      if ( kernel ) {
+        const cwd = filepath
+          ? path.dirname(path.resolve(filepath))
+          : getDocumentDirectory();
+        if ( cwd != kernel.cwd ) {
+          console.log( 'must prompt; different:', cwd, kernel.cwd );
+          store.dispatch(sendNotification.create({
+            title: "Notebook folder changed",
+            message:
+              "The kernel executing your code thinks your notebook is still " +
+              "in the old folder. Would you like to launch a new kernel in " +
+              "the new folder?",
+            level: "warning",
+            action: {
+              label: "Launch new kernel",
+              callback(): void {
+                store.dispatch(
+                  actions.launchKernelByName({
+                    kernelSpecName: kernel.kernelSpecName,
+                    cwd,
+                    kernelRef: createKernelRef(),
+                    selectNextKernel: true,
+                    contentRef
+                  })
+                );
+              }
+            }
+          }));
+        } else {
+          console.log( 'must not prompt; same:', cwd, kernel.cwd );
+        }
+      }
 
       return dependencies.contentProvider
         .save(serverConfig, filepath, saveModel)
--- a/applications/desktop/__tests__/renderer/menu.spec.ts
+++ b/applications/desktop/__tests__/renderer/menu.spec.ts
@@ -722,35 +722,6 @@ describe("initMenuHandlers", () => {
   });
 });
 
-describe("triggerWindowRefresh", () => {
-  test("does nothing if no filename is given", () => {
-    const store = {
-      dispatch: jest.fn()
-    };
-
-    expect(menu.triggerWindowRefresh(store, null)).toBeUndefined();
-  });
-  test("sends a SAVE_AS action if given filename", () => {
-    const props = {
-      contentRef: "123"
-    };
-
-    const store = {
-      dispatch: jest.fn()
-    };
-    const filepath = "dummy-nb.ipynb";
-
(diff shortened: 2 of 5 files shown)


nathancarter:
I think I'm going to need some help with the compile failures.  I did `yarn build:desktop` just fine of course while testing, but `yarn build:all` fails with errors on files that work fine in `yarn build:desktop`.  Being new to webpack and nteract both, I think I'll need some tips.

packages/epics/src/contents.ts
```diff
@@ -457,14 +460,55 @@ export function saveAsContentEpic(
...
       const serverConfig = selectors.serverConfig(host);
 
+      const kernelRef = selectors.kernelRefByContentRef( state, { contentRef } );
+      const kernel = selectors.kernel( state, { kernelRef } );
+      if ( kernel ) {
+        const cwd = filepath
+          ? path.dirname(path.resolve(filepath))
+          : getDocumentDirectory();
+        if ( cwd != kernel.cwd ) {
+          console.log( 'must prompt; different:', cwd, kernel.cwd );
```
captainsafia, line 470: Let's remove this console.log from the commit. These epics are used in both the nteract apps and other applications as well so we want to avoid extraneous logs.

packages/epics/src/contents.ts
```diff
@@ -457,14 +460,55 @@ export function saveAsContentEpic(
...
+          store.dispatch(sendNotification.create({
```
captainsafia, line 471: There's no need to call `store.dispatch` here. Simply returning `of(sendNotification.create()` should do the trick. The redux-observable middleware that is responsible for processing this function will take care of dispatching the action.

packages/epics/src/contents.ts
```diff
@@ -457,14 +460,55 @@ export function saveAsContentEpic(
...
+                    kernelRef: createKernelRef(),
+                    selectNextKernel: true,
+                    contentRef
+                  })
+                );
+              }
+            }
+          }));
+        } else {
+          console.log( 'must not prompt; same:', cwd, kernel.cwd );
```
captainsafia, line 494: We can remove this else statement and the `console.log` since we want to avoid extraneous logs.

packages/epics/src/contents.ts
```diff
@@ -457,14 +460,55 @@ export function saveAsContentEpic(
       const host = selectors.currentHost(state) as JupyterHostRecord;
       const serverConfig = selectors.serverConfig(host);
 
+      const kernelRef = selectors.kernelRefByContentRef( state, { contentRef } );
+      const kernel = selectors.kernel( state, { kernelRef } );
```
captainsafia, line 464: A type-error is thrown here because kernelRef can be null. We should add a guard here to avoid passing a null value to the `selectors.kernel` function.

```js
if (kernelRef) {
  kernel = selectors.kernel(state, { kernelRef })
} else {
  return EMPTY;
}
```

captainsafia (commented):
Thanks for submitting this PR, @nathancarter!

The issue that is causing the build to fail is in the TypeScript build step of our code, usually indicating type errors in the code. I have left comments inline where these type errors can be addressed.

packages/epics/src/contents.ts
nathancarter: Whoops!  That was obviously not supposed to stay in.

packages/epics/src/contents.ts
nathancarter: Yeah that was a mistake too.

packages/epics/src/contents.ts
nathancarter: I can't actually return at that point because the function isn't complete...should I just `sendNotification()`?

packages/epics/src/contents.ts
nathancarter: I can't return there, but I can wrap it in a guard, which does fix that type complaint, as you say.

packages/epics/src/contents.ts
nathancarter: Tried that...it doesn't actually send the notification if you don't pass it to `store.dispatch`.

nathancarter:
My main question was about the build errors regarding importing of packages.  It complains that it can't find packages like `@nteract/core` and so forth, even though it has no trouble finding them when building the app...I think I just don't know the build structure enough to know what's going on, sorry.

captainsafia:
> My main question was about the build errors regarding importing of packages. It complains that it can't find packages like @nteract/core and so forth, even though it has no trouble finding them when building the app...I think I just don't know the build structure enough to know what's going on, sorry.

The `build:desktop:watch` step does not do any type-checking, but the `build:all` step does. If there is a type error in one package (like @nteract/epics) that package won't be built. As a result, any packages that depend on it (like @nteract/core) will also not be built. The consequence is that any imports of this package will fail to resolve because the package (@nteract/core) was not actually built.

packages/epics/src/contents.ts
```diff
@@ -30,6 +32,7 @@ import {
   tap
 } from "rxjs/operators";
 import urljoin from "url-join";
+import * as path from "path";
```
captainsafia, line 35: Noticing this now: We'll need to use an alternate package that is compatible with both Node and browser scenarios. This package is intended to be used in Node (aka Electron or server-side) and in the browser. This import will fail downstream if folks are using it from the browser. As an alternative, we can use [path-browserify](https://www.npmjs.com/package/path-browserify) here so that this package will continue to work in browsers and on Node.

packages/epics/src/contents.ts
captainsafia: > I can't actually return at that point because the function isn't complete...should I just sendNotification()?

There's a few different patterns for this. The simplest is to return the `sendNotification.create` action alongside the `actions.changeFileName` and `actions.saveAsFulfilled` actions below.

For context, I suspect that doing `store.dispatch(sendNotifications.create(...))` in the desktop app works because `store` is bound to the global window object within the desktop app. However, this is not the case for folks using nteract apps in the browser or as a package so this would cause a null-ref in their code.

packages/epics/src/contents.ts
nathancarter: Is the suggestion therefore that I should `npm install path-browserify`?  I just want to be sure I don't add a new dependency if I'm not supposed to.

I was imitating the import that's already part of [`menu.tsx` here at this line](https://github.com/nteract/nteract/blob/master/applications/desktop/src/notebook/menu.tsx#L8), but I guess this is the node/browser distinction you're talking about.  Is `menu.tsx` for server-side only and `contents.ts` for both?

captainsafia: > I was imitating the import that's already part of menu.tsx here at this line, but I guess this is the node/browser distinction you're talking about. Is menu.tsx for server-side only and contents.ts for both?

Correct. `menu.tsx` is specifically for the desktop app so you can import Node packages there.

> Is the suggestion therefore that I should npm install path-browserify? I just want to be sure I don't add a new dependency if I'm not supposed to.

Correct and close. Since we have a monorepo setup you'll actually need to run the following command.

```
yarn workspace @nteract/epics add path-browserify
```

It basically says, add `path-browserify` as a dependency of the package `@nteract/epics` which is associated with the `packages/epics` directory in the monorepo.

nathancarter:
OK I've got a lot of build errors here and my process seems to go like this:
 1. Guess and see if it helps.
 1. If it doesn't help, ask here and do whatever the experts say.  (So far that's one expert. :) )
 1. Repeat with next error.

This is not very efficient for any of us, especially since sometimes the fixes expose new things I did wrong because I just don't know the enormous webpack/epic/myth/yarn/electron/web/etc. situation.  I think it's no longer productive for us to continue this workflow.  I think someone familiar with nteract's internals should take the reorganization I've done and tweak all the edges until it builds and passes the tests.

captainsafia:
> This is not very efficient for any of us, especially since sometimes the fixes expose new things I did wrong because I just don't know the enormous webpack/epic/myth/yarn/electron/web/etc. situation. I think it's no longer productive for us to continue this workflow. I think someone familiar with nteract's internals should take the reorganization I've done and tweak all the edges until it builds and passes the tests.

Totally understand. Thanks for getting the ball rolling on this. Can you give me access to your fork of the nteract repo? It'll allow me to push commits to the branch so we'll be able to track authorship for the parts you did.

nathancarter:
Thank you for understanding!  I've invited you as a collaborator.

(I also start online teaching this week so I don't have as many free cycles as I did when I started tinkering here!)

I look forward to poking my head in later and seeing what you've changed, so I know how I *should* have done it. :)

captainsafia:
Closing in favor of #5020

Issue #4981, opened 2020-03-10
RFC: Saying goodbye to the monorepo


captainsafia:
Fair warning: I'm penning this issue in mild frustration after working through an nteract pre-release (#4980) and confronting the not-so-good aspects of our monorepo.

To put it simply: our monorepo is too gosh darn big (for our maintenance capacity and scenario).

The monorepo in this repo currently contains quite a few packages. Some UI focused, some core SDK focused, some transforms, and some odds-and-ends that seem to be there largely as a matter of convience.

While the monorepo makes the development process on the desktop and web apps easy, it's a huge burden when it comes to releases, users finding and exploring code, and new contributors being able to make contributions to the project.

I'm proposing that we break up the monorepo we have in this GH repo. I've mentioned it before, but I'd like to outline the seperation that I think should happen.

* Transforms and outputs should be moved to their own monorepo under nteract/outputs
* Component packages should be moved to their own monorepo under nteract/components
* CodeMirror and Monaco editors should be moved to their own monorepo under nteract/editors
* commutable should be its own package
* The packages under @nteract/core can remain under this repo in addition to the myths
* Generally useful packages like styled-blueprintjsx and ansi-to-react should be their own repos
* fs-observable should be its own package

Overall, I hope that this split will make it easier to release, contribute to, and understand our package ecosystem.

One thing we do lose with this splitting is the ability to do integration tests within the desktop app more seamlessly. For this, I think we should lean more heavily on unit tests for validations.

Also, there is a cost to pay up-front with setting up the ops for each new repo.

Any other pros/cons I am missing? Overall thoughts?

cc: @rgbkrk @ramantehlan @CrystallineCat @willingc

EDIT: Here is a checklist of least-frequently modified/well-isolated packages that I am working on moving out of the monorepo.

- [x] styled-blueprintjsx
- [x] octicons
- [x] logos
- [x] ansi-to-react
- [x] directory-listing
- [x] dropdown-menu
- [x] fs-observable
- [x] ironicons
- [x] mathjax

rgbkrk:
The biggest pain point is likely commutable and `@nteract/core`, since they rely on the same data structures.

I agree with splitting out some of the super clear-cut packages like `ansi-to-react` and `styled-blueprintjsx`. Same thing with `@nteract/outputs` and the transforms.

ramantehlan:
[...]

PR #4993, opened 2020-03-13
[WIP] Move more packages out of monorepo

captainsafia:
Moving out more of the packages outline in #4981.

I also pinned TypeScript to 3.7.3 since 3.8 contained some configuration changes that affect the `noEmit` option we had configured for the nteract desktop app build.


willingc:
When I run locally I'm getting a bunch of import errors for the tests. Perhaps due to the movements out of the monorepo.

willingc (commented):
Looks like the imports are causing CircleCI to hang and timeout too

captainsafia:
> Looks like the imports are causing CircleCI to hang and timeout too

I saw the same as well. Been working on figuring out why and will push more commits once I am done.

In the meantime, I'll add the WIP label to this to let folks know that it is still in progress.

captainsafia:
This is getting rather bothersome so I'm closing this in favor of PRs migrating each of the packages one by one to make it easier to debug.

PR #4996, opened 2020-03-15
[WIP] Add Autocompletion for IJulia

willingc:
Addresses #4415


captainsafia:
@willingc What else do we need to do so that this PR is mergeable?

willingc:
@captainsafia I think that we need to fix a wider issue with autocomplete. It wasn't working with the Python kernel.

captainsafia:
@willingc Do you get completions when you type Ctrl-Space in the editor? Typeahead completions were temporarily moved a while back since a lot of users found them arbitrary and annoying and we didn't have a configurability layer for it.

willingc:
Closing since the solution isn't correct and the rebase mucked things up.

Issue #4902, opened 2020-01-31
RFC: Integrate IKernelProvider into codebase


captainsafia:
Inspired by the work @gitanuj and @languy have brought over to the repo with the concent of the IContentProvider, I'd like to introduce the concept of an IKernelProvider. This RFC outlines the problem and why I believe that IKernelProvider is the right solution to implement.

**Problem**
Currently, code for handling fundamental interactions with the kernel (like interrupt and restart) is distributed across different epics within @nteract/epics and the nteract desktop app. Furthermore, the API interface across the two epics is inconsistent. For example, a different error message text might be returned from each epic or one epic might handle some errors that the other doesn't.

Finally, there is a lot of differences between the ways that 0mq-backed kernels and WebSocket-backed kernels handle their statuses. A unified IKernelProvider interface would clean up the amount of 

**Solution**
Since all kernels have the same fundamental interactions, we need an abstraction layer over these interactions so that we don't have to implement too much duplicative logic and so that the interface for both actions is the same across nteract desktop and web clients.

I propose that we implement an IKernelProvider interface, which can be injected as a dependency into the epic middleware, in the same fashion as the IContentProvider.

**Implementation**

The IKernelProvider should have the following interface:

```js
interface IKernelProvider {
    interrupt: (kernelRef) => InterruptKernelActions;
    shutdown: (kernelRef) => ShutdownKernelActions;
    start: (kernelRef?) => StartKernelActions; // And creates the KernelRecord
    restart: (kernelRef) => RestartKernelActions;
    get: (kernelRef) => FetchKernelActions;
    list: (hostRef) => ListKernelActions;
}
```

In implementing this IKernelProvider model, we will have to migrate our current 0mq and Websocket-based kernels to this model.

0mq-based kernels will build on top of enchannel-zmq-backend for all their interactions.

WebSocket-based kernels will use the `sessions` API for all interactions listed above (except interrupt and restart).

**Questions**
- Did I miss any necessary kernel actions in the interface above?
- Are there any noticeable differences between 0mq and Websocket-based kernels that would make this abstraction difficult to implement.
- Thoughts overall?

cc: @languy @gitanuj @rgbkrk

rgbkrk:
I think the only extra difficulty is that websockets have fairly different reconnection and error logic from the zeromq sockets. That can probably be abstracted within here though.

gitanuj:
[...]

PR #4998, opened 2020-03-16
Expand comm epic doc

willingc:
In preparation for #4902

### DIFF
--- a/packages/epics/docs/comms.md
+++ b/packages/epics/docs/comms.md
@@ -1,15 +1,34 @@
 # Comm Epics
 
-Comm epics are epics that support interacting with the Jupyter Messaging Protocol via comm messages. Comm messages are an arbitrary data exchange format built on top of the Jupyter messaging protocol.
+Comm epics, more formally Communication Epics, are epics that support interacting with the
+[Jupyter Messaging Protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html)
+via comm messages.
 
-A comm 
+A few years ago, Jupyter added a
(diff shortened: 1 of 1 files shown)


packages/epics/docs/comms.md
```diff
@@ -1,15 +1,34 @@
...
+Comm epics, more formally Communication Epics, are epics that support interacting with the
+[Jupyter Messaging Protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html)
+via comm messages.
 
-A comm 
+A few years ago, Jupyter added a
+[custom messaging](https://jupyter-client.readthedocs.io/en/stable/messaging.html#custom-messages)
+system for developers to add their own objects with Frontend and Kernel-side components,
+and allow them to communicate with each other. To do this, IPython adds a notion of a `Comm`,
+which exists on both sides (Frontend and Kernerl), and can communicate in either direction.
```
captainsafia:
```suggestion
which exists on both sides (Frontend and Kernel), and can communicate in either direction.
```

captainsafia: Nit: Let's use "front end" for the noun and "front-end" for the adjective per [the content guide](https://content-guide.nteract.io/#/word-list).

captainsafia (approved):
Looks good overall! I left some comments and suggestions inline.

packages/epics/docs/comms.md
willingc: That's so cool that there is a content guide.

Issue #2780, opened 2018-04-18
Add new .ipynb file by right clicking on the folder directory


qzcool:
I am not sure if this should be a core functionality of Jupyter working with OS. It is convenient and user-friendly for new users who are used to create new files by right-clicking.

Ankk98:
@qzcool 
I would like to work on this functionality.
Can you guide me on this?

I think we need to work separately for windows, osx, various Linux based os.

I would like to start with ubuntu 18lts.

Ankk98:
https://askubuntu.com/questions/88480/adding-extra-options-to-right-click-menu#88485

But how can we add a functionality so that such a menu is added when we install nteract in ubuntu?
Also, there can be different desktop environments, different versions of operating systems have different features for e.g. https://itsfoss.com/add-new-document-option/

I think this can be done easily on windows as it has this functionality inbuilt.

jameskoch:
Take a look at [Electron](https://electronjs.org/); most of nteract's desktop integration features come from there.

jibraniqbal666:
I want to work on this, please tell me if someone is not working on it.

captainsafia:
Hello @jibraniqbal666! You're definitely welcome to work on it as a pull request for this effort has not been opened yet.

To get started, you can follow the instructions in our [contributing guide](https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md) to set up a local instance of nteract for development.

captainsafia:
We should still tackle this.

Because of OS limitations, this is something that we can only do on the Windows context menu. Not sure if it is feasible on Linux but have verified that macOS doesn't support this feature.

To get this working in Windows, we'll need to add some logic that registers metadata to the Windows registry on app launch.

PR #5005, opened 2020-03-17
Add "create new notebook" context menu for Linux

kshitijsubedi:

For Linux System Only ;
Tested on Nautilus and Nemo File Manager (Working)

Work on Issue ( https://github.com/nteract/nteract/issues/2780 )


applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,45 @@
...
+
+var fs = require('fs-extra');
+
+var home = process.env['HOME'];
+var tdir = '/Templates';
+var fileName = '/Templates/New_ipython_file.ipynb';
+var path = home+fileName;
+
+// json data
+var jsonData = '{"cells": [{"cell_type": "code","execution_count": null,"metadata": {},"outputs": [],"source": []}],"metadata": {"kernelspec": {"display_name": "Python 3","language": "python","name": "python3"}},"nbformat": 4,"nbformat_minor": 2}';
```
captainsafia: We can use the commutable back to create the new notebook here. You can read the code/documentation for it in the packages/commutable directory?

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,45 @@
+//  new .ipnb file option in Right Click Menu on Linux
```
captainsafia: Can you prettify this file so the formatting looks right?

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { dialog } from "electron";
+
+var fs = require('fs-extra');
+
+var home = process.env['HOME'];
+var tdir = '/Templates';
```
captainsafia: Can we move the template to the ~/.jupyter directory under ~/.jupyter/nteract-templates?

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,45 @@
...
+    );
+  } else{
+      if (!fs.existsSync(home+tdir)){
+	  fs.mkdirSync(home+tdir);
+	  }
+	else
+	{
+		console.log(" Templates directory Exists !!");
+	}
+	console.log(path);
```
captainsafia: Nit: We can remove the console.log here.

captainsafia (commented):
Left some comments inline.

Also, looks like the TypeScript build is failing. Let me know if you need help with it.

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: ok doing it

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: my bad , forgot to remove after testing

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: File Managers checks the template file from either '/home/user/Templates' folder or '/usr/share/templates' so the ipynb template is placed on home's Templates directory.

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: I was unaware of the commutable my bad :smile: I will try it asap

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: If we want to place template on ~/.jupyter folder, we have to create a symlink on the Templates folder anyway. Thought that will be double work for same thing.

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: made new commit with prettified code

applications/desktop/src/main/menu.ts
```diff
@@ -104,6 +104,11 @@ const helpDraft = {
   label: "Help",
   role: "help",
   submenu: [
+    {
+      label:"Create new .ipynb to system menu",
```
captainsafia: Nit: I would update this label to say "Add context menu item" or something like that.

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
+import { dialog } from "electron";
+
+var fs = require("fs-extra");
+
+var home = process.env["HOME"];
+var tdir = "/Templates";
+var fileName = "/Templates/New_ipython_file.ipynb";
+var path = home + fileName;
```
captainsafia: Let's use `path.join` here to join the segments together instead of string concatenation.

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
...
+    dialog.showErrorBox(
+      "This works only for Linux as of now .",
+      "Other Platform will be supported in future"
+    );
+  } else {
+    if (!fs.existsSync(home + tdir)) {
+      fs.mkdirSync(home + tdir);
+    }
+
+    fs.outputFile(path, jsonContent, (err:any) => {
```
captainsafia: Nit: Can we properly type the error instead of using any?

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (err) return console.error(err);
+      dialog.showMessageBox({
+        title: "Successfully installed.",
+        message: "You may now be able to create new .ipynb file from anywhere.",
```
captainsafia: Nit: "You can now create new notebook files from anywhere."

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
+import { dialog } from "electron";
+
+var fs = require("fs-extra");
+
+var home = process.env["HOME"];
+var tdir = "/Templates";
+var fileName = "/Templates/New_ipython_file.ipynb";
```
captainsafia: Nit: Let's call the file "jupyter-notebook.ipynb" since IPython is an older term.

captainsafia (commented):
Thanks for addressing the feedback. Left a few more comments inline. Looking forward to getting this in.

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: Done the changes

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: I still couldn't figure out how to use commutable on this specific code.

captainsafia: Take a look at how this is done in the [desktop app](https://github.com/nteract/nteract/blob/master/applications/desktop/src/notebook/epics/loading.ts#L95). Does that help?

applications/desktop/src/main/newfile-entry.ts
captainsafia: Got it! This might be worth a comment in the code...

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: That helped a lot. took me a while to understand but its done now

applications/desktop/src/main/menu.ts
```diff
@@ -104,6 +104,11 @@ const helpDraft = {
   label: "Help",
   role: "help",
   submenu: [
+    {
+      label:"Add Context Menu Item",
+      click:() => addRightClickMenu()
+
```
captainsafia: Super nit: Remove extraneous whitespace here.

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
+import { dialog } from "electron";
+import {
+  monocellNotebook,
+  toJS
+} from "@nteract/commutable";
+
+let fs = require("fs-extra");
```
captainsafia: Why not import using ES6 imports here?

```
import fs from "fs-extra";
import path from "path";
```

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
...
+let path = require("path");
+
+let home = process.env["HOME"];
```
captainsafia: Since none of these variables are re-assigned to, they can all be `const x = ...`.

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
...
+let fileName = "/Templates/jupyter-notebook.ipynb";
+let lpath = path.join(home + fileName);
+let xx=toJS(monocellNotebook);
```
captainsafia: Nit: You can probably make this `const jsonContent = JSON.stringify(toJS(monocellNotebook))`.

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
...
+let jsonContent = JSON.stringify(xx);
+
+export const addRightClickMenu = () => {
+  if (process.platform === "win32") {
```
captainsafia: We'll want this check to work on macOS too so we will probably want to check that the platform is not Linux or update this to check if the platform is macOS or Windows.

applications/desktop/src/main/newfile-entry.ts
```diff
@@ -0,0 +1,36 @@
...
+    dialog.showErrorBox(
+      "This works only for Linux as of now .",
```
captainsafia: Nit: "This feature is only supported on Linux. Other platforms will be supported in the future."

captainsafia (approved):
Looks good! Left a few more comments inline that should be addressed before merging.

applications/desktop/src/main/newfile-entry.ts
kshitijsubedi: fixed all suggested changes .

captainsafia (approved):
Looks good!

PR #5007, opened 2020-03-18
Move fs-observable out of monorepo

captainsafia:
Part of #4981.

The issues that caused the tests to timeout turned out to be in the migrated package so I've fixed them there and released a new version.


willingc (approved):
LGTM. 

Tried to rerun the tests. They are still timing out.

captainsafia:
@willingc At some point, I had imagined that this was working locally and pushed. Wrong was I!

It's fixed now (for real) this time. I moved one test that was mapping Observables to promises to use the RxJS test scheduler to properly simulate time in the tests.

PR #5008, opened 2020-03-18
Migrate enchannel-zmq-backend out of monorepo

captainsafia:
Part of #4981


willingc (approved):
Looks good. One question: Are the mocks still needed by the other tests?

captainsafia:
> One question: Are the mocks still needed by the other tests?

They are still needed for the zeromq kernel tests in the desktop app.

PR #5010, opened 2020-03-18
Cells created using commutable have outputs collapsed by default

miduncan:
This changes the cell creator so that it has cell outputs collapsed by default


captainsafia (approved):
Thanks for fixing the broken build on this PR.

Is there any other work that needs to be done for this collapse-support work?

miduncan:
Not that I know of, this should be the last of it. Thanks!

Issue #5013, opened 2020-03-21
Opening a notebook in Windows desktop app errors out


captainsafia:
**Application or Package Used**
nteract desktop app

**Describe the bug**
Opening a notebook in the desktop app on Windows causes an error with the following stacktrace.

```
Uncaught TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
    at subscribeTo (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/util/subscribeTo.js?:39)
    at subscribeToResult (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/util/subscribeToResult.js?:20)
    at SwitchMapSubscriber._innerSub (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/operators/switchMap.js?:59)
    at SwitchMapSubscriber._next (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/operators/switchMap.js?:49)
    at SwitchMapSubscriber.Subscriber.next (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/Subscriber.js?:63)
    at FilterSubscriber._next (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/operators/filter.js?:42)
    at FilterSubscriber.Subscriber.next (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/Subscriber.js?:63)
    at Notification.observe (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/Notification.js?:27)
    at QueueAction.ObserveOnSubscriber.dispatch (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/operators/observeOn.js?:48)
    at QueueAction.AsyncAction._execute (webpack:///C:/Users/safia/github.com/nteract/nteract/node_modules/rxjs/_esm5/internal/scheduler/AsyncAction.js?:68)
```

**To Reproduce**
1. Clone repository and follow setup steps.
2. Run `yarn start`.
3. Open any example notebook.
4. Observe the error above appears in the console.

**Expected behavior**
Notebooks should be able to open.

[...]

PR #5018, opened 2020-03-25
Move enchannel and fs-observable back into monorepo (temp)

captainsafia:
Resolution for #5013 so that we can get a production release out next week.

As it turns out, there seems to be an issue with requiring packages that have rxjs as a dependency. I'm not sure what I did wrong with moving contents out of the monorepo since I made sure that:

- rxjs was a peer dependency in the package
- rxjs was installed as a dev dependency (tried it with rxjs in production dependencies as well)

Which is a pattern similar to what other packages that depend on rxjs, like [redux-observable](https://github.com/redux-observable/redux-observable/blob/master/package.json#L67).

There's probably a way to make this work but I'm putting this in temporarily to unblock this month's release.

We'll see if we can get around to a more permanent solution for the March release. Otherwise, this will need to come in for the April release.


willingc (approved):
Makes sense to me. Good job on tracking this down.

Issue #4799, opened 2019-12-25
Add a preferences window


CrystallineCat:
Follow up of #4728.

This involves the following steps:
- [x] Factor out and generalize the config-related things into a new package.
- [ ] Factor out and generalize the windowing-related things into a new package.
- [x] Create a schema with all config options in `desktop`.
- [x] Add capability to watch files to `fs-observable` (#4800).
- [x] Change the config system to watch the config file and load/save config options from the schema.
- [x] Figure out the UI guidelines for preferences windows on Win/Mac/Linux.
- [ ] Add a new package that renders a preferences window and tie it into the menu.

CrystallineCat:
UI guidelines on preferences:
- [MacOS](https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/preferences/)
- [Windows](https://docs.microsoft.com/en-us/windows/uwp/design/app-settings/guidelines-for-app-settings)
- [Gnome](https://developer.apple.com/design/human-interface-guidelines/macos/app-architecture/preferences/)'s pretty unhelpfully state that we shouldn't have preferences, and just "[m]ake an effort to ensure that your application design works for everybody without the need to change its settings."

captainsafia:
Taking a look at this now! I'm looking forward to improving the abstractions we have around config management.

I was recently discussing how the core SDK should encourage developers to place all config in the nteract config state as opposed to creating additional fields in redux.

I think creating a separate package for managing config should help with this.

With regard to the options window, I think it would be good if the config window made sense in both a web and desktop context so that we can easily port it to the web app when needed.

Overall, your plan seems sensible. Let me know if you need help with anything.

PR #5031, opened 2020-04-01
Centralize all configuration related stuff into a mythic-configuration package, add a preferences menu and refactor the menu code

CrystallineCat:
Progress on #4799. Partial fix for #963.

TFW you break a task into small pieces, and then make the pieces huge by scope-creep refactoring... :joy_cat:

**Important:** I ran into a bug in `redux-observable`, so I upgraded to `latest` (`1.0.0` -> `2.0.0-alpha.0`, [Changelog](https://github.com/redux-observable/redux-observable/blob/master/CHANGELOG.md)) for a fix. This means **ActionsObservable is gone**, it is just Observable now.

### This PR does all of the following:
- Removes all configuration-related code everywhere.
- Make configuration state private.
- Creates a mythic package `mythic-configuration`, with a filesystem backend that monitors a config file and dispatches actions on detecting change.
- `mythic-configuration` has an interface for declaring configuration options. Doing so returns a selector that can be used to get the value for the option from a state object.
- The options that are declared are also stored and those that have enumerable values are displayed in a preferences menu.
- To implement this, I completely rewrote the menu code. The menu is now purely declarative, commands are reified as objects, and there is now a list of accelerators.
- There is now a config setting for custom accelerators.

(Screencast of the above features.)

### Still TODO:
- [ ] Tests
- [ ] Docs
- [ ] Giant rebase

### Out of scope for this PR:
- Adding file locking. I couldn't make sense of the available options and am not confident at this time that I'll not produce deadlocks. At the moment, if we read the file in the middle of a write, we likely get a syntax error, which we just ignore.
- Implementing a way to store the config in applications/web (using browser storage?).


CrystallineCat:
Ok, this is ready to be reviewed. I'm going to wait with rebasing until #5032 is in.

CrystallineCat:
Ok, I think I should close this and break it into smaller pieces, and submit new PRs for each piece...

Issue #4974, opened 2020-03-04
autocomplete brakets?


microwolf:
**Is your feature request related to a problem? Please describe.**
Maybe I'm not setting this up right? I found it very inconvenient to not have the autocomplete of (), [], {}, etc. while using nteract.

captainsafia:
Thanks for opening this issue, @microwolf!

The CodeMirror editor in nteract is currently not configured with the `closeBrackets` option enabled. If you are interested in adding support for this, I'd be happy to guide you through the process for contributing the change.

Tangent: Between this and other CodeMirror-config related requests that we get, we should consider adding a `codemirrorConfig` property to nteract's configuration so that users can set the codemirror configuration themselves.

captainsafia:
@CindyPotvin is taking a look at this!

CindyPotvin:
I'm working first on moving the existing `cursorBlinkRate` to a `codemirrorConfig` section so I have a place to put the new configurations. 
@captainsafia, do I need to be retrocompatibility with the old `cursorBlinkRate` config? I don't think it's that big of a deal to lose that configuration for users, and it would make that part of the code more complex for a single case.

captainsafia:
> @captainsafia, do I need to be retrocompatibility with the old cursorBlinkRate config? I don't think it's that big of a deal to lose that configuration for users, and it would make that part of the code more complex for a single case.

Nope. Since we have a UI for configuring the cursorBlinkRate, we don't have to retrocompat here.

captainsafia:
@CindyPotvin How is this going?

CindyPotvin:
@captainsafia I have the new configurations on my branch, including the one to auto close brackets. A few of them are only applied when starting a new code block (or maybe restarting the editor, didn't check yet) so I'm taking a look at this before sending the pull request.

captainsafia:
> A few of them are only applied when starting a new code block (or maybe restarting the editor, didn't check yet) so I'm taking a look at this before sending the pull request.

Sounds good! Feel free to ping on Slack if you need help with this.

captainsafia:
This, and support for other CodeMirror options, will ship in tomorrow's April release!

PR #5032, opened 2020-04-02
Expand support for CodeMirror configuration

CindyPotvin:

Issue #4974 - Adds multiple codeMirror configurations as options in the menu

Added the following configurations:
* codeMirror.showCursorWhenSelecting
* codeMirror.autoCloseBrackets
* codeMirror.matchBrackets (shows matching brackets with an underline)
* codeMirror.smartIndent
* codeMirror.tabSize (2, 3 and 4)
* codeMirror.lineNumbers (shows line numbers)

Also, cursorBlinkRate was also moved under codeMirror configurations, and all configurations now 
use the same action.

Omitted the indentWithTabs configuration even if it would have been nice for people that prefer 
tabs. The default CodeMirror behaviour does not add tabs when starting at the beginning of the line; only when the previous line is already indented and then it matches it with a tab. It would require a
bit more work.


applications/desktop/src/notebook/menu.tsx
```diff
@@ -237,13 +237,15 @@ export function dispatchPublishGist(
   }
```
CindyPotvin, line 161: The linted auto-fixed a lot of small things in this file, but there are a few changes too.

applications/desktop/__tests__/renderer/menu.spec.ts
```diff
@@ -195,7 +195,7 @@ describe("dispatchCopyCell", () => {
   });
 });
```
CindyPotvin, line 197: I was not sure if you wanted tests for all the new configurations since they're all dispatching the same action anyway.

packages/editor/src/index.tsx
```diff
@@ -366,7 +390,7 @@ export default class CodeMirrorEditor extends React.Component<
       if (!isConfigurable(optionName)) {
         continue;
       }
-      if (this.props[optionName] !== this.props[optionName]) {
+      if (this.props[optionName] !== prevProps[optionName]) {
```
captainsafia: Great catch!

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
```diff
@@ -27,7 +27,19 @@ const makeMapStateToProps = (state: AppState, ownProps: ComponentProps) => {
...
+      false
+    );
+    let autoCloseBrackets = state.config.get(
+      "codeMirror.autoCloseBrackets",
+      false
+    );
+    let matchBrackets = state.config.get("codeMirror.matchBrackets", true);
+    let smartIndent = state.config.get("codeMirror.smartIndent", true);
+    let tabSize = state.config.get("codeMirror.tabSize", 4);
+    let lineNumbers = state.config.get("codeMirror.lineNumbers", false);
```
captainsafia: Would it be possible to abstract this a little bit more so that it is less repetitive and gives users the flexibility to configure things that are supported by CodeMirror without us having to explicitly expose them?

```js
const defaults = { matchBrackets: true, tabseSize: 4, ... };
const codeMirrorOptions = Object.assign({}, defaults, state.config.get("codeMirror"));

...

return {
  codeMirrorOptions
}
```

Note that this change will likey involve removing some things like the `fullOption` method defined in the CodeMirrror editor class and the configurable.ts check.

packages/editor/src/index.tsx
```diff
@@ -314,7 +318,27 @@ export default class CodeMirrorEditor extends React.Component<
       this.props.editorFocused !== nextProps.editorFocused;
     const cursorBlinkRateChanged =
       this.props.cursorBlinkRate !== nextProps.cursorBlinkRate;
-    return valueChanged || editorFocusedChanged || cursorBlinkRateChanged;
+    const showCursorWhenSelectingChanged =
```
captainsafia: Instead of comparing each value, perhaps we can use lodash's isEqual helper to check if any of the values have changed.

```
hasChanged = _.isEqual(this.props.codeMirrorOptions. prevProps.codeMirrorOptions)
```

Note: that this is assumes we change the expected interval of the editor to take an entire `codeMirrorOptions` object instead of each individual key.

applications/desktop/__tests__/renderer/menu.spec.ts
captainsafia: This is fine. The extra tests here wouldn't be super meaningful.

applications/desktop/src/main/menu.ts
```diff
@@ -44,23 +44,162 @@ function createSender(
...
+    click: createSender("menu:theme", { key: "theme", value: "dark" })
   }
 ];
-const blink_menu = [
-  // TODO: replace the with one `type: 'checkbox'` item once we have state to
-  // know which way it should be set initially.
+// Definitions for all the codeMirror configurations that can be set from the menu.
+// TODO: replace Yes/No `type: 'checkbox'` item once we have state to get current
+// configuration to know which way it should be set initially.
+const codemirror_config_menu = [
```
captainsafia: Nit: `codeMirrorConfigMenu`.

captainsafia (commented):
Looks great so far! I left some comments inline about how we can improve this a bit more.

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
CrystallineCat: Hmm... I think that binding too closely to CodeMirror would be a mistake here. We should have, e.g., a `tabSize` option that works the same regardless of editor used. Additionally, with the changes I'm going to do with #4799 each config option will have to have metadata associated with it anyway, so each will need to be exposed explicitly anyway.

I'd only make two changes here: remove all the `codeMirror.` prefixes to make the options universal and change `lineNumbers` to `showLineNumbers` to be more explicit.

CindyPotvin: @CrystallineCat I fully agree with the idea, but what do you think about replacing `codeMirror.` by `codeEditor.` instead? It would still keep the code editor configurations grouped.

captainsafia: The pickle with assuming the options are universal is that different editors have different ways of configuring the same property. For example, enabling/disabling line numbers in CodeMirror is done by setting the `lineNumbers` property to `true` or `false`. In Monaco, this is done by setting a `lineNumbers` property [an enum value](https://microsoft.github.io/monaco-editor/api/modules/monaco.editor.html#linenumberstype).

My intention behind making the bindings specific to CodeMirror was to allow users to configure options for each editor specifically. For example:

```json
{
  "codeMirror": {
    "lineNumbers": true
  },
  "monaco": {
    "lineNumbers": "relative"
  }
}
```

Depending on the editor that the user has selected in their environment, a different configuration object would be passed down and each editor would support the configuration depending on its internal logic.

I haven't looked fully at your latest PR, @CrystallineCat, so let me know if I am missing something in my understanding of this.

.vscode/tasks.json
```diff
@@ -0,0 +1,16 @@
+{
```
CindyPotvin: Extra file to remove

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
+{
```
CindyPotvin: Extra file to remove

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
CindyPotvin: I'm working on the other changes, I'll wait for the conclusion to this to see if I need to rename.

CindyPotvin: I'm also not convinced about the renaming from `lineNumbers` to `showLineNumbers`, especially if we go with what @captainsafia says. The `lineNumbers` name is directly from the CodeMirror configuration name so it's easier to search for in the docs. We would also require extra code to handle this specific case, while no other editor is supported right now.

CrystallineCat: I'm fine with doing it the way @captainsafia suggested, if you think that is best. Exposing the editor configuration directly doesn't match with what I had in mind for the preference window, but I can see its usefulness. I'll adapt my plans.

@captainsafia do you have anything written on the UX of supporting multiple different editors? I see there is a configuration option for it already in, but is it already possible to use it? Also, is it does it only affect code cells, or also markdown cells? I'm struggling a bit with thinking the whole UX through atm...

captainsafia: > @captainsafia do you have anything written on the UX of supporting multiple different editors?

No exactly this but I can explain it here.

> I see there is a configuration option for it already in, but is it already possible to use it?

Yes. It is possible to use it.

> Also, is it does it only affect code cells, or also markdown cells? 

It affects all cell types.

Here's a breakdown of how it works.

We start off in @nteract/stateful-components. That's the package that we use to render everything nowadays. In particular, the [Editor](https://github.com/nteract/nteract/blob/master/packages/stateful-components/src/inputs/editor.tsx) component reads from the `state.config.editorType` property and selects the editor to render from its children. It should be implemented like this.

```js
import Editor from "@nteract/stateful-components";

<Editor id={id} contentRef={contentRef}>
{
     {
        codemirror: (props: PassedEditorProps) => (
          <ConnectedCodeMirrorEditor {...props} />
        ),
        ace: (props: PassedEditorProps) => (<ConnectedAceEditor {...props}/>
      }
}
</Editor>
```

`PassedEditorProps` is a set of props that are common (or relatively common) across all editors. These are things like the `value` which represents the contents of the editor and `editorFocused`. The `Editor` component is responsible for mapping these properties from the Redux store and passing them to the one of the child editors that is rendered.

To this end, I am imagining that if we group editor configurations by editor type then we can do something like this in the `mapStateToProps` of the editor.

```js
function mapStateToProps(props: OwnProps) {
   const editorType = state.config.editorType;
   // All the config associated with this particular editor
   const config = state.config.get(editorType);
   return {
     ...
     config
      ...
    };
}
```

With this change, `config` would be a property in the `PassedEditorProps` interface so that each editor will receive a `config` object from the parent that matches its config options.

Let me know if this helps!

applications/desktop/__tests__/renderer/epics/config.spec.ts
```diff
@@ -3,8 +3,8 @@ import { ActionsObservable } from "redux-observable";
 import { saveConfigOnChangeEpic } from "../../../src/notebook/epics/config";
 
 describe("saveConfigOnChangeEpic", () => {
-  test("invokes a SAVE_CONFIG when the SET_CONFIG_AT_KEY action happens", done => {
-    const action$ = ActionsObservable.of({ type: "SET_CONFIG_AT_KEY" });
+  test("invokes a SAVE_CONFIG when the SET_CONFIG action happens", done => {
```
CindyPotvin, line 6: I renamed SET_CONFIG_AT_KEY to SET_CONFIG since it now receives the object to insert directly and not a key/value pair.

applications/desktop/src/main/menu.ts
```diff
@@ -41,26 +41,150 @@ function createSender(
...
     label: "Dark",
-    click: createSender("menu:theme", "dark")
+    click: createSender("menu:set-config", { theme: "dark" })
   }
 ];
-const blink_menu = [
-  // TODO: replace the with one `type: 'checkbox'` item once we have state to
-  // know which way it should be set initially.
+// Definitions for all the codeMirror configurations that can be set from the menu.
+// TODO: replace Yes/No `type: 'checkbox'` item once we have state to get current
```
CindyPotvin, line 55: Do you want me to create an issue instead of keeping this TODO in the code?

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
CindyPotvin: @captainsafia I did all the other changes requested if you want to take a second look.

applications/desktop/src/main/menu.ts
captainsafia: We have a bot that will automagically create issues from TODOs so this will work.

packages/editor/src/index.tsx
```diff
@@ -32,6 +32,8 @@ import { Tooltip } from "./components/tooltip";
 import CodeMirrorCSS from "./vendored/codemirror";
 import ShowHintCSS from "./vendored/show-hint";
 
+import isEqual from "lodash.isEqual";
```
captainsafia:
```suggestion
import isEqual from "lodash.isequal";
```

captainsafia (approved):
Looks good overall. Left some small comments inline.

PR #5048, opened 2020-04-13
Add catchError to all persistent kernel susbcribers

vijayupadya:
When unclean websocket close event is thrown (e.g. 1006 websocket close error code), this error percolates to each of the subscribers of the websocket subject. If any of these subscribers don't handle this error, all the source observable streams are getting shutdown resulting in unresponsive application. This change adds catchError to the kernel channel subscribers to ensure the websocket error doesn't percolate to the root and crash the app.

### DIFF
--- a/packages/epics/src/comm.ts
+++ b/packages/epics/src/comm.ts
@@ -1,7 +1,7 @@
 import { ofMessageType } from "@nteract/messaging";
 import { ofType, StateObservable } from "redux-observable";
 import { ActionsObservable } from "redux-observable";
-import { merge, empty, of } from "rxjs";
+import { merge, of } from "rxjs";
 import { map, switchMap, takeUntil, filter, catchError } from "rxjs/operators";
 
 import {
--- a/packages/epics/src/comm.ts
+++ b/packages/epics/src/comm.ts
@@ -1,7 +1,7 @@
 import { ofMessageType } from "@nteract/messaging";
 import { ofType, StateObservable } from "redux-observable";
 import { ActionsObservable } from "redux-observable";
-import { merge, empty } from "rxjs";
+import { merge, empty, of } from "rxjs";
 import { map, switchMap, takeUntil, filter, catchError } from "rxjs/operators";
 
 import {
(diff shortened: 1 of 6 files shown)


packages/epics/src/comm.ts
```diff
@@ -57,7 +57,10 @@ export const commListenEpic = (
                 action.payload.kernelRef === kernelRef
             )
           )
-        )
+        ),
+        catchError((error: Error) => {
+          return empty();
```
captainsafia: Here and below, let's return an error action instead of an empty Observable.

```js
return of(
    actions.executeFailed({
        error: new Error(
        "The WebSocket connection has unexpectedly disconnected."
        ),
        code: errors.EXEC_WEBSOCKET_ERROR,
        contentRef
    })
);
```

packages/epics/src/kernel-lifecycle.ts
```diff
@@ -20,7 +20,7 @@ import {
   switchMap,
   take,
   takeUntil,
-  timeout
+  timeout
```
captainsafia: Nit: Remove extra space.

captainsafia (commented):
Thanks for investigating and submitting this PR. Solid approach overall. Left a few comments inline.

packages/epics/src/comm.ts
vijayupadya: I update two locations. Is contentRef available in watchExecutionStateEpic for KillKernelSuccessful action?

packages/epics/src/kernel-lifecycle.ts
vijayupadya: fixed

packages/epics/src/comm.ts
captainsafia: Yes, it should be accessible under the `action.payload.contentRef` property of the `NewKernelAction`.

vijayupadya: Updated

captainsafia (approved):
Thanks for adding tests and addressing LGTM bot issues.

:shipit:

Issue #5057, opened 2020-04-18
Exectuing empty code cell causes a distracting notification


jarmokivekas:
**Application or Package Used**

nteract 0.22.4 desktop

**Describe the bug**

Executing a code cell with no content causes this distracting notification to appear:

![image](https://user-images.githubusercontent.com/3074453/79631853-40acca80-8164-11ea-92af-62c5f755ba76.png)

Sometimes notebooks just happen to have some empty cells. The notification needlessly flood the screen especially when going shift-enter through a notebook or using the "Restart and Run All Cells"

**To Reproduce**

1. Open a notebook
2. Create a code cell
3. Don't write anything in it
4. Press ctrl-enter
5. Notiication appears

**Expected behavior**

Executing an empty code cell should not create notification.

**Desktop:**

- OS: ubuntu 19.10
- Version: nteract 0.22.4

**Additional context**

Related to #4961, as the notification stems from the same set of pre-execution checks.

https://github.com/nteract/nteract/blob/c21cb9dd43c9f25165ead0167ec5643e52f62c57/packages/epics/src/execute/execute.ts#L371-L387

jarmokivekas:
Closing this issue as fixed by #5061

Issue #4961, opened 2020-02-27
Toggle "Execute failed" for markdown cells.


thomascamminady:
A common situation is to have code cells mixed with markdown cells. 
I tend to `shift+enter` through cells when opening a new notebook and find it rather irritating to get an error message when `shift+enter`ing through a markdown cell, see the image below.

![Screenshot_20200227_092530](https://user-images.githubusercontent.com/24543638/75425610-5d3a3c80-5943-11ea-9f09-ab4a48d3a8ff.png)

This does not happen when using e.g. `jupyter lab`. I see that this is a desired feature as it is implemented in `packages/epics/src/execute/execute.ts`:
![Screenshot_20200227_093146](https://user-images.githubusercontent.com/24543638/75426014-100a9a80-5944-11ea-8ef9-c2e4ff96f6f0.png)

But maybe there would be the chance to toggle this error message off? Or display a warning instead?

willingc:
Thanks @camminady for reporting this issue.

Confirming that I see the same behavior when running the example Python notebooks on 0.22. There is also a typo in the error box.

jarmokivekas:
The way I see it, if a markdown cell makes it all the way to the cell type check in https://github.com/nteract/nteract/blob/c21cb9dd43c9f25165ead0167ec5643e52f62c57/packages/epics/src/execute/execute.ts#L359
it's ok to throw an error, since endering markdown is isn't the kernel's task (that check is already more of a backend functionality). The fix should probably be more toward the frontend, i.e ctlr-enter or shift-enter on a markdown cell shouldn't send a request or execution in the first place, right?

jarmokivekas:
Closing this as fixed in #5061

PR #5061, opened 2020-04-19
Silence error notifications from empty or non-code cell execution

jarmokivekas:
This PR proposes a fix for #4961 and at least partially for #5057

The goal is to restore normal cell execution behaviour, where using shift-enter and ctrl-enter on any cell in a typical notebook doesn't cause any _Execution Failed_ notifications 

No _Execution Failed_ notifications for markdown cells (`EXEC_INVALID_CELL_TYPE`) or empty code cells (`EXEC_NO_SOURCE_ERROR`) should appear when using: 

- [x] shift-enter
- [x] ctrl-enter
- [x] play button on empty code cell
- [x] Cell > Run All
- [x] Cell > Run All Below
- [x] Runtime > Restart and Run All Cells

The control flow goes: `keyDown` > `executeFocusedCell` > `executeCell` > `sendExecuteRequest`.
The play button and _run all cells_ don't go through `executeFocusedCell`, so might out-of-scope them, since the need to be done separately. Alternatively everything could be handled in `sendExecuteRequest` where the errors are raised in the first place.

I'm a newbie with typescript and most of the codebase, I've put some toughts as comments in the committed code for review.

- [x] remove RID comments
- [x] remove WIP


packages/notebook-app-component/src/decorators/kbd-shortcuts/index.tsx
```diff
@@ -88,6 +88,10 @@ export class KeyboardShortcuts extends React.Component<Props> {
     if (focusedCell) {
       // NOTE: Order matters here because we need it to execute _before_ we
       // focus the next cell
+      // RID:JK It's easier to check for markdown and empty
+      // cells inside executeFocusedCell, because all the
+      // tooling for getting cells from a contentRef is already
+      // there.
```
BenRussert: I was thinking this issue probably comes up due to keyboard shortcuts when a user just wants to render markdown and skip to the next cell. One other possible solution would be something like: 

```ts

      const isExecutableCodeCell =
        cellMap.get(focusedCell).get("cell_type") === "code";

      if (isExecutableCodeCell) {
        executeFocusedCell({ contentRef });
      }

```

I don't know for certain which approach is better, but wonder what you think of something like the snippet above.

packages/epics/src/execute/execute.ts
```diff
@@ -372,6 +376,8 @@ export function sendExecuteRequestEpic(
              * We cannot execute cells with no content, so
              * we through an error action if this is the case.
              */
+            // RID:JK this is also causing distracting notifications
+            // in normal use. Should be handled similar to markdown cells
```
BenRussert: I think we could remove these comments if the PR resolves this problem. 

Depending on your final solution here, I might edit the docstring on line ~348 to reflect the new behavior. Something like "attempts to execute non-code cells will fail silently"

BenRussert (commented):
Thanks for taking this on! 

I left a comment below as a suggestion, but if you believe the epic approach is better I could be easily convinced.

packages/notebook-app-component/src/decorators/kbd-shortcuts/index.tsx
jarmokivekas: I think your suggestion is nicer that what I did in the executeFocusedCellEpic. I've reimplemented using this method.

jarmokivekas:
I pushed a new implementation, which silences the relevant errors thrown by `sendExecutionRequestEpic` by using `executionCanceled` instead of `executionFailed`. That seems to be the only place where all cells pass through regardless of how the execution of a cell or set of cells started (keyboard shortcut, cell play button, and the runtime menu, etc). Having the check here in `sendExecutionRequestEpic` and for it to be silent looks like the cleanest way to avoid re-implementing the same check in multiple places (once for all different execution methods).

Here is a notebook with different edge-case cells I used for testing this.

https://gist.github.com/jarmokivekas/4837c46775265cb33f5be1dfa0eb280d

jdetle:
> `throws an error when cell is not a code cell`
 
It looks like this behavior is changing, @BenRussert @captainsafia are we okay with changing this behavior?

> `"Cannot execute cells with no source content"` 

It looks like this is an unintended regression and needs to be fixed? I might be mistaken.

captainsafia:
> It looks like this behavior is changing, @BenRussert @captainsafia are we okay with changing this behavior?

This seems fine. Ideally, we would make it so that hitting that part of the codebase still throws an "EXECUTE_FAILED" action but the action shouldn't be reflected in the UI. We probably want to set up something in our notifications layer to filter out error types that we do not want to be reflected in the UI. That way the information is still available in logs but not available to the user.

jarmokivekas:
@captainsafia Ah, that so much sense. I'll revert the changes to the epic and look for a way to filter out `errors.EXEC_NO_SOURCE_ERROR` and `errors.EXEC_INVALID_CELL_TYPE` in the notifications.

jarmokivekas:
Commit 149fd46 has solution which doesn't alter the epics at all, and instead filters notifications based on error code in the buleprintjs tooling. This seems to work well regardless of how the execution is initiated from the user interface.

jarmokivekas:
While ignoring warning notifications, there is also this notification, which pops up every time I use Runtime > Restart and Run All Cells (fairly often in my workflow). Despite the error message, there is not actually any user-noticable error, it's also a little distracting. But I guess errors messages have reasons, so maybe going overboard with hiding them isn't necessarily the right solution.

image

Also do warnings get logged somehow in the console or a file? I couldn't find any place to see them.

captainsafia:
> Despite the error message, there is not actually any user-noticable error, it's also a little distracting. But I guess errors messages have reasons, so maybe going overboard with hiding them isn't necessarily the right solution.

I believe this issue is happening because of a bug in our shutdown logic for 0mq-based kernels. It's fine to disable it for now since the user cannot take any action on it.

> Also do warnings get logged somehow in the console or a file? I couldn't find any place to see them.

Warnings issued via `console.warn` statements do get logged in the dev tools console. Warning notifications do not at the moment.

packages/mythic-notifications/src/myths/send-notification.ts
```diff
@@ -34,6 +34,8 @@ export const sendNotification =
           title: titleFromAction(action),
           message: messageFromAction(action),
           level: "error",
+          code: action.payload.code
```
captainsafia: Can we implement this in a `codeFromAction` function as is done in the rest of this codebase?

packages/mythic-notifications/src/backends/blueprintjs.tsx
```diff
@@ -23,6 +23,15 @@ export const blueprintjsNotificationSystem =
         error: Intent.DANGER,
       }[msg.level ?? "info"];
 
+      /* We filter out some errors that common, but can be ignored
+       * by users safely to prevent distracting norifications in the UI */
+      if (
```
captainsafia: I'm not sure if this is the best place to add the logic for skipping displaying notifications. @CrystallineCat might have more opinions on this.

In either case, I'd recommend we create a `BLOCKED_ERROR_CODES` constant or something that contains the list of error codes that won't have errors shown by them and then making the if-statement check `if (BLOCKED_ERROR_CODES.contains(msg.code))`

captainsafia (commented):
This seems like a more sensible approach overall.

I looped @CrystallineCat in one of the comments to get her sense of how we should handle visible/invisible notifications.

packages/mythic-notifications/src/myths/send-notification.ts
jarmokivekas: Yes. I've updated to include at `codeFromAction` function

packages/mythic-notifications/src/backends/blueprintjs.tsx
jarmokivekas: I tried making a list of codes by using their defined `const`s by doing something like:

```typescript
import { errors } from "@nteract/types";
import { actions } from "@nteract/core";

const BLOCKED_ERROR_CODES = [
  errors.EXEC_NO_SOURCE_ERROR,
  errors.EXEC_INVALID_CELL_TYPE,
  actions.SHUTDOWN_REPLY_TIMED_OUT
];
```

But that gives me for a bunch of related files an error like this in `yarn build:all`:

```
error TS5055: Cannot write file 'packages/mythic-notifications/lib/types.d.ts' because it would overwrite input file.
```

Is there something I'm how imports are done?

For now I'm just matching against strings instead.

packages/epics/src/execute/execute.ts
```diff
@@ -379,6 +380,7 @@ export function sendExecuteRequestEpic(
                   error: new Error(
                     "Cannot execute cells with no source content."
                   ),
+                  // RID:JK We want to silence this in the Ui
```
jdetle: I haven't looked at nteract code in a bit, what does "RID" mean?

CrystallineCat:
Thanks for looping me in @captainsafia! I'm honestly not convinced silencing the notifications is the way to go.

Here's what I'd do:
```typescript
  errors.EXEC_NO_SOURCE_ERROR
```
Why can't we exec code cells with no source? I'm inclined to just remove that whole check and error and just send the kernel the empty code string. I would expect that the kernel just sends back an empty reply, but if it errors, then we could treat empty code cells like invalid cell types below...

```typescript
  errors.EXEC_INVALID_CELL_TYPE
```
Why are we generating exec actions for non-executable cells in the first place? I think we need to alter the UI to not generate those.

Create a function...
```typescript
const isExecutable = (cell) =>
    cell.cell_type === "code" && cell.source !== "";
```
...and check that before generating the action? I see no benefit from logging attempted execution of non-code cells, either.

```typescript
  actions.SHUTDOWN_REPLY_TIMED_OUT
```
If this action is harmless, it shouldn't be an error action (removing the `error` parameter should do the trick). That way it'll still get logged if that's turned on and won't generate a notification.

captainsafia:
> Why can't we exec code cells with no source? I'm inclined to just remove that whole check and error and just send the kernel the empty code string. 

This was done for scenarios where users are using the SDK to build their own notebook clients so that errors could be surfaced how they want in the host applications. Returning "empty()" from library functions was opaque in a lot of scenarios where users would not know what was causing the issue and would have to resort to debugging internal nteract source code to track it down.

> If this action is harmless, it shouldn't be an error action (removing the error parameter should do the trick). That way it'll still get logged if that's turned on and won't generate a notification.

This seems like a sensible approach. It's not an official Flux type but perhaps typing it as "warning" will strike a middle ground here.

jarmokivekas:
> Why are we generating exec actions for non-executable cells in the first place? I think we need to alter the UI to not generate those.

There is and older partial solution for this which does the check in the UI when using keyboard shortcuts. But turns out there are many ways to start a cell excution action, so it would be necessary to have a similar check in multiple places. I thougth it would be easier to have one check that covers  all the options, so I decided to try something else.

The previous solution is here: https://github.com/nteract/nteract/compare/master...jarmokivekas:98dd23b16aad2b58043028f6535ff52678defe7a

jarmokivekas:
> If this action is harmless, it shouldn't be an error action (removing the error parameter should do the trick). That way it'll still get logged if that's turned on and won't generate a notification.

I did a quick test to just remove the `error` parameter and a few other ways down grade the errors. The `error` parameter seems to be required, and can't be undefined. I traced it to this snippet of generated code in  `packages/actions/lib/actionTypes/kernel_execution.d.ts`:

```typescript
export declare const shutdownReplyTimedOut: (payload: HasKernel & {
    error: Error;
    code?: string | undefined;
}) => {
    type: "SHUTDOWN_REPLY_TIMED_OUT";
    payload: HasKernel & {
        error: Error;
        code?: string | undefined;
    };
    error: true;
};
```

This applies to `executeFailed` used for markdown cells and empty cells also.

CrystallineCat:
> The error parameter seems to be required, and can't be undefined. I traced it to this snippet of generated code in packages/actions/lib/actionTypes/kernel_execution.d.ts:

Apologies, I forgot the type definitions. That file is generated from [packages/actions/src/actionTypes/kernel_lifecycle.ts](https://github.com/nteract/nteract/blob/master/packages/actions/src/actionTypes/kernel_lifecycle.ts). There are two places you need to change for each action where you remove the `error` attribute, e.g.:
```diff
---
-export type ShutdownReplyTimedOut         = ErrorAction<typeof SHUTDOWN_REPLY_TIMED_OUT,    HasKernel>;
+export type ShutdownReplyTimedOut         = Action     <typeof SHUTDOWN_REPLY_TIMED_OUT,    HasKernel>;
---
-export const shutdownReplyTimedOut        = makeErrorActionFunction <ShutdownReplyTimedOut>     (SHUTDOWN_REPLY_TIMED_OUT);
+export const shutdownReplyTimedOut        = makeActionFunction      <ShutdownReplyTimedOut>     (SHUTDOWN_REPLY_TIMED_OUT);
---
```

Thanks for the additional info. I think that emitting a non-error action instead of `ExecuteFailed` might be better than changing the UI, then. Something like `ExecuteNoOp` with a payload giving the reason (empty code cell vs. non-code cell)?

Otherwise, in case you come to the conclusion that you want to keep the error actions, instead of filtering the notifications out, I'd not create them in the first place. That code is at [packages/mythic-notifications/src/myths/send-notification.ts#L32](https://github.com/nteract/nteract/blob/master/packages/mythic-notifications/src/myths/send-notification.ts#L32), specifically the `on:`-clause. That way, every notification in the log was actually displayed.

jarmokivekas:
Thanks for the pointer @CrystallineCat, it was very helpful :100: 

I think I'm pretty happy with how this is implemented now.

In summary:
- This would essentially down-grade the returned errors to cancellations instead.
- The cancellations still carry the same error code as before, so user code can reason about why it cancelled.
- Since they're not labelled as errors, there is no need to individually blacklist them in notification logic, since they won't trigger notifications.
- There are no changes to the notification subsystem.
- There are no changes to the user interface code: it still might attempt executing markdown cells and empty cells, depending on how execution is started. There are many ways to trigger a cell execution, and they would all need to make the same checks for cell type and content, resulting in duplicating code, which would be more difficult to maintain (in my opinon).

In manual testing this implementation seems work well.

captainsafia (approved):
Looks good on my end. I'll let others chime in as well.

CrystallineCat (approved):
This looks fine for me as well. Thank you for implementing this! :cat:

jarmokivekas:
I had like 30 no-op commits that made changes and then were later reverted back to master. Squashed them because it was bugging me. The force push didn't actually change anything, just clean the history.

Sorry if that's confusing.

PR #5064, opened 2020-04-20
Add 'markdownOptions' as a state config to tweak react-markdown behavior

gitanuj:
1. Update @nteract/markdown to 4.4.0
2. Expose `markdownOptions` as state config. This is plumbed down to `react-markdown` to tweak markdown rendering.
3. Add default for `linkTarget`. This forces all markdown hyperlinks to open in new tab.

### DIFF
--- a/packages/stateful-components/package.json
+++ b/packages/stateful-components/package.json
@@ -22,9 +22,11 @@
     "@nteract/commutable": "^7.2.10-alpha.0",
     "@nteract/core": "^13.1.0-alpha.0",
     "@nteract/fixtures": "^2.3.5-alpha.0",
+    "@nteract/markdown": "4.4.0",
     "@nteract/outputs": "^3.0.9",
     "@nteract/presentational-components": "^3.3.12-alpha.0",
     "immutable": "^4.0.0-rc.12",
+    "react-markdown": "^4.3.1",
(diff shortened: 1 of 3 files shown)


packages/stateful-components/package.json
```diff
@@ -22,9 +22,11 @@
     "@nteract/commutable": "^7.2.10-alpha.0",
     "@nteract/core": "^13.1.0-alpha.0",
     "@nteract/fixtures": "^2.3.5-alpha.0",
+    "@nteract/markdown": "4.4.0",
     "@nteract/outputs": "^3.0.9",
     "@nteract/presentational-components": "^3.3.12-alpha.0",
     "immutable": "^4.0.0-rc.12",
+    "react-markdown": "^4.3.1",
```
captainsafia: Let's move this to the `devDependencies` of this package since it is only used to add type definitions for the props it doesn't need to be included in production builds.

captainsafia (approved):
Looks good. Left one small comment inline. Pinged you off-GitHub about validation.

packages/stateful-components/package.json
gitanuj: Moved `react-markdown` to devDependencies instead.

PR #5073, opened 2020-04-24
Hide cell toolbar in exported PDF / print

### LINKED ISSUE #5062: Cell menu elements of active cell visible in pdf exports

- nteract desktop, v0.22.4 installed from snap
- nteract desktop, current master, in development mode

**Describe the bug**

Cell menu elements of active cell visible in pdf exports:

image

image
 

**To Reproduce**

1. Open a new notebook
2. Click a cell to make it focused
3. File > Export PDF
4. Open the pdf in a pdf viewer
5. Cell menu icons are still visible on the focused cell

**Expected behavior**

The play button, menu arrow, trash can, and the drop down menu itself should be hidden in pdf exports. This should be able to be done using css `@print` media style rules in the relevant components.

jarmokivekas:
This PR fixes #5062 

The `@media print` rule for `.nteract-cell-toolbar` in the default theme is overridden by something, and is not applied when exporting to pdf. Using `display: none !important` works too.

Other components have this rule in the component using `styled.div`, so this is more consistent with that.

### DIFF
--- a/packages/stateful-components/src/cells/toolbar.tsx
+++ b/packages/stateful-components/src/cells/toolbar.tsx
@@ -4,6 +4,7 @@ import { Dispatch } from "redux";
 
 import { ContentRef, actions, AppState, selectors } from "@nteract/core";
 import { CellType } from "@nteract/commutable";
+import styled from "styled-components";
 
 export interface ComponentProps {
   id: string;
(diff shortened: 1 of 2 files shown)


jdetle (approved):
This must have been missed in a migration to styled components, thank you @jarmokivekas

PR #5076, opened 2020-04-26
Fix CSS in print mode and resolution of relative references

### LINKED ISSUE #5056: Markdown cells do not render local images (regression)
nteract desktop, Version 0.22.4 (0.22.4), running on macOS

**Describe the bug**
Locally sourced images do not render, but images pulled from online sources do.

This is seemingly a regression from
https://github.com/nteract/nteract/issues/1242
https://github.com/nteract/nteract/issues/2127

**To Reproduce**

```
### Images
<img src="yaktocat.png"/>
```

**Expected behavior**
The image should show up equally well when sourced from online or local documents.

**Additional context**
Workaround by using URL for local access, i.e. `file:///Users/.../yaktocat.png`

captainsafia:
Closes #5056.

I also reverted the changes introduced in #5073 and fixes the original bug using pure CSS.

For context, we want to avoid having any styles within the stateful-components package as described in #4731 because:

- styled-components has negative performance repercussions in a lot of cases
- stateful-components is designed to be unopinionated about how styles are applied to a page

### DIFF
--- a/applications/desktop/src/notebook/index.tsx
+++ b/applications/desktop/src/notebook/index.tsx
@@ -23,6 +23,7 @@ import DataExplorer from "@nteract/data-explorer";
 import WidgetDisplay from "@nteract/jupyter-widgets";
 import * as MathJax from "@nteract/mathjax";
 import NotebookApp from "@nteract/notebook-app-component";
+import NotebookHelmet from "@nteract/notebook-app-component/decorators/notebook-helmet";
 import { Media } from "@nteract/outputs";
 
 import "@nteract/styles/app.css";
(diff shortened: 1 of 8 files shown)


jarmokivekas:
The CSS media print rules and image handling in html `img` tags look seem fine to me :+1:

#5056 specifically mentions html images, but markdown `alt` don't render local images either currently, regardless of `file://`, `/home/...` or `./`paths. Using `https.//...` images render normally in markdown. A different but related issue?

captainsafia:
@jarmokivekas Both Markdown images and HTML images end up being rendered the same way on the page, via a `` tag. The `base` tag is used to set the default path that images will be resolved to in both scenarios.

jarmokivekas (approved):

PR #5084, opened 2020-05-06
Always prettify notebook json when creating/saving

gitanuj:


gitanuj:
After further investigation found out that `content` type is expected to be json by Jupyter. It is up to whoever is writing the file to disk to decide whether to format it or not. That's the reason we use `stringifyNotebook` only in `LocalContentProvider` and `DOWNLOAD_CONTENT`.

PR #5085, opened 2020-05-06
Fix CSS for collapsed cell and remove outputExpanded

### LINKED ISSUE #4448: Implement UI support for collapsed property in notebook
We currently parse and style collapsed cells outputs via commutable and our component but don't give users an interface for toggling the collapsed state from the notebook.

The ideal implementation would allow users to toggle this via the cell menu so that it has a little bit more visibility to the user than the classic Jupyter UI and co-exists alongisde our other visibility toggles.

Verification for this should include tests for the reducers and actions and snapshot tests for the output component.

captainsafia:
Resolves #4448 

As it turns out, we already had the toggle expanded property in the cell menu but we were missing the CSS for it. This is implemented now.

image

### DIFF
--- a/packages/reducers/__tests__/core/entities/contents/notebook.spec.ts
+++ b/packages/reducers/__tests__/core/entities/contents/notebook.spec.ts
@@ -8,7 +8,7 @@ import {
   makeDisplayData,
   makeErrorOutput,
   makeExecuteResult,
-  makeStreamOutput
+  makeStreamOutput,
 } from "@nteract/commutable";
 import * as Immutable from "immutable";
 import uuid from "uuid/v4";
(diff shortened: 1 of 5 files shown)


packages/reducers/__tests__/core/entities/contents/notebook.spec.ts
```diff
@@ -786,7 +786,6 @@ describe("toggleOutputExpansion", () => {
       cells =>
```
CrystallineCat: Nit: change the `outputExpanded` in the name of the test:
```ts 
test("changes outputExpanded set", () => ...)
```

CrystallineCat (approved):
Looks good! Left one comment in the code.

badiku:
packages/styles/themes/base.css 
 .nteract-cell-outputs {

`  overflow-y: scroll`

doesn't this line you added force scroll bar to appear, even though there's only one line?

scrolly


I overrided this boring scroll bar like this :

```python
from IPython.display import HTML
style = '''
    <style>
        div.nteract-cell-outputs {overflow-y: auto }
    </style>
    '''
HTML(style)
```

PR #5086, opened 2020-05-06
Fix broken links

jruales:
Small PR to fix some broken links


captainsafia (approved):
Oh bother -- the configuration on this deployment must've stripped out the mapping of `*.html` to `*` when we migrated to now V2.

In any case, thanks for being proactive and submitting this fix!

PR #5096, opened 2020-05-12
Move rxjs-based packages out of monorepo

captainsafia:
When I originally moved these packages out of the monorepo, I was running into issues with the build on Windows. See https://github.com/nteract/nteract/pull/5018.

Turns out that those issues might have had something to do with my particular machine. This would make sense since they do not repro on macOS.

cc: @jdetle to re-confirm that this PR builds fine on his Windows machine.


captainsafia:
Tested this on a Windows VM and everything looks good.

PR #5103, opened 2020-05-13
Upgrade redux-observable to 2.0.0-alpha.0

CrystallineCat:
Progress on #4799. Formerly part of #5031.

I ran into a bug in `redux-observable`, so I upgraded to latest (`1.0.0` → `2.0.0-alpha.0`, [Changelog](https://github.com/redux-observable/redux-observable/blob/master/CHANGELOG.md)) for a fix. **This means `ActionsObservable` is gone, it is just `Observable` now.**

### DIFF
--- a/applications/desktop/src/notebook/epics/close-notebook.ts
+++ b/applications/desktop/src/notebook/epics/close-notebook.ts
@@ -5,7 +5,7 @@ import {
 } from "@nteract/core";
 import { ipcRenderer as ipc } from "electron";
 import { RecordOf } from "immutable";
-import { ActionsObservable, ofType, StateObservable } from "redux-observable";
+import { ofType, StateObservable } from "redux-observable";
 import { concat, empty, Observable, Observer, of, zip } from "rxjs";
 import {
   catchError,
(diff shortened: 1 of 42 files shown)


applications/desktop/src/notebook/epics/close-notebook.ts
```diff
@@ -29,20 +29,11 @@ import { KernelRecord, KernelRef } from "@nteract/types";
 import { Actions } from "../actions";
 
 export const closeNotebookEpic = (
-  action$: ActionsObservable<
-    | CloseNotebook
-    | coreActions.KillKernelFailed
-    | coreActions.KillKernelSuccessful
-  >,
+  action$: Observable<any>,
```
captainsafia, line 32: I assume you ran into some trouble with TypeScript here.

cc: @jayphelps

packages/epics/src/execute/execute.ts
```diff
@@ -197,7 +197,7 @@ type StopExecutionActions =
 type ExecuteStreamActions = StopExecutionActions | actions.SendExecuteRequest;
 
 export function createExecuteCellStream(
-  action$: ActionsObservable<ExecuteStreamActions>,
+  action$: Observable<ExecuteStreamActions>,
```
captainsafia, line 200: Great to see that the types worked out here. @jayphelps and I ran into some trouble with this when we tried upgrading to 2.0.0 a few months back.

captainsafia (approved):
Looks good!

Heads up, @jayphelps -- this PR upgrades nteract to the 2.0 of redux-observable. 

Also, I need to fix our prettier pre-commit git hook....

applications/desktop/src/notebook/epics/close-notebook.ts
CrystallineCat: Yeah, I got errors. But also, as I understand it, the outer typing was wrong, as this is before the `ofType(...)`, the observable would include all actions and thus be of a union of all possible action types. So I don't think `any` is particularly problematic here, although maybe `AnyAction` would be better.

PR #5104, opened 2020-05-13
Fix syntax highlighting getting lost on CodeMirror mode update

### LINKED ISSUE #5093: Executing cell in new notebook removes syntax highlighting

- nteract desktop, v 0.23.1
- Ubuntu 19.10

**Describe the bug**

Executing cell in python notebook removes syntax highlighting from the notebook.

**To Reproduce**

1. Open a new notebook:  _File_ > _New_ > _Python 3_
2. Write a some code in a code cell e.g. `import math` (completely arbitrary)
3. Execute the cell, _Ctrl-Enter_
4. The cell no longer has any syntax hilighting.

Some observations:

- If you create multiple cells, all cells lose highlighting when one of them is executed.
- I can't reproduce this by opening a file from disk, only in new notebooks.
- Focusing a new cell seems to restore the highlighting in the previously focused cell

**Expected behavior**

All code cell should have syntax highlighting all the time

**Screenshots**

Before executing cell:

image

After executing cell:

image

CrystallineCat:
Fixes #5093.

The issue was that the mode becomes an `Immutable.Map` after receiving mode info from the kernel, and that map was only properly converted to a JS object on the initial setting of editor options, while the code for updating the options passed it through unchanged.

### DIFF
--- a/packages/editor/src/index.tsx
+++ b/packages/editor/src/index.tsx
@@ -349,7 +349,11 @@ export default class CodeMirrorEditor extends React.Component<
       if (
         this.props.codeMirror[optionName] !== prevProps.codeMirror[optionName]
       ) {
-        this.cm.setOption(optionName, this.props.codeMirror[optionName]);
+        if (optionName === "mode") {
+          this.cm.setOption(optionName, this.cleanMode());
+        } else {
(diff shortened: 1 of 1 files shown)


captainsafia (approved):
Ahh -- good catch!

Issue #4781, opened 2019-12-18
RFC: Standardizing keyboard shortcut support across nteract apps


captainsafia:
**Context**

Currently in nteract, we provide two main applications: an Electron-based desktop app and a Jupyter extension. Each application has a different setup for how it provides keyboard shortcuts. The desktop app uses the wiring that Electron provides for keyboard shortcuts. The Jupyter extension uses the react-hotkeys package to configure keyboard shortcuts.

Ideally, we should have a unified implementation for keyboard shortcuts on both the desktop app and the electron app.

Electron provides support for [event-listener based keyboard shortcuts on the renderer process](https://electronjs.org/docs/tutorial/keyboard-shortcuts#shortcuts-within-a-browserwindow) so we could potentially use react-hotekeys there, if need be.

Refs: #4101, #4102

**Problem**

Keyboard shortcuts should be implemented in a standardized fashion across nteract apps so we don't have to use different technologies to provide keyboard shortcuts in each one.

**Proposed Solution**

1. Create a new @nteract/keymap package that will export a JSON file that contains a standardized schema for cross-application keymaps in nteract. Based on my research, the keymap should look something like this to support use in both `electron` and `react-hotkeys`.

```js
{
    // Should match the Redux action
    name: string;
    // Should match the format that is needed by Electron 
    accelerator: string;
    // A function that passes the hotkey string format as needed by react-hotkey
    hotkey: () => string;
    // A human-friendly description of the string 
    description: string; 
    // A human-friendly label for the keyboard shortcut
    label: string;
}
```

2. Update `src/main/menu.ts` to read from the keymap package for keyboard shortcuts for each action.

```js
import keys from "@nteract/keymap";

const menu = {
    new: {
      label: keys[new].label,
      accelerator: keys[new].accelerator;
    },
    open: {
      label: keys[open].label,
      click: () => {
        ...
      },
      accelerator: keys[open].accelerator
    }
};
```

3. Update the implementation of `react-hotkeys` in the Jupyter extension to use the same keymap package.

```js
import keys from "@nteract/keymap";
import { HotKeys, KeyMap } from "react-hotkeys";

const keyMap = keys.map(key => {
    return { key.name: key.hotkey() };
});

const handlers = keys.map(key => {
    return { key.name:  () => dispatch(actions.createAction<keyName>())
})

<HotKeys keyMap={keyMap} handlers={handlers}>
    <App/>
</Hotkeys>
```

[...]

PR #5105, opened 2020-05-14
Add support for configuring keyboard shortcuts

### LINKED ISSUE #5092: Keyboard shortcuts should be configurable

The auto-completion shortcut, Ctrl-Space, conflicts with a system shortcut (trigger Spotlight on macOS)

Make shortcuts configurable in nteract desktop.

A workaround is to change the system-wide shortcut to free up Ctrl-Space for nteract. I am not willing to do this because my setup work well, and nteract is _the only_ application that had this problem. All other programs that had Ctrl-Space set for auto-complete allowed it to be reconfigured.

**Additional context**

The lack of auto-completion makes nteract basically unusable.

captainsafia:
This PR adds support for configuring keyboard shortcuts for different actions in the nteract desktop app. The configurable actions are documented in docs/configuring-kbd-shortcuts.md.

When the nteract application launches, we store the config the user provides in the `nteract.json` file under a global CONFIG attribute in the main process of the application. The configuration system expects the custom keymaps to be configured under the `keyboardShortcuts` map as read from in applications/desktop/src/main/index.ts.

When we build the menu in applications/desktop/src/main/index.ts, we first look for accelerators defined by the user's config before using our own defaults.

This PR also exposes configuration for actions that aren't currently tied to a keyboard shortcut.

Closes #5092 #4781 (ish)

### DIFF
--- a/applications/desktop/src/main/index.ts
+++ b/applications/desktop/src/main/index.ts
@@ -14,12 +14,12 @@ import {
   Event,
   ipcMain as ipc,
   Menu,
-  Tray
+  Tray,
 } from "electron";
 import {
   mkdirpObservable,
   readFileObservable,
-  writeFileObservable
+  writeFileObservable,
 } from "fs-observable";
 import { forkJoin, fromEvent, Observable, Subscriber, zip } from "rxjs";
 import {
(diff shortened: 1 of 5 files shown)


jarmokivekas:
Related to this, the ctrl-enter and shift-enter shortcuts could probably be implemented this way too. They're currently implemented as a bunch of if-statements somewhere in a keydown event application logic. I don't think those action have menu items, so they might not right away fit this same pattern though.

captainsafia:
> Related to this, the ctrl-enter and shift-enter shortcuts could probably be implemented this way too. They're currently implemented as a bunch of if-statements somewhere in a keydown event application logic. I don't think those action have menu items, so they might not right away fit this same pattern though.

Yeah, we've always had the keyboard registrations for cell execution registered within the web application as opposed to within Electron's keyboard handling. The reason for this eludes me now but I suspect it has something to do with keeping the code consistent across the Jupyter extension and desktop app.

I agree that it would make sense to shift these to the Electron keyboard shortcut handling -- assuming this doesn't cause any unintended consequences.

captainsafia:
Closing this in favor of a different approach once the config story is sorted out.

PR #5106, opened 2020-05-14
Improve myths package and API

CrystallineCat:
Progress on #4799. Formerly part of #5031. <del>Needs #5103 to be merged first (branch based on that branch; changes from there will show up here, hopefully only until it is merged).</del>

Some improvements to the `myths` package needed for the `mythic-configuration` package.


packages/myths/src/epics.ts
```diff
@@ -1,32 +1,47 @@
...
+  myth: Myth<PKG, NAME, PROPS, STATE>,
+  dispatch: EpicFuncDefinition<STATE, PROPS, MythicAction>,
+  narrow: (source: Observable<MythicAction>) => Observable<MythicAction>
+) =>
+    (
+      action$: Observable<MythicAction>,
+      state$: StateObservable<RootState<PKG, STATE>>,
+    ) =>
+      action$.pipe(
+        narrow,
```
captainsafia: Nit: `filterBy` seems like a better verb to use here.

packages/myths/src/epics.ts
```diff
@@ -1,32 +1,47 @@
...
+        withLatestFrom(state$.pipe(map(state => state.__private__[pkg]))),
```
captainsafia, line 23: What patterns exist for accessing the overall state from inside a myth? For example, how does a myth get the code cells in a notebook?

packages/myths/src/package.ts
```diff
@@ -30,5 +39,46 @@ export const createMythicPackage =
 
         createMyth:
           makeCreateMyth(pkg, myths),
+
+        createSelector:
+          <T>(selector: Selector<STATE, T>) =>
+            (state?: MaybeRootState<PKG, STATE>) => {
+              const value = selector(state?.__private__?.[pkg]);
+              return (value as any).toJS !== undefined
+                ? (value as any).toJS()
```
captainsafia: Why not return the Immutable representation here?

One benefit is standardizing on JSON objects for everything. However, a red flag that popped up for me is running `toJS` too frequently on large Immutable objects.

packages/myths/src/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { Observable, of } from "rxjs";
+import { catchError } from "rxjs/operators";
+
+export const mapErrorTo = <T, O>(target: T, pred: (err: any) => boolean) =>
```
captainsafia: What's the purpose of exposing this `mapErrorTo` method as part of the API for myths?

It seems a little out-of-place here but perhaps there's some context to this that I am not aware of.

packages/myths/src/store.ts
```diff
@@ -7,7 +8,7 @@ import { MythicAction, MythicPackage } from "./types";
 export const makeConfigureStore = <STATE>() => <DEPS>(
   definition: {
     packages: MythicPackage[],
-    reducers?: ReducersMapObject<STATE, any>
+    reducers?: ReducersMapObject<Omit<STATE, "__private__">, any>
```
captainsafia: Should `Omit` be imported from the new types package or something?

captainsafia (commented):
Left some comments inline.

Can you write something about what the improvements are? I haven't played around much with myths so I'm not sure what gaps this PR fixes.

packages/myths/src/store.ts
CrystallineCat: `Omit` is a TS built-in: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk

packages/myths/src/package.ts
CrystallineCat: Hmm... the only use for `createSelector` in my next patch is to make a function that fetches a specific config value from the state:
```ts
configuration.createSelector(
  state => state?.current?.get(props.key) ?? props.defaultValue,
)
```

In most cases, that will be a `string` or `number`, but in cases it's an object, I think it's easier to have it in plain JS. The configuration values should be small enough for speed not to matter much. However, I could see it as a problem for the general case of getting an object from the private state. I might move the `toJS()` to the calling code, I'll have to think about it some more.

packages/myths/src/utils.ts
CrystallineCat: It's an operator to be used in epics, like this:
```ts
    readFileObservable(filename).pipe(
      mapErrorTo("{}", err => err.code === "ENOENT"),
      map(data => JSON.parse(data.toString())),
      // SyntaxError means the file is probably in the middle of a write
      mapErrorTo(undefined, err => err.name === "SyntaxError"),
      skipWhile(data => data === undefined),
      map(mergeConfig.create),
    )
```
It's not directly related to `myths`, but can be helpful in creating `MythicAction`s, like above. Not sure if there's a better place for it. Maybe calling the file `utils.ts` suggests the wrong idea; it's not used within the package, but rather alongside it.

packages/myths/src/epics.ts
CrystallineCat: None, at the moment. I've yet to need this, so I've not implemented anything related to it. I'm planning to tackle this if I run into a case that needs it. TBH, I'd prefer if I didn't need to access non-local state at all to simplify dependencies.

CrystallineCat:
Thank you for your review! I'll add some documentation for the changes on my next commit.

CrystallineCat:
Ok, I'm done with the changes. I hope the documentation is readable. :cat:

packages/myths/README.md
```diff
@@ -58,9 +58,44 @@ const configureStore = makeConfigureStore<NonPrivateState>()({
...
+        when: action => action.error ?? false,
+        dispatch: (action, state, addToSum_) =>
+          of(addToSum_.create(-state.get("sum") / 2)),
+      },
+    ],
+
+  });
+```
+
+Use `thenDispatch: []` to define actions then should be generated from actions of the type being defined, and `andAlso: []` to
```
captainsafia: They are guaranteed to be dispatched to the Redux store, correct? Perhaps it would be helpful to be explicit about this.

```suggestion
Use `thenDispatch: []` to define actions then should be dispatched when actions of the type defined are dispatched, and `andAlso: []` to
```

captainsafia (approved):
Docs look good. I left one small suggestion for the README.

Once things are in a state you like with the myths package, we can write some more long-form docs for docs.nteract.io.

Issue #4963, opened 2020-02-28
Paste images into Markdown cells


riedel:
One nice application of computational notebooks is structural engineering. Here ppl use complicated commercial programs to derive results. Often those results are included as (even as vector?) images. One example of a commonly used program (in Germany) is here: https://www.yumpu.com/de/document/read/5958801/bundige-fussplatte-mit-schubdollen-bautext

A major showstopper is that currently you cannot easily include images from screenshots and copy & paste. The paste handler could generate data URLs or save the image under a temporary file name. There vscoder or atom plugins for this that could be integrated, I guess: https://atom.io/packages/markdown-image-paste, https://marketplace.visualstudio.com/items?itemName=njLeonZhang.markdown-image-paste

An alternatively one would need to save images first, which is a horror for a natural workflow. I think it would be a real benefit for some usecases

willingc:
Thanks for the suggestions @riedel. Do you have any interest in creating this helpful feature?

riedel:
Sorry for answering late.  But when posting this I actually looked at the code and found the developer documentation a bit confusing. It talks a lot about the SDK but not about how to contribute and extend the standalone version. Probably it is intuitive to someone who is used to develop electron applications (with lerna??) . Further the code structure is not completely clear to me. https://github.com/nteract/nteract/blob/master/CONTRIBUTING.md is very generic. 

Actually this also does not sound too encouraging:
> You'll need knowledge of JavaScript (ES6), React, RxJS, Redux, and TypeScript to help out with this project. 

Sorry!

fiedl:
@riedel @willingc @captainsafia 

I'm considering to implement this in the context of https://github.com/nteract/nteract/issues/5081. But I do have a question concerning the expected behaviour:

If you copy an image and paste it into the notebook, what would you expect to happen?

- (a) The image files stay at their original locations and are just linked into the notebook. The image files would be linked using relative paths if they reside within the notebook's directory. They would be linked using absolute paths if they reside outside the notebook's directory.
- (b) The image files are copied into the notebook's directory and linked using relative paths.
- (c) The image files are embedded as attachment in the ipynb file. — I'm not entirely sure if this is possible.

captainsafia:
I'm in favor of approach (a) for the following reasons.

* (c) is not possible currently since we don't support parsing cell attachments when we deserialize the notebook.
[...]

PR #5113, opened 2020-05-17
[Work in progress] Dropping or pasting images into the notebook

### LINKED ISSUE #5081: Feature request: Dropping image onto the ui creates markdown cell to display the image
I would like to be able to drag-and-drop an image file (png/jpg/...) into the nteract ui to display the image in the notebook. This could be used, for example, for sketches, schematics, or other resources.

drop-image-onto-nteract

Possible approach: Listen to the drop event, and create a new markdown cell before the cell the image has been dropped upon. The new markdown cell get the content

    

where `/path/to/the/image` is the path to the image from the drop event.

I would be willing to invest some time to implement this, but I would need some pointers:

- Where would I add the event listener?

  ```js
  document.addEventListener('drop', function(event) {});
  ```

- What is the api to add a markdown cell with a certain content?

- How could I add the new markdown cell *before* the cell the drop has occurred on?

Any help would be very much appreciated. Thanks!

fiedl:
This pull request adds new methods to quickly add images to notebooks.

As I'm fairly new to typescript, I would welcome your suggestions, especially regarding the applied patterns.

This pull request resolves [#5081 Feature request: Dropping image onto the ui creates markdown cell to display the image](https://github.com/nteract/nteract/issues/5081) and possibly [#4963 Paste images into Markdown cells](https://github.com/nteract/nteract/issues/4963).

This pull request adds the [plist](https://github.com/TooTallNate/plist.js) package as dependency in order to parse the clipboard content for file paths on macOS.

Special thanks to @captainsafia and @MSeal for your help!

## Dropping images

2020-05-18 nteract drag-to-embed

Dragging image files into the notebook creates new markdown cells, which contain the image as base64 source.

When dropping several images at once, each image creates its own markdown cell, because it would be hard to distinguish the images when editing the markdown cell due to the long base64 strings.

When sharing notebooks, the shared notebooks will also contain the embedded images.

## Dropping images with copy modifier

2020-05-18 nteract drag-to-copy

Dragging image files into the notebook while holding down the copy modifier key (<kbd>⌥</kbd> on macOS) copies the image files into the notebook directory and creates a new markdown cell, which contains references to the images.

The images are referred to using a relative path.

Using this method, only one new markdown cell is created rather than one for each image.

If image files occupying the same file name already exist within the notebook directory, a notification is shown giving the user the option to replace the existing file, or to dismiss the operation.

2020-05-18 nteract drag-to-replace

When sharing notebooks, one has to make sure also to share the image files, because the images are not embedded as blob within the notebook itself.

## Dropping images with link modifier

2020-05-18 nteract drag-to-link

Dragging image files into the notebook while holding down the link modifier key (<kbd>control</kbd> in macOS) creates a new markdown cell, which contains references to the images.

The images are referred to using their original paths in the file system. They are not copied into the notebook directory using this method.

When sharing notebooks, one has to make sure also to share the image files, because the images are not embedded as blob within the notebook itself.

## Pasting image files from the clipboard

2020-05-18 nteract paste-file

When copying one or several image files into the clipboard, they can be pasted into the notebook using the "paste" command or the corresponding keyboard shortcut (<kbd>command</kbd>+<kbd>V</kbd> in macOS).

This creates new markdown cells, which contain the image as base64 source.

When sharing notebooks, the shared notebooks will also contain the embedded images.

## Pasting an image from the clipboard

2020-05-18 nteract paste-image

When the clipboard contains an image blob, the image can be pasted into the notebook using the "paste" command or the corresponding keyboard shortcut (<kbd>command</kbd>+<kbd>V</kbd> in macOS).

This creates a new markdown cell, which contains the image as base64 source.

When sharing notebooks, the shared notebooks will also contain the embedded image.


fiedl:
@captainsafia May I ask for your help again? I was able to [resolve](https://github.com/nteract/nteract/pull/5113/commits/7ba80c7da622f9b1cb86d963ff9cace870e833ed) some of the issues suggested by `yarn build:all:ci`. But I'm unsure how to resolve these two:

```
[2020-05-18 01:52:38] fiedl@fiedl-mbp ~/code/nteract fiedl/drop-images ⚡
▶ yarn build:all:ci

applications/desktop/src/notebook/global-events.ts:6:24 - error TS7016: Could not find a declaration file for module 'plist'. '/Users/fiedl/code/nteract/applications/desktop/node_modules/plist/index.js' implicitly has an 'any' type.
  Try `npm install @types/plist` if it exists or add a new declaration (.d.ts) file containing `declare module 'plist';`

6 import * as plist from "plist";
                         ~~~~~~~

applications/desktop/src/notebook/global-events.ts:131:55 - error TS2345: Argument of type 'Event' is not assignable to parameter of type 'ClipboardEvent'.
  Property 'clipboardData' is missing in type 'Event' but required in type 'ClipboardEvent'.

131   window.addEventListener('paste', (event) => onPaste(event, contentRef, store));
                                                          ~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:3689:14
    3689     readonly clipboardData: DataTransfer | null;
                      ~~~~~~~~~~~~~
    'clipboardData' is declared here.
```

Details: https://circleci.com/gh/nteract/nteract/7628?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

captainsafia:
> applications/desktop/src/notebook/global-events.ts:6:24 - error TS7016: Could not find a declaration file for module 'plist'. '/Users/fiedl/code/nteract/applications/desktop/node_modules/plist/index.js' implicitly has an 'any' type.
  Try `npm install @types/plist` if it exists or add a new declaration (.d.ts) file containing `declare module 'plist';`


Installing the @types/pslist dependency per the instructions should do the trick here. You'll have to use yarn though: `yarn workspace nteract install @types/plist --dev`.

> Property 'clipboardData' is missing in type 'Event' but required in type 'ClipboardEvent'.

Try setting the `event` type in the outer function to `ClipboardEvent` as in `window.addEventListener('paste', (event: ClipboardEvent) => onPaste(event, contentRef, store))`

MSeal:
I really like the approach you took to the functionality side of control modifiers for setting the image paste behavior. I'm assuming these modifiers translate to Cntl- key combinations that are similar for Linux/Windows? We likely want to document how this works as well as this is a nice hidden gem others will probably not easily discover how to control on their own.

fiedl:
From the documentation of [`DataTransfer.dropEffect`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect), I‘m assuming that the platform-specific key bindings would apply, e.g. <kbd>Ctrl</kbd> on Linux and Windows for a copy operation.

> dropEffect will be initialized based on what action the user is requesting. How this is determined is platform specific, but typically the user can press modifier keys such as the alt key to adjust the desired action.

I‘ve tested the current implementation in tutorial sessions with students in the meantime. The behavior works very well so far, except for two issues:

- Occasionally the current cell appears to lose its focus while dragging in an image, causing the image cell to be created at the beginning of the notebook. For longer notebooks this is really annoying, because one then has to cut and paste the cell to the proper location within the notebook. I will have to look into how to maybe get the cell the user is dropping the image onto from the event and insert the image right before this cell (or something similar).

- The other issue is that the base64 images increase the time that is needed to save the notebook very much, which is really annoying. A notebook with around 1MB of embedded images takes around 90 seconds to save. We really need to resolve this before introducing this feature.

  How are images that are created by python, e.g. plots, saved in the notebook? Can we adopt this mechanism here? Because I‘ve never seen long saving times there.

Thanks!

captainsafia:
> How are images that are created by python, e.g. plots, saved in the notebook? Can we adopt this mechanism here? Because I‘ve never seen long saving times there.

Unfortunately, most visualizations end up being embedded as base64 images. nteract maintains an in-memory model of the notebook that is serialized to JSON so I suspect that the issue is in the time it takes to serialize the large in-memory representation to a JSON string that is stored in the notebook. There are several options here:

- (Best option, imo): Since the base64 image is a string, we can compress it using a string compression technique to reduce the size of the string that is stored on disk and in memory.
- (Meh option): Use the approach of copying over the image file to the appropriate directory.
- (Most expensive option, worth exploring but probably seperately): Improve the performance of the serialization and saving step

MSeal:
> (Best option, imo): Since the base64 image is a string, we can compress it using a string compression technique to reduce the size of the string that is stored on disk and in memory.

lz-string can do this well -- however you should take into account the fact it should still be loadable in other front-ends if possible. Saving as a new media type of lz-base64-png might be in order and let other front-ends similarly adopt decompression for that type over time.

fiedl:
> How are images that are created by python, e.g. plots, saved in the notebook? Can we adopt this mechanism here? Because I‘ve never seen long saving times there.

I've just tested this:

1. Prepare a test image `test.png` with about 1 MB file size.
2. Embed this image in a notebook using

   ```python
   from IPython.display import Image
   Image(filename = "test.png")
   ```

   Execute the cell and save the notebook.

   There is no noticeable time delay saving the notebook. The notebook file size is about 1.4 MB.

   The image is persistently stored within the notebook.

3. Embed the same image as base-64 source inside a markdown cell.

   ```html
   <img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAABZAAAAPfCAY ...
   ```

   Save the notebook.

   Saving the notebook this time took about 14 minutes. During this time, the notebook ui is frozen. The notebook filesize is about 1.4 MB as well.

   The image is persistently stored within the notebook.

So, in both cases, the image is persistently stored within the notebook. But only when embedding the image as base-64 string within a markdown cell [3], saving the notebook takes a lot of time.

I'm looking into how [2] stores the image data within the notebook.

```json
# image-test.ipynb

{
  "cells": [
    {
      "cell_type": "code",
      "source": [
        "from IPython.display import Image\n",
        "Image(filename = \"test.png\")"
      ],
      "outputs": [
        {
          "output_type": "execute_result",
          "execution_count": 1,
          "data": {
            "image/png": [
              "iVBORw0KGgoAAAANSUhEUgAABZAAAAPfCAY

...
``` 

I'm thinking about how we could best apply the same method [2] of saving the data to this pull request.

Is there an api to create a cell directly from a json source?

Then I could just create a cell like this:

```json
{
  "cell_type": "code",
  "source": [],
  "outputs": [
    {
      "output_type": "execute_result",
      "data": {
        "image/png": [
          "iVBORw0KGgoAAAANSUhEUgAABZAAAAPfCAY ..."
        ],
        "text/plain": [
          "<IPython.core.display.Image object>"
        ]
      }
    }
  ],
  "execution_count": 1,
  "metadata": {
    "collapsed": true,
    "outputExpanded": false,
    "jupyter": {
      "source_hidden": true,
      "outputs_hidden": false
    },
  }
}
```

This way, the image would be displayed, persistently stored, and saving would not cause a time delay.

Do you see any downsides to this approach?

fiedl:
@captainsafia [wrote](https://github.com/nteract/nteract/issues/5081#issuecomment-627090928):

> after another conversation I had with a dev I'm thinking that the createCellBelow and createCellAbove actions should be set up to take an ImmutableCell object. I can submit a PR for this.

Could this help in this situation as well, @captainsafia, where I would like to add the "outputs" "data" by hand?

From [commutable/src/cells.ts](https://github.com/nteract/nteract/blob/master/packages/commutable/src/cells.ts), I was thinking of something like this, maybe: (But I'm unsure of the initialisation syntax)

```js
let newCell = ImmutableCodeCell({
  "source": [],
  "outputs": [
    {
      "output_type": "execute_result",
      "data": {
        "image/png": [
          "iVBORw0KGgoAAAANSUhEUgAABZAAAAPfCAY ..."
        ],
  }}]
});

store.dispatch(
  actions.createCellAbove(newCell)
);
```

fiedl:
Thanks, @captainsafia, for https://github.com/nteract/nteract/pull/5102!

fiedl:
Creating new markdown or code cells with the new `createCellAbove` from https://github.com/nteract/nteract/pull/5102 works fine, thanks! I think, we're getting close! However, I'm currently unsuccessful in creating cells with image output.

For example, I've tried:

```js
  let newCell = emptyCodeCell
    .set("source", "")
    .set("outputs", Immutable.fromJS([
      {
        "output_type": "display_data",
        "data": {
          [imageType]: [
            base64Image + "\n"
          ],
          "text/plain": [
            "<IPython.core.display.Image object>"
          ]
        },
        "metadata": Immutable.Map()
      }
    ]))
    .set("metadata", Immutable.fromJS({
      "collapsed": true,
      "outputExpanded": false,
      "jupyter": {
        "source_hidden": true,
        "output_hidden": false
      }
    }));
```

When adding this cell, the `source_hidden` parameter takes effect. The image, however, is not displayed.

Furthermore, I've tried to save the notebook in order to inspect the the stored content for debugging. But the notebook can't be saved as I'm getting:

> ```
> Uncaught (in promise) TypeError: You provided an invalid object where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
>     at subscribeTo (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/util/subscribeTo.js?:39)
>     at subscribeToResult (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/util/subscribeToResult.js?:20)
>     at MergeMapSubscriber._innerSub (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/operators/mergeMap.js?:84)
>     at MergeMapSubscriber._tryNext (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/operators/mergeMap.js?:78)
>     at MergeMapSubscriber._next (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/operators/mergeMap.js?:61)
>     at MergeMapSubscriber.Subscriber.next (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/Subscriber.js?:63)
>     at FilterSubscriber._next (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/operators/filter.js?:42)
>     at FilterSubscriber.Subscriber.next (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/Subscriber.js?:63)
>     at Notification.observe (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/Notification.js?:27)
>     at QueueAction.ObserveOnSubscriber.dispatch (webpack:////Users/fiedl/code/nteract/node_modules/rxjs/_esm5/internal/operators/observeOn.js?:48)
> ```

I'm guessing that I must have done something wrong initializing the new cell. But I'm sorry, I don't see how to approach this. Can you see where this goes wrong?

captainsafia:
> I'm guessing that I must have done something wrong initializing the new cell. But I'm sorry, I don't see how to approach this. Can you see where this goes wrong?

I don't think this is related to the changes you made but might be related to a commit that you fetched (#5096 ) when rebasing to get my create cell changes.

- Are you testing on Windows?
- Did this issue occur with saving before pulling in the latest changes?
- Does a `yarn clean` and `yarn` resolve the issue?

EDIT: I just pushed another fix that might resolve this. Try now?

fiedl:
> Are you testing on Windows?

I'm on macOS.

>  I just pushed another fix that might resolve this. Try now?

Thanks, this resolved the saving issue.

However, I'm having trouble inserting even the most simple code cell with output.

- Inserting a markdown cell works fine.

  ```js
  let newCell = emptyMarkdownCell.set("source", "foo");
  store.dispatch(actions.createCellAbove({contentRef: contentRef, cell: newCell}));
  ``` 

  The markdown content is automatically rendered.

- Inserting this simple code cell works, but the output is not displayed.

  ```js
  let newCell = emptyCodeCell.set("source", "foo").set("outputs", ["foo"]);
  store.dispatch(actions.createCellAbove({contentRef: contentRef, cell: newCell}));
  ```

  Also, this cannot be saved due to:

  > immCell.outputs.map(...).toArray is not a function

- Inserting this code cell with a little more structure works, but the output is not be displayed:

  ```js
  let newCell = emptyCodeCell.set("source", "foo").set("outputs", Immutable.fromJS(
    [
      {
        "output_type": "stream",
        "name": "stdout",
        "text": ["hello foo\n"]
      }
    ]
  ));
  store.dispatch(actions.createCellAbove({contentRef: contentRef, cell: newCell}));
  ```

  Saving does work, but the saved notebook does not contain the output:

  ```js
  // test.ipynb
  "cells": [
    {
      "cell_type": "code",
      "source": [
        "foo"
      ],
      "outputs": [
        null
      ],
  ```

Do I need to trigger some kind of processing or rendering of code cells manually?

captainsafia:
@fiedl @nteract/commutable provides some APIs for creating cell outputs.

In particular, you can use `createImmutableOutput` to create the appropriate output.

```js
import { createImmutableOutput } from "@nteract/commutable";

let newCell = emptyCodeCell.set("source", "foo").set("outputs", createImmutableOutput({ output_type: "stream" ... })));
```

You might also want to use `push` instead of `set` to avoid overriding whatever outputs exist in the cell.

fiedl:
Thank you, this works! 🎉 

```js
let newCell = emptyCodeCell
  .set("source", "foo")
  .set("outputs", Immutable.List([
    createImmutableOutput({
      output_type: "stream",
      name: "stdout",
      text: "hello foo"
    })
  ]));

store.dispatch(actions.createCellAbove({contentRef: contentRef, cell: newCell}));
```

fiedl:
I've just tried to adopt this for adding images.

- The image is displayed after dropping it in.
- The image is persistently saved.
- Saving the notebook with embedded image does not add any noticeable time delay 👍 

But: The image is displayed within an output box with scroll bar and thereby cut off. Do you know how to get rid of that? (Neither `collapsed` nor `outputExpanded` has any effect.)

Bildschirmfoto 2020-05-30 um 03 25 39

fiedl:
Strange: nteract v0.23.1 does not cut off the image.

Bildschirmfoto 2020-05-30 um 13 08 58

Might be related to https://github.com/nteract/nteract/pull/5085?

Clicking "Toggle Expanded Output" in the ui of the cell does not help. When having the output expanded, the `.nteract-cell-outputs` element does get an extra `expanded` class. But, according to the inspector, the `.nteract-cell-outputs.expanded` element still has `max-height: 300px`, which seems to overrule the `height: 100%`.

EDIT: Yep, this fixes the issue:

```
.nteract-cell-outputs.expanded {
  height: 100%;
  max-height: 100%;
}
```

captainsafia:
Thanks for fixing the unrelated scroll bar issue, @fiedl. Do you think we're ready to take the WIP label off this PR and do a full review or is there anything else that you'd like to spruce up?

fiedl:
@captainsafia I think the core functionality is done, but there are a couple of things I wanted to look into before requesting a review.

- [ ] Check if the new cells can be inserted before the cell the user is dropping the images onto (rather than before the currently selected cell).

  Do you know if there is a way to get the cell that corresponds to a DOM object, e.g. the `.nteract-cell`, and select it as "current cell" before calling `insertCellAbove`?

  > Occasionally the current cell appears to lose its focus while dragging in an image, causing the image cell to be created at the beginning of the notebook. For longer notebooks this is really annoying, because one then has to cut and paste the cell to the proper location within the notebook. I will have to look into how to maybe get the cell the user is dropping the image onto from the event and insert the image right before this cell (or something similar). — https://github.com/nteract/nteract/pull/5113#issuecomment-632085169
  
- [ ] I have not looked into testing, yet. I'm unfamiliar to jest, yet. So I don't now if there is a good way to test pasting and dropping.

- [ ] I've tried the new feature on macOS. I wanted to give it a try under linux before requesting a review. Unfortunately, I can't test it on Windows at the moment. Does CI cover multiple operating systems?

  I think there might be operating-system differences when pasting multiple images from the clipboard, because I'm currently relying on [`NSFilenamesPboardType`](https://github.com/nteract/nteract/issues/4963#issuecomment-627561034) for that. For dropping images and pasting a single image, it should be the same for all operating systems.

- [x] There is also a menu item "Paste". Check if this is already the same as the paste event in `global-events.ts`. Otherwise they may need to be linked somehow.

- [ ] Also, I had no time to look into [refetching images after copy-and-replacing them](https://github.com/nteract/nteract/issues/5081#issuecomment-629897711). But I would regard that as nice-to-have only.

If there are timeline requirements, feel free to skip these issues and merge at your discretion :)

fiedl:
Drag&drop on ubuntu works fine; however pasting from the clipboard (one file, several files, image data) does not work.

Screenshot from 2020-05-31 22-47-54

fiedl:
I've had a look at the clipboard data on ubuntu using [copyq](https://github.com/hluk/CopyQ).

When copying one or several files on ubuntu 20.04, the clipboard contains `text/uri-list` with:

```
file:///home/fiedl/Desktop/Screenshot from 2020-05-31 22-47-22.png
file:///home/fiedl/Desktop/Screenshot from 2020-05-31 22-47-54.png
file:///home/fiedl/Desktop/Screenshot from 2020-06-01 23-37-02.png
```

When copying image data, the clipboard contains `image/png`, `image/tiff`, as well as other formats. It does not have the `public.tiff` key as on macOS.

fiedl:
I've modified the code to also consider the clipboard keys used by ubuntu. This, by itself, works nicely.

However, I somehow can't access the file system on ubuntu with `fs` (which is needed to get the base64 content of image files).

> ```
> Uncaught Error: ENOENT: no such file or directory, open '/home/fiedl/Screenshot.png'
> at Object.fs.readFileSync
> ```

The path itself is correct and when trying this out in an interactive node console

```
$ node --version
v10.19.0

$ node
fs.readFileSync("/home/fiedl/Screenshot.png")
<Buffer ...>
```

it does work. Is there an extra layer of security in electron on ubuntu? Or might the node version of electron be a different version than the one used from the console?

MSeal:
And `ls` from that same terminal show the file is there? The only references I could find to issues specific to Ubuntu for this were on really old versions of node :/

captainsafia:
@fiedl To make things a little bit easier to review, I'd recommend we create a feature branch for this work and merge smaller PRs into that feature branch.

We can factor out each of the tasks that you listed [here](https://github.com/nteract/nteract/pull/5113#issuecomment-636515409) into an issue and tackle them bit by bit.

If that makes sense to you, I'll create a feature branch and change the target branch here then we can review this change in its current state.

The review process might also reveal some other things we need to do.

fiedl:
I'm comfortable with both ways, @captainsafia 

- (a) consider this pull request as ready for review now; remove the "work-in-progress" state;  and we'll make the additional stuff (linux support, etc.) separate issues against master

- (b) create a longer-term feature branch as target for this pull request; and we'll make the additional stuff separate issues against this feature branch

If you're happy with this pull request being merged into master in it's current state, (a) might be the easier way :)

captainsafia:
Let's go with (b). I'll give this a review later today. Thanks!

applications/desktop/src/notebook/global-events.ts
```diff
@@ -35,6 +39,65 @@ export function onBeforeUnloadOrReload(
...
+  let imagePaths = ((event.dataTransfer) ? Array.from(event.dataTransfer.files) : [])
+    .filter(file => file.type.match(/image.*/))
+    .map(file => file.path);
+
+  let copyImagesToNotebookDirectory: boolean = (event.dataTransfer?.effectAllowed == "copy");
+  let linkImagesAndKeepAtOriginalPath: boolean = (event.dataTransfer?.effectAllowed == "link");
+  let embedImagesInNotebook: boolean = !(copyImagesToNotebookDirectory || linkImagesAndKeepAtOriginalPath);
+
+  insertImages({
+    imagePaths: imagePaths,
```
captainsafia: Nit: We can use the shorthand `imagePaths` instead of `imagePaths: imagePaths` here.

applications/desktop/src/notebook/global-events.ts
```diff
@@ -35,6 +39,65 @@ export function onBeforeUnloadOrReload(
...
+      .filter(filePath => /[\w-]+\.(png|jpg|jpeg|heic|gif|tiff)/.test(filePath));
+
+    insertImages({
+      imagePaths: imagePaths,
+      embedImagesInNotebook: true,
+      contentRef: contentRef,
+      store: store
+    });
+
+    event.preventDefault();
```
captainsafia: Nit: Place the `event.preventDefault()` at the top of the function body by convention.

applications/desktop/src/notebook/insert-images.ts
```diff
@@ -0,0 +1,210 @@
...
+          [imageType]: [
+            base64Image + "\n"
+          ]
+        },
+        "metadata": {}
+      })
+    ]))
+    .set("metadata", Immutable.fromJS({
+      "collapsed": false,
+      "outputExpanded": true,
```
captainsafia: The `outputExpanded` flag is deprecated so you don't need to provide it anymore. Just setting collapsed should do the trick.

captainsafia (commented):
Posted some comments inline.

I gave this a go on macOS and it functions as expected.

I also tried it in my Windows VM. I dragged a file from Explorer to nteract and a new code cell was created.

However, the image was not displayed in the output. When I inspected the store, it looked like the output was included in the programmatically generated cell, but nothing was showing up in the DOM.

I edited the cell and made a few other modifications to trigger a re-render in React and it was still empty.

This seems weird because the renderer is the same across Windows and macOS. There's some funkiness here.

@jdetle Can you verify this on your Windows machine to sanity-check?

applications/desktop/src/notebook/global-events.ts
fiedl: Cool, I didn't know that.

applications/desktop/src/notebook/global-events.ts
fiedl: Is it alright if I move the `event.preventDefault()` to the top of the `if` blocks rather than to the top of the function body? I thought I should only prevent the default if the clipboard has the `NSFilenamesPboardType` or `public.tiff` keys. Otherwise, this function should not handle the event.

fiedl:
@captainsafia Thanks for the review and your helpful suggestions!

applications/desktop/src/notebook/global-events.ts
captainsafia: Yep -- that works.

captainsafia (approved):
I create https://github.com/nteract/nteract/issues/5148 to track follow-up work on this feature branch.

I'm OK with merging this into the feature branch and tackling the follow-up items now.

Give me the green light, @fiedl and I can merge.

fiedl:
Sure, thanks!

captainsafia:
Merging this into the feature branch. Onward and upward!

Issue #3364, opened 2018-09-25
SageMath: LaTex rendering of output not showing in nteract


dkumor:
I am on Arch Linux, running latest nteract from AUR, along with latest jupyter.

When running with the SageMath 8.3 kernel in Jupyter notebook and Jupyter lab, equations are printed with MathJax if I either have:
```
%display latex
```
or use the `pretty_print` function. 

This does not happen in nteract: the above run without error, but do not output anything. That is, there is no output to the cell, when there should be a MathJax equation. Likewise, nteract does not display equations in existing sage notebooks.

To reproduce: start SageMath kernel in nteract, and run:

```
%display latex
x^2
```
Nothing is shown. When running a notebook without `%display latex`, a non-latex output `x^2` is shown.

Furthermore: the notebook created by nteract doesn't show the equations when displayed in a standard jupyter notebook. I need to re-run the cell for the math to show up.

I have attached a zip file with 2 notebooks: [notebooks.zip](https://github.com/nteract/nteract/files/2417449/notebooks.zip)
1) nteract.ipynb - the notebook created by nteract which does not display the math in nteract nor jupyter itself.
2) notebook.ipynb - the same notebook which had  the cell re-run in jupyter notebook. The math now shows in jupyter, but not in nteract.

What's weird is that I can't see any reason for this in the notebook code itself- both seem to be identical in cell contents...

willingc:
Hi @dkumor,

Thanks for the report. The latest release is [0.12.3](https://github.com/nteract/nteract/releases/tag/v0.12.3). If you are able to try with the latest release, please let us know if this still is happening.

dkumor:
I have tried with the latest release of nteract and SageMath 8.4 - there was no change in the issue. The same thing happens on two separate machines (both running arch). 

I have since noticed that JupyterLab needs to re-run cells for existing math to appear. nteract does not show the math at all, no matter what is done.

willingc:
Thanks for giving it a try and sharing your results.

captainsafia:
Re-opening this issue as it still occurs.

Upon further inspection, it seems that the issue with it not displaying at all might be related to the use of the `%display` magic.

When I tried the following, I was able to see the output.

```
%%latex
x^2
```

However, even though it was rendered with our MathJax component, it was still not displayed in formatted LaTeX. I suspect this is due to the MathJax component not working.

We'll need to get a test notebook with more LaTeX examples and run more validations to see what is going on here.

[...]

PR #5128, opened 2020-05-26
Try loading mathjax asynchronously and delay

willingc:
When loading and configuring mathjax, this PR adds async loading and tests whether changing
`delayStartupUntil` from `configure` to `onload`
https://docs.mathjax.org/en/v2.7-latest/configuration.html#configuring-mathjax-after-it-is-loaded

Will fix the reformatting done by VSCode


willingc:
Partially addresses: #3364

captainsafia (approved):
Interesting. We should make this change in [mathjax-electron](https://github.com/nteract/mathjax-electron) to bring this improvement to the desktop app as well.

rgbkrk:
Thanks all!

PR #5137, opened 2020-06-03
Centralize configuration as mythic-configuration

### LINKED ISSUE #5092: Keyboard shortcuts should be configurable

### LINKED ISSUE #5134: Does not save the theme
**Application**
nteract desktop

**Describe the bug**
nteract does not save the theme

**To Reproduce**

1. Preferences -> Theme -> Dark
2. Restart app

**Expected behavior**
I was expecting a dark theme:(

**Desktop:**

- OS: macOS and Ubuntu
- Version: 0.23.3

**Additional context**
- This problem also exists in version 0.23.2, but not in 0.23.1. 
- nteract saves theme settings to the nteract.json file, but ignores the theme parameter at startup

CrystallineCat:
Progress on #4799. Last part of #5031 :cat:. Fixes #5134. Fixes #5092.

Implements the `mythic-configuration` package. See #5031 for details.


applications/desktop/src/main/index.ts
```diff
@@ -70,41 +77,17 @@ const fullAppReady$ = zip(electronReady$, prepareEnv).pipe(first());
...
 
 const prepJupyterObservable = prepareEnv.pipe(
   mergeMap(() =>
     // Create all the directories we need in parallel
     forkJoin(
       // Ensure the runtime Dir is setup for kernels
       mkdirpObservable(jupyterPaths.runtimeDir()),
-      // Ensure the config directory is all set up
-      mkdirpObservable(jupyterConfigDir)
+      // The config directory is taken care of by the configuration myths
```
captainsafia, line 88: Wonderful! I love that myths enable us to clean-up a lot of the cruft that we have lingering around in the desktop app.

applications/jupyter-extension/nteract_on_jupyter/app/store.ts
```diff
@@ -14,7 +14,6 @@ export const configureStore = makeConfigureStore<AppState>()({
   reducers: {
     app: reducers.app,
     comms: reducers.comms,
-    config: reducers.config,
```
captainsafia, line 17: Is there a reason we don't add the configuration myth here?

For people who are using the nteract core SDK to build their own web apps, we'll need to have a migration story from the config being stored in the main state to being managed by a myth.

The Jupyter extension is often the best place to show an example of this.

package.json
```diff
@@ -32,6 +32,7 @@
     "lint:ts:fix": "tslint --config tslint.json --project tsconfig.base.json --fix",
     "build:all": "tsc -b",
     "build:all:ci": "tsc -b --verbose",
+    "build:all:cleanly": "npm run clean && yarn && npm run build:all",
```
captainsafia: Nit: Use `yarn` for all of these commands.

packages/epics/src/contents.ts
```diff
@@ -155,12 +137,21 @@ export function downloadString(
   FileSaver.saveAs(blob, filename);
 }
 
+// FIXME: should be createConfigOption(), but that gets called twice during some
+//        tests... hence the abuse of below function, which allows duplicate
+//        definitions
+const { selector: autoSaveInterval } = transferConfigOptionFromRenderer({
```
captainsafia: Why does `createConfigOption` get called twice?

This epic is part of the SDK so it may be running in non-Electron environments where there is no renderer/main process distinction.

packages/mythic-configuration/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```javascript
+import { configuration } from "@nteract/mythic-configuration";
+import { makeConfigureStore } from "@nteract/myths";
+
+export const configureStore = makeConfigureStore({
+  packages: [configuration],
+});
+
+store.dispatch(setConfigFile("/etc/app.conf"));
```
captainsafia, line 28: We'll need to support getting a config from a file and from a string for non-desktop app scenarios.

packages/mythic-configuration/README.md
```diff
@@ -0,0 +1,93 @@
...
+```javascript
+export const {
+  selector: defaultKernel,
+  action: setDefaultKernel,
+} = createConfigOption({
+  key: "defaultKernel",
+  label: "Default kernel on startup",
+  valuesFrom: "kernelspecs",
+  defaultValue: "python3",
+});
```
captainsafia, line 48: Nit: Might want to add another option to the docs that showcases creating a config option that doesn't have a `valuesFrom` property.

My initial thinking when I saw this code sample was that `valuesFrom` could be provided to any type of config but no-ops in all cases but "kernelSpecs".

packages/mythic-configuration/src/backends/filesystem.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Map } from "immutable";
+import * as path from "path";
+import { concat, of } from "rxjs";
+import { ignoreElements, map, mapTo, skipWhile } from "rxjs/operators";
+import { ConfigurationBackend } from "..";
+import { loadConfig } from "../myths/load-config";
+import { mergeConfig } from "../myths/merge-config";
+import { setConfigBackend } from "../myths/set-config-backend";
+
+const filesystemConfigurationBackend = (filename: string) => ({
```
captainsafia, line 12: Mentioned above but we should consider having an in-memory configuration provider as well.

packages/mythic-configuration/src/backends/filesystem.ts
```diff
@@ -0,0 +1,43 @@
...
+    readFileObservable(filename).pipe(
+      mapErrorTo("{}", err => err.code === "ENOENT"),
+      map(data => JSON.parse(data.toString())),
+      // SyntaxError means the file is probably in the middle of a write
+      mapErrorTo(undefined, err => err.name === "SyntaxError"),
+      skipWhile(data => data === undefined),
+      map(mergeConfig.create),
+    ),
+
+  save: (current: Map<string, any>) =>
```
captainsafia, line 31: I like the fact that we have some nicely-named API methods for interacting with the config file.

packages/mythic-configuration/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      state => state?.current?.getIn(props.key.split("."))
+        ?? props.defaultValue,
+    ),
+    action: (value: TYPE) => setConfigAtKey.create({ key: props.key, value }),
+  };
+
+  return options[props.key];
+};
+
+export const createDeprecatedConfigOption = <TYPE>(
```
captainsafia, line 63: What is creating a deprecated config options helpful for?

applications/desktop/src/main/index.ts
```diff
@@ -40,6 +42,11 @@ log.info("args", argv);
 
 const notebooks = argv._.filter((x) => /(.ipynb)$/.test(x));
 
+ipc.on("transfer-config-options-to-main",
```
captainsafia, line 45: Not sure where it is happening but when quitting the app I get a couple of uncaught exceptions bubbling from this file.

Screen Shot 2020-06-03 at 4 40 17 PM
Screen Shot 2020-06-03 at 4 42 21 PM

packages/mythic-configuration/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "keywords": [
+    "nteract"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+  },
+  "dependencies": {
+    "@nteract/myths": "^0.1.0",
```
captainsafia, line 19: Should we recommend that people include @nteract/myths as a peer dependency when developing with this pattern?

packages/notebook-app-component/src/decorators/undoable/undoable-cell-delete.tsx
```diff
@@ -1,11 +1,19 @@
...
 import React from "react";
 import { connect } from "react-redux";
 import { Dispatch } from "redux";
 import styled from "styled-components";
-
 import UndoableDelete from "./undoable-delete";
 
+export const {
+  selector: deleteDelay,
+} = createConfigOption({
```
captainsafia: Do we anticipate that for more complex apps a developer will have a `config.ts` file in their nteract application where all the config options are created and then import selectors from that central file?

Trying to get a feel of the dev experience for this.

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
```diff
@@ -1,7 +1,79 @@
-import { connect } from "react-redux";
-
 import { AppState, ContentRef, selectors } from "@nteract/core";
 import CodeMirrorEditor from "@nteract/editor";
+import { createConfigCollection, createConfigOption, createDeprecatedConfigOption, HasPrivateConfigurationState } from "@nteract/mythic-configuration";
+import { connect } from "react-redux";
+
+const codeMirrorConfig = createConfigCollection({
```
captainsafia, line 6: Can `createConfigCollection` be called multiple times within the same application instance?

I assume yes, but just checking.

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
```diff
@@ -64,11 +124,10 @@ const makeMapStateToProps = (state: AppState, ownProps: ComponentProps) => {
       }
     }
 
-    const codeMirror = Object.assign(
-      {},
-      { ...codeMirrorDefaults, mode },
-      codeMirrorCurrentConfig
-    );
+    const codeMirror = {
+      ...mode,
```
captainsafia: Syntax highlighting appears to be broken in this branch. It's trying to pull the mode styles from `text/plain/text/plain.js` instead of `text/plain.js`. Not sure what might be going on here.

captainsafia (commented):
Cool stuff! I left some comments inline. Most of them are how this API will function for users from non-desktop scenarios.

package.json
CrystallineCat: Done. I used `npm` because the other scripts all seem to use it, but `yarn` seems to work just as well.

packages/epics/src/contents.ts
CrystallineCat: I'm not sure why it gets called twice. I changed the duplicate detection so that it only throws if the spec it gets called with is actually different than the stored spec, so calling it twice with the same parameters no longer is a problem.

packages/mythic-configuration/src/backends/filesystem.ts
CrystallineCat: Done.

packages/mythic-configuration/src/index.ts
CrystallineCat: I'm going to add this to the documentation, they are mainly for renaming config keys. In our case, we currently have `cursorBlinkRate` and `codeMirror.cursorBlinkRate` from the CodeMirror options patches. So the following code will make all `cursorBlinkRate` into `codeMirror.cursorBlinkRate`, unless `codeMirror.cursorBlinkRate` is already defined, then delete the `cursorBlinkRate` key.

```ts
createDeprecatedConfigOption({
  key: "cursorBlinkRate",
  changeTo: (value: number) => ({
    "codeMirror.cursorBlinkRate": value,
  }),
});
```

applications/desktop/src/main/index.ts
CrystallineCat: Ah! Hmm... this doesn't happen on here (on Ubuntu), but I think this is because the `notebook`-thread might close too early, since I get:
```
nteract: [228198:0604/112401.457033:WARNING:x11_util.cc(1424)] X error received: serial 128, error_code 3 (BadWindow), request_code 4, minor_code 0 (Unknown)
```
In general, I tend to get plenty of `X` or `Gtk` errors and warnings each run, that I'm hoping that the upgrade of electron in #5115 will fix some of those. Btw, could you please post a comment in that PR stating what still needs to be done to get it merged? I'd like to help tackle that after this merges.

Anyway, I had type errors that looked similar to those errors, so I think I know what the problem is... I'm going to do a quick fix that should hopefully solve it for now, I'll be taking a look at rewriting the windowing code soon.

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
CrystallineCat: Yeah, that one just generates a function that takes `state` and collects all config options with a common prefix, respecting defaults. It doesn't modify global state, so calling it multiple times is fine. I should document the function in the README.

packages/stateful-components/src/inputs/connected-editors/codemirror.tsx
CrystallineCat: Fixed. NB: `mode`'s typing here is wrong, it can be e.g. a simple string. I made a comment, but didn't look in detail why that is the case.

packages/mythic-configuration/package.json
CrystallineCat: What would be the benefit? I don't really know much about deployment, so I can't tell what is right here.

packages/mythic-configuration/README.md
CrystallineCat: That's the job of the back end. We already support arbitrary back ends, so this shouldn't be a problem.

applications/jupyter-extension/nteract_on_jupyter/app/store.ts
CrystallineCat: I don't know how the configuration flow in the extension or the web app is supposed to go. I made a pure in-memory back end, which is used by default and works fine in e.g. the test cases. I could add another back end that e.g. uses `localStorage`, if that's what would make sense for the extension/web app, but I'd prefer if I had a description of how it is supposed to work, as I'm not familiar with either.

If you make a new issue and describe what needs to happen, I'll implement it right after this PR. :cat:

packages/notebook-app-component/src/decorators/undoable/undoable-cell-delete.tsx
CrystallineCat: I've been trying to keep the selector as local as possible. E.g. the auto save interval config option is defined right next to the epic that uses it and the selector isn't even exported. That way the config option is only defined if the file with the epic is executed and there aren't any cross-dependencies.

This might not work for some options, though, so in some cases collecting them in a file might be a better idea.

CrystallineCat:
Thank you for the quick and detailed review! I think I addressed everything.

applications/jupyter-extension/nteract_on_jupyter/app/store.ts
captainsafia: > I don't know how the configuration flow in the extension or the web app is supposed to go. I made a pure in-memory back end, which is used by default and works fine in e.g. the test cases

That implementation mimics what our current in-memory implementation does, reading and writing values to the Redux state but not persisting anything beyond the lifetime of the application.

> I could add another back end that e.g. uses localStorage, if that's what would make sense for the extension/web app, but I'd prefer if I had a description of how it is supposed to work, as I'm not familiar with either.

That's a good idea (although not required for parity with what we already have). It might be good to have myself or someone else try out implementation `LocalStorageConfigurationProvider` so that we can build some familiarity with this API.

package.json
```diff
@@ -222,7 +223,7 @@
     "remark-parse": "^8.0.0",
     "remark-stringify": "^8.0.0",
     "rimraf": "^3.0.0",
-    "rxjs": "^6.3.3",
+    "rxjs": "^6.5.5",
```
captainsafia: Any particular reason for upgrading to `6.5.5` here?

packages/editor/src/index.tsx
```diff
@@ -55,8 +55,6 @@ export type CodeMirrorEditorProps = {
   tip?: boolean;
   focusAbove?: (instance: Editor) => void;
   focusBelow?: (instance: Editor) => void;
-  // _Our_ theme, not the codemirror one we use
-  theme: string;
```
captainsafia, line 59: Nice job spotting this clean-up. We don't actually did to refresh on theme change if it handled by the theming component.

captainsafia (approved):
Ship it! 🚢

It would be good if we had some forthcoming docs on how to migrate to this pattern for people who are current reading/writing config from `state.config` using our current set of selectors and actions.

package.json
CrystallineCat: Oh! Thanks for catching that. When I did a build from clean, I had a compiler error in `node_modules/rxjs`, and I thought upgrading would fix that... which it didn't, but I forgot to revert the change.

Issue #5145, opened 2020-06-04
createOnDiskMediaBundle is very expensive when saving a notebook with large data section


bryanchen-d:
**Application or Package Used**
@nteract/core@13.1.1-alpha.0

**Describe the bug**
Saving a notebook with a large data payload in outputs takes a long time.

**To Reproduce**

1. load a notebook with a display output contains a large data payload, in my case, the size of data is about 250KB.
2. save the notebook
3. Then the UI will be frozen for like 10~ seconds

**Expected behavior**
UI should recover after 1-2 seconds and be responsive.

I did a profiling and figure out that almost all the time is spent in remultiline at https://github.com/nteract/nteract/blob/7983006937e212c4f8a021f159e8dea934ea796b/packages/commutable/src/primitives.ts#L225

As per comments around line 213: 
https://github.com/nteract/nteract/blob/7983006937e212c4f8a021f159e8dea934ea796b/packages/commutable/src/primitives.ts#L212-L215
Can we just return the mediaBundle instead of reformatting the multilines?

captainsafia:
Closed via #5146.

PR #5146, opened 2020-06-05
Skip reformatting mediaBundle

bryanchen-d:
This is to fix issue #5145


captainsafia (approved):
Shipping for perf!

We might be able to strike a balance here between running re-multiline everywhere and making incremental fixes.

PR #5151, opened 2020-06-09
Updates cell styles

ryanlanciaux:
This PR updates the Cell styles. In this paradigm, the Cell actions are no longer in a floating toolbar and are now in a popout menu. 

cell

## TODO
There are a couple of things that need to occur before moving this from a Draft to a final PR.

- [ ] Update tests
- [x] Run the application in more scenarios
- [x] Ensure that the Toolbar does not expand outside the visible area when on the last Cell.


captainsafia:
Gave this a spin. Exciting to see the new changes.

One piece of feedback I have is the interaction with the input is hidden.

Currently, it just hides the text. Typically, users will want to entire input area to be hidden and the toolbar elements to be moved adjacent to the output.

Also, I feel like there should be a bit of a margin in the output so that it lines up with the cell toolbar.

ryanlanciaux:
@captainsafia Thanks a ton for this feedback! Do you think it would make sense if I switched the layout to where the toolbar elements is on the left of both the input and the output? I think that may cover both the margin and condensing things a bit when the input is hidden. 

Thanks!

captainsafia:
>  Do you think it would make sense if I switched the layout to where the toolbar elements is on the left of both the input and the output?

I think I am visualizing the same thing you are with this description and it sounds like a sensible approach.

ryanlanciaux:
Okay great! I've updated this PR with the styling updates so the sidebar stays next to the output when the input is hidden:

image

Additionally, I've updated the color of the sidebar. I'm still seeing some failing tests, and I'm trying to determine how this may have broken them.

packages/presentational-components/package.json
```diff
@@ -9,7 +9,9 @@
   "dependencies": {
     "@blueprintjs/core": "^3.7.0",
     "@blueprintjs/select": "^3.2.0",
-    "react-syntax-highlighter": "^12.0.0"
+    "react-syntax-highlighter": "^12.0.0",
+    "@types/classnames": "^2.2.10",
```
captainsafia: Let's move this to `devDependencies`.

packages/presentational-components/package.json
```diff
@@ -26,9 +28,11 @@
   "license": "BSD-3-Clause",
   "devDependencies": {
     "@types/styled-components": "^4.1.4",
+    "@types/react-toggle": "^4.0.2",
     "color-parse": "^1.3.8",
     "color-space": "^1.16.0",
     "css": "^2.2.4",
-    "styled-components": "^4.4.1"
+    "styled-components": "^4.4.1",
+    "react-toggle": "^4.1.1"
```
captainsafia: This should be under `dependencies` I believe.

packages/notebook-app-component/src/derived-components/toolbar.tsx
```diff
@@ -5,151 +5,340 @@
...
                   className="executeButton"
                 >
-                  <span className="octicon">
-                    <TriangleRightOcticon />
-                  </span>
-                </button>
+                  <Icons.Play />
+                </CircularButton>
               ) : null}
-              <DropdownMenu
```
captainsafia, line 57: We might want to set some tabindexes on the menu items.

I can tab from the `Play` button to the `Toolbar` icon and press enter to show the toolbar. From that point on, I cannot tab through the menu options in the toolbar.

packages/stateful-components/src/cells/toolbar.tsx
```diff
@@ -43,11 +45,9 @@ class CellToolbar extends React.Component<
 > {
   render() {
     return (
-      <div className="nteract-cell-toolbar">
```
captainsafia: The `nteract-cell-toolbar` classname is part of the public API that we provide as part of stateful components so we will need to keep it around.

Was there a particular reason it was removed?

packages/styles/themes/default.css
```diff
@@ -56,19 +56,51 @@
   color: var(--theme-app-fg);
 }
 
-.nteract-cell {
+.nteract-cell{
```
captainsafia:
```suggestion
.nteract-cell {

```

packages/styles/themes/default.css
```diff
@@ -56,19 +56,51 @@
...
+  display: flex;
+  flex-direction: column;
+  background: var(--theme-cell-bg, white);
+  border: 1px solid #EEEFF0;
+  box-shadow: 0px 2px 4px rgba(41, 49, 58, 0.075);
+  border-radius: 4px;
+  font-family: 'Source Code Pro', sans-serif;
+}
+
+.nteract-cell-row{
```
captainsafia:
```suggestion
.nteract-cell-row {

```

packages/styles/themes/default.css
```diff
@@ -56,19 +56,51 @@
...
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+}
+
+.nteract-cell-container:focus .nteract-cell, .nteract-cell-container:focus-within .nteract-cell, .nteract-cell-container.selected .nteract-cell{
```
captainsafia:
```suggestion
.nteract-cell-container:focus .nteract-cell,
.nteract-cell-container:focus-within .nteract-cell,
.nteract-cell-container.selected .nteract-cell {
```

packages/styles/themes/default.css
```diff
@@ -56,19 +56,51 @@
   color: var(--theme-app-fg);
```
captainsafia: Nit: Can you keep this file unchanged and move the styles related to a new theme to a new file?

captainsafia (changes requested):
Looks good overall! This is really close to being mergeable.

One issue I noticed is with markdown cells. The markdown previewer applies its own gutter styling which results in an extra gray area inside the input text.

Screen Shot 2020-06-16 at 9 57 00 PM

packages/notebook-app-component/src/derived-components/toolbar.tsx
ryanlanciaux: I think this should be working now.

packages/stateful-components/src/cells/toolbar.tsx
ryanlanciaux: Sorry about that - there was no real reason. I've added it back in.

packages/styles/themes/default.css
ryanlanciaux: That would definitely make things a bit cleaner. I've created a CSS file just to contain these styles.

captainsafia (approved):
Looks good overall. One last detail I noticed is the color of the toolbar and elements is a little off in dark mode.

This wasn't included in the original design so we might have to be creative with how it ends up looking here.

Screen Shot 2020-06-24 at 1 34 31 AM

captainsafia:
Merging this now so that we can tackle to follow-ons in smaller PRs.

PR #5153, opened 2020-06-09
App Toolbar

ryanlanciaux:
The intent of this PR is to add a new Application level toolbar. The toolbar performs various actions that are typically in the Menu and cell-level toolbars. 

## Toolbar 
toolbar

## Sidebar

sidebar

## TODO

- [ ] Ensure tests work
- [x] Integrate sidebar


packages/styles/app.css
```diff
@@ -18,10 +18,6 @@ body {
   margin: 0;
 }
 
-#app {
```
captainsafia, line 21: Happy that we are finally getting read of this.

packages/stateful-components/src/index.tsx
```diff
@@ -7,6 +7,7 @@ import MarkdownCell from "./cells/markdown-cell";
 import RawCell from "./cells/raw-cell";
 import CellToolbar, { CellToolbarContext } from "./cells/toolbar";
 import StatusBar, { StatusBarContext } from "./notebook/status-bar";
+import AppToolbar, { AppToolbarProps, AppToolbarContext } from "./AppToolbar";
```
captainsafia: Super nit: Can we stick with snake case file names?

captainsafia (commented):
Looking nice so far!

Now that I am seeing it in action, the downward facing chevrons at the top of the toolbar are a little too subtle to notice in the light theme. Maybe we should give them the same background color effect as the sidebar toggle icon?

packages/presentational-components/src/components/icons/icons.tsx
```diff
@@ -0,0 +1,358 @@
+import React, { HTMLAttributes, FC } from "react";
```
captainsafia: Similar to the other icons, the HTML attributes here trigger runtime warnings from React because they are not camel-cased.

packages/types/src/entities/sidebar.ts
```diff
@@ -0,0 +1,11 @@
+import * as Immutable from "immutable";
```
captainsafia, line 1: Instead of having this be a part of the entities state, can we use the myths model to implement the sidebar toggling feature.

It will allow us to keep the `core.entities` property specific to the the "core" nteract bits (messaging, communication with the kernel, output display).

I provided an explainer of how to do this here: https://github.com/nteract/nteract/issues/5164#issuecomment-646429855

I realize this piece of feedback requires considerable refactor. Take a look at the myths model and let me know if you think it is feasible to do in this PR or if it should be done as a follow-on.

packages/styles/sidebar.css
```diff
@@ -0,0 +1,95 @@
+.sidebar-container {
```
captainsafia, line 1: Can we hide overflow on this element so that the scrollbar isn't triggered when you show the sidebar?

captainsafia (approved):
Looks good overall.

The biggest feedback is around how we handle the actions/state for toggling the sidebar.

I proposed packaging it into a [myth](https://github.com/nteract/nteract/tree/master/packages/myths) to avoid polluting our core state with UI related elements.

Let me know what you think about being able to make this change.

If we end up pursuing it, we can probably implement it as part of the notebook-app-component package instead of its own myth package since it's relatively small.

packages/types/src/entities/sidebar.ts
ryanlanciaux: A myth would make a lot of sense. It seems like that value doesn't quite fit in core entities and the myth would clean that up quite a lot. 

Thank you - the comment seems pretty logical in how to implement these myths. That said, what if that came as a follow-up PR to this? I thought it was feasible but the more I've tried to reconfigure this, the more I realize I'm breaking things in the process. I'm totally happy continuing down this path in this PR but wondering if that may be more effective to break that effort down a little further.

captainsafia: @ryanlanciaux Doing it in a follow-on PR makes sense. Considering the upcoming release, I'm eager to merge things in and split up the remaining polish work.

captainsafia:
@ryanlanciaux Looks like we need to resolve the merge conflicts on this now that cell updates have been merged. I'll let you do that since you have a better sense of what should go where.

ryanlanciaux:
@captainsafia Yea - I can update this :D Wondering if it'd make sense to merge the Command Palette first (as that will likely cause more conflicts).

PR #5158, opened 2020-06-11
Command palette

ryanlanciaux:
The intent of this PR is to add a command palette to nteract. There is some ongoing work, but the initial keyboard shortcut functionality is present. 

## TODO
- [x] Connect the palette buttons with the actions (currently they just console.log)
- [x] When the palette is open, arrow keys should select the various command buttons
- [x] Ensure the correct keyboard shortcuts are used
- [ ] Ensure the tests are not broken
- [ ] Updates based on the console warnings

## Using Commands
command_palette

## Keyboard interaction
command_palette


packages/presentational-components/src/components/CommandPalette/CommandButtonRow.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { HTMLAttributes, FC } from "react";
```
captainsafia: Nit: Can we keep with the project convention and use kebab case for the filenames here?

packages/presentational-components/src/components/Icons/Icons.tsx
```diff
@@ -0,0 +1,362 @@
+import React, { HTMLAttributes, FC } from 'react';
```
captainsafia: Ditto here with regard to using kebab-case for filenames.

packages/stateful-components/src/command-palette/command-palette.tsx
```diff
@@ -0,0 +1,67 @@
...
+  addCellBelow: () => void;
+  addCellAbove: () => void;
+  restartAndRun: () => void;
+  hideOutput: () => void;
+  hideInput: () => void;
+  convertToMarkdown: () => void;
+}
+
+export interface ComponentProps {
+  id: string;
```
captainsafia: We can use the `CellId` type here. It's imported from the `@nteract/commutable` package.

packages/presentational-components/src/components/Icons/Icons.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+export const Markdown = () => {
```
captainsafia: We get a lot of runtime errors here from React complaining about HTML attributes not being camel-cased (e.g. `fill-rule` not `fillRule`). We might want to try mapping the SVGs to React components using SVGR.

packages/presentational-components/src/components/CommandPalette/CommandPalette.tsx
```diff
@@ -0,0 +1,129 @@
...
+              Hide Menu Bar
+              {shortCut.map((shortcutKey) => (
+                <KeyTag mini>{shortcutKey}</KeyTag>
+              ))}
+            </KeyTag>
+          </div>
+          <div className="command-palette-input-row">
+            <label htmlFor="commandFilter">Filter commands</label>
+            <input
+              onFocus={}
```
captainsafia: Is this set to null intentionally?

packages/notebook-app-component/src/notebook-apps/app-command-palette.tsx
```diff
@@ -0,0 +1,256 @@
...
+} from "@nteract/presentational-components";
+import { GlobalHotKeys, KeyMap } from "react-hotkeys";
+import { ContentRef } from "@nteract/core";
+import {
+  CommandContainer,
+  CommandDispatchProps,
+  CommandContext,
+} from "@nteract/stateful-components";
+
+const isMac = os.platform() === "darwin";
```
captainsafia, line 18: This won't work if the Command Palette is used in a nteract-based app in a browser context.

This is out of scope for your work but can you create an issue so that we can follow-up and implement something that will work in desktop and browser-based environments.

captainsafia (approved):
Approving this now since all the feedback is relatively minor. Looking good!

It looks like the build errors that are crashing in CI are caused by two things:

1. Not having `@types/classnames` as a dev dependency
2. Setting `onFocus={}` in the CommandPalette component is a no-go

packages/presentational-components/src/components/Icons/Icons.tsx
ryanlanciaux: Thanks for mentioning svgr - this looks great! I'll switch these over to use this.

packages/presentational-components/src/components/CommandPalette/CommandPalette.tsx
ryanlanciaux: Thank you for catching that! Sorry it was something I was meaning to change from an inline to calling a class method and didn't finish wiring it up.

PR #5170, opened 2020-06-20
Move comms state into core.entities

### LINKED ISSUE #2749: Move comms state + reducers into core.entities
Since comms has unique IDs and each entry is really associated with only one notebook document, we might as well move them up to entities.

captainsafia:
Closes #2749.


willingc (approved):
LGTM. Thanks Safia for the cleanup.

PR #5177, opened 2020-06-24
Polish up interactions with cell execute button

captainsafia:
This PR polishes up some of the interactions with the cell play button.

- It should always appear if the cell has not been executed yet
- If the cell has been executed it should only appear on hover


ryanlanciaux (approved):
This is great. I appreciate that this updates to a clearer title also.

PR #5180, opened 2020-06-26
Decrease whitespace in app area

captainsafia:
After playing around with the new UI, I noticed that the amount of whitespace can overwhelm the content in the notebook.

This PR decreases the amount of whitespace on some elements to ensure that user content has the focus of attention.


packages/styles/toolbar.css
```diff
@@ -15,8 +15,6 @@
   box-sizing: border-box;
   box-shadow: none;
   width: 100%;
-  height: 40px;
```
ryanlanciaux, line 18: I'm seeing the animation jump a little - if I add the 40px back in it seems to have the previous effect (but applies a height that may not be desired).

ryanlanciaux (approved):
This is great - it definitely makes the spacing a bit cleaner and gives more room to the content.

packages/styles/toolbar.css
captainsafia: I see. I think you addressed this in the other PR. If not, we can do it in a follow-up PR.

ryanlanciaux: Ok great - I can add a follow-up 👍👍

PR #5191, opened 2020-07-01
Add info on customizing keyboard shortcuts

CrystallineCat:
@captainsafia Here's the PR for keyboard shortcut config docs. I admit I should have doc'ed this earlier/made clearer that this was part of the patch, but I was mentally exhausted at the end.

I'd be grateful if you could put it in the Changelog. :cat:


captainsafia (approved):
Hurrah!

PR #5209, opened 2020-07-16
Fix import of webpack merge in dev script

captainsafia:
cc: @willingc @emeeks


willingc (commented):
Thanks @captainsafia 💐

PR #5210, opened 2020-07-17
Mythic multiselect

emeeks:
Adds a simple myth for multiselecting cells exposing three functions: `selectCell`, `unselectCell` and `clearSelectedCells` which you can use to add or remove cell ids from a list of cell ids.

Includes tests but is not wired up to any functionality.

### DIFF
--- a/packages/mythic-multiselect/__tests__/multiselect.spec.ts
+++ b/packages/mythic-multiselect/__tests__/multiselect.spec.ts
@@ -18,8 +18,8 @@ describe("multiselect", () => {
       })
     );
 
-    expect(state.selectedCells.size).toBeLessThan(
-      originalState.selectedCells.size
+    expect(originalState.selectedCells.size).toBeLessThan(
+      state.selectedCells.size
     );
 
     const nextState = multiselect.rootReducer(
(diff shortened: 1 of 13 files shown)


willingc (approved):
LGTM. Thanks @emeeks. I will leave this for the amazing @captainsafia to merge. Thanks again for all the help yesterday Safia.

willingc:
Not sure why the test is failing.

packages/mythic-multiselect/__tests__/multiselect.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    );
+
+    const nextState = multiselect.rootReducer(
+      state,
+      unselectCell.create({
+        id: "test2",
+      })
+    );
+
+    expect(nextState.selectedCells.size).toBeLessThan(state.selectedCells.size);
```
willingc: Will there always be at least 1 in the `size`?

emeeks: No, the test was miswritten slightly and got fixed in the later commits.

Issue #5222, opened 2020-07-25
`yarn build:desktop:watch` broken


jruales:
Commit "Fix WebPack" in PR #5115 actually fixed webpack for webpack.prod.js but broke it for webpack.dev.js (Now there's an error when running `yarn build:desktop:watch`). The "Fix WebPack" change should be reverted for webpack.dev.js.

In webpack.prod.js, the import is `const merge = require("webpack-merge");`, so it makes sense to use `merge.merge()`.

But in webpack.dev.js, the import is `const { merge } = require("webpack-merge");`, so only `merge()` should be used directly.

_Originally posted by @jruales in https://github.com/nteract/nteract/pull/5115#issuecomment-663809525_

jruales:
Also seems like a regression of #5209

CrystallineCat:
Thank you for bringing this up! I think I missed the change made by #5209 when rebasing #5115. I'll fix this momentarily.

CrystallineCat:
Oops, didn't see you already made a PR. :cat:

PR #5223, opened 2020-07-25
Fix webpack-merge

jruales:
Fix for #5222


willingc (approved):
Thanks @jruales and @CrystallineCat.

PR #5234, opened 2020-07-25
Fix webpack config I broke in rebase

### LINKED ISSUE #5222: `yarn build:desktop:watch` broken

CrystallineCat:
Fixes #5222.

### DIFF
--- a/applications/desktop/webpack.dev.js
+++ b/applications/desktop/webpack.dev.js
@@ -3,7 +3,7 @@ const { merge } = require("webpack-merge");
 
 const { commonMainConfig, commonRendererConfig } = require("./webpack.common");
 
-const rendererConfig = merge.merge(commonRendererConfig, {
+const rendererConfig = merge(commonRendererConfig, {
   devtool: false, // Turn off webpack's default source maps, as they generate
   plugins: [      // conflicting information with the plugin.
     new webpack.SourceMapDevToolPlugin({


CrystallineCat:
So, this is exactly the same change as #5223, but the build failed there, and is green here... I'm going to leave both PRs up, hoping someone will help me figure out the non-determinism with circleci with respect to installing electron... :scream_cat:

willingc:
@CrystallineCat I reran the CircleCI tests on #5223. It looks like the earlier failure was a spurious CircleCI failure. It's passing now on #5223. I don't see anything that should prevent merging #5223 or this PR. Thanks for looking into it. Leaving for you to merge your favorite :wink:

willingc (approved):

willingc (approved):

CrystallineCat:
@willingc Thank you! :smile_cat: This isn't the first time I ran into this issue with circleci; I had intermittent failures on the electron v9 branch, too. I thought it was caused by not having a prebuilt version available, but it looks like this isn't the cause. I didn't know about rerunning the tests; I'll look into that in the future. :cat:

willingc:
@CrystallineCat All the CI systems have some intermittent failures. I suspect due to latency or momentary network issues. There's a little restart icon in CircleCI (upper right). If you don't see it, we can look at bumping privileges or just ping me/Safia to restart. Thanks :D

Issue #5241, opened 2020-07-30
CRLF line endings in loaded notebooks break magics


kycutler:
**Application or Package Used**
Any nteract connected to an IPython kernel

**Describe the bug**
When a notebook is opened containing a magic followed by a CRLF line ending, IPython returns an error that the magic could not be found. See the IPython issue [ipython/ipython#12471](https://github.com/ipython/ipython/issues/12471)

Until this is fixed in IPython, nteract could mitigate this issue my not sending CRLF line endings to the kernel.

**To Reproduce**

* Open a notebook containing a magic followed by a CRLF line ending, e.g.:
  ```json
    {"cells":[{"cell_type":"code","source":["%ls\r\n"],"outputs":[],"execution_count":null,"metadata":{}}],"metadata":{},"nbformat":4,"nbformat_minor":2}
  ```
* Connect to an IPython kernel and execute the cell
* An error will be returned

**Expected behavior**
The magic should execute successfully.

**Screenshots**
![image](https://user-images.githubusercontent.com/67761731/88984015-5ceef600-d281-11ea-8cc3-d292f547ac2c.png)

**Additional context**
Note that this does not happen if the cell content is entered / edited in nteract, because CodeMirror defaults to [replacing all line endings with LF](https://codemirror.net/doc/manual.html#option_lineSeparator). However if nteract switches to use another editor like Monaco this may be an issue when editing, especially on Windows systems.

captainsafia:
Thanks for reporting this issue, @kycutler!

Given that it's an issue with the ipython kernel and not the nteract front-end, we should contribute the fix to ipython so that all front-ends can benefit from the fix. I'll close this issue in favor of the one opened in the ipython repo.

PR #5242, opened 2020-07-30
Normalize newlines before sending to kernel

kycutler:
Fix for #5241


captainsafia:
Thanks for taking the time to draft this PR, @kycutler!

Let's try to resolve this in ipython so that all front-ends can benefit from the fix.

PR #5244, opened 2020-08-03
Update linting to use eslint and plugin for typescript

willingc:
- Add config for eslint
- Remove deprecated tslint and config
- Autofix a few errors
- Relax rules from errors to warning so linting passes as before


package.json
```diff
@@ -229,14 +236,13 @@
...
     "ts-node": "^8.0.0",
-    "tslint": "^5.12.1",
     "tslint-config-prettier": "^1.17.0",
     "tslint-eslint-rules": "^5.4.0",
     "tslint-react": "^4.0.0",
     "tslint-react-a11y": "^1.0.0",
     "tslint-react-perf": "^0.3.0",
     "typedoc": "^0.17.0",
-    "typescript": "^3.7.3",
+    "typescript": "^3.9.7",
```
captainsafia: Do we need the Typescript upgrade for some reason?

willingc: Nope, I don't think so. Will double check.

willingc: Good catch. Anything over 3.3 is supported.

captainsafia (approved):
Awesome! Thank you so much, Carol!

PR #5249, opened 2020-08-05
[nteract/web] Add language detection for editor and move functions out of main file.

### LINKED ISSUE #5248: Make the below mode dynamic by identifying the language on open
https://github.com/nteract/nteract/blob/70441972824c94ba68335f77658292a3d1c88a91/applications/web/pages/p/[...params].tsx#L642-L646

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 70441972824c94ba68335f77658292a3d1c88a91 when #5240 was merged. cc @ramantehlan.

ramantehlan:
This PR is to add multiple language support to nteract web editor by identifying the language of active/opened file and update it in codeMirror mode.

This PR also moves out abstract function out of main file, to util files. 

The three files created are:
1. **helper.ts**: To store general helper function like getLanguage. 
2. **Github.ts**: To store Github related function like createCommit, upload etc. 
3. **Icons.tsx**: To move out icons from the main file, so they can be used in other pages if required. 

> This closes #5248

### DIFF
--- a/applications/web/pages/p/[...params].tsx
+++ b/applications/web/pages/p/[...params].tsx
@@ -39,7 +39,6 @@ const Binder = dynamic(() => import("../../components/Binder"), {
 const BINDER_URL = "https://mybinder.org";
 
 function getPath(params){
-  console.log(params)
     const filepathSegments = params.slice(4);
     let filepath;
     if (typeof filepathSegments !== "string") {
@@ -244,7 +243,7 @@ function addBuffer(e){
     //updateQuery( provider, org, repo, gitRef)
   }
 
- function  oauthGithub(){
+ function  OAuthGithub(){
    if ( localStorage.getItem("token") == undefined ){
         window.open('https://github.com/login/oauth/authorize?client_id=83370967af4ee7984ea7&scope=repo,read:user&state=23DF32sdGc12e', '_blank');
         window.addEventListener('storage', getGithubUserDetails)
@@ -358,7 +357,7 @@ return (
               <MenuItem >
                     { loggedIn
                           ? <Avatar userImage={userImage} username={username} userLink={userLink} / >
-                          : <Button onClick={ () => oauthGithub()} text="Connect to Github" icon={githubIcon} />
+                          : <Button onClick={ () => OAuthGithub()} text="Connect to Github" icon={githubIcon} />
                     }
                   </MenuItem>
             </Menu>
--- a/applications/web/pages/p/[...params].tsx
+++ b/applications/web/pages/p/[...params].tsx
@@ -28,7 +28,7 @@ import { Dialog, Shadow, DialogRow, DialogFooter } from '../../components/Dialog
 import { FilesListing } from "../../components/FilesListing"
 import { Layout, Header, Body, Side, Footer} from "../../components/Layout"
 import NextHead from "../../components/Header";
-import { getLanguage } from "../../util/helpers"
+import { getLanguage, getPath } from "../../util/helpers"
 import { uploadToRepo, checkFork } from "../../util/github"
 import { runIcon, saveIcon, menuIcon, githubIcon, consoleIcon, pythonIcon, serverIcon, commitIcon } from "../../util/icons"
 
@@ -37,18 +37,6 @@ const Binder = dynamic(() => import("../../components/Binder"), {
 });
 
 const BINDER_URL = "https://mybinder.org";
-
-function getPath(params){
-    const filepathSegments = params.slice(4);
-    let filepath;
-    if (typeof filepathSegments !== "string") {
-      filepath = filepathSegments.join("");
-    } else {
-      filepath = filepathSegments;
-    }
-
-    return filepath
-  }
 
 function useInput(val: string | undefined ){
   const [value, setValue] = useState(val);
--- a/applications/web/pages/p/[...params].tsx
+++ b/applications/web/pages/p/[...params].tsx
@@ -12,10 +12,6 @@ import { WithRouterProps } from "next/dist/client/with-router";
 import  { withRouter , useRouter } from "next/router";
 import { Octokit } from "@octokit/rest";
 
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faPlay, faPlus, faSave, faBars, faTerminal, faServer} from '@fortawesome/free-solid-svg-icons'
-import { faGithubAlt,  faPython } from '@fortawesome/free-brands-svg-icons'
-
 import Notebook from "@nteract/stateful-components";
 import dynamic from "next/dynamic";
 import { Host } from "@mybinder/host-cache";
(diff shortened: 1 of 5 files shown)


applications/web/util/helpers.ts
```diff
@@ -0,0 +1,60 @@
+export const getLanguage =  (extention: string) =>  {
```
ramantehlan: Adding support for all the languages at once will be a little hard, so we will keep adding support for more language as we move ahead. It's very easy to add it. we just need to add a case here with the extensions as its value and assign language based on these [supported modes](https://codemirror.net/mode/) on code mirror.

applications/web/util/icons.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
```
ramantehlan, line 1: I was a little confused about moving them to components or to util folder, but I decided to move it to util as there is no new component we are introducing in this file.

applications/web/pages/p/[...params].tsx
```diff
@@ -471,12 +223,12 @@ function addBuffer(e){
...
         setFilepath(data["path"])
-        if ( isNotebook(data["name"]) )
-            setFileType("notebook")
-        else
-            setFileType("other")
+        setFileType(fileName.split('.').pop())
     })
     }
+   
+   let extention = fileName.split('.').pop()
```
Comment (willingc) [inline applications/web/pages/p/[...params].tsx]:
```suggestion
   let extension = fileName.split('.').pop()

```

applications/web/pages/p/[...params].tsx
```diff
@@ -471,12 +223,12 @@ function addBuffer(e){
...
+   setLang( getLanguage( extention ) )
```
Comment (willingc) [inline applications/web/pages/p/[...params].tsx]:
```suggestion
   setLang( getLanguage( extension ) )

```

applications/web/util/github.ts
```diff
@@ -0,0 +1,229 @@
+import { Octokit } from "@octokit/rest";
+
+
+export const listForks = (owner, repo) => {
+   return new Promise(function(resolve, reject) {
+    // This is get the fork of the active repo
```
willingc:
```suggestion
    // This gets the fork of the active repo
```

applications/web/util/github.ts
```diff
@@ -0,0 +1,229 @@
...
+         repo,
+      }).then(({data}) => {
+         resolve({data})
+      }).catch((e) => {
+         reject(e)
+      })
+   });
+}
+
+export const createFork = (octo, owner, repo) => {
```
willingc: Do we need to check if credentials are sufficient to create a fork here?

applications/web/util/github.ts
```diff
@@ -0,0 +1,229 @@
...
+          repo,
+        }).then(() => {
+            resolve()
+        }).catch( (e) => {
+            reject(e)
+        })
+    })
+}
+
+// checkFork is to perform 3 checks
```
willingc:
```suggestion
// checkFork performs 3 checks
```

applications/web/util/github.ts
```diff
@@ -0,0 +1,229 @@
...
+// A: If user is the owner of repo.
+// B: If fork Exist.
```
willingc:
```suggestion
// B: If fork exists.
```

applications/web/util/helpers.ts
willingc:
```suggestion
export const getLanguage =  (extension: string) =>  {
```

applications/web/util/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+      let language = "markdown"
+      
+      switch( extention.toLowerCase() ) {
```
willingc:
```suggestion
      switch( extension.toLowerCase() ) {
```

willingc: I would try to order the most frequently used extensions python, javascript, julia, etc. earlier in the case statements to prevent unnecessary extra checks.

willingc (commented):
Hi @ramantehlan, This is looking good. A couple of things:

- typo on `extension`: make sure you correct all instances (I think I flagged most of them)
- do we need to recheck the user and sufficient credentials before taking add, change, or delete commands in GitHub

applications/web/util/github.ts
ramantehlan: Thank you for pointing it out. 

We are checking the `GitHub token`, which is the only **user** credential required to do the fork. We check it every time we fetch the user info(on the loading of the page), if ever the token is invalid, it is removed. If no token is present, save option won't be rendered, hence the fork can't be initiated. 

And If the owner and repo are missing, or it doesn't exist, it will throw an error.

applications/web/util/helpers.ts
ramantehlan: Good point! Fixed now. :+1:

ramantehlan:
The number of silly spelling and grammar mistakes I have done in this PR is not going to age well. :crying_cat_face:

applications/web/util/github.ts
```diff
@@ -0,0 +1,229 @@
+import { Octokit } from "@octokit/rest";
+
+
+export const listForks = (owner, repo) => {
```
captainsafia: Let's replace all these `new Promise` invocations in the file with async/await.

```
export const listForks = async (owner, repo) => {
    const octo = new Octokit();
    const  data = await octo.repos.listForks({ owner, repo });
    return data;
}
```

captainsafia: Let's also add types to the function signatures where possible.

```
listForks = (owner: string, repo: string)
```

applications/web/util/helpers.ts
```diff
@@ -0,0 +1,62 @@
+export const getLanguage =  (extension: string) =>  {
```
captainsafia, line 3: Let's use an npm package for this (I'm sure it exists!) instead of bringing this complexity into our codebase.

captainsafia (commented):
Left some initial feedback.

applications/web/util/github.ts
```diff
@@ -0,0 +1,191 @@
...
+
+export const createNewCommit = async (
+    octo: Octokit,
+    org: string,
+    repo: string,
+    message: string,
+    currentTreeSha: string,
+    currentCommitSha: string
+) =>
+    (await octo.git.createCommit({
```
captainsafia: Let's make this two explicit invocations.

```
const commit = await octo.git.createCommit();
return commit.data;
```

applications/web/util/github.ts
```diff
@@ -0,0 +1,191 @@
...
+    octo: Octokit,
+    owner: string,
+    repo: string,
+    blobs,
+    paths: string[],
+    parentTreeSha: string
+) => {
+  const tree = blobs.map(({ sha }, index) => ({
+        path: paths[index],
+        mode: `100644`,
```
captainsafia: Assign this mode to a constant so there's no magic numbers in the code.

captainsafia (approved):
Looks good. Left a few comments inline before merging.

applications/web/util/github.ts
ramantehlan: Sure, but what's the benefit of doing that?

captainsafia: Clearer to read.

Issue #5190, opened 2020-06-30
nteract desktop hangs on launch


chameroy:
I just installed nteract desktop 0.24 on Windows 10 (2004) and the application seems to hang during launch, never proceeding past the splash screen.

chameroy:
It now works. I had originally uninstalled 0.23.3 prior to installing 0.24. When I couldn't get 0.24 to start up, I uninstalled 0.24 and reinstalled 0.23.3. However, when I then ran 0.23.3, it automatically updated to 0.24. After that, 0.24 now starts.

chameroy:
I'm also encountering the startup issue on MacOS 10.15.5. However, unlike with Windows, going the update route does not get me around the issue.

ctakasaka:
> 
> 
> I'm also encountering the startup issue on MacOS 10.15.5. However, unlike with Windows, going the update route does not get me around the issue.

Although I'm on Windows, this may work for you. I just had to select the nteract icon in my menu bar and forced it to open a new notebook with some engine. Then I could use nteract normally since the window then appeared. I think you could find something similar in your menu bar up top.

chameroy:
Yes, it did! Even while the loading screen is still showing, the top bar menus are already available and I could use the file menu to create a new window.

Thanks. At least there's a workaround for both platforms for right now.

For the developers: would it be possible to not force updates in case a new release isn't working properly? In both of my cases, I tried to downgrade to 0.23.3, but was then forced to update to the new version.

chameroy:
Also, clicking on a notebook file in macOS also successfully starts up the application. I don't know about Windows. I guess the only issue might be that the application does not start up with a new window.

ichit:
I am on windows 10. Version 0.24 works  perfectly  well, and i must confess it has a cool GUI. Thanks developers. You guys are doing great work .

captainsafia:
> For the developers: would it be possible to not force updates in case a new release isn't working properly? In both of my cases, I tried to downgrade to 0.23.3, but was then forced to update to the new version.

You can disable auto-updates by setting the `NTERACT_DESKTOP_DISABLE_AUTO_UPDATE` environment variable to "1".

I'm not sure why you're experiencing inconsistent behavior with launching.

If you'd like to help us debug the issue further, you can try to see if you can reproduce the issue when running a development instance of nteract by cloning the repository and building the application from the source.

That'll help identify what hiccups occur during the initialization that prevent the app from starting.

neolee:
[...]

PR #5252, opened 2020-08-08
Fix the logic getting the default kernel

CrystallineCat:
Maybe fix for #5190?

If the default kernel for the config system isn't found, it'll currently hang at the splash screen due to wrong logic. This fixes that.


captainsafia (approved):
Doesn't hurt to try it!

PR #5257, opened 2020-08-11
[nteract/web] Add notebook component and wrap up for beta and GSoC.

### LINKED ISSUE #5225: Implement iterm.js here to connect with the termianl | This can be also done whe...
https://github.com/nteract/nteract/blob/9b3a9d50013c13d66188d2911c349f60e9230824/applications/web/components/Console.tsx#L26-L31

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 9b3a9d50013c13d66188d2911c349f60e9230824 when #5167 was merged. cc @ramantehlan.

ramantehlan:
ss

This PR is to finish adding all the tasks required to launch it as beta and wrap up the GSoC period. 

### UI/UX Integration
The list of components that needs to be created and added is given below, more might be added later:

- [x] Layout
- [x] BinderMenu
- [x] Input
- [x] Menu
- [x] Button 
- [x] Avatar
- [x] Dialog
- [x] Landing Page
- [x] Auth page
- [x] Editor `Using @nteract/editor | Later we can use @nteract/monaco-editor `
- [x] Console
- [x] FileExplorer
- [x] Notification `Using new component | Later we can use nteract legacy lib`
- [x] Notebook rendering.

### Github Integration
This PR will also allow OAuth authentication with Github, and fetch user public info. 

- [x] Complete authentication
- [x] Fetch user's public info
- [x] Fetch public files and folders 
- [x] List and create a fork for repo
- [x] Save back changes to the repo or the fork 

### MyBinder Integration
- [x] Launch MyBinder instance on start.
- [x] Communicate with JupyterServer. Fetch info like sessions etc.
- [x] Execute the command on MyBinder instance
- [x] Fetch content from Github and not jupyter server.
- [x] Save back changes done to the notebook.


packages/stateful-components/src/cells/cells.tsx
```diff
@@ -28,9 +28,10 @@ interface CellProps {
   contentRef: ContentRef;
 }
 
-export class Cells extends React.Component<StateProps & ComponentProps> {
-  render() {
-    const { cellOrder, contentRef, children } = this.props;
+type Props = StateProps & ComponentProps
+
+export const Cells = (props: Props) => {
```
captainsafia: Is there a reason for this change? For consistency with the other components in this package, we should keep this as a class component (or make all the others functional components but that's a bigger change).

packages/stateful-components/src/cells/cells.tsx
```diff
@@ -67,17 +68,16 @@ export class Cells extends React.Component<StateProps & ComponentProps> {
...
     );
-  }
 }
 
 export const makeMapStateToProps = (
   state: AppState,
   ownProps: ComponentProps
-): ((state: AppState) => StateProps) => {
-  const mapStateToProps = (state: AppState) => {
+): ((state: AppState, ownProps: ComponentProps) => StateProps) => {
```
captainsafia, line 75: Good catch!

packages/stateful-components/__tests__/cells/cells.spec.tsx
```diff
@@ -10,14 +10,15 @@ describe("makeMapStateToProps", () => {
...
   });
   it("returns the correct length of cell ids for notebook content", () => {
     const state = mockAppState({ codeCellCount: 3 });
-    const contentRef = state.core.entities.contents.byRef.keySeq().first();
-    const ownProps = { contentRef };
-    const result = makeMapStateToProps(state, ownProps)(state);
+    const contentRef = state.core.entities.contents.byRef.keySeq().first()
+    const ownProps = { contentRef: contentRef as string};
+    const result = makeMapStateToProps(state, ownProps)(state, ownProps);
+    console.log(result)
```
captainsafia: Remove this `console.log`.

applications/web/components/BinderMenu.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { FC, HTMLAttributes, useState } from "react";
 import styled from "styled-components";
 import { Button } from "./Button";
-import { Input } from "./Input";
+import { Inp } from "./Input";
```
captainsafia: Any reason for this rename? The previous naming scheme is a lot clearer.

applications/web/components/Input.tsx
```diff
@@ -58,7 +58,7 @@ export interface Props extends HTMLAttributes<HTMLInputElement | HTMLSelectEleme
   autoFocus?: boolean;
 }
 
-export const Input: FC<Props> = (props: Props) => {
+export const Inp: FC<Props> = (props: Props) => {
```
captainsafia: See above about maintaining the `Input` naming here. If there is a conflict with another component consider something more descriptive instead of an abbreviation.

applications/web/util/github.ts
```diff
@@ -1,28 +1,41 @@
...
+*/
 import { Octokit } from "@octokit/rest";
 
+export const ghGetContent = async (octo, org: string, repo: string, ref: string, fileName: string) => {
+  const data = octo.repos.getContents({ owner: org, repo: repo, ref: ref, path: fileName })
+  return data
+}
 
-export const listForks = async (owner: string, repo: string) => {
+export const ghListForks = async (owner: string, repo: string) => {
```
captainsafia: We don't have to prefix the API with `gh` here. Since all these APIs are under the github file, they can all be used under the same namespace. Such as:

```
import * as gh from "./github";

gh.listForks();
```

applications/web/components/Console.tsx
```diff
@@ -1,10 +1,12 @@
 import React, { FC, HTMLAttributes } from "react";
```
captainsafia, line 1: Run formatting over this file.

applications/web/util/helpers.ts
```diff
@@ -1,18 +1,66 @@
...
+
+  return {
+    value,
+    onChange: handleChange
+  }
+}
+
+export function createNotebookModel(filePath: string, content?: string): IContent<"notebook"> {
+  const name = filePath
+  // tslint:disable-next-line no-bitwise
```
captainsafia: I don't think you need this as there's no bitwise logic in the next line.

applications/web/components/Binder.tsx
```diff
@@ -11,72 +12,100 @@ import {
...
+    if (props.host.endpoint != "") {
+      props.setAppHost(makeJupyterHostRecord({
+        ...props.host,
+        origin: props.host.endpoint
+      }));
+    }
+
+  }, [props.host])
+
+  return (<> {
```
captainsafia: You don't need all the render fragments here.

```
if (contentFlag) {
  return <NotebookApp contentRef={contentRef} />
}
```

applications/web/pages/p/index.tsx
```diff
@@ -1,66 +1,53 @@
+// Other
```
captainsafia: Nit: Remove this.

applications/web/pages/p/index.tsx
```diff
@@ -1,66 +1,53 @@
+// Other
 import React, { FC, HTMLAttributes, useState, useEffect } from "react";
 import { WithRouterProps } from "next/dist/client/with-router";
 import { withRouter, useRouter } from "next/router";
+import { connect } from "react-redux";
 import { Octokit } from "@octokit/rest";
-
-import Notebook from "@nteract/stateful-components";
+import moment from "moment";
```
captainsafia: We try to avoid the `moment`dependency in the codebase since it is quite large. Based on the usage here, you can use the standard `Date` object here and some string formatting. Take a look at the example in [this status bar](https://github.com/nteract/nteract/blob/main/packages/notebook-app-component/src/derived-components/status-bar.tsx) which takes advantage of a smaller library.

applications/web/pages/p/index.tsx
```diff
@@ -1,66 +1,53 @@
...
-  function handleChange(e: React.FormEvent<HTMLInputElement>) {
-    setValue(e.currentTarget.checked);
-  }
-
-  return {
-    value,
-    onChange: handleChange
-  }
+export interface StateProps {
+  contents: any
```
captainsafia: This can be properly typed. You probably want the type `ContentRecord` here.

applications/web/pages/p/index.tsx
```diff
@@ -1,66 +1,53 @@
...
-  function handleChange(e: React.FormEvent<HTMLInputElement> | React.FormEvent<HTMLSelectElement>) {
-    setValue(e.currentTarget.value);
-  }
-
-  return {
-    value,
-    onChange: handleChange
-  }
+export interface ComponentProps extends HTMLAttributes<HTMLDivElement> {
+  router: any
```
captainsafia: Type this as well. You can find the type in the next/router package.

applications/web/pages/p/index.tsx
```diff
@@ -82,38 +69,95 @@ export const Main: FC<WithRouterProps> = (props: Props) => {
...
 
+  /******************
+    Effect Hooks
+   *****************/
 
-  // This Effect runs only when the username change
   useEffect(() => {
-    // Check if user has a token saved
+    // To check if Github token exist, if yes, get user details
     if (localStorage.getItem("token") != undefined) {
```
captainsafia: Nit: Can condense this if-statement into a single message?

applications/web/pages/p/index.tsx
```diff
@@ -173,54 +241,64 @@ export const Main: FC<WithRouterProps> = (props: Props) => {
...
+          setFileContent(atob(data["content"]))
+        })
+      }
+    }
+
   }
 
-  function updateVCSInfo(event, provider, org, repo, gitRef) {
-    // TODO: Add a loading experience for the users and error if project not found
+  function updateVCSInfo(event, providerP, orgP, repoP, gitRefP) {
```
captainsafia: Nit: use `previousProvider` instead of `providerP` here fo more clarity.

captainsafia (commented):
Left some comments inline.

I tried to run a cell with these commits and ran into some issues. The Binder instance was successfully connected.

I captured some error traces from the run below.

Screen Shot 2020-09-06 at 5 47 31 PM

Screen Shot 2020-09-06 at 5 47 21 PM

packages/stateful-components/src/cells/cells.tsx
ramantehlan: It's a big task, which can be achieved in different PR. so I will keep it as a class component.

ramantehlan:
Thank you for the review @captainsafia. 

While trying to reproduce the cell not running issue, I observed that It happens only when we open a notebook before it is connected to the server, so it doesn't work even after it gets connected to the server. If we reload the page, it starts to work again. I will open an issue for the same.

captainsafia (approved):
Merging now so we can iterate in smaller PRs.

PR #5258, opened 2020-08-13
Introduce mythic-windowing and move splash screen logic there

CrystallineCat:
Progress on #4799.

Plan is to move as much of the electron-related stuff out of desktop as possible, and make things based on actions/myths.

This PR introduces the general package infra for that and uses it to make the splash screen.

### DIFF
--- a/packages/mythic-windowing/package.json
+++ b/packages/mythic-windowing/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@nteract/mythic-windowing",
   "version": "0.1.0",
-  "description": "TODO",
+  "description": "A windowing system based on electron and the myths redux framework",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "nteractDesktop": "src/index.ts",
--- a/packages/mythic-windowing/package.json
+++ b/packages/mythic-windowing/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@nteract/mythic-windowing",
+  "version": "0.1.0",
+  "description": "TODO",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "nteractDesktop": "src/index.ts",
+  "scripts": {},
+  "repository": "https://github.com/nteract/nteract/tree/master/packages/mythic-windowing",
+  "keywords": [
+    "nteract"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+  },
+  "dependencies": {
+    "@nteract/myths": "^0.2.0",
+    "electron": "7.3.2"
+  },
+  "peerDependencies": {
+  },
+  "author": "nteract contributers",
+  "license": "BSD-3-Clause"
+}
--- a/packages/mythic-windowing/src/myths/window-registry.ts
+++ b/packages/mythic-windowing/src/myths/window-registry.ts
@@ -1,0 +1,14 @@
+import { windowing } from "../package";
+
+
+export const registerWindow = <T>() =>
+  windowing.createMyth("registerWindow")<{ id: string; window: T }>({
+    reduce: (state, action) =>
+      state.setIn(["windows", action.payload.id], action.payload.window),
+  });
+
+export const deregisterWindow =
+  windowing.createMyth("deregisterWindow")<string>({
+    reduce: (state, action) =>
+      state.deleteIn(["windows", action.payload]),
+  });
--- a/packages/mythic-windowing/src/myths/window-registry.ts
+++ b/packages/mythic-windowing/src/myths/window-registry.ts
@@ -1,8 +1,15 @@
 import { windowing } from "../package";
+import { Window, WindowingBackend } from "../types";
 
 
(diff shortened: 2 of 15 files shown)


packages/mythic-windowing/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "keywords": [
+    "nteract"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+  },
+  "dependencies": {
+    "@nteract/myths": "^0.2.0",
```
captainsafia, line 19: Perhaps these should be peerDeps?

packages/mythic-windowing/src/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  windows: Immutable.Map<string, Window>;
+}
+
+export interface WindowingBackend<WINDOW> {
+  showWindow: (props: WindowProps) => Observable<MythicAction>;
+  closeWindow: (id: string, window?: WINDOW) => Observable<MythicAction>;
+}
+
+export interface WindowProps {
+  id: string;
```
captainsafia: Nit: Can we define a `WindowRef` type here?

packages/mythic-windowing/src/myths/window-registry.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Window, WindowingBackend } from "../types";
+
+
+export const setWindowingBackend =
+  windowing.createMyth("setWindowingBackend")<WindowingBackend<Window>>({
+    reduce: (state, action) =>
+      state.set("backend", action.payload),
+  });
+
+export const registerWindow =
```
captainsafia: Nit: `register` and `deregister` are a little hard to graph for these APIs. Are open/close fair replacements?

CrystallineCat: What do you mean by *graph*? Open/close don't really fit, but associate/deassociate maybe?

packages/mythic-windowing/package.json
CrystallineCat: I don't know, I don't really understand peer deps. I'm going to read up on them and maybe change this...

packages/mythic-windowing/src/myths/window-registry.ts
captainsafia: *awakens from the grave*

Terrible typo here. I meant "hard to grok".

I see what you mean about close/open not fitting. In that case, I'm fine with deregister/register.

packages/mythic-windowing/package.json
CrystallineCat: Hmm... I'm still not sure I fully understand peer deps, but I don't think this is a case for them. `myths` doesn't use shared state and we do actually depend on its API.

rgbkrk: This does seem like a case for keeping this as a regular dependency. 👍

rgbkrk:
This looks pretty good to me! I've been out of the loop for a bit so it looks like it's time for me to learn about the myths here. 😄

captainsafia (approved):
Looks good! Thanks.

Sidenote: for some of these myths that are desktop-app specific, I wonder if we could relocate them to inside the "applications/desktop" directory and avoid having them listed as packages. This way, they don't end up being picked up by lerna when we publish.

CrystallineCat:
Thanks all! :cat:

@captainsafia What is the downside of them being picked up by `lerna`? Doesn't that need to happen to publish `desktop` anyway? Or am I misunderstanding something?

captainsafia:
> @captainsafia What is the downside of them being picked up by lerna? Doesn't that need to happen to publish desktop anyway? Or am I misunderstanding something?

If they are picked up by lerna, they'll be published as individual npm packages. For some myths, like the configuration myth, this is a good idea. For myths that aren't as generalizable to both the desktop and web apps, we probably want to avoid shipping additional npm package.

See also #4981. Having a lot of packages defined increases the time needed to ship the repo.

CrystallineCat:
Ah, thank you! However, I don't think `mythic-windowing` is a desktop-only package; for the moment my goal is to encapsulate `electron`, but it could conceivably work with different backends. Some of its functionality, e.g. the `openExternalUrl` myth I'm about to PR, could be of use to `web` as well, and one of the reasons I'm encapsulating the windowing is to enable different desktop utilities. I actually tried to start a tool that would just display the `jupyter` config, found kernels, etc., a while ago to help debug kernel issues, but couldn't figure out how the whole `desktop` windowing stuff worked. `mythic-windowing` is supposed to make that kind of stuff easier.

I still have to figure out how to make the dependency on electron optional, though... I tried a general approach with mythic packages all having backends, but I couldn't figure out how to properly do type inference.

captainsafia:
Ah I see. In that case, we'll have to figure out how to reduce the ship cost for these packages.

Supposedly, it's possible to set up automated releases for lerna-based packages with [lerna-semantic-release](https://github.com/atlassian/lerna-semantic-release) so that might help with the burden of having a lot of smaller packages in the monorepo.

Issue #5055, opened 2020-04-17
Include golang syntax highlighting support in Codemirror


vsivsi:
**Is your feature request related to a problem? Please describe.**
Using Nteract with the [Gophernotes](https://github.com/gopherdata/gophernotes) kernel generally works great! Except there's no syntax highlighting! It works fine in JupyterLab and classic Jupyter notebooks.

First reported here: https://github.com/gopherdata/gophernotes/issues/158

**Describe the solution you'd like**
Enable support for golang in CodeMirror.  Possibly here(?): 

https://github.com/nteract/nteract/blob/master/packages/editor/src/index.tsx#L211

**Describe alternatives you've considered**
I guess I could just use JupyterLab...

**Additional context**

Jupyterlab:

<img width="508" alt="image" src="https://user-images.githubusercontent.com/428642/79600312-d03a7680-809b-11ea-9024-a6c95ee2e146.png">

Nteract:

<img width="508" alt="image" src="https://user-images.githubusercontent.com/428642/79600362-e47e7380-809b-11ea-8450-f0c54bc1fb69.png">

captainsafia:
The good news is that syntax coloring is relatively easy to add in nteract. It's controlled by CodeMirror and we can add support for Go by adding the Go codemirror mode [here](https://github.com/nteract/nteract/blob/master/packages/editor/src/index.tsx#L222).

```
require("codemirror/mode/go/go");
```

Would you be interested in submitting a PR with this change?

Interestingly enough, I'm surprised that Go syntax highlighting works in JupyterLab because it doesn't look like that CodeMirror mode is included [there](https://github.com/jupyterlab/jupyterlab/blob/f0153e0258b32674c9aec106383ddf7b618cebab/packages/codemirror/src/mode.ts) so perhaps there is a way to automatically download codemirror modes when a notebook from a particular type is loaded.

daxian-dbw:
@captainsafia I was about to add entries for syntax coloring for c#, but found that `codemirror/mode/clike/clike` is already there, which should cover c# (https://codemirror.net/mode/clike/). Why is c# still not properly syntax colored on macOS?

<img src="https://user-images.githubusercontent.com/127450/80034274-ff3d5780-84a2-11ea-8931-96b8ab81e87b.png" width="300" height="200">

Gozala:
[...]

PR #5269, opened 2020-08-24
Add codemirror mode for nim

stisa:

This adds a codemirror mode for syntax highlighting of [nim](https://nim-lang.org)
If there's some other plugin mechanism in place to load modes for languages not supported by codemirror please let me now and I'll try to use that.

Screenshot:
Annotazione 2020-08-24 171711


willingc:
Hi @stisa, 

We would be happy to add support for nim but taking a different approach than this PR. This PR adds Go support and outlines our [preferred approach](https://github.com/nteract/nteract/issues/5055#issuecomment-615554313).

We recommend that you create an external package likely using much of the code in this PR.

Thanks! Let us know if you have questions. cc/ @captainsafia

stisa:
Hi! Thanks for the reply.  
Makes sense, I'll have a look at doing that.  

Basically I need to:
- create a npm package for the mode
- add the package to the editor's dependecies ([here?](https://github.com/nteract/nteract/blob/main/packages/editor/package.json#L19))
- require that in `index.tsx`

Does that sound correct?

willingc:
@captainsafia Please correct me if I'm wrong. 

@stisa I don't think you need to add the package to the dependencies. Much like Rust or Go, you would require it in `packages/editor/src/index.tsx`

stisa:
> @stisa I don't think you need to add the package to the dependencies. Much like Rust or Go, you would require it in `packages/editor/src/index.tsx`

@willingc I'm pretty sure it's needed, to my understanding rust and go worked like that because they're already present in codemirror's package ([here](https://github.com/codemirror/CodeMirror/tree/master/mode)), they just weren't `required` by nteract.  As my mode is external, a dependency is needed for npm to pull it from the registry. 

(sorry if I'm wrong, I don't know much about javascript development)

willingc:
Thanks @stisa. Your points make sense to me. @captainsafia thoughts?

captainsafia:
Unfortunately, we will need to include it in the package.json since it is no bundled with the default set of syntax highlighters that CodeMirror provides.

I say "unfortunately" because this is one of those things that I'd like to avoid having as a dependency of the package and instead allow people to configure in the desktop app.

I don't have any ideas on how to make that possible now but I am open to hearing ideas from others!

stisa:
Yeah I understand adding dependencies just to support a language isn't ideal.

One way to do it could be like with jupyter notebooks, where a special file `kernel.js` is loaded from the kernelspec folder, eg. in `loadMode` if no mode is found try looking for a `<language>.mode.js` in the kernelspec folder (or similar) that provides a `register` function, but I'm not sure how doable it would be with nteract's architecture.

PR #5301, opened 2020-09-23
Bugfix for Monaco editor parameter widget not closing

pwang347:
- [x] I have run through the TEST_PLAN.md to ensure that my change does not break anything else. (validated in Windows, though for GeoJSON I've only confirmed that the behaviour is the same as main but either way the leaflet loads but renders nothing)

## Context
We are currently getting a bug related to the Monaco parameter widget not closing when focus shifts to a different cell (as well as the case where a different function is hovered). These issues (https://github.com/microsoft/vscode-python/issues/7851 and https://github.com/microsoft/vscode-python/issues/8007 respectively) were resolved in the [vscode-python](https://github.com/microsoft/vscode-python) repo, and this PR largely overhauls the same fix with minor differences. For specific resolution commits, see https://github.com/microsoft/vscode-python/commit/fa75cf0c70b70c08bd3ebf454b99c86e7a610aef and https://github.com/microsoft/vscode-python/commit/231f27428b4bd36f660e26c3bf53b5ec41a22ded.

## Tests
As it's a bit difficult to simulate the Intellisense popup and mouse interactions to reliably test, I've omitted them for now. Please let me know if there are other test scenarios that may be relevant here!

For manual testing, the main scenario was:
* Parameter widget popup now correctly hides when mouse hovers over to another cell (and conversely, that when mouse doesn't leave the cell the popup should remain visible)

### General case
monaco_popup_old

### Multiple popups
monaco_popup_multiple_old

### General case
monaco_popup_fix

**NOTE**: it is still possible to have more than one popup for a single cell, though at this point it is less obvious that this is a bug but rather preference. This behaviour is the same as in vscode.

### Multiple popups
monaco_popup_multiple_fix

Scrolling also has the behaviour of sliding the popup along with it, but it seems that the same behaviour is present in vscode.

#### Still focusable
monaco_popup_focus_fix

The popup only closes when your mouse moves outside of the region defined by the current active editor + the popup itself.

#### Screenreader
monaco_popup_screenreader_fix


changelogs/current_changelog.md
```diff
@@ -180,7 +180,7 @@ Provide a bulleted list of new features or improvements and a reference to the P
 
 #### Bug Fixes
 
-Provide a bulleted list of bug fixes and a reference to the PR(s) containing the changes.
+- Hide Monaco parameter widget when mouse moves into another editor cell.
```
joruales, line 183: Missing the link to this PR

ghost: Missing a link to this PR

pwang347: Fixed!

pwang347:
Unrelated, but seems like scroll is also broken with long parameter text for this widget, while the function signature/hover widget is fine.
image
image

captainsafia:
@vivek1729 is more familiar with this code base so I've requested his review as well.

Were you able to reproduce the original issues in the @nteract/monaco-editor?

pwang347:
@captainsafia Thanks! I actually spoke to Vivek just earlier and he says he will take a closer look at the code.

Also yep, it seems that this is the default behaviour of monaco and it can be repro'd in the [playground](https://microsoft.github.io/monaco-editor/playground.html#interacting-with-the-editor-adding-a-command-to-an-editor-instance) as well. The closest issue I found was [this](https://github.com/microsoft/monaco-editor/issues/1929), but it seemed largely unaddressed.

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -5,9 +5,17 @@ import * as React from "react";
 import { completionProvider } from "./completions/completionItemProvider";
 import { ContentRef } from "@nteract/core";
 import { DocumentUri } from "./documentUri";
+import debounce from "lodash/debounce";
```
vivek1729: We'll have to add the `lodash` dependency to Monaco editor's `package.json`

vivek1729: You'd probably want to import `debounce` from `lodash.debounce` module now that we are importing just that API. https://www.npmjs.com/package/lodash.debounce

packages/monaco-editor/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "@nteract/core": "^14.0.0",
     "@nteract/messaging": "^7.0.10",
+    "lodash.debounce": "^4.0.6",
```
vivek1729, line 28: Great to see this import pattern where we depend upon just the API we need.

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: good catch, updated!

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -5,9 +5,17 @@ import * as React from "react";
 import { completionProvider } from "./completions/completionItemProvider";
 import { ContentRef } from "@nteract/core";
 import { DocumentUri } from "./documentUri";
+import debounce from "lodash.debounce";
```
ghost, line 8: consider using debounce from RXJS to avoid extra dependencies

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -83,13 +91,22 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+  private mouseMoveListener?: monaco.IDisposable;
+
   constructor(props: IMonacoProps) {
     super(props);
     this.calculateHeight = this.calculateHeight.bind(this);
     this.onBlur = this.onBlur.bind(this);
     this.onDidChangeModelContent = this.onDidChangeModelContent.bind(this);
     this.onFocus = this.onFocus.bind(this);
     this.resize = this.resize.bind(this);
+    this.hideAllOtherParameterWidgets = debounce(this.hideAllOtherParameterWidgets.bind(this), 150);
```
kjcho-msft: Is 1500ms latency good? isn't it too long?

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -242,6 +259,23 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+        this.hideParameterWidget();
+      });
+
+      this.focusEditorWidgetListener = this.editor.onDidFocusEditorWidget(() => {
+        this.hideAllOtherParameterWidgets();
+      });
+
+      if (this.editor) {
+        this.mouseMoveListener = this.editor.onMouseMove((e: any) => {
+          if (!this.coordsInsideEditor(e.event?.pos?.x, e.event?.pos?.y)) {
```
kjcho-msft: Doesn't onMouseMove just work? Can you add detail on what coordsInsideEditor does?

packages/monaco-editor/src/MonacoEditor.tsx
captainsafia: We need to use `lodash.debounce` method here since the RxJS one returns an Observable-typed object.

pwang347:
Might be a good idea for me to concretize the different potential behaviours (feel free to add more, this is by no means exhaustive), and what should be done so that we can have a better discussion:

Firstly, here are the scenarios that I think we can probably agree on at the moment:
* The parameter help popup should close when you click anywhere except where the cursor is blinking and the popup itself
* You should be able to focus and click on the text to copy it (clicking on the popup will never close it)
  * Not fully reflected in PR yet. Right now, if your mouse leaves the current cell while on the popup it will close.
  * I had a discussion with Joaquin and the conclusion was that it would make a lot of sense to keep if someone needs to copy text / scroll a long comment, and even current signature popup (hover) allows you to focus and copy from it
* Screen readers should be able to access the popup without it closing

Some that I'm a bit unsure of at the moment:
* Is it a problem to display another popup overlaid on top of the parameter help popup? E.g. the signature help popup when you hover another function name. What vscode has now is kinda weird, their code is set up so that hovering on a different cell will hide the active parameter help popup, but not when it's in the current cell
  * There is no way to ever exceed two popups at once with this issue still (I'm not aware of any shortcuts that generate popups at least) so maybe this could be considered as marginally useful for preserving context
* So... should the popup 1) hide only when you click something that's not the popup? 2) hide the moment your mouse leaves it (make it hide like the other hover popup)? 3) hide when the mouse leaves both the editor area and the popup? 4) another variant of 1?
  * (1) hide on click
    * Pros: when you're copying text and scrolling it's much harder to accidentally close the popup
    * Cons: leads to the overlaying "issue" mentioned above
  * (2) hide on mouse leaving the popup
    * Pros: does not have overlaying "issue" mentioned above
    * Cons: clearly, it's unlikely that your mouse is always directly on top of the popup (it could be anywhere) meaning that any slight movement of the mouse will usually end up dismissing the popup; user will need to be trained to know that if they want the popup to remain they need to keep their mouse on wherever the popup spawns
  * (3) hide on mouse leaving the popup + editor area **<- VS Code is doing this**
    * Pros: fixes overlaying "issue" when you try to hover into a function in a different cell
    * Cons: doesn't fix the "issue" when you try to hover into a function in the same cell, and also no guarantees the mouse will be in the editor area. But generally speaking is a lot more likely that the mouse being where the popup appears.
  * (4) hide on click and also when other popups appear
    * Pros: resolves overlaying "issue", also has benefits of harder to accidentally close the popup when your mouse moves too far accidentally when scrolling or copying.
    * Cons: ?

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: I think it's to just limit the hiding (of popups in other editors) to only when the mouse isn't over the currently focused editor. In this case I think it may be cheaper to call coordsInsideEditor each move event than the hide parameter which does a heavier task of needing to do several querySelections.

I think this is a good point though, if it's okay to do this when you're not hovering on top of the current editor then why do it elsewhere?

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: I think this is mainly for performance concerns, since each move event will trigger the query selections; in terms of behaviour it seems to be reactive enough

Edit: reduced to 50ms for more responsiveness

packages/monaco-editor/src/MonacoEditor.tsx
ghost: I was thinking of something like the following
```
var eventStream = new rxjs.Subject();

var subscription = eventStream.pipe(rxjs.operators.debounceTime(200ms)).subscribe(x => console.log('Next: ' + x))

eventStream.next("foo");
eventStream.next("bar");
eventStream.next("baz");

// only "baz" is printed to console, after 200ms
```

Example fiddle: [link](https://jsfiddle.net/y0bnha1w/1/)

Although admittedly, using `lodash.debounce` will be less cryptic.

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: Nevermind you're right, it shouldn't be necessary as it is right now (just a minor optimization), though we will need to check that we aren't hovering over a popup still

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: I think I would personally prefer `lodash.debounce` here given that there's no idiomatic RXJS way to work with the  `editor.on` event listeners and the import itself doesn't seem too large (1 JS file ~300 lines). What do you think?

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -242,6 +263,21 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       if(this.props.cursorPositionHandler){
         this.props.cursorPositionHandler(this.editor, this.props);
       }
+
+      // When editor loses focus, hide parameter widgets (if any currently displayed).
+      this.blurEditorWidgetListener = this.editor.onDidBlurEditorWidget(() => {
```
pwang347, line 267: TODO: with the recent changes, it actually seems like these other two events are unneeded, will do some tests around removing these

Edit: this one seems to be required to close when tabbing, but the other one is okay to remove (the logic for hideParameterWidget can probably be simplified a lot still)

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -242,6 +263,21 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+        this.hideParameterWidget();
+      });
+
+      this.focusEditorWidgetListener = this.editor.onDidFocusEditorWidget(() => {
```
pwang347: ditto

pwang347:
For consistency with vscode, the current approach will be (3). With the latest changes the popup should remain open as long as the mouse is either above the currently focused editor or the popup itself!

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -441,4 +477,157 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       }
     }
   }
+
+  /**
+   * This will hide the parameter widget if the user is not hovering over
+   * the parameter widget for this monaco editor.
+   *
+   * Notes: See issue https://github.com/microsoft/vscode-python/issues/7851 for further info.
+   * Hide the parameter widget if all of the following conditions have been met:
```
ghost: Is it really if ALL of them are met? It seems like it's actually if ( (2nd one) and (3rd one) ) or if (1st one)?

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -441,4 +477,157 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+
+    // Find all elements that the user is hovering over.
+    // It's possible the parameter widget is one of them.
+    const hoverElements: Element[] = Array.prototype.slice.call(document.querySelectorAll(':hover'));
+
+    // These are the classes that will appear on a parameter widget when they are visible.
+    const parameterWidgetClasses = ['editor-widget', 'parameter-hints-widget', 'visible'];
+
+    // Find the parameter widget the user is currently hovering over.
+    let parameterWidget = hoverElements.find(item => {
```
ghost: Maybe rename to something like "isParameterWidgetHovered" to show it's a boolean and to show its true meaning

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -441,4 +477,157 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+  /**
+   * Hides the parameters widgets related to other monaco editors.
+   * Use this to ensure we only display parameters widgets for current editor (by hiding others).
+   *
+   * @private
+   * @returns
+   * @memberof MonacoEditor
+   */
+  private hideAllOtherParameterWidgets() {
+    const rootSelector = document.querySelectorAll('.nteract-cells');
```
ghost: Is it reasonable to assume that .nteract-cells will exist? I guess that technically nteract users can just mix and match stuff from nteract, so please make sure that this element .nteract-cells can be expected to exist

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: Resolved offline, will keep lodash

packages/monaco-editor/src/MonacoEditor.tsx
pwang347: I guess it didn't really make sense to use an intermediary anyway, so I've removed it!

ghost (approved):

vivek1729 (approved):

kjcho-msft (approved):

captainsafia:
Merging ahead of today's release.

PR #5307, opened 2020-09-29
Move styles from css to styled components to resolve build issue

vivek1729:

The package added a css file for perf optimziations. It turns out that the css files were not getting packaged into the `lib` folder as the typescript compiler did not pick it up as part of the `build` process. Also took a look at the [`styles`](https://github.com/nteract/nteract/tree/main/packages/styles) package which has a bunch of css files but it seems like it is wired up differently and does not really output a `lib` folder as the compilation target. We therefore followed the pattern for the `codemirror` editor and moved the styles to a styled component.

In the future we can explore build configuration changes to properly copy static assets (if any) to the compiler output folder.


packages/monaco-editor/src/monacoStyle.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/*
+For the following components, we use the inherited width values from monaco-container.
+On resizing the browser, the width of monaco-container will be calculated
+and we just use the calculated width for the following components
+So we don't need to use Monaco editor's layout() function which is expensive operation and causes performance issues on resizing.
+*/
+
+export default createGlobalStyle`
+.monaco-container .monaco-editor {
```
pwang347: nit: tabbing

pwang347 (approved):

packages/monaco-editor/src/monacoStyle.ts
vivek1729: Done.

captainsafia:
Any other validation we need to do on this?

pwang347:
@captainsafia I think to validate this we just need to verify that 1) there are no errors when running a project using the package and 2) that the styles are applied

Both Vivek and I tested the former earlier and things seemed to be okay.

The second one has me a bit confused though, I noticed that styles in general from the file weren't taking effect when I ran the nteract electron app. In fact, in the [documentation for styled components](https://styled-components.com/docs/api#createglobalstyle) it seems like the typical usage for `createGlobalStyle` involves adding it to the render tree.

ie.

```
<div>
  <GlobalStyle/>
  ...
</div>
```

Does this mean we should 1) rename `monacoEditorCSS` to `MonacoEditorCSS` (uppercase) and 2) add it as a child of the parent div in `render`?

The odd part though is that the `CodeMirrorCSS` and `ShowHintCSS` styles in CodeMirror aren't doing this at all, and also don't show up unless I add them to the render tree. Are these not supposed to show up by default and only when the export is used? If so, @vivek1729 do we want to make the current style change a default since it's a performance fix?

vivek1729:
That's a good point @pwang347. I followed the pattern defined in CodeMirror and I didn't see the `CodeMirrorCSS` getting applied explicitly anywhere. Does it mean they automatically get applied since it is a global style? What do you suggest @captainsafia?

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -6,8 +6,9 @@ import { completionProvider } from "./completions/completionItemProvider";
 import { ContentRef } from "@nteract/core";
 import { DocumentUri } from "./documentUri";
 import debounce from "lodash.debounce";
-import "./style.css";
+import monacoEditorCSS from "./monacoStyle";
```
pwang347: I think if this is being used anywhere it will still need to be capitalized to avoid [this](https://stackoverflow.com/questions/37414304/typescript-complains-property-does-not-exist-on-type-jsx-intrinsicelements-whe) issue

pwang347:
I just dug in a bit more out of boredom and noticed this PR:
https://github.com/nteract/nteract/pull/4093

Seems like it might be problematic to leave it here if it re-runs? Though the purpose isn't to drastically change style currently maybe it's dangerous to leave the css file around as per the PR comments (in case people decide to add more to the styling that may cause perf issues on re-render).

As for how Code Mirror CSS is working, it seems like it's... not? It looks like this change removed it and introduced a dependency on a style from the codemirror repo:
https://github.com/nteract/nteract/pull/4181/files

Some more cleanup happened here, though it seems largely unrelated:
https://github.com/nteract/nteract/commit/95c2abc0509657eee22ca36fa3740ac8a3f9a577#diff-09dab28ca09ac4ac6c052be5a7535a31

^Edit: missed that the style components were also re-added in this commit

captainsafia:
> As for how Code Mirror CSS is working, it seems like it's... not? It looks like this change removed it and introduced a dependency on a style from the codemirror repo:
https://github.com/nteract/nteract/pull/4181/files

Yeah, I made this change when I was doing stateful-components work. styled-component's re-rendering logic has a significant perf cost so I just nixed it out in favor of all CSS everything. More info on this at https://github.com/nteract/nteract/issues/4885#issuecomment-581191883.

@vivek1729 Might be best to add the stylesheet under @nteract/styles/monaco and document importing it at the top-level?

Does having it in a style-component dampen the perf improvements at all?

vivek1729:
@captainsafia, that was a good catch. The styled component brings a perf hit. I was not able to crunch the numbers but it was noticeable. 

> Might be best to add the stylesheet under @nteract/styles/monaco and document importing it at the top-level?

Adding it to the `styles` package makes sense. Where should we add this `css` import then? Or do you suggest we just document it and not explicitly import the css anywhere?

pwang347:
Hi all, just a gentle ping!

So it seems that the dilemma right now is between:

1) Using a CSS file: better performance, but needs some sort of copy script in `package.json` most likely either as a separate build step, or using a hook like `prepublish` / `prepublishOnly`
2) Using a styled component: worse performance, needs to be manually applied by package consumers at top level

What might be some tradeoffs we're willing to take here? Are there any other possible options aside from these?

captainsafia:
> Or do you suggest we just document it and not explicitly import the css anywhere?

Yes, I was suggesting this is as a first step.

> Using a CSS file: better performance, but needs some sort of copy script in package.json most likely either as a separate build step, or using a hook like prepublish / prepublishOnly

Correct. This is assuming we don't want to go with approach of shipping and documenting that folks will need to import it in their top-level application.

We'll also want to make sure the CSS import works regardless of what webpack configuration the parent app has (which is tricky).

So my proposal is:

1. Ship + document to start.
2. If the above seems to cumbersome or confusing for folks, investigate updating our build scripts to support the CSS file.

> Using a styled component: worse performance, needs to be manually applied by package consumers at top level

A styled-component doesn't need to be manually applied at the top-level if we include it in the monaco editor itself. The drawback with it is the perf impact which ends up defeating the purpose of this change in the first place.

vivek1729:
Thanks for clarifying. Your suggestion totally makes sense. I have made the corresponding changes and imported the css at the top level for the nteract apps. I have also verified the changes by building locally and doing a release build of the desktop app by doing `yarn dist`.
cc: @captainsafia, @pwang347

pwang347:
LGTM if css is being applied properly with the new import setup!

pwang347 (approved):

captainsafia (approved):
:shipit:

PR #5331, opened 2020-10-21
Fix: Set kernel execution state after launch

nicholaswold:

This fix is meant to set the execution state of a kernel after receiving a `kernel_info_reply` message. See: https://github.com/nteract/nteract/discussions/5324

**The problem:** In cases where kernels take a long time to boot, calls to the Jupyter Server's sessions API would return an `execution_state` of `starting`. Once in this state, we never go back to update the status of the kernel, even after establishing a websocket with the server and receiving a `kernel_info_reply`. For UI actions that switch on particular kernel execution states, the only resolution to this was to refresh the page.

**The fix:** The code here simply updates the execution state of the kernel to `launched`, much like what we do in `zeromq-kernels.ts`. In the event of a long boot, this will ensure that the state does eventually get set to something.

### DIFF
--- a/packages/epics/src/kernel-lifecycle.ts
+++ b/packages/epics/src/kernel-lifecycle.ts
@@ -67,13 +67,13 @@ export const watchExecutionStateEpic = (
           catchError((error: Error) => {
             return of(
               actions.executeFailed({
-                  error: new Error(
+                error: new Error(
                   "The WebSocket connection has unexpectedly disconnected."
-                  ),
-                  code: errors.EXEC_WEBSOCKET_ERROR,
-                  contentRef: (action as actions.NewKernelAction).payload.contentRef  
+                ),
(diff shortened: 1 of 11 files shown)


packages/epics/src/kernel-lifecycle.ts
```diff
@@ -139,7 +139,8 @@ export function acquireKernelInfo(
           actions.setKernelInfo({
             kernelRef,
             info
-          })
+          }),
+          actions.setExecutionState({ kernelStatus: "launched", kernelRef })
```
captainsafia: Consider using the `KernelStatus` enum here and elsewhere.

https://github.com/nteract/nteract/blob/49983455451c1038a59408537e8698ebca9f074a/packages/types/src/entities/kernels.ts#L40

captainsafia (approved):
🙏  Thanks!

PR #5338, opened 2020-10-27
Fix print to PDF and move formatting to stylesheet

captainsafia:
- Fix print to PDF feature
- Move styles to CSS

TBD:
- Fix bug where printToPDF only prints what is in viewport

### DIFF
--- a/applications/desktop/src/common/commands/by-menu/file.tsx
+++ b/applications/desktop/src/common/commands/by-menu/file.tsx
@@ -168,40 +168,25 @@ export const ExportPDF: DesktopCommand<ReqContent> = {
     const basename = path.basename(notebookName, ".ipynb");
     const basepath = path.join(path.dirname(notebookName), basename);
     const pdfPath = `${basepath}.pdf`;
-    const model = selectors.notebookModel(state, props);
-
-    // TODO: we should not be modifying the document to print PDFs
-    //       and we especially shouldn't be relying on all these actions to
-    //       run through before we print...
-    const unexpandedCells = selectors.notebook.hiddenCellIds(model);
-    yield* unexpandedCells.map(
-      id => actions.toggleOutputExpansion({ id, ...props }),
-    );
-
-    let data: any;
 
     try {
-      data = await promisify(remote.getCurrentWindow().webContents.printToPDF)(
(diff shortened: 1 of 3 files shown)


captainsafia:
Taking this out of draft mode to review these improvements. We still need to figure out why it is only printing one page.

jarmokivekas:
I didn't yet have time to test this, but I'm happy to see activity on this issue. I kinda miss this feature, even keep a v0.23 installed just to have access to it :grin:

willingc (approved):
I tried this locally and it does capture part of the notebook as PDF and exports. I noticed that if I scrolled to the bottom of the notebook that it would export that section of the notebook.

Merging as is. We can work on the multipage PDF in a separate PR. Thanks @captainsafia :sunny:

applications/desktop/src/common/commands/by-menu/file.tsx
```diff
@@ -168,40 +168,25 @@ export const ExportPDF: DesktopCommand<ReqContent> = {
...
-      data = await promisify(remote.getCurrentWindow().webContents.printToPDF)(
-        { printBackground: true },
-      );
-    }
-    finally {
-      // Restore the modified cells to their unexpanded state.
-      yield* unexpandedCells.map(
-        id => actions.toggleOutputExpansion({ id, ...props }),
-      );
+      const data = await remote.getCurrentWindow().webContents.printToPDF({ printBackground: false });
```
willingc, line 173: If I had to guess I think this is the line that we'll need to update.

captainsafia:
> I tried this locally and it does capture part of the notebook as PDF and exports. I noticed that if I scrolled to the bottom of the notebook that it would export that section of the notebook.

Yep! The issue is that it only prints what is currently on the page. I was able to workaround this by adding a `* { overflow: visible !important; }` style on the page for `@media print`. This prints both pages but also prints an additional set of 600 blank pages.

So it appears that there might be a sweet spot we need to land with the CSS here. It might be that we need to disable hiding the overflow on the cell components or something.

willingc:
Gee thanks @captainsafia. I just about spit my drink out when you mentioned 600 blank pages! Thanks for the humor.

PR #5353, opened 2020-11-15
Extract the kernelspecs handling out of the start up logic

### LINKED ISSUE #5352: Light-colored icon for system tray
`nteract` system tray icon is black and is not visible with dark styled desktop themes.

A light-colored icon for system tray.

None.

**Additional context**
None.

CrystallineCat:
Progress on #4799. Might fix #5352.

* I extracted all the kernelspec start up logic from `index.ts` into `kernel-specs.ts` to simplify the former in preparation of moving stuff to `mythic-windowing`.
* Also took a shot at fixing #5352, but cannot verify completely (see comment there).


applications/desktop/src/main/index.ts
```diff
@@ -85,49 +86,30 @@ const jupyterConfigDir = join(app.getPath("home"), ".jupyter");
...
-  mergeMap(() =>
-    // Create all the directories we need in parallel
-    forkJoin(
-      // Ensure the runtime Dir is setup for kernels
-      mkdirpObservable(jupyterPaths.runtimeDir()),
-      // The config directory is taken care of by the configuration myths
-    )
-  ),
-);
-
```
CrystallineCat, line 99: Removed the above, because `jupyterPath.runtimeDir` is referenced nowhere else in the repo. Also, shouldn't the kernels be managing/creating that one?

rgbkrk (approved):

willingc (commented):
I pulled this locally and am seeing an increase in kernel errors. I'm not sure if it is related to this PR or not, but I don't see the same with current release.

CrystallineCat:
@willingc Which kernels and what kind of errors? I went through a number of the example notebooks on `python3` and `node_nteract` and wasn't able to reproduce.

willingc:
@CrystallineCat Both Python and Julia. I'm using notebooks that are a bit more complex than the examples. I used the same on HEAD and 0.26 and wasn't seeing the same thing. That said, kernels are finicky things so who knows. I wouldn't hold up merging.

applications/desktop/src/main/index.ts
willingc: We probably want to set the `runtime_dir` to `data_dir/runtime` [ref](https://jupyter.readthedocs.io/en/latest/use/jupyter-directories.html#runtime-files) or be able to call `jupyter_core.paths.jupyter_runtime_dir` if needed.[ref](https://jupyter-core.readthedocs.io/en/latest/paths.html#jupyter_core.paths.jupyter_runtime_dir). For new installations this shouldn't matter but if folks have things other than the defaults, it may.

If I have preinstalled kernels, it may be an issue. Probably a few tests to cover the kernel basics would be helpful.

willingc (approved):
Thanks @CrystallineCat for this PR. I think it makes good sense to split things up more logically. My only concern is that we test the basic cases: new kernel, existing kernel.

CrystallineCat:
Thank you! :cat:

PR #5361, opened 2020-11-16
Use KernelStatus enum more in codebase

### LINKED ISSUE #5349: KernelStatus not matching with state being added to redux store
nteract/packages/reducers

**Describe the bug**
The status of the kernel being set in the redux store does not seem to match the `KernelStatus` types. 

**To Reproduce**

State being set in store: 
`LAUNCH_KERNEL_BY_NAME` status is set to `"launching"`
https://github.com/nteract/nteract/blob/5d25c23fd23e2e20aa6abb43b84c37f8a60ebeea/packages/reducers/src/core/entities/kernels.ts#L42-L50

`KernelStatus` for launching == `"starting"` 
https://github.com/nteract/nteract/blob/49983455451c1038a59408537e8698ebca9f074a/packages/types/src/entities/kernels.ts#L40-L46

**Expected behavior**
KernelStatus should reflect the state being set in the redux store to enable comparisons. Not sure if this is intended and I'm just missing some intermediate step :)

captainsafia:
Closes https://github.com/nteract/nteract/issues/5349.

We weren't using the `KernelStatus` enum everywhere throughout the codebase. Comments inline for some of these changes.

### DIFF
--- a/applications/desktop/src/notebook/epics/zeromq-kernels.ts
+++ b/applications/desktop/src/notebook/epics/zeromq-kernels.ts
@@ -469,7 +469,7 @@ export function watchSpawn(
             // We both set the state and make it easy for us to log the error
             observer.next(
               actions.setExecutionState({
-                kernelStatus: "process errored",
+                kernelStatus: KernelStatus.Error,
                 kernelRef: action.payload.kernelRef
               })
             );
@@ -479,7 +479,7 @@ export function watchSpawn(
           spawn.on("exit", () => {
             observer.next(
               actions.setExecutionState({
-                kernelStatus: "process exited",
+                kernelStatus: KernelStatus.NotConnected,
                 kernelRef: action.payload.kernelRef
               })
             );
@@ -488,7 +488,7 @@ export function watchSpawn(
           spawn.on("disconnect", () => {
             observer.next(
               actions.setExecutionState({
-                kernelStatus: "process disconnected",
+                kernelStatus: KernelStatus.NotConnected,
                 kernelRef: action.payload.kernelRef
               })
             );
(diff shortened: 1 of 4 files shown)


applications/desktop/src/notebook/epics/zeromq-kernels.ts
```diff
@@ -488,7 +488,7 @@ export function watchSpawn(
           spawn.on("disconnect", () => {
             observer.next(
               actions.setExecutionState({
-                kernelStatus: "process disconnected",
+                kernelStatus: KernelStatus.NotConnected,
```
captainsafia, line 491: Since it is no longer possible to send/receive messages with the child process after a disconnect, I think "NotConnected" is a reasonable transition here.

packages/reducers/src/core/entities/kernels.ts
```diff
@@ -31,7 +31,7 @@ const byRef = (state = Map(), action: Action): Map<unknown, unknown> => {
       typedAction = action as actionTypes.KillKernelFailed;
       return state.setIn(
         [typedAction.payload.kernelRef, "status"],
-        "failed to kill"
+        KernelStatus.Idle
```
captainsafia, line 34: This is a bit meh. If a kernel failed to kill, then it might still be idle. Then again, maybe `KernelStatus.Error` is the right thing to do here?

willingc: Yeah, I'm thinking `KernelStatus.Error` is probably better.

willingc (approved):
Looks good. Thanks @captainsafia

rgbkrk (approved):
Love the consistency!

PR #5380, opened 2020-11-30
Fix processing of kernelspecs request

captainsafia:
Resolves an issue that I noticed was introduced in https://github.com/nteract/nteract/pull/5353.

In the code below:

https://github.com/nteract/nteract/blob/0626e0a48cfc1355fce385ed92de3539f67ceacc/applications/desktop/src/main/index.ts#L72-L74

`last()` returns an `Observable` which is then processed as a `Kernelspecs` object in the code below when it is not.

https://github.com/nteract/nteract/blob/0626e0a48cfc1355fce385ed92de3539f67ceacc/applications/desktop/src/notebook/epics/zeromq-kernels.ts#L168-L171

To resolve this, we subscribe to the Observable and pass the `Kernelspecs` object from there.

### DIFF
--- a/applications/desktop/src/main/index.ts
+++ b/applications/desktop/src/main/index.ts
@@ -69,9 +69,11 @@ ipc.on("show-message-box", async (event: IpcMainEvent, arg: any) => {
   event.sender.send("show-message-box-response", response);
 });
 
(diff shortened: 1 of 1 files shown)


CrystallineCat (approved):
Thank you for fixing this!

PR #5405, opened 2020-12-18
Fix cursor position issue when invoking language cell magics in Monaco editor

pwang347:
## Checklist

## Context
When invoking cell magics to change the monaco editor language the cursor position will always be erroneously set to the position prior to completing the magic. This is because the prop update for `language` happens before the internal update of the editor cursor position occurs.

cellmagic1

## Solution
Move entire `setModel` logic onto the event queue to ensure that we have the latest cursorPosition.

cellmagic2

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -324,25 +324,39 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
     // Apply new model to the editor when the language is changed.
     const model = this.editor.getModel();
     if (model && language && model.getModeId() !== language) {
-      const newUri = DocumentUri.createCellUri(contentRef, id, language);
-      if (!monaco.editor.getModel(newUri)) {
-        // Save the cursor position before we set new model.
-        const position = this.editor.getPosition();
 
-        // Set new model targeting the changed language.
-        this.editor.setModel(monaco.editor.createModel(value, language, newUri));
-        this.addEditorTopMargin();
+        // Get a reference to the current editor
+        const editor = this.editor;
 
-        // Restore cursor position to new model.
-        if (position) {
-          this.editor.setPosition(position);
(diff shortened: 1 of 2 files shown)


barryt2 (approved):

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -324,25 +324,39 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+            // Set new model targeting the changed language.
+            editor.setModel(monaco.editor.createModel(value, language, newUri));
+            this.addEditorTopMargin();
+
+            // Restore cursor position to new model.
+            if (position) {
+              editor.setPosition(position);
+            }
+
+            // Set focus
```
jruales, line 351: Why was this "Set focust" added here? I see that there's another `Set focus` after the `setTimeout`?

pwang347: The focus is necessary here because setting the model will unfocus, and by placing the set model code onto the event queue we no longer guarantee the focus to be set immediately after

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -324,25 +324,39 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
-      }
+        setTimeout(() => {
+          const newUri = DocumentUri.createCellUri(contentRef, id, language);
+          if (!monaco.editor.getModel(newUri)) {
+            // Save the cursor position before we set new model.
+            const position = editor.getPosition();
+
+            // Set new model targeting the changed language.
+            editor.setModel(monaco.editor.createModel(value, language, newUri));
+            this.addEditorTopMargin();
```
captainsafia, line 344: Curious: why do we need to call this here?

captainsafia (approved):

PR #5407, opened 2020-12-21
Fix issues with handling of kernel completions

pwang347:
## Checklist

## Context
There were two issues with the previous approach of sanitization:
1. Anything that had a "." in the completion was being rewritten to be only the part trailing the "." when this wasn't always what we wanted. E.g. folders generally support "." and file names with extensions do too!

### [OLD] File path with "."
intellisense3

2. We also want to apply a similar logic to paths, e.g. for a path `some/path` we want to only be suggested `path` instead of `some/path` because otherwise we end up with `some/some/path` after accepting the completion

### [OLD] Full file paths were being offered
intellisense5

### [OLD] File path with whitespace
intellisense4

## Solution
We should make use of the context available for the completion. In other words, in the example above we should check that we are currently receiving suggestions off of `some/` and use this to our advantage by removing it from the suggestion.

### [FIXED] Paths with "."
intellisense1

### [FIXED] Paths with whitespace
intellisense2

## Known issues
See https://github.com/ipython/ipython/issues/2463 - we just make sure that the completion up to that point works

### DIFF
--- a/packages/monaco-editor/src/completions/completionItemProvider.ts
+++ b/packages/monaco-editor/src/completions/completionItemProvider.ts
@@ -272,7 +272,9 @@ class CompletionItemProvider
       // This is then used to figure out what we should actually insert.
       // example: context = "a/path/to/some." and text = "some.folder.name" should produce "folder.name"
       const completionContext = context.substr(context.lastIndexOf("/") + 1);
-      text = text.substr(completionContext.length);
+      if (text.startsWith(completionContext)) {
+        text = text.substr(completionContext.length);
+      }
     }
 
     for (let i = 0; i < percentCount; i++) {
--- a/packages/monaco-editor/src/completions/completionItemProvider.ts
+++ b/packages/monaco-editor/src/completions/completionItemProvider.ts
@@ -152,7 +152,7 @@ class CompletionItemProvider
       if (typeof match === "string") {
         const text = this.sanitizeText(match, context);
         const inserted = this.getInsertText(text, context);
-        const filtered = this.getFilterText(text);
+        const filtered = this.getFilterText(text, context);
         return {
           kind: this.adaptToMonacoCompletionItemKind(unknownJupyterKind),
           label: text,
@@ -190,7 +190,7 @@ class CompletionItemProvider
         }
 
         const text = this.sanitizeText(match.text, context);
-        const filtered = this.getFilterText(text);
+        const filtered = this.getFilterText(text, context);
         const inserted = this.getInsertText(text, context, percentCount);
         return {
           kind: this.adaptToMonacoCompletionItemKind(match.type),
@@ -232,7 +232,8 @@ class CompletionItemProvider
       // We should return only the last part of the path, wrapped in double quotes
       const completionIsPathWithWhitespace = text.startsWith('"') && text.endsWith('"') && text.length > 2; // sanity check: not empty string
       if (completionIsPathWithWhitespace && text.substr(1).startsWith(context)) { // sanity check: the context is part of the suggested path
-        return `"${text.substr(context.length+1)}`;
+        const toRemove = context.substr(0, context.lastIndexOf("/") + 1);
+        return `"${text.substr(toRemove.length+1)}`;
       }
 
       // Otherwise, display the most specific item in the path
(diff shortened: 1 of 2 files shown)


packages/monaco-editor/src/completions/completionItemProvider.ts
```diff
@@ -137,14 +137,26 @@ class CompletionItemProvider
       matches = results.metadata._jupyter_types_experimental;
     }
 
+    // retrieve the text that is currently typed out which is used to determine completion
+    const startPos = model.getPositionAt(results.cursor_start);
```
barryt2, line 141: is cursor_start and cursor_end always defined in results?

barryt2 (approved):

packages/monaco-editor/src/completions/completionItemProvider.ts
pwang347: yes, the interface `CompletionResults` has required number types for these two values

pwang347:
Verified same scenarios on Python 2.7.15.

Note that the following scenario
```
myarray = []
myarray.
```
Isn't provided by the Python 2.7.15 kernel at all

pwang347:
@captainsafia please let me know if there are any next steps before merging! thanks

packages/monaco-editor/src/completions/completionItemProvider.ts
```diff
@@ -203,17 +215,36 @@ class CompletionItemProvider
...
    * @param text Text of Jupyter completion item
    */
-  private sanitizeText(text: string) {
-    const index = text.lastIndexOf(".");
-    return index > -1 && index < text.length - 1
-      ? text.substring(index + 1)
-      : text;
+  private sanitizeText(text: string, context: string) {
+    // If we have whitespace within a path, we should just return quotes around that
+    if (context.includes("/") && text.startsWith('"') && text.endsWith('"') && text.length > 2 && text.substr(1).startsWith(context)) {
```
captainsafia: Nit: Can you assign these conditionals to some reader-friendly values? Eg.

```
const isItGreen = value === "green";
```

Also, the comment states "If we have whitespace within a path, we should just return quotes around that" but I don't see where we are asserting that the `text` contains whitespace?

packages/monaco-editor/src/completions/completionItemProvider.ts
```diff
@@ -203,17 +215,36 @@ class CompletionItemProvider
...
+      return `"${text.substr(context.length+1)}`;
```
captainsafia: Why only a beginning quote and not an end quote?

captainsafia (commented):
Thanks for your patience. I was OOF.

Looks good overall. The GIFs were help. I added a few questions to help address some gaps in my understanding.

Also maybe we want to move some of this logic to a separate utility file within the completions directory that we reference here so it's a little clearer that these address a subset of the completions.

packages/monaco-editor/src/completions/completionItemProvider.ts
pwang347: Done! I also updated the comment so that hopefully it makes a bit more sense.

For simplicity, I'm only checking for the presence of quotes for something in the form of "a/path with spaces" because that's what the kernel returns for any paths that contain spaces.

However, I also first make sure that we are currently typing a path, and later that the completion contains the current typed context to make sure it doesn't just work for any string.

packages/monaco-editor/src/completions/completionItemProvider.ts
pwang347: Since `text` here is guaranteed to be of form `"some/path with spaces"` I'm just dropping the `"some/` part of it with the substring, so to get `"path with spaces"` I still need to add one back to the beginning

packages/monaco-editor/src/completions/completionItemProvider.ts
pwang347: Also, re: moving to another utility file - we don't seem to be handling too many cases here at the moment, just 1) path completions 2) member completions (`object.member`) and 3) magics (`%`). Would it make sense to just keep them here for now so it's easier to read?

captainsafia (approved):
Fair enough wrt not moving to a separate file. I think a combination of the code comments + variable names gives us what we need here.

PR #5418, opened 2021-01-12
Add new configuration prop to register custom commands

resaldiv:
### Checklist

### Context
In a monaco editor instance, a command can be used upon the acceptance of a [completion item](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.completionitem.html). This is helpful to trigger specific actions and log events, so it would be nice to add support to register custom commands in the monaco editor.

### Solution
A way to achieve this, is by adding a command handler to the monaco configuration props (similar to the shortcuts handler). In this way, a user can call the [addCommand](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandalonecodeeditor.html#addcommand) method and add some logic on what type of action will be triggered.

```
<MonacoEditor
    ...
    commandHandler={myCommandHandler}
/>

myCommandHandler(editor) {
    // Add some logic on what type of action will be triggered
    const commandHandler = () => {
        alert('My command is executing');
    };

    const myCommand = editor.addCommand(0, commandHandler);
    
    // The alert will be triggered upon the acceptance of this completion item
    const myCompletionItem = {
        ...
        command: myCommand
    }
}
```

This PR adds an optional prop to the `IMonacoConfiguration`, so that a user can register and handle custom commands in the editor.

### Side notes
While running the `TEST_PLAN.md` I opened the **Explore Your World with GeoJSON** example and the maps look like this (not sure if it's a bug or not):
image

### DIFF
--- a/changelogs/current_changelog.md
+++ b/changelogs/current_changelog.md
@@ -176,7 +176,7 @@ Provide a bulleted list of breaking changes and a reference to the PR(s) contain
 
 #### New Features
 
-Provide a bulleted list of new features or improvements and a reference to the PR(s) containing these changes.
+- Add support to register custom commands using the command handler ([#5418](https://github.com/nteract/nteract/pull/5418))
 
 #### Bug Fixes
 
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -317,6 +317,11 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       this.props.cursorPositionHandler(this.editor, this.props);
     }
 
+    // Handle custom commands
+    if (this.editor && this.props.commandHandler) {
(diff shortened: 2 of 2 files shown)


captainsafia:
> A way to achieve this, is by adding a command handler to the monaco configuration props (similar to the shortcuts handler). In this way, a user can call the addCommand method and add some logic on what type of action will be triggered.

Can you share a code sample of how this API would be used?

resaldiv:
@captainsafia thanks for taking a look into this! I updated the description to include an example.

captainsafia (approved):
Looks fine but cc @vivek1729 since has more insight on the Monaco stuff

vivek1729 (approved):

resaldiv:
Thank you @captainsafia and @vivek1729 for the review. I did a minor change to include this functionality when the component gets updated.

Issue #5445, opened 2021-02-11
Unable to load mythic-configuration in browser due to node package reference


erem-stripe:
**Application or Package Used**
mythic-configuration

**Describe the bug**
Importing mythic-configuration into a browser setting results in this stack trace:
```
ERROR in ./node_modules/mkdirp/index.js
Module not found: Error: Can't resolve 'fs' in '.../node_modules/mkdirp'
 @ ./node_modules/mkdirp/index.js 2:9-22
 @ ./node_modules/fs-observable/lib/index.js
 @ ./node_modules/@nteract/mythic-configuration/lib/backends/filesystem.js
 @ ./node_modules/@nteract/mythic-configuration/lib/index.js
 @ ./node_modules/@nteract/stateful-components/lib/config-options.js
 @ ./node_modules/@nteract/stateful-components/lib/index.js
```

**To Reproduce**

1. Start with a react application in the browser
2. Add `mythic-configuration` as a dependency
3. Attempt to import mythic-configuration
4. Note the stack trace

**Expected behavior**
I expect no stack trace. I expect mythic-configuration not to import any node-specific modules by default.

**Desktop (please complete the following information):**

- OS: macOS
- Browser: chrome
- Version: 88

**Additional context**
The stack trace tells the story, but essentially by the root cause is that `mythic-configuration/index.js` imports `backends/filesystem.js` which imports `fs-observable`, which is node-specific and therefore breaks in the browser.

If index.js did a dynamic import of filesystem only as requested it would probably fix this ticket.

captainsafia:
@erem-stripe Thanks for the PR! After taking a look at it, I decided it's best to remove the top-level export of `setConfigFile` from mythic-configuration altogether. Since it's backend specific (and really only used for the desktop app), it doesn't make sense to export this as part of the top-level API.

PR #5390, opened 2020-12-03
Remove electron dependency from mythic-notifications

captainsafia:
Closes #5389.

@CrystallineCat mythic-notifications is used in nteract web but the mythic-window dependency brings in Electron which uses Node APIs that are not in the browser.

Temporarily removing the APIs that use mythic-windowing to resolve this issue to unblock using the packages in web environments.

PR #5446, opened 2021-02-12
Fix mythic-configuration: do not import node packages by default

erem-stripe:
Fixes issue #5445
https://github.com/nteract/nteract/issues/5445

### DIFF
--- a/changelogs/current_changelog.md
+++ b/changelogs/current_changelog.md
@@ -24,6 +24,8 @@ Provide a bulleted list of new features or improvements and a reference to the P
 
 Provide a bulleted list of bug fixes and a reference to the PR(s) containing the changes.
 
+- Fixed mythic-configuration, allowing it to be imported in a browser setting ([Issue #5445](https://github.com/nteract/nteract/issues/5445))
+
 ## Core SDK Packages
 
 ### @nteract/actions ([publish-version-here])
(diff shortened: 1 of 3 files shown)


erem-stripe:
Converting this to draft bc it turns out that, in the webpack setting, our nice little try/catch only tries and doesn't catch

erem-stripe:
Ok unit tests pass and I've verified that this works for my use-case. It's a little janky with the require-fu and I'm open to feedback on that.

Regarding the test plan, I'm getting out-of-memory exceptions trying to get `dist:all` to work but I'm hoping the reviewer can help me by running through the test plan?

erem-stripe:
Also, I noticed the nteract-web doesn't blow up in this way which had me going crazy! Turns out it's because of [this next.js configuration](https://github.com/nteract/nteract/blob/main/applications/web/next.config.js#L8-L13)

I think this PR (or a less janky version of it) is still worth it because not every web-based consumer of the UI libraries should have to remember to vigilantly stub out `fs` just to load a configuration package!

captainsafia (approved):
Thanks for the code change and the test!

I did a similar fix over in https://github.com/nteract/nteract/pull/5390 but missed that we had another Node-specific dependency in the platform-agnostic dependencies.

PR #5480, opened 2021-04-01
add docs deploy file

willingc:
use the now.json file from the old docs repo.


willingc:
Closing as unneeded. Was able to configure with CLI alias and GUI.

PR #5492, opened 2021-04-09
User guide edits

jeromefv:
# Edits
- Changes to user guides
- Updated home page language

### DIFF
--- a/docs/creating-data-visualizations-in-nteract.md
+++ b/docs/creating-data-visualizations-in-nteract.md
@@ -7,12 +7,12 @@ Data visualizations are a great way to communicate large amounts of information 
 Matplotlib is one of the most popular plotting tools in the open source ecosystem. A variety of academic and industry contexts use the Python library and it is a great starting point for anyone looking to dive into data visualization.
 
 The nteract example notebook below demonstrates getting started with Matplotlib.
-Install the required dependencies for this example notebook before continuing. Python setup, Installing the dependencies varies depending on your Python setup.
+Install the required dependencies for this example notebook before continuing. For [Python](https://www.python.org/downloads/) setup, install the dependencies according to your use case.
 
-The example below of the necessary dependency installations refer to a global Python3 installation.
+The example below of the necessary dependency installations refer to a global Python installation.
 
 ```
-$ python3 -m pip install pandas requests matplotlib seaborn
+$ python -m pip install pandas requests matplotlib seaborn
 ```
 
 After installation, open the example notebook in the menu bar.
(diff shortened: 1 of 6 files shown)


docs/kbd-shortcuts.md
```diff
@@ -47,54 +50,58 @@ The syntax is the following:
...
-* ZoomOut
-* ZoomReset
+nteract recognizes changes and displays new keyboard shortcuts in the menu after saving the config file. 
+
+> NOTE: No restart is required for updates.
+
+| **Supported commands:**| 
+| --- |
+| About |
+| BringAllToFront |
```
captainsafia, line 60: Maybe we should add a second column to these that describes what they do?

We can probably also remove the "About" from this since it just brings up the about page.

docs/creating-data-visualizations-in-nteract.md
```diff
@@ -1,53 +1,76 @@
...
 
 ### Plotting with Matplotlib and seaborn
 
-Matplotlib is one of the most popular plotting tools in the open source ecosystem. It's used in a variety of academic and industry contexts and is a great starting point for anyone looking to dive into data visualization.
+Matplotlib is one of the most popular plotting tools in the open source ecosystem. A variety of academic and industry contexts use the Python library and it is a great starting point for anyone looking to dive into data visualization.
 
-To get started with Matplotlib, we'll use another nteract example notebook. Before we get started, make sure you have the dependencies needed for this example notebook installed. The installation process will vary depending on your Python setup, but here is an example of the dependencies that need to be installed assuming a global Python3 installation.
+The nteract example notebook below demonstrates getting started with Matplotlib.
+Install the required dependencies for this example notebook before continuing. Python setup, Installing the dependencies varies depending on your Python setup.
```
captainsafia: I assume Python install instructions will be added or linked to from here?

docs/creating-data-visualizations-in-nteract.md
```diff
@@ -1,53 +1,76 @@
...
+
+The example below of the necessary dependency installations refer to a global Python3 installation.
```
captainsafia: Since Python 2 is out of support, I think we should just say "Python installation" and use `python".

docs/kbd-shortcuts.md
jeromefv: Removed the "About" line, though I think that the supported commands are self-explanatory for a developer. If there are any specifics with these commands that aren't clear in the name or the context, it would be a good add!

docs/creating-data-visualizations-in-nteract.md
jeromefv: I added a [Python](https://www.python.org/downloads/) link to that text as well as cleaned up the broken appositive for line 10.

docs/kbd-shortcuts.md
captainsafia: Fair enough. Can we normal case the columsn: e.g. `BringAllToFront` -> "Bring all windows to front"

willingc:
Thanks @jeromefv for the PR and @captainsafia for reviewing.

PR #5494, opened 2021-04-13
Move to the new updateOptions API to dynamically update editor options

vivek1729:
As we bumped up monaco editor to the latest version, we transition to the new [`updateOptions`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandalonecodeeditor.html#updateoptions) API to dynamically update some editor constructor options.

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -381,9 +381,12 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         }, 0);
     }
     
+    const monacoUpdateOptions: monaco.editor.IEditorOptions & monaco.editor.IGlobalEditorOptions = { readOnly: this.props.readOnly };
     if (theme) {
-      monaco.editor.setTheme(theme);
+      monacoUpdateOptions.theme = theme;
     }
+
+    this.editor.updateOptions(monacoUpdateOptions);
 
     // In the multi-tabs scenario, when the notebook is hidden by setting "display:none",
     // Any state update propagated here would cause a UI re-layout, monaco-editor will then recalculate


captainsafia:
@vivek1729 What's the benefit of this API over the previous one?

vivek1729:
@captainsafia, the older version of monaco editor did not have a generic API to update editor options dynamically. We therefore had to rely on specific getters/setters for individual props like `theme`. This completely simplifies the update process of editor props for re-renders. So instead of using specific methods like `setTheme`, `setPosition`, `setValue` etc, we can just move to a single `updateOptions` call for a set of configuration props.

captainsafia:
@vivek1729 Got it. Should we be replacing instances of `setValue` and other properties with this then?

vivek1729:
That's a good point @captainsafia. On digging deeper, it seems like the [`updateOptions`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandalonecodeeditor.html#updateoptions) API accepts a subset of properties ([IEditorOptions](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.ieditoroptions.html) and [IGlobalEditorOptions](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.iglobaleditoroptions.html)) that we pass in as [constructor options](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandaloneeditorconstructionoptions.html) when creating the editor. So the `value` property is a constructor option that has a dedicated setter namely `setValue`. Same goes for the position prop. So, I think we can continue using the `setValue` method and move to `updateOptions` for configuration properties like `theme`, `readOnly` etc highlighted in the two interfaces I linked earlier.

captainsafia (approved):
It looks like we're already using these for the two properties we care about.

vivek1729:
Thanks @captainsafia. Ok to merge now?

PR #5499, opened 2021-04-14
update gh-actions CI using official docs

willingc:


.github/workflows/ci-workflow.yml
```diff
@@ -26,13 +31,18 @@ jobs:
         run: yarn test
   
   validate_dev_setup:
```
captainsafia: Can we actually nuke the validate_dev_setup issue? After fixing the import issue that was failing the build yesterday, there are still transient issues with installing electron on the VM. All these failures make these checks way less helpful.

willingc: Sure thing. I will remove and we can always add back later.

PR #5507, opened 2021-04-17
CI cleanup

willingc:
- Update GH Action with HARD_LINK by @captainsafia 
- Add caching to GH Action
- Update README with GH Action badge
- Remove Azure Pipelines config


willingc:
@captainsafia This may still be flaky from time to time re: dependency install and electron. I've added the ability to restart and start runs manually to the actions config file.

willingc:
Going to go ahead and merge this and see how it does for the next few days.

PR #5522, opened 2021-04-21
Fix monaco editor theme conflicts

vivek1729:
Monaco editor by default comes with the following themes: `vs`, `vs-dark` and `hc_black`. While the package allows us to inherit and define custom themes, I noticed that we override the `vs-dark` theme in nteract. This can have unintended side effects on a consumer of this package. For instance, consider an app that has its own monaco editor package dependency and also for some reason relies on the nteract/moanco editor package. If the app already has an instance of monaco editor loaded with the default `vs-dark` theme (not from the `nteract/monaco-editor`) package,  loading the nteract/monaco-editor package would dynamically update the theme with our custom overrides. This happens because monaco editor has a global singleton for themes. This is already tracked as an [issue](https://github.com/microsoft/monaco-editor/issues/1713).

We update the theme definitions and prefix our strings with `nteract` to reduce the changes of theme name collisions and accidental overrides in scenarios I described above.

### DIFF
--- a/packages/monaco-editor/src/theme.ts
+++ b/packages/monaco-editor/src/theme.ts
@@ -42,7 +42,7 @@ monaco.editor.defineTheme(DarkThemeName, customMonacoDarkTheme);
 /**
  * The custom high contrast light theme with customized background
  */
-export const HCLightThemeName = "nteract-hclight";
+export const HCLightThemeName = "nteract-hc-light";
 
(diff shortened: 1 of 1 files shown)


jruales (approved):

willingc (approved):
Looks good. Thanks @vivek1729

PR #5534, opened 2021-05-04
Docs Edit and Proofreading

jeromefv:
# Proofread, edit
- Guides for SDK Groups
    - Components, Core, Editors, Jupyter, Mythic
        - Changes to text and formatting as needed

### DIFF
--- a/docs/groups/components-group.md
+++ b/docs/groups/components-group.md
@@ -10,13 +10,13 @@ The **Components** group is the set of SDK packages for displaying and styling n
 
 ## /stateful-components
 
-The @nteract/stateful-components package exports a set of React components that are connected to the Redux and an unconnected set of components.
+The /stateful-components package exports a set of React components that are connected to the Redux and an unconnected set of components.
 
 ### Using /stateful-components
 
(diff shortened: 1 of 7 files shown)


docs/groups/components-group.md
```diff
@@ -10,13 +10,13 @@ The **Components** group is the set of SDK packages for displaying and styling n
 
 ## /stateful-components
 
-The @nteract/stateful-components package exports a set of React components that are connected to the Redux and an unconnected set of components.
+The /stateful-components package exports a set of React components that are connected to the Redux and an unconnected set of components.
```
captainsafia: I think listing the full package name is better since that is how it'll appear in the installation commands and in their package.json.

jeromefv: Sounds good! In that case, I'll change each reference when first introducing the package to have the pattern of `@nteract/specific-package` with the code formatting included. I believe that's already the case for most references that way, so it shouldn't change too much otherwise.

willingc (approved):
Thanks @jeromefv for the PR and @captainsafia for the review.

PR #5538, opened 2021-05-13
Use 'scrolled' instead of 'collapsed' metadata property for output scrolling

### LINKED ISSUE #5527: Non-standard use of "collapsed" metadata property
nteract web

**Describe the bug**
in #4972, nteract started to use the "collapsed" metadata property to indicate whether an output is shown in a scrolled div or otherwise shown in its entirety. However, the standard field for that is not "collapsed", but rather "scrolled" (see official ipynb schema [here](https://github.com/jupyter/nbformat/blob/373d926154d47fa09c453958720786f8ee8aa737/nbformat/v4/nbformat.v4.schema.json#L252))

In Jupyter Classic, this is what an output with `"scrolled": true` looks like (which is what we want):
image

Whereas this is an output with `"collapsed": true`
image

Note that in a collapsed output there's a clickable rectangle that shows up and that allows expanding the output:
image

jruales:
Using "scrolled" instead of "collapsed" for representing outputs with a scrolled div, to better align with [official ipynb format](https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json).

### DIFF
--- a/packages/commutable/src/cells.ts
+++ b/packages/commutable/src/cells.ts
@@ -25,7 +25,6 @@ export const makeCodeCell = Record<CodeCellParams>({
   cell_type: "code",
   execution_count: null,
   metadata: ImmutableMap({
-    collapsed: true,
     jupyter: ImmutableMap({
       source_hidden: false,
       outputs_hidden: false,
--- a/packages/commutable/src/v3.ts
+++ b/packages/commutable/src/v3.ts
@@ -86,6 +86,8 @@ export interface HeadingCell {
 
 export interface CodeCell {
   cell_type: "code";
+  language: string;
+  collapsed: boolean;
   metadata: JSONObject;
   input: MultiLineString;
   prompt_number: number;
--- a/packages/commutable/src/v3.ts
+++ b/packages/commutable/src/v3.ts
@@ -86,8 +86,6 @@ export interface HeadingCell {
 
 export interface CodeCell {
   cell_type: "code";
-  language: string;
-  collapsed: boolean;
   metadata: JSONObject;
   input: MultiLineString;
   prompt_number: number;
--- a/changelogs/current_changelog.md
+++ b/changelogs/current_changelog.md
@@ -10,6 +10,8 @@
 
 Provide a bulleted list of breaking changes and a reference to the PR(s) containing those changes.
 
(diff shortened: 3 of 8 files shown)


packages/commutable/src/v3.ts
```diff
@@ -86,8 +86,6 @@ export interface HeadingCell {
 
 export interface CodeCell {
   cell_type: "code";
-  language: string;
```
miduncan: Why are we touching the language field?

miduncan (commented):
For existing notebooks with the collapsed field, how are we handling that? Are we just not doing anything with them?

packages/commutable/src/v3.ts
jruales: The `language` and `collapsed` fields are not being used here, so I removed them as cleanup

jruales:
> For existing notebooks with the collapsed field, how are we handling that? Are we just not doing anything with them?

@miduncan At the moment, we're not doing anything about those fields. Do you think that we should wipe that field from any notebook we open? Note that the collapsed field indicates in Jupyter classic that an output is collapsed like this:
image

packages/commutable/src/cells.ts
```diff
@@ -25,7 +25,6 @@ export const makeCodeCell = Record<CodeCellParams>({
   cell_type: "code",
   execution_count: null,
   metadata: ImmutableMap({
-    collapsed: true,
```
captainsafia, line 28: Why not default to `scrolled=true` here?

jruales: - Jupyter Notebook by default doesn't define it
- The actual default value we would want is `"auto"`, but I haven't implemented `"auto"` scrolling yet. ("scrolled" can be true, false, or "auto", and it's an optional property)
- If the `collapsed` property is set to `true`, then it would not be able to go back to `"auto"` by toggling. At least in Jupyter Notebook, the value only acts as "auto" as default, and then when you toggle it, it just goes to true or false and never back to "auto". So I wouldn't want notebooks out there to all have "scrolled":"true" since otherwise the auto scrolling would not apply to those existing notebooks

packages/commutable/src/v3.ts
jruales: I've reverted the changes to v3.ts since they seem to be for an older version of commutable and I don't want to touch that

jruales:
@miduncan @captainsafia any other comments?

captainsafia (approved):
This LGTM. I suspect the implementation of `auto` support will shift things quite a bit.

miduncan (approved):

MSeal:
Thanks @jruales !

PR #5553, opened 2021-05-27
Add 'auto' scrolled mode

jruales:
Added support for `"scrolled": "auto"` for outputs (see "scrolled" property in [ipynb format schema](https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json)), replicating behavior in classic Jupyter Notebook.

When "auto" scroll mode is set for a cell (or when "scrolled" is not defined, as "auto" is the default), then a cell's output will have scroll enabled if and only if its output's height exceeds a certain tall threshold (currently 1800px). The scrolling can still be set manually by setting `"scrolled": true` or `"scrolled": false`, or in the nteract UI by using the "Extended Output" toggle.


jruales:
@captainsafia @miduncan could I please get some feedback on the approach I took for "auto" scroll mode before I add more tests, run through the TEST_PLAN.md, and mark the PR as ready for review? I ended up adding state to the Outputs component rather than to the Redux store since I think that whether the auto scroll is currently scrolling or not is more of a detail of this component and about the presentation rather than notebook's state itself. An advantage of this approach is that it's self-contained within the Outputs component.

packages/stateful-components/__tests__/outputs/index.spec.tsx
```diff
@@ -72,6 +72,6 @@ describe("makeMapStateToProps", () => {
     const ownProps = { contentRef, id };
     const result = makeMapStateToProps(state, ownProps)(state);
     expect(result.hidden).toBe(false);
-    expect(result.expanded).toBe(false);
+    expect(result.scrolledValue).toBe("auto");
   });
 });
```
miduncan, line 77: Do we have a test for the "Extended output" option you mentioned in the docs?

packages/stateful-components/src/outputs/index.tsx
```diff
@@ -34,6 +65,36 @@ export class Outputs extends React.PureComponent<ComponentProps & StateProps> {
...
+    // "scrollHeight is a measurement of the height of an element's content, including content not visible on the screen due to overflow"
+    const heightOfOutputs = wrapperDiv.scrollHeight;
+    const shouldScroll = this.autoScrollShouldScroll(heightOfOutputs);
+    if (shouldScroll !== this.state.computedScrolledValue) {
+      this.setState({ computedScrolledValue: shouldScroll })
+    }
+  }
+
+  autoScrollShouldScroll(heightOfOutputs: number) {
+    return heightOfOutputs > AUTO_SCROLL_HEIGHT_THRESHOLD;
```
miduncan: nit: I think of threshold as the point where auto scroll would start, so should this be >=?

miduncan (approved):

packages/stateful-components/__tests__/outputs/index.spec.tsx
jruales: Done

packages/stateful-components/src/outputs/index.tsx
jruales: Done

Issue #5547, opened 2021-05-25
Multi-client kernel support 


shibbas:
Happy to make this contribute this change to nteract! Please let me know if there is any feedback on the proposed solution below. 

**Application or Package Used**
nteract on jupyter. And nteract packages. 

**Describe the bug**
When you open the same notebook in two browser sessions, the websocket connection of the second session stomps on the websocket connection of the first terminating the cell execution subscriptions. 

**To Reproduce**
Create a notebook with the following code:
```python
import time;

for i in range(100):
    print(i)
    time.sleep(1)
```
Start execution and open the same notebook in another browser. Execution shows [*] in the first browser, but output stops updating. 

**Expected behavior**
The first notebook execution should continue and the second notebook execution if started gets queued up. This seems to be Jupyter behavior. 

**Proposed solution** 
The disconnects are happening because the server session ID is being used when generating a websocket URL. It looks like nteract only supports single client sessions per notebook instance. 

[Jupyter messaging protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html)  supports client session ids. From the docs: 

> A client session id, in message headers from a client, should be unique among all clients connected to a kernel. When a client reconnects to a kernel, it should use the same client session id in its message headers. When a client restarts, it should generate a new client session id.

This client session id is also used as the sessionId when creating the websocket to differentiate the connection between multiple clients in Jupyter. 

Using kernelRef as the client sessionID and storing both the client sessionId as well as the server session ID (since it's used for session API) can enable the multi-client support.

PR #5557, opened 2021-06-02
[Change] add support for multiclients in remote connection

shibbas:
## Checklist

## Change
Adding support for multiple web clients to connect to same remote kernel

This addresses the issue as described here: https://github.com/nteract/nteract/issues/5547 

Tested the repro using the nteract_on_jupyter app which seems to be using the remote kernel props. The desktop app doesn't look like it's using websockets. 

| Before: | After: |
|--------------|-------------|
| When the second client's cell execution starts, the first client's websocket closes and the output stream errors out but the kernel is still busy so the second client does not execute until the kernel becomes idle again. |First client's cell execution completes to the end while second client's request queues up and starts processing when kernel becomes idle.|
|multiclient-before|    multiclient-after |

### DIFF
--- a/packages/epics/src/websocket-kernel.ts
+++ b/packages/epics/src/websocket-kernel.ts
@@ -69,12 +69,14 @@ export const launchWebSocketKernelEpic = (
         mergeMap(data => {
           const session = data.response;
 
-          const sessionId = castToSessionId(session.id);
+          const sessionId = castToSessionId(kernelRef);
+          const remoteSessionId = castToSessionId(session.id);
 
           const kernel: RemoteKernelProps = Object.assign({}, session.kernel, {
             type: "websocket",
             info: null,
             sessionId,
+            remoteSessionId,
             cwd,
             channels: kernels.connect(
               serverConfig,
@@ -139,8 +141,8 @@ export const changeWebSocketKernelEpic = (
       if (!oldKernel || oldKernel.type !== "websocket") {
         return empty();
       }
-      const { sessionId } = oldKernel;
-      if (!sessionId) {
+      const { sessionId, remoteSessionId } = oldKernel;
+      if (!sessionId || !remoteSessionId) {
         return empty();
       }
 
@@ -163,7 +165,7 @@ export const changeWebSocketKernelEpic = (
           };
           // The sessions API will close down the old kernel for us if it is
           // on this session
-          return sessions.update(serverConfig, sessionId, sessionPayload).pipe(
+          return sessions.update(serverConfig, remoteSessionId, sessionPayload).pipe(
             mergeMap(({ response: session }) => {
               const kernel: RemoteKernelProps = Object.assign(
                 {},
@@ -171,6 +173,7 @@ export const changeWebSocketKernelEpic = (
                 {
                   type: "websocket",
                   sessionId,
+                  remoteSessionId,
                   cwd,
                   channels: kernels.connect(
                     serverConfig,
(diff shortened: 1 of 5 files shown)


shibbas:
@willingc, @codeseal - PR for multiclient kernel support. Please let me know if there is any feedback!

captainsafia:
Requesting review from Matt and Carol who have more context here since I missed the meeting where this was discussed. TY in advance for the review.

shibbas:
Please let me know if you need me to provide any more information. The proposed solution is also described on the issue I had opened before: https://github.com/nteract/nteract/issues/5547

captainsafia:
@shibbas Not sure if it would help but might be useful to nudge folks about this at tomorrow's dev meeting. We can try to do a group review of the PRs then and get them merged in.

shibbas:
> @shibbas Not sure if it would help but might be useful to nudge folks about this at tomorrow's dev meeting. We can try to do a group review of the PRs then and get them merged in.

Thanks @captainsafia. I missed today's meeting. Will join next week.

captainsafia:
Heads up that automatic releases are now enabled on the repo (see #5568 for more info). To support automated releases, please make sure that the commits in this PR adhere to the conventional commit format (see https://www.conventionalcommits.org/en/v1.0.0/). Once this PR is merged, a release will be triggered for the modified packages based on the commit message.

willingc (approved):
@MSeal This looks good to me.

cc/ @rgbkrk @ivanov FYI

rgbkrk (approved):

packages/epics/src/websocket-kernel.ts
```diff
@@ -69,12 +69,14 @@ export const launchWebSocketKernelEpic = (
         mergeMap(data => {
           const session = data.response;
 
-          const sessionId = castToSessionId(session.id);
+          const sessionId = castToSessionId(kernelRef);
+          const remoteSessionId = castToSessionId(session.id);
```
rgbkrk, line 73: Interesting. This tells me we likely should have a `sessionRef` since the refs are supposed to be for the local id-like entities.

willingc: So basically swapping out `session.id` for `sessionRef` or `remoteSessionRef`.

MSeal (approved):
LGTM, the change is simple and to the point which I like

packages/epics/src/websocket-kernel.ts
rgbkrk: `remoteSessionRef` is really our session ID. From [`packages/core/plan.md`](https://github.com/nteract/nteract/blob/df958a94565520df1645b54ba7218b24be4e1a6a/packages/core/plan.md#core-concepts):

- `ref` - an internal _reference_ to an entity upon _recognition_, e.g. kernels, hosts, kernelspec collections, etc.
- `id` - likely an external identifier, e.g. with /api/kernels/9092, 9092 is the id

> We use the term _recognition_ over _creation_ because we want to have a way to
reference an entity _before_ we get a response from some api. A good example is
having a _ref_ for an active kernel before the kernel has been launched with a
jupyter notebook server. Since there will be a proliferation of id-strings, the
internal ones are called _refs_ and they are only meant for use inside the
application--i.e., they have no meaning externally. The external id-string that
will typically be found is called `id`.

shibbas:
Thanks for the review!

packages/epics/src/websocket-kernel.ts
shibbas: I see. In this case the sessionID is really _the_ clientSessionID which is sent to the server and remoteSessionID _is_ an external entity. Should I change these to be sessionRef (for the clientSessionID) and keep the sessionID to identify the ID returned by the server? I used remoteSessionId because it was clearer indication of where the ID was coming from. 

Happy to change if that's the recommendation! @rgbkrk

willingc:
@captainsafia Does this PR need anything to be merged for CI/changelog?

Issue #5549, opened 2021-05-25
Provide the execute_reply.status on cell execution 


shibbas:
Happy to contribute this to nteract. Feedback appreciated.  

**Is your feature request related to a problem? Please describe.**
Currently it looks like the "ok" | "aborted" | "error" status from the kernel on cell execution is not being raised during cell execution along with update_cell_status or updating execution counts. 

This is part of the messaging protocol that is coming directly from the kernel which is helpful to understand what happened to the execution on the kernel. https://jupyter-client.readthedocs.io/en/stable/messaging.html#execution-results

While information on the success or failure of the cell execution can be interpreted by the execution of the cell and the status, it will be more deterministic to provide the information coming directly from the kernel. 

**Describe the solution you'd like**
Raise an action on cell execution on the cell execution result on CELL_EXECUTION_SUCCESS ("ok") | CELL_EXECUTION_FAILED ("error") | CELL_EXECUTION_ABORTED ("aborted")

willingc:
"aborted" gets lost completely.

MSeal:
The described solution sounds like the right path. We should make a follow-up issue to render the cell specially for abort events.

PR #5561, opened 2021-06-05
[Fix] Update Cell execution result and align client and server state

shibbas:
## Checklist

## Context

Related Issue - https://github.com/nteract/nteract/issues/5549

This PR addresses a few things: 
- BREAKING: Uses the jupyter messaging protocol defaults for execute_request.stop_on_error. 
- BREAKING: Removes the EXECUTE_CANCEL_ALL and instead uses the kernel returned state in execute_reply status to determine whether to cancel a cell or not. 
- Adds the execute_reply status to the transient metadata via a new action that can be used by applications to determine the success / failure of the cell or whether the cell was aborted. 

## Proposal

It looks like client and kernel states on the cell execution get out of alignment around cell errors. Jupyter defaults stop_on_error for execute_request to true which stops execution on the server for the queued up cells. However, the default in nteract/messaging is set to false, so even though on error the client cells are no longer shown as executing, the execution is till happening on the kernel causing the client and kernel to get out of sync. 

### Before 
Due to the error in cell 3, execution on cells 4 and 5 is stopped, but the execution is still happening on the kernel. So when cell 5 is executed, the printed value is something that was never technically executed by the client. (Expected: 1, got 2)
cellstate-before

### After
Server sends "aborted" for cells 4, 5 which cancels the cell execution and re-running the cell 5 shows the expected output keeping client and kernel in sync. 
cellstate-after


changelogs/current_changelog.md
```diff
@@ -166,10 +167,12 @@ Provide a bulleted list of bug fixes and a reference to the PR(s) containing the
 #### Breaking Changes
 
 Provide a bulleted list of breaking changes and a reference to the PR(s) containing those changes.
+- executeRequest uses Jupyter default for stop_on_error = True. ([PR5561](https://github.com/nteract/nteract/pull/5561))
 
 #### New Features
 
 Provide a bulleted list of new features or improvements and a reference to the PR(s) containing these changes.
+- Add support for retreiving executionErrors including aborted requests ([PR5561](https://github.com/nteract/nteract/pull/5561))
```
pwang347: nit: retrieving

packages/messaging/src/index.ts
```diff
@@ -245,6 +245,26 @@ export const executionCounts = () => (source: Observable<JupyterMessage>) =>
...
+ * One of: 'ok' OR 'error' OR 'aborted' from the messaging protocol: https://jupyter-client.readthedocs.io/en/stable/messaging.html#execution-results
+ */
+export const executionStatuses = () => (source: Observable<JupyterMessage>) =>
+source.pipe(
+  ofMessageType("execute_reply"),
+  map(entry => entry.content.status)
+);
+
+/**
+ * Get all execution error from the observable of the jupyter messages
```
pwang347: nit: errors

packages/epics/__tests__/execute/execute.spec.ts
```diff
@@ -124,6 +128,62 @@ describe("executeCellStream", () => {
       )
     );
 
+    expect(emittedActions).toContainEqual(
```
pwang347, line 131: are we not able to assert some order with the actions here?

packages/types/src/errors/execute.ts
```diff
@@ -51,3 +51,7 @@ export const EXEC_ERROR_IN_CELL_STREAM = "EXEC_ERROR_IN_CELL_STREAM";
  * that were sent in the meantime might fail.
  */
 export const EXEC_EPIC_ERROR = "EXEC_EPIC_ERROR";
+/** 
+ * An error returned by the kernel on cell run as an "error" message type.
+ */
+export const EXEC_CELL_RUNTIME_ERROR = "EXEC_CELL_RUNTIME_ERROR"
```
pwang347: nit: newline

packages/epics/__tests__/execute/execute.spec.ts
shibbas: I don't think the order here really matters that much except when it comes to cancelation since it would cancel the stream but that's not handled here. but in createExecuteCellStream

shibbas:
Putting this back into draft to think through using errors message type vs. execute_reply.status for error.

captainsafia:
Heads up that automatic releases are now enabled on the repo (see #5568 for more info). To support automated releases, please make sure that the commits in this PR adhere to the conventional commit format (see https://www.conventionalcommits.org/en/v1.0.0/). Once this PR is merged, a release will be triggered for the modified packages based on the commit message.

MSeal:
@shibbas any input needed to help get the last issue you mentioned putting it back into draft status?

PR #5587, opened 2021-07-24
fix: update editor resize handler to incorporate the new contentsizechange event and adopt padding option

vivek1729:
Monaco 0.20 introduced a [`getContentHeight`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.icodeeditor.html#getcontentheight) API that enables us to get
the current editor height without resorting to hacks around line counts.
This also handles cases such as code folding. See this issue for
details: https://github.com/microsoft/monaco-editor/issues/794

Also added more type checking in the editor instantiation, leveraged the
new [padding](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.ieditorpaddingoptions.html) option to do away with the TopMargin work around.

**Screenshots**:

Before:
monacoResizeNteractBefore

After:

monacoResizeNteractAfter

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -118,42 +118,31 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       if (this.props.onChange) {
         this.props.onChange(this.editor.getValue(), e);
       }
-
-      this.calculateHeight();
     }
   }
 
   /**
-   * Adjust the height of editor
+   * Adjust the height of editor container
    *
-   * @remarks
-   * The way to determine how many lines we should display in editor:
-   * If numberOfLines is not set or set to 0, we adjust the height to fit the content
-   * If numberOfLines is specified we respect that setting
+   * @param height Expected height of the editor container
+   * We check the editor's content height and set the container height to match it
+   *
    */
-  calculateHeight() {
+  calculateHeight(height?: number) {
     // Make sure we have an editor
     if (!this.editor) {
       return;
     }
 
-    // Make sure we have a model
-    const model = this.editor.getModel();
-    if (!model) {
-      return;
+    if (typeof height === "undefined") {
+      // Retrieve content height directly from the editor if no height provided as param
+      height = this.editor.getContentHeight();
     }
-
     if (this.editorContainerRef && this.editorContainerRef.current) {
-      const expectedLines = this.props.numberOfLines || model.getLineCount();
-      // The find & replace menu takes up 2 lines, that is why 2 line is set as the minimum number of lines
-      const finalizedLines = Math.max(expectedLines, 1) + 1;
-      const lineHeight = this.editor.getOption(monaco.editor.EditorOption.lineHeight);
-      const contentHeight = finalizedLines * lineHeight;
-
-      if (this.contentHeight !== contentHeight) {
-        this.editorContainerRef.current.style.height = contentHeight + "px";
+      if (this.contentHeight !== height) {
+        this.editorContainerRef.current.style.height = height + "px";
         this.editor.layout();
-        this.contentHeight = contentHeight;
+        this.contentHeight = height;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -118,42 +118,31 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       if (this.props.onChange) {
```
barryt2: with the removal of this.calculateHeight(), I recommend you to combine the 2 if statements into a single if statement. e.g. if (this.editor && this.props.onChange)

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -118,42 +118,31 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
     if (this.editorContainerRef && this.editorContainerRef.current) {
-      const expectedLines = this.props.numberOfLines || model.getLineCount();
-      // The find & replace menu takes up 2 lines, that is why 2 line is set as the minimum number of lines
-      const finalizedLines = Math.max(expectedLines, 1) + 1;
-      const lineHeight = this.editor.getOption(monaco.editor.EditorOption.lineHeight);
-      const contentHeight = finalizedLines * lineHeight;
-
-      if (this.contentHeight !== contentHeight) {
-        this.editorContainerRef.current.style.height = contentHeight + "px";
+      if (this.contentHeight !== height) {
```
barryt2: I suggest you combine this if statement with the one above since there is no other logic needed after it enters the if statement block above. e.g. if (this.editorContainerRef && this.editorContainerRef.current && this.contentHeight !== height)

barryt2 (approved):

packages/monaco-editor/src/MonacoEditor.tsx
vivek1729: Done.

packages/monaco-editor/src/MonacoEditor.tsx
vivek1729: Done.

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -229,8 +214,6 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         this.props.onDidCreateEditor(this.editor);
       }
 
-      this.addEditorTopMargin();
```
kycutler, line 232: Since the padding was previously added in this class, should we add the default `padding` option here too to maintain the default sizing?

vivek1729: Yeah, that's a good point. I added the padding option when we instantiate the editor here but we could provide a default padding of 10px out of the box as well. Since we were adding a 10px top margin, may be we could just defaul to a 10px top padding. What do you think?

kycutler: You mean 12px, yeah? And yep that sounds good, plus a bottom padding of 5px since the height calculation added 1 line of height so we also had (1*17px - 12px = 5px) of extra height at the bottom.

vivek1729: Yep, makes sense. I've added the default padding so that folks relying on the monaco-editor package don't have to provide any other input options to have the same experience. Ofcourse, we support overriding all the editor options via props.

vivek1729:
@captainsafia , @willingc please let me know if you have any additional feedback. Thanks!

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -114,47 +114,32 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
-      const contentHeight = finalizedLines * lineHeight;
-
-      if (this.contentHeight !== contentHeight) {
-        this.editorContainerRef.current.style.height = contentHeight + "px";
-        this.editor.layout();
-        this.contentHeight = contentHeight;
-      }
+    if (this.editorContainerRef && this.editorContainerRef.current && (this.contentHeight !== height)) {
+      this.editorContainerRef.current.style.height = height + "px";
+      this.editor.layout({ width: this.editor.getLayoutInfo().width, height });
```
kycutler, line 146: When the content is wider than the container, does this cause overflow / disable scrolling? Thinking it might still be better to let Monaco determine the size based on the parent container just in case.

kycutler (approved):

MSeal (approved):

Issue #5590, opened 2021-07-30
Monaco editor should take into consideration that the value prop changed before calling setValue


barryt2:
**Application or Package Used**
nteract monaco-editor package.

**Describe the bug**
The Monaco editor should take into consideration that the value prop has changed before calling setValue with a new value. 

Current code in componentDidUpdate:

```js
// Ensures that the source contents of the editor (value) is consistent with the state of the editor
if (this.editor.getValue() !== this.props.value) {
  this.editor.setValue(this.props.value);
}
```

The use case for taking in consideration of "the value prop has changed" is to allow folks who may want to do some performance/scalability optimizations for large notebook scenarios is to cache the typing text externally outside of the editor and later committing the cached text to the redux store versus committing each typed character immediately to the redux store. Each redux store change per character typed could be expensive to re-render. The issue without taking "the value prop has changed" into consideration is that the current code above would cause the Monaco editor to re-render the using the old value prop that we know is not updated in the redux store to reflect what has been typed into the Monaco editor. So based on the current condition statement above, it will see that the Monaco editor text (up-to-date text value) is different from value prop (original text value), so it will call setValue with the old text value reverting what the user has typed. In the new proposal, we will detect that the value has actually changed and the value is different between the prop versus the text shown in the Monaco editor before calling setValue.

New code in componentDidUpdate:

```js
// Ensures that the source contents of the editor (value) is consistent with the state of the editor
// and the value has actually changed.
if (this.editor.getValue() !== this.props.value && prevProps.value !== this.props.value) {
  this.editor.setValue(this.props.value);
}
```

**Screenshots**
N/A

**Desktop (please complete the following information):**

- Os: Windows
- Browser Chrome
- Version 92

**Additional context**
N/A

PR #5591, opened 2021-07-30
fix: consider value prop changed before calling setvalue on monaco editor

barryt2:
Updated Monaco editor to take into consideration that the value prop has changed before calling setValue with a new value. 

Fixes issue: #5590 

Old code in componentDidUpdate:

```js
// Ensures that the source contents of the editor (value) is consistent with the state of the editor
if (this.editor.getValue() !== this.props.value) {
  this.editor.setValue(this.props.value);
}
```

The use case for taking in consideration of "the value prop has changed" is to allow folks who may want to do some performance/scalability optimizations for large notebook scenarios is to cache the typing text externally outside of the editor and later committing the cached text to the redux store versus committing each typed character immediately to the redux store. Each redux store change per character typed could be expensive to re-render. The issue without taking "the value prop has changed" into consideration is that the current code above would cause the Monaco editor to re-render the using the old value prop that we know is not updated in the redux store to reflect what has been typed into the Monaco editor. So based on the current condition statement above, it will see that the Monaco editor text (up-to-date text value) is different from value prop (original text value), so it will call setValue with the old text value reverting what the user has typed. In the new proposal, we will detect that the value has actually changed and the value is different between the prop versus the text shown in the Monaco editor before calling setValue.

New code in componentDidUpdate:

```js
// Ensures that the source contents of the editor (value) is consistent with the state of the editor
// and the value has actually changed.
if (this.editor.getValue() !== this.props.value && prevProps.value !== this.props.value) {
  this.editor.setValue(this.props.value);
}
```

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -317,7 +317,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
     }
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: IMonacoProps) {
     if (!this.editor) {
       return;
     }
@@ -334,7 +334,8 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
     }
 
     // Ensures that the source contents of the editor (value) is consistent with the state of the editor
-    if (this.editor.getValue() !== this.props.value) {
+    // and the value has actually changed.
+    if (this.editor.getValue() !== this.props.value && prevProps.value !== this.props.value) {
       this.editor.setValue(this.props.value);
     }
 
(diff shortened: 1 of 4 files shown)


vivek1729 (approved):

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -315,7 +315,8 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
     }
 
     // Ensures that the source contents of the editor (value) is consistent with the state of the editor
-    if (this.editor.getValue() !== this.props.value) {
+    // and the value has actually changed.
+    if (this.editor.getValue() !== this.props.value && prevProps.value !== this.props.value) {
```
kycutler: nit: since getValue() [does work to normalize line endings](https://github.com/microsoft/vscode/blob/4f113276b635f3443efc1d3d912e407951575ef1/src/vs/editor/common/model/textModel.ts#L758), it would be slightly more efficient to reverse these conditions and take advantage of short-circuiting in the case where the prop is unchanged.

kycutler (approved):

packages/monaco-editor/src/MonacoEditor.tsx
barryt2: good point. swapped the order.

vivek1729 (approved):

PR #5598, opened 2021-09-02
fix: batch dom writings to avoid layout thrashing

bryanchen-d:

Monaco editor's layout() is a major source that cause reflows, which causes UI responsiveness issue. By isolate DOM operations and batch the readings in one group and writings in another group can significantly reduce the times of reflows and improve the performance.


bryanchen-d:
the optimization is done with a different PR.

PR #5601, opened 2021-09-22
fix: bump react-redux to 7.2.5

captainsafia:
Closes #5588

PR #5608, opened 2021-10-05
fix: mark react-redux package as a peer dependency

vivek1729:

A recent [Pull request ](https://github.com/nteract/nteract/pull/5601) bumped up the mono repo to leverage `react-redux` v7 😃 . This is awesome as it sets us up to take several advantages of new apis. In that change, I noticed that `react-redux` was marked as a peer dependency as well as a regular dependency for some nested packages. Specifically:
- myths
- notebook-app-component
- connected-components
- stateful-components

While, it makes sense to manage `react-redux` as a regular dependency at the top level or for the individual applications, we'd want to ship nested nteract packages like myths etc to claim this package as a peer dependency only. This allows the users to manage a single copy of these core modules (like they already do for `react` right now).

### DIFF
--- a/packages/connected-components/package.json
+++ b/packages/connected-components/package.json
@@ -16,7 +16,6 @@
     "@nteract/commutable": "^7.4.5",
     "@nteract/core": "^15.1.9",
     "@nteract/types": "^7.1.9",
-    "react-redux": "7.2.5",
     "redux": "^4.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


barryt2 (approved):

MSeal (commented):
I'd consider also moving redux to the peerDependencies section. Thoughts @captainsafia ?

captainsafia (approved):

PR #5613, opened 2021-10-22
Only show the last input prompt

### LINKED ISSUE #5592: Input prompts don't disappear after entry
nteract desktop

**Describe the bug**
Input prompts do not disappear after a reply is sent. Thus if there are multiple prompts (for example with multiple `input()` calls, or when using the pdb debugger), prompts accumulate endlessly. Also the input text is synchronized across all prompts, and remembered after clearing the cell outputs.

**To Reproduce**

1. Connect to a python kernel
2. Add a code cell with multiple `input()` calls and execute it
3. Enter a value into each prompt and click "Submit"
4. Notice that the input prompts do not disappear. In addition, the input text is synchronized across all prompts.

**Expected behavior**
When clicking "Submit" on an input prompt, the input box should disappear. Jupyter and JupyterLab follow this behavior, and also append the prompt text and reply to stdout.

**Screenshots**
nteractprompts

- OS: Windows 10
- Version 0.28.0

CrystallineCat:
Fixes #5592.

Changed the input prompts to only show the last one.

NB: I'm not sure this is the correct way to fix this. Maybe we should store the value of each and turn them read-only once they have been submitted?

### DIFF
--- a/packages/stateful-components/__tests__/outputs/input-prompts.spec.tsx
+++ b/packages/stateful-components/__tests__/outputs/input-prompts.spec.tsx
@@ -1,6 +1,6 @@
 import { mount } from "enzyme";
-import Immutable from "immutable";
-import React from "react";
+import * as Immutable from "immutable";
+import * as React from "react";
 
 import { PromptRequest } from "../../src/outputs/input-prompts";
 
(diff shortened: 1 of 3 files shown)


captainsafia (approved):
LGTM except for fixing the commit message to support automated releases. Manually validated with following code.

```python
a = input("A: ")
b = input("B: ")
c = input("C: ")
```

```python
a, b, c
```

CrystallineCat:
Commit message should be good now. Thank you!

kycutler:
I'm curious about this fix; why do we need to keep around old prompts in the state if we only show the last one in the UI?

PR #5656, opened 2022-03-11
fix: import sub modules instead of the entire lodash to reduce build size

bryanchen-d:
Change importing the entire lodash package to just importing its sub modules to reduce the build size.


captainsafia (approved):
We had done this work elsewhere in the repo but it looks like there was a regression with the newer stuff. Maybe there is a lint rule we can add for this...

PR #5692, opened 2022-06-27
fix: capture both text and editor widget states on focus change

vivek1729:
Monaco editor has specific APIs - [`onDidFocusEditorWidget`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.IStandaloneCodeEditor.html#onDidFocusEditorWidget), [`onDidBlurEditorWidget`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.IStandaloneCodeEditor.html#onDidBlurEditorWidget),  [`hasWidgetFocus`](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.IStandaloneCodeEditor.html#hasWidgetFocus) which allow us to subscribe to and query focus states for both the editor text area i.e. the cursor as well as the variety of editor widgets that monaco editor supports.

We adopt these APIs for completeness. Otherwise, host applications providing focus change handlers would miss events when an editor widget eg: context menu or the command bar gain/lose focus.

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -237,7 +237,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       this.toggleEditorOptions(!!this.props.editorFocused);
 
       if (this.props.editorFocused) {
-        if (!this.editor.hasTextFocus()) {
+        if (!this.editor.hasWidgetFocus()) {
           // Bring browser focus to the editor if text not already in focus
           this.editor.focus();
         }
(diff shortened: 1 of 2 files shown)


barryt2 (approved):

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -275,11 +275,6 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         this.props.cursorPositionHandler(this.editor, this.props);
       }
 
-      // When editor loses focus, hide parameter widgets (if any currently displayed).
-      this.blurEditorWidgetListener = this.editor.onDidBlurEditorWidget(() => {
```
kycutler, line 279: Nit: looks like this variable can be removed as well

kycutler (approved):

packages/monaco-editor/src/MonacoEditor.tsx
vivek1729: Good catch!

PR #5694, opened 2022-07-05
feat: add support for Markdown cell attachments

christopher-o-toole:
The Jupyter notebook file specification states that Markdown and raw cells have an optional field for [cell attachments](https://nbformat.readthedocs.io/en/latest/format_description.html#cell-attachments), but this field is not supported by `@nteract/commutable`. By supporting attachments for markdown cells, it will be possible to embed images within the notebook that can then be referenced by Markdown cells. An added bonus is better compatibility with JupyterLab notebook files as any images that are pasted or dropped onto a markdown cell are stored as cell attachments (see [here](https://github.com/jupyterlab/jupyterlab/blob/cb117e3c0767f3fcfd734f3cc84b058b22b414c9/packages/cells/src/widget.ts#L1420) for the JupyterLab implementation)

### DIFF
--- a/packages/commutable/__tests__/v4.spec.ts
+++ b/packages/commutable/__tests__/v4.spec.ts
@@ -165,6 +165,27 @@ describe("cell ids", () => {
         initial
       );
     });
+    it("can read notebook containing a markdown cell with attachments", () => {
+      const sampleMarkdownCell = createMarkdownCell({
+        attachments: Immutable.Map({
+          "foo.gif": {
+            "image/gif": ["<mulitline-base64", "-string>"]
+          }
+        })
+      }).toJS();
+      const notebook = getNotebook({
+        cells: [
+          {
+            ...sampleMarkdownCell,
+            id: "markdown-test-cell"
+          }
+        ]
+      });
+      const immNotebook = fromJS(notebook);
+      const markdownCell = immNotebook.getIn(["cellMap", "markdown-test-cell"]).toJS();
+
+      expect(markdownCell).toHaveProperty(["attachments", "foo.gif", "image/gif"], "<mulitline-base64-string>");
+    });
   });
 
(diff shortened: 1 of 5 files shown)


packages/commutable/src/v4.ts
```diff
@@ -70,6 +71,7 @@ export interface CodeCell {
 }
 
 export interface MarkdownCell {
+  attachments?: { [mime_type: string]: MultiLineString }
```
vivek1729: Looking at the [nbformat](https://nbformat.readthedocs.io/en/latest/format_description.html#cell-attachments), it is not clear to me how the `filepath` is captured in this type.  Shouldn't filepaths be the top level keys for `attachments`?

packages/commutable/__tests__/v4.spec.ts
```diff
@@ -41,6 +41,19 @@ describe("cellToJS", () => {
     expect(cellToJS(codeCell).cell_type).toBe("code");
     expect(cellToJS(markdownCell).cell_type).toBe("markdown");
   });
+  it("can create and convert a markdown cell with attachments", () => {
```
vivek1729, line 44: For completeness, should we also consider adding an example with a multi line string?

packages/commutable/src/v4.ts
christopher-o-toole: I had corrected the immutable type without correcting the JS type, my bad.

packages/commutable/__tests__/v4.spec.ts
christopher-o-toole: Added a test to ensure that a notebook containing a markdown cell with attachments can be parsed to an immutable and converted back to a JS type successfully.

vivek1729 (approved):

kycutler (approved):

PR #5695, opened 2022-07-06
fix: add maxHeight and autoFitContentHeight options

bryanchen-d:
Add two options in monaco-editor component
* autoFixContentHeight - this is already on by default, adding this option to allow users to disable it by setting it to false
* maxHeight - allow users to set a max height of the editor

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -71,8 +71,8 @@ export interface IMonacoConfiguration {
   lineNumbers?: boolean;
   /** automatically adjust size to fit content, default is true */
   autoFitContentHeight?: boolean;
-  /** set a max height in number of pixels */
-  maxHeight?: number;
+  /** set a max content height in number of pixels, this only works when autoFitContentHeight is true*/
+  maxContentHeight?: number;
   /** set height of editor to fit the specified number of lines in display */
   numberOfLines?: number;
   indentSize?: number;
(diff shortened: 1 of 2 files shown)


packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -131,10 +135,20 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
+    const shouldProceed = this.props.autoFitContentHeight ?? true;
+    if(!shouldProceed) {
+      return;
+    }
+
     if (typeof height === "undefined") {
       // Retrieve content height directly from the editor if no height provided as param
       height = this.editor.getContentHeight();
+
+      if(this.props.maxHeight) {
```
shibbas: should this be outside the height undefined check since this is a max height? since you don't want content Height to override it? I think when the content changes, calculate height is called again with contentHeight

shibbas (approved):

packages/monaco-editor/src/MonacoEditor.tsx
bryanchen-d: good point, let me update that.

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -131,10 +135,20 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
       return;
     }
 
+    const shouldProceed = this.props.autoFitContentHeight ?? true;
```
shibbas, line 138: what is the relation of maxHeight to autoFitContentHeight? If there is a max height, should that still be respected since it's unrelated to content height changes?

bryanchen-d: updated maxHeight to maxContentHeight and also the comment header for this option is only respected when autoFitContentHeight is true.

vivek1729 (approved):

PR #5705, opened 2022-08-26
BREAKING CHANGE: Monaco version upgrade to 0.33.0

dileepyavan:

 [Monaco editor change log for 0.33.0 version](https://github.com/microsoft/monaco-editor/blob/main/CHANGELOG.md#0330)

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -215,7 +215,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         fixedOverflowWidgets: true,
         find: {
           addExtraSpaceOnTop: false, // pops the editor out of alignment if turned on
-          seedSearchStringFromSelection: "always", // default is true
+          seedSearchStringFromSelection: "always", // default is "always"
           autoFindInSelection: "never" // default is "never"
         },
         language: this.props.language,
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -215,7 +215,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         fixedOverflowWidgets: true,
         find: {
           addExtraSpaceOnTop: false, // pops the editor out of alignment if turned on
-          seedSearchStringFromSelection: true, // default is true
+          seedSearchStringFromSelection: "always", // default is true
           autoFindInSelection: "never" // default is "never"
         },
         language: this.props.language,
@@ -356,7 +356,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
 
     // Apply new model to the editor when the language is changed.
     const model = this.editor.getModel();
-    if (model && language && model.getModeId() !== language) {
+    if (model && language && model.getLanguageId() !== language) {
       // Get a reference to the current editor
       const editor = this.editor;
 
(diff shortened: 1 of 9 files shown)


barryt2 (approved):

vivek1729:
@dileepyavan I just noticed that nteract also uses  monaco-editor in their jupyter extension application. Would you be able to update the webpack config for that project as well? Here's the [package.json](https://github.com/nteract/nteract/blob/main/applications/jupyter-extension/nteract_on_jupyter/package.json#L35) and [webpack config](https://github.com/nteract/nteract/blob/main/applications/jupyter-extension/nteract_on_jupyter/webpack.config.js#L12) for that project and I found it referencing an older version for the monaco-editor webpack plugin which won't be compatible with v 0.33.x

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -215,7 +215,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         fixedOverflowWidgets: true,
         find: {
           addExtraSpaceOnTop: false, // pops the editor out of alignment if turned on
-          seedSearchStringFromSelection: true, // default is true
+          seedSearchStringFromSelection: "always", // default is true
```
captainsafia:
```suggestion
          seedSearchStringFromSelection: "always", // default is "always"
```

packages/monaco-editor/src/theme.ts
```diff
@@ -34,7 +34,7 @@ export const customMonacoDarkTheme: monaco.editor.IStandaloneThemeData = {
   inherit: true,
   rules: [],
   colors: {
-    "editor.background": "#111"
+    "editor.background": "#111111"
```
captainsafia, line 37: Interesting -- is there a particular reason we can't leverage the hex code shorthand anymore?

captainsafia (commented):
Is there a particular bug fix/feature we're looking to get from this major upgrade?

~Also, to capture the breaking change for conventional commits correctly, it'll have to be communicated via a `!` in the message. The "BREAKING CHANGE" is only applicable in the commit body AFAIK ([ref](https://www.conventionalcommits.org/en/v1.0.0/)).~

It looks like some commits have the right naming. I think that should be enough.

packages/monaco-editor/src/theme.ts
dileepyavan: Monaco editor came up with new validation which checks for the length of the code and if its less than 6, throws invalid color code error.

dileepyavan:
> 



> @dileepyavan I just noticed that nteract also uses monaco-editor in their jupyter extension application. Would you be able to update the webpack config for that project as well? Here's the [package.json](https://github.com/nteract/nteract/blob/main/applications/jupyter-extension/nteract_on_jupyter/package.json#L35) and [webpack config](https://github.com/nteract/nteract/blob/main/applications/jupyter-extension/nteract_on_jupyter/webpack.config.js#L12) for that project and I found it referencing an older version for the monaco-editor webpack plugin which won't be compatible with v 0.33.x

@vivek1729  Thanks for the review, I have updated the PR with mentioned changes.

dileepyavan:
> Is there a particular bug fix/feature we're looking to get from this major upgrade?
> 
> ~Also, to capture the breaking change for conventional commits correctly, it'll have to be communicated via a `!` in the message. The "BREAKING CHANGE" is only applicable in the commit body AFAIK ([ref](https://www.conventionalcommits.org/en/v1.0.0/)).~
> 
> It looks like some commits have the right naming. I think that should be enough.

@captainsafia We requested monaco editor for a change regarding streaming hover results in case of multiple hover providers. This was added in https://github.com/microsoft/monaco-editor/blob/main/CHANGELOG.md#0310-10122021.

vivek1729 (approved):
Thanks Dileep! Your changes LGTM.

vivek1729:
@captainsafia , I am merging the PR as the changes look complete to me. @dileepyavan can address any additional feedback you might have as a follow up PR.

PR #5707, opened 2022-09-15
fix: add skipLayoutWhenHidden prop

bryanchen-d:
Added a skipLayoutWhenHidden prop in monaco-editor to skip calling editor.layout() in cases when the monaco-editor is hidden by its parent or ancestor container.


vivek1729 (commented):
Do you think we should update the `calculateHeight` method to include this check? It's called when the content height changes or during the `componentUpdate` lifecycle of the editor component.

bryanchen-d:
> Do you think we should update the `calculateHeight` method to include this check? It's called when the content height changes or during the `componentUpdate` lifecycle of the editor component.

Hmm, thanks for reminding me that, maybe a better place is to put the check under `updateEditorLayout`.

vivek1729 (approved):

kycutler (approved):

PR #5731, opened 2023-02-10
fix: optimize resizing perf by using resize observer and batch the DOM reads and writes separately

bryanchen-d:
This PR fixed several issues with resizing perf
* Use ResizeObserver to detect resize changes. Previously it was using window resize event which is not enough to trigger the layout of monaco editors, there are other cases like container div gets resized, or changed from hidden to visible. 
* Use editor.layout to control the height instead of setting height on the container, which simplify the write operations on DOM.
* Add skipLayoutWhenNotInViewport property, when set to true, we only do the layout when the editor is in viewport.

### DIFF
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -135,13 +135,14 @@ export default class MonacoEditor extends React.Component<IMonacoProps> implemen
     }
   }
 
-  getContainerDimension() {
-    const container = this.editorContainerRef.current;
-    if (!container || !container.offsetHeight || !container.offsetWidth) {
+  getLayoutDimension() {
+    const container = this.editor?.getContainerDomNode();
+    if (!container) {
       return undefined;
     }
 
-    return { width: container.offsetWidth, height: container.offsetHeight };
+    // use clientHeight and clientWidth to align with what monaco editor does for getting the dimension when it is undefined.
+    return { width: container.clientWidth, height: container.clientHeight };
   }
 
   isContainerHidden() {
--- a/packages/monaco-editor/src/MonacoEditor.tsx
+++ b/packages/monaco-editor/src/MonacoEditor.tsx
@@ -121,7 +121,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
     this.onBlur = this.onBlur.bind(this);
     this.onDidChangeModelContent = this.onDidChangeModelContent.bind(this);
     this.onFocus = this.onFocus.bind(this);
-    this.resize = this.resize.bind(this);
+    this.onResize = this.onResize.bind(this);
     this.hideAllOtherParameterWidgets = this.hideAllOtherParameterWidgets.bind(this);
     this.handleCoordsOutsideWidgetActiveRegion = debounce(
       this.handleCoordsOutsideWidgetActiveRegion.bind(this),
@@ -136,10 +136,18 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
   }
 
   isContainerHidden() {
-    return !this.editorContainerRef.current?.offsetHeight;
+    const container = this.editorContainerRef.current;
+    return !container?.offsetParent || !container?.offsetHeight;
   }
 
-  updateEditorLayout(layout?: monaco.editor.IDimension) {
+  /**
+   * call into monaco editor to layout the editor
+   */
+  layout(layout?: monaco.editor.IDimension) {
+    this.editor?.layout(layout);
+  }
+
+  requestLayout(layout?: monaco.editor.IDimension) {
     if (!this.editor) {
       return;
     }
@@ -151,9 +159,9 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
     }
 
     if (this.props.batchLayoutChanges === true) {
-      scheduleEditorForLayout(this.editor, layout);
+      scheduleEditorForLayout(this, layout);
     } else {
-      this.editor.layout(layout);
+      this.layout(layout);
     }
   }
 
@@ -191,7 +199,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
        * This causes a forced layout by the browser
        * We pass in the expected width and height to as an optimization to avoid the forced layout
        */
-      this.updateEditorLayout({ width: this.editor.getLayoutInfo().width, height });
+      this.requestLayout({ width: this.editor.getLayoutInfo().width, height });
       this.contentHeight = height;
     }
   }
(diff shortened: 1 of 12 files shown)


barryt2 (approved):

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -134,25 +135,40 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
     if (!this.editor) {
       return;
     }
 
-    // when skipLayoutWhenHidden is true and the editor's parent or ancestor container is hidden, 
-    // we will not layout the editor. 
-    if(this.props.skipLayoutWhenHidden && this.isContainerHidden()) {
+    // when skipLayoutWhenHidden is true and the editor's parent or ancestor container is hidden,
+    // we will not layout the editor.
+    if (this.props.skipLayoutWhenHidden && this.isContainerHidden()) {
```
kycutler: Nit: reuse `shouldLayout()` here

packages/monaco-editor/src/layoutSchedule.ts
```diff
@@ -1,24 +1,38 @@
...
+  layoutTimer = null;
+  const editorsToLayout = [];
+  for (const [editor, layout] of editorsInSchedule) {
+    if (editor.shouldLayout()) {
+      editorsToLayout.push([editor, layout]);
+    }
+  }
+
+  for (const [editor, layout] of editorsToLayout) {
+    (editor as IEditor).layout(layout as monaco.editor.IDimension | undefined);
```
kycutler: Curious why the two loops here? Can we just do the `layout()` call in the first loop and remove `editorsToLayout`?

packages/monaco-editor/src/layoutSchedule.ts
```diff
@@ -1,24 +1,38 @@
...
+}
+
 /**
  * For monaco editors, we need to call layout() on any editors that might have changed size otherwise the view will look off.
  * These updates often happen together with other editors, such as when the window resizes.
  * In order to avoid layout thrashing, we batch these layout calls together and perform them all at once in a RAF timeout.
  */
-export function scheduleEditorForLayout(editor: monaco.editor.IEditor, layout: monaco.editor.IDimension | undefined) {
-  editorsToLayout.set(editor, layout);
+export function scheduleEditorForLayout(editor: IEditor, layout?: monaco.editor.IDimension) {
```
kycutler: Making this optional defeats the purpose of the batching, as it results in the editor layout being calculated after the timeout, forcing a synchronous layout and possible layout thrashing if multiple editors are resizing.

packages/monaco-editor/src/resizeObserver.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Interface for resizable components.
+ */
+export interface IResizable {
+  onResize(): void;
+}
+
+/**
+ * ResizeObserver that monitors the size of the element and calls the resizable component.
+ */
+const monitoredResizables = new Map<Element, IResizable>();
```
kycutler: Nit: this would be a great place to use a [`WeakMap`](https://caniuse.com/mdn-javascript_builtins_weakmap)

packages/monaco-editor/src/layoutSchedule.ts
kycutler: Oh wait I see it was also optional before -- nevertheless can you check if the layout thrashing has regressed? We may want to e.g. calculate the layout here if it is not passed in

packages/monaco-editor/src/layoutSchedule.ts
bryanchen-d: The first loop would give all editors a chance to check if they are hidden just in case there maybe a force synchronization.
Then the second loop is actually the one to batch the layout operations.

packages/monaco-editor/src/layoutSchedule.ts
bryanchen-d: Updated to batch the reads of DOM before calling the actual layout method of editors.

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -103,67 +118,120 @@ export interface IMonacoConfiguration {
...
+
+  requestLayout(layout?: monaco.editor.IDimension): void {
     if (!this.editor) {
       return;
     }
 
-    // when skipLayoutWhenHidden is true and the editor's parent or ancestor container is hidden, 
-    // we will not layout the editor. 
-    if(this.props.skipLayoutWhenHidden && this.isContainerHidden()) {
+    if (!this.isInViewport) {
```
kycutler, line 226: Should we check `this.shouldLayout()` before this? Otherwise we could cache a zero-size dimension if the editor is hidden?

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -174,29 +242,71 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
...
       height = Math.min(height, this.props.maxContentHeight);
     }
-    
-    if (this.editorContainerRef && this.editorContainerRef.current && this.contentHeight !== height) {
-      this.editorContainerRef.current.style.height = height + "px";
+
+    const containerElement = this.editorContainerRef.current;
+    if (containerElement && this.containerHeight !== height) {
+      this.containerHeight = height;
+      containerElement.style.height = height + "px";
```
kycutler: I'm starting to think it makes more sense to just do this in `layout()` -- i.e. make `layout()` require the dimension parameter, and then also set the style on the container element (if `autoFitContentHeight` is true). We could even apply the max height there as well. Then we can schedule everything through `requestLayout` and really ensure that all of the DOM writes happen together (whereas here it is possible that this write triggers a sync layout if another editor is in the batch as well).

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -262,6 +372,7 @@ export default class MonacoEditor extends React.Component<IMonacoProps> {
         },
         theme: this.props.theme,
         value: this.props.value,
+        dimension: this.props.initialDimension,
```
kycutler, line 333: Should we skip the `calculateHeight()` call below if this is set?

kycutler (approved):

packages/monaco-editor/src/MonacoEditor.tsx
bryanchen-d: Updated to call `shouldLayout()` before calling the layout

packages/monaco-editor/src/MonacoEditor.tsx
bryanchen-d: Agree, I updated the logic removing the height style on the container.

packages/monaco-editor/src/MonacoEditor.tsx
```diff
@@ -106,97 +121,146 @@ export type IMonacoProps = IMonacoComponentProps & IMonacoConfiguration;
...
-       * With no params, the layout method queries the DOM to get the parent container dimensions
-       * This causes a forced layout by the browser
-       * We pass in the expected width and height to as an optimization to avoid the forced layout
-       */
-      this.updateEditorLayout({ width: this.editor.getLayoutInfo().width, height });
-      this.contentHeight = height;
+    if (this.props.batchLayoutChanges === true) {
+      scheduleEditorForLayout(this, layout);
+    } else {
+      const layout = this.getLayoutDimension();
```
kycutler: Nit: we can still avoid the DOM read when coming from `onDidContentSizeChange` by preferring the `layout` parameter here

bryanchen-d: Ah yes, let me update that.

PR #5738, opened 2023-03-25
fix: contextualize cell and line magic for autocompletion

barryt2:
Make a fix to contextualize cell magic and line magic suggested by auto-completion. Currently, auto-completion suggests cell and line magics in cases when it should not. And if the user commits the suggestion, it leads to cell that will fail to execute successfully because it is invalid.

Take for example the following cell code:
```js
print("hello");
%h
```

When the "h" is typed, autocompletion will suggest %%html cell magic. In this case, it should not have been suggested. 

Cell magics should only be suggested if everything prior to the cell magic is whitespace and the cell magic starts on the first column of the line.

Line magics should only be suggested if the line magic starts on the first column of any given line.

### DIFF
--- a/packages/monaco-editor/src/completions/completionItemProvider.ts
+++ b/packages/monaco-editor/src/completions/completionItemProvider.ts
@@ -134,7 +134,6 @@ class CompletionItemProvider implements monaco.languages.CompletionItemProvider 
     }
 
     const isWhitespaceFromCellStart = this.isWhitespaceFromCellStart(model, startPos);
-    const isWhitespaceFromLineStart = this.isWhitespaceFromLineStart(model, startPos);
 
     return completionItems
       .map((completionItem: CompletionMatch, index: number) => {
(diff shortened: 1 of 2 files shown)


packages/monaco-editor/src/completions/completionItemProvider.ts
```diff
@@ -126,193 +101,212 @@ class CompletionItemProvider
...
+  private isCellMagic(text: string) {
+    return text.startsWith("%%");
+  }
+
+  /**
+   * Whether text is a line magic.
+   */
+  private isLineMagic(text: string) {
+    return text.startsWith("%") && !this.isCellMagic(text);
   }
```
bryanchen-d: I can see the usage of these two methods are 
```typescript
if(isCellMagic(text)){
   ...
} else if(isLineMagic(text)){
  ...
}
```
In which case isCellMagic most of time get called twice, then better we optimize it since it is going to be called very frequently, I am thinking something like
```typescript
private getMagicPrefix(text){
    if(text.startsWith("%")){
       // so we can skip one 'startsWith" call most of then time
        return text.startsWith("%%")? "%%":"%"; // return string or enum should be equivalent.
    }
    else{
        return null;
    }
}


....
// later when we try to remove the percentages characters:
if(magicPrefix) { completionText=completionText.replace(magicPrefix, "")}
```

packages/monaco-editor/src/completions/completionItemProvider.ts
```diff
@@ -27,19 +17,18 @@ const jupyterToMonacoCompletionItemKind: {
...
 };
 
 /**
  * Completion item provider.
  */
-class CompletionItemProvider
-  implements monaco.languages.CompletionItemProvider {
+class CompletionItemProvider implements monaco.languages.CompletionItemProvider {
   private channels: Channels | undefined;
+  private regexWhitespace = new RegExp(/^\s*$/);
```
bryanchen-d, line 28: Does this regex match multiple lines of whitespaces?

bryanchen-d (approved):

packages/monaco-editor/src/completions/completionItemProvider.ts
barryt2: That makes sense, updated.

PR #5747, opened 2023-07-15
fix: normalize cell source line endings to LF

barryt2:
Make a fix in Nteract to normalize line endings to LF (\n) for all cell sources. This is to align with and be more compatible with other notebook solutions. In 2021, Jupyter Lab, VS Code, and Google Colab notebooks had made similar updates to normalize line endings to LF.

- Jupyter Lab: [reference](https://github.com/jupyterlab/jupyterlab/pull/11141)
- VS Code Notebook: [reference](https://github.com/microsoft/vscode/pull/133762)
- Google Colab: [reference](https://github.com/jupyterlab/jupyterlab/issues/9465#issuecomment-802215114)

CodeMirror editor also already normalizes its content to use LF line endings per its [docs](https://codemirror.net/5/doc/manual.html). CodeMirror is one of the two editors supported by Nteract.  Monaco is the other editor Nteract supports and this PR has updates to configure Monaco to also normalize line ending to LF.

The Nteract packages that will contain the normalization updates are the following:
 
@nteract/commutable
- On cell model creators

@nteract/reducers
- On cell source content updates
 
@nteract/monaco-editor
- On Monaco editor creation

### DIFF
--- a/packages/commutable/src/structures.ts
+++ b/packages/commutable/src/structures.ts
@@ -1,6 +1,6 @@
 import { v4 as uuid } from "uuid";
 
-import { CellId, createCellId } from "./primitives";
+import { CellId, createCellId, normalizeLineEndings } from "./primitives";
 
 import { ImmutableCell, makeCodeCell, makeMarkdownCell } from "./cells";
 
@@ -64,7 +64,7 @@ export function appendCell(
 ): CellStructure {
   return {
     cellOrder: cellStructure.cellOrder.push(id),
-    cellMap: cellStructure.cellMap.set(id, immutableCell)
+    cellMap: cellStructure.cellMap.set(id, immutableCell.setIn(["source"], normalizeLineEndings(immutableCell.source)))
   };
 }
 
@@ -108,7 +108,7 @@ export function insertCellAt(
 ): ImmutableNotebook {
   return notebook.withMutations(nb =>
     nb
-      .setIn(["cellMap", cellId], cell)
+      .setIn(["cellMap", cellId], cell.setIn(["source"], normalizeLineEndings(cell.source)))
       .set("cellOrder", nb.get("cellOrder").insert(index, cellId))
   );
 }
--- a/packages/commutable/src/structures.ts
+++ b/packages/commutable/src/structures.ts
@@ -1,6 +1,6 @@
 import { v4 as uuid } from "uuid";
 
-import { CellId, createCellId, normalizeLineEndings } from "./primitives";
+import { CellId, createCellId } from "./primitives";
 
 import { ImmutableCell, makeCodeCell, makeMarkdownCell } from "./cells";
 
@@ -64,7 +64,7 @@ export function appendCell(
 ): CellStructure {
   return {
     cellOrder: cellStructure.cellOrder.push(id),
-    cellMap: cellStructure.cellMap.set(id, immutableCell.setIn(["source"], normalizeLineEndings(immutableCell.source)))
+    cellMap: cellStructure.cellMap.set(id, immutableCell)
   };
 }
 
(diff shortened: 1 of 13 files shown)


packages/commutable/src/structures.ts
```diff
@@ -108,7 +108,7 @@ export function insertCellAt(
 ): ImmutableNotebook {
   return notebook.withMutations(nb =>
     nb
-      .setIn(["cellMap", cellId], cell)
+      .setIn(["cellMap", cellId], cell.setIn(["source"], normalizeLineEndings(cell.source)))
```
rgbkrk: 👍

rgbkrk (approved):
This isn't changing the interface and CodeMirror was already doing this automatically so I say go forth.

packages/commutable/src/structures.ts
```diff
@@ -64,7 +64,7 @@ export function appendCell(
 ): CellStructure {
   return {
     cellOrder: cellStructure.cellOrder.push(id),
-    cellMap: cellStructure.cellMap.set(id, immutableCell)
+    cellMap: cellStructure.cellMap.set(id, immutableCell.setIn(["source"], normalizeLineEndings(immutableCell.source)))
```
bryanchen-d: This will generate a different `ImmutableCell` object in the cellMap, I wonder if consumers of `appendCell` may assume the input `immutableCell` to be present in the cellMap and this change could break that.
Same as in the `insertCellAt` function.

Instead of doing the normalization here, can we do the normalization in the creator of ImmutableCell? Like makeCodeCell, etc.

bryanchen-d: I just noticed that you already do the normalization below in the creators, in that case, these two changes may not be needed?

packages/reducers/src/core/entities/contents/notebook.ts
```diff
@@ -590,10 +591,14 @@ function setInCell(
...
 ): RecordOf<DocumentRecordProps> {
-  return state.setIn(
-    ["notebook", "cellMap", action.payload.id].concat(action.payload.path),
-    action.payload.value
-  );
+  const { id, path } = action.payload;
+  let value = action.payload.value;
+
+  if (path.length === 1 && path[0] === "source") {
+    value = normalizeLineEndings(value) ?? "";
```
bryanchen-d: nit: so if we are doing normalization anyway when writing cell source, should we also apply `demultiline` here too?

packages/commutable/src/structures.ts
barryt2: This change is still required because the cell creator methods you mention (containing normalization) are private to the creation of the notebook model. When a user calls public API makeCodeCell, makeMarkdownCell, etc., those are simple object structure creators with default values and do not have a chance to introduce extra logic such as normalization on them.

packages/reducers/src/core/entities/contents/notebook.ts
barryt2: Good observation. Updated.

packages/commutable/src/structures.ts
kycutler: Can we wrap the immutable factories? E.g. have a helper like:
```typescript
function normalizedSourceRecord<T extends { source?: string }>(recordFn: Record.Factory<T>): Record.Factory<T> {
  // Transparently wrap the factory, but overwrite the source with its normalized value
  return function (...args: Parameters<typeof recordFn>) {
    const res = recordFn.apply(this, args);
    return res.set("source", normalizeLineEndings(res.source));
  };
}
```

And then wrap the cell creation functions:
```typescript
const makeCodeCell = normalizedSourceRecord(Record({
  // ...
}));
```

barryt2: Updated. Moved normalization logic into the make cell functions.

bryanchen-d (approved):

kycutler (approved):

